PR #19, opened 2020-10-06
Newshell documentation

ret2libc:
Feedback is welcome. Of course, only way to see if these instructions work is to try to convert some commands :D


thestr4ng3r (commented):
I think there should also be a short guideline about how help messages should be formulated. In particular:
`Finds all occurrences` vs. `Find all occurrences` vs. `find all occurrences`. Imo it should always be `Find all occurrences`.

PR #28, opened 2020-10-12
Wrap summaries when printing cmd help ##newshell

ret2libc:
**Detailed description**

We may have long summaries(or small screens :D) and if you run `?` in those cases the summaries would go to a new line and they would make it hard to understand the help. This PR tries to improve the situation by wrapping the summary and printing it aligned.

By the way, if you prefer I can split things so that you can review the `rz_str_wrap` API separately from its usages. As you prefer.

**Test plan**

...

**Additional info**
Before:
image
After:
image
image


ITAYC0HEN (approved):
Loving it! The code looks good, and the screenshots as well.Likely worth that another experienced dev will go over it before. merging

wargio (approved):

PR #75, opened 2020-11-09
Initial version of newshell autocompletion

ret2libc:
This (quite big) PR adds support for autocompletion based on all the info available in "newshell".

Limitations of existing autocompletion:
- it is based on few assumptions about the structure of the input, which makes it very fragile and less useful (e.g. if you put a space in the first place, no autocompletion; if you use `$(` it won't detect that a new command can be suggested; etc.)
- it assumes only one argument should be autocompleted. It has no info about other arguments
- it requires manual changes every time a new command is added/removed and/or arguments accepted by a command are changed

Advantages of `cfg.newshell.autocompletion`:
- whenever possible it is based on tree-sitter syntax tree, so it can easily autocomplete new commands/new arguments everywhere one is expected (e.g. after `$(`, `@@c:`, `.`, etc.)
- it gets all the information about available commands/arguments from RzCmd, so whenever a new command/argument is added, it automatically supports it
- it supports more than one argument

Current limitations/issues:
- only file, env variables, zign spaces, and "choices" (see `za` `<type>` field) are autocompleted.
- commands that are not converted to "newshell" style are not autocompleted.
- as a command you are writing may not be valid yet while writing it, some sort of guessing is needed anyway to try to understand which token should be autocompleted (e.g. `zs "hell<tab>` is not valid according to the parser, because the `"` is not completed, so we need to try and see what is the proper thing to do).


XVilka (approved):
LGTM.

PR #91, opened 2020-11-16
Use cmd_descs_generate.py to update src dir and fallback if no PyYAML

ret2libc:
People with no PyYAML can still build everything, but they won't be able to update commands, of course. Also, the files in the src directory are automatically updated when you build with meson.

@thestr4ng3r I think this is a good tradeoff.

### DIFF
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -1,0 +1,20 @@
+name: "Mixed linter and checks"
+
+on:
+  push:
+  pull_request:
+  
+jobs:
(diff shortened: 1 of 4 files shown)


thestr4ng3r (approved):
Yep I think this is good. It makes developing slightly easier because you don't have to manually run the generator and removes the dependency.

But I think it's important that we also add a check in the ci to ensure the yaml and c/h are always consistent.

thestr4ng3r (approved):
Perfection.

PR #138, opened 2020-12-08
Use meson in FreeBSD/OpenBSD CI

ret2libc:


XVilka (approved):
I had to merge this quickly to prevent timeouts on SourceHut side. Was asked to prevent this in the future.

PR #140, opened 2020-12-09
Updated Dockerfile to follow the build instructions from rizin README.

0pendev:
Signed-off-by: Francisco Freitas <contact@ffreitas.io>

 

**Detailed description**

Looking at the Dockerfile I saw that it was not using the rizin build instructions to use meson. Another problem was that the project request a meson version that is not available in debian:10. I got the error message : 
`meson.build:1:0: ERROR:  Meson version is 0.49.2 but project requires >=0.50.1.`

I moved the image to OpenSUSE/leap as I usually build with this image.

I did not added rz-ghidra in the build because I did not know if it was usefull. But it could make sense to add it by default like in Cutter.

I removed the addition to the sudo group. If a user needs to install a new package he can just use the argument --user root when starting the container. If privileged are needed (for debugging for example) the user just needs to use the argument --privileged when starting the container. 

**Test plan**

I think I might need more help here. I simply ran a simple analysis on /bin/ls : 
image

**Closing issues**

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -31,13 +31,13 @@
 #
 
 # Using debian 10 as base image.
-FROM debian:10
+FROM opensuse/leap:15.2
 
 # Label base
 LABEL rizin latest
 
 # Radare version
-ARG RZ_VERSION=master
+ARG RZ_VERSION=dev
 # rz-pipe python version
 ARG RZ_PIPE_PY_VERSION=1.4.2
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -31,13 +31,13 @@
 #
 
 # Using debian 10 as base image.
-FROM debian:10
+FROM opensuse/leap:15.2
```
wargio: not super sure about the distro choice (just an issue regarding the image size)

wargio (approved):
Looks great! i before merging this i just want to check if for the others is ok this distro, but besides that, thank you :)

ret2libc (commented):
Thanks a lot! I was actually going to work on this but postponed until tomorrow... The way i bypassed the meson problem was to just install meson and ninja from pip, as that gets you an updated one. Opensuse could work as well, though probably most people are more familiar with Debian/Ubuntu than with suse

Dockerfile
XVilka: I think it's better to stick to Debian here. Or Ubuntu, at least.

XVilka (changes requested):

Dockerfile
thestr4ng3r: Even better alpine

0pendev: Debian is not an option as the meson version is not working. I'll check alpine to try to limit its size.

0pendev: Might try installing from pip as suggested by @ret2libc

ret2libc: Alpine with pip should work, yeah!

Dockerfile
```diff
@@ -60,48 +60,37 @@ VOLUME ["/mnt"]
...
+	apt-get install -y \
+	gcc \
+	git \
+	python3-pip \
+	${with_arm64_as:+binutils-aarch64-linux-gnu} \
+	${with_arm32_as:+binutils-arm-linux-gnueabi} \
+	${with_ppc_as:+binutils-powerpc64le-linux-gnu} && \
+	pip3 install meson ninja && \
+	cd /mnt && \
+	git clone https://github.com/rizinorg/rz-pipe && \
```
ret2libc: set `RZ_PIPE_PY_VERSION` default to `master` and use 
```suggestion
	git clone -b "$RZ_PIPE_PY_VERSION" https://github.com/rizinorg/rz-pipe && \
```

Dockerfile
```diff
@@ -60,48 +60,37 @@ VOLUME ["/mnt"]
...
 # Cleanup
-RUN DEBIAN_FRONTEND=noninteractive dpkg --add-architecture i386 && \
-  apt-get update && \
-  apt-get install -y \
-  curl \
-  wget \
-  gcc \
-  git \
-  bison \
-  pkg-config \
```
ret2libc, line 71: I think this is needed?

Dockerfile
0pendev: Forgot to add the branch. I'm adding it.

Dockerfile
0pendev: bison, pkg-config are not used when building with meson (or at leasts it builds and the rizin binary seems  ok).

I only install those packages :  gcc  git python3-pip.

Do you have a use case where bison and pkg-config are used to build when using meson ?

0pendev:
As people where not that happy with the base image change I went back to debian.


Alpine does not package binutils for the architectures specified in the Dockerfile.
```
/ # apk search binutils
binutils-gold-2.34-r1
mingw-w64-binutils-2.34-r0
binutils-avr-doc-2.34-r0
mingw-w64-binutils-doc-2.34-r0
binutils-doc-2.34-r1
binutils-dev-2.34-r1
binutils-2.34-r1
binutils-avr-2.34-r0
```

Dockerfile
ret2libc: Then it's good :) I thought it meson/ninja needed that, but probably it's just to build against the libraries.

wargio:
For me debian is ok!! Thank you again for this contrib

Dockerfile
ret2libc: Just saw the review request. Did you already add this?

wargio: @ret2libc search for `$RZ_VERSION`

ret2libc: No, that is different @wargio . This should use `RZ_PIPE_PY_VERSION`, not `RZ_VERSION`.

XVilka:
@ret2libc I think the docker CI job should be triggered on the change of Dockerfile and scripts that are used by it.

Dockerfile
0pendev: Sorry I thought I already added it. It's done now.

ret2libc (approved):
Thanks a lot!

ret2libc:
Merged!

0pendev:
> Merged!

Nice !

PR #144, opened 2020-12-09
Avoid reporting undefined error on FreeBSD

ret2libc:
On some platforms (e.g. FreeBSD) by default -Wl,--no-undefined is passed
to the compiler, thus when generating the util library there is an
error, because `environ` is defined only in an executable. Let's
automatically add a linker option to ignore those errors on FreeBSD.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -348,8 +348,8 @@ userconf.set10('WITH_GPL', true)
 ok = cc.has_header_symbol('sys/personality.h', 'ADDR_NO_RANDOMIZE')
 userconf.set10('HAVE_DECL_ADDR_NO_RANDOMIZE', ok)
 
-if not cc.compiles('extern char **environ;\nint main(int argc, char **argv) { (void)environ; return 0; }')
-  error('Environ is not found')
+if host_machine.system() == 'freebsd'
+  add_project_link_arguments('-Wl,--unresolved-symbols,ignore-in-object-files', language: 'c')
 endif
 
 lrt = []
--- a/.builds/freebsd.yml
+++ b/.builds/freebsd.yml
@@ -46,4 +46,5 @@ tasks:
         # https://todo.sr.ht/~sircmpwn/builds.sr.ht/274
         ln -s ${HOME}/rizin-testbins test/bins
         # Running the test suite
+        cd test
         rz-test -L -o results.json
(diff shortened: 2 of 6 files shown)


wargio (approved):

wargio:
failed again

yossizap (approved):

kazarmy:
1. Not sure whether you've noticed this, but Clang on OpenBSD is 10.0.1, but on FreeBSD is 8.0.1.
1. Try `__attribute__((weak_import))`?

ret2libc:
>     1. Not sure whether you've noticed this, but Clang on OpenBSD is 10.0.1, but on FreeBSD is 8.0.1.

Yeah I've noticed that! Indeed in my local FreeBSD env I have 10.0.1 and it worked. If possible, I'd like a solution that also works with older clang in FreeBSD!

XVilka:
Just use the flag. I think fixing tests is more important.

ret2libc:
> Just use the flag. I think fixing tests is more important.

I found the way to fix it without having to manually specify `-Dl_undef=false`. If possible I think it's really better for freebsd users to not have to add special flags. I tested it and a simple `meson build; ninja -C build` works well now!

XVilka (approved):
Amazing!

meson.build
```diff
@@ -348,8 +348,8 @@ userconf.set10('WITH_GPL', true)
 ok = cc.has_header_symbol('sys/personality.h', 'ADDR_NO_RANDOMIZE')
 userconf.set10('HAVE_DECL_ADDR_NO_RANDOMIZE', ok)
 
-if not cc.compiles('extern char **environ;\nint main(int argc, char **argv) { (void)environ; return 0; }')
-  error('Environ is not found')
+if host_machine.system() == 'freebsd'
+  add_project_link_arguments('-Wl,--unresolved-symbols,ignore-in-object-files', language: 'c')
```
kazarmy, line 352: Shouldn't you just use `--warn-unresolved-symbols`?

wargio: it is green!!!

ret2libc: I think that option is slightly different?

XVilka: Let's merge it as is, further improvements can be made separately. We need all CIs up and running before big changes come.

wargio:
@ret2libc can i squash?

PR #147, opened 2020-12-09
Version 0.1.0

wargio:
We still use the original r2 version number.
Let's start with the version 1.0.0-dev


ret2libc:
I think we said to go with just two numbers for the time being. So 1.0-git.

wargio:
ops, my bad.

ITAYC0HEN (approved):
Can consider to use team@rizin.re which might be accessible for more people in the future. but it is not critical as we don't expect many emails anyways

wargio:
> Can consider to use [team@rizin.re](mailto:team@rizin.re) which might be accessible for more people in the future. but it is not critical as we don't expect many emails anyways

done

XVilka (approved):
LGTM. While you are on it - can you please do the same for rz-ghidra, rz-pm, rz-pipe? cc @thestr4ng3r

configure
```diff
@@ -1,8 +1,5 @@
 #!/bin/sh
 # This script was automatically generated by ACR v1.9.4
-# @author: pancake <nopcode.org>
```
XVilka: These fields are generated by acr. Please just regenerate this file.

XVilka (changes requested):

configure
wargio: i do not know who to do that.

XVilka:
See also https://github.com/rizinorg/rizin/issues/174

wargio:
so we want to drop also `-git` ?

XVilka (approved):

ret2libc (approved):

PR #148, opened 2020-12-09
- is not a valid function name in bash

ret2libc:
Also, RZPM_PREFIX should not have the "/prefix" suffix


wargio:
i can't tell the difference in your commit.

wargio (approved):
good catch

PR #164, opened 2020-12-10
Use RPATH when the prefix is not the standard /usr

ret2libc:
This is to make it easier for people to install rizin without dealing
with LD_LIBRARY_PATH, while at the same time avoiding RPATH when things
are installed in the system-wide path (for most systems, at least). In
any case, people can disable this behaviour by using `-Dlocal=disabled`.

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -38,11 +38,12 @@ $ ninja -C build # or `meson compile -C build`
 $ sudo ninja -C build install # or `meson install -C build`
 ```
 
-As not all systems look for libraries in `/usr/local` subdirectories, you may
-have to set `LD_LIBRARY_PATH` to the proper path (e.g. `/usr/local/lib64` or
-`/usr/local/lib`). Otherwise, if you don't want to change your `LD_LIBRARY_PATH`
-you can use `meson -Dlocal=true build` in the first step to use `RPATH` and make
-sure the rizin binary can find its libraries by itself.
+NOTE: when `--prefix=/usr` is not used, meson will set `RPATH` to ensure that
+libraries can be found on the system without having to deal with
+`LD_LIBRARY_PATH` or ld settings. This is done to ensure a simple
+installation process out-of-the-box, but if you don't want this behaviour and
+you know what you are doing, you can still use `-Dlocal=disabled` to avoid
+using `RPATH`.
 
 ### Build system-wide, in `/usr`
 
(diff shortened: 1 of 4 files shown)


meson.build
```diff
@@ -562,9 +562,19 @@ if get_option('b_sanitize').contains('undefined')
   pkgcfg_sanitize_libs += ' -lubsan'
 endif
 
+use_rpath = false
+if host_machine.system() != 'windows'
+  if get_option('local').enabled() and get_option('default_library') == 'shared'
+    use_rpath = true
+  elif (get_option('local').auto() or get_option('local').enabled()) and get_option('prefix') != '/usr' and get_option('default_library') == 'shared'
+    use_rpath = true
```
kazarmy:
```suggestion
if host_machine.system() != 'windows' and get_option('default_library') == 'shared'
  if get_option('local').enabled() or (get_option('local').auto() and get_option('prefix') != '/usr')
    use_rpath = true
```

kazarmy (changes requested):

meson.build
ret2libc: Thanks

kazarmy (approved):

kazarmy:
Not sure every distro follows the LSB though.

ret2libc:
> Not sure every distro follows the LSB though.

Right. I will update the documentation specifying that for simplicity it sets RPATH in some cases and that they should use `-Dlocal=disabled` if they don't want RPATH. Those distros/systems that wants particular things can still do whatever they want, but at that point we don't need to provide documentation for them because they should know what they do.

kazarmy:
sgtm

ret2libc:
@kazarmy please check BUILDING.md now

BUILDING.md
```diff
@@ -74,28 +76,30 @@ $ ninja -C build install
 ```
 
 The `install` step will install rizin in `~/.local/bin`, so make sure to add it
-to your `PATH` variable. As most systems don't look for libraries in
-`~/.local/lib`/`~/.local/lib64`, you will have to set `LD_LIBRARY_PATH`
-accordingly or, if you prefer, use `meson -Dlocal=true --prefix=~/.local build`
-instead of just `meson --prefix=~/.local build`.
+to your `PATH` variable (e.g. `export PATH=$PATH:~/.local/bin`).
+
+NOTE: meson will set `RPATH` to ensure that libraries can be found on the
```
kazarmy, line 81:
```suggestion
NOTE TO DISTRIBUTION PACKAGERS: meson will set `RPATH` to ensure that libraries can be found on the
```

This is just a suggestion.

ret2libc: I would personally leave it for everyone, it's not really just for distro packegers. Is it ok if i leave it as-is?

kazarmy: I think so

PR #166, opened 2020-12-10
Reduce Coupling between RzCore and Tasks ##core

thestr4ng3r:
(More context in the comments)

This makes the low-level tasks code almost completely independent of RzCore. RzCore then puts a layer on top if the tasks to allow having a task-local cons context. This also introduces function tasks in addition to just command tasks.

### DIFF
--- a/librz/core/cmd_tasks.c
+++ b/librz/core/cmd_tasks.c
@@ -5,6 +5,10 @@
 #include "cmd_descs.h"
 
 static int task_enqueue(RzCore *core, const char *cmd, bool transient) {
+	if (rz_sandbox_enable (0)) {
+		eprintf ("This command is disabled in sandbox mode\n");
+		return -1;
+	}
 	RzCoreTask *task = rz_core_cmd_task_new (core, true, cmd);
 	if (!task) {
 		return -1;
@@ -33,6 +37,10 @@ static int task_output(RzCore *core, int tid) {
 }
 
 static int task_break(RzCore *core, int tid) {
+	if (rz_sandbox_enable (0)) {
+		eprintf ("This command is disabled in sandbox mode\n");
+		return -1;
+	}
 	if (!tid) {
 		return -1;
 	}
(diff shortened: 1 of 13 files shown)


ret2libc:
Are you removing sandbox feature from everywhere?
Also, why do you want to decouple them? Do you intend to move tasks outside of RzCore?

thestr4ng3r:
Yeah I think we should remove sandbox completely. I see no way of it being remotely safe.

Yes, the underlying task scheduling mechanism could be completely ripped out from core and moved to util. This will make it easier to get an overview over since task code is more self-contained, but more importantly, we can make other modules below core access the task api. This means we can for example move the breaking and yielding functionality completely from cons to tasks and have it local by passing a pointer to the scheduler to for example a complex analysis function to allow reentrant cooperative multitasking everywhere.

thestr4ng3r:
But in the short run, my plan is to just be able to run arbitrary functions in tasks instead of being forced to us a command. This will make it possible to reintroduce async rz-ghidra in cutter.

XVilka:
I think sandbox should be removed in a separate PR. Otherwise it's fine.

librz/core/cmd_tasks.c
```diff
@@ -40,7 +41,10 @@ static int task_break(RzCore *core, int tid) {
 		eprintf ("This command is disabled in sandbox mode\n");
 		return -1;
 	}
-	if (tid) {
+	if (!tid) {
```
XVilka, line 44: What about using `rz_return_val_if_fail()` instead?

librz/core/task.c
```diff
@@ -625,12 +526,316 @@ RZ_API int rz_core_task_del (RzCoreTaskScheduler *scheduler, int id) {
...
+	}
+	const char *cmd;
+	if (task == core->tasks.main_task) {
+		cmd = "";
+	} else {
+		cmd = ((CmdTaskCtx *)task->runner_user)->cmd;
+	}
+	switch (mode) {
+	case 'j':
+		{
```
XVilka: Can you please use PJ API while you are on it?

test/unit/test_task.c
```diff
@@ -0,0 +1,246 @@
...
+
+	LocalCtx a_ctx = {
+		.shared = &shared_ctx,
+		.id = 1,
+		.want_to_consume = 3
+	};
+	RzCoreTask *a = rz_core_task_new (&sched, runner, local_ctx_free, &a_ctx);
+	int a_id = a->id;
+	rz_core_task_enqueue (&sched, a);
+	a = NULL; // ownership moved to the scheduler, don't touch anymore!
```
XVilka, line 85: We will see this pattern often probably. Might be useful to create a macros or a function for this.

XVilka (changes requested):

librz/core/cmd_tasks.c
thestr4ng3r: It's not an assertion-level failure, the user could have passed 0, but that is the main task which should not be breaked.

librz/core/cmd_tasks.c
```diff
@@ -77,6 +81,9 @@ RZ_IPI RzCmdStatus rz_tasks_delete_handler(RzCore *core, int argc, const char **
 		return RZ_CMD_STATUS_ERROR;
 	}
 	int tid = rz_num_math (core->num, argv[1]);
+	if (!rz_core_task_is_cmd (core, tid)) {
+		return -1;
```
ret2libc:
```suggestion
		return RZ_CMD_STATUS_ERROR;
```

librz/core/cmd_tasks.c
```diff
@@ -98,6 +105,9 @@ RZ_IPI RzCmdStatus rz_tasks_wait_handler(RzCore *core, int argc, const char **ar
 	if (argc == 2) {
 		tid = rz_num_math (core->num, argv[1]);
 	}
+	if (!rz_core_task_is_cmd (core, tid)) {
+		return -1;
```
ret2libc:
```suggestion
		return RZ_CMD_STATUS_ERROR;
```

ret2libc (changes requested):
Nice thing overall!! I follow xvilka saying that it would be better to use pj. Also, here or in a very next PR it would be great to have doc for the other API functions and maybe an overall description of tasks as used in rizin (the fact it is collaborative, etc.)

librz/core/task.c
```diff
@@ -625,12 +526,316 @@ RZ_API int rz_core_task_del (RzCoreTaskScheduler *scheduler, int id) {
...
+		break;
+	}
+	}
+}
+
+RZ_API void rz_core_task_list(RzCore *core, int mode) {
+	RzListIter *iter;
+	RzCoreTask *task;
+	if (mode == 'j') {
+		rz_cons_printf ("[");
```
wargio: Same here. PJ should be used.

wargio (changes requested):

test/unit/test_task.c
thestr4ng3r: What pattern do you mean exactly?

ret2libc (approved):

PR #169, opened 2020-12-12
Remember project file after save and add Ps without args

thestr4ng3r:

**Detailed description**

After saving a project explicitly using `Ps file.rzdb` you can just type `Ps` to save to the same file again.

**Test plan**

run the tests

### DIFF
--- a/librz/core/cmd_project.c
+++ b/librz/core/cmd_project.c
@@ -3,9 +3,20 @@
 #include <rz_project.h>
 
 RZ_IPI RzCmdStatus rz_project_save_handler(RzCore *core, int argc, const char **argv) {
-	RzProjectErr err = rz_project_save_file (core, argv[1]);
+	const char *file;
+	if (argc == 1) {
+		file = rz_config_get (core->config, "prj.file");
+		if (!file || !*file) {
+			eprintf ("There is no project file associated with the current session yet.\n"
+				"Specify the file explitily as Ps <file.rzdb> or set it manually with e prj.file.\n");
+			return RZ_CMD_STATUS_ERROR;
+		}
+	} else { // argc == 2 checked by the shell
+		file = argv[1];
+	}
+	RzProjectErr err = rz_project_save_file (core, file);
 	if (err != RZ_PROJECT_ERR_SUCCESS) {
-		eprintf ("Failed to save project: %s\n", rz_project_err_message (err));
+		eprintf ("Failed to save project: %s to file %s\n", file, rz_project_err_message (err));
 	}
 	return RZ_CMD_STATUS_OK;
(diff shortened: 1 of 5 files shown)


librz/core/cmd_project.c
```diff
@@ -3,9 +3,20 @@
 #include <rz_project.h>
 
 RZ_IPI RzCmdStatus rz_project_save_handler(RzCore *core, int argc, const char **argv) {
-	RzProjectErr err = rz_project_save_file (core, argv[1]);
+	const char *file;
+	if (argc == 1) {
+		file = rz_config_get (core->config, "prj.file");
+		if (!file || !*file) {
```
ret2libc:
```suggestion
		if (RZ_STR_ISEMPTY (file)) {
```

librz/core/cmd_project.c
```diff
@@ -3,9 +3,20 @@
...
+			eprintf ("There is no project file associated with the current session yet.\n"
+				"Specify the file explitily as Ps <file.rzdb> or set it manually with e prj.file.\n");
```
ret2libc:
```suggestion
				"Specify the file explitily as Ps <file.rzdb> or set it manually with `e prj.file=<project-path>`.\n");
```

librz/core/cmd_project.c
```diff
@@ -3,9 +3,20 @@
...
+			return RZ_CMD_STATUS_ERROR;
+		}
+	} else { // argc == 2 checked by the shell
+		file = argv[1];
+	}
+	RzProjectErr err = rz_project_save_file (core, file);
 	if (err != RZ_PROJECT_ERR_SUCCESS) {
-		eprintf ("Failed to save project: %s\n", rz_project_err_message (err));
+		eprintf ("Failed to save project: %s to file %s\n", file, rz_project_err_message (err));
```
ret2libc: should you switch the arguments?

ret2libc (changes requested):

librz/core/cmd_project.c
thestr4ng3r: The format string was nonsense

ITAYC0HEN:
What about defaulting to save as `path/to/file/filename.rzdb` upon executing `Ps`? We can assume that most likely, people would like to save it with `[filename].rzdb`. IDA assumes it, Binja, and I think even Cutter. It is very probable, and can't hurt (?)

thestr4ng3r:
Yeah could be convenient. It will need a bit of extra logic to for example make sure not to overwrite any other file and to actually determine the binary filename, which is not always possible so should go into another pr.

ret2libc (approved):

PR #172, opened 2020-12-12
Fix a crash in dts+ command with empty register arenas ##debug

yossizap:

**Detailed description**

Some architectures may not support reading all `RzRegisterTypes` in their `reg_read` implementation. This caused `dts+` to crash the program when `rz_debug_add_checkpoint` attempted to copy b->size bytes(`rz_reg_arena_new` allocates a minimum of 1 byte) from an empty arena.

**Test plan**

Attempt to run `dts+` with a debug backend that doesn't support all arenas in `reg_read`.


ITAYC0HEN (approved):

librz/debug/dsession.c
```diff
@@ -63,6 +63,10 @@ RZ_API bool rz_debug_add_checkpoint(RzDebug *dbg) {
 	rz_debug_reg_sync (dbg, RZ_REG_TYPE_ALL, 0);
 	for (i = 0; i < RZ_REG_TYPE_LAST; i++) {
 		RzRegArena *a = dbg->reg->regset[i].arena;
+		// Ignore empty register arenas
+		if (!a->bytes) {
```
ret2libc: How can this happen? I see that there is a `rz_reg_arena_new` function and that can't return an arena with bytes == null. I think these kind of things shall be caught much before.

ret2libc (changes requested):

librz/debug/dsession.c
yossizap: When [`rz_reg_fit_arena`](https://github.com/rizinorg/rizin/blob/dev/librz/reg/arena.c#L134) is called after [register profile creation](https://github.com/rizinorg/rizin/blob/dev/librz/reg/profile.c#L237) the size is recalculated and freed if it's smaller than 1:
https://github.com/rizinorg/rizin/blob/4f51dd53514a4ae8ee7d694629ca3a2e3ca15fae/librz/reg/arena.c#L146-L153

and since [`rz_debug_reg_sync`](https://github.com/rizinorg/rizin/blob/dev/librz/debug/dreg.c#L8) receives size 0 when performing `reg_read` from the debug backend the arena will remain null:

https://github.com/rizinorg/rizin/blob/4f51dd53514a4ae8ee7d694629ca3a2e3ca15fae/librz/debug/dreg.c#L69-L72

> rz_reg_arena_new function and that can't return an arena with bytes == null

As I wrote in the description, that's what caused the segfault since b->size was defaulted to 1. I am in favor of making that new() return null for size 0 but I am not sure if something else might break. Do we want to look into it in this PR?

wargio: i think this is the way to go. i would still have a safe memcpy define or inline which we can use to also avoid this stuff..

ret2libc: IMO there are some invariants that we should try to maintain to make the code easy to understand and avoid all these `if`/`else` scattered everywhere. For example, an arena is created in `rz_reg_arena_new` with the assumption that, if it is created, `->arena` is allocated and it is at least `->size` bytes. Now, this condition should *always* be respected IMHO, so you don't have to check this everywhere. I think it makes sense to try and see if removing that `size < 1` check in `rz_reg_arena_new` works.

librz/reg/arena.c
```diff
@@ -168,15 +168,19 @@ RZ_API int rz_reg_fit_arena(RzReg *reg) {
 
 RZ_API RzRegArena *rz_reg_arena_new(int size) {
 	RzRegArena *arena = RZ_NEW0 (RzRegArena);
```
ret2libc, line 170: Add a rz_return_val_if_fail (size >=0)

librz/include/rz_types.h
```diff
@@ -333,6 +333,7 @@ static inline void *rz_new_copy(int size, void *data) {
 }
 // TODO: Make RZ_NEW_COPY be 1 arg, not two
 #define RZ_NEW_COPY(x,y) x=(void*)malloc(sizeof(y));memcpy(x,y,sizeof(y))
+#define RZ_MEMCPY_S(dest, dmax, src, smax) memcpy (dest, src, (smax < dmax) ? smax : dmax)
```
XVilka: Could you please add here an `#ifdef` with MSVC and newer GCC/Glibc (C11 should have support for `memcpy_s()`):
- https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/memcpy-s-wmemcpy-s?view=msvc-160
- https://en.cppreference.com/w/c/string/byte/memcpy

You will need to check for `__STDC_LIB_EXT1__ ` in case of GCC/Clang/Glibc

XVilka (changes requested):

librz/include/rz_types.h
XVilka: Moreover, we can make an inline function and put it in the utils instead. It will allow to check the types of the arguments, that `smax`, `dmax` should be size_t, not a signed integer, etc.

librz/reg/arena.c
```diff
@@ -168,15 +168,19 @@ RZ_API int rz_reg_fit_arena(RzReg *reg) {
...
+	}
+
+	arena->size = size;
+	if (size < 1) {
+		return arena;
+	}
+
+	if (!(arena->bytes = calloc (1, size + 8))) {
+		RZ_LOG_ERROR ("Failed to allocate arena bytes.\n");
+		RZ_FREE (arena);
```
wargio, line 183: if you freeing `arena`, you need to assign it to `NULL` or `return NULL`. otherwise we are gonna have a `use after free`

wargio (changes requested):

librz/reg/arena.c
```diff
@@ -168,15 +168,19 @@ RZ_API int rz_reg_fit_arena(RzReg *reg) {
...
-			arena->size = size;
-		}
+	if (!arena) {
+		RZ_LOG_ERROR ("Failed to allocate RzRegArena.\n");
+		return NULL;
+	}
+
+	arena->size = size;
+	if (size < 1) {
+		return arena;
```
wargio, line 178: i would suggest to also free and return NULL here.

wargio (changes requested):

librz/include/rz_types.h
ret2libc: Actually, instead of relying on checking various macros... Just use `meson` to check if the function is available. You can look at https://github.com/rizinorg/rizin/blob/dev/meson.build#L355 . That is imho the best way to see if some functions/features are available in the compiler.

librz/reg/arena.c
ret2libc: This is not an error. It's a valid use case, but if size is 0 (`size` should be of type `size_t`, to make sure it is always positive or 0) you don't want to allocate `arena->bytes` for nothing.

librz/reg/arena.c
ret2libc: `RZ_FREE` does free + assign NULL.

librz/reg/arena.c
yossizap: switched to size_t

librz/include/rz_types.h
yossizap: I stayed with regular `memcpy` in `rz_mem_copy` since `memcpy_s` has a different return value and we are going to set C99 anyways(?), I don't see a reason to split the implementation with ifdefs when `rz_mem_copy` does the same thing.

librz/debug/dsession.c
```diff
@@ -64,7 +64,9 @@ RZ_API bool rz_debug_add_checkpoint(RzDebug *dbg) {
 	for (i = 0; i < RZ_REG_TYPE_LAST; i++) {
 		RzRegArena *a = dbg->reg->regset[i].arena;
 		RzRegArena *b = rz_reg_arena_new (a->size);
-		memcpy (b->bytes, a->bytes, b->size);
+		if (a->size) {
```
ret2libc: is this check really needed? I think rz_mem_copy just do nothing if a->size = 0, right?

yossizap: You're right but I don't want to trigger assertions from `rz_mem_copy`(and those should stay there since dest/src could be NULL with size >0 somewhere else in the future)
 
```
WARNING: rz_mem_copy: assertion 'dest && src' failed (line 46)                                                        
WARNING: rz_mem_copy: assertion 'dest && src' failed (line 46)                                                        
```

librz/util/mem.c
```diff
@@ -36,6 +36,11 @@ RZ_API void rz_mem_copyloop(ut8 *dest, const ut8 *orig, int dsize, int osize) {
 	}
 }
 
+RZ_API void *rz_mem_copy(void *dest, size_t dmax, const void *src, size_t smax) {
+	rz_return_val_if_fail (dest && src, NULL);
```
ret2libc, line 43:
```suggestion
	rz_return_val_if_fail ((smax == 0 || dmax == 0) || (dest && src), NULL);
```

ret2libc: Also, after this  you could just add an `if (smax == 0 || dmax == 0)` to skip the `memcpy` call at all, if you want. By doing this we don't need to have an `if (...)` at every place that calls `rz_mem_cpy`

yossizap: The suggested change will print an unnecessary assertion even though smax/dmax=0 doesn't warrant a warning, added a simple if on entry.

ret2libc: yeah :D

ret2libc (approved):
Looks good to me now! Let's see tests.
I guess you tested it manually as well and debugging still works, right?

yossizap:
> I guess you tested it manually as well and debugging still works, right?

Yep with native Linux and gdbserver

XVilka (approved):
LGTM. We should probably to fish out also other places where `memcpy()` called and use our `rz_mem_copy()` instead.

ret2libc:
@wargio please review again as you requested changes. If it's ok for you, I think you can directly merge.

wargio (approved):

PR #176, opened 2020-12-13
Rename more Block API and remove hard-coded x86 logic ##analysis

thestr4ng3r:
Before these changes, when you called `afb+` to manually add a block, it would create the block and then call `fcn_recurse()` with a depth of 1 on it. This also includes variable analysis and xrefs which is why they are in the edited tests.

However, `fcn_recurse()` was ONLY hard-coded for x86 there, for any other arch you would get nothing set here. The actual reason why it was called there was only to get the op offsets in the block set up, which is especially important for x86 but also for other archs, but `fcn_recurse()` is a total overkill for that. Now there is the function `rz_analysis_block_analyze_ops()` which does what we want here and nothing else.
It makes sense to explicitly do nothing else in `afb+` because it is meant to be a relatively low-level operation. Setting the op offsets is necessary because there is no other way for the user to get them set and they are more of an internal cache. Other things like variables and xrefs can instead be done with dedicated commands.

In addition to that, I did a the renaming any remaining `bb`-style api to `block` so `RzAnalysisBlock` api can be considered clean and more or less stable now. (Functions are another story)


wargio:
you seems to have broken some stuff

wargio:
we need to understand what was that logic needed for.

thestr4ng3r:
I know, the whole code is a mess. I have a partial refactor locally but it's not ready yet.

librz/analysis/block.c
```diff
@@ -952,3 +952,169 @@ RZ_API void rz_analysis_block_automerge(RzList *blocks) {
...
+}
+
+/**
+ * Successively disassemble the ops in this block, update the contained op addrs
+ * and create xrefs if necessary.
+ * This will not move or resize the block itself or touch anything else around it,
+ * it is primarily useful when creating or editing blocks after full function analysis.
+ */
+RZ_API void rz_analysis_block_analyze_ops(RzAnalysisBlock *block) {
+	rz_return_if_fail (block);
```
wargio, line 1069: should you check if blocksize is > 0 ?

test/db/cmd/cmd_pd2
```diff
@@ -158,16 +153,16 @@ EXPECT=<<EOF
...
+|       :   0x00560f95      mov dword [esp + 0x10], ebx
 |       :   0x00560f99      push dword [esp]
 |       :   0x00560f9c      pop ebx
 |       :   0x00560f9d      push ebx
 |       :   0x00560f9e      mov ebx, esp
 |       :   0x00560fa0      add ebx, 4
 |       :   0x00560fa6      add ebx, 4
 |       :   0x00560fa9      push ebx
-|       :   0x00560faa      push dword [var_1ch]
+|       :   0x00560faa      push dword [esp + 4]
```
wargio, line 165: is this a regression?

librz/analysis/block.c
thestr4ng3r: Not in the assert, but yeah that wouldn't hurt.

test/db/cmd/cmd_pd2
thestr4ng3r: Good question. I've added some stuff to the description since the reason basically describes the whole pr.

XVilka (approved):
Awesome.

kazarmy:
> It makes sense to explicitly do nothing else in afb+ because it is meant to be a relatively low-level operation. Setting the op offsets is necessary because there is no other way for the user to get them set and they are more of an internal cache. Other things like variables and xrefs can instead be done with dedicated commands.

This is fine I think.

The real problem with afb+ is that you can define a basic block that isn't a valid basic block (e.g. the basic block straddles a jump, or it ends in the middle of an instruction etc.). Don't know what happens if you do that, but I don't think it's going to be pretty.

wargio (approved):
looks great to me

thestr4ng3r:
> > It makes sense to explicitly do nothing else in afb+ because it is meant to be a relatively low-level operation. Setting the op offsets is necessary because there is no other way for the user to get them set and they are more of an internal cache. Other things like variables and xrefs can instead be done with dedicated commands.
> 
> This is fine I think.
> 
> The real problem with afb+ is that you can define a basic block that isn't a valid basic block (e.g. the basic block straddles a jump, or it ends in the middle of an instruction etc.). Don't know what happens if you do that, but I don't think it's going to be pretty.

This is perfectly fine, it is up to the caller to determine if what they set actually makes any sense. `afb+` also takes explicit arguments for the jump targets. So what you set is what you will see.

PR #182, opened 2020-12-14
Fix external AppVeyor tests that are XX

kazarmy:
This is the same pr as #180 but with the branch name renamed from `fix-external-xx-windows` to `windows-fix-external-xx`.

PR #180, opened 2020-12-14
Fix external AppVeyor tests that are XX

kazarmy:
**Your checklist for this pull request**
- [X] I've read the [guidelines for contributing](https://github.com/rizinorg/rizin/blob/master/DEVELOPERS.md) to this repository
- [X] I made sure to follow the project's [coding style](https://github.com/rizinorg/rizin/blob/master/DEVELOPERS.md#code-style)
- [X] I've added tests that prove my fix is effective or that my feature works (if possible)
- [ ] I've updated the documentation and the [rizin book](https://github.com/rizinorg/rizinbook) with the relevant information (if needed)

**Detailed description**

The pr fixes the XX AppVeyor tests by rewriting them using internal commands and marking the external version as broken. Also moved the broken `rz-bin` tests into `tools/rz_bin`.

@ret2libc I'd be happy to fix #110 so that it works with this pr. AppVeyor should be green asap though, since it is a pain to check the AppVeyor logs to see whether there is additional redness.

**Test plan**

All builds are green, including AppVeyor.

**Closing issues**

...

PR #183, opened 2020-12-14
Fix external AppVeyor tests that are XX

kazarmy:
This is the same pr as #182 but with the host repository being `rizinorg` instead of `kazarmy`.


kazarmy:
Closed because #180 was merged.

PR #184, opened 2020-12-14
Check if submodules are downloaded and warn user if they are not

ret2libc:
Similar to how this is done in libvirt/meson.build. As people may often
forget to clone with recurse-submodules, this saves them the trouble of
getting an error and understanding what's wrong.

**Test plan**

Test with .git dir and without, with -Dno_git=true and without.


karliss:
I am not a fan of build system messing with version control unless manually enabled or doing anything network related.  How about just adding a sanity check and printing a error message suggesting that user probably missed a step. Something like "Source for x not found, did you forgot to run 'git foo`?". It usually needs to be done just once so having to do the step manually after receiving a reminder isn't as bad. The check might be even doable by just checking folders for files without touching git.

thestr4ng3r:
Agree with @karliss for the reasons he mentioned. Also, it especially should NOT run `update` since I might have checked out a different commit on purpose and then the build system would mess that up again.

XVilka:
It might be good to have but certainly should be not enabled by default.

ret2libc:
Ok thanks for the feedback. Please have a look at it now.

thestr4ng3r:
Warning sgtm in case we decide not to use subprojects/wrap.

karliss:
Descripton and possibly name for the option needs to be updated now.

meson.build
```diff
@@ -49,6 +49,12 @@ gittap = ''
 gittip = 'unknown'
 
 if git_exe.found()
+  git_dir_exists = run_command(py3_exe, '-c', '__import__("sys").exit(0 if __import__("os").path.isdir(".git") else 1)')
+  sdb_src_exists = run_command(py3_exe, '-c', '__import__("sys").exit(0 if not __import__("os").listdir("shlr/sdb/") else 1)')
+  if git_exe.found() and not get_option('no_git') and git_dir_exists.returncode() == 0 and sdb_src_exists.returncode() == 0
```
thestr4ng3r: `git_exe.found()` already checked before

XVilka:
I suppose it will be superseded by https://github.com/rizinorg/rizin/pull/191

ret2libc:
> I suppose it will be superseded by #191

Well, not exactly now... Because ACR/Make needs sdb, capstone, etc. in the shlr directories. So considering we still support it, we can't really move away from submodules :(

ret2libc:
I think this is ready for review.

thestr4ng3r (approved):

PR #186, opened 2020-12-14
Finish removing www related content

xarkes:
I think these were remains from https://github.com/rizinorg/rizin/pull/63.


XVilka:
It breaks http tests and precisely why I didn't remove it from the beginning. We need to decide what to do with HTTP pipes first. See also https://github.com/rizinorg/rizin/issues/168

XVilka:
Please keep the empty index.html for now and it should be good to merge.

xarkes:
Ah okay, I was willing to remove this useless `www` folder but if that's useful for some test I think we can just close this PR for the moment.

wargio (approved):

Maijin:
And lot still there check:

* WWWROOT
* httpui
* shlr/www
* use_webui

ret2libc:
Just to understand... What to do with this? Is www stuff still needed or not?

xarkes:
From my PoV it is not, hence this PR.
For some reason a test requires this folder I believe, although I did not check if the test was still relevant or not...
This is the failing test: `[XX] db/archos/linux-x64/http =h` -> `RZ_NOPLUGINS=1 rizin -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -Qc !scripts/test-webserver.sh --`
With the corresponding file: https://github.com/rizinorg/rizin/blob/dev/test/scripts/test-webserver.sh

I can also remove them from my curent PR if you think that's the thing to do.

ret2libc:
Ok I've looked a bit at it, and it seems that the index.html is really needed only for the `curl` command that is used to check if the web server is up and running. I think that the `curl` command can be replaced with `netcat` or something like that, to check if the connection is open (or similar).

Also, is https://github.com/rizinorg/rizin/blob/dev/test/scripts/test-webserver.bat even used in https://github.com/rizinorg/rizin/blob/dev/test/db/archos/windows-x64/http ? That file still references r2, so I'm not sure if/how it works :/

IMO we can remove these things.

ret2libc:
Oops, I probably written things in the wrong way :D I meant we can probably remove the `www` things, but the `=h` test could have been changed a bit to use `nc`/`netcat` instead of `curl` to test if the server was up.

Anyway, also removing everything is fine for me because I really don't like how the whole "remote"/"rtr" part is done.

XVilka:
Lets wait for AppVeyor first, just in case.

Maijin:
@xarkes  - You are also missing the `http.` eval variables in cconfig and associated code - a lot to trim there as well

xarkes:
I can do it but probably in a separate PR.

Maijin:
Sounds good to me 👍

XVilka:
Could you please open a new PR and close this one? Otherwise AppVeyor seems not triggered. Then if green - lets merge it.

XVilka:
I should note, we should test if `http://` rz-pipe still works after this removal.

PR #193, opened 2020-12-15
Switch to ubuntu-20.04 to fix gcov issue

ret2libc:
Reading a bit online, it seems some solved the gcov issue by using a newer gcc. So this PR moves our github CI to ubuntu-20.04. THis change was anyway going to happen soon because github CI is moving ubuntu-latest from 18.04 to 20.04.

It seems it actually makes the gcov job much faster and I don't see the problem about "merge function 0".


XVilka:
Please also test Windows with AppVeyor.

ret2libc:
> Please also test Windows with AppVeyor.

It's not necessary. I've only touched files that are not considered by appveyor

XVilka (approved):
It's great.
Interestingly, there seem to be a variable size bug somewhere in the debugging code:
```
0x00000000
0x00400590
0x00000003
0x00400590
0xfffffffffffff30e
0x00000000
0x0040052f
```

ret2libc:
> It's great.

It is! It makes linux-meson-gcc job much faster!

> Interestingly, there seem to be a variable size bug somewhere in the debugging code:
> 
> ```
> 0x00000000
> 0x00400590
> 0x00000003
> 0x00400590
> 0xfffffffffffff30e
> 0x00000000
> 0x0040052f
> ```

Yeah... I have just "fixed" the test for now because I wanted to see how the jobs behaved, but this should be fixed, if it is an actual problem.

kazarmy (approved):
Why is this still in draft?

ret2libc:
> Why is this still in draft?

I need to *really* fix the debug tests.

kazarmy:
Don't be surprised though to find that each debug test fix is an adventure in itself.

test/db/archos/linux-x64/dbg_step_back
```diff
@@ -14,7 +14,7 @@ dk 9
 EOF
 EXPECT=<<EOF
 0x00000001
-0x00000001
+0x736c6974756572
```
ret2libc, line 17: checked this on my machine, fedora 32, and this value is 0. So i assume it depends on the system. The other values remain the same. Maybe a better test would be to just check rbx, r12 and rip.

test/db/archos/linux-x64/dbg_step_back
```diff
@@ -37,16 +37,16 @@ dk 9
 EOF
 EXPECT=<<EOF
 0x00400574
-0x00000000
 0x00400590
-0x00000003
 0x00400590
+0xfffffffffffff30e
```
ret2libc, line 42: these values are simply not stable, as they may contain whatever data... so I'd say it is alright to just adjust the tests. In the long term we probably want to have better way to test debug tests.

PR #195, opened 2020-12-15
Add -std=c99, for older gcc ##build

mscherer:
While modern compilers seems to default to c99, GCC 4.8,
the version on RHEL 7 don't, despites supporting C99.

 

- [X] I've updated the documentation and the [rizin book](https://github.com/rizinorg/rizinbook) with the relevant information (if needed)

**Detailed description**

From time to time, I build radare2 on my laptop running RHEL 7. I often found that things didn't compiled because gcc 4.8 default to C89 and not C99, despites supporting it. I never found where to fix that in r2 build system, but since rizin aim to clean that, I spent time to fix the root cause in meson.

The C99 code show this error:
```
FAILED: librz/flag/librz_flag.so.0.1.0-git.p/serialize_flag.c.o 
cc -Ilibrz/flag/librz_flag.so.0.1.0-git.p -I. -I.. -I../librz/include -I../shlr/spp -I../shlr/sdb/src -I../shlr/zip/zlib -pipe -D_FILE_OFFSET_BITS=64 -Wall -Winvalid-pch -g -fvisibility=hidden -D_GNU_SOURCE -fPIC -pthread -DRZ_PLUGIN_INCORE=1 -MD -MQ librz/flag/librz_flag.so.0.1.0-git.p/serialize_flag.c.o -MF librz/flag/librz_flag.so.0.1.0-git.p/serialize_flag.c.o.d -o librz/flag/librz_flag.so.0.1.0-git.p/serialize_flag.c.o -c ../librz/flag/serialize_flag.c
In file included from ../librz/flag/serialize_flag.c:6:0:
../librz/flag/serialize_flag.c: In function ‘flag_load_cb’:
../librz/flag/../util/serialize_helper.h:36:3: erreur: ‘for’ loop initial declarations are only allowed in C99 mode
   for(RJson *child = json->children.first; child; child = child->next) { \
   ^
../librz/flag/serialize_flag.c:174:2: note: in expansion of macro ‘KEY_PARSER_JSON’
  KEY_PARSER_JSON (ctx->parser, json, child, {
  ^
../librz/flag/../util/serialize_helper.h:36:3: note: use option -std=c99 or -std=gnu99 to compile your code
   for(RJson *child = json->children.first; child; child = child->next) { \
   ^
../librz/flag/serialize_flag.c:174:2: note: in expansion of macro ‘KEY_PARSER_JSON’
  KEY_PARSER_JSON (ctx->parser, json, child, {
```

**Test plan**

Test plan include "look at the CI" and see if things still build. The current build git HEAD do not work on RHEL 7 due to the git command to reset capstone clone not supporting -C, but I will also take a look at it (for now, I just bypassed it).

**Closing issues**

...


thestr4ng3r (changes requested):
The correct way to do this is to set the meson `c_std` option in the `project()` call, see https://mesonbuild.com/howtox.html#set-default-cc-language-version

However, iirc in r2 we decided explicitly to support C89 so the error you see would be considered a but should be fixed in code instead of using C99. Maybe we would like to reconsider for rizin, but fixing it would be trivial too. Summoning @ret2libc @XVilka @wargio for discussion.

ret2libc:
I am not sure r2 has compiled without c99 support in the last 5 years, at least. In particular, https://github.com/radareorg/sdb/blob/master/config.mk#L34 was introduced about 5 years ago. Moreover, right now to compile tree-sitter, IIRC, we need gnu99/c99 at least. Last, right now there are variable declarations interleaved with regular statements, which is not standard c90 anyway. 

So, IMHO we can switch to c99.

thestr4ng3r:
Yeah if tree sitter needs it then I think there is no reason not to switch to C99 for everything.

ret2libc:
https://github.com/rizinorg/rizin/commit/48c30dfd99da9465fb3e4c68b5b4e43969ca2ee8 this was required to compile tree-sitter on RHEL 6.

wargio:
TBH i would be against that `for(int ....)` but i understand the need of that so i will agree on using it by default.

wargio (approved):

XVilka:
We use C99 in some places for some time, so I suggest it should be the minimum required version. Maybe even specify this in the building instructions document.

XVilka:
Can we enforce the feature for all compilers? Not just GCC.

thestr4ng3r:
Yes so we eventually decided to specify C99 as the C version for rizin. @mscherer please update the pr according to the link I sent above. @XVilka This will then automatically apply to all compilers.

meson.build
```diff
@@ -1,4 +1,4 @@
-project('rizin', 'c', license: 'LGPL3', meson_version: '>=0.55.0')
+project('rizin', 'c', license: 'LGPL3', meson_version: '>=0.50.0', default_options: ['c_std=c99'])
```
ret2libc: no please, keep 0.55.0, because there are things we use that require at least 0.54.something and we are going to switch to meson subprojects soon, which require  0.55.0 anyway.

ret2libc (changes requested):

meson.build
mscherer: oups, wrong merge, sorry

ret2libc (approved):

ret2libc:
It seems this discovered some "problems" within SDB :(

ret2libc:
This needs first appveyor to be sure it works there.

XVilka:
Lets close this one then in favor of https://github.com/rizinorg/rizin/pull/250

PR #201, opened 2020-12-16
Remove folding

Maijin:

**Detailed description**

Removing folding feature that's adding complexity in core .c files (disasm.c, visual.c)

### DIFF
--- a/librz/analysis/serialize_analysis.c
+++ b/librz/analysis/serialize_analysis.c
@@ -13,13 +13,13 @@
  *
  * /
  *   /blocks
- *     0x<addr>={size:<ut64>, jump?:<ut64>, fail?:<ut64>, traced?:true, folded?:true, colorize?:<ut32>,
+ *     0x<addr>={size:<ut64>, jump?:<ut64>, fail?:<ut64>, traced?:true, colorize?:<ut32>,
  *               fingerprint?:"<base64>", diff?: <RzAnalysisDiff>, switch_op?:<RzAnalysisSwitchOp>,
  *               ninstr:<int>, op_pos?:[<ut16>], stackptr:<int>, parent_stackptr:<int>,
  *               cmpval:<ut64>, cmpreg?:<str>}
  *   /functions
  *     0x<addr>={name:<str>, bits?:<int>, type:<int>, cc?:<str>, stack:<int>, maxstack:<int>,
- *               ninstr:<int>, folded?:<bool>, pure?:<bool>, bp_frame?:<bool>, bp_off?:<st64>, noreturn?:<bool>,
+ *               ninstr:<int>, pure?:<bool>, bp_frame?:<bool>, bp_off?:<st64>, noreturn?:<bool>,
  *               fingerprint?:"<base64>", diff?:<RzAnalysisDiff>, bbs:[<ut64>], imports?:[<str>], vars?:[<RzAnalysisVar>],
  *               labels?: {<str>:<ut64>}}
  *   /xrefs
@@ -254,9 +254,6 @@ static void block_store(RZ_NONNULL Sdb *db, const char *key, RzAnalysisBlock *bl
 	if (block->traced) {
 		pj_kb (j, "traced", true);
 	}
-	if (block->folded) {
-		pj_kb (j, "folded", true);
-	}
 	if (block->colorize) {
 		pj_kn (j, "colorize", (ut64)block->colorize);
 	}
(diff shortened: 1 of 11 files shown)


XVilka:
I am not sure about this one - folding is useful in the visual interface.

Maijin:
imho It's adding complexity, it's not working, tested properly and almost no users knows about it. I think may be good to add it back later once disasm.c/visual.c is decluttered.

thestr4ng3r:
Personally, I am ok with removing since I never used this. Don't know about others though.

XVilka:
I just checked with:
```
rizin /bin/ls
[0x00006b60]> aaa
[0x00006b60]> afF
[0x00006b60]> Vp
```
And it works:

image

Maijin:
Yeah but in visual as soon as there is more complexity everytime there is a new addition, folding needs to be taken in the code/display. It impacts scrolling/mouse wheels gets sluggish, visual display etc. You can see the number of files impacted, where it should barely touch a couple, imho not good enough to be part of a cleaned-up disasm.c / visual.c.

kazarmy:
Perhaps only basic blocks should be folded, and a folded function is a special case of a folded superblock? Anyway, it would be great if we could see broken folding examples.

thestr4ng3r:
I think instead of folding individual blocks and functions, it might make more sense to be able to make groups of blocks and then fold/unfold such groups. So you can get a more high-level graph with maybe some complex parts being collapsed into a single folded block. That would indeed require rewriting this from scratch.

Maijin:
Just an example - https://asciinema.org/a/J4gA4dvXbwDWfp8GoLZXOrwYT folding a function with z, then it's showing weird .int32 and scroll seems broken afterwards :/

wargio:
tbh, looks really weird.

kazarmy:
Yeah, the code probably needs an answer to the question "What happens if you seek into the middle of a folded block/function?" among others.

thestr4ng3r:
Considering all of this, I think it's better to remove.

ret2libc:
I agree with removing this. As @XVilka said, it works, yes, but as soon as you do any other thing it's kinda unusable imo. So +1 from me.

XVilka (approved):
LGTM to remove then.  I created [an issue](https://github.com/rizinorg/rizin/issues/203) for implementing this feature from scratch. Let's merge this PR.

XVilka:
Just please fix this warning/error:
```
../librz/core/visual.c: In function ‘rz_core_visual_disasm_down’:
../librz/core/visual.c:4056:22: error: variable ‘f’ set but not used [-Werror=unused-but-set-variable]
  RzAnalysisFunction *f = NULL;
                      ^
```

librz/core/visual.c
```diff
@@ -4112,26 +4056,22 @@ RZ_API void rz_core_visual_disasm_down(RzCore *core, RzAsmOp *op, int *cols) {
 	RzAnalysisFunction *f = NULL;
 	f = rz_analysis_get_fcn_in (core->analysis, core->offset, 0);
```
wargio: i think you need to remove this and the previous line.

wargio (changes requested):

librz/analysis/serialize_analysis.c
```diff
@@ -13,13 +13,13 @@
  *
  * /
  *   /blocks
- *     0x<addr>={size:<ut64>, jump?:<ut64>, fail?:<ut64>, traced?:true, folded?:true, colorize?:<ut32>,
```
ret2libc, line 16: @thestr4ng3r I think we can consider the project format changeable until 1.0.0, right?

librz/core/agraph.c
```diff
@@ -4362,18 +4347,6 @@ RZ_API int rz_core_visual_graph(RzCore *core, RzAGraph *g, RzAnalysisFunction *_
 				agraph_toggle_callgraph (g);
 			}
 			break;
-		case 'Z':
```
ret2libc, line 4365: Actually.. Is there anything wrong with the folding on the ascii-graph? I think it works well enough... or not?

ret2libc (approved):

librz/core/agraph.c
Maijin: @ret2libc Doesn't work see:

image

ret2libc: Did you press z while in graph mode or while in visual mode?

Maijin: while in graph mode

ret2libc: ahhh sorry, I was doing the `z`, not `Z`. Ok then!

PR #207, opened 2020-12-16
Use async-signal-safe functions after fork in multithread program

ret2libc:
From the commit description:   

    Do not use mutexes after fork nor setenv
    
    If pipe2 is not available on a system, we try to simulate the atomic
    setting of O_CLOEXEC in different ways (depending on whether O_CLOEXEC
    is at least available or not). However this involves using mutexes. When
    forking, only the calling thread is forked, so it is important to lock
    the mutex before the fork. However, in the child between the fork and
    the exec* it is recommended to use only async-safe functions. mutex
    operations are not async-safe, so we just keep track of whether we are
    in the child or not and avoid using the mutex as necessary.


thestr4ng3r:
For clarification: It's not just async-safe but async-signal-safe. See `man pthread_atfork` for example.

librz/util/subprocess.c
```diff
@@ -562,6 +571,17 @@ RZ_API RzSubprocess *rz_subprocess_start(
...
+	// save current environment and update it with the provided one.
+	// Do this before rz_sys_fork, because after that only async-safe functions
+	// shall be used and setenv is not.
+	HtPP *o_env = ht_pp_new (env_val_dup, env_kv_free, NULL);
+	size_t i;
+	for (i = 0; i < env_size; i++) {
+		char *val = getenv (envvars[i]);
+		ht_pp_insert (o_env, envvars[i], val);
+		rz_sys_setenv (envvars[i], envvals[i]);
+	}
```
thestr4ng3r: I don't think we should do this before the fork if possible since then for a short amount of time you will mess up the parent process's env. For example if we later start a debugee with the subprocess api, it would leak the child env into the parent.
Some alternatives:
* Something like https://phabricator.kde.org/R285:bb7c80717a60be7ae00cdf7e05c787217d4403c0 with `execvpe()` though that is a GNU extension but `execle()` is not.
* Maybe manually setting `environ` directly

librz/util/sys.c
```diff
@@ -1456,9 +1456,15 @@ RZ_API int rz_sys_pipe_close(int fd) {
 // Use this lock to wraps pipe, close, exec*, system to ensure all pipe file
 // descriptors are either created AND set as CLOEXEC or not created at all.
 static RzThreadLock *sys_pipe_mutex;
+bool is_child = false;
```
thestr4ng3r:
```suggestion
static bool is_child = false;
```

librz/util/sys.c
```diff
@@ -1456,9 +1456,15 @@ RZ_API int rz_sys_pipe_close(int fd) {
...
+
+static void set_child(void) {
+	is_child = true;
+}
 
 __attribute__ ((constructor)) static void sys_pipe_constructor(void) {
 	sys_pipe_mutex = rz_th_lock_new (true);
+	pthread_atfork (NULL, NULL, set_child);
```
thestr4ng3r: Does this still work in threads that are not the main thread? From the man page:
```
The pthread_atfork() function registers fork handlers that are to be executed when fork(2) is called by this thread.
```

ret2libc: Mmmm good point... I need to investigate, it seems it doesn't from what you say.

librz/util/subprocess.c
thestr4ng3r: I think setting `environ` manually might be the best solution. For a bit of context, I took a quick look at glibc code of `setenv()`: https://sourceware.org/git/?p=glibc.git;a=blob;f=stdlib/setenv.c;h=c6c6d02fff8d0a172f63869ddb47109cb4f19e63;hb=HEAD#l116 It uses a lock so that would be directly an example where a `setenv()` implementation can deadlock after `fork()`. Basically what we would have to do is write something similar as this `__add_to_environ()` but without locking.

ret2libc: What about just using `execve`?

librz/util/sys.c
ret2libc: The "solution", since we have a `rz_sys_fork` anyway, could be to just implement the `pthread_atfork` logic in `rz_sys_fork`. That should work.

librz/util/subprocess.c
thestr4ng3r: `execve` takes a raw file path while `execvp*` also take into account the `PATH`. Depends of course on what is desired but I for rz-test at least it would be the latter.

ret2libc:
@thestr4ng3r please have a look at the current solution. The only possible problem I see here is that by using alloca it could be a problem for environments where the stack is small. WDYT?

thestr4ng3r (approved):
Looks very nice now! Hopefully we haven't forgotten any other weird issue.

PR #212, opened 2020-12-17
Don't check reg a0 if analysis plugin is null

kazarmy:

**Detailed description**

This pr fixes the following problem on asan (https://github.com/rizinorg/rizin/runs/1564668218#step:17:166):

null-reg-a0

**Test plan**

All builds are green, including the asan build.

**Closing issues**

...

### DIFF
--- a/librz/core/cbin.c
+++ b/librz/core/cbin.c
@@ -738,11 +738,16 @@ RZ_API void rz_core_analysis_type_init(RzCore *core) {
 }
 
 RZ_API void rz_core_analysis_cc_init(RzCore *core) {
-	const char *dir_prefix = rz_config_get (core->config, "dir.prefix");
 	const char *analysis_arch = rz_config_get (core->config, "analysis.arch");
-	int bits = core->analysis->bits;
 	Sdb *cc = core->analysis->sdb_cc;
+	if (!strcmp (analysis_arch, "null")) {
+		sdb_reset (cc);
+		RZ_FREE (cc->path);
+		return;
+	}
 
+	const char *dir_prefix = rz_config_get (core->config, "dir.prefix");
+	int bits = core->analysis->bits;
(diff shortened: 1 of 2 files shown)


librz/core/cbin.c
```diff
@@ -738,11 +738,16 @@ RZ_API void rz_core_analysis_type_init(RzCore *core) {
 }
 
 RZ_API void rz_core_analysis_cc_init(RzCore *core) {
-	const char *dir_prefix = rz_config_get (core->config, "dir.prefix");
 	const char *analysis_arch = rz_config_get (core->config, "analysis.arch");
-	int bits = core->analysis->bits;
 	Sdb *cc = core->analysis->sdb_cc;
+	if (!strcmp (analysis_arch, "null")) {
+		sdb_reset (cc);
```
ret2libc, line 744: why this? I don't see similar exit code in the other paths of this function.

kazarmy: the null analysis plugin won't have a calling convention ever so might as well exit early ... or are you saying i don't need to `sdb_reset()`?

kazarmy: if i'm not mistaken , `sdb_reset()` clears the cc sdb, basically saying that there's no calling conventions defined, which sounds reasonable to me

ret2libc: The latter. Why do you need to call `sdb_reset`?

kazarmy: Well `rz_core_analysis_cc_init()` does clear the cc sdb before doing anything with it:
https://github.com/rizinorg/rizin/blob/6601ae97596be4979675c8c358ee240aec0ebbbd/librz/core/cbin.c#L756-L757
so that tells me that i cannot assume that the cc sdb is empty when entering the function.

And since there are no calling conventions defined for the null analysis plugin, i should ensure that the sdb is empty before exiting.

ret2libc: Mmmm ok then!

ret2libc (approved):

wargio (approved):

wargio:
CI needs to be restarted

PR #221, opened 2020-12-18
Remove Java

Maijin:

**Detailed description**

The Java plugins were never been really integrated with core, the commands are all in a different format, not tested and most do not work.

Most users rely on decompiler for java, but if java is considered important these 4 plugins should be rewritten to fit the rizin core or added in r2pm later.

The code is also not integrated, overlapping with ranal and ops.h, DSOjson and PJ, with an upcoming refactor of rasm/ranal it'd be good to remove this plugin from main repo once and for all.


thestr4ng3r:
I think a regular asm/analysis plugin for Java should stay. I remember already removing a lot of the custom analysis for java previously to let it use the regular loop. Maybe more specific code can be removed but I would at least keep the basic plugins if possible.

Maijin:
@thestr4ng3r it still uses ops.h, classes.h which are overlapping with the rest of rzanal.

thestr4ng3r:
Yes, I think this should be refactored to use the regular RzAnalysis things. Or ar least remove the things that are currently not relevant for simple analysis like disassembling and getting cfgs of functions.

XVilka:
:+1: for refactoring :-1: for removal. Java is a "core" architecture that should be supported and improved over time.

Maijin:
Imho it should be refactored after the rasm/ranal fix and therefore should be removed to ease that merge complexity.

ret2libc:
I agree it is not well integrated. I also agree I'd like to keep java in some ways, however I've never used this java plugin myself, as @Maijin said I use other java tools for that.

Is it broken?

Also, maybe we can postpone this decision until we are actually going to review RzBin/RzAsm/RzAnalysis/RzArch. At that point, if we notice java plugin is too much of an outsider we can remove it.

Maijin:
The java command (core) def not working and not tested. 

Also wanted to get your attention to the fact that the fingerprint of java is about 15k+ LOC over 73 files (not normal).

ret2libc:
> Also wanted to get your attention to the fact that the fingerprint of java is about 15k+ LOC over 73 files (not normal).

Yeah this is a very good point indeed and it is definitely a lot. I would say this should be done in a separate plugin?

wargio:
i would not remove it

Maijin:
okay let's rediscussed that when you work on rasm/ranal then, i thought would be harder to remove but it's actually not that bad, so we can always do it later.

PR #223, opened 2020-12-18
Add Exclusions to RzConfig Loading

thestr4ng3r:

**Detailed description**

More exclusions should follow, but these are the two most important ones for Cutter.
For example Cutter needs `scr.interactive` to be set to false, otherwise you will get some weird behavior in a few places.
But for command line rizin, you also don't want everything to be loaded. Also here, when you load a project that was created in Cutter or any other environment where `scr.interactive` was false or `scr.color` was set to something else, you don't suddenly want your shell to change but keep your local settings.

The solution here is to just exclude certain config vars when loading.

### DIFF
--- a/test/unit/test_serialize_config.c
+++ b/test/unit/test_serialize_config.c
@@ -38,7 +38,7 @@ bool test_config_load() {
 
 	Sdb *db = ref_db ();
 	sdb_set (db, "sneaky", "not part of config", 0);
-	bool loaded = rz_serialize_config_load (db, config, NULL);
+	bool loaded = rz_serialize_config_load (db, config, NULL, NULL);
 	sdb_free (db);
 	mu_assert ("load success", loaded);
 
@@ -50,9 +50,35 @@ bool test_config_load() {
 	mu_end;
 }
 
+bool test_config_load_exclude() {
+	static const char * const exclude[] = {
+		"somestring",
+		"someint",
+		NULL
+	};
+	RzConfig *config = rz_config_new (NULL);
+	rz_config_set (config, "somestring", "someoldvalue");
+	rz_config_set_i (config, "someint", 123);
+	rz_config_set_i (config, "somebiggerint", 0);
+	rz_config_lock (config, true);
+
+	Sdb *db = ref_db ();
+	bool loaded = rz_serialize_config_load (db, config, exclude, NULL);
+	sdb_free (db);
+	mu_assert ("load success", loaded);
+
+	mu_assert_eq (rz_list_length (config->nodes), 3, "count after load");
+	mu_assert_streq (rz_config_get (config, "somestring"), "someoldvalue", "excluded config string");
+	mu_assert_eq_fmt (rz_config_get_i (config, "someint"), (ut64)123, "excluded config int", "%"PFMT64u);
+	mu_assert_eq_fmt (rz_config_get_i (config, "somebiggerint"), (ut64)0x1337, "loaded config bigger int", "0x%"PFMT64x);
+	rz_config_free (config);
+	mu_end;
+}
+
 int all_tests() {
 	mu_run_test (test_config_save);
 	mu_run_test (test_config_load);
+	mu_run_test (test_config_load_exclude);
 	return tests_passed != tests_run;
 }
 
--- a/librz/config/serialize_config.c
+++ b/librz/config/serialize_config.c
@@ -23,17 +23,40 @@ RZ_API void rz_serialize_config_save(RZ_NONNULL Sdb *db, RZ_NONNULL RzConfig *co
 	}
 }
 
+typedef struct load_config_ctx_t {
+	RzConfig *config;
+	HtPP *exclude;
+} LoadConfigCtx;
+
 static bool load_config_cb(void *user, const char *k, const char *v) {
-	RzConfig *config = user;
-	RzConfigNode *node = rz_config_node_get (config, k);
+	LoadConfigCtx *ctx = user;
+	if (ctx->exclude && ht_pp_find_kv (ctx->exclude, k, NULL)) {
+		return true;
+	}
+	RzConfigNode *node = rz_config_node_get (ctx->config, k);
 	if (!node) {
 		return 1;
 	}
-	rz_config_set (config, k, v);
+	rz_config_set (ctx->config, k, v);
 	return 1;
 }
 
-RZ_API bool rz_serialize_config_load(RZ_NONNULL Sdb *db, RZ_NONNULL RzConfig *config, RZ_NULLABLE RzSerializeResultInfo *res) {
-	sdb_foreach (db, load_config_cb, config);
+/**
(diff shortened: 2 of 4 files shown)


ret2libc:
What is this for exactly? Could you explain more?

thestr4ng3r:
Description updated.

karliss:
Instead of exceptionlist maybe there needs to be classification for each option whether it describes  current file/project or execution environment and user  preferred data display mode. More radical approach would be splitting them completely so that they never mix. Third approach I can think of is having two layers of config and ability to unset the value with ability to inspect and change both of them, similar to how it's done in Jetbrains IDEs and VSCode.

ret2libc:
I don't like too much this change.

> Instead of exceptionlist maybe there needs to be classification for each option whether it describes current file/project or execution environment and user preferred data display mode

This sounds slightly better, however it is probably hard to decide what user expect to be loaded from project and what from his own "environment file".

Can't you workaround this in Cutter code?

thestr4ng3r:
> Instead of exceptionlist maybe there needs to be classification for each option whether it describes current file/project or execution environment and user preferred data display mode.

Something like this could be done by adding a flag to `RzConfigNode` for example. However, depending on how it's done, this has some conceptual implications:
* Assuming we call the flag file-specific/user-specific or something like this. This would mean that RzConfig, which by itself is an independent config module independent of reverse-engineering purposes, would now get more features that actually belong into RzCore.
* Assuming we call the flag "do not serialize" or "do not deserialize", this would mean that at the time the RzConfig is initially set up, there will be exactly a fixed set of config vars that are serializable/deserializable and it will be impossible to for example get a full deserialization if it is desired somewhere else, considering that the serialization functions may also be useful for other purposes, like undo/redo, realtime collab, ...
* One alternative that still gives us at least some flexibility would be to have something like a "category bitmask" in the RzConfigNode. So you have 64bits that you can each freely assign to categories like "user-setting", "screen", ... and when (de)serializing you could pass a bitmask of what you want to have exactly. That would still be less flexible than direct lists, but maybe not too bad.

That's why I initially decided to go with the exceptionlist since, despite looking a bit dirty, it is actually quite flexible and also in-line with how RzConfig works, which is by using string keys. And the responsibility of specifying the list is in core serialization, where it belongs.

> More radical approach would be splitting them completely so that they never mix.

This would have the same implications as described in the second bullet point above. Also I am not sure if it really makes sense, or if it is even possible to do such a hard split with how the settings currently are.

> Third approach I can think of is having two layers of config and ability to unset the value with ability to inspect and change both of them, similar to how it's done in Jetbrains IDEs and VSCode.

Not sure exactly what you mean by this.

> Can't you workaround this in Cutter code?

In Cutter, yes, but it would be very dirty. Also it doesn't affect only Cutter, as mentioned in the description.

karliss:
By the way consider the things I mentioned more as longer term ideas not necessarily something for this issue.

>Not sure exactly what you mean by this.

Ignore the Jetbrains part I think they have changed the way it works, but Vscode works that way. I'll try to explain it once more. Third approach means maintaining multiple sets of configuration variables.  In the simplest case 2 - gobal settings and project settings. When deciding currently effective value project settings are checked first. If the value in project settings isn't set, global setting value is used. The default values can be considered as third layer. Things like `~/.rizinrc' would modify the global settings. How to fully expose it in rizin text interface is a bit trickier question. Should there be two commands or an argument for modifying each configuration set? Should there be a way to save the current global config or user always needs to modify .rizinrc and restart the program, due to rizin analysis performance restarting can be quite painful. This approach somewhat reduces need to  clasify  each setting as global or project setting.

One more approach to look at the whole situation is whether you view rizin config as storage of key value pairs or as interface with commands `e var` for reading value and `e var=value` for changing value. If you view it only as user interface it doesn't mean everything needs to be stored the same way or use the `e var` interface when saving project state.   In some of the cases `e var=value` is already used as an interface for changing a value somwhere in one of the `core->foo->bar` structures which the project already intentionally save or not save.

thestr4ng3r:
Ok, now I see what you mean. These are indeed more like ideas for the future.

@karliss @ret2libc Thoughts about how to proceed for now?

test/unit/test_serialize_config.c
```diff
@@ -50,9 +50,35 @@ bool test_config_load() {
 	mu_end;
 }
 
+bool test_config_load_exclude() {
+	static const char * const exclude[] = {
+		"somestring",
+		"someint",
+		NULL
+	};
+	RzConfig *config = rz_config_new (NULL);
```
ret2libc, line 59: what about making the `exclude` list a constructor argument so that a `RzConfig` can be created with different exclude-lists based on the need? Maybe also a setter, so Cutter can take `core->config` and apply his own list of things he does not want to save. Do I understand it right that, if not saved, the default/current value is applied?

ret2libc: No sorry, that's actually not the best thing probably... Such configuration belongs to the serialization process, not to the single classes. Maybe we should put it into `RzSerializeResultInfo`?

thestr4ng3r: Not sure what exactly you mean by putting it into `RzSerializeResultInfo` but this struct is purely for returning the result.

ret2libc: You definitely know more about the serialization stuff, but IMHO the list of things you do not want to serialize is a property of the "serialization process". It's not something of `RzConfig` but not even something that `RzCore` should choose on its own.

Right now the output of the serialization process is just passed as a single `Sdb`. Maybe this should actually be `RzSerialization` which includes an `Sdb *` and some configuration fields (e.g. in this case the list of config variables to exclude from the serialization/deserialization). WDYT?

thestr4ng3r: > the list of things you do not want to serialize is a property of the "serialization process". It's not something of `RzConfig` but not even something that `RzCore` should choose on its own.

Yes. This is why the exclusion list here is currently an argument of `rz_serialize_config_load()` instead of for example a member of `RzConfigNode`.
I also agree that core should rather not choose this at `RzCore` creation time, but I think the right place is where the `RzConfig` instance is loaded in the context of a project, which is here: https://github.com/rizinorg/rizin/pull/223/files#diff-e10dc9ab50cb92db60c5eaf3dd32a456141636e478a2e6b320f25d3c28892e08R57


> Right now the output of the serialization process is just passed as a single `Sdb`. Maybe this should actually be `RzSerialization` which includes an `Sdb *` and some configuration fields (e.g. in this case the list of config variables to exclude from the serialization/deserialization). WDYT?

Yeah I think it can make sense to have such a struct that holds possibly also multiple configuration fields for the (de)serialization. But I would rather keep it clearly separated from what is the output and what is input/config. So until now we only have one output and one config, which I think is good enough as just two args.

ret2libc: Okok, makes sense. Let's consider this if configuration stuff becomes more though.

ret2libc (approved):

PR #226, opened 2020-12-19
rz_sys_system should not depend on LIBC_HAVE_FORK

ret2libc:
This PR should fix current failures in AppVeyor. In particular, it moves all the arch-dependent code to sys.c and it implements `rz_sys_system` in different ways based on the availability of the host machine. Moreover it changes the signature of `rz_sandbox_system` so that it always does a `system`. In some cases `rz_sandbox_system` was called to actually execute just a `/bin/sh -c` in the current process, but I changed those few callers to just call `rz_sys_execl`, because I think "system" has a "standard meaning" which is `fork+exec(/bin/sh -c)`.


ret2libc:
Somehow it's this commit that broke dev, though it's not clear to me why.

librz/util/sandbox.c
```diff
@@ -210,7 +210,6 @@ RZ_API int rz_sandbox_system(const char *x, int n) {
 		eprintf ("sandbox: system call disabled\n");
 		return -1;
 	}
-#if LIBC_HAVE_FORK
 #if LIBC_HAVE_SYSTEM
```
ret2libc: maybe all this thing should be inside `rz_sys_system`.

ret2libc:
@XVilka AppVeyor still not running on PR :(

librz/util/sandbox.c
```diff
@@ -210,7 +210,6 @@ RZ_API int rz_sandbox_system(const char *x, int n) {
...
 	if (n) {
 #if APPLE_SDK_IPHONEOS
```
kazarmy:
```suggestion
#if APPLE_SDK_IPHONEOS && LIBC_HAVE_FORK
```

kazarmy (changes requested):

librz/util/sandbox.c
ret2libc: I've actually changed the whole thing so that `rz_sys_system` implements the right thing depending on the availability of `system`/`fork` on the host machine. So now `rz_sandbox_system` just checks if sandboxing is enabled and then it calls `rz_sys_system`.

XVilka (approved):

thestr4ng3r (approved):

librz/util/sandbox.c
kazarmy: Oh ok

Issue #179, opened 2020-12-14
Is a rizin installation side by side possible with radare2?


Petross404:
I want to package rizin for gentoo. Before installing rizin, do I have to uninstall radare2 first or they are separate installations?

XVilka:
> I want to package rizin for gentoo

That would be awesome, thank you!

> separate installations

They are - you can see https://aur.archlinux.org/packages/rizin-git/ for inspiration.

Please let us know if you meet any issues so we can fix them before 1.0 is released.

Petross404:
The last lines of the emerge:
```
OUTPUT(rz-test elf64-x86-64)
/usr/lib/gcc/x86_64-pc-linux-gnu/10.2.0/../../../../x86_64-pc-linux-gnu/bin/ld: total time in link: 0.076870
make[2]: Leaving directory '/mnt/Volume_3/Gentoo/temp/portage/dev-util/rizin-9999/work/rizin-9999/binrz/rz-test'
make[1]: Leaving directory '/mnt/Volume_3/Gentoo/temp/portage/dev-util/rizin-9999/work/rizin-9999/binrz'
>>> Source compiled.
>>> Test phase [not enabled]: dev-util/rizin-9999

>>> Install dev-util/rizin-9999 into /mnt/Volume_3/Gentoo/temp/portage/dev-util/rizin-9999/image
make -j5 -l4 DESTDIR=/mnt/Volume_3/Gentoo/temp/portage/dev-util/rizin-9999/image install 
mkdir -p "/mnt/Volume_3/Gentoo/temp/portage/dev-util/rizin-9999/image/usr/share/doc/rizin-9999"
/usr/lib/portage/python3.8/ebuild-helpers/xattr/install -d "/mnt/Volume_3/Gentoo/temp/portage/dev-util/rizin-9999/image/usr/share/doc/rizin-9999"
/usr/share/doc/rizin-9999
for FILE in doc/* ; do \
        if [ -f $FILE ]; then /usr/lib/portage/python3.8/ebuild-helpers/xattr/install -m 644 $FILE "/mnt/Volume_3/Gentoo/temp/portage/dev-util/rizin-9999/image/usr/share/doc/rizin-9999" || true ; fi; \
done
[...]

Issue #160, opened 2020-12-10
macOS homebrew support


kciredor:
Having (mostly) everything scripted when (re-)provisioning my laptop, it would be nice for rizin to be added to homebrew like radare2 https://github.com/Homebrew/homebrew-core/blob/master/Formula/radare2.rb.

That way it's easy to just add another brew install and it'll be updated upon every brew update.

XVilka:
It makes sense to do after the first release is out.

ret2libc:
Hi @kciredor ! Would you be available to help us with that? It should be similar to radare2, but using `meson`/`ninja` instead of `configure`/`make` for build/install steps.

Maijin:
Some docs - https://github.com/toolmantim/tap-release

XVilka:
A heads up for the maintainers - we [tagged the first release](https://github.com/rizinorg/rizin/releases/tag/0.1.0), building artifacts and writing the release notes. If you planned to craft a new tap - it's a good time to do so.

ret2libc:
Hi @kciredor . I'm planning to work relatively soon on providing repositories for linux distros so that they can remain updated with Rizin versions without having to re-download the release every time. It will also help us at least until distros will start packaging Rizin.

However, many of us are not working on Mac (not usually at least) so it would be great to receive some help on that. Are you still interested in the homebrew part? Would you be available to help with a basic formula and maybe look at what we need to do/setup to use https://github.com/toolmantim/tap-release ? It would be great to setup something like that, because that way we can just automatically update the formula when we release a new update, without requiring manual intervention.

Let us know please. Thanks.

pcolladosoto:
Hi all! According to the current [`INSTALL.md`](https://github.com/rizinorg/rizin/blob/fc24887c5cabe649cd22bf681c9280666d4f883e/INSTALL.md) I believe there's no macOS Homebrew tap available for Rizin.

I usually work on a macOS machine and I would be more than happy to lend a hand and get that tap up and running.

Before doing so I would really appreciate it if you could tell me wether this is something you'd like: I'd hate to just bog you down with something you're not really seeking...

Thanks a ton for Rizin 😸

ret2libc:
I'm not sure it's still worth it after all TBH... Rizin is already in homebrew core formulas and it is updated right now so maybe we can just rely on that. What do you think?

pcolladosoto:
[...]

PR #229, opened 2020-12-21
Add packagers.md

teasp00n:
Add initial information for people wanting to create their own Rizin packages.

### DIFF
--- a/doc/PACKAGERS.md
+++ b/doc/PACKAGERS.md
@@ -16,6 +16,8 @@ the Rizin source and all dependencies using
 If you change branches run `git submodule update --init --recursive` to ensure
 you are tracking the correct versions of the submodules.
 
+See [BUILDING.md][] for more details.
+
 Building
 --------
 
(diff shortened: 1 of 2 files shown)


XVilka:
Maybe @Petross404 also wants to take a look from the Gentoo PoV (https://github.com/rizinorg/rizin/issues/179)

XVilka:
Also @kciredor (https://github.com/rizinorg/rizin/issues/160)

doc/PACKAGERS.md
```diff
@@ -0,0 +1,83 @@
...
+We will look at each step below.
+
+Cloning Source
+--------------
+
+Rizin uses git submodules to track versions of dependencies. You can check out
+the Rizin source and all dependencies using
+`git clone --recursive https://github.com/rizinorg/rizin`.
+
+If you change branches run `git submodule update --init --recursive` to ensure
```
ret2libc, line 16: I would add a link to https://github.com/rizinorg/rizin/blob/dev/BUILDING.md .

doc/PACKAGERS.md
```diff
@@ -0,0 +1,83 @@
...
+
+Rizin uses the `meson` build system. Command line flags passed to meson can
+change how Rizin is built.
+
+To define the base install location for Rizin use the `--prefix` flag when
+invoking `meson`. For system installs it is common to use `/usr`. If in doubt
+check your distributions packaging guidelines.
+
+Rizin uses the Capstone disassembly engine and supports versions 3, 4, and 5.
+By default we pull version 4 and statically link it into the Rizin executables.
```
ret2libc:
```suggestion
By default we use a custom version of Capstone based on v4 and statically link it into the Rizin executables.
```

doc/PACKAGERS.md
```diff
@@ -0,0 +1,83 @@
...
+Some distributions might prefer that a system version of Capstone be
+dynamically linked at runtime. To do this, use the `-Duse_sys_capstone=true`
+command line option when running `meson`. See `capstone.md` for more details.
```
ret2libc: I would remove the reference to `capstone.md` and I would prefer to keep everything here in one file.

doc/PACKAGERS.md
```diff
@@ -0,0 +1,83 @@
...
+
```
ret2libc, line 60: I would add some paragraphs to mention the other compile-time options like `use_sys_magic`, `use_sys_zip`, etc. and add a reference to https://github.com/rizinorg/rizin/blob/dev/meson_options.txt . I think they are quite important for distros that want to avoid embedded code.

doc/PACKAGERS.md
```diff
@@ -0,0 +1,83 @@
...
+Building
+--------
+
+Rizin uses the `meson` build system. Command line flags passed to meson can
+change how Rizin is built.
+
+To define the base install location for Rizin use the `--prefix` flag when
+invoking `meson`. For system installs it is common to use `/usr`. If in doubt
+check your distributions packaging guidelines.
+
```
ret2libc, line 30: I would add something like:
```
If you do not use `/usr` as a prefix, you may want to use `-Dlocal=disabled` to avoid `RPATH` in the installed binaries.
```

doc/PACKAGERS.md
```diff
@@ -0,0 +1,83 @@
...
+Rizin uses the Capstone disassembly engine and supports versions 3, 4, and 5.
+By default we pull version 4 and statically link it into the Rizin executables.
+Some distributions might prefer that a system version of Capstone be
+dynamically linked at runtime. To do this, use the `-Duse_sys_capstone=true`
+command line option when running `meson`. See `capstone.md` for more details.
+
+Once you are happy with the flags you have passed to `meson` to configure your
+build you need to actually compile Rizin using `ninja`. You can do this with
+`ninja -C build`.  The `-C` flag tells `ninja` to change into the directory
```
ret2libc: Just mention the minimum things necessary and then reference BUILDING.md to avoid duplications.

doc/PACKAGERS.md
```diff
@@ -0,0 +1,83 @@
...
+---------
+
+The final step is creating a package from the build outputs. To do this we will
+invoke `DESTDIR=$PKGDIR ninja -C build install`.
+
+Recall in the building step how we defined a `--prefix` to choose where the
+software was going to be installed? If we just ran `ninja -C build install`
+then we would install the software into that prefix which would be `/usr` using
+the example above. That isn't what we want when we are packaging software for
+distribution! Defining `DESTDIR` allows us to choose a base location to install
```
ret2libc:
```suggestion
distributions! Defining `DESTDIR` allows us to choose a base location to install
```

doc/PACKAGERS.md
```diff
@@ -0,0 +1,83 @@
...
+There are files we want to include in the package that aren't installed using
+`ninja` so we need to add these by hand.
+
+To do this we can use `install` to create directories with the desired attributes
+and copy the files in.
+
+For example:
+
+```sh
+install -dm644 "${pkgdir}/usr/share/doc/rizin"
```
ret2libc: instead of `${pkgdir}` i would use `${destdir}` to make it a bit clearer what this refers to

doc/PACKAGERS.md
```diff
@@ -0,0 +1,83 @@
...
+Once you are happy with the flags you have passed to `meson` to configure your
+build you need to actually compile Rizin using `ninja`. You can do this with
+`ninja -C build`.  The `-C` flag tells `ninja` to change into the directory
+`build` before building. If you do not specify a `target` the default target of
+`all` will be used.
+
+Packaging
+---------
+
+The final step is creating a package from the build outputs. To do this we will
```
ret2libc: "To do this it may be required to invoke ... , based on the distribution you are targeting."

Not all distros actually require to set DESTDIR manually (e.g. for Fedora I don't need to do that).

doc/PACKAGERS.md
```diff
@@ -0,0 +1,83 @@
...
+
+Building
+--------
+
+Rizin uses the `meson` build system. Command line flags passed to meson can
+change how Rizin is built.
+
+To define the base install location for Rizin use the `--prefix` flag when
+invoking `meson`. For system installs it is common to use `/usr`. If in doubt
+check your distributions packaging guidelines.
```
ret2libc, line 29: Also
```
If you want to specify different directories for binaries, libraries, header files, etc. you may want to look at `--bindir`, `--libdir`, `--includedir` or check `meson setup --help` for more options. For extra control over the directories used by Rizin, have a look at options `rizin_sdb`, `rizin_zigns`, etc. in [link to meson_options.txt file]
```

ret2libc (changes requested):
Thanks a lot! I think this is going to be useful for others ;)

I added some comments here and there.
Also, could you add a small note at the end of `BUILDING.md` to reference this file as well? Just a short sentence should be enough, I think (e.g. "If you are trying to build Rizin to create a package for a distribution, look at [link to this file]").

ret2libc (approved):
Awesome! I think this is already a quite good starting point and things can be improved from here as we go!

Thanks again for this! I think it will be very useful once we start doing the first releases.

XVilka (approved):
Looks great!

XVilka:
Lets merge it, further improvements can be done as separate PRs.

PR #234, opened 2020-12-21
Fix #218: End pdf on function max bound

### LINKED ISSUE #218: Buggy formatting of disassembling output
### Work environment

| OS/arch/bits (mandatory)                             | Arch Linux 64 bit
| File format of the file you reverse (mandatory)      | PE
| Architecture/bits of the file (mandatory)            | x86-32
| rizin -v full output, **not truncated** (mandatory)  | rizin 0.1.0-git 25430 @ linux-x86-64 git.0.1.0-git commit: e08a87ee8947e77eed8e4ca659e6e87f0852ef2d build: 2020-12-17__16:18:26

### Expected behavior
The disassembly outlines should look clean and the output should stay clear

### Actual behavior
When disassembling the `entry0` function in this binary, the output is a mess

image

- Open the attached file in rizin `$ rizin 5bae961fec67565fb88c8bcd3841b7090566d8fc12ccb70436b5269456e55c00`
- Analyze it using `aaa`
- run `pdf`, you should be on `entry0`
- observe the buggy output as shown in the image above

### Additional Logs, screenshots, source-code,  configuration dump, ...
[5bae961fec67565fb88c8bcd3841b7090566d8fc12ccb70436b5269456e55c00.zip](https://github.com/rizinorg/rizin/files/5714726/5bae961fec67565fb88c8bcd3841b7090566d8fc12ccb70436b5269456e55c00.zip)

kazarmy:

**Detailed description**

This pr ends pdf output on function max bound.

**Test plan**

All builds are green.

**Closing issues**

Fixes #218.

### DIFF
--- a/librz/core/disasm.c
+++ b/librz/core/disasm.c
@@ -5211,6 +5211,9 @@ toro:
 	rz_cons_break_push (NULL, NULL);
 	for (i = idx = ret = 0; addrbytes * idx < len && ds->lines < ds->l; idx += inc, i++, ds->index += inc, ds->lines++) {
 		ds->at = ds->addr + idx;
+		if (ds->pdf && ds->at >= rz_analysis_function_max_addr (ds->pdf)) {
+			break;
+		}
 		ds->vat = rz_core_pava (core, ds->at);
 		if (rz_cons_is_breaked ()) {
 			RZ_FREE (nbuf);
--- a/test/db/analysis/x86_64
+++ b/test/db/analysis/x86_64
@@ -2614,14 +2614,12 @@ diff: type: new
 |       .-> 0x00000004      c1ebfd         shr ebx, 0xfd
 |           0x00000007      bb2c000000     mov ebx, 0x2c               ; ','
 \           0x0000000c      cc             int3
-..
 
 / 16: fcn.00000000 ();
 |           0x00000000      b8210000c1     mov eax, 0xc1000021         ; '!'
 |       `=< 0x00000005      ebfd           jmp 4
 |           0x00000007      bb2c000000     mov ebx, 0x2c               ; ','
 \           0x0000000c      cc             int3
-..
 
 #
 offset: 0x00000000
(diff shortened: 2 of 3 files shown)


ret2libc (approved):

librz/core/disasm.c
```diff
@@ -5211,6 +5211,9 @@ RZ_API int rz_core_print_disasm(RzPrint *p, RzCore *core, ut64 addr, ut8 *buf, i
 	rz_cons_break_push (NULL, NULL);
 	for (i = idx = ret = 0; addrbytes * idx < len && ds->lines < ds->l; idx += inc, i++, ds->index += inc, ds->lines++) {
 		ds->at = ds->addr + idx;
+		if (ds->pdf && ds->at >= rz_analysis_function_max_addr (ds->pdf)) {
+			break;
```
thestr4ng3r, line 5215: I think the right place to fix this would be here:
https://github.com/rizinorg/rizin/blob/da4703d757f25b51c520deb83deb55e9ee98af45/librz/core/cmd_print.c#L5389-L5403
You can see it is starting to disassemble at `fcn->addr` which is already wrong because the addr of a function is its entrypoint, not necessarily its lowest address. For how `pdf` is meant to work, it should probably just disassemble from the min address to the max address and the whole logic there about `RZ_MAX (linearsz, realsz);` and so on is probably superfluous.

thestr4ng3r (changes requested):

librz/core/disasm.c
kazarmy: I think users would be surprised though if `pdf` starts from min address rather than entrypoint, for functions where min address != entrypoint.

thestr4ng3r: This would still have been the correct place to check, even if starting to disassemble at the entrypoint. Imho the merged fix is wrong, or at least hiding the real issue.

PR #239, opened 2020-12-21
Remove Pseudo-Sandbox

thestr4ng3r:
The sandbox is practically impossible to keep remotely secure so instead of claiming false security, better to remove it completely.


librz/util/sys.c
```diff
@@ -1826,3 +1791,170 @@ RZ_API int rz_sys_fork(void) {
...
+		free (wepath);
+	}
+#else // __WINDOWS__
+	ret = open (epath, perm, mode);
+#endif // __WINDOWS__
+	free (epath);
+	return ret;
+}
+
+RZ_API FILE *rz_sys_fopen (const char *path, const char *mode) {
```
ret2libc:
```suggestion
RZ_API FILE *rz_sys_fopen(const char *path, const char *mode) {
```

librz/util/sys.c
```diff
@@ -1826,3 +1791,170 @@ RZ_API int rz_sys_fork(void) {
...
+		free (wepath);
+#else // __WINDOWS__
+		ret = fopen (epath, mode);
+#endif // __WINDOWS__
+	}
+	free (epath);
+	return ret;
+}
+
+RZ_API int rz_sys_kill(int pid, int sig) {
```
ret2libc, line 1929: these things are copied-pasted from sandbox.c right?
I think it would be better to check the presence of `kill` in meson.build and base the implementation of these things on `HAVE_KILL` instead of having know ourselves what platforms/compiler support a given function.

ret2libc (approved):

librz/util/sys.c
thestr4ng3r: Yeah, they are 1:1 copies of what was left in sandbox after removing all pure stubs. The approach sounds good to me but I would rather do that in another pr.

ret2libc: absolutely.

PR #250, opened 2020-12-23
Add -std=c99, for older gcc ##build

ret2libc:
See https://github.com/rizinorg/rizin/pull/195. This is just to trigger appveyor.


XVilka:
As I suspected - we should use pure "c99":
```
meson.build:1:0: ERROR: Value "gnu99" (of type "string") for combo option "C language standard to use" is not one of the choices. Possible choices are (as string): "none", "c89", "c99", "c11".
```

XVilka:
```
FAILED: sdb.p/shlr_sdb_src_main.c.o 
gcc -Isdb.p -I../shlr/sdb/src -fdiagnostics-color=always --coverage -pipe -D_FILE_OFFSET_BITS=64 -Wall -Winvalid-pch -std=c99 -g -Werror -Wno-cpp -MD -MQ sdb.p/shlr_sdb_src_main.c.o -MF sdb.p/shlr_sdb_src_main.c.o.d -o sdb.p/shlr_sdb_src_main.c.o -c ../shlr/sdb/src/main.c
../shlr/sdb/src/main.c: In function ‘insertkeys’:
../shlr/sdb/src/main.c:243:30: error: implicit declaration of function ‘strdup’; did you mean ‘strcmp’? [-Werror=implicit-function-declaration]
  243 |      char *v, *kv = (char *) strdup (args[i]);
      |                              ^~~~~~
      |                              strcmp
../shlr/sdb/src/main.c:243:21: error: cast to pointer from integer of different size [-Werror=int-to-pointer-cast]
  243 |      char *v, *kv = (char *) strdup (args[i]);
      |                     ^
```

ret2libc:
> As I suspected - we should use pure "c99":
> 
> ```
> meson.build:1:0: ERROR: Value "gnu99" (of type "string") for combo option "C language standard to use" is not one of the choices. Possible choices are (as string): "none", "c89", "c99", "c11".
> ```

This is unfortunately a bug in meson. It is supposed to work and there is already a fix, but it did not land in 0.56.0. It should be fixed in 0.56.1 though.

ret2libc:
And we do need gnu99 in some cases (though honestly I don't remember why and maybe it is possible to fix the codebase to not require it?)

ret2libc:
At this point, given the bug in meson I suggest to either:
1) wait for this change until 0.56.1 is release
2) just use `add_project_arguments` as it was in the original PR.

cc @thestr4ng3r

thestr4ng3r:
What exactly is the bug? Do you mean that you can't specify `gnu99`? I think the best option would be to make the code work with `c99`.

XVilka:
According to Meson's [issue](https://github.com/mesonbuild/meson/issues/7599#issuecomment-678403760) MSVC just fallbacks to `c99` anyway in case of `gnu99`. Thus, I think, it is the same as we set `c99` on our side.

XVilka:
For `strdup` error we can just define `#define _POSIX_C_SOURCE 200809L` for non-MSVC compilers. Or something like:
```py
add_project_arguments('-D_GNU_SOURCE', language : 'c')
add_project_arguments('-D_GNU_SOURCE', language : 'cpp')
```

ret2libc:
> For `strdup` error we can just define `#define _POSIX_C_SOURCE 200809L` for non-MSVC compilers. Or something like:
> 
> ```python
> add_project_arguments('-D_GNU_SOURCE', language : 'c')
> add_project_arguments('-D_GNU_SOURCE', language : 'cpp')
> ```

I tried having a look at this and I still think the right solution is to use `c_std=gnu99`. Which *should* fallback to `c99` on windows, but doesn't because of an issue with recent meson. IMHO in the meantime we can just do something like:
```
if cc.has_argument('--std=gnu99')
   add_project_arguments('--std=gnu99', language: ['c', 'cpp'])
elif cc.has_argument('--std=c99')
   add_project_arguments('--std=c99', language: ['c', 'cpp'])
endif
```
This should be required only until meson fixes the regression with regard to `c_std`.

What are the reasons for not using gnu99? (which is just a GCC thing afaik).

XVilka (approved):

PR #243, opened 2020-12-21
Implement rz_subprocess_start_opt API to choose pipes

ret2libc:
**Your checklist for this pull request**
- [x] I've read the [guidelines for contributing](https://github.com/rizinorg/rizin/blob/master/DEVELOPERS.md) to this repository
- [x] I made sure to follow the project's [coding style](https://github.com/rizinorg/rizin/blob/master/DEVELOPERS.md#code-style)
- [x] I've added tests that prove my fix is effective or that my feature works (if possible)
- [ ] I've updated the documentation and the [rizin book](https://github.com/rizinorg/rizinbook) with the relevant information (if needed)

**Detailed description**

So far the `RzSubprocess` API always created pipes for stdin/stdout/stderr, but this is not always desirables as in some cases a program may want to directly work with the terminal (e.g. `vim`). As we want to use subprocess API to implement pipe and `!` commands, I'm adding this API so that I can specify when I want pipes and when not.

**Test plan**

Windows still needs to be implemented.

Issue #209, opened 2020-12-17
Enumerate all third-party code and libraries used in Rizin


XVilka:
Currently some of the libraries we use are imported during the fetching process as a submodules (using system-wide version or download as an archive). But a lot of external dependencies are in fact just copied inside the sources. Some are quite diverged from the original, e.g. `shlr/tcc`, while others just outdated. We should enumerate all such occurences and list them in a separate documentation file, with instructions where it came from and how to update. Or additional scripts on automating it. Also we should note what licenses every such piece of code uses.

### Managed as subprojects

- `subprojects/yxml`, **YXML**, version in `subprojects/yxml/meson.build`. Can be updated from https://g.blicky.net/yxml.git/
- `subprojects/xxhash`, xxhash, version in `subprojects/xxhash/meson.build`

###  Copied files

- `shlr/heap/include/rz_jemalloc` - copied files from **jemalloc** library. **NOTE!**: We should not blindly update this jemalloc since we should be able to parse both old (pre-5.x) heap versions and new ones. So it might be needed to reworked completely, see also https://github.com/rizinorg/rizin/issues/157 and https://github.com/rizinorg/rizin/issues/158
- `shlr/mpc` - MPC, **BSD** license, taken from https://github.com/orangeduck/mpc Will be closed by  https://github.com/rizinorg/rizin/issues/631
- `librz/magic` - **BSD**, taken from OpenBSD, somewhat outdated, has custom modifications (**v 1.29 2009/11/11**?) #721

### Managed as subprojects

- `shlr/capstone` - almost clean **capstone** 4.0.2, but there are 4 patches applied. Most of them can either be dropped or upstreamed, only the x86-16 one is a bit tricky

###  Copied files

- pieces from **binutils-gdb** were used in many different places (opcodes library/libopcodes, libiberty, BFD library):
    1. `librz/asm/arch/arc/gnu/`
    2. `librz/asm/arch/arm/aarch64/`
    4. `librz/asm/arch/cris/gnu/`
    5. `librz/asm/arch/hppa/gnu/`
    6. `librz/asm/arch/lanai/gnu/`
    7. `librz/asm/arch/mips/gnu/`
    8. `librz/asm/arch/nios/gnu/`
    9. `librz/asm/arch/xtensa/gnu/`
    10. `librz/asm/arch/tricore/gnu/`
    11. `librz/asm/arch/sparc/gnu/`
    13. `librz/asm/arch/vax/`
[...]

PR #251, opened 2020-12-23
Use mu_main to reduce boilerplate in unit tests

ret2libc:
* Disable colors on Windows because terminal looks bad otherwise
* Use environment on Windows to set PATH and make sure unit tests use
  the compiled libraries and not the installed ones
* Define ssize_t on Windows as well by using SSIZE_T
* A bit of refactoring in rz_file_mmap API
* Remove io_mmap plugin and just keep io_default, which does a mmap if possible or just directly access the file if not possible.
* Simplify io_default code by just using RzBuffer API
* Add unit test for io_default
* Add unit test for rz_file_mmap API
* Allow IO maps of size 0. This is useful when creating a new file (e.g. `rizin -w newfile`) and you use `r` to resize the file. By having a map of 0 size, the map will be resized as well. Without this, the newfile cannot be easily mmaped and viewed in `om`.

 

**Detailed description**

I have marked test_rzpipe as broken on windows because it does not work for now and I want to fix that by using the improved RzSubprocess API. 

This is needed for https://github.com/rizinorg/rizin/pull/243 , so I can have unit tests to make sure stuff works well on linux & windows.

**Test plan**

AppVeyor should be green even with tests enabled


librz/magic/file.h
```diff
@@ -46,9 +46,6 @@
 /* Do this here and now, because struct stat gets re-defined on solaris */
 #include <sys/stat.h>
 #include <stdarg.h>
-#ifdef _MSC_VER
```
XVilka, line 49: One more embedded file for https://github.com/rizinorg/rizin/issues/209

XVilka (approved):

librz/magic/file.h
ret2libc: Yeah, however I am not sure how to use this if not embedding. AFAIR this is slightly different from the linux `magic` and I don't think distros have a `magic-bsd` package.

XVilka: Yes, embedding is fine. We just should note imported version somewhere and our changes.

XVilka:
See also the changed error message:
```
[XX] C:\projects\rizin\test\db\tools\rz_find rz-find can't open
ANSICON=1 RZ_NOPLUGINS=1 rizin -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -Qc '!rz-find -n  -f 0x00010000 -t 0x00020000 -r -X -s 250382 incorrect_file | grep incorrect_file' -
-- stderr
@@ -1,2 +1,2 @@
-rz_sys_open: No such file or directory
+rz_sys_open: (0x2) The system cannot find the file specified.
 Cannot open file 'incorrect_file'
```
https://ci.appveyor.com/project/rizinorg/rizin/builds/37056378/job/5pmtiku169x60fi5#L4020

thestr4ng3r (commented):
Changes lgtm but the history should be cleaned up if possible to split into `mu_main`, `mmap` removal, windows-specific stuff...

ret2libc:
> Changes lgtm but the history should be cleaned up if possible to split into `mu_main`, `mmap` removal, windows-specific stuff...

Done. WDYT now?

test/unit/test_buf.c
```diff
@@ -154,27 +154,54 @@ bool test_r_buf_mmap(void) {
 	write (fd, content, length);
 	close (fd);
 
-	b = rz_buf_new_mmap (filename, RZ_PERM_RW);
+	b = rz_buf_new_mmap (filename, O_RDWR, 0);
```
yossizap, line 157: Might be out of scope for this PR but I think that `rz_sys_open` should accept `RZ_PERM` and use your `rizinperms2open` in the same manner as `rz_io_def_mmap_create_new_file` to keep things consistent. 
Especially when `O_CREAT` is implicitly toggled in `RZ_PERMS` for WR while having to set it explicitly with functions that accept `O_*` perms.

yossizap (approved):

test/unit/test_buf.c
ret2libc: I agree. I started doing that but it required many changes all over the place so I ended up keeping it like this for now.

thestr4ng3r (approved):

PR #263, opened 2020-12-25
Add rz-test in codecov ignore list

Maijin:
Add rz-test in codecov ignore list


kazarmy:
Should we test the tester itself?

ret2libc:
Who test the tester? 🤪

PR #265, opened 2020-12-26
Remove EMSCRIPTEN__TODO and TODO

Maijin:
Remove Emscripten and TODO

### DIFF
--- a/librz/util/debruijn.c
+++ b/librz/util/debruijn.c
@@ -7,7 +7,6 @@
 // char* peda_charset =
 //    "A%sB$nC-(D;)Ea0Fb1Gc2Hd3Ie4Jf5Kg6Lh7Mi8Nj9OkPlQmRnSoTpUqVrWsXtYuZvwxyz";
 
-//TODO(crowell): Make charset configurable, to allow banning characters.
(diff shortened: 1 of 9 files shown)


wargio (commented):
i'm ok for emscripten but dunno for those TODOs

librz/util/debruijn.c
```diff
@@ -7,7 +7,6 @@
 // char* peda_charset =
 //    "A%sB$nC-(D;)Ea0Fb1Gc2Hd3Ie4Jf5Kg6Lh7Mi8Nj9OkPlQmRnSoTpUqVrWsXtYuZvwxyz";
 
-//TODO(crowell): Make charset configurable, to allow banning characters.
```
XVilka, line 10: I think this TODO should stay, it's really useful feature.

XVilka (changes requested):

librz/util/debruijn.c
Maijin: If it's useful then It should be moved to an issue, not lost in-line TODO

wargio: i agree with Maijin. that's a valid point.

XVilka (approved):

PR #260, opened 2020-12-25
Set CI buildtypes to highest opt level possible and set default buildtype to debugoptimized

kazarmy:
**Your checklist for this pull request**
- [X] I've read the [guidelines for contributing](https://github.com/rizinorg/rizin/blob/master/DEVELOPERS.md) to this repository
- [X] I made sure to follow the project's [coding style](https://github.com/rizinorg/rizin/blob/master/DEVELOPERS.md#code-style)
- [ ] I've added tests that prove my fix is effective or that my feature works (if possible)
- [ ] I've updated the documentation and the [rizin book](https://github.com/rizinorg/rizinbook) with the relevant information (if needed)

**Detailed description**

This pr:

1. Sets the `buildtype` for CI builds to `release` (`-O3`) except the asan and Codecov builds. This is done because compiler optimizers increase performance but can also change code behavior (especially if the code invokes UB) so this needs to be tested.
  1.1 The asan `buildtype` is set to `debugoptimized` (`-O2 -g`) for hopefully sane stack traces.
  1.2 The Codecov `buildtype` is set to `debug` because `debugoptimized` (as seen in https://github.com/rizinorg/rizin/pull/260#issuecomment-751470698) produces wonky Codecov reports.

2. Set the default `buildtype` to `debugoptimized` due to reasons given in #256 i.e. users might leave performance on the table otherwise.

**Test plan**

All meson builds are green, including Windows and asan builds.

**Closing issues**

Closes #256.

PR #268, opened 2020-12-26
rz-test: Use RZ_FREE instead of free in _result_info_free()-related code

kazarmy:

**Detailed description**

This pr sets pointers of `_result_info_free()`-related code to NULL after freeing (using `RZ_FREE` or manually), intended as part of the effort to track down the following OpenBSD bug (https://builds.sr.ht/~xvilka/job/377421#task-test-636):

openbsd-uaf

Cherry-picked from #260.

**Test plan**

All builds, except maybe the OpenBSD build, are green.

**Closing issues**

...

### DIFF
--- a/binrz/rz-test/run.c
+++ b/binrz/rz-test/run.c
@@ -353,8 +353,8 @@ RZ_API void rz_test_asm_test_output_free(RzAsmTestOutput *out) {
 	if (!out) {
 		return;
 	}
-	free (out->disasm);
-	free (out->bytes);
+	RZ_FREE (out->disasm);
+	RZ_FREE (out->bytes);
 	free (out);
 }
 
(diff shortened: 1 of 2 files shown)


thestr4ng3r:
In theory this shouldn't be necessary so not sure if we should merge it, but for debugging the idea looks good to me.

kazarmy:
This code might conveniently convert a UAF on OpenBSD to a null pointer error on some more accessible OS that has ASAN available (both currently and in the future). So I think it's worth the merge. Theoretically the performance hit is miniscule.

kazarmy:
Going to `memset` all the things.

kazarmy:
Unfortunately there's nothing significant to `memset` 😞

thestr4ng3r:
We should try to extract the coredump next time this happens in OpenBSD. Sourcehut gives you ssh access after a build fail but only for limited time so we will have to be fast, cc @XVilka

kazarmy:
The OpenBSD uaf seems to have disappeared.

thestr4ng3r:
I think it's non-deterministic. Probably some race condition.

PR #269, opened 2020-12-26
Added support for regex in test output ##test

yossizap:

**Detailed description**

REGEXP_OUT/ERR is used to apply regex on the program's stdout/stderr which will then be compared with EXPECT/_ERR.

**Test plan**

Check that tests that use regex pass.

### DIFF
--- a/test/db/archos/linux-x64/dbg_oo
+++ b/test/db/archos/linux-x64/dbg_oo
@@ -95,3 +95,13 @@ Hello world!
 EOF
 RUN
 
+NAME=ood check for ptrace errors
+FILE=bins/elf/analysis/x86-helloworld-gcc
+CMDS=<<EOF
+ood
+EOF
+REGEXP_ERR=([a-zA-Z-]+\s+)
+EXPECT_ERR=<<EOF
+Process with PID attach File -helloworld-gcc  reopened in read-write mode
+EOF
+RUN
--- a/binrz/rz-test/run.c
+++ b/binrz/rz-test/run.c
@@ -170,11 +170,13 @@ RZ_API bool rz_test_check_cmd_test(RzSubprocessOutput *out, RzCmdTest *test) {
 		return false;
 	}
 	const char *expect_out = test->expect.value;
-	if (expect_out && strcmp (out->out, expect_out) != 0) {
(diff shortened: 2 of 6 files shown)


test/db/archos/linux-x64/dbg_oo
```diff
@@ -95,3 +95,13 @@ Hello world!
 EOF
 RUN
 
+NAME=ood check for ptrace errors
+FILE=bins/elf/analysis/x86-helloworld-gcc
+CMDS=<<EOF
+ood
+EOF
+REGEXP_ERR=([a-zA-Z-]+\s+)
```
yossizap, line 103: The implementation from OpenBSD in rz_regex is somewhat limited, I wasn't able to use negative/positive lookahead, non-capturing groups, \W, \D, \S, \b and had other weird behavior compared to python.

XVilka: Maybe it could be updated to the more recent version? I doubt modern version of it doesn't have support for those.

yossizap: I couldn't find anything related in [regcomp.c](https://github.com/openbsd/src/commits/master/lib/libc/regex). Based on their [commit history](https://github.com/openbsd/src/commits/master/lib/libc/regex) this code isn't really maintained.

Also, their [man page](https://man.openbsd.org/regex.3) states `This is an alpha release with known defects.` and `regexec() performance is poor. This will improve with later releases. nmatch exceeding 0 is expensive; nmatch exceeding 1 is worse.`(rz_regex_exec). I think that we should replace this implementation completely but we can still merge this PR with limited regex support for now if we are satisfied with the rz_regex API.

XVilka (approved):
Then it's fine as is, I think. LGTM. We could try to use different implementation later.

test/db/archos/linux-x64/dbg_oo
ITAYC0HEN: I am fine with this API. Also, it wouldn't be too painful to replace the implementation with others. just need to make sure that the search functionalities will be okay still

ITAYC0HEN (approved):

PR #278, opened 2020-12-28
Import patches from radare2

XVilka:
DO NOT SQUASH

**Detailed description**

Additionally, I added a simple macros `HT_PTR(x)` to be expanded into `(ut64)(size_t) x` since it's heavily used for portability reasons when calling the `ht_*` and `sdb_*` API.


librz/analysis/block.c
```diff
@@ -839,34 +839,34 @@ static bool automerge_predecessor_successor_cb(ut64 addr, void *user) {
...
-		ht_up_insert (ctx->predecessors, (ut64) block, ctx->cur_pred);
+		ht_up_insert (ctx->predecessors, HT_PTR(block), ctx->cur_pred);
 	}
 	return true;
 }
 
 static bool automerge_get_predecessors_cb(void *user, const ut64 k, const void *v) {
 	AutomergeCtx *ctx = user;
-	const RzAnalysisFunction *fcn = (const RzAnalysisFunction *)k;
+	const RzAnalysisFunction *fcn = (const RzAnalysisFunction *)(size_t)k;
```
wargio: why that `(size_t)` is needed? k is already 64 bits

librz/debug/p/native/linux/linux_debug.c
```diff
@@ -92,7 +92,7 @@ int linux_handle_signals(RzDebug *dbg, int tid) {
 		//ptrace (PTRACE_SETSIGINFO, dbg->pid, 0, &siginfo);
 		dbg->reason.type = RZ_DEBUG_REASON_SIGNAL;
 		dbg->reason.signum = siginfo.si_signo;
-		dbg->stopaddr = (ut64)siginfo.si_addr;
+		dbg->stopaddr = (ut64)(size_t)siginfo.si_addr;
```
wargio: my this cast is needed?

librz/debug/p/debug_native.c
```diff
@@ -1510,7 +1510,7 @@ static int rz_debug_native_map_protect (RzDebug *dbg, ut64 addr, int size, int p
 		"sc@syscall(%d);\n"
 		"main@global(0) { sc(%p,%d,%d);\n"
 		":int3\n"
-		"}\n", num, (void*)addr, size, io_perms_to_prot (perms));
+		"}\n", num, (void*)(size_t)addr, size, io_perms_to_prot (perms));
```
wargio: why this is needed?

wargio (changes requested):

XVilka:
It's not 64bit on ARMv7 I suppose.

librz/analysis/block.c
XVilka: On ARMv7 the pointer size will be 32bit, thus it will lead to the warning "casting pointer from integer of different size" or something like that.

librz/debug/p/debug_native.c
XVilka: Same as above.

wargio (approved):
resolved

librz/analysis/block.c
```diff
@@ -839,34 +839,34 @@ static bool automerge_predecessor_successor_cb(ut64 addr, void *user) {
 		return true;
 	}
 	bool found;
-	RzAnalysisBlock *pred = ht_up_find (ctx->predecessors, (ut64)block, &found);
+	RzAnalysisBlock *pred = ht_up_find (ctx->predecessors, HT_PTR(block), &found);
```
ret2libc: I think this is just wrong.. :/ Instead of using this new `HT_PTR` thing we should just use `ht_pp_find`, which has a `void *` as key and a `void *` as value. By doing so you don't need to convert between ut64 and void * .

ret2libc (approved):
`HT_PTR` is not a big deal, just a shorter way to write `(ut64)(size_t)`, but I think 1) it is a bad name in some cases, when for example HT is not used at all 2) in many cases it is possible to use `ht_pp` instead of `ht_up` to solve the problem entirely.

Anyway, ok for now since we haven't really changed these parts of the code so far.

eagleoflqj:
> `HT_PTR` is not a big deal, just a shorter way to write `(ut64)(size_t)`, but I think 1) it is a bad name in some cases, when for example HT is not used at all 2) in many cases it is possible to use `ht_pp` instead of `ht_up` to solve the problem entirely.
> 
> Anyway, ok for now since we haven't really changed these parts of the code so far.

`ht_pp` deals with string instead of pointer. Can't replace `ht_up`.
https://github.com/rizinorg/sdb/blob/26c300a8cf3ddcae2af1bc28a458b06748ce67c7/src/ht_pp.c#L5-L9

ret2libc:
> > `HT_PTR` is not a big deal, just a shorter way to write `(ut64)(size_t)`, but I think 1) it is a bad name in some cases, when for example HT is not used at all 2) in many cases it is possible to use `ht_pp` instead of `ht_up` to solve the problem entirely.
> > Anyway, ok for now since we haven't really changed these parts of the code so far.
> 
> `ht_pp` deals with string instead of pointer. Can't replace `ht_up`.
> https://github.com/rizinorg/sdb/blob/26c300a8cf3ddcae2af1bc28a458b06748ce67c7/src/ht_pp.c#L5-L9

No, `ht_pp` deals with pointers. It's just that `ht_pp_new`, `ht_pp_new0` and `ht_pp_new_size` by default use strings for convenience and because there were supposed to be many use cases for strings, however there is a `ht_pp_new_opt` function where you can pass a `HtPPOptions` pointer where you define how the Ht should behave. In particular you probably want to set at least `hashfn` to compute the hash used in the Ht and `cmp` function to compare the values stored in the hashtable. All other fields can probably be NULL/0 but of course it depends on the use case.

yossizap (approved):
LGTM

XVilka:
I could skip the commit in the question then in favor of the later, better solution if you prefer.

XVilka:
@ret2libc @wargio @eagleoflqj merged without the ARMv7 fixes and `HT_PTR`. This could be fixed differently and separately.

wargio:
ok!

PR #281, opened 2020-12-29
ACCESS_CMP: Prevent compiler from assuming expression signedness

kazarmy:

**Detailed description**

This pr fixes the following failure in #260 (https://github.com/rizinorg/rizin/runs/1617745700#step:15:87). It's cherry-picked from b48c93f1e4bc736752ef31be8d54494e3f0d0dbc since it can affect optimized builds outside #260:

test_analysis_var-fail

Apparently, it's currently only relevant because of optimization and the following line in the unit tests. Not sure why you'd want to relocate a function to the higher half of memory tbh:
https://github.com/rizinorg/rizin/blob/cf5334773037b0403d5a9603305499d0ec42780a/test/unit/test_analysis_var.c#L165

**Test plan**

All builds are green.

**Closing issues**

...

### DIFF
--- a/librz/analysis/var.c
+++ b/librz/analysis/var.c
@@ -6,7 +6,7 @@
 #include <rz_core.h>
 #include <rz_list.h>
 
-#define ACCESS_CMP(x, y) ((x) - ((RzAnalysisVarAccess *)y)->offset)
+#define ACCESS_CMP(x, y) ((st64)((ut64)(x) - (ut64)((RzAnalysisVarAccess *)y)->offset))
 
 RZ_API bool rz_analysis_var_display(RzAnalysis *analysis, RzAnalysisVar *var) {
 	char *fmt = rz_type_format (analysis->sdb_types, var->type);


XVilka (approved):
The fix makes sense but the need for it is indeed questionable.

kazarmy:
@ret2libc If you find any holes in the code's logic or in my logic (I may not be as knowledgeable as I think -- this is not a troll), just let me know. But as long as

1. The system running rizin uses two's complement (which according to https://en.wikipedia.org/w/index.php?title=Signed_number_representations&oldid=993293474#History is basically every contemporary system out there).
2. The _difference_ is inside the `st64`'s range (though if you're doing `0x8000000000000000 - 0`, I'll be asking what on earth are you doing there).

everything should be fine.

karliss:
Please just stop using (x-y) as comparison operator it doesn't matter what casts you do. It is a cool hack that works in limited situations. I consider it acceptable if you are comparing bytes where sufficient range is guaranteed due to integer promotion rules.  But having a sorting function or other comparison based algorithm which breaks if you use the full range permitted by chosen integer type because you are too lazy to write few more characters is bad idea.

karliss:
Solutions that work always:

* `(x > y) - (x < y)` 
* pair of ternary operators
* Use boolean comparison operator, which can simply be implemented as x < y.

kazarmy:
> - Use boolean comparison operator, which can simply be implemented as x < y.

An expression using just `<` isn't a proper argument to `rz_vector_lower_bound` as used here:
https://github.com/rizinorg/rizin/blob/18217491d40c744f7a95657e3433ec0a609ace28/librz/analysis/var.c#L313
though it might work.

----

> - (x > y) - (x < y)
> - pair of ternary operators

I couldn't afford the brainpower to verify these more complicated expressions, so I just took the simplest expression that worked.

ret2libc:
I agree with what @karliss said and I like `(x > y) - (x < y)` the most.

kazarmy:
Ok tried that, but neither

```c
#define ACCESS_CMP(x, y) (((x) > ((RzAnalysisVarAccess *)y)->offset) - ((x) < ((RzAnalysisVarAccess *)y)->offset))
```
nor
```c
#define ACCESS_CMP(x, y) (((ut64)(x) > (ut64)((RzAnalysisVarAccess *)y)->offset) - ((ut64)(x) < (ut64)((RzAnalysisVarAccess *)y)->offset))
```

work under either gcc or clang `-O2` optimization (Ubuntu 20.04).

And that is one ugly expression.

kazarmy:
The first expression doesn't work because when `x = 0x8000000000000110` and `y = 0x7fffffffffff0120` and both `x` and `y` are `st64`:
1. `x` is negative because the MSB is set.
2. `y` is positive.
3. `(x > y) - (x < y)` == `(-ve > +ve) - (+ve < -ve)` == `(0) - (1)` == `-1` which is not what we want.

Idk why the second expression doesn't work though.

kazarmy:
The first expression doesn't work because when `x = 0x8000000000000110` and `y = 0x7fffffffffff0120` and both `x` and `y` are `st64`:
1. `x` is negative because the MSB is set.
2. `y` is positive.
3. `(x > y) - (x < y)` == `(-ve > +ve) - (-ve < +ve)` == `(0) - (1)` == `-1` which is not what we want.

Idk why the second expression doesn't work though.

kazarmy:
Surprisingly, the second expression doesn't work even without optimization, possibly because the semantics have changed. In the absence of a hypothesis as to why it doesn't work, I'm going to declare this pr the least evil option, though a cast is an operator with defined semantics just like `+` and `-` so it's not that evil.

karliss:
>  (x > y) - (x < y) == (-ve > +ve) - (-ve < +ve) == (0) - (1) == -1 which is not what we want.

I thought that's exactly what we want. If x=0x8000000000000110 is really supposed to represent the corresponding negative number and the second number is supposed to represent positive number then returning -1 is correct result.  Just like with 1 and 5, 1-5=-4 you get negative number when first one is smaller.

if those are supposed to be positive numbers and both of them are stored as such it also works correctly.
`(x>y)-(x<y)==(0x8000000000000110 > 0x7fffffffffff0120) - (0x8000000000000110 < 0x7fffffffffff0120)==1-0 = 1` again correct result.

I think the first problem here might be clarfiying what those numbers are supposed to represent and if their data type matches what they are representing.

karliss:
Looked at the code, seems like both are indeed st64 both from the logical and variable declaration perspective. If it fails with first expression then it means that something else might be broken. I will look into this.

karliss:
Screenshot from 2020-12-30 17-44-28

This is supposed to be array sorted by offset(st64) . Not surprising that binary search doesn't work correctly when array isn't correctly sorted. Will continue debugging to find out where the incorrect sorting comes from.

kazarmy:
Emmm ok the first entry isn't in its proper position and has a really high offset. Good luck!

kazarmy:
I would just like to add that the scope of this pr is just to prevent different behavior of https://github.com/rizinorg/rizin/blob/18217491d40c744f7a95657e3433ec0a609ace28/librz/analysis/var.c#L313 when under `-O2` optimization. It's not an attempt to fix what ails `librz/analysis/var.c`.

kazarmy:
I would just like to add that the scope of this pr is just to prevent different behavior of https://github.com/rizinorg/rizin/blob/18217491d40c744f7a95657e3433ec0a609ace28/librz/analysis/var.c#L313 when under `-O2` optimization compared to without optimization. It's not an attempt to fix what ails `librz/analysis/var.c`.

karliss:
I don't think that hiding test failures by piling bugs on top of each other so that binary search "works" on unsorted array is helpful. Assuming that I didn't make an error while doing additional testing it produces unsorted array with all 3 versions of comparison operator, just with the one you committed test doesn't trigger test failure.

If the test is broken it needs to be fixed or marked as broken not changed so that you don't see that it's broken.

kazarmy:
> ... just with the one you committed test doesn't trigger test failure.

That's probably because it _is_ a drop-in replacement for the original, except for the following line https://github.com/rizinorg/rizin/blob/cfc57aef3cc634eb291e8850975db65ac7ada1b1/librz/include/rz_vector.h#L163 where it exhibits _expected_ behavior (as defined by tests) even under the combination of optimization and signed underflow when doing `x - y` (if `x` and `y` were signed).

> If the test is broken it needs to be fixed or marked as broken not changed so that you don't see that it's broken.

For this particular case, I actually prefer that the test author do the fixing (if it needs to be fixed).

PR #285, opened 2021-01-02
SWIG Bindings

thestr4ng3r:


XVilka:
1. I think we should put basic tests of the resulting bindings to check how it works, even in this PoC, for a couple functions at least.
2. It might be useful to check something non-Python, Lua or JavaScript.

XVilka:
@xarkes regarding the java plugin, see @wargio Java refactoring, maybe could help: https://github.com/rizinorg/rizin/pull/804

xarkes:
Will check that later - I think the main issue right now is that Rizin API is not very well suited for such automated generation.
A first required step is to improve its API so we can move forward with fancy bindings.

XVilka:
Closing this since @wingdeans works on this anyway

PR #288, opened 2021-01-03
Update building instruction in README

ITAYC0HEN:
**Detailed description**

Slight formatting of the building instructions in the README.

I suggest to decide what we want to keep

Either (v1)
```
$ meson build
$ ninja -C build
$ sudo ninja -C build install
```

or (v2)
```
$ meson build
$ meson compile -C build
$ sudo meson install -C build
```

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -37,8 +37,8 @@ distribution in `/usr`.
 
 ```
 $ meson build
-$ ninja -C build # or `meson compile -C build`
-$ sudo ninja -C build install # or `sudo meson install -C build`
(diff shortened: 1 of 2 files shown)


ret2libc (commented):
As you realized, `meson` does not have the `uninstall` option, so it's probably better to stay with `ninja`.

Maijin (commented):
Please also update https://github.com/rizinorg/rizin/blob/dev/BUILDING.md and the rzbook.

ITAYC0HEN:
I want to keep both options in building.md. So I only improved ux in formatting

Maijin (approved):

Issue #272, opened 2020-12-27
Zignatures search during aaa gives wrong result


p371jah:
Zignature search was tested on a binary builded from this source:
```c
#include <stdio.h>

int main() {
	char buf[5];
	FILE *f = fopen ("test", "r");
	fseek (f, 5, 1);
	fread (buf, 5, 1, f);
	fclose (f);
}
```
This is how `main` looks in the related object file. Ideally I want to see almost the same result after providing `rizin` appropriate zignatures of functions `fopen`, `fseek`, `fread` and `fclose` from `stdio` lib. 
```asm
[0x00000000]> pdf @ sym.main
            ;-- section..text_mn_4:
            ;-- .text$mn:
            ;-- $LN4:
/ 111: int sym.main (int argc, char **argv, char **envp);
|           ; var int64_t var_20h @ rsp+0x20
|           ; var int64_t var_28h @ rsp+0x28
|           ; arg int argc @ rdi
|           ; arg char **argv @ rsi
|           0x00000130      push rbx                                   ; [04] -r-x section size 111 named .text_mn_4
|           0x00000132      sub rsp, 0x30
|           0x00000136      mov rax, qword [sym.imp.__security_cookie] ; [0xfc0:8]=0
|           0x0000013d      xor rax, rsp
|           0x00000140      mov qword [var_28h], rax
|           0x00000145      lea rdx, sym.___C__01KDCPPGHE_r            ; 0xdf0 ; "r"; RELOC 32 ??_C@_01KDCPPGHE@r@ @ 0x00000df0
|           0x0000014c      lea rcx, sym.___C__04CEJDCDCH_test         ; 0xe00 ; "test"; RELOC 32 ??_C@_04CEJDCDCH@test@ @ 0x00000e00
|           0x00000153      call fopen
|           0x00000158      mov edx, 5
|           0x0000015d      mov rcx, rax
|           0x00000160      mov rbx, rax
|           0x00000163      lea r8d, [rdx - 4]
|           0x00000167      call fseek
|           0x0000016c      mov edx, 5
|           0x00000171      lea rcx, [var_20h]
|           0x00000176      mov r9, rbx
|           0x00000179      lea r8d, [rdx - 4]
|           0x0000017d      call fread
|           0x00000182      mov rcx, rbx
|           0x00000185      call fclose
|           0x0000018a      xor eax, eax
|           0x0000018c      mov rcx, qword [var_28h]
|           0x00000191      xor rcx, rsp
[...]

PR #292, opened 2021-01-03
Detect wrapped function in z/f

p371jah:

**Detailed description**

Related to #272 
`z/f` is more accurate than `z/` and IMO it should be used by default in `aaa`. But `z/f` can't detect wrapped function if a wrapper jumps unconditionally to it.
Proposed solution: 
- check if the first BB of function jumps unconditionally to the next
- search from addr of the next BB
- search from addr of the first BB

Potential issues:
- creating new functions while iterating over the function list
- ugly way to split a function, no error handling
- accessing objects after splitting

Before (`aa` + `zo` + `z/f`)
```asm
[0x140001394]> pdf @ 0x1400033cc
        ╎   ; CALL XREF from main @ 0x140001033
┌ 196: fcn.1400033cc (int64_t arg1, int64_t arg2);
│ bp: 0 (vars 0, args 0)
│ sp: 4 (vars 4, args 0)
│ rg: 2 (vars 0, args 2)
│       ╎   0x1400033cc      mov r8d, 0x40                             ; '@' ; 64
│       └─< 0x1400033d2      jmp 0x140003310
[0x140001394]> pd 4 @ 0x140003310
│           ; CODE XREF from fcn.1400033cc @ 0x1400033d2
│           0x140003310      mov qword [var_10h], rbx
│           0x140003315      mov qword [var_18h], rsi
│           0x14000331a      push rdi
│           0x14000331b      sub rsp, 0x30

```
After
```asm
[0x140001394]> pdf @ 0x1400033cc
        ╎   ; CALL XREF from main @ 0x140001033
┌ 11: fcn.1400033cc ();
│       ╎   0x1400033cc      mov r8d, 0x40                             ; '@' ; 64
└       └─< 0x1400033d2      jmp sym.___common_fsopen_D__YAPEAU_iobuf__QEBD0H_Z
[0x140001394]> pd 4 @ 0x140003310
            ; CODE XREF from fcn.1400033cc @ 0x1400033d2
            ;-- sign.bytes.sym.___common_fsopen_D__YAPEAU_iobuf__QEBD0H_Z_0:
            ;-- fcn.140003310:
┌ 185: sym.___common_fsopen_D__YAPEAU_iobuf__QEBD0H_Z (int64_t arg1, int64_t arg2);
│ bp: 0 (vars 0, args 0)
│ sp: 4 (vars 4, args 0)
│ rg: 2 (vars 0, args 2)
│           0x140003310      mov qword [var_10h], rbx
│           0x140003315      mov qword [var_18h], rsi
│           0x14000331a      push rdi
│           0x14000331b      sub rsp, 0x30
```

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/librz/core/cmd_zign.c
+++ b/librz/core/cmd_zign.c
@@ -580,7 +580,7 @@ struct ctxSearchCB {
 	bool rad;
 	int count;
 	const char *prefix;
-	const RzAnalysisBlock *wrap_bb;
+	const RzAnalysisBlock *wrapper_bb;
 };
 
 static void apply_name(RzCore *core, RzAnalysisFunction *fcn, RzSignItem *it, bool rad) {
@@ -675,13 +675,15 @@ static int searchHitCB(RzSignItem *it, RzSearchKeyword *kw, ut64 addr, void *use
 	apply_flag (ctx->core, it, addr, kw->keyword_length, kw->count, ctx->prefix, ctx->rad);
 	RzAnalysisFunction *fcn = rz_analysis_get_fcn_in (ctx->core->analysis, addr, 0);
 	// TODO: create fcn if it does not exist
-	if (fcn && ctx->wrap_bb && ctx->wrap_bb->jump == addr) {
+	if (fcn && ctx->wrapper_bb && ctx->wrapper_bb->jump == addr) {
+		ut64 wrapper_size = ctx->wrapper_bb->size;
+		ctx->wrapper_bb = NULL;
 		RzAnalysisFunction *newfcn = rz_analysis_create_function (ctx->core->analysis, NULL, addr, RZ_ANALYSIS_FCN_TYPE_FCN, NULL);
-		int size = rz_analysis_function_realsize (fcn) - ctx->wrap_bb->size;
+		ut64 size = rz_analysis_function_realsize (fcn) - wrapper_size;
 		int flen = rz_analysis_fcn (ctx->core->analysis, newfcn, addr, size, RZ_ANALYSIS_REF_TYPE_CALL);
 		/* rz_analysis_fcn() returns RZ_ANALYSIS_RET_ERROR if BBs already exist */
 		if (flen == RZ_ANALYSIS_RET_ERROR) {
-			rz_analysis_function_resize (fcn, ctx->wrap_bb->size);
+			rz_analysis_function_resize (fcn, wrapper_size);
 		} else {
 			rz_analysis_function_delete (newfcn);
 			newfcn = NULL;
@@ -822,7 +824,7 @@ static bool search(RzCore *core, bool rad, bool only_func) {
 	const char *mode = rz_config_get (core->config, "search.in");
 	bool useBytes = rz_config_get_i (core->config, "zign.match.bytes");
 	const char *zign_prefix = rz_config_get (core->config, "zign.prefix");
-	int maxsz = rz_config_get_i (core->config, "zign.maxsz");
+	ut64 maxsz = rz_config_get_i (core->config, "zign.maxsz");
 
 	struct ctxSearchCB metsearch_ctx = { core, rad, 0, NULL };
 	RzSignSearchMetrics sm;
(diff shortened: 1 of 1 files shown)


librz/core/cmd_zign.c
```diff
@@ -674,6 +675,21 @@ static int searchHitCB(RzSignItem *it, RzSearchKeyword *kw, ut64 addr, void *use
...
 	RzAnalysisFunction *fcn = rz_analysis_get_fcn_in (ctx->core->analysis, addr, 0);
 	// TODO: create fcn if it does not exist
+	if (fcn && ctx->wrapper_bb && ctx->wrapper_bb->jump == addr) {
+		ut64 wrapper_size = ctx->wrapper_bb->size;
+		ctx->wrapper_bb = NULL;
+		RzAnalysisFunction *newfcn = rz_analysis_create_function (ctx->core->analysis, NULL, addr, RZ_ANALYSIS_FCN_TYPE_FCN, NULL);
+		ut64 size = rz_analysis_function_realsize (fcn) - wrapper_size;
+		int flen = rz_analysis_fcn (ctx->core->analysis, newfcn, addr, size, RZ_ANALYSIS_REF_TYPE_CALL);
+		/* rz_analysis_fcn() returns RZ_ANALYSIS_RET_ERROR if BBs already exist */
+		if (flen == RZ_ANALYSIS_RET_ERROR) {
```
p371jah, line 685: @thestr4ng3r maybe you know a better way to split a function

thestr4ng3r: I think it's very questionable whether the zignature search should modify functions and basic blocks at all. Especially `rz_analysis_fcn()` should be avoided at all costs since it is the actual analysis loop that could mess up a lot of things. Are there any other cases where more than just adding a flag, renaming and maybe applying types happens? Otherwise I would suggest to just add a flag in the middle of the function and leave it at that.

Anyway, in case we do decide to split functions here, you first have to be aware that `rz_analysis_get_fcn_in()` used above is a very fuzzy and half-deprecated thing. Since both functions and basic blocks can overlap, considering a single function containing `addr` here is conceptually wrong. Intuitively I would say `rz_analysis_get_function_at()` should be used above and for your wrapped functions, you will want to save the wrapper function next to the `wrapper_bb`.

Once you really do have the correct function that at this point contains the wrapper bb and the rest of the function, as well as a reference to this wrapper bb, you can call `rz_analysis_function_relocate()` on the function and set its entrypoint to the target of the wrapper bb. Then you can remove the block from the function with `rz_analysis_function_remove_block()`, create a new one for the wrapper function with `rz_analysis_create_function()` and add the bb to that one with `rz_analysis_function_add_block()`. Basically the other way around as you currently did it.
Also, be aware of the refcounting of the blocks! If you didn't explicitly call `rz_analysis_block_ref()` and you didn't get the block reference out of a function that did this for you, after removing the block from the last function, it would be freed automatically!

librz/core/cmd_zign.c
```diff
@@ -858,8 +874,23 @@ static bool search(RzCore *core, bool rad, bool only_func) {
...
-				int fcnlen = rz_analysis_function_realsize (fcni);
-				int len = RZ_MIN (core->io->addrbytes * fcnlen, maxsz);
+				ut64 fcnlen = rz_analysis_function_realsize (fcni);
+				ut64 len = RZ_MIN (core->io->addrbytes * fcnlen, maxsz);
+				if (rz_list_length (fcni->bbs) > 1) {
+					const RzAnalysisBlock *bb = NULL;
+					bb = rz_list_first (fcni->bbs);
+					if (bb->addr != fcni->addr) {
+						bb = rz_analysis_get_block_at (core->analysis, fcni->addr);
+					}
```
thestr4ng3r, line 884: I would rather get the block like this:
```c
bb = rz_analysis_get_block_at (core->analysis, fcni->addr);
if (rz_list_contains (bb->fcns, fcni)) {
	...
}

librz/core/cmd_zign.c
Maijin: @thestr4ng3r In future I could see zignature also containing color highlights of BBs, comments, calling convention, signatures, local flags, and much more that, could really be handy imho (to be discussed ofc)

thestr4ng3r: Perhaps it should be a two-step process in some way. First searching and then optionally applying to functions.

p371jah: I understand that my code can break some internal data. That's why I have made draft PR. Anyway now I think it was bad idea to sent it. It would be better to implement a complete solution, not an interim one.

PR #293, opened 2021-01-03
Fix crash after tracing an invalid opcode ##trace

yossizap:

**Detailed description**

`rz_debug_trace_ins_before` fails when an instruction is invalid(according to `rz_analysis_op`) and sets `dbg->cur_op` to NULL. This lead to a segfault in `rz_debug_trace_ins_after` after stepping since it was trying to iterate over cur_op->access.
An invalid instruction won't be properly traced but debug will stop and an error will be printed for further investigation, the user will still be able to step over it backwards and forwards.

**Test plan**

This is a bit problematic to add a test to since it depends on an invalid instruction that won't crash the debugee and I am planning to add support for the unsupported instructions in another PR. You can test by tracing from debug start to main in Linux and you'll probably find some unsupported CET instruction: `ood`->`dts+`->`db main`->`dc`

### DIFF
--- a/librz/debug/debug.c
+++ b/librz/debug/debug.c
@@ -973,7 +973,7 @@ RZ_API int rz_debug_step(RzDebug *dbg, int steps) {
 			dbg->session->maxcnum++;
 			dbg->session->bp = 0;
 			if (!rz_debug_trace_ins_before (dbg)) {
-				eprintf ("trace_ins_before: failed");
+				eprintf ("trace_ins_before: failed\n");
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


ret2libc (approved):
Ok, so the "proper fix", if I get it right, is to implement the instruction in the disassembler?

yossizap:
To this case, yes. But we should be able to skip unknown instructions without crashing regardless.

wargio (approved):

ITAYC0HEN:
> To this case, yes. But we should be able to skip unknown instructions without crashing regardless.

That is actually a fantastic point. On the same note, I'd offer to do the same with External calls, and trigger an event to say that it is unsupported, and set return value to 0 or something. This will be a life saver in many ways

XVilka (approved):

PR #294, opened 2021-01-04
Migration to PJ and various memleak fixes

XVilka:
DO NOT SQUASH

**Detailed description**


wargio (commented):
can't we use the PJ pointer to check if the output is JSON or not, instead of the `int mode` variable?

Maijin:
Can you add a test for `arS`?

officialcjunior:
> Can you add a test for `arS`?

How exactly should a file be opened to test that?

XVilka:
@wargio we can, but it's out of the scope of this PR. Can be done later.

XVilka:
Ugnore the `jni.h` failed test, it will be green once https://github.com/rizinorg/rizin-testbins/pull/4 is merged.

librz/debug/dsession.c
```diff
@@ -576,6 +580,7 @@ static bool deserialize_checkpoints_cb(void *user, const char *cnum, const char
 	// Extract RzRegArena's from "registers"
 	const RJson *regs_json = rz_json_get (chkpt_json, "registers");
 	if (!regs_json || regs_json->type != RZ_JSON_ARRAY) {
+		free (json_str);
```
ret2libc, line 583: there should be also `rz_json_free (chkpt_json);` after this line AFAIK

test/unit/test_str.c
```diff
@@ -558,14 +558,20 @@ bool test_rz_str_encoded_json(void) {
 
 	const char *array = rz_str_encoded_json (invalidJsonString, len, PJ_ENCODING_STR_ARRAY);
 	mu_assert_streq (array, "084,104,105,115,032,105,115,032,109,121,032,226,032,115,097,109,112,108,101,032,194,169,032,115,116,114,105,110,103,010", "string as array of uchar");
```
ret2libc: use `mu_assert_streq_free` instead of `mu_assert_streq` + `free`.

librz/cons/line.c
```diff
@@ -41,6 +41,7 @@ RZ_API void rz_line_free(void) {
 	// XXX: prompt out of the heap?
 	free ((void *)I.prompt);
 	I.prompt = NULL;
+	//rz_list_free (I.kill_ring);
```
ret2libc, line 44: why this commented line?

ret2libc: I think it is right to uncomment it.

XVilka: It causes double free segfault then.

ret2libc: Why? Does it happen in radare2 as well? Maybe it should be `rz_list_free`+`I.kill_ring=NULL`? (though this is not *the* solution)

XVilka: It doesn't happen in radare2:  https://github.com/radareorg/radare2/commit/06b735441d299e90695235495982354bef1fcc41 - note the green status.

XVilka: I can just remove this line if you want.

ret2libc: Adding this free is right in theory... Could you try uncommenting it so I can see where exactly the double free happens?

XVilka: What about merging this first, then try? It's grown quite big.

PR #299, opened 2021-01-04
Implement `e` commands in newshell ##newshell

### LINKED ISSUE #115: Autocomplete for `e asm.` doesn't work
### Work environment

| rizin -v full output, **not truncated** (mandatory)  | https://github.com/rizinorg/rizin/commit/c5021f7c8a60f9c95b29a3f75922b17f3ec38bc3

### Expected behavior

`e asm.` then press `<TAB>` it should show available options or autocomplete their names when you start typing, like it was before the new shell.

### Actual behavior

No autocompletion.

ret2libc:
  - [x] Update https://book.rizin.re/configuration/evars.html: see https://github.com/rizinorg/book/pull/12

**Detailed description**

This PR converts `e` commands to newshell. Some commands  were changed to make them more consistent with others. In particular:
- `e??` was removed and replaced with `el`
- `ed` was removed because it edits only one configuration file while Rizin uses multiple from different locations and merges them all together
- `evj` became `elJ` (`J` used as a suffix to mean "verbose json")
- `e.` removed (not sure why one would not want to use a space but use a `.`)
- `e,` removed because `e` now supports multiple get/sets without the need for an additional command
- autocompletion for evals implemented (also options and booleans are autocompleted).
- add a `flags` type to `RzCmdArg` to specify if a command is an array, it's the last, etc. In this way we can easily have arrays of various types, without the need to creates multiples types (e.g. ARRAY_STRING, ARRAY_EVAL, etc.)

**Test plan**

I expect some existing tests to fail, I will update them once I see which are failing.

**Closing issues**

Closes https://github.com/rizinorg/rizin/issues/115


Maijin (approved):
It makes no difference to me, If it makes for simpler code to maintain and easier check to see if a user has type something wrong, good enough for me.

librz/core/cmd_eval.c
```diff
@@ -660,3 +667,135 @@ RZ_IPI int rz_cmd_eval(void *data, const char *input) {
...
+			return RZ_CMD_STATUS_ERROR;
+		}
+		char *key = rz_list_get_n (l, 0);
+
+		if (llen == 1 && rz_str_endswith (key, ".")) {
+			// no value was set, only key with ".". List possible sub-keys.
+			rz_config_list (core->config, key, false);
+		} else if (llen == 1) {
+			// no value was set, show the value of the key
+			rz_cons_printf ("%s\n", rz_config_get (core->config, key));
```
ret2libc: Check first if the key is valid and print a message if not. (not `(null)`).

librz/core/cmd_eval.c
```diff
@@ -660,3 +667,135 @@ RZ_IPI int rz_cmd_eval(void *data, const char *input) {
...
+	for (i = 1; i < argc; i++) {
+		RzList *l = rz_str_split_duplist_n (argv[i], "=", 2, false);
+		if (!l) {
+			return RZ_CMD_STATUS_ERROR;
+		}
+		size_t llen = rz_list_length (l);
+		if (!llen) {
+			return RZ_CMD_STATUS_ERROR;
+		}
+		char *key = rz_list_get_n (l, 0);
```
ret2libc, line 659: check if `key` is empty and print a warning. Moreover, suggest the correct syntax without spaces.

wargio (approved):

XVilka (approved):
Code looks fine, I tested and it works. Much better than before the newshell actually. Good job!
I noticed somewhat unrelated bug though:
If you run `ed` it tries to open `${HOME}/.rizinrc` instead of the actually used config. It should open the corresponding config that is currently used. In my case it's `${HOME}/.config/rizin/rizinrc`.

binrz/rz-test/run.c
```diff
@@ -89,9 +89,9 @@ static char *convert_win_cmds(const char *cmds) {
 static RzSubprocessOutput *run_rz_test(RzTestRunConfig *config, ut64 timeout_ms, const char *cmds, RzList *files, RzList *extra_args, bool load_plugins, RzTestCmdRunner runner, void *user) {
 	RzPVector args;
 	rz_pvector_init (&args, NULL);
-	rz_pvector_push (&args, "-escr.utf8=0");
-	rz_pvector_push (&args, "-escr.color=0");
-	rz_pvector_push (&args, "-escr.interactive=0");
+	rz_pvector_push (&args, "-e scr.utf8=0");
```
ret2libc: Why? I think this is wrong, because by doing this there are actually two arguments ("-e" and "scr.utf8=0"). I think there should be no problem in keeping these lines as they were before. @XVilka

ret2libc:
> I noticed somewhat unrelated bug though:
> If you run `ed` it tries to open `${HOME}/.rizinrc` instead of the actually used config. It should open the corresponding config that is currently used. In my case it's `${HOME}/.config/rizin/rizinrc`.

Let's just kill the `ed` command. Rizin considers all rizinrc files in multiple places, so it's not clear what this command should modify.

PR #300, opened 2021-01-04
Use /MT for VS2019 build

kazarmy:

**Detailed description**

This is an attempt to reobtain a sub-1-hour VS2019 build on AppVeyor.

**Test plan**

All builds are green.

**Closing issues**

...


XVilka:
By the way, is it possible to split somehow these AppVeyor long lines to the multiline somehow? Reading differences and the file is a pain :/

ret2libc (commented):
I would think that buildtype=release also sets debug=false. Isn't that so?

kazarmy:
> I would think that buildtype=release also sets debug=false. Isn't that so?

Yes but it also sets `-Doptimization=3` which sets `/Ob2` (assuming I'm reading [1] correctly) which does more aggressive inlining, which might break the CPU's instruction cache and actually slow things down.

[1] https://github.com/mesonbuild/meson/blob/67ee65f47ed0872040fcce3db68a34005af9a03d/mesonbuild/compilers/mixins/visualstudio.py#L81-L88

kazarmy:
> By the way, is it possible to split somehow these AppVeyor long lines to the multiline somehow? Reading differences and the file is a pain :/

Might be possible with `^` (https://ss64.com/nt/syntax-esc.html)

kazarmy:
Hmm closing this because performance is much worse than `release` 😞

ret2libc:
On my windows system the fastest build is without `default-library=static` and without `static-runtime`. Moreover, I noticed that `ninja` auto-detects too many jobs on windows and I have to be careful sometimes because otherwise it consumes all memory on my system. So maybe one thing to try is to set `-j` to a reasonable value for appveyor.

Moreover, I tried in the paste with `msbuild` instead of `ninja`, but it was actually even worse in terms of compilation time!

kazarmy:
> Moreover, I tried in the paste with msbuild instead of ninja, but it was actually even worse in terms of compilation time!

My experience tells me that building rizin with `msbuild` is going to be like building cutter with `ninja`, but on the bright side ... https://xkcd.com/303/

kazarmy:
> So maybe one thing to try is to set -j to a reasonable value for appveyor.

It appears that compilation still completes in a reasonable amount of time, but if you hover the mouse over the `[**]` rz-test output lines in the latest AppVeyor log on `dev`, you'll notice that the timestamps increase in a somewhat worrying manner, at least compared to when the VS2019 build was sub-50 minutes.

kazarmy:
@ret2libc You might want to check whether `rizin -` has slowed down on Windows.

ret2libc:
> @ret2libc You might want to check whether `rizin -` has slowed down on Windows.

Slowed down compared to when?

kazarmy:
> Slowed down compared to when?

cfc57aef3cc634eb291e8850975db65ac7ada1b1

XVilka:
Could be one of these:
- https://github.com/rizinorg/rizin/commit/d25cea5420e306af32f5e99d7417aa3e1d45accf
- https://github.com/rizinorg/rizin/commit/7316b71619a8025e076f6462a6141badb40b8628

ret2libc:
> Could be one of these:
> 
>     * [d25cea5](https://github.com/rizinorg/rizin/commit/d25cea5420e306af32f5e99d7417aa3e1d45accf)
> 
>     * [7316b71](https://github.com/rizinorg/rizin/commit/7316b71619a8025e076f6462a6141badb40b8628)

These two changes are mainly about mmap and the io_default plugin and none of them is used when doing `rizin -`.

ret2libc:
But it anyway looks like some of my changes drastically increased the time for one of the appveyor jobs :(

ret2libc:
https://github.com/rizinorg/rizin/commit/7316b71619a8025e076f6462a6141badb40b8628#diff-30d8f6be6320feeacf686be94f48c70869b52630e01ea625f0f15adc0d57c3e4R131 I think this is the problem. Not sure why I added it. Let's try to remove it.

XVilka (approved):

kazarmy:
LGTM ... let's just do 1 more check...

kazarmy:
Meh ... let's just do the check on dev since it's red anyway

ret2libc:
Ok that was it. Times for appveyor dropped again to ~30/40 mins, instead of ~1hr

PR #309, opened 2021-01-05
More Web removal

Maijin:
More WebUI removal. Still a bunch of http eval intricated linked with rz-agent and rtr.


wargio:
isn't the proxy stuff needed for the PDB ?

Maijin:
I didn't remove the http backend, only rtr

ret2libc:
Mmm are you completely removing the possibility to run commands over HTTP? @thestr4ng3r I remember you had an opinion about this.

@Maijin is it still possible to contact a remote rizin somehow?

XVilka:
Having some of the HTTP stuff is required for http:// pipes in rz-pipe.

Maijin:
@wargio This variable isn't used in pdb related code.
@ret2libc  There are still a lot of code related to remote connection, see rtr.c (still tons of http code), rz-agent.c (still tons of http dup code), this is only rtr_http.c related aka =h and =H

Maijin:
@XVilka I haven't removed io_http nor socket_http.c

wargio:
ok, so for me i'm ok with this.
for the http server i think it is better to use an external library

XVilka:
@wargio we have an issue for this: https://github.com/rizinorg/rizin/issues/168

XVilka (approved):
I am ok with the removal if it doesn't break `http://` pipes for rz-pipe and @thestr4ng3r has no objections.

ITAYC0HEN (approved):

librz/core/rtr_http.c
```diff
@@ -1,562 +0,0 @@
...
-								http_logf (core, "File '%s' not found\n", path);
-								rz_socket_http_response (rs, 404, "File not found\n", 0, headers);
-							}
-						}
-						free (path);
-					}
-				} else {
-					rz_socket_http_response (rs, 403, "", 0, NULL);
-				}
-			} else if (!strncmp (rs->path, "/cmd/", 5)) {
```
thestr4ng3r, line 274: This does break rz-pipe with `http://`. Removal of http while still keeping rz-pipe intact would mean only removing the file-serving but keeping the cmd stuff.
We might want to reconsider keeping http rz-pipe since there is also the tcp one but imo this should be discussed.

ret2libc:
@thestr4ng3r could you provide directions to move this forward?

thestr4ng3r:
If we want to keep the rz-pipe functionality, the removal should probably be re-done from scratch, only removing the file-serving things.

XVilka:
@Maijin do you plan to work on this further?

Maijin:
No but this should still be done though

PR #326, opened 2021-01-06
Added regex documentation to test/README.md

yossizap:

**Detailed description**

...

**Test plan**

...

### DIFF
--- a/test/README.md
+++ b/test/README.md
@@ -115,13 +115,43 @@ Example commands tests for the other `db/` folders:
 	EOF
 	RUN
 
+It is also possible to match specific parts of the output in `EXPECT` and `EXPECT_ERR` using
+regex(with `REGEXP_OUT` and `REGEXP_ERR` respectively) in case some of the test's output is dynamic:
+
+	NAME=bp rebase
+	FILE=bins/elf/analysis/pie
+	ARGS=-d
+	CMDS=<<EOF
+	aa
+	db main
+	db~main
+	doc
+	db~main
+	EOF
+	REGEXP_OUT=([a-zA-Z="]+\s+)
+	EXPECT=<<EOF
+	x sw break enabled valid cmd="" cond="" name="main" pie"
+	x sw break enabled valid cmd="" cond="" name="main" pie"
+	EOF
+	RUN
+
+Without the regex that filtered out the non-deterministic file path and addresses, the expected output would have been the following:
+
+```
+0x566495c5 - 0x566495c6 1 --x sw break enabled valid cmd="" cond="" name="main" module="/home/user/rizin/test/bins/elf/analysis/pie"
+0x000005c5 - 0x000005c6 1 --x sw break enabled valid cmd="" cond="" name="main" module="/home/user/rizin/test/bins/elf/analysis/pie"
+```
+
 * **NAME** is the name of the test, it must be unique
 * **FILE** is the path of the file used for the test
 * **ARGS** (optional) are the command line argument passed to rizin (e.g -b 16)
 * **CMDS** are the commands to be executed by the test
-* **EXPECT** is the expected output of the test
+* **EXPECT** is the expected output of the test from stdout
+* **EXPECT_ERR** (optional) is the expected output of the test from stderr. Can be specified in addition or instead of `EXPECT`
 * **BROKEN** (optional) is 1 if the tests is expected to be fail, 0 or unspecified otherwise
 * **TIMEOUT** (optional) is the number of seconds to wait before considering the test timeout
+* **REGEXP_OUT** (optional) apply given regex on stdout before comparing the ouput to `EXPECT` (e.g. `REGEXP_OUT=([a-zA-Z]+)`)
+* **REGEXP_ERR** (optional) apply given regex on stderr before comparing the ouput to `EXPECT_ERR`
 
 You must end the test by adding RUN keyword
 
--- a/test/README.md
+++ b/test/README.md
@@ -45,52 +45,51 @@ to build Rizin).
 # Failure Levels
 
 A test can have one of the following results:
-* success: The test passed, and that was expected.
-* fixed: The test passed, but failure was expeced.
-* broken: Failure was expected, and happened.
-* failed: The test failed unexpectedly. This is a regression.
+* **success**: The test passed, and that was expected.
+* **fixed**: The test passed, but failure was expected.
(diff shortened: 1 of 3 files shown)


test/README.md
```diff
@@ -119,9 +119,12 @@ Example commands tests for the other `db/` folders:
 * **FILE** is the path of the file used for the test
 * **ARGS** (optional) are the command line argument passed to rizin (e.g -b 16)
 * **CMDS** are the commands to be executed by the test
-* **EXPECT** is the expected output of the test
+* **EXPECT** is the expected output of the test from stdout
+* **EXPECT_ERR** is the expected output of the test from stderr. Can be specified in addition or instead of `EXPECT`
```
ret2libc: (optional)

test/README.md
```diff
@@ -119,9 +119,12 @@ Example commands tests for the other `db/` folders:
...
 * **BROKEN** (optional) is 1 if the tests is expected to be fail, 0 or unspecified otherwise
 * **TIMEOUT** (optional) is the number of seconds to wait before considering the test timeout
+* **REGEX_OUT** (optional) apply given regex on stdout before comparing the ouput to `EXPECT` (e.g. `REGEX_OUT=([a-zA-Z]+)`)
```
ret2libc: Could you also make an example for clarity? Something like "suppose the output is ....... and you want to check that the word Hello is followed by a number that can change on each execution of the test, you would write the test in this way"

test/README.md
```diff
@@ -99,29 +98,59 @@ The JSON tests `db/json` are executed on 3 standard files (1 ELF, 1 MachO, 1 PE)
...
+FILE=bins/elf/analysis/pie
+ARGS=-d
+CMDS=<<EOF
+aa
+db main
+db~main
+doc
+db~main
+EOF
+REGEXP_OUT=([a-zA-Z="]+\s+)
```
ret2libc: I'm not sure I understand how this work.. Does this regex removes every char from output that is not a-z, A-Z, =, ", spaces ?

yossizap: it only matches a sequence a-z, A-Z, =, " followed by whitespace
Since we don't have negative lookahead these things get a bit tricky.. Will a simpler example be better?

ret2libc: Sorry, maybe it's just me who does not understand....

How this regex "converts"
```
 0x566495c5 - 0x566495c6 1 --x sw break enabled valid cmd="" cond="" name="main" module="/home/user/rizin/test/bins/elf/analysis/pie"
```
into 
```
x sw break enabled valid cmd="" cond="" name="main" pie"
```
?

What is more or less the operation that rz-test does if you would write it in python with the `re` module? (if you know python, otherwise just explaining in words could be fine). Sorry again, but I want to use this regex thing but i'm still not sure I understand how.

yossizap: The regex is applied and all matches are printed as a single output. I tried playing with having the user specify the number of matches but it felt error prone. This format seems intuitive to me since that's what I expect when I regex in a text editor/grep, we can improve the documentation or test format if it isn't clear.

```python
>>> import re
>>> string = "0x566495c5 - 0x566495c6 1 --x sw break enabled valid cmd=\"\" cond=\"\" name=\"main\" module=\"/home/user/rizin/test/bins/elf/analysis/pie\""
>>> res = re.findall("([a-zA-Z=\"]+\s+)", string)
>>> print(res)
['x ', 'sw ', 'break ', 'enabled ', 'valid ', 'cmd="" ', 'cond="" ', 'name="main" ']
>>> expect = "".join(res)
>>> print(expect)
'x sw break enabled valid cmd="" cond="" name="main" '
```

You will also get the same result with
```sh
echo "0x566495c5 - 0x566495c6 1 --x sw break enabled valid cmd=\"\" cond=\"\" name=\"main\" module=\"/home/user/rizin/test/bins/elf/analysis/pie\"" | grep -E "([a-zA-Z=\"]+\s+)"
```

the string `pie"`(the only part of that word which started with letters and ends with a whitespace) isn't matched in these example because the given strings were terminated without a whitespace.

XVilka (approved):

test/README.md
```diff
@@ -99,29 +98,59 @@ The JSON tests `db/json` are executed on 3 standard files (1 ELF, 1 MachO, 1 PE)
...
+0x566495c5 - 0x566495c6 1 --x sw break enabled valid cmd="" cond="" name="main" module="/home/user/rizin/test/bins/elf/analysis/pie"
+0x000005c5 - 0x000005c6 1 --x sw break enabled valid cmd="" cond="" name="main" module="/home/user/rizin/test/bins/elf/analysis/pie"
+```
 
 * **NAME** is the name of the test, it must be unique
 * **FILE** is the path of the file used for the test
 * **ARGS** (optional) are the command line argument passed to rizin (e.g -b 16)
 * **CMDS** are the commands to be executed by the test
-* **EXPECT** is the expected output of the test
+* **EXPECT** is the expected output of the test from stdout
```
ret2libc:
```suggestion
* **EXPECT** is the expected output of the test from stdout. If `REGEXP_FILTER_OUT` is used, `EXPECT` matches only the filtered output.
```

test/README.md
```diff
@@ -99,29 +98,59 @@ The JSON tests `db/json` are executed on 3 standard files (1 ELF, 1 MachO, 1 PE)
...
+* **EXPECT_ERR** (optional) is the expected output of the test from stderr. Can be specified in addition or instead of `EXPECT`
 * **BROKEN** (optional) is 1 if the tests is expected to be fail, 0 or unspecified otherwise
 * **TIMEOUT** (optional) is the number of seconds to wait before considering the test timeout
+* **REGEXP_FILTER_OUT** (optional) apply given regex on stdout before comparing the ouput to `EXPECT` (e.g. `REGEXP_FILTER_OUT=([a-zA-Z]+)`)
```
ret2libc:
```suggestion
* **REGEXP_FILTER_OUT** (optional) apply given regex on stdout before comparing the ouput to `EXPECT` (e.g. `REGEXP_FILTER_OUT=([a-zA-Z]+)`). This is similar to piping stdout to `grep -E "<regex>"` and then matching only the matched text with `EXPECT`.
```

ret2libc (commented):
Thanks a lot @yossizap ! I've made some small modifications, but I think it's clear now how to use these.

ret2libc (approved):

PR #339, opened 2021-01-10
Shell error messages improvements

ret2libc:

**Detailed description**

* When a command does not exist it shows a better error message making clear the command does not exist. This works only if the whole commands sub-tree is handled by newshell, otherwise things are redirected to the old-style handlers, which may or may not make it clear a command does not exist.
* When a user passes the wrong number of arguments the help for that command is automatically shown to the user
* When a user runs things like `wv5? <arg>` which in general might be a command with an arg, we show the error `Command 'wv5?' does not exist. Did you want to see the help? Try "wv5?" without any argument.`


Maijin:
"[...]because not existent" is not English, you may want to choose one of the following:

* [...] because non-existent
* [...] because it doesn't exist (I prefer this one)

Same with NOTEXISTING, I'd use NONEXISTING or something used in other similar projects (if that's a thing).

XVilka (approved):
Would be very helpful for users, indeed.

librz/core/cmd.c
```diff
@@ -4635,6 +4635,23 @@ DEFINE_HANDLE_TS_FCN_AND_SYMBOL(arged_command) {
...
 		const char *cmdname = rz_cmd_parsed_args_cmd (pr_args);
 		eprintf ("Wrong number of arguments passed to `%s`, see its help with `%s?`\n", cmdname, cmdname);
+		RzCmdDesc *cd = rz_cmd_get_desc (state->core->rcmd, cmdname);
+		if (cd) {
+			char *cmdname_help = rz_str_newf ("%s?", cmdname);
+			if (!cmdname_help) {
+				goto err;
+			}
+			RzCmdParsedArgs *help_pra = rz_cmd_parsed_args_newcmd (cmdname_help);
+			char *help_msg = rz_cmd_get_help (state->core->rcmd, help_pra, true);
```
wargio, line 4648: `help_msg` is not checked of its value. so if something goes wrong, it will print `(null)`

wargio (changes requested):

librz/core/cmd.c
wargio: is `help_pra` dynamically allocated? if yes then it needs to be freed, otherwise probably it should be a `const` in there

ret2libc: fixed

Maijin (approved):

PR #348, opened 2021-01-12
Fixed AVR analysis plugin and double free in ds_update_ref_lines

wargio:
**DO NOT SQUASH**

This aims to fix the AVR analysis plugin issues (the `WARNING: Analysis plugin avr did not return disassembly`) and some small issues i have found while fixing it, unrelated to AVR.


XVilka (approved):
Certainly an improvement. Good job! :+1: 
Tested and it works fine for the ROP search.

ret2libc (commented):
I don't know this architecture, sorry. :(

librz/asm/arch/avr/avr_disasm.h
```diff
@@ -91,8 +65,20 @@ struct _disassembledInstruction {
...
+	uint32_t longAddress;
+	/* A copy of the AVR long instruction, we need to keep this so we know information about the
+	 * instruction (mnemonic, operands) after we've read the next 16-bits from the program file. */
+	disassembledInstruction longInstruction;
+};
+typedef struct _avrDisassembleContext avrDisassembleContext;
+
 /* Disassembles an assembled instruction, including its operands. */
-static int disassembleInstruction(disassembledInstruction *dInstruction, const assembledInstruction aInstruction);
+int disassembleInstruction(avrDisassembleContext *context, disassembledInstruction *dInstruction, const assembledInstruction aInstruction);
```
thestr4ng3r, line 81:
```suggestion
int disassembleInstruction(avrDisassembleContext *context, disassembledInstruction *dInstruction, const assembledInstruction aInstruction);
```
Better call it `avr_disassemble_instruction` or something like that.

What is the relationship between `librz/asm/arch/avr/avr_disasm.h`, `librz/asm/arch/avr/disasm.h` and `librz/asm/arch/avr/avr_instructionset.c`? Wouldn't it maybe make more sense to put them all in a single header and prefix everything in there with avr, including the structs and making them follow regular rizin naming conventions? So maybe not prefix with `rz_` but at least get rid of lowerCamelCase stuff.

librz/asm/arch/avr/disasm.h
```diff
@@ -0,0 +1,6 @@
+#ifndef AVR_DISASSEMBLER_H
+#define AVR_DISASSEMBLER_H
+
+int avr_decode (char *out, ut64 addr, cut8 *buf, int len);
```
thestr4ng3r, line 4:
```suggestion
int avr_decode(char *out, ut64 addr, cut8 *buf, int len);
```

thestr4ng3r (commented):
Overall +1 from me, it is already a pure enhancement over the old state, especially considering the `avrDisassembleContext` change.
When finished, please split/squash the changes in sensible commits.

librz/asm/arch/avr/avr_disasm.h
wargio: the namings and files comes from the original ripped code, which i have not touched.
I have added some headers because the code was a big `#include "file.c"`

wargio:
the PR it is finished. i thought there was an issue with AVR and memory mappings, but in reality the issue reside in RzIOMap

wargio:
> I don't know this architecture, sorry. :(

@ret2libc the issue is that ROP code was checking for the `mnemonic` variable being populated in AVR analysis plugin (which was not)

thestr4ng3r (approved):

PR #358, opened 2021-01-14
Move seek history from RzIO to RzCore

ret2libc:

**Detailed description**

- Move RzIO undo implementation to RzCore and provide API to seek and save in one shot.
- Slightly improve undo/redo movements in visual mode (in particular with cursor it could happen that while undoing it moved to the wrong location)
- Use two stacks to implement the history instead of detailed implementation as before. One stack for undos, another for redos. Once an element is undone, it is moved to the redos stack. Once a redo element is redone it is moved back in the undos stack.
- Simplify the printing of the seek history and move all the printing logic in cmd_seek.c

TODOs:
- [x] Limit the size of the undos queue with a variable cfg.seek.histsize
- [x] Add cfg.seek.silent config var in cconfig.c

**Test plan**

Try in visual mode if undo/redo works well. This includes with/without cursor. With `:s ....`. With `g` (in visual mode). With quick jumps (e.g. numbers near jumps/calls).

**Closing issues**

...


wargio:
i love you.

wargio (approved):
Just one comment on the json output: do we really need a variable to know the last location is the current one?

XVilka (commented):
Code looks fine overall. Please fix the broken tests.

ret2libc:
> Just one comment on the json output: do we really need a variable to know the last location is the current one?

It's not always the last one, because if you have done an undo, you will have redos. So the current one may be in the middle of the array. Elements before the current one are "undos", elements after the current one are "redos".

wargio:
> > Just one comment on the json output: do we really need a variable to know the last location is the current one?
> 
> It's not always the last one, because if you have done an undo, you will have redos. So the current one may be in the middle of the array. Elements before the current one are "undos", elements after the current one are "redos".

makes sense. thanks.

ret2libc:
> > > Just one comment on the json output: do we really need a variable to know the last location is the current one?
> > 
> > 
> > It's not always the last one, because if you have done an undo, you will have redos. So the current one may be in the middle of the array. Elements before the current one are "undos", elements after the current one are "redos".
> 
> makes sense. thanks.

It is explained at https://github.com/rizinorg/rizin/pull/358/files#diff-8c2d566471f828f6863a5539d3bc2b695a1c76157f39f5a4b2050d6cc0569b6eR352 btw.

librz/include/rz_vector.h
```diff
@@ -140,7 +140,7 @@ RZ_API void *rz_vector_shrink(RzVector *vec);
 
 #define rz_vector_foreach_prev(vec, it) \
 	if (!rz_vector_empty (vec)) \
-		for (it = (void *)((char *)(vec)->a + (((vec)->len - 1)* (vec)->elem_size)); (char *)it != (char *)(vec)->a; it = (void *)((char *)it - (vec)->elem_size))
+		for (it = (void *)((char *)(vec)->a + (((vec)->len - 1)* (vec)->elem_size)); (char *)it != (char *)(vec)->a - (vec)->elem_size; it = (void *)((char *)it - (vec)->elem_size))
```
ret2libc, line 143: @thestr4ng3r FYI

test/db/cmd/undoredo
```diff
@@ -217,16 +235,16 @@ s+
 sj
 EOF
 EXPECT=<<EOF
-[{"offset":0,"current":true}]
-[{"offset":0},{"offset":10,"current":true}]
-[{"offset":0},{"offset":10,"current":true},{"offset":20}]
-[{"offset":0,"current":true},{"offset":10},{"offset":20}]
+[{"offset":0,"cursor":0,"current":true}]
```
ret2libc, line 238: I can remove the `current` field when it is not true... However I preferred to always include it, so users can know what fields to expect. WDYT?

wargio: no, keep it.

ret2libc:
I think this is ready for last reviews.

librz/include/rz_vector.h
thestr4ng3r: Can you add a dedicated test for this in `test/unit/test_vector.c`?

ret2libc: Done! ;)

thestr4ng3r (approved):

PR #360, opened 2021-01-14
Autodetect RISC-V architecture when using GDB Remote protocol

XVilka:
DO NOT SQUASH

**Detailed description**


shlr/gdb/src/libgdbr.c
```diff
@@ -48,27 +48,27 @@ int gdbr_init(libgdbr_t *g, bool is_server) {
...
 
 	const char *regprofile = gdbr_get_reg_profile (arch, bits);
 	if (!regprofile) {
 		eprintf ("cannot find gdb reg_profile\n");
-		return -1;
+		return false;
 	}
 	if (!gdbr_set_reg_profile (g, regprofile)) {
-		return -1;
+		return false;
```
wargio, line 66: Please add `free (regprofile);` before this return (memleak).

shlr/gdb/src/libgdbr.c
```diff
@@ -48,27 +48,27 @@ int gdbr_init(libgdbr_t *g, bool is_server) {
...
 	if (!g) {
-		return -1;
+		return false;
 	}
 	if (g->target.valid && g->registers) {
-		return 0;
+		return true;
 	}
 
 	const char *regprofile = gdbr_get_reg_profile (arch, bits);
```
wargio: remove `const` since the returned pointer is dynamically allocated by strdup

wargio (changes requested):

shlr/gdb/src/libgdbr.c
```diff
@@ -48,27 +48,27 @@ int gdbr_init(libgdbr_t *g, bool is_server) {
...
 	}
 	g->target.arch = arch;
 	g->target.bits = bits;
 	g->target.valid = true;
 
-	return 0;
+	return true;
 }
 
 const char *gdbr_get_reg_profile(int arch, int bits) {
```
wargio: also here, remove `const`. this returns a a string from strdup

wargio (changes requested):

shlr/gdb/src/gdbclient/xml.c
```diff
@@ -554,6 +554,9 @@ static int _resolve_arch(libgdbr_t *g, char *xml_data) {
 		if (strstr (xml_data, "com.apple.debugserver.arm64")) {
 			g->target.arch = RZ_SYS_ARCH_ARM;
 			g->target.bits = 64;
+		} else if (strstr (xml_data, "org.gnu.gdb.riscv")) {
+			g->target.arch = RZ_SYS_ARCH_RISCV;
+			g->target.bits = 64;
```
yossizap, line 559: It seems that gdb doesn't provide any information in the headers to differentiate 32bit and 64bit riscv, both have `org.gnu.gdb.riscv` and `<architecture>riscv</architecture>`. We should determine the `g->target.bits` based on the bitsize of "pc".

64:
https://github.com/bminor/binutils-gdb/blob/1fa1262d5a865a0f52c381f6465a3965e95c7904/gdb/testsuite/gdb.arch/riscv-tdesc-loading-02.xml
32:
https://github.com/bminor/binutils-gdb/blob/1fa1262d5a865a0f52c381f6465a3965e95c7904/gdb/testsuite/gdb.arch/riscv-tdesc-loading-04.xml

shlr/gdb/include/reg/riscv.h
```diff
@@ -0,0 +1,53 @@
...
+	"=A3	a3\n"
+	"=A4	a4\n"
+	"=A5	a5\n"
+	"=A6	a6\n"
+	"=A7	a7\n"
+	"=R0	a0\n"
+	"=R1	a1\n"
+	"=SN	a7\n" // ABI: syscall numer
+	"gpr	zero	.32	0	0\n" // seems to be allocated in the arena too
+	// RV64I regs (ABI names)
```
yossizap: This file should be renamed riscv64.h, adding riscv32.h is up to you

shlr/gdb/include/reg/riscv.h
```diff
@@ -0,0 +1,53 @@
...
+	"=A2	a2\n"
+	"=A3	a3\n"
+	"=A4	a4\n"
+	"=A5	a5\n"
+	"=A6	a6\n"
+	"=A7	a7\n"
+	"=R0	a0\n"
+	"=R1	a1\n"
+	"=SN	a7\n" // ABI: syscall numer
+	"gpr	zero	.32	0	0\n" // seems to be allocated in the arena too
```
yossizap: Should be .64 according to `Unprivileged Spec v. 20191213` 5.1 
> (RV64I widens the integer registers and supported user address space to 64 bits (XLEN=64 in Figure 2.1).

XLEN is the same for all registers including x0/zero in that figure. Also in gdb:
https://github.com/bminor/binutils-gdb/blob/1fa1262d5a865a0f52c381f6465a3965e95c7904/gdb/testsuite/gdb.arch/riscv-tdesc-loading-02.xml#L6

yossizap (changes requested):
I want to test this with qemu before merging

XVilka:
Failing test is 
```
[XX] db/tools/rz_asm rz-asm -k

RZ_NOPLUGINS=1 rizin -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -Qc '!rz-asm -k darwin 0x90' -

-- stdout

@@ -0,0 +1 @@

+0000ffff
```
It looks completely wrong (the test itself), probably should be rewritten or dropped.

wargio (approved):

shlr/gdb/src/gdbclient/xml.c
wargio: Probably is better to fix this with another PR

wargio:
Mark it as broken

librz/asm/arch/arm/armass64.c
```diff
@@ -1285,222 +1285,118 @@ static bool handlePAC(ut32 *op, const char *str) {
...
+		free (ops.mnemonic);
 		return false;
 	}
 	/* TODO: write tests for this and move out the regsize logic into the mov */
 	if (!strncmp (str, "mov", 3)) {
 		*op = mov (&ops);
-		return *op != -1;
-	}
-	if (!strncmp (str, "cb", 2)) {
+	} else if (!strncmp (str, "cb", 2)) {
```
ret2libc, line 1294: all these strncmp should be replaced with `rz_str_startswith` to be "safer", meaning I don't have to check that the number really corresponds to the length of the string :)

yossizap (approved):
LGTM, tested with riscv64 QEMU. We'll need to put more love into this in the future.

librz/asm/arch/arm/armass64.c
XVilka: Yes,  but it turned out that compiler doesn't optimize strlen for the `const char *` strings:
- Clang 11: https://godbolt.org/z/rnYE38
- GCC 10.2: https://godbolt.org/z/4TGK6n

XVilka:
Imported a couple more fixes - PJ refactoring, memleaks, and marked `rz-asm -k` test as broken (I have no idea what it is supposed to test at all...).

librz/asm/arch/arm/armass64.c
ret2libc: That's probably because our implementation of rz_str_startswith should go in the rz_str.h file and should be static inline. That way the compiler can easily inline it and it will be as good as strncmp with a hard-coded value.

ret2libc: Anyway, a change for another day :)

ret2libc (approved):

PR #361, opened 2021-01-14
AVR: Fixed profile, (null) instruction and analysis

wargio:

**Detailed description**

- Fixed missing SN alias in AVR profile (`Warning: No SN reg alias for current architecture.`).
- Fixed "(null)" instruction
- Fixed analysis when unknown type is found.


wargio:
The failed test behavior is random. before it didn't print anything. then i got a different nomenclature.

XVilka:
> The failed test behavior is random. before it didn't print anything. then i got a different nomenclature.

Seems that test happens only with Clang - it is probably it initializes some variable with zeroes by default, while GCC - doesn't

XVilka (approved):

librz/asm/arch/avr/disasm.c
```diff
@@ -38,11 +44,13 @@ int avr_decode (char *out, ut64 addr, cut8 *buf, int len) {
...
-	if (printDisassembledInstruction (&context, out, dins, opt) < 0) {
+		opsize = 4;
+	} else if (printDisassembledInstruction (&context, out, dins, opt) < 0) {
 		strcpy (out, "invalid");
 		return -1;
 	}
-	return 2;
+	if (out[0] == '.' || !out[0]) {
+		strcpy (out, "invalid");
+	}
```
kazarmy, line 54: Why return `opsize` and not `-1`?

kazarmy, line 54: Why return `opsize` and not `-1`?

kazarmy: Sorry dup.

wargio: because the size can be 2 or 4 and if i return -1 it will skip 2 bytes instead of 4

thestr4ng3r (approved):

librz/asm/arch/avr/disasm.c
kazarmy: The "return `opsize`" is inconsistent with the rest of the function, but `avr_decode` is used in a few places and I think the code works ... wish it could be less confusing.

kazarmy (approved):

PR #379, opened 2021-01-17
Clear syscall and sysregs sdbs if their files are not found

kazarmy:

**Detailed description**

This pr clears the syscall and sysregs sdbs if their files are not found, to prevent usage of sdbs that are inappropriate for the os/arch/bits/cpu.

**Test plan**

All builds should be green, including the Travis builds.

**Closing issues**

...


ITAYC0HEN (approved):

XVilka (approved):

XVilka:
Ideally, it should be something like:
```c
static bool openDatabase(Sdb **db, const char *name) {
	char *file = rz_str_newf ( RZ_JOIN_3_PATHS ("%s", RZ_SDB, "%s.sdb"),
		rz_sys_prefix (NULL), name);
	if (rz_file_exists (file)) {
		if (db && *db) {
			sdb_reset (*db);
			sdb_open (*db, file);
		} else {
			*db = sdb_new (0, file, 0);
		}
	} else {
             free (file);
             return false;
       }
       free (file);
       return true;
```
And checking the result in the place where it was called, imho.

test/db/analysis/avr
```diff
@@ -673,36 +672,36 @@ EXPECT=<<EOF
...
-0x000001d0 1 syscall.read.2
-0x00000260 1 syscall.read.3
+0x000001d0 1 syscall.0.2
+0x00000260 1 syscall.0.3
 0x0000029e 164 fcn.0000029e
-0x000002ae 1 syscall.read.5
-0x000002f2 1 syscall.inotify_rm_watch.1
-0x00000318 1 syscall.inotify_rm_watch.0
-0x00000330 1 syscall.inotify_rm_watch
+0x000002ae 1 syscall.0.5
```
ret2libc, line 678: I'm not sure I understand why these need to be changed. Can you explain?

XVilka: There are no such syscalls on the AVR architecture.

wargio (approved):

test/db/analysis/avr
ret2libc: There are, I guess, if you have specified os=linux as it was before.
Was this test just completely wrong?

Also, are there syscalls anyway in those places?

wargio: yes, the test was wrong.

kazarmy:
> Ideally, it should be something like:

the suggestion lgtm

Issue #350, opened 2021-01-13
test/unit/test_dwarf and test/unit/test_dwarf_info fail on Alpine Linux static build


XVilka:
### Work environment

| Questions                                            | Answers
|------------------------------------------------------|--------------------
| OS/arch/bits (mandatory)                             | Alpine Linux (Edge) x86_64
| File format of the file you reverse (mandatory)      | DWARF
| Architecture/bits of the file (mandatory)            | x86/64
| rizin -v full output, **not truncated** (mandatory)  | https://github.com/rizinorg/rizin/commit/4805f845d3a72d058b79048e809a594f4d3634f1

### Expected behavior

All tests pass

### Actual behavior

Segfaults on the `test_big_endian_dwarf2` from `test/unit/test_dwarf` and `test_dwarf2_big_endian` from `test/unit/test_dwarf_info`.

```
74/74 vector               OK             0.00s

The output from the failed tests:

29/74 dwarf                FAIL           0.47s (killed by signal 11 SIGSEGV)

--- command ---
10:24:02 /__w/rizin/rizin/rizin/build/test/unit/test_dwarf
--- stdout ---
test_dwarf_cpp_empty_line_info OK
-------

30/74 dwarf_info           FAIL           0.37s (killed by signal 11 SIGSEGV)

--- command ---
10:24:02 /__w/rizin/rizin/rizin/build/test/unit/test_dwarf_info
--- stdout ---
test_dwarf3_c OK
-------


Summary of Failures:

29/74 dwarf                FAIL           0.47s (killed by signal 11 SIGSEGV)
30/74 dwarf_info           FAIL           0.37s (killed by signal 11 SIGSEGV)

Ok:                 72  
Expected Fail:      0   
Fail:               2   
Unexpected Pass:    0   
Skipped:            0   
Timeout:            0   

Full log written to /__w/rizin/rizin/rizin/build/meson-logs/testlog.txt
ninja: job failed: /usr/bin/meson test --no-rebuild --print-errorlogs
ninja: subcommand failed
```

### Steps to reproduce the behavior
```yaml
  build-static:
    name: Build static
    # FIXME: ignore error until all tests properly pass
    continue-on-error: true
    if: contains(github.head_ref, 'dist') || contains(github.head_ref, 'static') || contains(github.ref, 'release-') || github.event_name == 'schedule'
    runs-on: ubuntu-20.04
    container: alpine:edge
    steps:
    - name: Install dependencies
[...]

PR #386, opened 2021-01-18
Release preparations

XVilka:

**Detailed description**

Checking if all release machinery is ready.

**Test plan**

All should be green.


XVilka:
So only missing things are:
- [x] 2 failing tests for Alpine static build: https://github.com/rizinorg/rizin/issues/350
- [ ] Unpublished [rz-pipe.py](https://github.com/rizinorg/rz-pipe) on [PyPI](https://pypi.org/).

ret2libc:
The pipe2 error is because that still uses the acr/make build system. But honestly... I would not release a DEB that way...  @thestr4ng3r  ?

AFAIK there is also https://man7.org/linux/man-pages/man7/debhelper.7.html that has integration with meson/ninja so we should be able to do something with that, but I haven't looked at it yet. I suggest we drop  the .deb file for now. In any case, distros hopefully will package that for us at some point :)

XVilka:
@ret2libc there is **zero** documentation on how to do this, so the old way could work for now. It's not an error, just a warning.

thestr4ng3r:
Imo we should not release debs like this at all.

XVilka:
I remember some people use `*.deb` files. Ideally, we should build those binaries statically, with Alpine and musl, and pack then into the deb, so they will work almost anywhere without the 3rd party dependencies. But that's a task for the 0.2.0 release instead.

XVilka:
Regarding porting building the Debian package we can see https://github.com/linuxmint/xed/tree/master/debian as an example.

ret2libc:
I agree with @thestr4ng3r on not releasing .deb files this way. I'm honestly not sure they will work on various debian/ubuntu versions :(

XVilka:
Removed `*.deb` packages building and uploading then.

ret2libc:
I think the build dir used to run `meson dist` should be done with `enable_tests=false`. We probably do not want to ship rz-test as part of the release. Also, We should probably find a way to remove from the release tarball all the test/db stuff?

librz/util/sys.c
```diff
@@ -1378,6 +1378,9 @@ RZ_API void rz_sys_info_free(RSysInfo *si) {
 
 #if __UNIX__ && HAVE_PIPE2
 #include <fcntl.h>
+#ifndef _GNU_SOURCE
```
ret2libc: this is not necessary. It is already defined in meson.build

XVilka: It was necessary for the ACR-based `sys/debian.sh`. But because we removed the use of it, I will remove those lines too.

thestr4ng3r:
> Also, We should probably find a way to remove from the release tarball all the test/db stuff?

I think you should be able to run all tests from a release tarball too. The main question there would be about the testbins, but the makefile (acr/buildsystem-independent) in there can fetch it.

XVilka:
Interesting bug on Alpine that isn't related to the debugging:
```
[XX] db/tools/rz_ax rz-ax -t "1234567890 GMT-1"
RZ_NOPLUGINS=1 rizin -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -Qc '!rz-ax -t "1234567890 GMT-1" | cut -d " " -f 1,2' -
-- stdout
@@ -1 +1 @@
-Fri Feb
+Fri May
```

XVilka:
@ret2libc @thestr4ng3r it's ready now, please check & test the artifacts/etc

.github/workflows/ci.yml
```diff
@@ -375,27 +375,34 @@ jobs:
       working-directory: rizin
     - name: Compile with meson
       run: |
-        CFLAGS="-static" LDFLAGS="-static" meson --prefix=${HOME}/.local --buildtype release --default-library static build
+        mkdir -p ${HOME}/rizin-static
+        CFLAGS="-static" LDFLAGS="-static" meson --prefix=${HOME}/rizin-static --buildtype release --default-library static build
         ninja -C build && ninja -C build install
+        tar -C ${HOME}/rizin-static --strip-components=3 --xz -cf rizin-static.tar.xz ${HOME}/rizin-static
```
ret2libc: why this? isn't the created tarball already in good shape?

XVilka: It isn't

XVilka: Doesn't help anyway

.github/workflows/ci.yml
```diff
@@ -793,41 +720,14 @@ jobs:
...
+    - name: Upload rizin static x86_64 tar.gz
       uses: actions/upload-release-asset@v1
       env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       with:
           upload_url: ${{ steps.create_release.outputs.upload_url }}
-          asset_path: ./rizin-dev-ubuntu-1804-deb/rizin-dev_amd64.deb
-          asset_name: rizin-dev-ubuntu-1804_${{ steps.extract_version.outputs.branch }}_amd64.deb
+          asset_path: ./rizin-static/rizin-static.tar.xz
+          asset_name: rizin-${{ steps.extract_version.outputs.branch }}-static-x86_64.tar.gz
```
ret2libc: why the conversion from tar.xz to tar.gz?

.github/workflows/ci.yml
```diff
@@ -375,27 +375,34 @@ jobs:
...
       uses: actions/upload-artifact@v2
       with:
         name: test-results-static
-        path: test/results.json
+        path: rizin/test/results.json
+    - name: Upload static build
+      uses: actions/upload-artifact@v2
+      with:
+        name: rizin-static.tar.xz
+        path: rizin/rizin-static.tar.xz
```
ret2libc, line 403: if i try to download the file, I find another archive inside this one. I think there's something wrong.

XVilka: It's GitHub failure - it always creates ZIP files.

.github/workflows/ci.yml
XVilka: No conversion, forgot to change.

.github/workflows/ci.yml
XVilka: See https://github.com/actions/upload-artifact/issues/39

ret2libc (approved):

PR #375, opened 2021-01-16
Convert `s` commands to newshell

ret2libc:
**Your checklist for this pull request**
- [x] I've read the [guidelines for contributing](https://github.com/rizinorg/rizin/blob/master/DEVELOPERS.md) to this repository
- [x] I made sure to follow the project's [coding style](https://github.com/rizinorg/rizin/blob/master/DEVELOPERS.md#code-style)
- [x] I've added tests that prove my fix is effective or that my feature works (if possible)
- [x] I've updated the documentation and the [rizin book](https://github.com/rizinorg/book) with the relevant information (if needed). See https://github.com/rizinorg/book/pull/23
- [ ] Implement autocompletion

**Detailed description**

* Rename `sj`/`s*` and move them under `sH`. Remove `s=`
* Use `sd` for "seek delta", which moves relative to the current offset
* Enforce <cmd> <arg> syntax. So no more `s+16` but `s +16`. Command
  `sHr` is for redo history, `sHu` for undo history. If one wants to move
  relative to the current address, he needs to do `sd +10`/`sd -10`.
* Remove `ss` commands in favour of cmd.seek.silent. This was anyway
  used mainly for scripts, so no need to replicate all `s` commands
  under `ss`.

**Test plan**

Try all various s commands, visual mode undo/redo.

**Closing issues**

PR #401, opened 2021-01-20
Use space between 'so' and argument

ret2libc:
See https://github.com/rizinorg/rizin/pull/375#pullrequestreview-571700406


XVilka (approved):

XVilka:
While you are on it, what about just changing to the API?

wargio (approved):

kazarmy (approved):

PR #418, opened 2021-01-21
Fixes for jump table ##analysis

### LINKED ISSUE #304: Switch statement edges duplicated and attached in wrong place
### Work environment

| OS/arch/bits (mandatory)                             | macOS/x86_64 on arm64 (with rosetta)
| File format of the file you reverse (mandatory)      | machO
| Architecture/bits of the file (mandatory)            | x86 64
| rizin -v full output, **not truncated** (mandatory)  | rizin 0.1.0 0 @ darwin-x86-64 git.0.1.0 commit: unknown build: 2021-01-04__22:56:50

Not sure why commit hash wasn't properly included in `rizin -v`. Should be 4ac005d4894ac5f13b007784aad4d6147ebf1296

### Expected behavior

Switch statement edges are attached to the basic block which ends with computed jump

### Actual behavior

Switch statement edges are added twice and to two basic blocks which precede the block with indirect jump.

* Open /bin/ls on macOS
* aaa
* view CFG for main function.

### Additional Logs, screenshots, source-code,  configuration dump, ...

```asm
0x100004153      mov     byte [0x100008504], 1 ; switch statement edges attached here 
0x10000415a      mov     r12d, 0x10 ; and here
0x100004160      lea     r13, str.1_ABCFGHLOPRSTUWabcdefghiklmnopqrstuvwx ; 0x100007a9c
0x100004167      lea     rbx, [0x1000048e8]
0x10000416e      mov     edi, r15d ; int argc
0x100004171      mov     rsi, r14  ; const char * *argv
0x100004174      mov     rdx, r13  ; const char *optstring
0x100004177      call    getopt    ; sym.imp.getopt ; int getopt(int argc, const char * *argv, const char *optstring)
0x10000417c      lea     ecx, [rax - 0x25]
0x10000417f      cmp     ecx, 0x53
0x100004182      ja      case.default.0x10000418b
0x100004184      movsxd  rax, dword [rbx + rcx*4]
0x100004188      add     rax, rbx
;-- switch
0x10000418b      jmp     rax       ; switch table (84 cases) at 0x1000048e8 ; the edges should be around here
```

Screenshot made using Cutter because it's easier to fit more info. But it was also confirmed using rizin from command line.

The part where switch edges get attached
Screenshot 2021-01-04 at 23 18 24

Thee part where edges should be attached
Screenshot 2021-01-04 at 23 22 51

This is just a guess, but maybe the basic block gets split after the switch edges are already created due to code jumping in the middle of it but the splitting logic leaves the edges attached to first part instead of second.

I will later try to make a test executable for this.

GustavoLCR:

**Detailed description**

* Try to detect shift of first case value
* Detect when block where switch table is attached is split by analysis
* Fix attaching switch table to wrong block
* Add/fix tests

**Test plan**

Tests pass

**Closing issues**

Fixes #304


wargio (approved):

wargio:
you need to rebase

karliss:
Tested using the executable in #304 it now looks correct.

librz/analysis/jmptbl.c
```diff
@@ -44,10 +44,26 @@ static void apply_switch(RzAnalysis *analysis, ut64 switch_addr, ut64 jmptbl_add
 // analyze a jmptablle inside a function // maybe rename to rz_analysis_fcn_jmptbl() ?
 RZ_API bool rz_analysis_jmptbl(RzAnalysis *analysis, RzAnalysisFunction *fcn, RzAnalysisBlock *block, ut64 jmpaddr, ut64 table, ut64 tablesize, ut64 default_addr) {
 	const int depth = 50;
-	return try_walkthrough_jmptbl(analysis, fcn, block, depth, jmpaddr, table, table, tablesize, tablesize, default_addr, false);
```
XVilka, line 47: We need to simplify these functions in the future...

XVilka (approved):

test/db/analysis/x86_32
```diff
@@ -629,7 +629,7 @@ aaa
 f~?case
 EOF
 EXPECT=<<EOF
-90
+91
```
ret2libc, line 632: I think this PR is changing this test in a more important way.
Old behaviour:
Screenshot 2021-01-23 at 16 06 05
New behaviour:
Screenshot 2021-01-23 at 16 06 16
Jump table:
Screenshot 2021-01-23 at 16 06 42

The basic block shown in the first two images is 0x10000192c + 4294965314 (modulo 32bit), which as you can see is the selected case of 1...14, etc., as shown in the old behaviour, not the new one.

ret2libc (changes requested):

test/db/analysis/x86_32
karliss: I am guessing that it's the result if implementing "shift of first case value" support. The question if in this case it's false posivite for shift logic or are the new results  more correct.

karliss: image

The code subtracts 49 from the index before doing lookup in jump table. Seems likely that compiler removed first 49 unused entrie to reduce table size.

test/db/analysis/x86_64
```diff
@@ -3391,38 +3391,109 @@ f~case.
 EOF
 EXPECT=<<EOF
 0x140001023 1 switch.0x140001023
+0x140001f5b 1 switch.0x140001f5b
```
ret2libc, line 3394: Maybe not for this PR, but this is really weird... There is a switch table here but there is no function defined. However at 0x140001e38 you can see `jbe   0x140001f40`, which goes just there. It's strange no basicblock was created there.

ret2libc (approved):

Issue #42, opened 2020-10-20
Tons of flags are shown at offset 0x00


ITAYC0HEN:
```
$ rizin /bin/ls
[0x00000000]> s 0
[0x00000000]> pd 1
            ;-- section.:
            ;-- section..comment:
            ;-- section..shstrtab:
            ;-- segment.LOAD0:
            ;-- segment.GNU_STACK:
            ;-- segment.ehdr:
            ;-- __ctype_toupper_loc:
            ;-- getenv:
            ;-- cap_to_text:
            ;-- sigprocmask:
            ;-- __snprintf_chk:
            ;-- raise:
            ;-- free:
            ;-- abort:
            ;-- __errno_location:
            ;-- strncmp:
            ;-- _ITM_deregisterTMCloneTable:
            ;-- localtime_r:
            ;-- _exit:
            ;-- strcpy:
            ;-- __fpending:
            ;-- isatty:
            ;-- sigaction:
            ;-- iswcntrl:
            ;-- wcswidth:
            ;-- localeconv:
            ;-- mbstowcs:
            ;-- readlink:
            ;-- clock_gettime:
            ;-- setenv:
            ;-- textdomain:
            ;-- fclose:
            ;-- opendir:
            ;-- getpwuid:
            ;-- bindtextdomain:
            ;-- dcgettext:
            ;-- __ctype_get_mb_cur_max:
            ;-- strlen:
            ;-- __lxstat:
            ;-- __stack_chk_fail:
            ;-- getopt_long:
            ;-- mbrtowc:
            ;-- strchr:
            ;-- getgrgid:
            ;-- __fxstatat:
            ;-- snprintf:
            ;-- __overflow:
            ;-- strrchr:
            ;-- gmtime_r:
            ;-- lseek:
            ;-- __assert_fail:
            ;-- __strtoul_internal:
            ;-- fnmatch:
            ;-- memset:
            ;-- ioctl:
            ;-- getcwd:
            ;-- strspn:
            ;-- closedir:
            ;-- __libc_start_main:
            ;-- memcmp:
            ;-- _setjmp:
            ;-- fputs_unlocked:
            ;-- calloc:
            ;-- strcmp:
            ;-- signal:
            ;-- dirfd:
            ;-- fputc_unlocked:
            ;-- getpwnam:
            ;-- __memcpy_chk:
            ;-- syscall:
            ;-- sigemptyset:
            ;-- __gmon_start__:
            ;-- memcpy:
[...]

PR #426, opened 2021-01-21
Don't create flags at UT64_MAX

p371jah:

**Detailed description**

...

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/librz/core/cbin.c
+++ b/librz/core/cbin.c
@@ -2350,7 +2350,7 @@ static int bin_symbols(RzCore *r, PJ *pj, int mode, ut64 laddr, int va, ut64 at,
 					rz_flag_item_set_realname(fi, n);
 					fi->demangled = (bool)(size_t)sn.demname;
 				} else {
-					if (fn) {
+					if (fn && addr != UT64_MAX) {
 						eprintf("[Warning] Can't find flag (%s)\n", fn);
 					}
 				}
--- a/librz/flag/flag.c
+++ b/librz/flag/flag.c
@@ -711,7 +711,7 @@ RZ_API RzFlagItem *rz_flag_set_next(RzFlag *f, const char *name, ut64 off, ut32 
  * The realname of the item will be the same as the name.
  * NULL is returned in case of any errors during the process. */
 RZ_API RzFlagItem *rz_flag_set(RzFlag *f, const char *name, ut64 off, ut32 size) {
-	rz_return_val_if_fail(f && name && *name, NULL);
+	rz_return_val_if_fail(f && name && *name && off != UT64_MAX, NULL);
 
 	bool is_new = false;
 	char *itemname = filter_item_name(name);
--- a/librz/flag/flag.c
+++ b/librz/flag/flag.c
@@ -711,7 +711,10 @@ RZ_API RzFlagItem *rz_flag_set_next(RzFlag *f, const char *name, ut64 off, ut32 
  * The realname of the item will be the same as the name.
  * NULL is returned in case of any errors during the process. */
 RZ_API RzFlagItem *rz_flag_set(RzFlag *f, const char *name, ut64 off, ut32 size) {
-	rz_return_val_if_fail(f && name && *name && off != UT64_MAX, NULL);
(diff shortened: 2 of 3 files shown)


XVilka:
It broke 3 tests:
```
[XX] db/formats/mach0/leb128 macho trie malformed uleb128
RZ_NOPLUGINS=1 rizin -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -Qc  bins/mach0/macho-trie-bad-export-info-malformed-uleb128
-- stderr
@@ -1,2 +1,3 @@
 malformed uleb128
 malformed uleb128
+WARNING: rz_flag_set: assertion 'f && name && *name && off != UT64_MAX' failed (line 714)

[XX] db/esil/x86_64 mem read
RZ_NOPLUGINS=1 rizin -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -Qc 'e io.va=true
e esil.debug=true
e asm.arch=x86
e asm.bits=64
aei
aeim
aer rsp=0x00108000
.aer*
aer rdi=0xFFFFFFFFFFFFFFFF
ae rdi,rsp,=[8]
.aer*
ae rsp,[8],rax,=
.aer*
?v rax
' -
-- stdout
@@ -1 +1 @@
-0xffffffffffffffff
+0x0

[XX] db/cmd/cmd_i it2
RZ_NOPLUGINS=1 rizin -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -Qc 'mkdir .tmp
cp bins/mach0/fatmach0-3true .tmp/it2-modfile
o .tmp/it2-modfile
oo+
it
wx 0x90
it
o--
rm .tmp/it2-modfile
' --
-- stderr
@@ -1,3 +1,4 @@
+WARNING: rz_flag_set: assertion 'f && name && *name && off != UT64_MAX' failed (line 714)
 File has been modified.
 - md5 84144530144b53a704c84d9c65ab5f92
 + md5 9a99cca7bf3e27ee1c8964141cfad053
```

p371jah:
About `db/esil/x86_64 mem read`: `aer` command execute `.dr*` (equivalent of `.aer*`)

XVilka (approved):

p371jah:
Should `offset == UT64_MAX` be treated as an error? Not sure that skipping it silently is ok. But numerous warning messages are not ok too.

ret2libc (commented):
Why do we need to consider UT64_MAX not valid? I guess this assumption is unfortunately done in many other places, right? :(

p371jah:
Between `0` and `UT64_MAX` I would prefer the second one.

p371jah:
But somewhere in rizin code `0` is used for invalid addresses, and as a result you get this https://github.com/rizinorg/rizin/issues/42

ret2libc:
> Between `0` and `UT64_MAX` I would prefer the second one.

Oh well, yeah :D though I'd say there should be something else to indicate an invalid address, because the entire ut64 range is valid. But yeah, between the two I prefer UT64_MAX, but then you have the same problem only at a different (less common) address.

p371jah:
Well, since the issue is relates mostly to symbols it would be better to use `undefined address` term.

ret2libc:
I am not sure about this... 
`f world @ 0xffffffffffffffff` will not work anymore with this PR.

wargio:
yes but `0xffffffffffffffff` is an impossible address

ITAYC0HEN:
> yes but `0xffffffffffffffff` is an impossible address

but it might be a possible value (crypto, etc...)

wargio:
Ok, so if we want to consider valid any address space, then we need to refactor all the methods to be like: `RZ_API bool my_function(...., ut64 *returned_address);` so we can check if the function has failed and allow UT64_MAX.

ret2libc:
> Ok, so if we want to consider valid any address space, then we need to refactor all the methods to be like: `RZ_API bool my_function(...., ut64 *returned_address);` so we can check if the function has failed and allow UT64_MAX.

Yes I actually think this is the only real way. Rizin represents an address space of 64 bit, so you can't use one of those addresses as "invalid". You need to have some other ways to mark the invalid addr.

Issue #383, opened 2021-01-18
Refactor Panels mode (`librz/core/tui/panels.c`) to use API instead of commands


XVilka:
Currently the code of Panels modes heavily relies on running commands instead of just calling the proper API functions.
To reduce the overhead, code cruft, and code breakage in case of the command changes it is important to remove all those command calls and call the proper API directly.
```
$ rg "rz_core_cmd[0f_]*\(" librz/core/tui/panels.c                                                                                                                                                                                 rg "rz_core_cmd[0f_]*\(" librz/core/tui/panels.c
1666:         rz_core_cmd(core, buf, 1);
6937:                 (void)rz_core_cmd0(core, cmd);
6943:                 (void)rz_core_cmd0(core, cmd);
6951:                 (void)rz_core_cmd0(core, cmd);
6957:                 (void)rz_core_cmd0(core, cmd);
6963:                 (void)rz_core_cmd0(core, cmd);
6969:                 (void)rz_core_cmd0(core, cmd);
6975:                 (void)rz_core_cmd0(core, cmd);
6986:                 (void)rz_core_cmd0(core, cmd);
6997:                 (void)rz_core_cmd0(core, cmd);
7008:                 (void)rz_core_cmd0(core, cmd);
7014:                 (void)rz_core_cmd0(core, cmd);
7020:                 (void)rz_core_cmd0(core, cmd);
```

ret2libc:
I think this can probably be moved to a later release. It's just refactoring and we seem to have already a lot of things in our hands.

PR #431, opened 2021-01-22
Refactor panels' Analyze menu to use more API ##refactor

officialcjunior:

**Detailed description**

This refactors panels mode a bit to use more API instead of commands.
This commit makes `cmd_analysis_blocks` and `cmd_analysis_calls`
a part of RZ_API and uses it while different options from the
Analyze menu of the Visual Panel mode is triggered.

Some other APIs are also used here instead of `rz_core_cmd0`

**Test plan**
Some variables' names might need to be tweaked. 
Also needs to write tests, if it is required.

**Closing issues**

Related to #383


XVilka (approved):
Will be simplified even more once `a` commands ported to the newshell. :+1:

officialcjunior:
Looks like I made a mistake while naming it.
It should be `rz_core_cmd_analysis_blocks` and `rz_core_cmd_analysis_calls` isn't it?
Forgot to add the `_core_` part.

librz/core/panels.c
```diff
@@ -4827,7 +4827,7 @@ void __panel_breakpoint(RzCore *core) {
 }
 
 void __panel_continue(RzCore *core) {
-	rz_core_cmd(core, "dc", 0);
+	rz_debug_continue(core->dbg);
```
ret2libc: The command and this do slightly different things here :( I suggest you refactor the `dc` command in `cmd_debug.c` and wrap all the command does now within a `RZ_IPI rz_debug_continue_oldhandler (void *data, const char *input)` (it will be useful for when I/someone will convert existing debug commands to newshell).

librz/core/panels.c
```diff
@@ -3571,7 +3571,7 @@ int __watch_points_cb(void *user) {
 
 int __references_cb(void *user) {
 	RzCore *core = (RzCore *)user;
-	rz_core_cmdf(core, "aar");
+	rz_core_analysis_refs (core, "\0");
```
ret2libc:
```suggestion
	rz_core_analysis_refs (core, "");
```
should be enough

librz/core/panels.c
```diff
@@ -3530,13 +3530,13 @@ int __program_cb(void *user) {
...
 	RzCore *core = (RzCore *)user;
-	rz_core_cmdf(core, "aab");
+	rz_cmd_analysis_blocks (core, "\0");
 	return 0;
 }
 
 int __calls_cb(void *user) {
 	RzCore *core = (RzCore *)user;
-	rz_core_cmdf(core, "aac");
+	rz_cmd_analysis_calls(core, "\0", false, false);
```
ret2libc:
```suggestion
	rz_cmd_analysis_calls(core, "", false, false);
```

librz/core/panels.c
```diff
@@ -3530,13 +3530,13 @@ int __program_cb(void *user) {
 
 int __basic_blocks_cb(void *user) {
 	RzCore *core = (RzCore *)user;
-	rz_core_cmdf(core, "aab");
+	rz_cmd_analysis_blocks (core, "\0");
```
ret2libc:
```suggestion
	rz_cmd_analysis_blocks (core, "");
```

librz/core/panels.c
```diff
@@ -3518,7 +3518,7 @@ int __function_cb(void *user) {
 
 int __symbols_cb(void *user) {
 	RzCore *core = (RzCore *)user;
-	rz_core_cmdf(core, "aa");
+	rz_core_analysis_all (core);
```
ret2libc: this needs to be tested a bit. DId you already tried to see if it works well. I have some doubts just because it seems the `aa` command in `cmd_analysis.c` does a lot more than just this function.

ret2libc (commented):
Looks good overall, thanks a lot!

librz/core/panels.c
officialcjunior: >I have some doubts just because it seems the aa command in 
> cmd_analysis.c does a lot more than just this function.

Yes, you are right about that.
https://github.com/rizinorg/rizin/blob/dev/librz/core/cmd_analysis.c#L9421-L9424

>DId you already tried to see if it works well.

Well, there changes were quite visible on a test binary I had used. So, I just went with it. 
I have not went down in detail.



Maybe, it's better to revert it?

librz/core/panels.c
officialcjunior: Thank you for the reply.
The overall idea is clear. But there are some particulars I'm not sure of.
So, I'll push a commit regarding this and let you request changes there.

> (it will be useful for when I/someone will convert existing debug commands to newshell)

Do tell if there are any others :)

librz/core/cmd_debug.c
```diff
@@ -4659,6 +4651,19 @@ static void consumeBuffer(RzBuffer *buf, const char *cmd, const char *errmsg) {
 	rz_cons_printf("\n");
 }
 
+RZ_IPI void rz_debug_continue_oldhandler (void *data, const char *input) {
```
ret2libc: did you run `git-clang-format`? There should be no space here AFAIK.

librz/core/cmd_descs.h
```diff
@@ -55,6 +55,7 @@ RZ_IPI int rz_cmd_bsize(void *data, const char *input);
 RZ_IPI int rz_cmd_cmp(void *data, const char *input);
 RZ_IPI int rz_cmd_meta(void *data, const char *input);
 RZ_IPI int rz_cmd_debug(void *data, const char *input);
+RZ_IPI void rz_debug_continue_oldhandler(void *data, const char *input);
```
ret2libc: This is an auto-generated file, as the header on top says. You should modify `cmd_descs.yaml`.
In particular, find the section for `d` command and add something like:
```yaml
  subcommands:
    - name: dc
      cname: debug_continue_oldhandler
      summary: Continue execution
      type: RZ_CMD_DESC_TYPE_OLDINPUT
```

Then when you re-run `ninja -C build` it will update both `cmd_descs.h` and `cmd_descs.c`. Make sure to have `PyYAML` python module installed.

officialcjunior: Sorry about that, I'll fix that.

ret2libc: No problem ;)

officialcjunior: How/what should the function return? Since, an `RZ_IPI` returning an `int` is what the `.yaml` generated.

```
		eprintf("Cannot continue, run ood?\n");
		return false;
	}
	rz_debug_continue(core->dbg);
	return true;
}
```
Would that do?

ret2libc: return a number `>= 0` when everything goes fine, `-1` when there are errors.
So it should become:
```
                 eprintf("Cannot continue, run ood?\n");
		return -1;
	}
	rz_debug_continue(core->dbg);
	return 0;
}
```

officialcjunior: Did that. But the CI fails somehow.

I think we have to make this available/included to `panels.c` to fix the build.

And the broken tests, too.

librz/core/panels.c
```diff
@@ -4827,7 +4827,7 @@ void __panel_breakpoint(RzCore *core) {
 }
 
 void __panel_continue(RzCore *core) {
-	rz_core_cmd(core, "dc", 0);
+	rz_debug_continue_oldhandler (core, "");
```
ret2libc:
```
../librz/core/panels.c: In function ‘__panel_continue’:
../librz/core/panels.c:4830:2: warning: implicit declaration of function ‘rz_debug_continue_oldhandler’; did you mean ‘rz_debug_continue_syscalls’? [-Wimplicit-function-declaration]
 4830 |  rz_debug_continue_oldhandler (core, "");
      |  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |  rz_debug_continue_syscalls

```

I think you should `#include "cmd_descs.h"`

librz/core/panels.c
ret2libc: > Maybe, it's better to revert it?

No, it seems alright.

ret2libc (dismissed):

ret2libc (changes requested):
Make sure to add the include to remove the warning.

ret2libc (approved):

librz/core/panels.c
```diff
@@ -1,6 +1,7 @@
 /* Copyright rizin 2014-2020 - Author: pancake, vane11ope */
 
 #include <rz_core.h>
+#include <cmd_descs.h>
```
ret2libc:
```suggestion
#include "cmd_descs.h"
```

ret2libc (approved):
LGTM, however there are tests failing but they are not caused by this. It seems GitHub CI has finally switched to ubuntu-20.04, so we need to update the debug tests because they are very fragile.

ret2libc:
> LGTM, however there are tests failing but they are not caused by this. It seems GitHub CI has finally switched to ubuntu-20.04, so we need to update the debug tests because they are very fragile.

Actually it seems they happen only in this PR, so they might be real problems.

librz/core/cmd_debug.c
```diff
@@ -4659,6 +4651,20 @@ static void consumeBuffer(RzBuffer *buf, const char *cmd, const char *errmsg) {
 	rz_cons_printf("\n");
 }
 
+RZ_IPI int rz_debug_continue_oldhandler(void *data, const char *input) {
+	RzCore *core = (RzCore *)data;
+	rz_reg_arena_swap(core->dbg->reg, true);
+#if __linux__
+		core->dbg->continue_all_threads = true;
```
ret2libc: wrong indentation here.

librz/core/cmd_debug.c
```diff
@@ -4659,6 +4651,20 @@ static void consumeBuffer(RzBuffer *buf, const char *cmd, const char *errmsg) {
...
+#endif
+	if (rz_debug_is_dead(core->dbg)) {
+		eprintf("Cannot continue, run ood?\n");
+		return -1;
```
ret2libc: I suspect this has to do with this. Try to return 0 here. Unfortunately existing commands often just return a "success" return value even when something goes wrong. This is an interesting point to re-evaluate when we convert `d` subcommands to newshell.

ret2libc (changes requested):

librz/core/cmd_debug.c
officialcjunior: And 1 when it succeeds?

ret2libc: No, 0/1 is the same IIRC. As said, this return mess 0 vs negative vs positive vs bool vs whatever would finally be resolved when switching to newshell, where we have well-defined return states https://github.com/rizinorg/rizin/blob/dev/librz/include/rz_cmd.h#L23

ret2libc:
Found the issue, my bad. `dc` is not just the command `dc` but it is actually the name of the whole group `dc` (see `dc?`). So if we want to switch it to newshell, we have to convert the whole tree, not just the specific `dc` command.

Apply this patch with `git am  <patch>` please:

```diff
From d1fe5086b0a4a49fafb046c3ccd9837a05369d78 Mon Sep 17 00:00:00 2001
From: Riccardo Schirone <sirmy15@gmail.com>
Date: Tue, 26 Jan 2021 15:04:25 +0100
Subject: [PATCH] Make the whole 'dc' tree a OLDINPUT command

---
 librz/core/cmd_debug.c | 53 ++++++++++++++++++++----------------------
 1 file changed, 25 insertions(+), 28 deletions(-)

diff --git a/librz/core/cmd_debug.c b/librz/core/cmd_debug.c
index 63effeb964..e76ad82578 100644
--- a/librz/core/cmd_debug.c
+++ b/librz/core/cmd_debug.c
@@ -4311,13 +4311,22 @@ static bool cmd_dcu(RzCore *core, const char *input) {
 	return true;
 }
 
-static int cmd_debug_continue(RzCore *core, const char *input) {
+RZ_IPI int rz_debug_continue_oldhandler(void *data, const char *input) {
+	RzCore *core = (RzCore *)data;
 	int pid, old_pid, signum;
 	char *ptr;
 	// TODO: we must use this for step 'ds' too maybe...
-	switch (input[1]) {
+	switch (input[0]) {
 	case 0: // "dc"
-		rz_debug_continue_oldhandler(core, input);
+		rz_reg_arena_swap(core->dbg->reg, true);
+#if __linux__
+		core->dbg->continue_all_threads = true;
+#endif
+		if (rz_debug_is_dead(core->dbg)) {
+			eprintf("Cannot continue, run ood?\n");
+			break;
+		}
+		rz_debug_continue(core->dbg);
 		break;
 	case 'a': // "dca"
 		eprintf("TODO: dca\n");
@@ -4348,7 +4357,7 @@ static int cmd_debug_continue(RzCore *core, const char *input) {
 		break;
 	case 'c': // "dcc"
 		rz_reg_arena_swap(core->dbg->reg, true);
-		if (input[2] == 'u') {
+		if (input[1] == 'u') {
 			rz_debug_continue_until_optype(core->dbg, RZ_ANALYSIS_OP_TYPE_UCALL, 0);
 		} else {
 			rz_debug_continue_until_optype(core->dbg, RZ_ANALYSIS_OP_TYPE_CALL, 0);
@@ -4361,8 +4370,8 @@ static int cmd_debug_continue(RzCore *core, const char *input) {
 	case 'k':
 		// select pid and rz_debug_continue_kill (core->dbg,
 		rz_reg_arena_swap(core->dbg->reg, true);
-		signum = rz_num_math(core->num, input + 2);
-		ptr = strchr(input + 3, ' ');
+		signum = rz_num_math(core->num, input + 1);
+		ptr = strchr(input + 2, ' ');
 		if (ptr) {
 			int old_pid = core->dbg->pid;
 			int old_tid = core->dbg->tid;
@@ -4377,12 +4386,12 @@ static int cmd_debug_continue(RzCore *core, const char *input) {
 		}
 		break;
 	case 's': // "dcs"
-		switch (input[2]) {
+		switch (input[1]) {
 		case '*':
 			cmd_debug_cont_syscall(core, "-1");
 			break;
 		case ' ':
-			cmd_debug_cont_syscall(core, input + 3);
+			cmd_debug_cont_syscall(core, input + 2);
 			break;
 		case '\0':
 			cmd_debug_cont_syscall(core, NULL);
@@ -4417,24 +4426,26 @@ static int cmd_debug_continue(RzCore *core, const char *input) {
 		return 1;
 	}
 	case 'u': // "dcu"
-		if (input[2] == '?') {
+		if (input[1] == '?') {
 			rz_core_cmd_help(core, help_msg_dcu);
-		} else if (input[2] == '.') {
+		} else if (input[1] == '.') {
 			cmd_dcu(core, "cu $$");
 		} else {
-			cmd_dcu(core, input);
+			char *tmpinp = rz_str_newf("cu %s", input + 2);
+			cmd_dcu(core, tmpinp);
+			free(tmpinp);
 		}
 		break;
 	case ' ':
 		old_pid = core->dbg->pid;
-		pid = atoi(input + 2);
+		pid = atoi(input + 1);
 		rz_reg_arena_swap(core->dbg->reg, true);
 		rz_debug_select(core->dbg, pid, core->dbg->tid);
 		rz_debug_continue(core->dbg);
 		rz_debug_select(core->dbg, old_pid, core->dbg->tid);
 		break;
 	case 't':
-		cmd_debug_backtrace(core, input + 2);
+		cmd_debug_backtrace(core, input + 1);
 		break;
 	case '?': // "dc?"
 	default:
@@ -4651,20 +4662,6 @@ static void consumeBuffer(RzBuffer *buf, const char *cmd, const char *errmsg) {
 	rz_cons_printf("\n");
 }
 
-RZ_IPI int rz_debug_continue_oldhandler(void *data, const char *input) {
-	RzCore *core = (RzCore *)data;
-	rz_reg_arena_swap(core->dbg->reg, true);
-	#if __linux__
-		core->dbg->continue_all_threads = true;
-	#endif
-	if (rz_debug_is_dead(core->dbg)) {
-		eprintf("Cannot continue, run ood?\n");
-		return 0;
-	}
-	rz_debug_continue(core->dbg);
-	return 0;
-}
-
 RZ_IPI int rz_cmd_debug(void *data, const char *input) {
 	RzCore *core = (RzCore *)data;
 	RzDebugTracepoint *t;
@@ -5031,7 +5028,7 @@ RZ_IPI int rz_cmd_debug(void *data, const char *input) {
 		break;
 	case 'c': // "dc"
 		rz_cons_break_push(static_debug_stop, core->dbg);
-		(void)cmd_debug_continue(core, input);
+		(void)rz_debug_continue_oldhandler(core, input + 1);
 		follow = rz_config_get_i(core->config, "dbg.follow");
 		rz_cons_break_pop();
 		break;
-- 
2.26.2

```

librz/core/cmd_analysis.c
```diff
@@ -9340,16 +9340,16 @@ static int cmd_analysis_all(RzCore *core, const char *input) {
 	case 'c': // "aac"
 		switch (input[1]) {
 		case '*': // "aac*"
-			cmd_analysis_calls(core, input + 1, true, false);
+			rz_cmd_analysis_calls (core, input + 1, true, false);
```
ret2libc: make sure to run `git-clang-format` to respect the clang-format style. In particular, no more spaces before function calls.

librz/core/cmd_debug.c
```diff
@@ -4425,24 +4426,26 @@ static int cmd_debug_continue(RzCore *core, const char *input) {
...
 	case 'u': // "dcu"
-		if (input[2] == '?') {
+		if (input[1] == '?') {
 			rz_core_cmd_help(core, help_msg_dcu);
-		} else if (input[2] == '.') {
+		} else if (input[1] == '.') {
 			cmd_dcu(core, "cu $$");
 		} else {
-			cmd_dcu(core, input);
+			char *tmpinp = rz_str_newf("cu %s", input + 2);
```
ret2libc, line 4435: Ok the last errors are probably here.. I think I messed up a bit with the call to `cmd_dcu`. Could you check what `cmd_dcu` expects (in the old code) and update this accordingly?

ret2libc:
```diff
diff --git a/librz/core/cmd_debug.c b/librz/core/cmd_debug.c
index 7ab3236823..757407ee5e 100644
--- a/librz/core/cmd_debug.c
+++ b/librz/core/cmd_debug.c
@@ -4315,6 +4315,7 @@ RZ_IPI int rz_debug_continue_oldhandler(void *data, const char *input) {
        RzCore *core = (RzCore *)data;
        int pid, old_pid, signum;
        char *ptr;
+       rz_cons_break_push(static_debug_stop, core->dbg);
        // TODO: we must use this for step 'ds' too maybe...
        switch (input[0]) {
        case 0: // "dc"
@@ -4452,6 +4453,14 @@ RZ_IPI int rz_debug_continue_oldhandler(void *data, const char *input) {
                rz_core_cmd_help(core, help_msg_dc);
                return 0;
        }
+       int follow = rz_config_get_i(core->config, "dbg.follow");
+       rz_cons_break_pop();
+       if (follow > 0) {
+               ut64 pc = rz_debug_reg_get(core->dbg, "PC");
+               if ((pc < core->offset) || (pc > (core->offset + follow))) {
+                       rz_core_cmd0(core, "sr PC");
+               }
+       }
        return 1;
 }
 
@@ -5014,10 +5023,8 @@ RZ_IPI int rz_cmd_debug(void *data, const char *input) {
                eprintf("TODO: transplant process\n");
                break;
        case 'c': // "dc"
-               rz_cons_break_push(static_debug_stop, core->dbg);
                (void)rz_debug_continue_oldhandler(core, input + 1);
-               follow = rz_config_get_i(core->config, "dbg.follow");
-               rz_cons_break_pop();
+               follow = 0;
                break;
        case 'm': // "dm"
                cmd_debug_map(core, input + 1);

```

ret2libc (approved):

PR #433, opened 2021-01-22
Remove support for ACR/Make

### LINKED ISSUE #393: Kill ACR/Make and switch completely to Meson
We chose to switch to meson by default and kill ACR/Make. This bug is to track the removal of ACR/Makefiles from the entire codebase so that we can just maintain one buildsystem and remove a lot of scripts files all around.

Before doing so, we must ensure that all CI jobs use meson.

**Additional context**
- [x] Migrate TinyCC builds to Meson/Ninja
- [x] Migrate CentOS builds to Meson/Ninja
- [ ] Use `project` version field in meson.build instead of taking it from configure.acr
- [x] Remove configure.acr / configure
- [x] Remove all .mk files
- [x] Remove all Makefiles
- [x] Remove all related sys/ scripts
- [ ] Remove all references from BUILDING.md, book and various documentation

See also https://github.com/rizinorg/rizin/pull/191 and https://github.com/rizinorg/rizin/issues/337

### LINKED ISSUE #432: Use `v` prefix in all release artifacts
Instead of rizin-src-0.1.0 we want to have rizin-src-v0.1.0. Same for all the artifacts.

ret2libc:
**DO NOT SQUASH**

- [x] I've updated the documentation and the [rizin book](https://github.com/rizinorg/book) with the relevant information (if needed). See https://github.com/rizinorg/book/pull/28

**Detailed description**

Remove all ACR/Makefiles and many shell scripts in ./sys

TODO from the related issue:
- [x] Use project version field in meson.build instead of taking it from configure.acr
- [x] Remove all references from BUILDING.md, book and various documentation

**Test plan**

All green, except TCC build because meson does not support it. However, I have already a working patch for meson in https://github.com/ret2libc/meson/commit/fff2169843819beb45743449b9a614e32d47726d . It probably needs a bit of cleaning and then preparing it for upstream, but it works (tested locally).

**Closing issues**

Closes https://github.com/rizinorg/rizin/issues/393
Closes https://github.com/rizinorg/rizin/issues/432


shlr/meson.build
```diff
@@ -286,9 +291,10 @@ qnx_files = [
 qnx_inc = [platform_inc, include_directories('qnx/include')]
 
 librzqnx = static_library('rzqnx', qnx_files,
-  dependencies: [rz_socket_dep, rz_util_dep],
+  dependencies: [rz_util_dep, rz_socket_dep],
```
yossizap, line 287: Maybe we should document that the dependencies have to be written in their linking order to support TCC since it's not standard meson behavior. Might be frustrating for someone to figure out why tinycc broke after modifying meson files.

shlr/meson.build
```diff
@@ -286,9 +291,10 @@ qnx_files = [
...
   include_directories: qnx_inc,
-  implicit_include_directories: false
+  implicit_include_directories: false,
+  pic: true,
```
yossizap: Same as the library linking order - maybe this should be in a section with meson guidelines in DEVELOPERS.md?
Will it be an issue to just pass it as a C flag universally?

.github/workflows/tcc.yml
```diff
@@ -19,6 +19,7 @@ jobs:
     name: ubuntu-tcc-test
     runs-on: ubuntu-latest
     steps:
+    # TODO: need to add support for TCC in meson
```
yossizap: Just disable it until your meson PR is merged? It's not like it's a real target, we can merge this PR without TinyCC and revisit it later.

shlr/meson.build
ret2libc: Actually I need to check whether this is really necessary... maybe by improving the TCC support on meson I did this can be avoided. Need to check better though, but I wanted to start pushing the PR to see if something else broke.

XVilka:
can you temporarily enable also the usual `dist-*` jobs for this PR too? So we will see what pre-prelease CI tasks break if any.

shlr/meson.build
ret2libc: Ok indeed this is not actually necessary. I was able to fix the tcc code in meson to avoid this, the same it is done when using other compilers.

XVilka:
What about just using Meson from your branch https://github.com/ret2libc/meson/tree/tiny-cc only in TinyCC CI job? This way we can test both and proceed with actually merging this PR right now. 
Please also fix the "CodeQL" build.

shlr/meson.build
ret2libc: Ok I actually found out that all these reordering was not needed, it was just my tcc plugin in Meson that was broken. However, I'd still keep this commit because it removes some unnecessary dependencies.

meson.build
```diff
@@ -1,5 +1,5 @@
 project('rizin', 'c',
-  version: '0.2.0-git',
+  version: 'v0.2.0-git',
```
thestr4ng3r, line 2: I think this should be `0.2.0-git`, only tag names should be `v...` but can be discussed.

ret2libc: Do you expect the tarball to contain rizin-0.2.0-git directory or rizin-v0.2.0-git?

thestr4ng3r: Doesn't matter I think. I find `v0.2.0` a bit nicer in the tarball since the file itself is called like that too but would be fine with both.

ret2libc: so, if we use `v0.2.0-git`, the directory within the tarball is called `rizin-v0.2.0-git`. If we use `0.2.0-git` the directory is called `rizin-0.2.0-git`.Do you want to keep just the tag as `v{version}` and then all the artifacts as `{version}`(without the `v`)?

thestr4ng3r: No, I prefer `v0.2.0` for the artifact names and the folder inside of it. So unless having the `v` in the `project()` has any negative implications, I think the state in the pr here is good.

librz/core/cmd_help.c
```diff
@@ -876,7 +876,6 @@ RZ_IPI int rz_cmd_help(void *data, const char *input) {
 			pj_ks(pj, "arch", RZ_SYS_ARCH);
 			pj_ks(pj, "os", RZ_SYS_OS);
 			pj_ki(pj, "bits", RZ_SYS_BITS);
-			pj_ks(pj, "tap", RZ_GITTAP);
```
wargio, line 879: Is this the git version stuff that you removed?

wargio (approved):
why `test/Makefile` was not deleted?

librz/core/cmd_help.c
ret2libc: Yes. https://github.com/rizinorg/rizin/pull/433/files#diff-30d8f6be6320feeacf686be94f48c70869b52630e01ea625f0f15adc0d57c3e4L65 It essentially took the last git tag. However, as said, I don't know why we would want that. I think we only care about RZ_VERSION.

ANyway, given how big this diff is maybe it's better if I extract in a separate PR this commit so you can review it better.

ret2libc:
> why `test/Makefile` was not deleted?

Because it just contains the `bins` target, which is used to download the binary directory. I think it's the best option.

librz/core/cmd_help.c
ret2libc: Ok look at https://github.com/rizinorg/rizin/pull/480 first.

thestr4ng3r (approved):

PR #464, opened 2021-01-26
Fix "‘%s’ directive argument is null" warnings

kazarmy:

**Detailed description**

This pr fixes warnings like the following in #260 (https://github.com/rizinorg/rizin/pull/260/checks?check_run_id=1763197110#step:12:607 -- also search for `-Wformat-overflow=`):

directive-argument-null

**Test plan**

All builds are green.

**Closing issues**

...

### DIFF
--- a/librz/asm/arch/tms320/c55x_plus/c55plus_decode.c
+++ b/librz/asm/arch/tms320/c55x_plus/c55plus_decode.c
@@ -234,7 +234,7 @@ static char *decode_ins(st32 hash_code, ut32 ins_pos, ut32 ins_off, ut32 *ins_le
 	// get pseudo instruction
 	ins = ins_str[1 + 2 + hash_code * 4];
 	if (!ins /*|| ins_str[4 * hash_code] == 0*/) {
-		fprintf(stderr, "Invalid instruction %s /hash %x\n", ins, hash_code);
+		fprintf(stderr, "Invalid instruction hash %x\n", hash_code);
 		*err_code = -1;
 		return NULL;
 	}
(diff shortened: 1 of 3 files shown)


librz/asm/arch/tms320/c55x_plus/c55plus_decode.c
```diff
@@ -234,7 +234,7 @@ static char *decode_ins(st32 hash_code, ut32 ins_pos, ut32 ins_off, ut32 *ins_le
 	// get pseudo instruction
 	ins = ins_str[1 + 2 + hash_code * 4];
 	if (!ins /*|| ins_str[4 * hash_code] == 0*/) {
-		fprintf(stderr, "Invalid instruction %s /hash %x\n", ins, hash_code);
```
ret2libc, line 237: why removed the instruction printed?

ret2libc: ah nothing sorry, ignore me

ret2libc (approved):

librz/core/cmd_eval.c
```diff
@@ -238,7 +238,7 @@ static void nextpal(RzCore *core, int mode) {
 						rz_list_free(files);
 						return;
 					}
-					eprintf("%s %s %s\n", nfn, curtheme, fn);
+					eprintf("%s %s %s\n", nfn ? nfn : "(null)", curtheme, fn);
```
XVilka, line 241: By the way, I think we can use `rz_str_get{2}()` in these cases

ret2libc: If we want to use such thing though we should make the function static inline and move it to rz_str.h because it is so small that a call would be wasted time

wargio (approved):

librz/core/cmd_eval.c
wargio: agree

PR #481, opened 2021-01-28
Remove obsolete iphone.md

Maijin:
Remove obsolete iphone.md


wargio:
how old is this?

ret2libc (commented):
I didn't remove them because we don't have any "alternative" version of this... But I think these things can't be easily supported/maintained unless people have many jailbroken devices :) Probably people who go that route can also find their way to compile Rizin there as well.

thestr4ng3r (approved):
iPhone OS was renamed to iOS in 2010 and it's still mentioned like that in this file so I think it's a bit obsolete.

PR #483, opened 2021-01-28
Use OLDINPUT command descriptor for =! to allow passing cmd to IO plugin

ret2libc:

**Detailed description**

This is just a temporary solution, because new commands should be registered to RzCmd and not executed through an IO plugin. The problem here is that `=!pid`(and similar) were considered as a command name, but newshell was not aware of any command named `=!pid`. Actually, the command is `=!` and it passes whatever is after that to the IO plugin. By transforming the `=!` command in OLDINPUT, newshell automatically tries to remove letter after letter from `=!pid` until it finds a command descriptor and then it passes the whole input string to the proper handler.

**Test plan**

Execute `gdbserver :12345 /bin/ls` on one shell.
On another do `rizin -d gdb://localhost:12345` and then try `=?` and `=!?` and `=!pid`. You should be able to see a list of gdb-specific commands and execute them.

**Closing issues**

...


librz/core/cmd_remote.c
```diff
@@ -168,15 +168,14 @@ RZ_IPI RzCmdStatus rz_remote_send_handler(RzCore *core, int argc, const char **a
...
+RZ_IPI int rz_io_system_run_oldhandler(void *data, const char *input) {
+	RzCore *core = (RzCore *)data;
+	char *res = rz_io_system(core->io, input);
 	if (res) {
 		rz_cons_printf("%s\n", res);
 		free(res);
 	}
-	free(args);
-	return RZ_CMD_STATUS_OK;
+	return 0;
```
wargio, line 178: why?

ret2libc: because I had to change this handler to OLDINPUT handler, which returns a int not a RzCmdStatus. And in "oldshell" 0 means "all's good!".

wargio: oh, i see.

wargio (approved):

ret2libc:
@yossizap please confirm this works well for the usecase you reported before merging this.

yossizap (approved):
LGMT. Tested with gdb and it's in parity with the original functionality.

PR #486, opened 2021-01-29
Fixes for Dalvik and DEX

XVilka:
**DO NOT SQUASH**

**Detailed description**

Notable changes are fixes in Dalvik and DEX parsing, dropping old MODE_DIST from the rz-diff code, various memory leaks.

### DIFF
--- a/librz/analysis/dwarf_process.c
+++ b/librz/analysis/dwarf_process.c
@@ -102,6 +102,7 @@ static bool strbuf_rev_prepend_char(RzStrBuf *sb, const char *s, int c) {
 	char *sb_str = sb->ptr ? sb->ptr : sb->buf;
 	char *pivot = strrchr(sb_str, c);
 	if (!pivot) {
+		free(ns);
 		return false;
 	}
 	size_t idx = pivot - sb_str;
@@ -869,12 +870,12 @@ static const char *map_dwarf_reg_to_ppc64_reg(ut64 reg_num, VariableLocationKind
 	case 6: return "r6";
 	case 7: return "r7";
 	case 8: return "r8";
-	case 9: return "r8";
-	case 10: return "r9";
-	case 11: return "r10";
-	case 12: return "r11";
-	case 13: return "r12";
-	case 14: return "r13";
+	case 9: return "r9";
+	case 10: return "r10";
+	case 11: return "r11";
+	case 12: return "r12";
+	case 13: return "r13";
+	case 14: return "r14";
 	case 15: return "r15";
 	case 16: return "r16";
 	case 17: return "r17";
--- a/librz/cons/dietline.c
+++ b/librz/cons/dietline.c
@@ -368,7 +368,7 @@ RZ_API int rz_line_hist_cmd_up(RzLine *line) {
 		setup_hist_match(line);
 		if (line->history.match) {
 			int i;
-			for (i= line->history.index - 1; i >= 0; i--) {
+			for (i = line->history.index - 1; i >= 0; i--) {
 				if (match_hist_line(line->history.data[i], line->history.match)) {
 					line->history.index = i;
 					break;
@@ -398,7 +398,7 @@ RZ_API int rz_line_hist_cmd_down(RzLine *line) {
 	if (line->history.match) {
 		int i;
 		for (i = line->history.index + 1; i < line->history.top; i++) {
-			if (match_hist_line(line->history.data[i], line->history.match)) {	
+			if (match_hist_line(line->history.data[i], line->history.match)) {
 				break;
 			}
 		}
@@ -1209,6 +1209,7 @@ static void __vi_mode(void) {
 			break;
 		case 'c':
 			I.vi_mode = INSERT_MODE; // goto insert mode
+			/* fall through */
 		case 'd': {
 			char c = rz_cons_readchar();
 			while (rep--) {
@@ -1293,6 +1294,7 @@ static void __vi_mode(void) {
 			break;
 		case 'a':
 			__move_cursor_right();
+			/* fall through */
 		case 'i':
 			I.vi_mode = INSERT_MODE;
 			if (I.hud) {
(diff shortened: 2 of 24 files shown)


thestr4ng3r (dismissed):

librz/analysis/dwarf_process.c
```diff
@@ -869,12 +870,12 @@ static const char *map_dwarf_reg_to_ppc64_reg(ut64 reg_num, VariableLocationKind
 	case 6: return "r6";
 	case 7: return "r7";
 	case 8: return "r8";
-	case 9: return "r8";
-	case 10: return "r9";
-	case 11: return "r10";
-	case 12: return "r11";
-	case 13: return "r12";
-	case 14: return "r13";
+	case 9: return "r9";
```
wargio, line 873: this stuff needs to be rewritten. i don't see any reason to have something like this

XVilka: @wargio good point, added a new commit to simplify this code.

librz/analysis/dwarf_process.c
```diff
@@ -857,42 +858,12 @@ static const char *map_dwarf_reg_to_x86_reg(ut64 reg_num, VariableLocationKind *
...
-	case 25: return "r25";
-	case 26: return "r26";
-	case 27: return "r27";
-	case 28: return "r28";
-	case 29: return "r29";
-	case 30: return "r30";
-	case 31: return "r31";
-	default:
+	} else if (reg_num < 32) {
+		return rz_str_newf("r%"PFMT64d, reg_num);
```
ret2libc: this is wrong. It now allocates a string every time and the function returns `const char *` so I doubt it is freed by the caller.

librz/cons/dietline.c
```diff
@@ -1293,6 +1294,7 @@ static void __vi_mode(void) {
 			break;
 		case 'a':
 			__move_cursor_right();
+			break;
```
ret2libc: is this right? I think there should be a fallthrough here as well, similar to `A`. If I get it right, this behaves like `a` in vim, so it's similar to `i` but you move first the cursor to the right.

librz/cons/dietline.c
```diff
@@ -1209,6 +1209,7 @@ static void __vi_mode(void) {
 			break;
 		case 'c':
 			I.vi_mode = INSERT_MODE; // goto insert mode
+			break;
```
ret2libc: did not check this... is this really right as well?

librz/cons/dietline.c
```diff
@@ -1888,6 +1890,7 @@ RZ_API const char *rz_line_readline_cb(RzLineReadCallback cb, void *user) {
 						break;
 					case 0x37: // HOME xrvt-unicode
 						rz_cons_readchar();
+						break;
```
ret2libc: same here... mmm need to test all these things.

librz/core/cfile.c
```diff
@@ -909,6 +909,7 @@ RZ_API RzCoreFile *rz_core_file_open(RzCore *r, const char *file, int flags, ut6
 	if (rz_io_is_listener(r->io)) {
 		rz_core_serve(r, fd);
 		rz_io_desc_free(fd);
+		fd = NULL;
```
ret2libc: why is this necessary? It seems it's a dead assignment, because `fd`is not used in `beach`.

librz/core/disasm.c
```diff
@@ -2416,7 +2416,6 @@ static void ds_update_ref_lines(RDisasmState *ds) {
 		free(ds->prev_line_col);
 		ds->refline = strdup("");
 		ds->refline2 = strdup("");
-		ds->prev_line_col = strdup("");
```
ret2libc, line 2419: I'm not sure why this... Do we know?

XVilka: It is assigned NULL later in the function.

librz/core/cfile.c
XVilka: Good catch! Removed that line.

librz/cons/dietline.c
XVilka: Yes, you are right. I added a fall-through comment to indicate that.

librz/cons/dietline.c
XVilka: Yes, was wrong as well. Added a fall-through comment to indicate that.

librz/cons/dietline.c
XVilka: Added a fall-through comment to indicate that.

librz/analysis/dwarf_process.c
wargio: i agree. @XVilka can you open an issue to refactor this? because this needs more work.

librz/core/disasm.c
ret2libc: Right!

librz/analysis/dwarf_process.c
XVilka: @ret2libc the simplification will involve changing more than just one function. Are you ok with merging this fix and do the rework of that file later?

librz/core/canalysis.c
```diff
@@ -2385,14 +2385,15 @@ RZ_API void rz_core_analysis_callgraph(RzCore *core, ut64 addr, int fmt) {
 	case RZ_GRAPH_FORMAT_DOT:
 		if (!is_html) {
 			const char *gv_edge = rz_config_get(core->config, "graph.gv.edge");
-			const char *gv_node = rz_config_get(core->config, "graph.gv.node");
+			char *gv_node = strdup(rz_config_get(core->config, "graph.gv.node"));
```
ret2libc, line 2388: this is going to leak because it is never freed (apart when it is reallocated).

ret2libc (commented):
LGTM apart from that small thing

ret2libc (approved):

Issue #382, opened 2021-01-18
Refactor Visual mode (`librz/core/tui/visual.c`) to use API instead of commands


XVilka:
Currently, the code of Visual modes heavily relies on running commands instead of just calling the proper API functions.
To reduce the overhead, code cruft, and code breakage in case of the command changes it is important to remove all those command calls and call the proper API directly.

```
[i] ℤ rg "rz_core_cmd[0f_]*\(" librz/core/tui/visual.c                                                                                                                                                                                 12:21:36 
683:          rz_core_cmd(core, buf, 0);
715:                  ch = rz_core_cmd0(core, cmd);
723:                  ch = rz_core_cmd0(core, cmd);
731:                  ch = rz_core_cmd0(core, cmd);
737:                  ch = rz_core_cmd0(core, cmd);
748:                  ch = rz_core_cmd0(core, cmd);
754:                  ch = rz_core_cmd0(core, cmd);
760:                  ch = rz_core_cmd0(core, cmd);
768:                  ch = rz_core_cmd0(core, cmd);
776:                  ch = rz_core_cmd0(core, cmd);
784:                  ch = rz_core_cmd0(core, cmd);
790:                  ch = rz_core_cmd0(core, cmd);
796:                  ch = rz_core_cmd0(core, cmd);
1992:                 rz_core_cmd0(core, "eco $(eco~...)");
1998:                 rz_core_cmd0(core, "s $(afb~...)");
2002:                 rz_core_cmd0(core, "s $(ii~...)");
2005:                 rz_core_cmd0(core, "s $(isq~...)");
2008:                 rz_core_cmd0(core, "s $(dm~...)");
2017:                 rz_core_cmdf(core, "sh~...");
2249:                         rz_core_cmd(core, buf, true);
2502:                         rz_core_cmd(core, buf, 0);
2529:                 rz_core_cmd(core, buf, 1);
2800:                         rz_core_cmd0(core, key_s);
2808:                         rz_core_cmd0(core, key_s);
2914:                                 rz_core_cmd(core, buf, 0);
2938:                                 rz_core_cmd(core, buf, 0);
3656:                 rz_core_cmd0(core, vi);
3670:         rz_core_cmd0(core, vi);
3712:                 rz_core_cmd0(core, cmd_str);
3841:         int ret = rz_core_cmd0(core, cmd);
```

officialcjunior:
> rz_core_cmd0(core, "pg");

[...]

PR #487, opened 2021-01-29
Refactor to use API in Visual mode

XVilka:
**DO NOT SQUASH**

- [x] made sure to follow the project's [coding style](https://github.com/rizinorg/rizin/blob/master/DEVELOPERS.md#code-style)

**Detailed description**

Refactor Visual mode to use API instead of `rz_core_cmd*` functions in some cases - debug continue until and resizing commands.

Also introduced new API calls for better understanding what config variables are bool and check it more granularly:
- `rz_config_get_b(RzConfig *cfg, const char *name)`
- `rz_config_set_b(RzConfig *cfg, const char *name, bool value)`

To reduce unnecessary `rz_core_cmdf()` calls in visual browse also switched `af` calls to the newly introduced `RZ_API bool rz_core_analysis_function_add()`.

**Test plan**

Go into Visual mode, e.g. `Vp` and try to press `s` or F8  in both debug and ESIL modes. Also try `g` key.
Also try to resize the file by pressing `r` (-1 byte) and `R` (+1 byte).

**Closing issues**

Partially addresses https://github.com/rizinorg/rizin/issues/382

### DIFF
--- a/librz/config/config.c
+++ b/librz/config/config.c
@@ -293,6 +293,9 @@ RZ_API bool rz_config_set_setter(RzConfig *cfg, const char *key, RzConfigCallbac
 	return false;
 }
 
+/**
+ * Returns the value of the config variable of \p name as a string
+ */
 RZ_API const char *rz_config_get(RzConfig *cfg, const char *name) {
 	rz_return_val_if_fail(cfg && name, NULL);
 	RzConfigNode *node = rz_config_node_get(cfg, name);
@@ -310,6 +313,11 @@ RZ_API const char *rz_config_get(RzConfig *cfg, const char *name) {
 	return NULL;
 }
 
+/**
+ * Reads the value of the config variable of \p name only and only if
+ * the variable is boolean, then tries to write back the inverted value.
+ * Returns true in case of success.
+ */
 RZ_API bool rz_config_toggle(RzConfig *cfg, const char *name) {
 	RzConfigNode *node = rz_config_node_get(cfg, name);
 	if (!node) {
@@ -327,6 +335,10 @@ RZ_API bool rz_config_toggle(RzConfig *cfg, const char *name) {
 	return true;
 }
 
+/**
+ * Reads the value of the config variable of \p name only and only if
+ * the variable is integer.
+ */
 RZ_API ut64 rz_config_get_i(RzConfig *cfg, const char *name) {
 	RzConfigNode *node = rz_config_node_get(cfg, name);
 	if (node) {
@@ -342,6 +354,22 @@ RZ_API ut64 rz_config_get_i(RzConfig *cfg, const char *name) {
 		return (ut64)rz_num_math(cfg->num, node->value);
 	}
 	return (ut64)0LL;
+}
+
+/**
+ * Reads the value of the config variable of \p name only and only if
+ * the variable is boolean. Returns false in case of the failure.
+ */
+RZ_API bool rz_config_get_b(RzConfig *cfg, const char *name) {
+	RzConfigNode *node = rz_config_node_get(cfg, name);
+	if (!node) {
+		return false;
+	}
+	if (!rz_config_node_is_bool(node)) {
+		RZ_LOG_DEBUG("(error: '%s' is not a boolean variable)\n", name);
+		return false;
+	}
+	return rz_str_is_true(node->value);
 }
 
 RZ_API const char *rz_config_node_type(RzConfigNode *node) {
@@ -386,7 +414,78 @@ static bool __is_true_or_false(const char *s) {
 	return s && (!rz_str_casecmp(s, "true") || !rz_str_casecmp(s, "false"));
 }
 
+/**
+ * Writes the boolean \p value in the config variable of \p name only and only if
+ * the variable is boolean.
+ */
+RZ_API RzConfigNode *rz_config_set_b(RzConfig *cfg, const char *name, bool value) {
+	RzConfigNode *node = NULL;
+	char *ov = NULL;
+	ut64 oi = 0;
+
+	rz_return_val_if_fail(cfg && cfg->ht, NULL);
+	rz_return_val_if_fail(!IS_NULLSTR(name), NULL);
+
+	node = rz_config_node_get(cfg, name);
+	if (node) {
+		if (rz_config_node_is_ro(node)) {
+			RZ_LOG_DEBUG("(error: '%s' config key is read only)\n", name);
+			return node;
+		}
+
+		oi = node->i_value;
+		if (node->value) {
+			ov = strdup(node->value);
+		}
+		if (rz_config_node_is_bool(node)) {
+			node->i_value = value ? 1 : 0;
+			char *svalue = strdup(rz_str_bool(value));
+			if (svalue) {
+				free(node->value);
+				node->value = svalue;
+			}
+		} else {
+			RZ_LOG_ERROR("(error: '%s' is not a boolean variable)\n", name);
+			free(ov);
+			return NULL;
+		}
+	} else {
+		if (!cfg->lock) {
+			node = rz_config_node_new(name, rz_str_bool(value));
+			if (!node) {
+				node = NULL;
+				goto beach;
+			}
+			node->flags = CN_RW | CN_BOOL;
+			node->i_value = value ? 1 : 0;
+			ht_pp_insert(cfg->ht, node->name, node);
+			if (cfg->nodes) {
+				rz_list_append(cfg->nodes, node);
+			}
+		} else {
+			RZ_LOG_ERROR("(locked: no new keys can be created (%s))\n", name);
+		}
+	}
+
+	if (node && node->setter) {
+		if (!node->setter(cfg->user, node)) {
+			if (oi != UT64_MAX) {
+				node->i_value = oi;
+			}
+			free(node->value);
+			node->value = strdup(ov ? ov : "");
+		}
+	}
+
+beach:
+	free(ov);
+	return node;
+}
+
 /* TODO: reduce number of strdups here */
+/**
+ * Writes the string \p value in the config variable of \p name.
+ */
 RZ_API RzConfigNode *rz_config_set(RzConfig *cfg, const char *name, const char *value) {
 	RzConfigNode *node = NULL;
 	char *ov = NULL;
@@ -516,6 +615,10 @@ RZ_API void rz_config_node_value_format_i(char *buf, size_t buf_size, const ut64
 	}
 }
 
+/**
+ * Writes the integer \p value in the config variable of \p name only and only if
+ * the variable is integer.
+ */
 RZ_API RzConfigNode *rz_config_set_i(RzConfig *cfg, const char *name, const ut64 i) {
 	char buf[128], *ov = NULL;
 	rz_return_val_if_fail(cfg && name, NULL);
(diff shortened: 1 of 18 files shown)


librz/config/config.c
```diff
@@ -386,6 +386,50 @@ static bool __is_true_or_false(const char *s) {
 	return s && (!rz_str_casecmp(s, "true") || !rz_str_casecmp(s, "false"));
 }
 
+RZ_API RzConfigNode *rz_config_set_b(RzConfig *cfg, const char *name, bool value) {
+	RzConfigNode *node = NULL;
+	char *ov = NULL;
+	ut64 oi = 0;
+
+	rz_return_val_if_fail(cfg && cfg->ht, NULL);
+	rz_return_val_if_fail(!IS_NULLSTR(name), NULL);
```
ret2libc: Use `RZ_STR_ISEMPTY` please. We probably should kill this `IS_NULLSTR`.

librz/config/config.c
```diff
@@ -386,6 +386,50 @@ static bool __is_true_or_false(const char *s) {
...
+				node->value = svalue;
+			}
+		} else {
+			RZ_LOG_DEBUG("(error: '%s' is not a boolean variable)\n", name);
+			return NULL;
+		}
+	}
+	if (node && node->setter) {
+		if (!node->setter(cfg->user, node)) {
+			if (oi != UT64_MAX) {
```
ret2libc, line 473: If you took this code from `rz_config_set_i`, it seems that `oi` comes from `node->i_value`. In this function instead it is always 0.

librz/config/config.c
```diff
@@ -386,6 +386,50 @@ static bool __is_true_or_false(const char *s) {
...
+				node->i_value = oi;
+			}
+			free(node->value);
+			node->value = strdup(ov ? ov : "");
```
ret2libc, line 477: also this seems slightly different from the `rz_config_set_i` code. Which one is right?

ret2libc: I would say this one, the other probably causes a use after free mmm

librz/config/config.c
```diff
@@ -386,6 +386,50 @@ static bool __is_true_or_false(const char *s) {
 	return s && (!rz_str_casecmp(s, "true") || !rz_str_casecmp(s, "false"));
 }
 
+RZ_API RzConfigNode *rz_config_set_b(RzConfig *cfg, const char *name, bool value) {
```
ret2libc, line 422: It seems you don't allow insertion of new evals in this function, while it is allowed in rz_config_set_i.
I agree with this choice in general, however i'm not sure if only done here is the right thing. Maybe we should allow insertion of new vars also here and think about RzConfig later. I think also RzConfig would benefit from a more general refactoring similar to newshell :D

ret2libc (changes requested):

librz/config/config.c
XVilka: I would kill insertion everywhere.

ret2libc: I agree, but probably a separate PR and last time i tried it caused some problems here and there because in the current state it is not always clear where a eval is really created the first time.

librz/config/config.c
XVilka: Made sure the code is the same everywhere.

XVilka:
I also noticed that the action for toggling hints was not the same in visual/panels mode and ASCII graphs mode. I made it the same, please let me know if you think it was the right move.

librz/config/config.c
XVilka: Ok, added insertion too.

librz/config/config.c
XVilka: It is set in line 418: `oi = node->i_value;`

ret2libc (changes requested):
If you are going to add new API, especially in RzConfig, we need:
- doxygen doc for the new methods (like rz_config_get_b and set_b)
- unit tests for those functions.

librz/config/config.c
```diff
@@ -344,6 +345,18 @@ RZ_API ut64 rz_config_get_i(RzConfig *cfg, const char *name) {
 	return (ut64)0LL;
 }
 
+RZ_API bool rz_config_get_b(RzConfig *cfg, const char *name) {
```
ret2libc, line 364: I see a problem here as `false` is both an error value but a valid value. We should probably add an argument for the error or make the return value the error and passing the value of the config var through the arg.

librz/config/config.c
```diff
@@ -386,14 +399,76 @@ static bool __is_true_or_false(const char *s) {
...
+		}
+	}
+
+	if (node && node->setter) {
+		if (!node->setter(cfg->user, node)) {
+			if (oi != UT64_MAX) {
+				node->i_value = oi;
+			}
+			free(node->value);
+			node->value = strdup(ov ? ov : "");
```
ret2libc, line 477: UAF reported by clang here.

librz/config/config.c
XVilka: In general I agree but it would make this API inconsistent with the `rz_config_get()` and `rz_config_get_i()` APIs.

ret2libc: That's why i said we need unit tests and I'm not sure this is the right time to do this... I would first refactor existing API and then add this new one. Is it really necessary for the purpose of pr? Can't you just use get_i for now and do a full refactoring with all the needed tests in a separate pr?

XVilka: It already improves the situation. I can do a separate refactoring changing the RzConfig API everywhere after this one is done. Is this fine for you?

ret2libc: Add at least doxygen and basic unit tests for coverage. This is the way to introduce new APIs

XVilka: Ok, will do.

librz/config/config.c
ret2libc: @XVilka have you seen this?

librz/core/canalysis.c
```diff
@@ -110,6 +120,21 @@ static char *getFunctionName(RzCore *core, ut64 addr) {
 	return (flag && flag->name) ? strdup(flag->name) : NULL;
 }
 
+static char *getFunctionNamePrefix(RzCore *core, ut64 off, const char *name, bool prefix) {
```
ret2libc: parts of these two functions was just removed/is going to be removed by @08A PR.

librz/core/canalysis.c
```diff
@@ -5784,6 +5809,138 @@ static bool analyze_noreturn_function(RzCore *core, RzAnalysisFunction *f) {
...
+				if (ref->type != RZ_ANALYSIS_REF_TYPE_CODE && ref->type != RZ_ANALYSIS_REF_TYPE_CALL) {
+					/* only follow code/call references */
+					continue;
+				}
+				if (!rz_io_is_valid_offset(core->io, ref->addr, !core->analysis->opt.noncode)) {
+					continue;
+				}
+				rz_core_analysis_fcn(core, ref->addr, fcn->addr, RZ_ANALYSIS_REF_TYPE_CALL, depth);
+				/* use recursivity here */
+#if 1
```
ret2libc: remove `#if 1`

librz/core/cfile.c
```diff
@@ -198,6 +198,58 @@ RZ_API int rz_core_file_reopen(RzCore *core, const char *args, int perm, int loa
 	return ret;
 }
 
+RZ_API bool rz_core_file_resize(RzCore *core, ut64 newsize) {
```
ret2libc: haven't looked too much at these two functions, but they share a lot of code and if we split in functions I think it should be to reduce code duplication, not to increase it.

librz/core/vmenus.c
```diff
@@ -3127,7 +3127,7 @@ RZ_API void rz_core_visual_analysis(RzCore *core, const char *input) {
 		case '-':
 			switch (level) {
 			case 0:
-				rz_core_cmdf(core, "af-0x%" PFMT64x, addr);
+				rz_core_cmdf(core, "af- 0x%" PFMT64x, addr);
```
ret2libc, line 3130: i'm not sure if this work yet, because some existing command actually *require* no space.

test/unit/meson.build
```diff
@@ -23,6 +23,7 @@ if get_option('enable_tests')
     'bitmap',
     'buf',
     'cmd',
+	'config',
```
ret2libc: indentation

test/unit/test_config.c
```diff
@@ -0,0 +1,56 @@
...
+}
+
+bool test_config_lock() {
+	RzConfig *cfg = rz_config_new(NULL);
+	cfg->lock = 1;
+
+	// string variables
+	rz_config_set(cfg, "foo.bar", "bla");
+	const char *bla = rz_config_get(cfg, "foo.bar");
+	mu_assert_eq(bla, NULL, "String variable (locked)");
```
ret2libc:
```suggestion
	mu_assert_null(bla, "String variable (locked)");
```

test/unit/test_config.c
```diff
@@ -0,0 +1,56 @@
...
+
+	// integer variables
+	rz_config_set_i(cfg, "universe.question", 42);
+	int answer = rz_config_get_i(cfg, "universe.question");
+	mu_assert_neq(answer, 42, "Integer variable (locked)");
+
+	// boolean variables
+	rz_config_set_b(cfg, "true.or.false", true);
+	bool what = rz_config_get_b(cfg, "true.or.false");
+	mu_assert_eq(what, false, "Boolean variable (locked)");
```
ret2libc:
```suggestion
	mu_assert_false(what, "Boolean variable (locked)");
```

ret2libc (commented):
I assume that most chunks of code that were deleted/added were actually just moved without much changes, right?

librz/core/vmenus.c
XVilka: It works, I tried.

librz/core/canalysis.c
XVilka: Addressed this thing, also removed `prefix` parameter since it's not used in the new code after @08A changes.

librz/config/config.c
XVilka: Fixed, also fixed in another function with the same pattern.

librz/core/cfile.c
XVilka: Made them API wrappers around the function with the same logic.

librz/config/config.c
```diff
@@ -386,14 +415,85 @@ static bool __is_true_or_false(const char *s) {
...
+		if (rz_config_node_is_bool(node)) {
+			node->i_value = value ? 1 : 0;
+			char *svalue = strdup(rz_str_bool(value));
+			if (svalue) {
+				free(node->value);
+				node->value = svalue;
+			}
+		} else {
+			RZ_LOG_ERROR("(error: '%s' is not a boolean variable)\n", name);
+			return NULL;
```
ret2libc, line 451: leak of `ov`. See checks reported by github.

librz/config/config.c
```diff
@@ -558,7 +664,7 @@ RZ_API RzConfigNode *rz_config_set_i(RzConfig *cfg, const char *name, const ut64
 		if (!ret) {
 			node->i_value = oi;
 			free(node->value);
-			node->value = ov ? ov : strdup("");
+			node->value = strdup(ov ? ov : "");
 			ov = NULL;
```
ret2libc: shouldn't you free ov before setting it to NULL?

librz/config/config.c
```diff
@@ -386,14 +415,87 @@ static bool __is_true_or_false(const char *s) {
...
+
+	if (node && node->setter) {
+		if (!node->setter(cfg->user, node)) {
+			if (oi != UT64_MAX) {
+				node->i_value = oi;
+			}
+			free(node->value);
+			node->value = strdup(ov ? ov : "");
+			free(ov);
+			return NULL;
```
ret2libc: why do you return NULL here? The same code path in `rz_config_set_i` seems to behave differently.

XVilka: Fixed.

ret2libc (approved):

PR #492, opened 2021-01-29
Add a `?` in the help message of `a` to show that `ab?` exists

officialcjunior:

**Detailed description**

Recently, a person from the community pointed out that here was no 
indication that help for ab (`ab?`) exists under the help under `a?`. 
This commits adds a `?` to solve that.

Also tweaks `help_msg_ab` a bit.

**Test plan**
Do we need that `[b]` before `[addr]`on that line? I'm not sure. If not, we can remove that.
https://github.com/officialcjunior/rizin/blob/0a14c7e53e4b707d49a28215ebca15ae86dc2040/librz/core/cmd_analysis.c#L14

See if there's anything else that could be improved. 

**Closing issues**

None

### DIFF
--- a/librz/core/cmd_analysis.c
+++ b/librz/core/cmd_analysis.c
@@ -12,7 +12,6 @@ static const char *help_msg_a[] = {
 	"aa", "[?]", "analyze all (fcns + bbs) (aa0 to avoid sub renaming)",
 	"a8", " [hexpairs]", "analyze bytes",
 	"ab", "[?] [addr]", "analyze block",
-	"abb", " [len]", "analyze N basic blocks in [len] (section.size by default)",
 	"ac", "[?]", "manage classes",
 	"aC", "[?]", "analyze function call",
 	"aCe", "[?]", "same as aC, but uses esil with abte to emulate the function",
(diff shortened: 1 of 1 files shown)


librz/core/cmd_analysis.c
```diff
@@ -11,7 +11,7 @@ static const char *help_msg_a[] = {
 	"a*", "", "same as afl*;ah*;ax*",
 	"aa", "[?]", "analyze all (fcns + bbs) (aa0 to avoid sub renaming)",
 	"a8", " [hexpairs]", "analyze bytes",
-	"ab", "[b] [addr]", "analyze block at given address",
+	"ab", "[?] [b] [addr]", "analyze block at given address",
```
ret2libc: no, we don't need the `b`. There is not only the `b`, but many others. So `[?]` is ok. With newshell it will be automatically printed by the code.

librz/core/cmd_analysis.c
```diff
@@ -11,7 +11,7 @@ static const char *help_msg_a[] = {
...
 	"abb", " [len]", "analyze N basic blocks in [len] (section.size by default)",
```
ret2libc: i believe this is already in the `ab` sub-menu. I'd say to remove this and just keep a nice description under `ab?`.

officialcjunior: Is edb0e63c3c910e8f9633508068a31ab4e7977b92 what you meant over here?
`ab` already had a proper description inside `ab?`

ret2libc: No i meant that if you look at `help_msg_ab` there is already `abb`, so I see no point in having `abb` listed also here. I think we can remove this and just keep the help under `help_msg_ab`.

officialcjunior: Oh yeah, didn't notice that, sorry.
Removed that.

ret2libc (approved):

PR #500, opened 2021-01-31
Add "initial view" test for Visual mode

kazarmy:

**Detailed description**

This pr adds a test for Visual mode, depicting the initial view after `V`.

**Test plan**

All builds are green.

**Closing issues**

...

### DIFF
--- a/test/db/archos/linux-x64/cmd_visual
+++ b/test/db/archos/linux-x64/cmd_visual
@@ -1,0 +1,30 @@
+NAME=initial Visual view
+FILE=bins/elf/ls
+CMDS=<<EOF
+e scr.interactive=true
+e scr.columns=80
+e scr.rows=12
+< q; V
+?e
+?e
+dr~^rip
+# TODO why is rip not at entry0?
+# TODO why can't I do "dr~^rip  # a comment here"?
+EOF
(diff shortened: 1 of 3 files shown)


XVilka:
Could you please add the same test but for Windows also?

kazarmy:
Okay e4a1bfec67f9ce1818f3bdadd1ded7e6109eabeb

ret2libc (approved):
Awesome, our first visual test? :D
BTW, is there any difference between the two tests? Can we move it under a common dir that is always executed or are there arch-specific things?

kazarmy:
> Awesome, our first visual test? :D

There were some tests under [db/cmd/cmd_visual](https://github.com/rizinorg/rizin/blob/7ce7fa6acc1042119902dfebd5bbf7bd60831284/test/db/cmd/cmd_visual) but they tested side effects after the `V` command, rather than the output of `V` itself.

> BTW, is there any difference between the two tests? Can we move it under a common dir that is always executed or are there arch-specific things?

The difference is an extraneous `^M` on the last line of the Linux `V` test output (it can be seen using `git diff --no-index`). It appears to be a bug but some spelunking through the source code is needed to fix it.

PR #501, opened 2021-01-31
Recursive help on newshell with json support

wargio:
# DO NOT SQUASH ME

**Detailed description**

Essentially this allows the support of `?* <cmd>` and `?*j <cmd>` on newshell.

Book reference here: https://github.com/rizinorg/book/pull/29

### DIFF
--- a/librz/core/cautocmpl.c
+++ b/librz/core/cautocmpl.c
@@ -89,10 +89,10 @@ err:
 	return NULL;
 }
 
-static bool do_autocmplt_cmdidentifier(RzCmd *cmd, const char *name, void *user) {
+static bool do_autocmplt_cmdidentifier(RzCmd *cmd, const RzCmdDesc *desc, void *user) {
 	struct autocmplt_cmdidentifier_t *u = (struct autocmplt_cmdidentifier_t *)user;
-	if (!strncmp(name, u->s, u->len)) {
-		rz_line_ns_completion_result_add(u->res, name);
+	if (!strncmp(desc->name, u->s, u->len)) {
+		rz_line_ns_completion_result_add(u->res, desc->name);
 	}
 	return true;
 }
@@ -103,7 +103,7 @@ static void autocmplt_cmdidentifier(RzCore *core, RzLineNSCompletionResult *res,
 		.s = s,
 		.len = len,
 	};
-	rz_cmd_foreach_cmdname(core->rcmd, do_autocmplt_cmdidentifier, &u);
+	rz_cmd_foreach_cmdname(core->rcmd, NULL, do_autocmplt_cmdidentifier, &u);
 }
 
 static void autocmplt_cmd_arg_file(RzLineNSCompletionResult *res, const char *s, size_t len) {
--- a/librz/core/cmd.c
+++ b/librz/core/cmd.c
@@ -4618,7 +4618,8 @@ DEFINE_HANDLE_TS_FCN_AND_SYMBOL(arged_command) {
 			if (!help_pra) {
 				goto err;
 			}
-			char *help_msg = rz_cmd_get_help(state->core->rcmd, help_pra, true);
+			bool use_color = state->core->print->flags & RZ_PRINT_FLAGS_COLOR;
+			char *help_msg = rz_cmd_get_help(state->core->rcmd, help_pra, use_color);
 			if (!help_msg) {
 				goto help_pra_err;
 			}
@@ -4760,55 +4761,116 @@ DEFINE_HANDLE_TS_FCN_AND_SYMBOL(redirect_command) {
 	return res;
 }
 
+typedef struct _search_help {
+	bool color;
+	RzStrBuf *sb;
+	PJ *pj;
+} RzHelpSearch;
+
+static bool help_search_cmd_desc_entry(RzCmd *cmd, const RzCmdDesc *cd, void *user) {
+	rz_return_val_if_fail(cd, false);
+	RzHelpSearch *hs = (RzHelpSearch *)user;
+	if (hs->pj) {
+		rz_cmd_get_help_json(cmd, cd, hs->pj);
+	} else {
+		rz_cmd_get_help_strbuf(cmd, cd, hs->color, hs->sb);
+	}
+	return true;
+}
+
+RZ_IPI RzCmdStatus rz_cmd_help_search_handler(RzCore *core, int argc, const char **argv, RzOutputMode mode) {
+	RzCmdStatus status = RZ_CMD_STATUS_OK;
+	RzCmdDesc *begin = NULL;
+
+	if (argc == 2) {
+		begin = rz_cmd_get_desc(core->rcmd, argv[1]);
+		if (!begin) {
+			status = RZ_CMD_STATUS_NONEXISTINGCMD;
+			goto exit_status;
+		}
+	}
+
+	RzHelpSearch hs = {
+		.color = core->print->flags & RZ_PRINT_FLAGS_COLOR,
+		.pj = NULL,
+		.sb = NULL,
+	};
+
+	if (mode & RZ_OUTPUT_MODE_JSON) {
+		hs.pj = pj_new();
+		if (!hs.pj) {
+			status = RZ_CMD_STATUS_ERROR;
+			goto exit_status;
+		}
+		pj_o(hs.pj);
+	} else {
+		hs.sb = rz_strbuf_new(NULL);
+		if (!hs.sb) {
+			status = RZ_CMD_STATUS_ERROR;
+			goto exit_status;
+		}
+	}
+
+	rz_cmd_foreach_cmdname(core->rcmd, begin, help_search_cmd_desc_entry, &hs);
+
+	if (mode & RZ_OUTPUT_MODE_JSON) {
+		pj_end(hs.pj);
+		rz_cons_printf("%s\n", pj_string(hs.pj));
+		pj_free(hs.pj);
+	} else {
+		char *help = rz_strbuf_drain(hs.sb);
+		rz_cons_printf("%s", help);
+		free(help);
+	}
+exit_status:
+	return status;
+}
+
 DEFINE_HANDLE_TS_FCN_AND_SYMBOL(help_command) {
 	size_t node_str_len = strlen(node_string);
 	if (node_str_len >= 2 && !strcmp(node_string + node_str_len - 2, "?*")) {
-		// TODO: get recursive help from RzCmdDesc
-		int detail = 0;
-		if (node_str_len > 3 && node_string[node_str_len - 3] == '?') {
-			detail++;
-			if (node_str_len > 4 && node_string[node_str_len - 4] == '?') {
-				detail++;
-			}
-		}
-		node_string[node_str_len - 2 - detail] = '\0';
-		recursive_help(state->core, detail, node_string);
-		return RZ_CMD_STATUS_OK;
-	} else {
-		TSNode command = ts_node_child_by_field_name(node, "command", strlen("command"));
-		char *command_str = ts_node_sub_string(command, state->input);
-		TSNode args = ts_node_child_by_field_name(node, "args", strlen("args"));
-		RzCmdParsedArgs *pr_args = NULL;
-		RzCmdStatus res = RZ_CMD_STATUS_INVALID;
-		if (!ts_node_is_null(args)) {
-			RzCmdDesc *cd = rz_cmd_get_desc(state->core->rcmd, command_str);
-			bool do_unwrap = cd && cd->type != RZ_CMD_DESC_TYPE_OLDINPUT;
-			pr_args = ts_node_handle_arg_prargs(state, node, args, 1, do_unwrap);
-			if (!pr_args) {
-				goto err_else;
-			}
-			rz_cmd_parsed_args_setcmd(pr_args, command_str);
-		} else {
-			pr_args = rz_cmd_parsed_args_newcmd(command_str);
-			if (!pr_args) {
-				goto err_else;
-			}
-		}
-
-		// let's try first with the new auto-generated help, if
-		// something fails fallback to old behaviour
-		char *help_msg = rz_cmd_get_help(state->core->rcmd, pr_args, state->core->print->flags & RZ_PRINT_FLAGS_COLOR);
-		if (help_msg) {
-			rz_cons_printf("%s", help_msg);
-			free(help_msg);
-			res = RZ_CMD_STATUS_OK;
-		}
-	err_else:
-		rz_cmd_parsed_args_free(pr_args);
-		free(command_str);
-		return res;
+		node_string[node_str_len - 2] = 0;
+		const char *argv[2] = { NULL, node_string };
+		return rz_cmd_help_search_handler(state->core, 2, argv, RZ_OUTPUT_MODE_STANDARD);
+	} else if (node_str_len >= 3 && !strcmp(node_string + node_str_len - 3, "?*j")) {
+		node_string[node_str_len - 3] = 0;
+		const char *argv[2] = { NULL, node_string };
+		return rz_cmd_help_search_handler(state->core, 2, argv, RZ_OUTPUT_MODE_JSON);
 	}
-	return RZ_CMD_STATUS_OK;
+
+	TSNode command = ts_node_child_by_field_name(node, "command", strlen("command"));
+	char *command_str = ts_node_sub_string(command, state->input);
+	TSNode args = ts_node_child_by_field_name(node, "args", strlen("args"));
+	RzCmdParsedArgs *pr_args = NULL;
+	RzCmdStatus res = RZ_CMD_STATUS_INVALID;
+	if (!ts_node_is_null(args)) {
+		RzCmdDesc *cd = rz_cmd_get_desc(state->core->rcmd, command_str);
+		bool do_unwrap = cd && cd->type != RZ_CMD_DESC_TYPE_OLDINPUT;
+		pr_args = ts_node_handle_arg_prargs(state, node, args, 1, do_unwrap);
+		if (!pr_args) {
+			goto err_else;
+		}
+		rz_cmd_parsed_args_setcmd(pr_args, command_str);
+	} else {
+		pr_args = rz_cmd_parsed_args_newcmd(command_str);
+		if (!pr_args) {
+			goto err_else;
+		}
+	}
+
+	// let's try first with the new auto-generated help, if
+	// something fails fallback to old behaviour
+	bool use_color = state->core->print->flags & RZ_PRINT_FLAGS_COLOR;
+	char *help_msg = rz_cmd_get_help(state->core->rcmd, pr_args, use_color);
+	if (help_msg) {
+		rz_cons_printf("%s", help_msg);
+		free(help_msg);
+		res = RZ_CMD_STATUS_OK;
+	}
+err_else:
+	rz_cmd_parsed_args_free(pr_args);
+	free(command_str);
+	return res;
 }
 
 DEFINE_HANDLE_TS_FCN_AND_SYMBOL(tmp_seek_command) {
--- a/librz/core/cmd_api.c
+++ b/librz/core/cmd_api.c
@@ -642,7 +642,7 @@ static size_t strbuf_append_calc(RzStrBuf *sb, const char *s) {
 	return strlen(s);
 }
 
-static void fill_modes_children_chars(RzStrBuf *sb, RzCmdDesc *cd) {
+static void fill_modes_children_chars(RzStrBuf *sb, const RzCmdDesc *cd) {
 	// RZ_CMD_DESC_TYPE_ARGV_MODES does not have actual children for
 	// the output modes, so we consider it separately
 	size_t i;
@@ -653,7 +653,7 @@ static void fill_modes_children_chars(RzStrBuf *sb, RzCmdDesc *cd) {
 	}
 }
 
-static size_t fill_children_chars(RzStrBuf *sb, RzCmdDesc *cd) {
+static size_t fill_children_chars(RzStrBuf *sb, const RzCmdDesc *cd) {
 	if (cd->help->options) {
 		return strbuf_append_calc(sb, cd->help->options);
 	}
@@ -663,7 +663,7 @@ static size_t fill_children_chars(RzStrBuf *sb, RzCmdDesc *cd) {
 
 	void **it;
 	bool has_other_commands = false;
-	RzCmdDesc *exec_cd = rz_cmd_desc_get_exec(cd);
+	const RzCmdDesc *exec_cd = rz_cmd_desc_get_exec((RzCmdDesc *)cd);
 	if (exec_cd) {
 		switch (exec_cd->type) {
 		case RZ_CMD_DESC_TYPE_ARGV_MODES:
@@ -706,7 +706,7 @@ static size_t fill_children_chars(RzStrBuf *sb, RzCmdDesc *cd) {
 	return res;
 }
 
-static bool show_children_shortcut(RzCmdDesc *cd) {
+static bool show_children_shortcut(const RzCmdDesc *cd) {
 	return cd->n_children || cd->help->options || cd->type == RZ_CMD_DESC_TYPE_OLDINPUT || cd->type == RZ_CMD_DESC_TYPE_ARGV_MODES;
 }
 
@@ -739,7 +739,7 @@ static void fill_wrapped_comment(RzCmd *cmd, RzStrBuf *sb, const char *comment, 
 	}
 }
 
-static size_t fill_args(RzStrBuf *sb, RzCmdDesc *cd) {
+static size_t fill_args(RzStrBuf *sb, const RzCmdDesc *cd) {
 	const RzCmdDescArg *arg;
 	size_t n_optionals = 0;
 	size_t len = 0;
@@ -836,11 +836,11 @@ static void fill_usage_strbuf(RzCmd *cmd, RzStrBuf *sb, RzCmdDesc *cd, bool use_
 	rz_strbuf_append(sb, "\n");
 }
 
-static size_t calc_padding_len(RzCmdDesc *cd, const char *name) {
+static size_t calc_padding_len(const RzCmdDesc *cd, const char *name, bool show_children) {
 	size_t name_len = strlen(name);
 	size_t args_len = 0;
 	size_t children_length = 0;
-	if (show_children_shortcut(cd)) {
+	if (show_children && show_children_shortcut(cd)) {
 		RzStrBuf sb;
 		rz_strbuf_init(&sb);
 		fill_children_chars(&sb, cd);
@@ -859,14 +859,14 @@ static size_t calc_padding_len(RzCmdDesc *cd, const char *name) {
 	return name_len + args_len + children_length;
 }
 
-static void update_minmax_len(RzCmdDesc *cd, size_t *max_len, size_t *min_len) {
-	size_t val = calc_padding_len(cd, cd->name);
+static void update_minmax_len(RzCmdDesc *cd, size_t *max_len, size_t *min_len, bool show_children) {
+	size_t val = calc_padding_len(cd, cd->name, show_children);
 	*max_len = val > *max_len ? val : *max_len;
 	*min_len = val < *min_len ? val : *min_len;
 }
 
-static void do_print_child_help(RzCmd *cmd, RzStrBuf *sb, RzCmdDesc *cd, const char *name, const char *summary, bool show_children, size_t max_len, bool use_color) {
-	size_t str_len = calc_padding_len(cd, name);
+static void do_print_child_help(RzCmd *cmd, RzStrBuf *sb, const RzCmdDesc *cd, const char *name, const char *summary, bool show_children, size_t max_len, bool use_color) {
+	size_t str_len = calc_padding_len(cd, name, show_children);
 	int padding = str_len < max_len ? max_len - str_len : 0;
 	const char *pal_args_color = "",
 		   *pal_opt_color = "",
@@ -918,7 +918,7 @@ static char *group_get_help(RzCmd *cmd, RzCmdDesc *cd, bool use_color) {
 
 	rz_cmd_desc_children_foreach(cd, it_cd) {
 		RzCmdDesc *child = *(RzCmdDesc **)it_cd;
-		update_minmax_len(child, &max_len, &min_len);
+		update_minmax_len(child, &max_len, &min_len, true);
 	}
 	if (max_len - min_len > MAX_RIGHT_ALIGHNMENT) {
 		max_len = min_len + MAX_RIGHT_ALIGHNMENT;
@@ -936,7 +936,7 @@ static char *argv_modes_get_help(RzCmd *cmd, RzCmdDesc *cd, bool use_color) {
 	fill_usage_strbuf(cmd, sb, cd, use_color);
 
 	size_t max_len = 0, min_len = SIZE_MAX;
-	update_minmax_len(cd, &max_len, &min_len);
+	update_minmax_len(cd, &max_len, &min_len, true);
 	max_len++; // consider the suffix letter
 	if (max_len - min_len > MAX_RIGHT_ALIGHNMENT) {
 		max_len = min_len + MAX_RIGHT_ALIGHNMENT;
@@ -1022,17 +1022,18 @@ static char *argv_get_help(RzCmd *cmd, RzCmdDesc *cd, size_t detail, bool use_co
 
 	switch (detail) {
 	case 1:
-		return rz_strbuf_drain(sb);
+		break;
 	case 2:
 		if (cd->help->description) {
 			rz_strbuf_appendf(sb, "\n%s\n", cd->help->description);
 		}
 		fill_details(cmd, cd, sb, use_color);
-		return rz_strbuf_drain(sb);
+		break;
 	default:
 		rz_strbuf_free(sb);
 		return NULL;
 	}
+	return rz_strbuf_drain(sb);
 }
 
 static char *fake_get_help(RzCmd *cmd, RzCmdDesc *cd, bool use_color) {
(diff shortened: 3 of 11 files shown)


librz/core/cmd_api.c
```diff
@@ -1684,8 +1765,7 @@ RZ_API char *rz_cmd_parsed_args_execstr(RzCmdParsedArgs *a) {
 }
 
 RZ_API const char *rz_cmd_parsed_args_cmd(RzCmdParsedArgs *a) {
-	rz_return_val_if_fail(a && a->argv && a->argv[0], NULL);
-	return a->argv[0];
+	return a && a->argv && a->argv[0] ? a->argv[0] : NULL;
```
ret2libc: why this change?

librz/core/cmd_api.c
```diff
@@ -1089,6 +1089,87 @@ static char *get_help(RzCmd *cmd, RzCmdDesc *cd, RzCmdParsedArgs *args, bool use
 	return NULL;
 }
 
+
+static bool child_help_json(RzCmd *cmd, RzCmdDesc *cd, PJ* j) {
```
ret2libc: I think when printing in JSON it makes sense to not only return the `args_str`, but also an array of `RzCmdDescArg`, with all the details. It can be much more useful to callers than the string itself.

librz/core/cmd.c
```diff
@@ -4780,20 +4780,102 @@ DEFINE_HANDLE_TS_FCN_AND_SYMBOL(redirect_command) {
...
+help_rec_err:
+		free (help_msg);
+		rz_cmd_parsed_args_free(help_pra);
+		free (cmdname_help);
+	}
+	return true;
+}
+
+
+static void foreach_cmd_search(RzCmd *cmd, RzCmdDesc *cd, PJ* pj) {
```
ret2libc: IMO you can more easily get the same info by adapting a bit `rz_cmd_foreach_cmdname`. Right now it accepts a callback that gets `RzCmd *, char *name, void *user`. I think you can extend it to get `RzCmd *, RzCmdDesc *, char *name, void *user` and use it here instead of re-implementing all the iteration of the various commands.

librz/core/cmd_api.c
```diff
@@ -1089,6 +1089,87 @@ static char *get_help(RzCmd *cmd, RzCmdDesc *cd, RzCmdParsedArgs *args, bool use
...
+		}
+		return group_get_help_json(cmd, cd, j);
+	case RZ_CMD_DESC_TYPE_ARGV:
+		return child_help_json(cmd, cd, j);
+	case RZ_CMD_DESC_TYPE_ARGV_MODES:
+		return argv_modes_get_help_json(cmd, cd, j);
+	case RZ_CMD_DESC_TYPE_FAKE:
+		return child_help_json(cmd, cd, j);
+	case RZ_CMD_DESC_TYPE_OLDINPUT:
+		return false;
```
ret2libc: As said, I believe it's better if you re-use `rz_cmd_foreach_cmdname`, because it is already tested and it also kinda works with OLDINPUT types (at least it gets the first level of them, it doesn't go into the inner commands because there is no good way).

librz/core/cmd_api.c
```diff
@@ -1089,6 +1089,87 @@ static char *get_help(RzCmd *cmd, RzCmdDesc *cd, RzCmdParsedArgs *args, bool use
...
+		pj_ks(j, "args", args);
+		free (args);
+	}
+	pj_ks(j, "desc", cd->help->summary ? cd->help->summary : "");
+	pj_end(j);
+
+	return true;
+}
+
+static bool argv_modes_get_help_json(RzCmd *cmd, RzCmdDesc *cd, PJ *j) {
```
ret2libc: Probably `argv_modes` and `argv` cmddesc printing can be refactored a lot to be almost equal.

librz/core/cmd_api.c
```diff
@@ -1089,6 +1089,87 @@ static char *get_help(RzCmd *cmd, RzCmdDesc *cd, RzCmdParsedArgs *args, bool use
...
+			if (cd->help->args_str) {
+				pj_ks(j, "args", cd->help->args_str);
+			} else {
+				RzStrBuf *sb = rz_strbuf_new(NULL);
+				fill_args(sb, cd);
+				char* args = rz_strbuf_drain(sb);
+				pj_ks(j, "args", args);
+				free (args);
+			}
+			pj_ks(j, "desc", summary);
```
ret2libc: this should be `summary`. Also, you should add `description` field for the actualy description field.

ret2libc (commented):
Please add a unit tests for `rz_cmd_get_help_json` and an integration test for `?*`.

librz/core/cmd_api.c
wargio: rz_return_val_if_fail is a macro that can be skipped if the right flags are defined (meant for debugging).

ret2libc: I know, indeed I assume that the condition in `rz_return_val_if_fail` is always true. Indeed `RzCmdParsedArgs` is created through `rz_cmd_parsed_args_new`, which always allocates `a->argv` and always puts a value in `a->argv[0]`. So I think the condition should stay how it was before and it can never be false. If it is, there's some error somewhere else.

librz/core/cmd_api.c
wargio: but this is only for JSON. i will use that if the output is standard.

librz/core/cmd.c
wargio: the issue with `rz_cmd_foreach_cmdname` is regarding the supported modes of each commands which is handled by `argv_modes_get_help`

ret2libc: Can we change `argv_modes_get_help_json`? I think it is not doing anything particular apart from computing the name+suffix and summary+suffix. The arguments should be the same no matter what the mode is.

Maybe `rz_cmd_foreach_cmdname` can temporarily compute name and suffix as necessary and replace `cd->name`/`cd->summary` with the changed one before passing the reference to the callback and then restore the original values after the callback has finished.

As said, ideally I would like to have just one mode to iterate over the command descriptor and since this seems very very similar to what is already in `rz_cmd_foreach_cmdname` it's IMO better to keep everything there and just provide the right callback.

wargio: i'm thinking to change the output to have always `"modes"` and have the modes as `["default","json", "table"]` etc..

ret2libc: The problem there is that right now `rz_cmd_foreach_cmdname` lists those as separate commands, because actually the user sees them as separate (z, zj, z*, zt, zl, etc.), so I think whatever API we use it should still show them separately (as said, because the names are actually different from a user point of view it's as they are separate.

test/unit/test_cmd.c
```diff
@@ -619,7 +620,7 @@ bool test_foreach_cmdname(void) {
 	rz_cmd_desc_argv_new(cmd, v_inner_cd, "v2", zd_handler, &fake_help);
 
 	RzList *res = rz_list_newf(free);
-	rz_cmd_foreach_cmdname(cmd, foreach_cmdname_cb, res);
+	rz_cmd_foreach_cmdname(cmd, NULL, foreach_cmdname_cb, res);
```
ret2libc, line 622: please add a test with non-NULL `begin` as well

test/db/cmd/cmd_help
```diff
@@ -242,4 +242,36 @@ EOF
...
\ No newline at end of file
+RUN
+
+NAME=recursive help
+FILE=-
+CMDS=<<EOF
+?* zi
+EOF
+EXPECT=<<EOF
+Usage: [0mzi[0m[j*qi][0m   # Show zignatures matching information[0m
```
ret2libc: isn't `scr.color` usually disabled when running tests?

ret2libc: I think the `Usage` line should not be present because all commands are already listed anyway and it makes the output a bit messy IMO.

librz/core/cmd_descs.c
```diff
@@ -598,6 +599,20 @@ static const RzCmdDescHelp remote_rap_bg_help = {
 	.args = remote_rap_bg_args,
 };
 
+static const RzCmdDescArg cmd_help_search_args[] = {
+	{
+		.name = "search_cmd",
+		.type = RZ_CMD_ARG_TYPE_STRING,
+		.optional = true,
+
```
ret2libc, line 607: was this created by the generator script? This empty line seems weird.

librz/core/cmd.c
```diff
@@ -4780,20 +4780,86 @@ DEFINE_HANDLE_TS_FCN_AND_SYMBOL(redirect_command) {
...
-		if (node_str_len > 3 && node_string[node_str_len - 3] == '?') {
-			detail++;
-			if (node_str_len > 4 && node_string[node_str_len - 4] == '?') {
-				detail++;
-			}
-		}
-		node_string[node_str_len - 2 - detail] = '\0';
-		recursive_help(state->core, detail, node_string);
-		return RZ_CMD_STATUS_OK;
+		// this needs to be properly handled.
```
ret2libc: what does it mean?

librz/core/cmd.c
```diff
@@ -4780,20 +4780,86 @@ DEFINE_HANDLE_TS_FCN_AND_SYMBOL(redirect_command) {
...
+	return true;
+}
+
+RZ_IPI RzCmdStatus rz_cmd_help_search_handler(RzCore *core, int argc, const char **argv, RzOutputMode mode) {
+	bool is_interactive = rz_cons_is_interactive(); 
+	rz_cons_set_interactive(false); 
+	PJ* pj = NULL;
+	RzCmdStatus status = RZ_CMD_STATUS_OK;
+
+	RzCmdDesc *begin = argc == 2 ? rz_cmd_get_desc(core->rcmd, argv[1]) : NULL;
```
ret2libc: avoid the ternary and just use a `if (argc == 2) ` where you set `begin = rz_cmd_get_desc` and you check `begin`.

librz/core/cmd.c
```diff
@@ -4780,20 +4780,86 @@ DEFINE_HANDLE_TS_FCN_AND_SYMBOL(redirect_command) {
...
+		free (help_msg);
+		rz_cmd_parsed_args_free(help_pra);
+		free (cmdname_help);
+	}
+	return true;
+}
+
+static bool help_search_cmd_desc_entry(RzCmd *cmd, RzCmdDesc* cd, const char *name, void *user) {
+	rz_return_val_if_fail(cd, false);
+	if (cd->type != RZ_CMD_DESC_TYPE_ARGV &&
```
ret2libc: I think we can handle all types that `rz_cmd_foreach_cmdname` can return, no need to filter them. We can handle OLDINPUT as well i think

librz/core/cmd_api.c
```diff
@@ -1890,14 +1976,14 @@ static void cmd_foreach_cmdname(RzCmd *cmd, RzCmdDesc *cd, RzCmdForeachNameCb cb
...
-			cb(cmd, cd->name, user);
+			cb(cmd, cd, cd->name, user);
 		}
 		break;
 	case RZ_CMD_DESC_TYPE_ARGV_MODES:
 		for (i = 0; i < RZ_ARRAY_SIZE(argv_modes); i++) {
 			if (cd->d.argv_modes_data.modes & argv_modes[i].mode) {
 				char *name = rz_str_newf("%s%s", cd->name, argv_modes[i].suffix);
-				cb(cmd, name, user);
+				cb(cmd, cd, name, user);
```
ret2libc: temporarily swap `cd->name = name` and `cd->summary = summary + summary_suffix`. By doing this here the method `rz_cmd_get_help_json` can be much simpler, in particular with regards to printing the `summary` JSON field.

librz/core/cmd_api.c
```diff
@@ -1089,6 +1089,92 @@ static char *get_help(RzCmd *cmd, RzCmdDesc *cd, RzCmdParsedArgs *args, bool use
...
+		pj_kb(j, "is_option", arg->flags & RZ_CMD_ARG_FLAG_OPTION);
+		if (arg->default_value) {
+			pj_kb(j, "default", arg->default_value);
+		}
+		pj_end(j);
+	}
+	pj_end(j);
+}
+
+RZ_API bool rz_cmd_get_help_json(RzCmd *cmd, RzCmdDesc *cd, const char* name, PJ *j) {
```
ret2libc: add the `description` field.

librz/core/cmd_api.c
```diff
@@ -1089,6 +1089,92 @@ static char *get_help(RzCmd *cmd, RzCmdDesc *cd, RzCmdParsedArgs *args, bool use
...
+	pj_a(j);
+	const char* argtype = NULL;
+	for (arg = cd->help->args; arg && arg->name; arg++) {
+		if (has_array) {
+			rz_warn_if_reached();
+			break;
+		}
+		pj_o(j);
+#define CASE_TYPE(x,y) case (x): \
+						argtype = (y); \
```
ret2libc: strange indentation?

librz/core/cmd_api.c
```diff
@@ -1089,6 +1089,92 @@ static char *get_help(RzCmd *cmd, RzCmdDesc *cd, RzCmdParsedArgs *args, bool use
...
+		}
+		pj_ks(j, "type", argtype);
+		pj_ks(j, "name", arg->name);
+		if (arg->type == RZ_CMD_ARG_TYPE_FAKE) {
+			pj_end(j);
+			continue;
+		}
+		pj_kb(j, "nospace", arg->no_space);
+		pj_kb(j, "required", !arg->optional);
+		pj_kb(j, "is_array", arg->flags & RZ_CMD_ARG_FLAG_ARRAY);
```
ret2libc: add `is_last`

librz/core/cmd_api.c
```diff
@@ -1089,6 +1089,92 @@ static char *get_help(RzCmd *cmd, RzCmdDesc *cd, RzCmdParsedArgs *args, bool use
...
+		pj_kb(j, "is_option", arg->flags & RZ_CMD_ARG_FLAG_OPTION);
+		if (arg->default_value) {
+			pj_kb(j, "default", arg->default_value);
+		}
+		pj_end(j);
```
ret2libc, line 1161: if arg is `choice`  add `choices` field as well.

librz/core/cmd.c
```diff
@@ -4780,20 +4780,86 @@ DEFINE_HANDLE_TS_FCN_AND_SYMBOL(redirect_command) {
 	return res;
 }
 
+static bool do_help_manual(RzCmd *cmd, const char* cmdname) {
```
ret2libc: check style `const char *cmdname`

librz/core/cmd.c
```diff
@@ -4780,20 +4780,86 @@ DEFINE_HANDLE_TS_FCN_AND_SYMBOL(redirect_command) {
...
+	char *cmdname_help = NULL;
+	char *help_msg = NULL;
+	RzCmdParsedArgs *help_pra = NULL;
+	if (cmdname) {
```
ret2libc: can `cmdname` really be NULL? If it can (i'm not sure), i'd prefer a `if (!cmdname)` and an early return rather than having everything indented.

librz/core/cmd.c
```diff
@@ -4780,20 +4780,86 @@ DEFINE_HANDLE_TS_FCN_AND_SYMBOL(redirect_command) {
...
+		cmdname_help = rz_str_newf("%s?", cmdname);
+		if (!cmdname_help) {
+			goto help_rec_err;
+		}
+		help_pra = rz_cmd_parsed_args_newcmd(cmdname_help);
+		if (!help_pra) {
+			goto help_rec_err;
+		}
+		help_msg = rz_cmd_get_help(cmd, help_pra, true);
```
ret2libc: here we would need the output of `print_child_help`, but it's not simple...

ret2libc (commented):
Some comments! Thanks ;)

test/db/cmd/cmd_help
wargio: yes; but unless that is a bug, that is the output :/

librz/core/cmd_descs.c
wargio: yes.

librz/core/cmd.c
wargio: `?` should be refactored to be a newshell command (and probably moved under a letter instead of using the question mark).

librz/core/cmd_api.c
wargio: can you tell me an example of command with that?

librz/core/cmd.c
wargio: no we can't.

librz/core/cmd_api.c
wargio: probably, but if i need to do that, then i prefer to add a method to generate a RzCmdDesc which outputs the correct information.

ITAYC0HEN:
Can we also use the `help [keyword]` as a command that searches in all help messages. I think it will be a significant addition to the user experience as this is an obvious command.


`help disas`, `help imports`, `help qword`, ... in addition to `?*~disas`, `?*~imports`, `?*~qword`

test/db/cmd/cmd_help
ret2libc: I think it's because you called rz_cmd_get_help with true instead of passing the value of scr.color

librz/core/cmd_api.c
ret2libc: What do you mean?

librz/core/cmd.c
ret2libc: this has nothing to do with the `?`. This piece of code is about handling things like `z?*`, but currently it does not seem to work correctly because it prints all commands and not just the `z` sub-commands.

librz/core/cmd_api.c
ret2libc: Grep in `cmd_descs.yaml` for `choices` and you'll find `sn`(seek next)  or `za`

ret2libc:
> Can we also use the `help [keyword]` as a command that searches in all help messages. I think it will be a significant addition to the user experience as this is an obvious command.
> 
> `help disas`, `help imports`, `help qword`, ... in addition to `?*~disas`, `?*~imports`, `?*~qword`

I think we can, yes, but I would keep this for a separate PR as it can be implemented on its own later.

ret2libc:
IMO if you do `?* zs` or `zs?*` you should get something like:
```
| zs [<zignspace>]     # Display/select zignspaces
| zsj [<zignspace>]    # Display/select zignspaces (JSON mode)
| zs* [<zignspace>]    # Display/select zignspaces (rizin mode)
| zs- [<zignspace>]    # Pop last zignspace or delete <zignspace>
| zs+ <zignspace>      # Push previous zignspace and set
| zsr <newname>        # Rename selected zignspace
```
Notice:
- the aligned summary on the right (as it is already done for regular sub-commands)
- the expanded modes (so if you want to filter for only commands that support e.g. JSON you can)
- no `Usage` line, just the summary line, like what is printed by `print_child_help`

test/db/cmd/cmd_help
wargio: oh, maybe.

librz/core/cmd.c
wargio: nono. that piece of code is the one that is handled by "?" OLDCMD

ret2libc: No :/ you can see from the if condition that it does `node_str_len >= 2 && !strcmp(node_string + node_str_len - 2, "?*")) {`. Also, you can see from the old code that it was calling `recursive_help`. So I'm quite sure this code is about `<cmd>?*`.

librz/core/cmd_api.c
wargio: i can't do this. it's too messy due `const` being in the structure.

ret2libc: Ah, `cd->help` is const :(

librz/core/cmd_api.c
ret2libc: Add doxygen doc to new APIs.

librz/include/rz_cmd.h
```diff
@@ -431,7 +431,7 @@ typedef struct rz_core_plugin_t {
 	RzCmdCb fini;
 } RzCorePlugin;
 
-typedef bool (*RzCmdForeachNameCb)(RzCmd *cmd, const char *name, void *user);
+typedef bool (*RzCmdForeachNameCb)(RzCmd *cmd, RzCmdDesc* desc, const char *name, void *user);
```
ret2libc:
```suggestion
typedef bool (*RzCmdForeachNameCb)(RzCmd *cmd, const RzCmdDesc* desc, const char *name, void *user);
```
So we can be sure callers won't change desc.

librz/core/cmd_api.c
ret2libc: The as you said, probably having a method to "clone" an existing `cd` is the right way. You can probably just allocate a RzCmdDesc and copy all-fields except for help->summary and cd->name. I think that's the "cleaner" way so the caller doesn't have to deal with summary/names itself.

librz/core/cmd.c
wargio: no. it's the code that is for eval numbers.

librz/core/cmd_api.c
wargio: yes, but that is heavier than me doing that check

librz/core/cmd.c
```diff
@@ -4760,55 +4764,131 @@ DEFINE_HANDLE_TS_FCN_AND_SYMBOL(redirect_command) {
...
+		pj_free(pj);
+	}
+exit_status:
+	rz_cons_set_interactive(is_interactive); 
+	return status;
+}
+
 DEFINE_HANDLE_TS_FCN_AND_SYMBOL(help_command) {
 	size_t node_str_len = strlen(node_string);
-	if (node_str_len >= 2 && !strcmp(node_string + node_str_len - 2, "?*")) {
```
ret2libc, line 4765: this can still be used to suppport the format `<cmd>?*`

wargio: no. that needs to go to `argv_modes` to be supported and tbh, i do not think is a good thing.

i highly prefer using `?* <prefix>`

ret2libc: Do support this for now. Because when you do `pd?*` the tree-sitter grammar will recognize a `help_command` and it will land here. So for now add it here, it should not be too hard as you already have all the methods.

librz/core/cmd_api.c
ret2libc: performances don't matter here. It's just a help stuff, it can be "slow" (and by slow i mean no-one will ever notice the slowness). The problem is that if you don't do this here, *every* caller of this function has to be sure to handle the things properly, which is super error prone.

Actually, the foreach function can just provide a `const RzCmdDesc *`, instead of `RzCmdDesc + name` , with the right name and summary already prefilled. This way the callers know what they are iterating without having to rely on *internal details*.

librz/core/cmd_api.c
```diff
@@ -1089,7 +1099,137 @@ static char *get_help(RzCmd *cmd, RzCmdDesc *cd, RzCmdParsedArgs *args, bool use
...
+				pj_s(j, *choice);
+			}
+			pj_end(j);
+		}
+		pj_end(j);
+	}
+	pj_end(j);
+}
+
+/**
```
ret2libc, line 1166: ??

librz/core/cmd_api.c
```diff
@@ -1089,7 +1099,137 @@ static char *get_help(RzCmd *cmd, RzCmdDesc *cd, RzCmdParsedArgs *args, bool use
...
+ * \param cmd reference to the RzCmd
+ * \param parent Parent command descriptor of the command being added
+ * \param name Base name of the command. New commands will be created with the proper suffix based on the supported \p modes
+ * \param modes Modes supported by the handler (see RzOutputMode). They can be put in OR to support multiple modes
+ * \param cb Callback that actually executes the command
+ * \param help Help structure used to describe the command when using `?` and `??`
+ */
+
+/**
+ * \brief generates a json output of the given help message description
```
ret2libc:
```suggestion
 * \brief Generates a JSON output of the given help message description
```

librz/core/cmd.c
wargio: yes true.

librz/core/cmd_api.c
wargio: it's the doxygen grammar.

ret2libc: Yes, i meant that there's a documentation more than necessary.

librz/core/cmd.c
```diff
@@ -4760,55 +4764,130 @@ DEFINE_HANDLE_TS_FCN_AND_SYMBOL(redirect_command) {
 	return res;
 }
 
+typedef struct _search_help {
+	bool color;
+	PJ *pj;
+} RzHelpSearch;
+
+static bool do_help_manual(RzCmd *cmd, const char *cmdname, bool color) {
```
ret2libc: this is not "exploding" the argv modes like it is done in json mode.

ret2libc (commented):
For some unknown reasons, if I do `?* zd` then I can't use the arrows anymore to see the commands history

librz/core/cmd.c
```diff
@@ -4760,55 +4764,130 @@ DEFINE_HANDLE_TS_FCN_AND_SYMBOL(redirect_command) {
...
+	cmdname_help = rz_str_newf("%s?", cmdname);
+	if (!cmdname_help) {
+		goto help_rec_err;
+	}
+	help_pra = rz_cmd_parsed_args_newcmd(cmdname_help);
+	if (!help_pra) {
+		goto help_rec_err;
+	}
+	ut32 pflags = color ? RZ_CMD_HELP_PRINT_USE_COLORS : 0;
+	help_msg = rz_cmd_get_help(cmd, help_pra, pflags);
```
ret2libc: The logic you are using here is wrong because it just prints commands when you ask for the help of the parent command (for example, if you do `?* zsr` you won't see anything because nobody is going to print its summary. When you do `?* zs` you see `zsr` just because `zs` is printing its children).

I think you should make have a way to ask for just the summary line, like:
```
rz_cmd_get_help(cmd, help_pra, RZ_CMD_HELP_PRINT_SHOW_SUMMARY)
```
assuming help_pra.cmd == 'zsr`:
```
| zsr <newname>        # Rename selected zignspace
```

librz/core/cmd.c
```diff
@@ -4760,55 +4764,130 @@ DEFINE_HANDLE_TS_FCN_AND_SYMBOL(redirect_command) {
...
+
+static bool do_help_manual(RzCmd *cmd, const char *cmdname, bool color) {
+	if (!cmdname) {
+		return false;
+	}
+	char *cmdname_help = NULL;
+	char *help_msg = NULL;
+	RzCmdParsedArgs *help_pra = NULL;
+
+	cmdname_help = rz_str_newf("%s?", cmdname);
```
ret2libc: maybe we can think about a new API `rz_cmd_desc_get_help` that directly takes a RzCmdDesc instead of a RzCmdParsedArgs.

wargio:
regarding `?* zd` there is a bug in `rz_cons`. if i remove the `rz_cons_printf` with an `eprintf` it works.

librz/core/cmd_api.c
```diff
@@ -1879,34 +2002,57 @@ RZ_API const RzCmdDescArg *rz_cmd_desc_get_arg(RzCmd *cmd, const RzCmdDesc *cd,
...
+}
+
 static void cmd_foreach_cmdname(RzCmd *cmd, RzCmdDesc *cd, RzCmdForeachNameCb cb, void *user) {
 	if (!cd) {
 		return;
 	}
 
 	void **it_cd;
 	size_t i;
+	char *name;
```
ret2libc: Just define these vars inside the `if (cd->d.argv_modes_data.modes & argv_modes[i].mode) {` as they are used only there.

wargio:
image

librz/core/cautocmpl.c
```diff
@@ -89,10 +89,11 @@ static struct guess_data_t *guess_next_autocmplt_token(RzCore *core, RzLineBuffe
 	return NULL;
 }
 
-static bool do_autocmplt_cmdidentifier(RzCmd *cmd, const char *name, void *user) {
+static bool do_autocmplt_cmdidentifier(RzCmd *cmd, const RzCmdDesc *desc, void *user) {
+	(void)desc;
```
ret2libc: Remove

librz/core/cmd.c
```diff
@@ -4760,55 +4761,123 @@ DEFINE_HANDLE_TS_FCN_AND_SYMBOL(redirect_command) {
...
 
+typedef struct _search_help {
+	bool color;
+	RzStrBuf *sb;
+	PJ *pj;
+} RzHelpSearch;
+
+static bool help_search_cmd_desc_entry(RzCmd *cmd, const RzCmdDesc *cd, void *user) {
+	rz_return_val_if_fail(cd, false);
+	if (cd->type != RZ_CMD_DESC_TYPE_ARGV &&
```
ret2libc: This whole check can be easily removed and everything still works. It has the advantage that you also get the summary of oldinput commands, though you can't ofc descend into them.

librz/core/cmd.c
```diff
@@ -4760,55 +4761,123 @@ DEFINE_HANDLE_TS_FCN_AND_SYMBOL(redirect_command) {
...
+		rz_cmd_get_help_json(cmd, cd, hs->pj);
+	} else {
+		rz_cmd_get_help_strbuf(cmd, cd, hs->color, hs->sb);
+	}
+	return true;
+}
+
+RZ_IPI RzCmdStatus rz_cmd_help_search_handler(RzCore *core, int argc, const char **argv, RzOutputMode mode) {
+	bool is_interactive = rz_cons_is_interactive();
+	rz_cons_set_interactive(false);
```
ret2libc: why this? is it to avoid the "Do you want to print X lines?"? I think it still makes sense to have that warning.

librz/core/cmd.c
ret2libc: However, I would add a a "type" field in the json output to export the fact that some commands are OLDINPUT, others are ARGV and others ARGV_MODES.

librz/core/cmd_api.c
```diff
@@ -1089,6 +1090,128 @@ static char *get_help(RzCmd *cmd, RzCmdDesc *cd, RzCmdParsedArgs *args, bool use
...
+	}
+	fill_args_json(cd, j);
+	pj_ks(j, "description", cd->help->description ? cd->help->description : "");
+	pj_ks(j, "summary", cd->help->summary ? cd->help->summary : "");
+	pj_end(j);
+	return true;
+}
+
+/**
+ * \brief Generates a text output of the given help message description
```
ret2libc: AFAIK this only generate the "summary view", so maybe it's worth explaining that a bit more here. Something like: `it only prints a summary of the command, with its name, arguments and the summary of the command`

ret2libc (commented):
Awesome!! Well done ;)

I added some minor stuff, but I think overall it is in good shape to be merged very soon.

One thing to consider at some point: right now stuff like `~?`, `|?`, `>?`, `@?`, etc. are not printed, so one cannot find them by using `?*`. The technical reason is that `rz_cmd_foreach_cmdname` only iterates over the "callable" commands, and of course `|` & co. are not commands but are rather tokens of the "statements" rizin-shell accepts. Now... not sure, do you think it makes sense to show them within `?*` anyway? @wargio @XVilka @thestr4ng3r  @ITAYC0HEN WDYT? other ideas?

librz/core/cmd.c
wargio: i would not do this. the only reason i'm doing something like this, is to not handle OLDINPUT and the fake one.

librz/core/cmd.c
wargio: Because before i was calling `<cmd>?` and some like `V` didn't like it and started the visual mode

librz/core/cautocmpl.c
wargio: i forgot that.

librz/core/cmd.c
ret2libc: Is it still necessary?

librz/core/cmd.c
ret2libc: Fake ones cannot arrive at this point because the foreach method discards them.
For oldinput ones, you don't need to handle anything in particular and it will provide slightly better results.

By removing this check:
```
[0x00000000]> w?*
| w <string>           # Write string
| wB <value>           # Set bits with given value
| wB- <value>          # Unset bits with given value
| wv <value>           # Write value as 4-bytes/8-bytes based on value
| wv1 <value>          # Write value of 1 byte
| wv2 <value>          # Write value of 2 byte
| wv4 <value>          # Write value of 4 byte
| wv8 <value>          # Write value of 8 byte
| w0 <len>             # Write <len> bytes with value 0x00
| w1+ [<n>]            # Increment a byte
| w1- [<n>]            # Decrement a byte
| w2+ [<n>]            # Increment a word
| w2- [<n>]            # Decrement a word
| w4+ [<n>]            # Increment a dword
| w4- [<n>]            # Decrement a dword
| w8+ [<n>]            # Increment a qword
| w8- [<n>]            # Decrement a qword
| w6d <base64>         # Write the base64-decoded bytes
| w6e <hexstring>      # Write the base64-encoded bytes
| wh                # whereis/which shell command
| we                # Extend write operations (insert bytes instead of replacing)
| wu                # Apply unified hex patch (see output of cu)
| wr                # Write <num> random bytes
| wA                # Alter/modify opcode at current seek (see wA?)
| wc                # Write cache commands
| wz                # Write zero-terminated string
| wt                # Write to file
| wf                # Write data from file, socket, offset
| ww                # Write wide string
| wx                # Write hexadecimal
| wa                # Write opcode, separated by ';'
| wb                # Write in current block with cyclic hexstring
| wm                # Set binary mask hexpair to be used as cyclic write mask
| wo                # Write in block with operation
| wd                # Duplicate N bytes from offset at current seek
| ws                # Write 1 byte for length and then the string
```
With this check instead:
```
| w <string>           # Write string
| wB <value>           # Set bits with given value
| wB- <value>          # Unset bits with given value
| wv <value>           # Write value as 4-bytes/8-bytes based on value
| wv1 <value>          # Write value of 1 byte
| wv2 <value>          # Write value of 2 byte
| wv4 <value>          # Write value of 4 byte
| wv8 <value>          # Write value of 8 byte
| w0 <len>             # Write <len> bytes with value 0x00
| w1+ [<n>]            # Increment a byte
| w1- [<n>]            # Decrement a byte
| w2+ [<n>]            # Increment a word
| w2- [<n>]            # Decrement a word
| w4+ [<n>]            # Increment a dword
| w4- [<n>]            # Decrement a dword
| w8+ [<n>]            # Increment a qword
| w8- [<n>]            # Decrement a qword
| w6d <base64>         # Write the base64-decoded bytes
| w6e <hexstring>      # Write the base64-encoded bytes

```

wargio:
The output is now aligned and i removed the checks as you have asked @ret2libc 
Please do not squash this.

wargio:
looks like `[XX] db/archos/linux-x64/dbg_gdbserver gdbserver continue` failed, but it's unrelated to this PR. probably failed randomly

librz/core/cmd_api.c
```diff
@@ -1089,6 +1090,146 @@ static char *get_help(RzCmd *cmd, RzCmdDesc *cd, RzCmdParsedArgs *args, bool use
...
+ * \brief Generates a JSON output of the given help message description
+ *
+ * \param cmd reference to RzCmd
+ * \param cd reference to RzCmdDesc
+ * \param j reference to PJ
+ *
+ * \return returns false if an invalid argument was given, otherwise true.
+ */
+RZ_API bool rz_cmd_get_help_json(RzCmd *cmd, const RzCmdDesc *cd, PJ *j) {
+	if (!cmd || !cd || !j) {
```
ret2libc: use `rz_return_val_if_fail`

librz/core/cmd_api.c
```diff
@@ -1089,6 +1090,146 @@ static char *get_help(RzCmd *cmd, RzCmdDesc *cd, RzCmdParsedArgs *args, bool use
...
+ *
+ * \param cmd reference to RzCmd
+ * \param cd reference to RzCmdDesc
+ * \param use_color output strings with color codes.
+ * \param sb reference to RzStrBuf
+ *
+ * \return returns false if an invalid argument was given, otherwise true.
+ */
+RZ_API bool rz_cmd_get_help_strbuf(RzCmd *cmd, const RzCmdDesc *cd, bool use_color, RzStrBuf *sb) {
+	if (!cmd || !cd || !sb) {
```
ret2libc: use `rz_return_val_if_fail`

ret2libc (approved):
Please fix the rz_return_val_if_fail, then i'm ok to merge!
Alignment looks awesome now!!

ret2libc (approved):

ret2libc:
Please update on current `dev` and merge when you want ;)

PR #505, opened 2021-02-01
Remove axF

Maijin:
Remove axF, a legacy command used to find xrefs that uses cmd0 + macros (also its untested and does not work)


wargio (approved):

ret2libc (approved):

XVilka (commented):
Good catch!

PR #506, opened 2021-02-01
Add rz_strf() Macro

thestr4ng3r:

**Detailed description**

Convenience macro for local temporary strings
\param buf Target buffer, **must** be an array type, not a pointer.

This eases the common pattern where a stack-allocated string of a fixed
size is created and filled with `snprintf()` to be used as a temporary string.
 
Example:
```c
char k[32];
char v[32];
sdb_set(db, rz_strf(k, "key.%d", 42), rz_strf(v, "val.%d", 123));
```

**Test plan**

see unit tests

### DIFF
--- a/librz/include/rz_util/rz_str.h
+++ b/librz/include/rz_util/rz_str.h
@@ -19,6 +19,25 @@ typedef enum {
 	RZ_STRING_ENC_GUESS = 'g',
 } RzStrEnc;
 
+/**
+ * \brief Convenience macro for local temporary strings
+ * \param buf Target buffer, **must** be an array type, not a pointer.
+ *
+ * This eases the common pattern where a stack-allocated string of a fixed
+ * size is created and filled with `snprintf()` to be used as a temporary string.
+ *
+ * Example:
+ *
+ *     char k[32];
+ *     char v[32];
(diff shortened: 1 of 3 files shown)


XVilka (approved):

librz/include/rz_util/rz_str.h
```diff
@@ -19,6 +19,25 @@ typedef enum {
...
+ * \param buf Target buffer, **must** be an array type, not a pointer.
+ *
+ * This eases the common pattern where a stack-allocated string of a fixed
+ * size is created and filled with `snprintf()` to be used as a temporary string.
+ *
+ * Example:
+ *
+ *     char k[32];
+ *     char v[32];
+ *     sdb_set(db, rz_strf(k, "key.%d", 42), rz_strf(v, "val.%d", 123));
```
ret2libc, line 33: I think it might make sense to define a INTMAXCHARS so that you can use it when defining the array size instead of random number where you are never sure whether that's enough or not

ret2libc (approved):

librz/include/rz_util/rz_str.h
thestr4ng3r: You mean like a constant for how many chars `%d` would take? Could be useful, yes.

PR #512, opened 2021-02-02
Windows subprocess: Remove \r only if followed by \n or not part of esc seq

kazarmy:

**Detailed description**

This pr should prevent an `\r` that is part of an escape sequence from being removed under Windows, and solves https://github.com/rizinorg/rizin/pull/500#issuecomment-770798278.

**Test plan**

All builds are green.

**Closing issues**

...


librz/util/subprocess.c
```diff
@@ -125,6 +125,19 @@ static LPWCH override_env(const char *envvars[], const char *envvals[], size_t e
 	return ret;
 }
 
+static void remove_cr_on_cond(char *str) {
+	char *start = str;
+	while (*str) {
+		if (str[0] == '\r' &&
+		    (str[1] == '\n' ||
+		     !(str - start >= 4 && *(str - 4) == '\x1b') /* HACK */)) {
```
ret2libc: ???

librz/util/subprocess.c
```diff
@@ -125,6 +125,19 @@ static LPWCH override_env(const char *envvars[], const char *envvals[], size_t e
 	return ret;
 }
 
+static void remove_cr_on_cond(char *str) {
```
ret2libc: what does this name mean exactly? I'm not sure I understand what the function is supposed to do based on the name.

librz/util/subprocess.c
kazarmy: Well that condition checks whether the character 4 spaces back is an escape char, and does not remove the `\r` if so ... it's marked `/* HACK */` because the condition shouldn't really be needed, but there seem to be cases where `\r` pops up in the output unexpectedly (e.g. https://ci.appveyor.com/project/rizinorg/rizin/builds/37565774/job/pe7srqflun29gnjm#L4825).

----

Why 4 spaces? Because the escape sequence that needs protection right now is
https://github.com/rizinorg/rizin/blob/154f34f8f1da6ac0223e143c54d4d0da4bbf25c5/librz/include/rz_cons.h#L584 but there are 2 other escape sequences that are similar
https://github.com/rizinorg/rizin/blob/154f34f8f1da6ac0223e143c54d4d0da4bbf25c5/librz/include/rz_cons.h#L583-L585

librz/util/subprocess.c
kazarmy: `remove_cr_on_condition`, but I think I should just shorten it to `remove_cr`

kazarmy: or maybe I should just rename it back to `remove_cr_from_crlf`

kazarmy: 5adf82014edf99fd428e379b4c3b66063951e800 ok renamed the function ... I chose `remove_cr` because it more accurately describes what the function is doing right now.

librz/util/subprocess.c
ret2libc: Let's see if I got the problem right:
rz_subprocess originally tried to remove `\r` so that `\r\n` becomes just `\n`, right?

kazarmy: Why is the `str[1] == '\n'` condition needed? It's because the `/* HACK */` condition protects too much.

kazarmy: > Let's see if I got the problem right:
rz_subprocess originally tried to remove \r so that \r\n becomes just \n, right?

That I think was the original intention, but just doing https://github.com/rizinorg/rizin/pull/512/commits/5bb007faaa1908ab0b823de4bb231e360a261727#diff-04d59abbf789e0c515594c6e252c1c2edec7975d252e1d59d6b767049db87d52R128-R137 resulted in the hot mess of https://ci.appveyor.com/project/rizinorg/rizin/builds/37567072/job/4bqkfj8f006p8pu9#L4083

ret2libc: I see... Do you have any idea why? I'm not a big fan of hacks, as you know :D
Maybe some `rz_str_ansi*` function can help? Like `rz_strstr_ansi`?

kazarmy: > Do you have any idea why?

No unfortunately ... looks like a major investigation in itself

> Maybe some `rz_str_ansi*` function can help? Like `rz_strstr_ansi`?

ok more precise code is 82b3c2436be1493633e8a37b81ae8e57545070c0 using `strncmp`

ret2libc (commented):
cc @thestr4ng3r please review this as well. I'm not super happy with the hack but I don't know what else we can do :/ do you have any idea?

ret2libc (approved):

PR #514, opened 2021-02-02
Remove all flaky download pdb test

ghost:

**Detailed description**
Those tests are flaky by definition.

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/test/db/formats/pdb
+++ b/test/db/formats/pdb
@@ -1,36 +1,3 @@
-NAME=PDB downloader check
-FILE=bins/pdb/user32.dll
-CMDS=<<EOF
-%RZ_CURL=1
-!!rz-bin -PP ${RZ_FILE} ~PDB
-EOF
-EXPECT=<<EOF
-PDB "user32.pdb" download success
-EOF
-RUN
-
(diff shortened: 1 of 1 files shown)


XVilka:
I am not sure it's a good idea. This feature should be tested somehow. Maybe just check the HTTP request instead? So it should match to what really server responds right now.

ghost:
> This feature should be tested somehow. Maybe just check the HTTP request instead? So it should match to what really server responds right now.

We can open an issue to fix it later.

This PR is a small patch to fix quickly this annoying behavior.

ret2libc (changes requested):
I'm actually against this without another test or something else in the same PR (things tends to get forgotten with "later" :D ) also, it just happens some times, but we can always restart the build. I'd prefer to have something better to test these things.

wargio:
not a good idea, but we should allow them to fail, but not crash.

wargio:
or we can run a python server that provides the pdb and the test will download from there.

kazarmy:
It would be great if we could have 3-colour builds: red and green as normal, and yellow if this test fails.

PR #525, opened 2021-02-06
Convert `q` command to newshell

officialcjunior:

**Detailed description**
Initial attempt to convert `q` commands to newshell.
Created respective handlers for the existing commands and handled the quitting with preference case with subcommands.

**Test plan**

~~- [x] Firstly, we can fix handling `rz_cmd_quit_choose_hist_proj` (need to fix the names, too), 
I don't think arguments are handled properly there.~~

- [x] Split `q[y\n][y\n]` into multiple commands and handlers 

- [x] Pass arguments to `cmd_Quit` accordingly.

**Closing issues**

None


librz/core/cmd_descs.yaml
```diff
@@ -590,9 +590,32 @@
...
+      args: []
+    - name: "q!"
+      cname: cmd_force_quit
+      summary: Force quit rizin
+      args: []
+    - name: "q!!"
+      cname: cmd_force_quit_without_history
+      summary: force quit rizin without saving history
+      args: []
+    - name: q
```
ret2libc: Maybe instead of this we can change a bit the way this thing is done... 
What if we just keep one `q` command that optionally accept arguments? First argument is "kill process", second argument is "save project". So if you want to kill and save project you would do `q y y`. If we go this route, just remove this and add the optional arguments in https://github.com/rizinorg/rizin/pull/525/files#diff-4f94ed53f57e9c49ad8a154a2d4e3defb758c0deaf69909274e560bb29a63938R598

@XVilka WDYT?

XVilka: While it makes sense this is not how Vim commands work (Rizin commands are largely inspired by Vim).

ret2libc: So you would stay with `qyy` `qyn` `qny` `qnn`?

XVilka: I think `qyy` is fine probably. Not 100% sure.

ret2libc: Ok, in this case @officialcjunior you should change this to be similar to https://github.com/rizinorg/rizin/blob/dev/librz/core/cmd_descs.yaml#L884

librz/core/cmd_descs.yaml
```diff
@@ -590,9 +590,32 @@
...
+      cname: cmd_quit_choose_hist_proj
+      summary: Quit rizin and choose to save projects and kill the process
+      args_str: " [y/n][y/n]"
```
ret2libc:
```suggestion
      args_str: "<y|n><y|n>"
```

librz/core/cmd_quit.c
```diff
@@ -32,6 +67,16 @@ static int cmd_Quit(void *data, const char *input) {
 	return -2;
 }
 
+RZ_IPI RzCmdStatus rz_cmd_force_quit_handler(RzCore *core, int argc, const char **argv) {
+	cmd_Quit(core, *argv);
```
ret2libc:
```suggestion
	cmd_Quit(core, argv[0] + 1);
```
The first character of argv[0] should be skip.

librz/core/cmd_quit.c
```diff
@@ -15,6 +15,41 @@ static const char *help_msg_q[] = {
...
 };
 
+RZ_IPI RzCmdStatus rz_cmd_quit_handler(RzCore *core, int argc, const char **argv) {
+	core->num->value = 0LL;
+	//exit (*input?rz_num_math (core->num, input+1):0);
+	//if (core->http_up) return false; // cancel quit when http is running
+	return RZ_CMD_STATUS_EXIT;
+}
+
+RZ_IPI RzCmdStatus rz_cmd_quit_choose_hist_proj (RzCore *core, int argc, const char **argv) {
```
ret2libc: this is not right. If we go down the route of having `qyn`/`qyy`/`qnn`/`qny`, the choose_project/choose_process are not separate arguments but you have to get those from the name of the command. You can just use different handlers in each case. Please see https://github.com/rizinorg/rizin/pull/525/files#r572686967 first.

ret2libc (changes requested):

librz/core/cmd_quit.c
officialcjunior: Yes, now I have understood that. And yes, we will need to use different handlers.
👍🏾

librz/core/cmd_quit.c
```diff
@@ -15,6 +15,34 @@ static const char *help_msg_q[] = {
...
 
+RZ_IPI RzCmdStatus rz_cmd_quit_handler(RzCore *core, int argc, const char **argv) {
+	core->num->value = 0LL;
+	//exit (*input?rz_num_math (core->num, input+1):0);
+	//if (core->http_up) return false; // cancel quit when http is running
+	return RZ_CMD_STATUS_EXIT;
+}
+
+RZ_IPI RzCmdStatus rz_quit_kill_save_handler(RzCore *core, int argc, const char **argv) {
+	core->num->value = 5;
```
ret2libc: Use some defines for these values (which I'm not sure what they mean... but let's keep them for now).
For example, RZ_QUIT_VALUE_KILL, RZ_QUIT_VALUE_SAVE, RZ_QUIT_VALUE_NOKILL, etc.

librz/core/cmd_quit.c
```diff
@@ -15,6 +15,34 @@ static const char *help_msg_q[] = {
 	NULL
 };
 
+RZ_IPI RzCmdStatus rz_cmd_quit_handler(RzCore *core, int argc, const char **argv) {
+	core->num->value = 0LL;
+	//exit (*input?rz_num_math (core->num, input+1):0);
```
ret2libc: Remove commented code please.

librz/core/cmd_descs.yaml
```diff
@@ -590,9 +590,42 @@
...
+      summary: Force quit rizin
+      args: []
+    - name: "q!!"
+      cname: cmd_force_quit_without_history
+      summary: force quit rizin without saving history
+      args: []
+    - name: "q"
+      cname: quit_with_preference
+      summary: Quit rizin and choose to save projects and kill the process
+      args_str: "<y|n><y|n>"
```
ret2libc:
```suggestion
      options: "<y|n><y|n>"
```

librz/core/cmd_descs.yaml
```diff
@@ -590,9 +590,42 @@
...
+      type: RZ_CMD_DESC_TYPE_INNER
+      subcommands:
+        - name: qyy
```
ret2libc: Do similar to `w1` / `w2`, etc.... Group `qyy` and `qyn` under `qy` and `qny`/`qnn` under `qn`.

librz/core/cmd_descs.yaml
```diff
@@ -590,9 +590,48 @@
...
+      summary: Force quit rizin
+      args: []
+    - name: "q!!"
+      cname: cmd_force_quit_without_history
+      summary: force quit rizin without saving history
+      args: []
+    - name: "q"
+      cname: quit_with_preference
+      summary: Quit rizin and choose to save projects and kill the process
+      options: "<y|n><y|n>"
```
ret2libc: actually... sorry, for consistency this should be:
```suggestion
      options: "<yn><yn>"
```

ret2libc (commented):
LGTM, just fix that last thing and we can merge IMO.

librz/core/cmd_descs.yaml
```diff
@@ -590,9 +590,48 @@
...
+      cname: cmd_force_quit
+      summary: Force quit rizin
+      args: []
+    - name: "q!!"
+      cname: cmd_force_quit_without_history
+      summary: force quit rizin without saving history
+      args: []
+    - name: "q"
+      cname: quit_with_preference
+      summary: Quit rizin and choose to save projects and kill the process
```
ret2libc:
```suggestion
      summary: Quit rizin and choose to kill the process and save projects
```
Because the first <yn> refers to the kill, the second to the projects.

ret2libc (changes requested):

librz/core/cmd_descs.yaml
officialcjunior: Oh, yeah, missed that one. All of the summaries seem to be fine.

ret2libc (approved):
Thanks!

Issue #491, opened 2021-01-29
Refactor Visual Menus `/librz/core/tui/vmenus*` to use API instead of commands


XVilka:
Currently the code of Visual menus heavily relies on running commands instead of just calling the proper API functions.
To reduce the overhead, code cruft, and code breakage in case of the command changes it is important to remove all those command calls and call the proper API directly.

```
[i] ℤ rg "rz_core_cmd[0f_]*\(" librz/core/tui/vmenus*                                                                                                                                                                                    librz/core/tui/vmenus_graph.c
421:                  rz_core_cmd0(core, cmd);

librz/core/tui/vmenus.c
496:                  rz_core_cmdf(core, ".dte %d", i);
498:          rz_core_cmd0(core, "x 64@r:SP");
502:          rz_core_cmdf(core, "dtd %d", delta);
585:  rz_core_cmd0(core, cmd);
850:                  rz_core_cmd0(core, "s $(afl~...)");
885:                          rz_core_cmd0(core, sortModes[sortMode % 4]);
```

ret2libc:
I think this can probably be moved to a later release. It's just refactoring and we seem to have already a lot of things in our hands.

PR #531, opened 2021-02-06
Refactor to use API in visual mode

XVilka:

**Detailed description**

Refactor to use the API directy instead of calling `rz_core_cmd()` for some actions in the Visual Menus (`Vv` mode).

**Test plan**

- Try to add and remove the functions in this mode, also to navigate debug traces.
- Try to run Rizin in debug mode `rizin -d /bin/ls` then enter Visual mode, show the registers, scroll a bit and press `$` to set PC to the selected address
- Try to add comment
- Try to add/remove local flags (labels)

**Closing issues**

Partially addresses https://github.com/rizinorg/rizin/issues/491

### DIFF
--- a/librz/core/visual.c
+++ b/librz/core/visual.c
@@ -1325,7 +1325,7 @@ static void addComment(RzCore *core, ut64 addr) {
 	if (rz_cons_fgets(buf, sizeof(buf), 0, NULL) < 0) {
 		buf[0] = '\0';
 	}
-	rz_core_cmdf(core, "\"CC %s\"@0x%08" PFMT64x, buf, addr);
+	rz_meta_set_string(core->analysis, RZ_META_TYPE_COMMENT, addr, buf);
 	rz_core_visual_showcursor(core, false);
 	rz_cons_set_raw(true);
 }
@@ -2524,9 +2524,9 @@ RZ_API int rz_core_visual_cmd(RzCore *core, const char *arg) {
 			break;
 		case '$':
 			if (core->print->cur_enabled) {
-				rz_core_cmdf(core, "dr PC=$$+%d", core->print->cur);
+				rz_core_debug_reg_set(core, "PC", core->offset + core->print->cur, NULL);
 			} else {
-				rz_core_cmd0(core, "dr PC=$$");
+				rz_core_debug_reg_set(core, "PC", core->offset, NULL);
 			}
 			break;
 		case '@':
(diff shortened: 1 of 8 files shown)


librz/core/vmenus.c
```diff
@@ -2229,7 +2230,7 @@ RZ_API int rz_core_visual_comments(RzCore *core) {
 		case ' ':
 		case '\r':
 		case '\n':
-			rz_core_cmdf(core, "s 0x%" PFMT64x, from);
+			rz_core_seek(core, from, true);
```
ret2libc:
```suggestion
			rz_core_seek_and_save(core, from, true);
```

ret2libc (commented):
LGTM apart from that thing.

officialcjunior (approved):

librz/core/visual.c
```diff
@@ -2771,25 +2778,21 @@ RZ_API int rz_core_visual_cmd(RzCore *core, const char *arg) {
 			visual_refresh(core);
 			break;
 		case ' ':
-		case 'V':
-			if (rz_config_get_b(core->config, "graph.web")) {
```
XVilka, line 2775: I removed this case since it doesn't work either and essentially shows also an ASCII graph too.

Maijin: I'd kill graph.web completely or open an issue to do it in another PR.

Maijin (approved):

librz/core/visual.c
```diff
@@ -1325,7 +1325,7 @@ static void addComment(RzCore *core, ut64 addr) {
 	if (rz_cons_fgets(buf, sizeof(buf), 0, NULL) < 0) {
 		buf[0] = '\0';
 	}
-	rz_core_cmdf(core, "\"CC %s\"@0x%08" PFMT64x, buf, addr);
+	rz_meta_set(core->analysis, RZ_META_TYPE_COMMENT, addr, 1, buf);
```
ret2libc: I see that the `CC` command code actually does `rz_meta_set_string`. Does it change anything?

librz/core/visual.c
```diff
@@ -2522,13 +2522,17 @@ RZ_API int rz_core_visual_cmd(RzCore *core, const char *arg) {
 		case 'c':
 			setcursor(core, !core->print->cur_enabled);
 			break;
-		case '$':
-			if (core->print->cur_enabled) {
```
ret2libc, line 2526: I would convert all these `dr <reg>=<val>` a bit differently. I think the command handler for `dr` does slightly different things (e.g. it consider whether the reg is flags or not), so I would refactor a bit that part and use a function there instead of doing this. See how I did it for `ar <reg>=<val>` with `rz_core_analysis_set_reg`.

ret2libc (changes requested):

librz/core/visual.c
officialcjunior: > I'd kill graph.web completely or open an issue to do it in another PR.


I'll look into that.

Also, if I were to remove that does that mean that the cases (if statements) like [these](https://github.com/rizinorg/rizin/blob/dev/librz/core/cmd_analysis.c#L8590-L85924) can just be removed?

librz/core/visual.c
XVilka: I think it's not required if we know the reg is PC which is the case here.

librz/core/visual.c
XVilka: Good catch, fixed!

librz/core/visual.c
Maijin: Right

librz/core/visual.c
ret2libc: I know, but I still think it's better to have the same code here, for consistency. In particular, the `dr <reg>=<val>` code should be refactored in a function similar to `rz_core_analysis_set_reg` that takes a `regname` and a `val` as arguments. For example, `dr` also does `rz_core_debug_regs2flags` which is not done here. Please introduce that API now, because otherwise later it will become almost impossible to understand that these pieces of code were actually using `dr`.

XVilka: Ok, done. I introduced new `RZ_IPI bool rz_core_debug_reg_set(RzCore *core, const char *regname, ut64 regval, const char *ostr)`. Note the two arguments, I know it's a bit ugly but I have to use both because of how `dr` is actually implemented. Let me know if you have better ideas.

XVilka:
I moved a bit of the code for the consistency too.

ret2libc (approved):

PR #540, opened 2021-02-07
Fix dead assignment detected by clang CI

ghost:

**Detailed description**

...

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/librz/io/p/io_rzpipe.c
+++ b/librz/io/p/io_rzpipe.c
@@ -16,7 +16,6 @@ static int __write(RzIO *io, RzIODesc *fd, const ut8 *buf, int count) {
 	char fmt[4096];
 	char *bufn, bufnum[4096];
 	int i, rv, rescount = -1;
-	char *res, *r;
 	if (!fd || !fd->data) {
 		return -1;
 	}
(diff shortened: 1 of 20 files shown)


librz/io/p/io_rzpipe.c
```diff
@@ -39,13 +38,8 @@ static int __write(RzIO *io, RzIODesc *fd, const ut8 *buf, int count) {
 		eprintf("rzpipe_write: error\n");
 		return -1;
 	}
-	res = rzpipe_read(RZP(fd));
```
ret2libc, line 42: sure about this?? Maybe `res` is dead, but the call could have other effects, right?

librz/util/buf.c
```diff
@@ -541,7 +541,6 @@ static st64 buf_format(RzBuffer *dst, RzBuffer *src, const char *fmt, int n) {
 			}
 
 			m = 1;
-			bigendian = true;
```
ret2libc, line 544: is this really dead? I'm not sure 100%

ret2libc (changes requested):

librz/util/buf.c
ghost: bigendian is assign in the switch line 461

ret2libc: it is not always set. Some cases do not set it and it may be used after that AFAIS.

ghost: case '0' - '9' -> continue (loop)
case 's', 'S', 'i', 'I', 'l', 'L', 'c' -> set bigendian
default -> return from the function

ghost: There is no side effect if the line 544 is removed

ret2libc: What happens if there is a number and then the fmt string ends? In that case, the previously set bigendian value would be considered, no?

ghost: bigendian is a local loop variable (line 458)

So the case of a number and then fmt strings ends as no impact

ghost: I don't see what I am missing.

ret2libc: Ok I think you are right!

ret2libc (approved):

PR #597, opened 2021-02-12
Refactor to use API in debug commands

XVilka:

**Detailed description**

Switch to use API instead of `rz_core_cmd*()` calls in `librz/core/cmd_debug.c` - all `d*` commands

**Test plan**

- Debug commands that involve ESIL - e.g. `des` and `desu` should behave as before
- `dcf` should behave as before
- `dc` command should behave as before
- `ds` commands should behave as before
- All tests should be green

### DIFF
--- a/librz/core/cmd_debug.c
+++ b/librz/core/cmd_debug.c
@@ -3880,7 +3880,7 @@ static void rz_core_debug_esil(RzCore *core, const char *input) {
 		if (rz_debug_esil_watch_empty(core->dbg)) {
 			eprintf("Error: no esil watchpoints defined\n");
 		} else {
-			rz_core_cmd0(core, "aei");
+			rz_core_analysis_esil_init(core);
 			rz_debug_esil_prestep(core->dbg, rz_config_get_i(core->config, "esil.prestep"));
 			rz_debug_esil_continue(core->dbg);
 		}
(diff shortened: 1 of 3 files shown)


ret2libc (approved):

librz/core/cmd_debug.c
```diff
@@ -3880,15 +3880,15 @@ static void rz_core_debug_esil(RzCore *core, const char *input) {
 		if (rz_debug_esil_watch_empty(core->dbg)) {
 			eprintf("Error: no esil watchpoints defined\n");
 		} else {
-			rz_core_cmd0(core, "aei");
+			rz_core_analysis_esil_init(core);
```
yossizap, line 3883: Aren't you supposed to also run `aeim`/`cmd_esil_mem` with `aei`? `rz_core_analysis_esil_init` doesn't call it

yossizap: Also, calling rz_core_analysis_esil_init for each step seems redundant, maybe check if `core->analysis->esil` is initialized?

ret2libc: I think `aei` command is not calling `aeim` as well.

XVilka: I preserved the logic as is. If you think it could be changed it should probably done in a separate PR. But usually yes, you need to run also the `aeim` too. Not sure why it wasn't done here before.

yossizap (approved):

PR #598, opened 2021-02-12
Get libuv from GitHub

kazarmy:

**Detailed description**

GitHub should be more reliable than https://dist.libuv.org/ and this pr should also solve the following error (https://github.com/rizinorg/rizin/runs/1884711099#step:9:6428):

libuv_ssl

**Test plan**

All builds are green.

**Closing issues**

...


XVilka:
I wonder why source hashes of the same distribution are different...

kazarmy:
From dist.libuv.org:

sha256_dist

From GitHub:

sha256_github

Could be due to name ordering in the tar.gz file.

kazarmy:
Oh dear

ret2libc:
I'm not sure the problem is about reliability. I believe the problem is that docker build containers have no internet access. I thought I did it with last commits that predownloaded all subprojects... But it seems it didn't work

kazarmy:
Well they seem to be doing `git clone` just fine.

ret2libc:
> Well they seem to be doing `git clone` just fine.

Yeah I guess you are right.
However I would expect the subprojects to not be downloaded each time because of the `meson subprojects download` I put before building the docker containers...

kazarmy:
Ok as a last try I'm going to try upgrading the Docker base image. If that doesn't work then perhaps the only option left on this track is to patch Meson itself which doesn't seem pleasant, so I'll probably close the pr instead.

ret2libc:
Ahhh, right! The problem is indeed within the dockerfile... It downloads rizin from git instead of using the current context. I think the `git clone -b "$RZ_VERSION" https://github.com/rizinorg/rizin.git && \` line should be removed/replaced with something else and probably there should be a directory somewhere that can be mapped in the container.

kazarmy:
Oh it didn't work :slightly_frowning_face:

> I'm not sure the problem is about reliability.

Yeah dist.libuv.org should work fine.

> It downloads rizin from git instead of using the current context. I think the git clone -b "$RZ_VERSION" https://github.com/rizinorg/rizin.git && \ line should be removed/replaced with something else and probably there should be a directory somewhere that can be mapped in the container.

I understand that you want to fetch code from outside the container ... sgtm

Ok closing this pr.

kazarmy:
Just curious to know whether using Ubuntu 20.04 as base image will work or not.

kazarmy:
Hmm I guess I need to study Docker some more 😅

PR #599, opened 2021-02-12
Move mpc to subprojects: keep it in-repo because small

ret2libc:


thestr4ng3r (approved):
Good idea to move out. I expect it to be replaced by tree-sitter in the long run.

XVilka (approved):

Issue #397, opened 2021-01-20
Refactor Graph mode (`librz/core/agraph.c`) to use API instead of commands


XVilka:
Currently the code of Graph rendering modes heavily relies on running commands instead of just calling the proper API functions.
To reduce the overhead, code cruft, and code breakage in case of the command changes it is important to remove all those command calls and call the proper API directly.

```
$ rg "rz_core_cmd[0f_]*\(" librz/core/agraph.c
3474:                 rz_core_cmd0(core, cmdv);
3981:         rz_core_cmd0(core, buf);
4032:                         rz_core_cmd0(core, cmd);
4370:                                 rz_core_cmd0(core, ".axfg @$FB");
4380:                         rz_core_cmd0(core, ".axtg $FB");
4386:                 rz_core_cmd0(core, ".dtg*");
4405:                         rz_core_cmd0(core, key_s);
4831:                         (void)rz_core_cmd0(core, cmd);
4837:                         (void)rz_core_cmd0(core, cmd);
4845:                         (void)rz_core_cmd0(core, cmd);
4851:                         (void)rz_core_cmd0(core, cmd);
4857:                         (void)rz_core_cmd0(core, cmd);
4863:                         (void)rz_core_cmd0(core, cmd);
4869:                         (void)rz_core_cmd0(core, cmd);
4877:                         (void)rz_core_cmd0(core, cmd);
4885:                         (void)rz_core_cmd0(core, cmd);
4893:                         (void)rz_core_cmd0(core, cmd);
4899:                         (void)rz_core_cmd0(core, cmd);
4905:                         (void)rz_core_cmd0(core, cmd);
```

Note, some of these calls are legit, e.g. when the command is associated with some keybinding:
```c
			cmd = rz_config_get(core->config, "key.f3");
			if (cmd && *cmd) {
				(void)rz_core_cmd0(core, cmd);
			}
```

PR #604, opened 2021-02-12
Refactor to replace `sr` and `aei` calls with the API ##refactor

officialcjunior:

**Detailed description**

Completely replace the places where `sr` and `aei` is called
in the codebase with its corresponding APIs.

Also have some minor improvements in certain places.

**Test plan**

All tests should be green and the CI should pass.

**Closing issues**
Partially solves #397


XVilka (approved):

librz/core/agraph.c
```diff
@@ -4238,10 +4238,12 @@ RZ_API int rz_core_visual_graph(RzCore *core, RzAGraph *g, RzAnalysisFunction *_
 			break;
 			// Those hardcoded keys are useful only for aegi, should add subcommand of ag to set key actions
 		case '1':
-			rz_core_cmd0(core, "so;.aeg*");
+			rz_core_seek_opcode(core, 4, false);
```
ret2libc: why 4?

officialcjunior: Oh yes, this is something I forgot to add as a WIP/to ask.
I was trying to find out the equivalent of calling `so`. When I ran that, four bytes got seeked. I just gave 4 even though I didn't realise how it's done.

I think we should pass 0 or something that's 'nothing'?

ret2libc: Pass `1`. In your case it went ahead of 4 bytes because probably the instruction at core->offset was 4 bytes.

XVilka (approved):

PR #618, opened 2021-02-13
Make Tests use x86.as only on x86 Host

thestr4ng3r:

**Detailed description**

x86.as depends an the `as` executable to be available, which usually on non-x86 isn't. I rewrote tests that didn't explicitly test this plugin to not use it and moved the others to archos.

**Test plan**

run tests on e.g. arm

### DIFF
--- a/test/db/archos/linux-x64/asm_x64_as
+++ b/test/db/archos/linux-x64/asm_x64_as
@@ -1,6 +1,5 @@
 NAME=rz-asm #1167 3
 FILE=-
-# BROKEN=1
 CMDS=!rz-asm -s att -a x86.as -b 64 "test %rbx, %rax"
 EXPECT=<<EOF
 4885d8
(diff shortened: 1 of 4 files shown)


XVilka (approved):

ret2libc (commented):
Why x86.as shouldn't be available on non-x86 machines? I think on Mac you can install gnu-binutils and you can have assemblers for other architectures anyway, no?

wargio:
i believe the issue is valid.

thestr4ng3r:
It is possible yes, like here for example:
https://github.com/rizinorg/rizin/blob/3f71f45c533e81fc6e5cc6c298709eeed3bf1144/.travis.yml#L48
You would have to set the `RZ_X86_AS` env var. I think the question is whether the entire test suite succeeding should depend on this to be available. An alternative would be to put it into `db/extras` or similar.

Maijin:
It's a good first step but it only underlines a deeper issue/enhancement to work on. Imho x86.as should be killed completely because there is already `e asm.assembler`.

Imho rizin should automatically list the assembler available in asm.assembler wether it's then embedded (e.g x86.nz but also all the one available for other architectures), `r2pm` when installed (keystone) and the system's one (nasm, or other etc.) and show that to user also in e `asm.assembler`...

ret2libc:
Anyway, ok to merge, do not let me block you ;)

PR #627, opened 2021-02-15
ci: push Docker tag also for tagged release

ret2libc:
**Detailed description**

Right now on Docker Hub you have the `stable` tag, but actually you may want the v1.0, etc. tags as well. I *think* this should address that problem, but it needs to be tested during the v0.1.1 release.

**Test plan**

Check when doing the v0.1.1 release if the v0.1.1 tag is pushed to Docker Hub


ret2libc:
I'm keeping this open as I think it makes sense to merge it as last before 0.1.1 goes out, to be sure we test if it really works.

XVilka (approved):
I guess the only way to know is to try.

PR #637, opened 2021-02-16
Extract common API to generate different types of graphs/outputs ##breaking

ret2libc:

**Detailed description**

Extract a common API to print graphs (`ag` commands) as needed. Few things to consider about this:
- unfortunately existing code did slightly different things for different types of graph/output so when possible i tried to refactor these differences and just use one single code to output a graph in different modes. What this means is that some commands now might output slightly different output. For example, `agc`/`agC` in json mode printed:
```
[
  {
    "name": "entry0",
    "size": 46,
    "imports": [
      "reloc.__libc_start_main"
    ]
  }
]
```
while now it prints:
```
{
  "nodes": [
    {
      "id": 0,
      "title": "entry0",
      "body": "",
      "out_nodes": [
        1
      ]
    },
    {
      "id": 1,
      "title": "reloc.__libc_start_main",
      "body": "",
      "out_nodes": [
        
      ]
    }
  ]
}
```
which is the common format.
- Only `agf` commands/subcommands was not refactored for now, because it involves a lot of changes, so right now `agf` is kinda of a special case, because it uses completely different code to print a graph and to write it in different formats.

Probably Cutter needs to be fixed for the `agc`/`agC` (and maybe other) cases, while `agf` would still work.

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/librz/core/cagraph.c
+++ b/librz/core/cagraph.c
@@ -399,6 +399,37 @@ RZ_IPI void rz_core_agraph_xrefs_create(RzCore *core, ut64 addr) {
 	rz_list_free(list);
 }
 
+RZ_IPI void rz_core_agraph_esil_create(RzCore *core, const char *expr) {
+	RzAnalysisEsilDFG *edf = rz_analysis_esil_dfg_expr(core->analysis, NULL, expr);
+	RzListIter *iter, *ator;
+	RzGraphNode *node, *edon;
+	RzStrBuf *buf = rz_strbuf_new("");
+	rz_list_foreach (rz_graph_get_nodes(edf->flow), iter, node) {
+		const RzAnalysisEsilDFGNode *enode = (RzAnalysisEsilDFGNode *)node->data;
+		char *esc_str = rz_str_escape(rz_strbuf_get(enode->content));
+		rz_strbuf_set(buf, esc_str);
+		if (enode->type == RZ_ANALYSIS_ESIL_DFG_BLOCK_GENERATIVE) {
+			rz_strbuf_prepend(buf, "generative:");
+		}
+		rz_strbuf_append(buf, "\n");
+		char title[32];
+		rz_agraph_add_node_with_color(core->graph, rz_strf(title, "%d", enode->idx), rz_strbuf_get(buf), -1);
+		free(esc_str);
+	}
+	rz_strbuf_free(buf);
+
+	rz_list_foreach (rz_graph_get_nodes(edf->flow), iter, node) {
+		const RzAnalysisEsilDFGNode *enode = (RzAnalysisEsilDFGNode *)node->data;
+		rz_list_foreach (rz_graph_get_neighbours(edf->flow, node), ator, edon) {
+			const RzAnalysisEsilDFGNode *edone = (RzAnalysisEsilDFGNode *)edon->data;
+			char u[32], v[32];
+			rz_core_agraph_add_edge(core, rz_strf(u, "%d", enode->idx), rz_strf(v, "%d", edone->idx));
+		}
+	}
+
+	rz_analysis_esil_dfg_free(edf);
+}
+
 RZ_IPI void rz_core_agraph_bb_create(RzCore *core, ut64 addr) {
 	RzAnalysisBlock *bb;
 	RzListIter *iter;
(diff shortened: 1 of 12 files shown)


ret2libc:
The refactoring is not perfect IMO, but probably a better one would require a much bigger investment in time which I do not think would pay off right now. I believe the `rz_core_agraph_print_type` API is more or less what we want and we can probably improve the inner details within it.

librz/core/cagraph.c
```diff
@@ -198,4 +202,441 @@ RZ_IPI void rz_core_agraph_print_gml(RzCore *core) {
...
+		}
+	}
+}
+
+static int funccall_analysis_ref_cmp(const RzAnalysisRef *ref1, const RzAnalysisRef *ref2) {
+	return ref1->addr != ref2->addr;
+}
+
+static void funccall_create_graph(RzCore *core, RzAnalysisFunction *fcn) {
+	bool refgraph = rz_config_get_i(core->config, "graph.refs");
```
XVilka, line 252: `rz_config_get_b()`

librz/core/cagraph.c
```diff
@@ -198,4 +202,441 @@ RZ_IPI void rz_core_agraph_print_gml(RzCore *core) {
...
+	RzList *refs = rz_analysis_function_get_refs(fcn);
+	RzList *calls = rz_list_new();
+	RzListIter *iter2;
+	RzAnalysisRef *fcnr;
+
+	rz_list_foreach (refs, iter2, fcnr) {
+		if (fcnr->type == 'C' && rz_list_find(calls, fcnr, (RzListComparator)funccall_analysis_ref_cmp) == NULL) {
```
XVilka, line 259: We should convert `fcnr->type` to enums as well.

librz/core/cagraph.c
```diff
@@ -198,4 +202,441 @@ RZ_IPI void rz_core_agraph_print_gml(RzCore *core) {
...
+	if (rz_list_empty(fcns)) {
+		return;
+	}
+	RzAnalysisFunction *fcn = rz_list_get_n(fcns, 0);
+
+	RzConfigHold *hc = rz_config_hold_new(core->config);
+	rz_config_hold_i(hc, "scr.color", "scr.utf8", "asm.marks", "asm.offset", "asm.lines",
+		"asm.cmt.right", "asm.cmt.col", "asm.lines.fcn", "asm.bytes", NULL);
+	/*rz_config_set_i (core->config, "scr.color", 0);*/
+	rz_config_set_i(core->config, "scr.utf8", 0);
```
XVilka, line 446: Some of those are boolean, please use `rz_config_set_b()` where appropriate.

XVilka (changes requested):

librz/core/cagraph.c
ret2libc: Yep, I actually didn't touch that code but mostly copied-pasted from other locations and forgot about the new set_b ;)

karliss:
As someone who started refactoring this in different direction  (so take my opinion with grain of salt) with `rz_graph_drawable` i think this PR are making some things worse. Things I don't like in this PR:
* functions converting graph from each respective rizin representation shouldn't be storing the result in single global(core->graph) variable, it makes the API worse and introduces problems with it getting overwritten
* converting from the common graph to selected output format should be separate function from getting the graph
* in some cases information about addresses is lost

ret2libc:
> As someone who started refactoring this in different direction (so take my opinion with grain of salt) with `rz_graph_drawable` i think this PR are making some things worse. Things I don't like in this PR:
> 
> * functions converting graph from each respective rizin representation shouldn't be storing the result in single global(core->graph) variable, it makes the API worse and introduces problems with it getting overwritten

I absolutely agree about this point, but  most of the existing codes related to graph was already using the global graph and I didn't mean to change too many "concepts" in one shot. In particular, most code was doing things like `rz_core_cmd(".agi*;agg")`. That's the main thing I intended to "fix".

> * converting from the common graph to selected output format should be separate function from getting the graph

That's why you have all the `rz_core_agraph_data_create`, `rz_core_agraph_imports_create`, etc. and then you have the `rz_core_agraph_print_json`, `rz_core_agraph_print_ascii`, etc. The `rz_core_agraph_print_type` API puts those things together as many times you want to create the graph and print it immediately. Does it make sense?

> * in some cases information about addresses is lost

This is probably true... I do wonder, however, if having those info in a "drawable graph" is really what you want, instead of having those info available somewhere else.

ret2libc:
Let's close this for now. I'm going to do smaller refactorings for the commands part and then we can think to something better for the graph mess.

PR #651, opened 2021-02-18
core: API for `aesp` and `afb` and remove `abb`,`aab`, `abr`, `ab.` ##breaking

ret2libc:

**Detailed description**

* Extract API for `aesp` command
* Remove `abb`, `aab`, `abr` commands as they were not tested at all and they were implementing another algorithm. Unfortunately, it could not even detect code by following a direct jmp.
* Extract API for `afb` sub-commands.
* Remove `ab.` because it is the same as `ab`.

**Test plan**

...

**Closing issues**

...


librz/core/panels.c
```diff
@@ -131,7 +131,7 @@ static char *menus_Debug[] = {
 };
 
 static const char *menus_Analyze[] = {
-	"Function", "Symbols", "Program", "BasicBlocks", "Calls", "References",
+	"Function", "Symbols", "Program", "Calls", "References",
```
XVilka, line 134: Any particular reason why you removed this from panels mode?

XVilka (approved):

librz/core/panels.c
ret2libc: Because it was executing `rz_cmd_analysis_blocks`, which is essentially the `aab` command, that I removed as well.

librz/core/cmd_analysis.c
```diff
@@ -97,7 +96,6 @@ static const char *help_msg_ab[] = {
 	"ab", " [addr]", "show basic block information at given address",
 	"ab.", "", "same as: ab $$",
 	"aba", " [addr]", "analyze esil accesses in basic block (see aea?)",
-	"abb", " [length]", "analyze N bytes and extract basic blocks",
```
kazarmy, line 100: Maybe `abb` can be redefined as "analyze N bytes and run `af` as often as possible"?

ret2libc: what does that mean?

kazarmy: 1. Run `af` on the first byte that isn't inside a function.
2. Do step 1 repeatedly until you go above `length`.

`af` might fail but that's a minor detail.

ret2libc: I would leave that for another day :/ It is not *that* simple imo.

PR #535, opened 2021-02-06
[WIP] Fix gdbserver test read packet failure

yossizap:
**Your checklist for this pull request**
- [x] I've read the [guidelines for contributing](https://github.com/rizinorg/rizin/blob/master/DEVELOPERS.md) to this repository
- [x] I made sure to follow the project's [coding style](https://github.com/rizinorg/rizin/blob/master/DEVELOPERS.md#code-style)
- [ ] I've added tests that prove my fix is effective or that my feature works (if possible)
- [ ] I've updated the documentation and the [rizin book](https://github.com/rizinorg/book) with the relevant information (if needed)

**Detailed description**

The tests seem to fail randomly after failing to read some packets but I can't reproduce it locally, trying different solutions

**Test plan**

CI is green after multiple re-runs

Issue #275, opened 2020-12-27
Replace TCC Header Parser by Tree-Sitter based one


ghost:
## Current State

There are currently two parsers for different parts of C syntax in rizin:

Parsing of C headers for struct definitions, enums, etc. using tcc into **base types** for the database:
https://github.com/rizinorg/rizin/blob/a3a339b11b1e241ee95201d7b666ff291c8e2100/librz/include/rz_parse.h#L73-L77

Parsing C Type expressions like `char *[42]` using a custom mpc-based parser into **ctypes** to be used on top of base types in variables, etc:
https://github.com/rizinorg/rizin/blob/a3a339b11b1e241ee95201d7b666ff291c8e2100/librz/include/rz_parse.h#L79-L115

The tcc one for base types has several limitations such as usage of global state and it also writes out directly the sdb records, which is super unflexible and hard to inspect. The mpc one for ctypes works ok, but mpc also has some aspects that make it annoying to use and extend (see the amount of strcmp in https://github.com/rizinorg/rizin/blob/a3a339b11b1e241ee95201d7b666ff291c8e2100/librz/parse/ctype.c).

## Solution

[tree-sitter-cpp](https://github.com/tree-sitter/tree-sitter-cpp/) should be used to replace tcc. The parser for base types should be written in such a way that it emits `RzAnalysisBaseType` structs and not raw sdb. Thus, the results will be able to be used more flexible or just stored into the database through the api.

Because a complete C(++) parser and thus presumably also tree-sitter-cpp must also parse cast expressions such as `(char *[42])<something>`, the parser for the `char *[42]` in this example could also be re-used directly for `RzParseCTypeType`, replacing mpc there.

## Conflicts/Dependencies

This has a little bit of a conflict with #371 if the parser is done first because the new parser will generate `RzAnalysisBaseType` objects, but #371 will modify the structure of it to use `RzParseCTypeType` instead of strings, forcing the newly written parser to be adapted too.
If however the parser is done second and #371 first, then perhaps the old mpc parser will have to be integrated in many more places, making refactoring to the tree sitter parser a bit harder.
I think the best solution to solve this will be to implement the **parser first**, and in the parser creating a `RzParseCTypeType` whenever such a type is encountered, but only in the very last step when constructing the `RzAnalysisBaseType`, converting that to a string. As such, this issue depends on #370.

[...]

PR #661, opened 2021-02-19
Fix uninitialized argument value detected by clang CI

ghost:

**Detailed description**
~The gdb code is really strange. I don't understand how it worked.
I fixed the warning when I prevent the token processing if j != 4.~

PS: Should we move shlr/gdb and shlr/tcc to subprojects?

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/librz/bin/mangling/microsoft_demangle.c
+++ b/librz/bin/mangling/microsoft_demangle.c
@@ -702,7 +702,7 @@ static size_t get_namespace_and_name(const char *buf, STypeCodeStr *type_code_st
 				tmp += state.amount_of_read_chars + 1;
 				read_len += state.amount_of_read_chars + 1;
 			}
-			char *demangled;
+			char *demangled = NULL;
 			if (parse_microsoft_mangled_name(tmp + 2, &demangled, &len) != eDemanglerErrOK) {
 				break;
 			}
--- a/librz/core/canalysis.c
+++ b/librz/core/canalysis.c
@@ -5749,11 +5749,12 @@ RZ_API void rz_core_analysis_esil(RzCore *core, const char *str, const char *tar
 		}
 		rz_core_analysis_esil_init_mem(core, NULL, UT64_MAX, UT32_MAX);
 	}
+	const char *spname = rz_reg_get_name(core->analysis->reg, RZ_REG_NAME_SP);
 	EsilBreakCtx ctx = {
 		&op,
 		fcn,
-		rz_reg_get_name(core->analysis->reg, RZ_REG_NAME_SP),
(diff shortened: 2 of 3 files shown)


wargio (approved):

XVilka:
1. There is no need to move `shlr/gdb` to subprojects since it's originally written for Radare2 and not used anywhere else, it's not an external library.
2. `shlr/tcc` shouldn't be a subproject since it contains a lot of the custom changes to the TinyCC code and should stay as is until [conversion to Tree-Sitter is done](https://github.com/rizinorg/rizin/issues/275). See also https://github.com/rizinorg/rizin/issues/209
3. We really need to run GDB tests on CI, see also https://github.com/rizinorg/rizin/pull/535

shlr/gdb/src/arch.c
```diff
@@ -76,27 +77,27 @@ gdb_reg_t *arch_parse_reg_profile(const char *reg_profile) {
...
-			if (j > PARSER_MAX_TOKENS - 1) {
+			if (j >= PARSER_MAX_TOKENS) {
 				break;
 			}
 			// Save the token
 			tok[j++] = strdup(tmp);
 		}
 		// Empty line, eww
-		if (j) {
+		if (j == 4) {
```
ret2libc: @yossizap could you have a look and approve if you feel it is ok?

yossizap: I don't think that this is right, 5 is the maximum number of tokens here but it can be higher(defined as 8 in profile.c).

This entire function is a duplication of `rz_reg_set_profile_string` from `librz/reg/profile.c` since that API isn't accessible here and RzReg isn't available(this is bad and hopefully will change soon). The implementation in `profile.c` also properly treats lines that start with `=` that are j==2 and there must be other edge cases here with flag tokens etc. I would keep it the way it was until we move this out of `shlr`(this is only reached if the xml protocol fails and the default regprofiles available here aren't really complex atm) or try to adopt `profile.c`'s implementation temporarily.

shlr/gdb/src/arch.c
```diff
@@ -115,9 +116,9 @@ gdb_reg_t *arch_parse_reg_profile(const char *reg_profile) {
 				rz_list_append(gdb_regs_list, reg);
 			}
 			// Clean up
```
yossizap: You forgot the comment up here

shlr/gdb/src/arch.c
ghost: I will revert this commit. Thanks

ret2libc (approved):

PR #664, opened 2021-02-20
asm.flags.middle: Don't split bb instruction

kazarmy:
**Your checklist for this pull request**
- [X] I've read the [guidelines for contributing](https://github.com/rizinorg/rizin/blob/master/DEVELOPERS.md) to this repository
- [X] I made sure to follow the project's [coding style](https://github.com/rizinorg/rizin/blob/master/DEVELOPERS.md#code-style)
- [X] I've added tests that prove my fix is effective or that my feature works (if possible)
- [ ] I've updated the documentation and the [rizin book](https://github.com/rizinorg/book) with the relevant information (if needed)

**Detailed description**

This pr fixes a TODO in the "basic block overlaps function start" test in the `analysis/x86_64` testfile (strictly speaking, it replaces the TODO with another that is less problematic), by not allowing flags to split basic block instructions.

**Test plan**

All builds are green.

**Closing issues**

...

PR #667, opened 2021-02-20
handleMidFlags: Reset ds->midflags on entry

kazarmy:

**Detailed description**

This pr resets `ds->midflags` on entry into `handleMidFlags()` and prevents further propagation down the disassembly of `ds->midflags` changes in `handleMidFlags()` (e.g. https://github.com/rizinorg/rizin/commit/a0f1511a8cfa6480f16c817c145fb83781a85daf#diff-ae53298c5641d2c8d7ccac1ebcc8982a969f4075a440962e6f16f1ffee8ceb23).

**Test plan**

All builds are green.

**Closing issues**

...

### DIFF
--- a/test/db/cmd/midflags
+++ b/test/db/cmd/midflags
@@ -395,3 +395,25 @@ EXPECT=<<EOF
   0x00000009      90             nop
 EOF
 RUN
+
+NAME=asm.flags.middle reset
+FILE=-
+CMDS=<<EOF
+e asm.arch=x86
+e asm.bytes=true
+e asm.bits=64
+e asm.bb.middle=false
+e asm.flags.middle=2
+wx b821c10010ebf9b8210011c1
+af @ 3
+f sym.dummy @ 9
+pd 4 @ 0
+EOF
+EXPECT=<<EOF
+|           0x00000000  ~   b821c10010     mov   eax, 0x1000c121
+|           0x00000005      ebf9           jmp   0
+            ;-- sym.dummy:
+            0x00000007  ~   b8210011c1     mov   eax, 0xc1110021       ; '!'
+            0x0000000c      0000           add   byte [rax], al
+EOF
+RUN
--- a/test/db/cmd/midflags
+++ b/test/db/cmd/midflags
@@ -412,8 +412,8 @@ EOF
 EXPECT=<<EOF
 |           0x00000000  ~   b821c10010     mov   eax, 0x1000c121
 |           0x00000005      ebf9           jmp   0
-            ;-- sym.dummy:
             0x00000007  ~   b8210011c1     mov   eax, 0xc1110021       ; '!'
-            0x0000000c      0000           add   byte [rax], al
+            ;-- sym.dummy:
+            0x00000009      0011           add   byte [rcx], dl
 EOF
 RUN
--- a/librz/core/disasm.c
+++ b/librz/core/disasm.c
@@ -1530,6 +1530,7 @@ static void ds_atabs_option(RDisasmState *ds) {
 }
 
 static int handleMidFlags(RzCore *core, RDisasmState *ds, bool print) {
+	ds->midflags = rz_config_get_i(core->config, "asm.flags.middle");
 	ds->hasMidflag = false;
 	if (ds->midcursor && core->print->cur != -1) {
 		ut64 cur = core->offset + core->print->cur;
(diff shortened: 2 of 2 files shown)


XVilka (approved):

test/db/cmd/midflags
```diff
@@ -395,3 +395,25 @@ EXPECT=<<EOF
...
+e asm.bits=64
+e asm.bb.middle=false
+e asm.flags.middle=2
+wx b821c10010ebf9b8210011c1
+af @ 3
+f sym.dummy @ 9
+pd 4 @ 0
+EOF
+EXPECT=<<EOF
+|           0x00000000  ~   b821c10010     mov   eax, 0x1000c121
```
ret2libc, line 413: Why is this better than the current output:
```
[0x00000000]> pd 4 @ 0
│           ; CODE XREF from fcn.00000003 @ 0x5
│           0x00000000  ~   b821c10010     mov   eax, 0x1000c121
┌ 9: fcn.00000003 ();
│ bp: 0 (vars 0, args 0)
│ sp: 0 (vars 0, args 0)
│ rg: 0 (vars 0, args 0)
│           0x00000003      0010           add   byte [rax], dl
│           0x00000005      ebf9           jmp   0
            0x00000007  ~   b8210011c1     mov   eax, 0xc1110021       ; '!'
```
??

kazarmy: Because `asm.flags.middle=2` aka realign on flag and there is a flag at address 9.

ret2libc: Isn't there a flag also at address 3? In particular, the user has even said there is a function starting at 3 and the new output in your PR hides that.

kazarmy: It's #664 that hides the function at 3. Why? Because realigning on 3 breaks graph disassembly output (see #664's test) by showing an instruction that is not executed by the CPU.

Due to the limitations of a linear disassembly view, you can still realign on 3 by setting `asm.bb.middle=true`.

kazarmy: You might argue that graph view should set `asm.flags.middle=1`. But then we can get into a situation where the linear view realigns at 3 regardless of the value of `asm.bb.middle`.

Imo, flags should not be able to break up the instruction flow in a basic block. Perhaps the only reason you want to do that is if the instructions are actually hiding something else, but right now (if I'm reading the code correctly), both the linear and graph disassembly will realign on a `str.` flag regardless.

That the flag at 3 is hidden unfortunately appears to be a bug.

kazarmy: > ... both the linear and graph disassembly will realign on a `str.` flag regardless.

Except on `asm.flags.middle=0` that is.

ret2libc: :/ you lost me! 

Who said what is executing and what is not? I mean, are you assuming that execution starts at 0?

I guess that in linear view there is no 100% perfect way to show these kind of situations, right?

kazarmy: > :/ you lost me!

Sorry.

> Who said what is executing and what is not? I mean, are you assuming that execution starts at 0?

The test does an `af @ 3`. There is a basic block starting from 0x0.

> I guess that in linear view there is no 100% perfect way to show these kind of situations, right?

Yes, but you can make it easy to toggle between views.

ret2libc: > > Who said what is executing and what is not? I mean, are you assuming that execution starts at 0?
> 
> The test does an `af @ 3`. There is a basic block starting from 0x0.

Right, so essentially with this new PR you preferred the bb at 0x0 instead of the one at 0x3, right? Why should it be better to show the one at 0 rather than the one at 3? Maybe We should just find a better way to clearly show both paths? Just saying, i'm not sure yet how.

kazarmy: > Right, so essentially with this new PR you preferred the bb at 0x0 instead of the one at 0x3, right? Why should it be better to show the one at 0 rather than the one at 3?

You can show the one at 3. Just set `asm.bb.middle=true` and by the way, this is the default value of `asm.bb.middle`. This PR has nothing to do with whether 0 or 3 is preferred.

> Maybe We should just find a better way to clearly show both paths?

Just do `agf` or `VV`.

ret2libc: Ok!

ret2libc (approved):

test/db/cmd/midflags
kazarmy: Tq! :tada::scream::joy::dancer:

PR #668, opened 2021-02-20
Fix memleak and various null derefs

wargio:
# DO NOT SQUASH ME

### DIFF
--- a/librz/util/print_code.c
+++ b/librz/util/print_code.c
@@ -95,9 +95,7 @@ static void print_c_code(RzPrint *p, ut64 addr, const ut8 *buf, int len, int ws,
 }
 
 RZ_API void rz_print_code(RzPrint *p, ut64 addr, const ut8 *buf, int len, char lang) {
-	if (!p || !buf) {
-		return;
-	}
+	rz_return_if_fail(p && buf);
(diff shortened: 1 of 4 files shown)


librz/util/print_code.c
```diff
@@ -95,6 +95,9 @@ static void print_c_code(RzPrint *p, ut64 addr, const ut8 *buf, int len, int ws,
 }
 
 RZ_API void rz_print_code(RzPrint *p, ut64 addr, const ut8 *buf, int len, char lang) {
+	if (!p || !buf) {
```
ret2libc: Use `rz_return_val_if_fail` here. If you are passing a NULL RzPrint you are doing something wrong, so better be warned than just seeing nothing happening maybe later.

librz/util/print_code.c
```diff
@@ -95,6 +95,9 @@ static void print_c_code(RzPrint *p, ut64 addr, const ut8 *buf, int len, int ws,
...
+		return;
+	}
```
wargio:
```suggestion
	rz_return_if_fail(p && buf);
```

ret2libc (approved):

PR #669, opened 2021-02-20
Disable function size by default ##breaking

ITAYC0HEN:

**Detailed description**

Until now, when disassembling a function or the top-part of it, Rizin will print the size of the function near the signature.
For example "44" in the first line here:

```
┌ 44: int main (char **argv);
│           ; var int32_t var_4h @ ebp-0x4
│           ; arg char **argv @ esp+0x24
│           0x000011cd      lea   ecx, [argv]
│           0x000011d1      and   esp, 0xfffffff0
 . . . 
 . . .
```

This pull request introduces a new configuration variable named `asm.fcn.size` that will be disabled by default.
In addition, it changes the name of the `asm.fcnsig` to `asm.fcn.signature` because Developers' lingo isn't users' lingo.

**Test plan**

disassemble using `pdf` and `pdf` as well; as graph, and observe how the `asm.fcn.size` changes when disabled and enabled.


ret2libc (approved):

kazarmy (approved):

Maijin (approved):

XVilka (changes requested):
I suggest to convert only half of the tests to the default output, so we test both modes, just in case. Apart from that LGTM.

Maijin:
No I would leave like this, but add a test for each options, just copy/pasta and add the eval and change the name. Tests are supposed to test one and one feature only, not multiple one's 👍

XVilka:
@Maijin majority of tests are "integration" tests since involve many aspects of the Rizin, especially in `analysis` category.

Maijin:
yeah but again, I think it'd be better to just copy/paste one or few of those and add an eval true/false to them to test both  functionnalities instead separately.

officialcjunior:
And, I'll handle documenting this new variable on the book. 👍🏽

test/db/cmd/cmd_pd
```diff
@@ -642,7 +642,7 @@ e asm.flags=false
 s sym.func.100004401
 af
 afn newname
-shu
+sHu
```
ITAYC0HEN:
```suggestion
```

ITAYC0HEN:
Would love help with fixing the last failing tests. Not sure what's going on in there

XVilka:
@ITAYC0HEN you forgot to add a test with the size enabled. Please do this in a separate PR.

test/db/cmd/cmd_pd
```diff
@@ -675,7 +675,38 @@ e asm.lines.fcn=false
...
+CMDS=<<EOF
+e asm.bytes=true
+e asm.lines.fcn=true
+s 0x004017c0
+af+ 0x004017c0 fcn2.0x004017c0
+afb+ 0x4017c0 0x4017c0 13 0x4017dd 0x4017cd
+pdf
+?e
+e asm.lines.fcn=false
+e asm.fcn.size=true
```
ITAYC0HEN, line 705: @XVilka I did here
I can add a few more of needed

XVilka: Ah, didn't notice. No worries then, should be fine.

Issue #569, opened 2021-02-10
Java bytecode disassembly/analysis/binary parsing code refactoring


XVilka:
The existing code of the JVM bytecode disassembly, analysis, and binary parsing is truly terrible and not very well integrated with the rest of Rizin code. I propose to:
- [x] Remove the dsojson.c and use RzUtil functions
- [x] Write tests for different JVM bytecode versions in our testsuite
- [x] Refactor the code to reduce duplication, maybe move it out of the `shlr/` directory
- [ ] Compare the results with other Java reversing tools
- [ ] Provide better integration of the same features with the DEX analysis
- [x] Various bugfixes and improvements along the way

ret2libc:
What about moving it out of rizin repository into extras until someone show proper interest in making this really better?

wargio:
Remove DsoJson here: https://github.com/rizinorg/rizin/pull/585

XVilka:
@wargio @ret2libc lets close this issue since majority of it had been done already. The only left part is assembly, right?
And maybe better integration with DEX

PR #671, opened 2021-02-21
Move java from shlr to core 

wargio:
# DO NOT SQUASH ME

**Detailed description**

- Splitted files, to have a better code handling (like stdout and java output from bin).
- Removed also dead code
- Removed dup code and used already tested and working methods like `rz_read_at_be32` instead of `UINT|USHORT` etc..
- Removed `yc_w_refs` and `i_mref` which never worked.
- Fixed `set_flags` method to work as intended.
- Fixed various parsing errors.
- Fixed many memory leaks.
- Refactoring core plugin with a proper example

Partially addresses https://github.com/rizinorg/rizin/issues/569


librz/asm/arch/java/code.c
```diff
@@ -122,7 +118,7 @@ RZ_API int java_print_opcode(RzBinJavaObj *obj, ut64 addr, int idx, const ut8 *b
 	}
 
 #if 0
```
XVilka: Remove `if 0`

librz/asm/arch/java/code.c
```diff
@@ -11,13 +11,9 @@
 #include <stdarg.h>
 #include "ops.h"
 #include "code.h"
-#include "class.h"
 
 #define V if (verbose)
```
XVilka: Maybe kill `V` too?

XVilka (changes requested):

Maijin:
As you have worked on this and therefore should better understand the commands, do you think you could add tests for the Java commands or at least open a follow-up issue with the commands that needs to be tested so that's covered?

wargio:
it's here the issue @Maijin https://github.com/rizinorg/rizin/issues/569
the next following steps is actually add tests, and fix all the issues.

librz/meson.build
```diff
@@ -197,9 +197,7 @@ bp_plugins = [
   'x86'
 ]
 
-core_plugins = [
-  'java'
-]
+core_plugins = []
```
ret2libc: as said, I think this is wrong. IMO we should keep the java core plugin. We have plugins for everything, I don't see why `java` should be a "core" feature of rizin instead of a plugin.

In particular, you could still easiy use newshell without the need to have it outside the plugin architecture. I would register the `java` command inside the `RzCorePlugin.init` method. You would not be able to use the yaml thing anymore, but you can anyway register the java commands from C.

ret2libc (changes requested):

librz/meson.build
wargio: reverted back

XVilka:
Four failed unit tests:
```
13/76 autocmplt            FAIL           0.41s (exit status 1)

--- command ---
23:19:18 /usr/home/build/rizin/build/test/unit/test_autocmplt
--- stdout ---
test_autocmplt_cmdid OK
test_autocmplt_newcommand ERR
[XX] Fail at line 139: there are 4 commands available: expected 4, got 5.

test_autocmplt_argid OK
test_autocmplt_quotedarg OK
test_autocmplt_newarg OK
test_autocmplt_fcn OK
test_autocmplt_eval OK
test_autocmplt_seek OK

--- stderr ---
-------

20/76 cmd                  FAIL           0.02s (exit status 1)

--- command ---
23:19:19 /usr/home/build/rizin/build/test/unit/test_cmd
--- stdout ---
test_parsed_args_noargs OK
test_parsed_args_onearg OK
test_parsed_args_args OK
test_parsed_args_nospace OK
test_parsed_args_newcmd OK
test_parsed_args_newargs OK
test_cmd_descriptor_argv ERR
[XX] Fail at line 102: root has 1 child: expected 1, got 2.

test_remove_cmd ERR
[XX] Fail at line 469: root has 2 commands as children: expected 2, got 3.

test_cmd_args OK
test_cmd_argv_modes ERR
[XX] Fail at line 539: zj, zJ and zq are considered in the help: expected Usage: [.][times][cmd][~grep][@[@iter]addr!size][|>pipe] ; ...
| z[jqJ] # z summary
, got Usage: [.][times][cmd][~grep][@[@iter]addr!size][|>pipe] ; ...
| java <subcmd> [<arg0> [<arg1> ...]] # Suite of java commands
| z[jqJ]                     # z summary
.

test_cmd_group_argv_modes ERR
[XX] Fail at line 585: zd is considered in the help: expected Usage: [.][times][cmd][~grep][@[@iter]addr!size][|>pipe] ; ...
| z[jqd] # z group summary
, got Usage: [.][times][cmd][~grep][@[@iter]addr!size][|>pipe] ; ...
| java <subcmd> [<arg0> [<arg1> ...]] # Suite of java commands
| z[jqd]                     # z group summary
.

test_foreach_cmdname ERR
[XX] Fail at line 625: count regular commands that can be executed: expected 9, got 10.
```

librz/asm/meson.build
```diff
@@ -224,8 +226,8 @@ rz_asm = library('rz_asm', rz_asm_sources,
     rz_syscall_dep,
     rz_flag_dep,
     rz_parse_dep,
+    rz_bin_dep,
```
ret2libc, line 229: why is `rz_bin_dep` needed? Could you try compiling it without this dependency?
I'm not a big fan of having asm depending on bin, if not really necessary.

librz/core/p/core_java.c
```diff
@@ -1938,12 +1598,66 @@ static bool rz_cmd_java_handle_print_exceptions(RzCore *core, const char *input)
 	return true;
 }
 
+RZ_IPI RzCmdStatus rz_cmd_java_handler(RzCore *core, int argc, const char **argv) {
+	bool res = false;
+	for (int i = 0; JAVA_CMDS[i].name; i++) {
+		if (!strncmp(argv[1], JAVA_CMDS[i].name, JAVA_CMDS[i].name_len)) {
+			res = JAVA_CMDS[i].handler(core, argc - 2, argc > 2 ? &argv[2] : NULL);
```
ret2libc: It would be awesome if the sub-handlers could directly return a RzCmdStatus, so for example you can return `RZ_CMD_STATUS_WRONG_ARGS` when the wrong number of arguments is passed to the sub-command.

ret2libc (commented):
Wow... quite a PR :D

librz/core/p/core_java.c
wargio: yes, i was thinking to do this. but then i just adapted the old ones to the current one. i will fix this tonight.

librz/asm/meson.build
wargio: yes it is, because the java asm relies on bin `class.c`.

ret2libc: :( :(

wargio: i think i can fix this later.

XVilka (approved):
I think decoupling it from RzBin dependency can be a separate PR. It's already a good improvement of the current state of affairs.

librz/core/p/core_plugin_example.c
```diff
@@ -0,0 +1,69 @@
+// SPDX-License-Identifier: LGPL-3.0-only
```
ret2libc, line 1: is this file necessary?

librz/core/p/core_java.c
```diff
@@ -1935,15 +1624,67 @@ static bool rz_cmd_java_handle_print_exceptions(RzCore *core, const char *input)
...
+	{
+		.name = "arg",
+		.type = RZ_CMD_ARG_TYPE_STRING,
+		.flags = RZ_CMD_ARG_FLAG_ARRAY,
+		.optional = true,
+	},
+	{ 0 },
+};
+
+static const RzCmdDescHelp cmd_java_help = {
```
ret2libc, line 1641: I think we would want in the future to have the "long help" to show also the type of arguments and their possible values (e.g. in the case of `java` command it would show that `subcmd` can be any of `help`, `set_flags`, etc.). WDYT?

librz/core/p/core_java.c
```diff
@@ -1935,15 +1624,67 @@ static bool rz_cmd_java_handle_print_exceptions(RzCore *core, const char *input)
...
+	}
+	return RZ_CMD_STATUS_NONEXISTINGCMD;
+}
+
+static const char *cmd_java_subcmd_choices[] = { "help", "set_flags", "prototypes", "resolve_cp", "calc_flags", "flags_str_at", "flags_str", "m_info", "f_info", "find_cp_const", "find_cp_value", "replace_cp_value", "replace_classname_value", "reload_bin", "summary", "exc", "calc_sz", "is_valid", NULL };
+static const RzCmdDescArg cmd_java_args[3] = {
+	{
+		.name = "subcmd",
+		.type = RZ_CMD_ARG_TYPE_CHOICES,
+		.optional = false,
```
ret2libc: I couldn't find an easy way to understand what subcommands were accepted. I suggest you either:
- make this argument optional and assume `help` as subcommands when nothing is specified. You could make this clear by also using the `default_value` field.
- provide a `description` in the `java` cmd descriptor to describe at least how to get additional help.

librz/core/p/core_java.c
```diff
@@ -1935,15 +1624,67 @@ static bool rz_cmd_java_handle_print_exceptions(RzCore *core, const char *input)
...
+	return RZ_CMD_STATUS_OK;
+}
+
+RZ_IPI RzCmdStatus rz_cmd_java_handler(RzCore *core, int argc, const char **argv) {
+	for (int i = 0; i < END_CMDS; i++) {
+		if (!strncmp(argv[1], JAVA_CMDS[i].name, JAVA_CMDS[i].name_len)) {
+			return JAVA_CMDS[i].handler(core, argc - 2, argc > 2 ? &argv[2] : NULL);
+		}
+	}
+	return RZ_CMD_STATUS_NONEXISTINGCMD;
```
ret2libc: this should probably return `WRONG_ARGS`, otherwise executing `java asdasd` returns `Command 'java' does not exist.` which is confusing.

librz/core/p/core_plugin_example.c
wargio: yes, this is the example file for core. i just tested and made sure it does compile. we could move this later into the book

librz/core/p/core_java.c
wargio: oh. i'll fix this tonight.

librz/core/p/core_plugin_example.c
wargio: this file is not compiled.

librz/core/p/core_java.c
```diff
@@ -1935,15 +1605,72 @@ static bool rz_cmd_java_handle_print_exceptions(RzCore *core, const char *input)
 			free(class_info);
 		}
 	}
-	return true;
+	return RZ_CMD_STATUS_OK;
+}
+
+RZ_IPI RzCmdStatus rz_cmd_java_handler(RzCore *core, int argc, const char **argv) {
+	if (argc < 2) {
```
ret2libc, line 1612: since you have made the `subcmd` argument optional, you should handle it here.

ret2libc (changes requested):

librz/core/p/core_java.c
wargio: when you exec `java` newshell will only provide `argv[][]= {{"java"}}` and argc is `1`.

ret2libc (approved):

ret2libc:
Approved because the `default_value` will be handled by the parser itself.

PR #687, opened 2021-02-22
core: split cmd_descs.yaml in smaller YAML files

### LINKED ISSUE #656: Split cmd_descs.yaml in sub-files
When we created cmd_descs.yaml we were aware it could become too big. To ease contributions and make it easier to read for humans, it is probably come the time to split it in separate files.

It would be nice to have `cmd_analysis.yaml`, `cmd_flags.yaml`, etc. As suggested by @karliss we could have a `parent` field at the root of the sub-files that specify where in the main file they should be placed.

-

**Additional context**
This would only change the way cmd_descs_generate.py works, we can still create one huge cmd_descs.c/cmd_descs.h file with all the info merged together.

ret2libc:
This is based on https://github.com/rizinorg/rizin/issues/656, so wait for that before merging this.

**Detailed description**

This PR splits the command descriptors in multiple files so that it becomes easier to modify the file and there are less levels of indentation. Some files like cmd_analysis could be split even more in the future. The way I did it is that each yaml file starts with:
```yaml
---
name: <name>
commands:
  .... usual list of commands ...
```

Then a command that has some subcommands could provide a string instead of a list of subcommands:
```yaml
- name: =
  summary: remote handling
  subcommands: cmd_remote
```

In this case, the YAML file with `name` equal to `cmd_remote` will be used to load the subcommands. All YAML files need to be passed to cmd_descs_generate.py and meson consider all of them as dependencies of that build target, so if any of those is changed, the cmd_descs.c file is recomputed.

**Test plan**

All builds as usual.

**Closing issues**

Closes https://github.com/rizinorg/rizin/issues/656


wargio:
tbh, i prefer 1 file

XVilka:
@wargio we converted only a small part of the Rizin commands and the file is already big. Imagine how huge it will become once everything is converted. Strong 👍 for the split.

XVilka (changes requested):
Since there will be many different files, what about putting them into the separate directory?
`librz/core/yaml/` or `librz/core/cmd_desc/` or something similar?

XVilka (approved):
Overall is good, but you need to fix formatting.

thestr4ng3r (approved):
It shouldn't be split too fine-grained but this kind of module-like one like the c files are is good I think.

kazarmy (approved):
Size does matter here, but even if the file size is too small, there's always `grep`.

PR #700, opened 2021-02-24
Migrating types command to the newshell

XVilka:
# **DO NOT SQUASH**

**Detailed description**

Migrating `t` commands to the newshell and fixes along the way.

**Test plan**

All green


librz/core/cmd_descs/cmd_descs.yaml
```diff
@@ -254,9 +254,8 @@ commands:
     summary: Seek commands
     subcommands: cmd_seek
   - name: t
-    cname: cmd_type
```
ret2libc, line 257: I think you should keep both `cname` and `type` here if you want the oldinput `t` handler to keep working.

wargio:
that `rz_type_list_c_nl_handler` is a horror, but is ok. no need to fix that now.
i think you can keep also oldshell, so then we can clean the code once we have converted all the commands

librz/core/cmd_type.c
```diff
@@ -1005,6 +1085,59 @@ RZ_API void rz_core_link_stroff(RzCore *core, RzAnalysisFunction *fcn) {
...
+	}
+	ls_free(l);
+}
+
+static void type_list_c_all(RzCore *core) {
+	Sdb *TDB = core->analysis->sdb_types;
+	// TODO: Change the logic maybe?
+	//eprintf("Specify the type");
+	//return RZ_CMD_STATUS_ERROR;
+	rz_core_cmd0(core, "tfc");
```
XVilka, line 1118: I will convert these `rz_core_cmd*()` calls to the API once all those converted to the newshell and new API.

librz/core/canalysis.c
```diff
@@ -7403,3 +7403,87 @@ RZ_IPI void rz_core_analysis_value_pointers(RzCore *core, RzOutputMode mode) {
 	free(tmp);
 	rz_config_set_i(core->config, "search.align", o_align);
 }
+
+static bool __nonreturn_print_commands(void *p, const char *k, const char *v) {
+	RzCore *core = (RzCore *)p;
+	if (!strncmp(v, "func", strlen("func") + 1)) {
+		char *query = sdb_fmt("func.%s.noreturn", k);
+		if (sdb_bool_get(core->analysis->sdb_types, query, NULL)) {
+			rz_cons_printf("tnn %s\n", k);
```
ret2libc, line 7412: to be honest these kind of things should be wrapped with `rz_cmd_escape_arg`. But it should have been done in manny places I guess...

librz/core/cmd_descs/cmd_type.yaml
```diff
@@ -0,0 +1,130 @@
...
+    args:
+      - name: type
+        type: RZ_CMD_ARG_TYPE_STRING
+        optional: true
+  - name: tn
+    summary: Manage noreturn function attributes and marks
+    subcommands:
+      - name: tn
+        cname: type_list_noreturn
+        summary: List all noreturn references
```
ret2libc:
```suggestion
        summary: List all noreturn references / Add a noreturn function.
```

librz/core/cmd_descs/cmd_type.yaml
```diff
@@ -0,0 +1,130 @@
...
+          - name: name
+            type: RZ_CMD_ARG_TYPE_STRING
+            optional: true
+      - name: tn-
+        cname: type_noreturn_del
+        summary: Remove the noreturn reference
+        args:
+          - name: name
+            type: RZ_CMD_ARG_TYPE_STRING
+            flags: RZ_CMD_ARG_FLAG_ARRAY
```
ret2libc, line 79: actually this is a bit out of place... I don't see why `tn-` accepts a list, but most other `-` commands don't. Can't we just have one arg only even here?

librz/core/cmd_descs/cmd_type.yaml
```diff
@@ -0,0 +1,130 @@
...
+      - name: tn-*
+        cname: type_noreturn_del_all
+        summary: Remove all noreturn references
+        args: []
+  - name: tt
+    summary: List loaded typedefs
+    subcommands:
+      - name: tt
+        cname: type_list_typedef
+        summary: List loaded typedefs
```
ret2libc:
```suggestion
        summary: List loaded typedefs / Show name for given type alias
```

librz/core/cmd_descs/cmd_type.yaml
```diff
@@ -0,0 +1,130 @@
...
+        args:
+          - name: type
+            type: RZ_CMD_ARG_TYPE_STRING
+            optional: true
+  - name: tu
+    summary: List loaded unions
+    subcommands:
+      - name: tu
+        cname: type_list_union
+        summary: List loaded unions
```
ret2libc:
```suggestion
        summary: List loaded unions / Show pf format string for given union
```

librz/core/cmd_type.c
```diff
@@ -208,7 +198,7 @@ static void __core_cmd_tcc(RzCore *core, const char *input) {
 		pj_free(pj);
 		rz_list_free(list);
 	} break;
-	case 'l': {
+	case RZ_OUTPUT_MODE_LONG: {
```
ret2libc, line 201: I don't see any command that you converted that accept the mode `RZ_OUTPUT_MODE_LONG`.

librz/core/cmd_type.c
```diff
@@ -1735,3 +1822,162 @@ RZ_IPI int rz_cmd_type(void *data, const char *input) {
 	}
 	return true;
 }
+
+RZ_IPI RzCmdStatus rz_type_cc_list_handler(RzCore *core, int argc, const char **argv, RzOutputMode mode) {
+	const char *cc = argc > 1 ? argv[1] : NULL;
+	if (cc) {
+		types_cc_print(core, cc);
```
ret2libc: considering this can be called also with `tccj <cc>` i might expect it to return JSON as well.

librz/core/cmd_type.c
```diff
@@ -1735,3 +1822,162 @@ RZ_IPI int rz_cmd_type(void *data, const char *input) {
...
+		types_cc_print_all(core, mode);
+	}
+	return RZ_CMD_STATUS_OK;
+}
+
+RZ_IPI RzCmdStatus rz_type_cc_del_handler(RzCore *core, int argc, const char **argv) {
+	const char *cc = argc > 1 ? argv[1] : NULL;
+	if (!cc) {
+		eprintf("Specify the calling convention");
+		return RZ_CMD_STATUS_ERROR;
```
ret2libc: WRONG_ARGS is better

ret2libc: Also, why are you checking? In general, the parser will automatically check for the right number of arguments

librz/core/cmd_type.c
```diff
@@ -1735,3 +1822,162 @@ RZ_IPI int rz_cmd_type(void *data, const char *input) {
...
+RZ_IPI RzCmdStatus rz_type_list_c_nl_handler(RzCore *core, int argc, const char **argv) {
+	type_list_c_all_nl(core);
+	return RZ_CMD_STATUS_OK;
+}
+
+RZ_IPI RzCmdStatus rz_type_define_handler(RzCore *core, int argc, const char **argv) {
+	const char *type = argc > 1 ? argv[1] : NULL;
+	if (!type) {
+		eprintf("Missing type definition");
+		return RZ_CMD_STATUS_ERROR;
```
ret2libc: same as before... WRONG_ARGS and is it really necessary?

librz/core/cmd_type.c
```diff
@@ -1735,3 +1822,162 @@ RZ_IPI int rz_cmd_type(void *data, const char *input) {
...
+RZ_IPI RzCmdStatus rz_type_kuery_handler(RzCore *core, int argc, const char **argv) {
+	const char *query = argc > 1 ? argv[1] : NULL;
+	Sdb *TDB = core->analysis->sdb_types;
+	char *output = NULL;
+	if (query) {
+		output = sdb_querys(TDB, NULL, -1, query);
+	} else {
+		output = sdb_querys(TDB, NULL, -1, "*");
+	}
+	if (output) {
```
ret2libc: I think it's better to have `if (!output) { eprintf("some error msg"); return ERROR;} rz_cons_print (....); return OK;`

librz/core/cmd_type.c
```diff
@@ -1735,3 +1822,162 @@ RZ_IPI int rz_cmd_type(void *data, const char *input) {
...
+	} else {
+		rz_core_analysis_noreturn_print(core, mode);
+	}
+	return RZ_CMD_STATUS_OK;
+}
+
+RZ_IPI RzCmdStatus rz_type_noreturn_del_handler(RzCore *core, int argc, const char **argv) {
+	if (argc == 1) {
+		eprintf("Specify the name or the address");
+		return RZ_CMD_STATUS_ERROR;
```
ret2libc: same as before, WRONG_ARGS and is it really necessary?

librz/core/cmd_type.c
```diff
@@ -1735,3 +1822,162 @@ RZ_IPI int rz_cmd_type(void *data, const char *input) {
...
+	}
+	for (int i = 1; i < argc; i++) {
+		noreturn_del(core, argv[i]);
```
ret2libc: calling `noreturn_del` as is is wrong, because that function splits the input by spaces... We don't need those kind of things in newshell, also considering that an argument in newshell could even contain spaces.

librz/core/cmd_type.c
```diff
@@ -1735,3 +1822,162 @@ RZ_IPI int rz_cmd_type(void *data, const char *input) {
...
+	}
+	return RZ_CMD_STATUS_OK;
+}
+
+RZ_IPI RzCmdStatus rz_type_noreturn_del_all_handler(RzCore *core, int argc, const char **argv) {
+	RzList *noretl = rz_core_analysis_noreturn(core);
+	RzListIter *iter;
+	char *name;
+	rz_list_foreach (noretl, iter, name) {
+		noreturn_del(core, name);
```
ret2libc: same

ret2libc (changes requested):
Do you know if some of these commands is not covered at all by tests?

librz/core/canalysis.c
XVilka: Good point, I will add it in my TODO. But it's out of the scope for the current PR, imho.

librz/core/cmd_descs/cmd_type.yaml
XVilka: In theory, yes. In practice other parts of the Rizin rely on the `tn` and `tn-` commands and call them with multiple arguments.

librz/core/cmd_type.c
XVilka: Good point, added. `tccl` shoudl work now.

XVilka:
@ret2libc addressed everything, I think there are tests for every `t` command converted and I also checked them manually. The only two missing comments are completely a valid point but a bit out of the scope of this PR and will be addressed later, once we finish `t` commands conversion and I will start on RzTypes API extraction.

librz/core/canalysis.c
XVilka: I was also thinking about creating some kind of small API like PJ for forming Rizin commands, and the same for the C-like output.

ret2libc: I think that should be discussed thoroughly  before starting with an implementation because I think it's something we should use in many places... I honestly don't like how PJ is always passed everywhere, because it makes it kinda "special" and you have to deal with it in too many places. `RzTable` might be a good step toward having a unified interface for output, but as we said in the past, many times the format of Json/table/simple output is too different... We should try to think about a very good solution.

librz/core/cmd_descs/cmd_type.yaml
ret2libc: I'm ok to not do this in this PR, but IMHO we should be consistent. Either all `-` commands accept a list, or none.

librz/core/cmd_type.c
```diff
@@ -219,46 +209,93 @@ static void __core_cmd_tcc(RzCore *core, const char *input) {
...
-	case 'k':
+	case RZ_OUTPUT_MODE_SDB:
 		rz_core_kuery_print(core, "analysis/cc/*");
 		break;
-	case ' ':
-		if (strchr(input, '(')) {
-			if (!rz_analysis_cc_set(core->analysis, input + 1)) {
-				eprintf("Invalid syntax in cc signature.");
+	default:
+		break;
```
ret2libc, line 228: add a `rz_warn_if_reached();`

ret2libc (approved):

librz/core/cmd_type.c
XVilka: Done, good idea.

librz/core/canalysis.c
XVilka: Created an issue for that to track and discuss. Feel free to edit/improve: https://github.com/rizinorg/rizin/issues/727

PR #706, opened 2021-02-24
Show overlapped flags if requested and show them differently

kazarmy:

**Detailed description**

This pr restores the showing of flags that are overlapped or _inside_ an instruction. Also, the flags are shown differently.

**Test plan**

All builds are green.

**Closing issues**

...


XVilka (changes requested):
Could you please also show them with a color different from the one that is used by normal flags?

kazarmy:
> Could you please also show them with a color different from the one that is used by normal flags?

Yes but it's tricky for at least the following reasons:

1. There is no color in the `ec` color palette for these kinds of flags.
2. Flags can actually have their own individual color through `fc`.

There may be other reasons that I've missed.

What I've done e7cdac1 is that I've colored the offset differently. I hope that will suffice:

overlapped_flag_color

XVilka (approved):

PR #709, opened 2021-02-24
Convert `dsu` commands to newshell

officialcjunior:

**Detailed description**

Convert the commands under `dsu`: `dsu`, `dsui`, `dsuir`, `dsue`, `dsuf`, `dsuo` to newshell.
Introduced `dbg_follow_seek_register()` to seek PC while dealing with `dbg.follow` in cmd_debug.c

**Test plan**

- [x] `dsuo` (which doesn't have tests yet) supposedly handles a number of opcodes as arguments. Not sure whether the current implementation handles that.

- [x] `dsu`, `dsue`, `dsui[r]` has written tests. Needs to add tests for the rest, when the implementation part is correct.

**Closing issues**

None.


librz/core/cmd_descs/cmd_debug.yaml
```diff
@@ -4,4 +4,47 @@ commands:
...
+    type: RZ_CMD_DESC_TYPE_OLDINPUT
+  - name: ds
+    summary: Debug step commands
+    subcommands:
+      - name: dsu
+        summary: Debug step until commands
+        subcommands:
+          - name: dsu
+            cname: cmd_debug_step_until
+            summary: Step until <address>
```
```suggestion
            summary: Step until <addr>
```

librz/core/cmd_descs/cmd_debug.yaml
```diff
@@ -4,4 +4,47 @@ commands:
...
+            args:
+            - name: addr
+              type: RZ_CMD_ARG_TYPE_RZNUM
+              flags: RZ_CMD_ARG_FLAG_LAST
```
ret2libc: no need for this, i've changed cmd_descs_generate.py to automatically set it when appropriate.

librz/core/cmd_descs/cmd_debug.yaml
```diff
@@ -4,4 +4,47 @@ commands:
...
+            summary: Step until <esil> expression matches
+            args:
+            - name: esil
+              type: RZ_CMD_ARG_TYPE_STRING
+          - name: dsuf
+            cname: cmd_debug_step_until_flag
+            summary: Step until pc == <flag> matching name
+            args:
+            - name: flag
+              type: RZ_CMD_ARG_TYPE_STRING
```
ret2libc: Please introduce a new type `RZ_CMD_ARG_TYPE_FLAG` for this. It will be useful once we start converting `f` commands as well and when i add autocompletion for flag names. If you use it now, it will be automatically autocompleted later ;)

librz/core/cmd_descs/cmd_debug.yaml
officialcjunior: You were referring to  just the `flags: ` descriptor, right?

ret2libc: yes, just `flags`.

librz/core/cmd_debug.c
```diff
@@ -4219,6 +4219,37 @@ RZ_IPI int rz_debug_continue_oldhandler(void *data, const char *input) {
...
+	step_until_inst(core, argv[1], false);
+	return RZ_CMD_STATUS_OK;
+}
+
+RZ_IPI RzCmdStatus rz_cmd_debug_step_until_instr_regex_handler(RzCore *core, int argc, const char **argv)  {
+	step_until_inst(core, argv[1], true);
+	return RZ_CMD_STATUS_OK;
+}
+
+RZ_IPI RzCmdStatus rz_cmd_debug_step_until_optype_handler(RzCore *core, int argc, const char **argv) {
```
ret2libc, line 4240: I think this should be refactored and the command seems to accept a list of optypes... See inside `step_until_optype`, there is a `rz_str_split_list`. So you should add the `ARRAY` flag to the argument and refactor the `step_until_optype` function to actually accept a list of optypes instead of doing the split itself.

ret2libc (changes requested):

librz/core/cmd_debug.c
officialcjunior: So, should `step_until_optype` accept an `RzList` ?

librz/core/cautocmpl.c
```diff
@@ -182,6 +182,18 @@ static void autocmplt_cmd_arg_macro(RzCore *core, RzLineNSCompletionResult *res,
 	}
 }
 
+static void autocmplt_cmd_arg_flag(RzCore *core, RzLineNSCompletionResult *res, const char *s, size_t len) {
+	RzFlagItem *item;
+	RzListIter *iter;
+	RzList *list = rz_flag_tags_list(core->flags, NULL);
```
ret2libc: These are not `flag_tags`.

Use `rz_flag_all_list`. For now pass `false` to `by_space`... I guess we can split FLAGs later, if needed, in FLAG_INSPACE and FLAG.

librz/core/cmd_debug.c
ret2libc: Yes

officialcjunior: Sorry for the late reply.
What should I use to convert `input`  to an `RzList` so that I can pass to it? That's where I was stuck.

officialcjunior: So, I have pushed a patch (breaking and probably incorrect) so that you can review a lot of other things as well.

librz/core/cmd_analysis.c
```diff
@@ -5043,9 +5043,12 @@ static void cmd_analysis_esil(RzCore *core, const char *input) {
 				case ' ': // "aesu"
 					until_addr = rz_num_math(core->num, input + 2);
 					break;
-				case 'o': // "aesuo"
-					step_until_optype(core, rz_str_trim_head_ro(input + 3));
+				case 'o': { // "aesuo"
+					char *optypes = strdup(rz_str_trim_head_ro((char *)input + 3));
+					RzList *optypes_list = rz_str_split_list(optypes, " ", 0);
```
ret2libc, line 5050: you need to free this list after using it.

librz/core/cmd_analysis.c
```diff
@@ -5043,9 +5043,12 @@ static void cmd_analysis_esil(RzCore *core, const char *input) {
 				case ' ': // "aesu"
 					until_addr = rz_num_math(core->num, input + 2);
 					break;
-				case 'o': // "aesuo"
-					step_until_optype(core, rz_str_trim_head_ro(input + 3));
+				case 'o': { // "aesuo"
+					char *optypes = strdup(rz_str_trim_head_ro((char *)input + 3));
```
ret2libc, line 5049: free this as well.

librz/core/cmd_debug.c
```diff
@@ -4280,9 +4310,12 @@ static int cmd_debug_step(RzCore *core, const char *input) {
 		case 'e': // dsue
 			step_until_esil(core, input + 3);
 			break;
-		case 'o': // dsuo
-			step_until_optype(core, input + 3);
+		case 'o': { // dsuo
+			char *optypes = strdup(rz_str_trim_head_ro((char *)input + 3));
```
ret2libc: free this and next var.

librz/core/cmd_descs/cmd_debug.yaml
```diff
@@ -4,4 +4,47 @@ commands:
   - name: dc
     cname: debug_continue_oldhandler
     summary: Continue execution
-    type: RZ_CMD_DESC_TYPE_OLDINPUT
\ No newline at end of file
+    type: RZ_CMD_DESC_TYPE_OLDINPUT
+  - name: ds
```
ret2libc, line 10: this should probably be of type `RZ_CMD_DESC_TYPE_OLDINPUT`, otherwise you will lose access to all others `ds` commands.

Question: have you try running your branch and checking if the `ds`/`dsu` commands you have replaced still work well? I suggest that when you do such kind of conversion to newshell, you first spend some time understanding how the commands behave before your PR and then check that they still make sense after that.

For example, in this case if you do `ds?` you'd notice that a lot of commands are now missing because this is not OLDINPUT type. Also, if you check `dsuo` you'd see that it is not working as implemented right now. It is important that you test, at least briefly, manually your PRs when doing these kind of changes, to ensure functionalities are still maintained. Unfortunately, automatic tests do not cover everything, so we should pay attention

librz/core/cmd_debug.c
```diff
@@ -4219,6 +4216,39 @@ RZ_IPI int rz_debug_continue_oldhandler(void *data, const char *input) {
...
+	return RZ_CMD_STATUS_OK;
+}
+
+RZ_IPI RzCmdStatus rz_cmd_debug_step_until_instr_regex_handler(RzCore *core, int argc, const char **argv)  {
+	step_until_inst(core, argv[1], true);
+	return RZ_CMD_STATUS_OK;
+}
+
+RZ_IPI RzCmdStatus rz_cmd_debug_step_until_optype_handler(RzCore *core, int argc, const char **argv) {
+	char *optypes = strdup(rz_str_trim_head_ro((char *)argv[0]));
```
ret2libc: As said, this does not work. Please have a look at how `RZ_CMD_ARG_FLAG_ARRAY` is used in other handlers. For example, I suggest looking at cmd_zign.yaml.

librz/core/cmd_debug.c
```diff
@@ -906,7 +904,6 @@ static int step_until_optype(RzCore *core, const char *_optypes) {
 cleanup_after_push:
 	rz_cons_break_pop();
 end:
-	free(optypes);
 	rz_list_free(optypes_list);
```
ret2libc: remove this

officialcjunior:
What's left here is to fix `dsuo`. It's also the only failing test.
I think, when we changed the `char *` to `RzList`, something didn't catch up in `step_until_optype`

ret2libc (dismissed):

test/db/archos/linux-x64/dbg_step
```diff
@@ -189,6 +189,33 @@ EXPECT=<<EOF
...
+0x00000002
+EOF
+RUN
+
+NAME=dbg.dsuo
+FILE=bins/elf/analysis/calls_x64
+ARGS=-a x86 -d
+CMDS=<<EOF
+dsuo jmp
+pi 1~[0]
```
ret2libc:
```suggestion
pi 1 @ rip~[0]
```

ret2libc (approved):
Good work! I think it will work now ;)

ret2libc:
Ok tests are still failing, but that's not @officialcjunior 's fault.
Reason for that is that there is an hack in debug mode that sets `cmd.prompt` to `.dr*`, but `.dr*` is executed only when commands are executed separately.

Please mark the test as BROKEN for now, I will create a new issue with all the details. Thanks for working on this and sorry for the wasted time on another bug.

officialcjunior:
I am not sure whether the code is alright, yet.

Regarding, the seek not moving: previously, the seek moved on the shell, but now on newshell, it doesn't.
```
$ rizin -d test/bins/elf/analysis/calls_x64 
[0x7f097d55e090]> dsuo jmp
[0x7f097d55ee56]> pi 1 @ rip~[0] 
jmp 
```
The seek moved above ^
```
$ ./build/binrz/rizin/rizin -d test/bins/elf/analysis/calls_x64 
[0x7f4996109090]> dsuo jmp
[0x7f4996109090]> pi 1 @ rip~[0]
jmp
```

Is this tied to the same bug, as well?

ret2libc:
> Is this tied to the same bug, as well?

Good point. It isn't.

See how you converted `rz_debug_continue_oldhandler` some time ago. There is a `dbg.follow` thing to respect and probably most debug commands need that before returning. I would extract that in a separate function and make sure you always call it in all newshell debug handlers.

officialcjunior:
Two others tests broke, one is related to the `.dr` thing, which I think can be labelled as broken, but the other one is related to PDB, which is confusing.
If there's something to be improved, I think it'll probably be around places where `follow` was previously used. Hoping to wrap this up in one more commit (if you find something nasty) :slightly_smiling_face:

librz/core/cmd_debug.c
```diff
@@ -832,29 +832,38 @@ static int step_until_inst(RzCore *core, const char *instr, bool regex) {
 	return true;
 }
 
-static int step_until_optype(RzCore *core, const char *_optypes) {
+static void dbg_follow_seek_register(RzCore *core) {
+	int follow = rz_config_get_i(core->config, "dbg.follow");
+	rz_cons_break_pop();
```
ret2libc: move this outside this function, because it is done only in one place, not everywhere.

librz/core/cmd_debug.c
```diff
@@ -832,29 +832,38 @@ static int step_until_inst(RzCore *core, const char *instr, bool regex) {
...
+	if (follow > 0) {
+		ut64 pc = rz_debug_reg_get(core->dbg, "PC");
+		if ((pc < core->offset) || (pc > (core->offset + follow))) {
+			rz_core_cmd0(core, "sr PC");
```
ret2libc: why this instead of `rz_core_seek_to_register` as it originally was?

ret2libc (changes requested):

librz/core/cmd_debug.c
officialcjunior: Sure, alright. You are referring to `rz_cons_break_pop();`, right?
Yeah, I'll remove it and place it at the end of `rz_debug_continue_oldhandler`

ret2libc: yes!

librz/core/cmd_debug.c
officialcjunior: My mistake. I had copied it from the old PR (where rz_core_cmd0 was used), not from the latest `dev`.

officialcjunior:
> Reason for that is that there is an hack in debug mode that sets `cmd.prompt` to `.dr*`, but `.dr*` is executed only when commands are executed separately.
> Please mark the test as BROKEN for now, I will create a new issue with all the details. Thanks for working on this and sorry for the wasted time on another bug.

One last broken test remaining. I hope I can label that as broken?
It doesn't explicitly have a `.dr`, but a `dr` 💭

librz/core/cmd_debug.c
```diff
@@ -5154,11 +5196,6 @@ RZ_IPI int rz_cmd_debug(void *data, const char *input) {
 		rz_core_cmd_help(core, help_msg_d);
 		break;
 	}
-	if (follow > 0) {
-		ut64 pc = rz_debug_reg_get(core->dbg, "PC");
-		if ((pc < core->offset) || (pc > (core->offset + follow))) {
-			rz_core_seek_to_register(core, "PC", false);
-		}
-	}
+	dbg_follow_seek_register(core);
```
ret2libc: I suspect the `pf` test that is failing is due to this. In particular, in the old code `follow` is always set to 0 except in few cases, but with your PR `follow` is always taken from `rz_config_get_i(core->config, "dbg.follow")`.

officialcjunior: You are right, when I tried bringing it back, that test seems to work.
I'll add it back. There's no other hacks to this, is there?

ret2libc: I hope not :D

officialcjunior: Tests pass, except clang-format. And that's because it couldn't fetch something properly. And I hope mine's clearly formatted.

You can probably merge the dev branch and that'll probably fix the failed test.

ret2libc (approved):

Issue #697, opened 2021-02-24
Allow to set the enum value for the operand.


XVilka:
I think it was supported already at some point and was either removed or stopped to work.

For example, if we have the following instruction for x86_64 disassembly:
```
4180fc07 cmp r12b, 7 
```
And 7 here is the part of some enum, e.g.:
```c
enum BLA {
    FOO = 5,
    OOF = 6,
    BAR = 7
};
```
We need a way to specify the operand as the type structure with either `tl` or `ah` commands.

PR #728, opened 2021-02-26
Migrating types command to the newshell

XVilka:
# **DO NOT SQUASH**

**Detailed description**

- Migrating `t` commands to the newshell and fixes along the way.
- Removed `touch` command
- Added `tek` command to print enum contents in the SDB format
- Added `tef` command to search for the matching enums by the value (necessary for https://github.com/rizinorg/rizin/issues/697)
- Removed `tfc` command and now `tf` behaves as one - it prints the function types in C format
- Moved old `tf` command to `tfk` to print SDB format

For example, lets define two enums:
```
td "enum BLA { A = 1, B = 0x80, C = 2 };"
td "enum FOO { G = 2, B = 0x90, P = 3254543};"
```
If you run `tef 2` it will show:
```
BLA.C
FOO.G
```
If you run `tef 1` it will show:
```
BLA.A
```

**Test plan**

All green

### DIFF
--- a/librz/core/cmd_descs/cmd_type.yaml
+++ b/librz/core/cmd_descs/cmd_type.yaml
@@ -109,16 +109,6 @@ commands:
           - name: type
             type: RZ_CMD_ARG_TYPE_STRING
             optional: true
-      - name: tfc
-        cname: type_function_c
-        summary: Show function definition in the C output format
-        modes:
-          - RZ_OUTPUT_MODE_STANDARD
-          - RZ_OUTPUT_MODE_JSON
-        args:
-          - name: type
-            type: RZ_CMD_ARG_TYPE_STRING
-            optional: true
   - name: tk
     cname: type_kuery
     summary: Perform SDB query on types database
--- a/librz/core/cmd_descs/cmd_type.yaml
+++ b/librz/core/cmd_descs/cmd_type.yaml
@@ -50,6 +50,51 @@ commands:
     args:
       - name: type
         type: RZ_CMD_ARG_TYPE_STRING
+  - name: te
+    summary: List loaded enums
+    subcommands:
+      - name: te
+        cname: type_list_enum
+        summary: List loaded enums / Show enum member
+        modes:
+          - RZ_OUTPUT_MODE_STANDARD
+          - RZ_OUTPUT_MODE_JSON
+          - RZ_OUTPUT_MODE_SDB
+        args:
+          - name: enum
+            type: RZ_CMD_ARG_TYPE_STRING
+            optional: true
+          - name: value
+            type: RZ_CMD_ARG_TYPE_STRING
+            optional: true
+      - name: teb
+        cname: type_enum_bitfield
+        summary: Show enum bitfield
+        args:
+          - name: enum
+            type: RZ_CMD_ARG_TYPE_STRING
+          - name: field
+            type: RZ_CMD_ARG_TYPE_STRING
+      - name: tec
+        cname: type_enum_c
+        summary: Show enum in the C output format
+        args:
+          - name: enum
+            type: RZ_CMD_ARG_TYPE_STRING
+            optional: true
+      - name: ted
+        cname: type_enum_c_nl
+        summary: Show enum in the C output format without newlines
+        args:
+          - name: enum
+            type: RZ_CMD_ARG_TYPE_STRING
+            optional: true
+      - name: tef
+        cname: type_enum_find
+        summary: Find enum and member by the member value
+        args:
+          - name: value
+            type: RZ_CMD_ARG_TYPE_STRING
   - name: tf
     summary: List loaded functions definitions
     subcommands:
--- a/librz/core/cmd_descs/cmd_type.yaml
+++ b/librz/core/cmd_descs/cmd_type.yaml
@@ -50,6 +50,30 @@ commands:
     args:
       - name: type
         type: RZ_CMD_ARG_TYPE_STRING
+  - name: tf
+    summary: List loaded functions definitions
+    subcommands:
+      - name: tf
+        cname: type_list_function
+        summary: List loaded function definitions / Show function signature
+        modes:
+          - RZ_OUTPUT_MODE_STANDARD
+          - RZ_OUTPUT_MODE_JSON
+          - RZ_OUTPUT_MODE_SDB
+        args:
+          - name: type
+            type: RZ_CMD_ARG_TYPE_STRING
+            optional: true
+      - name: tfc
+        cname: type_function_c
+        summary: Show function definition in the C output format
+        modes:
+          - RZ_OUTPUT_MODE_STANDARD
+          - RZ_OUTPUT_MODE_JSON
+        args:
+          - name: type
+            type: RZ_CMD_ARG_TYPE_STRING
+            optional: true
   - name: tk
     cname: type_kuery
     summary: Perform SDB query on types database
--- a/librz/core/cmd_descs/cmd_type.yaml
+++ b/librz/core/cmd_descs/cmd_type.yaml
@@ -150,6 +150,28 @@ commands:
         cname: type_noreturn_del_all
         summary: Remove all noreturn references
         args: []
+  - name: to
+    summary: Open C header file and load types from it
+    subcommands:
+      - name: to
+        cname: type_open_file
+        summary: Open C header file and load types from it
+        args:
+          - name: file
+            type: RZ_CMD_ARG_TYPE_FILE
+      - name: toe
+        cname: type_open_editor
+        summary: Open cfg.editor to edit types
+        args:
+          - name: type
+            type: RZ_CMD_ARG_TYPE_STRING
+            optional: true
+      - name: tos
+        cname: type_open_sdb
+        summary: Open SDB file and load types from it
+        args:
+          - name: file
+            type: RZ_CMD_ARG_TYPE_FILE
   - name: tt
     summary: List loaded typedefs
     subcommands:
--- a/librz/core/cmd_type.c
+++ b/librz/core/cmd_type.c
@@ -26,7 +26,6 @@ static const char *help_msg_t[] = {
 	"to", " <path>", "Load types from C header file",
 	"toe", " [type.name]", "Open cfg.editor to edit types",
 	"tos", " <path>", "Load types from parsed Sdb database",
-	"touch", " <file>", "Create or update timestamp in file",
 	"tp", "  <type> [addr|varname]", "cast data at <address> to <type> and print it (XXX: type can contain spaces)",
 	"tpv", " <type> @ [value]", "Show offset formatted for given type",
 	"tpx", " <type> <hexpairs>", "Show value for type with specified byte sequence (XXX: type can contain spaces)",
@@ -71,7 +70,6 @@ static const char *help_msg_to[] = {
 	"to", " -", "Open cfg.editor to load types",
 	"to", " <path>", "Load types from C header file",
 	"tos", " <path>", "Load types from parsed Sdb database",
-	"touch", " <file>", "Create or update timestamp in file",
 	NULL
 };
 
@@ -1535,14 +1533,6 @@ RZ_IPI int rz_cmd_type(void *data, const char *input) {
 		}
 		if (input[1] == ' ') {
 			types_open_file(core, input + 2);
-		} else if (input[1] == 'u') {
-			// "tou" "touch"
-			char *arg = strchr(input, ' ');
-			if (arg) {
-				rz_file_touch(arg + 1);
-			} else {
-				eprintf("Usage: touch [filename]");
-			}
 		} else if (input[1] == 's') { // "tos"
 			types_open_sdb(core, input + 3);
 		} else if (input[1] == 'e') { // "toe"
--- a/librz/core/cmd_type.c
+++ b/librz/core/cmd_type.c
@@ -37,7 +37,7 @@ static const char *help_msg_t[] = {
 };
 
 static const char *help_msg_tcc[] = {
-	"Usage: tfc", "[-name]", "# type function calling conventions",
+	"Usage: tcc", "[-name]", "# type function calling conventions",
 	"tcc", "", "List all calling convcentions",
 	"tcc", " r0 pascal(r0,r1,r2)", "Show signature for the 'pascal' calling convention",
 	"tcc", "-pascal", "Remove the pascal cc",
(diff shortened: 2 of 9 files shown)


librz/core/cmd_type.c
```diff
@@ -254,6 +253,163 @@ static void types_cc_print(RzCore *core, const char *cc, RzOutputMode mode) {
...
+			rz_list_foreach (list, iter, member) {
+				pj_kn(pj, member->name, rz_num_math(NULL, member->val));
+			}
+			pj_end(pj);
+		}
+		pj_end(pj);
+		rz_cons_printf("%s\n", pj_string(pj));
+		pj_free(pj);
+		rz_list_free(list);
+	} break;
```
ret2libc: put the break inside the  `{ .. }`. At least it will be with the right indentation.

test/db/cmd/types
```diff
@@ -278,18 +278,108 @@ RUN
...
+RUN
+
+NAME=tej
+FILE=-
+CMDS=<<EOF
+td "enum FOO { A=1, B=0x100, C=0xFF, D=4, E=5 };"
+td "enum BAR { G=0x800, Z=0xEE, P=4, E=5 };"
+tej FOO
+tej BAR
+tej
```
ret2libc, line 316: this should output an array, not just 2 objects, I think.

librz/core/cmd_type.c
```diff
@@ -1115,7 +1209,7 @@ static void type_list_c_all(RzCore *core) {
 	// TODO: Change the logic maybe?
 	//eprintf("Specify the type");
 	//return RZ_CMD_STATUS_ERROR;
-	rz_core_cmd0(core, "tfc");
+	rz_core_cmd0(core, "tfc"); // seems return nothing
```
ret2libc: use `types_function_print`?

librz/core/cmd_type.c
```diff
@@ -1883,6 +1919,84 @@ RZ_IPI RzCmdStatus rz_type_define_handler(RzCore *core, int argc, const char **a
...
+	// TODO: Reconsider the `te <enum_name> <member_value>` syntax change
+	const char *member_value = argc > 2 ? argv[2] : NULL;
+	if (enum_name) {
+		if (member_value) {
+			return types_enum_member_find(core, enum_name, member_value);
+		} else {
+			types_enum_print(core, enum_name, mode);
+		}
+	} else {
+		// A special case, since by default `te` returns only the list of all enums
```
ret2libc, line 1976: can't we just add support for `te` and `teq`?

librz/core/cmd_type.c
```diff
@@ -1883,6 +1919,84 @@ RZ_IPI RzCmdStatus rz_type_define_handler(RzCore *core, int argc, const char **a
 	return RZ_CMD_STATUS_OK;
 }
 
+RZ_IPI RzCmdStatus rz_type_list_enum_handler(RzCore *core, int argc, const char **argv, RzOutputMode mode) {
+	const char *enum_name = argc > 1 ? argv[1] : NULL;
+	// TODO: Reconsider the `te <enum_name> <member_value>` syntax change
```
ret2libc, line 1962: This is something that always bugs me... Having the same handler that does both listing and addition. I'm still not sure if it's a real problem or not and, if it is, how to improve the situation... Adding yet another command is a bit meh.

librz/core/cmd_type.c
```diff
@@ -1883,6 +1919,84 @@ RZ_IPI RzCmdStatus rz_type_define_handler(RzCore *core, int argc, const char **a
...
+RZ_IPI RzCmdStatus rz_type_enum_find_handler(RzCore *core, int argc, const char **argv) {
+	const char *enum_value = argc > 1 ? argv[1] : NULL;
+	return types_enum_member_find_all(core, enum_value);
+}
+
+RZ_IPI RzCmdStatus rz_type_list_function_handler(RzCore *core, int argc, const char **argv, RzOutputMode mode) {
+	const char *function = argc > 1 ? argv[1] : NULL;
+	Sdb *TDB = core->analysis->sdb_types;
+	if (function) {
+		// A special case, since by default `tf name` returns the output in SDB format
```
ret2libc: Same here. Can't we just support both STANDARD and SDB modes? For consistency, sdb output should be printed with `tfk`

ret2libc: I'm trying the commands, and I feel like `tfc` should really be just `tf` and `tf` should become `tfk`. THis would be consistent with other commands.

librz/core/cmd_type.c
```diff
@@ -1883,6 +1919,84 @@ RZ_IPI RzCmdStatus rz_type_define_handler(RzCore *core, int argc, const char **a
...
+		if (mode == RZ_OUTPUT_MODE_JSON) {
+			print_keys(TDB, core, stdiffunc, printfunc_json_cb, false);
+		} else {
+			print_keys(TDB, core, stdiffunc, printkey_cb, false);
+		}
+	}
+	return RZ_CMD_STATUS_OK;
+}
+
+RZ_IPI RzCmdStatus rz_type_function_c_handler(RzCore *core, int argc, const char **argv, RzOutputMode mode) {
```
ret2libc: in oldshell i'm not able to make `tfcj` work. Also, what would be the difference between `tfcj` and `tcj` in this new implementation? I think they are the same thing. Honestly, I can't think about a "C JSON" reprensetation. IMHO this should be merged with `tf`, as said above.

librz/core/cmd_descs/cmd_type.yaml
```diff
@@ -81,6 +150,28 @@ commands:
         cname: type_noreturn_del_all
         summary: Remove all noreturn references
         args: []
+  - name: to
```
ret2libc, line 143: Instead of using `load` as cname and `Load` as summary for most of these, i would rephrase a bit and use the word `open`/`Open` because i guess `to` stands for `type open`, meaning that you open a file to work on types. `to` opens a file and loads types from it, `toe` open the editor, `tos` open a sdb file and load types from it.

librz/core/cmd_type.c
```diff
@@ -1933,6 +2047,24 @@ RZ_IPI RzCmdStatus rz_type_noreturn_del_all_handler(RzCore *core, int argc, cons
...
+
+RZ_IPI RzCmdStatus rz_type_load_editor_handler(RzCore *core, int argc, const char **argv) {
+	const char *typename = argc > 1 ? argv[1] : NULL;
+	types_load_editor(core, typename);
+	return RZ_CMD_STATUS_OK;
+}
+
+RZ_IPI RzCmdStatus rz_type_load_sdb_handler(RzCore *core, int argc, const char **argv) {
+	const char *path = argc > 1 ? argv[1] : NULL;
+	types_load_sdb(core, path);
```
ret2libc: just do `types_load_sdb(core, argv[1])` and remove the previous line. This argument is anyway required.

librz/core/cmd_descs/cmd_type.yaml
```diff
@@ -81,6 +150,28 @@ commands:
...
+    summary: Load types from C header file
+    subcommands:
+      - name: to
+        cname: type_load_file
+        summary: Load types from C header file
+        args:
+          - name: file
+            type: RZ_CMD_ARG_TYPE_STRING
+      - name: toe
```
ret2libc, line 152: Other places use `!` as suffix for the editor, however i think there's no real consistent way at the moment. I like `e`, but it could be some commands already have it occupied... Maybe `E`? Not sure...

librz/core/cmd_type.c
```diff
@@ -1153,6 +1247,78 @@ static void type_define(RzCore *core, const char *type) {
...
+			fprintf(stderr, "%s", error_msg);
+			free(error_msg);
+		}
+	}
+	free(homefile);
+}
+
+static void types_load_editor(RzCore *core, const char *typename) {
+	Sdb *TDB = core->analysis->sdb_types;
+	char *str = rz_core_cmd_strf(core, "tc %s", typename);
```
ret2libc: `typename` might be null when called from the newshell handler, because the argument is optional. I think this is not alright.

librz/core/cmd_type.c
```diff
@@ -1933,6 +2047,24 @@ RZ_IPI RzCmdStatus rz_type_noreturn_del_all_handler(RzCore *core, int argc, cons
 	return RZ_CMD_STATUS_OK;
 }
 
+RZ_IPI RzCmdStatus rz_type_load_file_handler(RzCore *core, int argc, const char **argv) {
+	const char *path = argc > 1 ? argv[1] : NULL;
+	types_load_file(core, path);
```
ret2libc: just do `types_load_file(core, argv[1]);`

librz/core/cmd_type.c
XVilka: Agree on this.

librz/core/cmd_descs/cmd_type.yaml
XVilka: I can note it in my TODO but this is out of the scope of this PR, imho.

librz/core/cmd_type.c
XVilka: Done.

librz/core/cmd_type.c
XVilka: Fixed in https://github.com/rizinorg/rizin/pull/728/commits/636b315bebfa6e1e461c0c9910d8ea1b9ddc77a5

librz/core/cmd_type.c
XVilka: Removed it.

librz/core/cmd_type.c
XVilka: I am not sure about this to be honest. It makes sense to list only enum names by default, since there can be many.

librz/core/cmd_type.c
XVilka: It doesn't add the value, it can only be done through `td` or manually in SDB. But it gives fine grained access for the enum members.

librz/core/cmd_descs/cmd_type.yaml
ret2libc: It was just for discussion, to check if you have any idea about it.

test/db/cmd/types
ret2libc: @XVilka seen this?

librz/core/cmd_descs/cmd_type.yaml
```diff
@@ -50,6 +50,65 @@ commands:
...
+          - name: field
+            type: RZ_CMD_ARG_TYPE_STRING
+      - name: tec
+        cname: type_enum_c
+        summary: Show enum in the C output format
+        args:
+          - name: enum
+            type: RZ_CMD_ARG_TYPE_STRING
+            optional: true
+      - name: ted
```
ret2libc, line 85: this should probably be a subcommand of `tec` (e.g. `tecn`, because also `?e` has `?en` to avoid newlines).

librz/core/cmd_type.c
ret2libc: Right, but maybe we can consider STANDARD just the enum names and LONG the C version. This way you'd have just one command to list enums in different formats:
`te` -> just enum names (STANDARD)
`tej` -> enum in json (JSON)
`tek` -> enum in sdb (SDB)
`tel` -> enum in C (LONG)

librz/core/cmd_descs/cmd_type.yaml
XVilka: We have the same pattern for `tsd`, `tud`, etc. Do you suggest to convert all of them?

librz/core/cmd_type.c
XVilka: But then `te <enum name>` command wouldn't have any meaning like right now. Other types have output with `pf` string when you call e.g. `tu <union name>`.

test/db/cmd/types
XVilka: It will require passing `PJ` pointer around :(

librz/core/cmd_descs/cmd_type.yaml
```diff
@@ -81,6 +140,28 @@ commands:
...
         args: []
+  - name: to
+    summary: Open C header file and load types from it
+    subcommands:
+      - name: to
+        cname: type_open_file
+        summary: Open C header file and load types from it
+        args:
+          - name: file
+            type: RZ_CMD_ARG_TYPE_STRING
```
ret2libc: TYPE_FILE

librz/core/cmd_descs/cmd_type.yaml
```diff
@@ -81,6 +140,28 @@ commands:
...
+        args:
+          - name: type
+            type: RZ_CMD_ARG_TYPE_STRING
+            optional: true
+      - name: tos
+        cname: type_open_sdb
+        summary: Open SDB file and load types from it
+        args:
+          - name: file
+            type: RZ_CMD_ARG_TYPE_STRING
```
ret2libc: type file

test/db/cmd/types
ret2libc: it is necessary, I think. If you have a json command that returns more than one object, they should be wrapped in `[...]`.

librz/core/cmd_type.c
ret2libc: So many inconsistencies :( I even wonder why one would want to print "pf format" instead of something easier... I like more and more the idea of having `:<mode>` as a way to specify the output mode you want. So commands could just have `<cmd>:pf` or `<cmd>:sdb` `<cmd>:json` instead of having to find a letter for each case, ending up with many inconsistencies :(

XVilka: Yes, the idea of having the ":<mode>" is the good one.

librz/core/cmd_type.c
```diff
@@ -321,14 +319,23 @@ static void types_enum_print_all(RzCore *core, RzOutputMode mode) {
 	Sdb *TDB = core->analysis->sdb_types;
 	SdbKv *kv;
 	SdbListIter *iter;
+	PJ *pj = (mode == RZ_OUTPUT_MODE_JSON) ? pj_new() : NULL;
```
ret2libc: rz_core_pj_new @XVilka

test/db/cmd/types
XVilka: Done, albeit in an ugly way.

ret2libc (approved):
I'm ok with this, however please consider the various suggestions once you refactor further the whole types thing.

PR #731, opened 2021-02-26
BUILDING.md improvements

caribpa:

**Detailed description**
Currently the default build instructions create debug builds. The target audience for the `BUILDING.md` document should be the end user by default (who's usually not interested in debugging), moreover, the default build (optimized debug) blocks/breaks the build process in Windows 10 x86 (tested) due to the buggy `mspdbsrv.exe` process when generating `PDB` files (see [here](https://social.msdn.microsoft.com/Forums/en-US/9e58b7d1-a47d-4a76-943a-4f35090616e8/link-fatal-error-lnk1318?forum=vclanguage) for more info).


ret2libc (commented):
I'm ok with the change, however we might also say that people not interested in dev should use releases and those releases are built with --buildtype=release and whatnot. Having the users that follow BUILDING.md use debug builds might help in case they catch bugs and we need more info (e.g. stacktraces).

caribpa:
Ideally yes, but there are currently two issues:
1. Windows users may not be able to build Rizin with `--buildtype=debug` (as the `mspdbsrv.exe` blocks the building process 100% of the times, even when building with 1 thread, I have experienced this in Windows 10 x86)
2. Currently there's not defined release cycle, so the latest release is way behind from the `dev` branch, in fact I still see from time to time the fortune message saying that the latest release is obtained by building from rizin-git (`dev` branch is the default)

If we fix nº 2 (#703 😉), then we could confidently expect that users using the `dev` branch are interested in aiding the development (thus `--buildtype=debug` should the default).

ret2libc (approved):
Ok then! @XVilka @kazarmy @thestr4ng3r cc

wargio (approved):

kazarmy (approved):
Hmm maybe we should test the release build instead.

ret2libc:
> Hmm maybe we should test the release build instead.

Good point...  @caribpa if we advice for buildtype=release, we should also test it, so I would add `--buildtype=release` in `ci.yml` for, at least, `linux-meson-gcc-tests`, while leaving the others as they are now. THis way we test both build types.

kazarmy:
`linux-meson-gcc-tests` is needed for test coverage so you can't go to `buildtype=release` for that build, otherwise the Codecov reports will be wonky.

kazarmy:
> ... the Codecov reports will be wonky.

Oops the reports will probably not exist at all, since there's no debug information.

caribpa:
> Good point... @caribpa if we advice for buildtype=release, we should also test it, so I would add `--buildtype=release` in `ci.yml` for, at least, `linux-meson-gcc-tests`, while leaving the others as they are now. THis way we test both build types.

If the release build is composed by cherry-picking from the development build, and the development build is the one tested, then, shouldn't that mean that the release build changes are already tested?

kazarmy:
There's no cherry-picking involved here ... the release build is `buildtype=release`, the tested build is `buildtype=debug` (except Windows, which is `buildtype=release`) and unfortunately release optimization can change program behavior if code is not written correctly.

ret2libc:
Let's add --buildtype=release to linux-meson-clang-tests then.

kazarmy:
> Let's add --buildtype=release to linux-meson-clang-tests then.

I agree. It's just that Clang and gcc employ different optimizers and therefore for greater confidence, you should add `--buildtype=release` to a build that uses gcc, as well.

caribpa:
> There's no cherry-picking involved here ... the release build is `buildtype=release`, the tested build is `buildtype=debug` (except Windows, which is `buildtype=release`) and unfortunately release optimization can change program behavior if code is not written correctly.

It was an innocent joke 👼, I think that there should be two branches (`stable` and `dev`) with the two corresponding build tests. The `stable` branch gets the cherry-pics for the next release (and gets tested with `--buildtype=release`), whereas the usual commits to the `dev`branch is tested with `--buildtype=debug`. This way everything is covered.

XVilka (approved):

ret2libc:
> > There's no cherry-picking involved here ... the release build is `buildtype=release`, the tested build is `buildtype=debug` (except Windows, which is `buildtype=release`) and unfortunately release optimization can change program behavior if code is not written correctly.
> 
> It was an innocent joke 👼, I think that there should be two branches (`stable` and `dev`) with the two corresponding build tests. The `stable` branch gets the cherry-pics for the next release (and gets tested with `--buildtype=release`), whereas the usual commits to the `dev`branch is tested with `--buildtype=debug`. This way everything is covered.

The thing is that you are now telling people to use release even on dev, so if you test release only on stable, you are actually giving users something that you don't test

caribpa:
> The thing is that you are now telling people to use release even on dev, so if you test release only on stable, you are actually giving users something that you don't test

Of course that's not a good idea, that's why the last thing I said is that both `stable` and `dev` branches should be tested. In my opinion, the `stable` branch should contain the gradual changes that will make it to the next release without further modifications.

In case I am still not explaining myself, suppose the next release will be 0.2.0 and the team agreed that, from all the changes, only X, Y, Z will make it to 0.2.0. `dev` branch will have all the changes (A, B, ..., X, Y, Z), but `stable` branch will only receive the commits for changes X, Y, Z (cherry-picked from `dev` when they are ready). Now, whenever X, Y, or Z are ready to make it to `stable`, the changes get automatically tested by the `stable` CI pipeline via pull/merge request. And `dev` branch stays like now with its own CI pipeline running the tests.

Testing both `stable` and `dev` branches is the most robust approach to ensure all changes play nice.

caribpa:
Ultimately, the point of having a `stable` branch is to give the users a way for using the latest stable changes without having to wait for the next release. Following the previous example, suppose that the change X is finished in the `dev` branch, so it gets added to the `stable` branch; but note that the next [release](https://github.com/rizinorg/rizin/releases) won't be available until changes Y and Z are finished.
Thanks to the `stable` branch, the users can enjoy the change X without having to wait for the next release, and without being forced to use the `dev` branch.

ret2libc:
That's one other way to look at the stable/dev branches we have, yes. Actually, the main reason we used them is that in this way we can still do patch releases even if dev branch introduced breaking changes.

caribpa:
> Actually, the main reason we used them is that in this way we can still do patch releases even if dev branch introduced breaking changes.

Yes, in my distinction of `stable` vs `dev` branch, the `stable` branch also receives bug fixes. When I said that `stable` branch gets change X cherry-picked from `dev` branch also refers to bug fixes.

caribpa:
Basically you are giving the advanced user the chance of being bleeding-edge while being stable by compiling Rizin from the `stable` branch. On the other hand, the lazy user may just get their Rizin from the pre-compiled binaries from [releases](https://github.com/rizinorg/rizin/releases) or directly from their distro packages.

I'm proposing to do it a bit like the linux kernel (where I am one of those who compiles their own customized kernel from the latest stable changes).

ret2libc:
@kazarmy just to wrap up: are you adding `--buildtype=release` to the CI?

kazarmy:
Yes in #260, but even now I think it's safe to use `--buildtype=release` on the `dev` branch in the sense that all tests including the fuzz tests will pass.

ret2libc (approved):

PR #733, opened 2021-02-26
Replace SDB wtih HT in librz/core/agraph.c

p371jah:

**Detailed description**

...

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/librz/core/agraph.c
+++ b/librz/core/agraph.c
@@ -3,6 +3,7 @@
 #include <rz_core.h>
 #include <rz_cons.h>
 #include <rz_util/rz_graph_drawable.h>
+#include <ht_pu.h>
 #include <ctype.h>
 #include <limits.h>
 #include "core_private.h"
@@ -50,11 +51,6 @@ static const char *mousemodes[] = {
 
 #define NORMALIZE_MOV(x) ((x) < 0 ? -1 : ((x) > 0 ? 1 : 0))
 
-#define hash_set(sdb, k, v)    (sdb_num_set(sdb, sdb_fmt("%" PFMT64u, (ut64)(size_t)(k)), (ut64)(size_t)(v), 0))
-#define hash_get(sdb, k)       (sdb_num_get(sdb, sdb_fmt("%" PFMT64u, (ut64)(size_t)(k)), NULL))
-#define hash_get_rnode(sdb, k) ((RzGraphNode *)(size_t)hash_get(sdb, k))
-#define hash_get_rlist(sdb, k) ((RzList *)(size_t)hash_get(sdb, k))
-#define hash_get_int(sdb, k)   ((int)hash_get(sdb, k))
 /* don't use macros for this */
 #define get_anode(gn) ((gn) ? (RzANode *)(gn)->data : NULL)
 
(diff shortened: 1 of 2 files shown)


XVilka (approved):
Hah, SDB was used as hashtable before. Good catch! :+1:

librz/core/agraph.c
```diff
@@ -920,23 +916,30 @@ static int is_valid_pos(const RzAGraph *g, int l, int pos) {
...
+		for (int j = 0; j < g->layers[i].n_nodes; j++) {
 			RzGraphNode *gn = g->layers[i].nodes[j];
-			const RzList *Ln = hash_get_rlist(res, gn);
+			const RzList *Ln = ht_pp_find(res, gn, NULL);
 			const RzANode *an = get_anode(gn);
 
 			if (!Ln) {
 				RzList *vert = rz_list_new();
-				hash_set(res, gn, vert);
+				ht_pp_update(res, gn, vert);
```
ret2libc: This is just a `ht_pp_insert`, isn't it?

librz/core/agraph.c
```diff
@@ -1146,36 +1156,43 @@ static void place_nodes(const RzAGraph *g, const RzGraphNode *gn, int is_left, S
 	}
 
 	graph_foreach_anode (lv, itk, gk, ak) {
-		hash_set(res, gk, p);
-		hash_set(placed, gk, true);
+		ht_pu_update(res, gk, (ut64)(size_t)p);
```
ret2libc, line 1161: is also this just an insert? I think we should use `_insert` when we want to do just an insert.

librz/core/agraph.c
```diff
@@ -1146,36 +1156,43 @@ static void place_nodes(const RzAGraph *g, const RzGraphNode *gn, int is_left, S
...
 	if (!classes) {
 		return NULL;
 	}
 
-	Sdb *res = sdb_new0();
-	Sdb *placed = sdb_new0();
+	HtPUOptions pu_opt = { 0 };
+	HtPPOptions pp_opt = { 0 };
+	HtPU *res = ht_pu_new_opt(&pu_opt);
+	SetP *placed = (SetP *)ht_pp_new_opt(&pp_opt);
```
ret2libc, line 1178: this is risky.. But I see that setp was designed only for strings, which is IMHO wrong. A Set based on strings should be called like `SetS` or `SetString` or something like that. But I guess that's ok for now :(

ret2libc:
With this PR I get these warnings that are not present in dev:
```
WARNING: hash_get_int: assertion 'found' failed (line 1035)
WARNING: hash_get_int: assertion 'found' failed (line 1035)
WARNING: hash_get_int: assertion 'found' failed (line 1035)
```
This happens while working on ./test/bins/elf/ls, with the `main` function.

librz/core/agraph.c
p371jah: No, overrides are possible here.

p371jah:
@ret2libc You can achieve same result on dev by replacing `hash_get_int()` macro with with the below snippet. Anyway I have removed this assertion.
```c
static int hash_get_int(Sdb *sdb, const void *k) {
	ut32 cas;
	int val = sdb_num_get(sdb, sdb_fmt("%" PFMT64u, (ut64)(size_t)(k)), &cas);
	rz_return_val_if_fail(cas != 0, 0);
	return val;
}
```

ret2libc (approved):

XVilka (approved):

PR #742, opened 2021-03-01
Pull actively info-applying functions out of rz_core_bin_info()

thestr4ng3r:


librz/core/cbin.c
```diff
@@ -321,33 +345,90 @@ RZ_API int rz_core_bin_set_by_name(RzCore *core, const char *name) {
...
+	rz_asm_use(r->rasm, arch);
+
+	// ----
+	// inlined: rz_core_bin_info(r, RZ_CORE_BIN_ACC_ALL, NULL, RZ_MODE_SET, va, NULL, NULL);
+	ut64 loadaddr = binobj ? binobj->loadaddr : UT64_MAX;
+
+	// use our internal values for va
+	va = va ? VA_TRUE : VA_FALSE;
+
+	rz_core_bin_apply_strings(r, binfile);
```
ret2libc, line 379: why all these calls?

thestr4ng3r: This is all the code that `rz_core_bin_info(r, RZ_CORE_BIN_ACC_ALL, NULL, RZ_MODE_SET, va, NULL, NULL)`, which was there before was calling. The plan is to make it into individual functions `rz_core_bin_apply_*()` per info type and only for applying info, while `rz_core_bin_info()` will only do printing.

ret2libc: Ah, YES! If this is to split the set and the printing, then yes please :)

ret2libc (approved):

librz/core/cbin.c
thestr4ng3r: Exactly this.

Maijin (approved):

librz/core/cbin.c
```diff
@@ -29,6 +29,30 @@
 	if (binfile && binfile->rbin && binfile->rbin->verbose) \
 	eprintf
 
+static RZ_NULLABLE RZ_BORROW const RzList *core_bin_strings(RzCore *r, RzBinFile *file);
+static void _print_strings(RzCore *r, const RzList *list, PJ *pj, int mode, int va);
+static bool bin_raw_strings(RzCore *r, PJ *pj, int mode, int va);
+static int bin_info(RzCore *r, PJ *pj, int mode, ut64 laddr);
+static int bin_main(RzCore *r, PJ *pj, int mode, int va);
+static int bin_dwarf(RzCore *core, PJ *pj, int mode);
```
Maijin, line 37: Would it be possible to sort this list somehow?

thestr4ng3r: Yes, but these declarations are also only temporary until the are all unfolded and refactored.

thestr4ng3r:
minstr/maxstr/filter is missing in apply_strings, putting this to draft again until I figured out how to do that best without code duplication.

XVilka:
It looks good. In the future we might consider to split `librz/core/cmd_*.c` files into the different directory than `librz/core/cbin.c`, `librz/core/canalysis.c`, or vice versa.

librz/core/cbin.c
Maijin: ok

Maijin:
@thestr4ng3r if you're refactoring, may as well use the table API to have the ability to sort per column etc. 👍

thestr4ng3r:
Table API is a good idea, but should be done seperately.

Maijin (approved):

PR #749, opened 2021-03-02
build: improve documentation for static building

### LINKED ISSUE #735: Linux x64 - Cross-arch static compilation to x86 requires static_runtime=true
### Work environment

| OS/arch/bits (mandatory)                             | Kali Linux x64
| File format of the file you reverse (mandatory)      | n/a
| Architecture/bits of the file (mandatory)            | n/a
| `rizin -v` full output, **not truncated** (mandatory)  | n/a

### Expected behavior
Cross-arch compiling Rizin statically from x64 to x86 works when following the [BUILDING](https://github.com/rizinorg/rizin/blob/dev/BUILDING.md) guide (and forcing 32 bit build).

### Actual behavior
Cross-arch static compilation doesn't work without adding the `-Dstatic_runtime=true` meson flag.

```sh
$ git clone https://github.com/rizinorg/rizin
$ cd rizin
$ CFLAGS="-static -m32" LDFLAGS="-static -m32" meson --prefix="${PWD}/build/install" --default-library=static --buildtype=release build
$ ninja -C build
# 💔 - error: /usr/bin/ld: attempted static link of dynamic object `/usr/lib/gcc/x86_64-linux-gnu/10/../../../../lib32/libutil.so'
```

The full error is:
```
[...]
[1376/1550] Linking target binrz/rz-test/rz-test
FAILED: binrz/rz-test/rz-test 
cc  -o binrz/rz-test/rz-test binrz/rz-test/rz-test.p/rz-test.c.o binrz/rz-test/rz-test.p/load.c.o binrz/rz-test/rz-test.p/run.c.o -Wl,--as-needed -Wl,--no-undefined -Wl,-O1 -static -m32 -static -m32 -Wl,--start-group librz/util/librz_util.a librzspp.a subprojects/sdb/src/libsdb.a subprojects/zlib-1.2.11/libz.a -ldl -lm -pthread /usr/lib/gcc/x86_64-linux-gnu/10/../../../../lib32/libutil.so -Wl,--end-group
/usr/bin/ld: librz/util/librz_util.a(lib.c.o): in function `rz_lib_dl_open':
lib.c:(.text+0xa0): warning: Using 'dlopen' in statically linked applications requires at runtime the shared libraries from the glibc version used for linking
/usr/bin/ld: attempted static link of dynamic object `/usr/lib/gcc/x86_64-linux-gnu/10/../../../../lib32/libutil.so'
collect2: error: ld returned 1 exit status
[...]
```

**Important**: This error does **not** happen (meaning that the cross compilation is successful) when passing `-Dstatic_runtime=true` to meson:

```sh
$ git clone https://github.com/rizinorg/rizin
$ cd rizin
$ CFLAGS="-static -m32" LDFLAGS="-static -m32" meson --prefix="${PWD}/build/install" --default-library=static -Dstatic_runtime=true --buildtype=release build
$ ninja -C build
# ❤️ - Though with many warnings in the linking process saying:
# /usr/bin/ld: librz/util/librz_util.a(lib.c.o): in function `rz_lib_dl_open':
# lib.c:(.text+0xa0): warning: Using 'dlopen' in statically linked applications requires at runtime the shared libraries from the glibc version used for linking
```

Currently the `--default-library=static` meson flag doesn't seem to imply `-Dstatic_runtime=true` according to [this](https://github.com/rizinorg/rizin/blob/dev/meson.build#L20) (and the following) lines, but I don't know if this is wanted behavior.

Note that I purposely used Kali Linux as it is a popular distro in the InfoSec world, install `gcc-multilib` for being able to cross-compile.

ret2libc:
- [x] I've updated the documentation and the [rizin book](https://github.com/rizinorg/book) with the relevant information (if needed). See https://github.com/rizinorg/book/pull/35

**Detailed description**

When compiling statically, you not only want to statically compile your own libs/binaries, but you also want to search static dependencies, so we use `static_runtime` to tell meson to do so.

**Test plan**

rizin-static should still generate static binaries.

**Closing issues**

Closes https://github.com/rizinorg/rizin/issues/735

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -365,7 +365,7 @@ jobs:
     - name: Compile with meson
       run: |
         mkdir -p ${HOME}/rizin-static
-        CFLAGS="-static" LDFLAGS="-static" meson --prefix=${HOME}/rizin-static --buildtype release --default-library static -Dstatic_runtime=true -Dportable=true build
+        meson --prefix=${HOME}/rizin-static --buildtype release --default-library static -Dstatic_runtime=true -Dportable=true build
         ninja -C build && ninja -C build install
         tar -C ${HOME}/rizin-static --xz -cf rizin-static.tar.xz $(ls ${HOME}/rizin-static)
       working-directory: rizin
@@ -550,7 +550,7 @@ jobs:
     - name: Compile with meson
       run: |
         export PATH=${HOME}/.local/bin:${PATH}
-        CFLAGS="-static" LDFLAGS="-static" meson --buildtype release --default-library static --prefix=/tmp/android-dir -Dportable=true -Dblob=true build --cross-file .github/meson-android-${{ matrix.name }}.ini
+        meson --buildtype release --default-library static --prefix=/tmp/android-dir -Dportable=true -Dblob=true -Dstatic_runtime=true build --cross-file .github/meson-android-${{ matrix.name }}.ini
         ninja -C build && ninja -C build install
     - name: Create rizin-android-${{ matrix.name }}.tar.gz
       run: |
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -104,11 +104,12 @@ $ meson -Db_sanitize=address,undefined build
 
 ## Build fully-static binaries
 
-It may be useful to run Rizin just by using a single file, which can be copied
-on other systems if necessary.
+It may be useful to run Rizin just by using a single file, which can be
+copied on other systems if necessary. On *NIX systems, this adds the classic
+`-static` flag to the linker, while on Windows it uses `/MT`.
 
 ```
-$ CFLAGS="-static" meson --default-library=static build
+$ meson --default-library=static -Dstatic_runtime=true build
 ```
 
 ## Cross-compilation for Android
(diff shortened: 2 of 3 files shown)


ret2libc:
cc @caribpa This should "solve" your problem. Let's just use `static_runtime` to also set LDFLAGS=-static and we are good to go. Would this be ok?

caribpa:
Does it work on Windows?

I don't remember about `LDFLAGS` but I am certain that setting `CFLAGS` in a `cmd.exe` prior building results in a warning being returned by `ninja` in every file being compiled.

caribpa:
Windows doesn't seem to require setting `CFLAGS` and `LDFLAGS` to `-static`, just passing `--default-library=static` to meson is enough.

ret2libc:
> Windows doesn't seem to require setting `CFLAGS` and `LDFLAGS` to `-static`, just passing `--default-library=static` to meson is enough.

Are you talking about msvc? If yes, then please see https://github.com/rizinorg/rizin/pull/749/files#diff-30d8f6be6320feeacf686be94f48c70869b52630e01ea625f0f15adc0d57c3e4R113 .

caribpa:
My bad, I thought the meson `has_link_argument` checked if meson was called with `LDFLAGS="<flag>"`, instead of checking if the linker accepted the `<flag>` without warning or error (I found that information in the [meson reference](https://mesonbuild.com/Reference-manual.html)).

Then yes, I see no conflicts with Windows.

ret2libc:
> My bad, I thought the meson `has_link_argument` checked if meson was called with `LDFLAGS="<flag>"`, instead of checking if the linker accepted the `<flag>` without warning or error (I found that information in the [meson reference](https://mesonbuild.com/Reference-manual.html)).
> 
> Then yes, I see no conflicts with Windows.

Just to be sure, before merging this I will double-check on my Windows machine.

ret2libc:
@caribpa I also added a section for 32-bit compilation on 64-bit machines. See https://github.com/rizinorg/rizin/pull/749/files#diff-f1355fd54329147c830843ab133fed7716180c3242031487b861a2a3235d0078R143 however I'm now thinking that it might be worth being a bit more specific, because in some cases the environment is already all set to build 32-bit binaries (e.g. when running the x86-32 visual studio tools shell), so there is no need to create any cross-compile file.

caribpa:
I think that having cross-files already present in the repo would be really useful. When I looked on the internet about cross compilation for meson I found that many projects used cross-files (and the [official meson cross-compilation reference](https://mesonbuild.com/Cross-compilation.html) encourages their use), so we can't go wrong if we provide them.

Note that they could also be used for the CI testing pipeline, so that we catch errors and warnings for other archs early.

ret2libc:
> I think that having cross-files already present in the repo would be really useful. When I looked on the internet about cross compilation for meson I found that many projects used cross-files (and the [official meson cross-compilation reference](https://mesonbuild.com/Cross-compilation.html) encourages their use), so we can't go wrong if we provide them.
> 
> Note that they could also be used for the CI testing pipeline, so that we catch errors and warnings for other archs early.

Unfortunately it is impossible to provide any useful cross-compilation file, because it would depend on way too many things. Ubuntu provides some files, Fedora others, one wants to use a specific compiler, another user wants to use another setting, etc. etc. You cannot provide pre-done files. If you want to cross compile, let's say you are expected to know at least a bit your way. There are already examples for Android cross-compilation and i686, so I think everyone should be able to adapt them for their needs.

caribpa:
Well, yes, I understand that cross-files are there for targeting specific platforms and cannot be generalized. What I meant was to have, for example a `linux-x86_64`, `linux-i386`, `windows-x86_64`, `windows-i386`, etc for the mainstream platforms. This said, I am still not sure about all the dependencies of the project and if it is even possible to generalize, for example, all "linux" platforms in one file.
According to [this answer](https://stackoverflow.com/a/48714666), meson is able to find the dependencies from the `PKG_CONFIG_LIBDIR` variable, so maybe it is just a matter of telling the users to simply change the path of that variable to be able to support (for example) all "linux" platforms?

ret2libc:
> Well, yes, I understand that cross-files are there for targeting specific platforms and cannot be generalized. What I meant was to have, for example a `linux-x86_64`, `linux-i386`, `windows-x86_64`, `windows-i386`, etc for the mainstream platforms. This said, I am still not sure about all the dependencies of the project and if it is even possible to generalize, for example, all "linux" platforms in one file.

As said, even to have a linux-i386 cross-compile file is not possible easily :) there are just way too many variables. For example, are you cross-compiling to linux-i386 from x86-64 windows or linux? From arm or from x86_64? Are you using tools installed from your distribution to cross-compile or you are downloading them yourself? Are you using ubuntu, arch, fedora, rhel or what? There's just no point in having a file that would need to be changed 100% of the times. The instructions, I believe, are more than enough. If you(generic) want to cross-compile, take 3 mins to copy what we provide already and adapt it to your needs. We expect you to know your machine, your compiler, etc.

> According to [this answer](https://stackoverflow.com/a/48714666), meson is able to find the dependencies from the `PKG_CONFIG_LIBDIR` variable, so maybe it is just a matter of telling the users to simply change the path of that variable to be able to support (for example) all "linux" platforms?

It's not that simple AFAIK.

caribpa:
Understood, sorry for the noobness. I don't know meson nor how it does its thing so I imagined that it was able to do magic with effortless changes 😄

XVilka (approved):

XVilka:
Usually for cross-compilation it's easier to use containers, like we do with Alpine for Linux static builds.
Back in pre-container times I usually used either [`crosstool-ng`](https://crosstool-ng.github.io/) or [`buildroot`](https://buildroot.org/) for these cases.

ret2libc:
> Usually for cross-compilation it's easier to use containers, like we do with Alpine for Linux static builds.
> Back in pre-container times I usually used either [`crosstool-ng`](https://crosstool-ng.github.io/) or [`buildroot`](https://buildroot.org/) for these cases.

Yeah it may be worth re-writing the cross-compilation part to use dockcross, crosstool-ng or one of these similar tools.

ret2libc:
See https://github.com/rizinorg/rizin/pull/756 for final testing.

ret2libc:
Ok, dist- works.

PR #750, opened 2021-03-02
Prevent signed int difference overflow in java_switch_op()

kazarmy:

**Detailed description**

This pr is a fix for the broken fuzz tests in https://github.com/rizinorg/rizin/runs/1994325466 of #260. An explanation for the fix is [here](https://stackoverflow.com/questions/8024480/computing-integer-absolute-differences-in-overflow-safe-ways/8024536#8024536).

**Test plan**

All builds are green.

**Closing issues**

...

### DIFF
--- a/librz/analysis/p/analysis_java.c
+++ b/librz/analysis/p/analysis_java.c
@@ -55,7 +55,7 @@ static int java_switch_op(RzAnalysis *analysis, RzAnalysisOp *op, ut64 addr, con
 		ut32 default_loc = (ut32)(rz_read_at_be32(data, pos)), cur_case = 0;
 		op->switch_op = rz_analysis_switch_op_new(addr, min_val, max_val, default_loc);
 		pos += 12;
-		if (max_val > min_val && ((max_val - min_val) < (UT16_MAX / 4))) {
+		if (max_val > min_val && (((ut32)max_val - (ut32)min_val) < (UT16_MAX / 4))) {
 			//caseop = rz_analysis_switch_op_add_case(op->switch_op, addr+default_loc, -1, addr+offset);
 			for (cur_case = 0; cur_case <= max_val - min_val; pos += 4, cur_case++) {
 				//ut32 value = (ut32)(rz_read_at_be32 (data, pos));


librz/analysis/p/analysis_java.c
```diff
@@ -55,7 +55,7 @@ static int java_switch_op(RzAnalysis *analysis, RzAnalysisOp *op, ut64 addr, con
 		ut32 default_loc = (ut32)(rz_read_at_be32(data, pos)), cur_case = 0;
 		op->switch_op = rz_analysis_switch_op_new(addr, min_val, max_val, default_loc);
 		pos += 12;
-		if (max_val > min_val && ((max_val - min_val) < (UT16_MAX / 4))) {
+		if (max_val > min_val && (((ut32)max_val - (ut32)min_val) < (UT16_MAX / 4))) {
```
ret2libc, line 58: what is this check for exactly? Can't we just have `const ut32 min_val, max_val` instead of casting here? Again, what is the real purpose of this check?

kazarmy: Referring to https://docs.oracle.com/javase/specs/jvms/se15/html/jvms-6.html#jvms-6.5.tableswitch ...

> what is this check for exactly?

the check apparently prevents processing of Java jump tables that have more than `UT16_MAX / 4` + 1 entries, probably because they are considered improbably large and could cause the analysis to take a long time.

> Can't we just have `const ut32 min_val, max_val` instead of casting here?

Apparently not ... the spec above really actually states that they are signed.

> Again, what is the real purpose of this check?

See first answer above.

ret2libc: I see.

Shouldn't we check that both max_val and min_val are greater than 0 then? Otherwise e.g. if you have a negative `min_val`, the check `max_val > min_val` would always be true.

kazarmy: The check `max_val > min_val` is done using signed integers, so if I have a negative `min_val`, the check might not be true (e.g. if I have a negative `max_val` as well).

Note that the above is based on the logic of the spec, but if the spec implies that it is possible, well what can I say?

> the check apparently prevents processing of Java jump tables that have more than UT16_MAX / 4 + 1 entries, probably because they are considered improbably large and could cause the analysis to take a long time.

Btw, very large jump tables can also cause memory exhaustion, which is probably why the fuzz tests were failing.

ret2libc: OkOk, proceed as you think it's better :)

wargio (approved):
LGTM. does not impact my work (for now)

PR #778, opened 2021-03-06
Fix memory leak in rz_skiplist_insert()

p371jah:

**Detailed description**

...

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/librz/util/skiplist.c
+++ b/librz/util/skiplist.c
@@ -162,6 +162,9 @@ RZ_API RzSkipListNode *rz_skiplist_insert(RzSkipList *list, void *data) {
 	x = find_insertpoint(list, data, update, true);
 	// check whether the element is already in the list
 	if (x != list->head && !list->compare(x->data, data)) {
+		if (list->freefn && data) {
+			list->freefn(data);
+		}
 		return x;
 	}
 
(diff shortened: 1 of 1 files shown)


thestr4ng3r:
Wouldn't it perhaps make more sense to free the `data` that is currently in the list and replace it with the new one?

ret2libc:
A description of the PR with the mem leak it is trying to solve would help to understand if the fix is ok or not.

ret2libc (commented):
Maybe the function should just return NULL if the element already exists? Otherwise it replaces data with another data that is essentially the same because the compare function returns 0.

p371jah:
> Maybe the function should just return NULL if the element already exists? Otherwise it replaces data with another data that is essentially the same because the compare function returns 0.

So how to distinguish memory allocation error and duplication?

ret2libc:
> > Maybe the function should just return NULL if the element already exists? Otherwise it replaces data with another data that is essentially the same because the compare function returns 0.
> 
> So how to distinguish memory allocation error and duplication?

If you have a function that only does insert any duplication is an error, in some sense :D Do we need to differentiate the two?

XVilka:
@pelijah @ret2libc @thestr4ng3r  what was the decision on this one?

PR #793, opened 2021-03-08
Ensure command substitution can provide multiple arguments to some statements

### LINKED ISSUE #773: `/a jmp rsp @{$(dmm ~rizin[0,1])}` is not working
By the way @ret2libc, do you know why `/a jmp rsp @{$(dmm ~rizin[0,1])}` is not working (and separate shell expansions are required, like showed in the OP)?

_Originally posted by @caribpa in https://github.com/rizinorg/rizin/issues/766#issuecomment-791398622_

---

In grammar.js there is:
```
prec.right(1, seq($._simple_command, "@{", alias($.arg_brace, $.arg), alias($.arg_brace, $.arg), "}")),
```
thus the grammar expects two arguments between `@{...}`. However `${....)` is considered as a single argument (at least before the command expansion is performed) so the above command fails to be parsed.

A solution is probably to accept a variable number of arguments between `@{...}` and check  if there are exactly 2 in the `tmp_fromto_command` handler, which is executed after the parser has substituted the `$(..)` with the output.

ret2libc:

**Detailed description**

* go back to `@(<from> <to>)` syntax to search in a given range, instead of `@{...}` which was introduced in radare2 not long ago to ease the parsing of oldshell, however it required a lot of special handling in newshell. This simplifies the grammar without any special drawback I can see
* allow one to use command substitution in places like `<cmd> @($(?e 1 2))` or `<cmd> @@s:$(?e 1 2 3)`, etc. This is done by moving the validation to the statement handlers, instead of doing everything in the grammar itself. That means that the parser will accept any number of arguments for `@(` or for `@@s:` but the *statement* handlers will ensure the right number is passed to them (this happens after the command substitution is done).

**Test plan**

...

**Closing issues**

Closes #773


ret2libc:
cc @caribpa

XVilka:
Finally a syntax bit familiar to those who does bash/zsh programming.

XVilka:
Please don't forget to add a paragraph or two in Rizin book after.

caribpa:
> Finally a syntax bit familiar to those who does bash/zsh programming.

Well, the `@()` is indeed also used in bash and zsh, but for a different purpose, if `extglob` is enabled, `@()` matches only one of the given shell patterns (full documentation [here](https://www.gnu.org/software/bash/manual/html_node/Pattern-Matching.html)), for example `@(rizin|rz-gg|ls*)` 🙂

ret2libc:
> Please don't forget to add a paragraph or two in Rizin book after.

Where exactly... @(..) was never mentioned in the book, so I'm not sure. There would be also many other things to do to the book. I'd prefer to have better documentation for each `@` in rizin itself (e.g. `@{??` to get more details and example about `@{` etc.)

XVilka:
@ret2libc Likely in Scripting section, here: https://book.rizin.re/scripting/intro.html

XVilka (approved):

PR #795, opened 2021-03-08
Reuse analysis data in rz_core_analysis_type_match()

p371jah:

**Detailed description**

With these changes `aaft` takes 53s VS 93s for 18.7MB binary. 

...

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/librz/core/analysis_tp.c
+++ b/librz/core/analysis_tp.c
@@ -266,6 +266,19 @@ static void retype_callee_arg(RzAnalysis *analysis, const char *callee_name, boo
 	}
 }
 
+RzAnalysisOp *op_cache_get(HtUP *cache, RzCore *core, ut64 addr) {
+	RzAnalysisOp *op = ht_up_find(cache, addr, NULL);
+	if (op) {
+		return op;
+	}
+	op = rz_core_analysis_op(core, addr, RZ_ANALYSIS_OP_MASK_BASIC | RZ_ANALYSIS_OP_MASK_VAL);
+	if (!ht_up_insert(cache, addr, op)) {
+		rz_analysis_op_free(op);
+		return NULL;
+	}
+	return op;
+}
+
 #define DEFAULT_MAX  3
 #define REGNAME_SIZE 10
 #define MAX_INSTR    5
@@ -282,7 +295,7 @@ static void retype_callee_arg(RzAnalysis *analysis, const char *callee_name, boo
  * \param caddr addr of the callee
  */
 static void type_match(RzCore *core, char *fcn_name, ut64 addr, ut64 baddr, const char *cc,
-	int prev_idx, bool userfnc, ut64 caddr) {
+	int prev_idx, bool userfnc, ut64 caddr, HtUP *op_cache) {
 	Sdb *trace = core->analysis->esil->trace->db;
 	Sdb *TDB = core->analysis->sdb_types;
 	RzAnalysis *analysis = core->analysis;
@@ -349,15 +362,12 @@ static void type_match(RzCore *core, char *fcn_name, ut64 addr, ut64 baddr, cons
 			if (instr_addr < baddr) {
 				break;
 			}
-			RzAnalysisOp *op = rz_core_analysis_op(core, instr_addr, RZ_ANALYSIS_OP_MASK_BASIC | RZ_ANALYSIS_OP_MASK_VAL);
+			RzAnalysisOp *op = op_cache_get(op_cache, core, instr_addr);
 			if (!op) {
-				rz_analysis_op_free(op);
 				break;
 			}
-			RzAnalysisOp *next_op = rz_core_analysis_op(core, instr_addr + op->size, RZ_ANALYSIS_OP_MASK_BASIC | RZ_ANALYSIS_OP_MASK_VAL);
+			RzAnalysisOp *next_op = op_cache_get(op_cache, core, instr_addr + op->size);
 			if (!next_op || (j != idx && (next_op->type == RZ_ANALYSIS_OP_TYPE_CALL || next_op->type == RZ_ANALYSIS_OP_TYPE_JMP))) {
-				rz_analysis_op_free(op);
-				rz_analysis_op_free(next_op);
 				break;
 			}
 			RzAnalysisVar *var = rz_analysis_get_used_function_var(analysis, op->addr);
@@ -434,8 +444,6 @@ static void type_match(RzCore *core, char *fcn_name, ut64 addr, ut64 baddr, cons
 					__var_retype(analysis, var, name, type ? type : "int", memref, false);
 				}
 			}
-			rz_analysis_op_free(op);
-			rz_analysis_op_free(next_op);
 		}
 		size += analysis->bits / 8;
 		free(type);
@@ -449,10 +457,13 @@ static int bb_cmpaddr(const void *_a, const void *_b) {
 	return a->addr > b->addr ? 1 : (a->addr < b->addr ? -1 : 0);
 }
 
+void free_op_cache_kv(HtUPKv *kv) {
+	rz_analysis_op_free(kv->value);
+}
+
 RZ_API void rz_core_analysis_type_match(RzCore *core, RzAnalysisFunction *fcn) {
 	RzAnalysisBlock *bb;
 	RzListIter *it;
-	RzAnalysisOp aop = { 0 };
 	bool resolved = false;
 
 	rz_return_if_fail(core && core->analysis && fcn);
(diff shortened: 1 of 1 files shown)


XVilka:
Good idea, we might cache more stuff in RzAnalysis and upcoming RzTypes module.

XVilka (approved):

librz/core/analysis_tp.c
```diff
@@ -266,6 +266,29 @@ static void retype_callee_arg(RzAnalysis *analysis, const char *callee_name, boo
 	}
 }
 
+static RzAnalysisOp *op_cache_seek(RzPVector *cache, ut64 addr, ssize_t *p_idx) {
+	ssize_t idx = *p_idx;
+	if (rz_pvector_empty(cache)) {
+		return NULL;
+	}
+	for (int i = 0; i < 10 && idx >= 0; i++, idx--) {
```
ret2libc: isn't it faster to have a HT instead of a sorted Vector?

librz/core/analysis_tp.c
```diff
@@ -343,21 +367,45 @@ static void type_match(RzCore *core, char *fcn_name, ut64 addr, ut64 baddr, cons
...
+		ssize_t cache_idx = cache_len - 1;
 		// Backtrace instruction from source sink to prev source sink
-		for (j = idx; j >= prev_idx; j--) {
+		for (j = idx; j >= prev_idx; j--, cache_idx--) {
 			ut64 instr_addr = sdb_num_get(trace, sdb_fmt("%d.addr", j), 0);
 			if (instr_addr < baddr) {
 				break;
 			}
-			RzAnalysisOp *op = rz_core_analysis_op(core, instr_addr, RZ_ANALYSIS_OP_MASK_BASIC | RZ_ANALYSIS_OP_MASK_VAL);
+			RzAnalysisOp *op = NULL;
```
ret2libc: it seems to me the "get op from cache if available, otherwise call rz_core_analysis_op" can be refactored in a separate function

librz/core/analysis_tp.c
```diff
@@ -535,44 +589,77 @@ RZ_API void rz_core_analysis_type_match(RzCore *core, RzAnalysisFunction *fcn) {
...
+					break;
+				}
+			}
+			if (aop_cached) {
+				ret = aop->size;
+			} else {
+				ret = rz_analysis_op(analysis, aop, addr, buf + i, bsize - i, RZ_ANALYSIS_OP_MASK_BASIC | RZ_ANALYSIS_OP_MASK_VAL);
+				if (ret <= 0) {
+					i += minopcode;
+					addr += minopcode;
```
ret2libc: shouldn't we add `addr` even when the data is taken from the cache?

ret2libc (commented):
To summarize my doubts:
1) Ht instead of Vector. What do you think?
2) function to "get from cache or analyze with rz_analysis_op AND put in the cache immediately"

librz/core/analysis_tp.c
p371jah: No, we don't need random access. Actually this loop stops after 2 or even 1 iteration.

librz/core/analysis_tp.c
p371jah: Yes, this function checks last entries of the vector and sets start index.

p371jah: If analysis data for a given addr is not presented then `rz_core_analysis_op()` is called

librz/core/analysis_tp.c
p371jah: Ofc, it is done in the end of the loop body. BTW it is not my code.

ret2libc: Oh i see, sorry

librz/core/analysis_tp.c
ret2libc: why? You are looking for an entry in the cache with `op->addr == addr`, right? So you want to have immediate access to the element which has `addr` as key, don't you?

p371jah: As I understood `type_match()` analyzes instructions of a basic block in reverse order. The vector (aka cache) contains analysis data of all instructions of a BB. So we can just iterate over the vector in a reverse order. It is much faster than `ht_up_find()`, isn't it? And moreover we need to get two entries from the cache at every iteration. In case of a vector we can just use `vector[i]` + `vector[i+1]`. Why `op_cache_seek()` is needed? The vector can contain some entries which were not added to ESIL trace. So `op_cache_seek()` mostly just returns `vector[cache_idx]` but it can also decrease `cache_idx` value to skip some entries.

p371jah:
@ret2libc 
1. As I said HT is not good for that case
2. I doubt that this function can make `rz_core_analysis_type_match()` look better

ret2libc:
>     1. As I said HT is not good for that case

Maybe it is not, I'm still failing to understand everything in this function :(

>     2. I doubt that this function can make `rz_core_analysis_type_match()` look better

On this I disagree. `type_match` is already a quite long/complicated function and in its current state this PR is just making things even harder to read/understand because the cache seems too complicated to use. If we want to have a cache, it should be well wrapped in an API with cache_get/cache_set. Right now it is a mix of use `cache_seek` and manual looking into the cache, with magic checks to see if the element got from the cache is actually what we wanted... Then there's a free/no-free approach based on whether the element was in cache or not.

Just some proposals:
- can't you put in the cache opcodes that are not already in the cache so that you don't have to keep track of what is cached and what not? This way you don't free anything and you let the cache vector free all items at the end. Can the cache grow too big in this case?
- can you add a `cache_seek_next` separated function to get the next element in the cache? Users of the cache functions should not be that involved in dealing with indexes, etc.
- can we get rid of that `if (not cache) add to cache` and just move it inside the cache_seek function so that when an item is not found in the cache it is automatically added?

p371jah:
> If we want to have a cache, it should be well wrapped in an API with cache_get/cache_set.

We need to use some sort of cache in `rz_core_analysis_type_match()` because analysis stuff is repeated 2-4 times for every instruction. Well I can move some code in separate functions but `type_match()` and `rz_core_analysis_type_match()` will remain complicated.

> can't you put in the cache opcodes that are not already in the cache so that you don't have to keep track of what is cached and what not?

I expect that all ops are pushed to the vector. But `rz_pvector_push()` can fail so we need to track what is cached to avoid memory leaks. Memory leak can also happen in case of cache miss in `type_match()`.

> Users of the cache functions should not be that involved in dealing with indexes, etc.

Whole `type_match()` work is based on indexes. And what do you mean under `Users`? It is internal cache inside `rz_core_analysis_type_match()`.

ret2libc:
> > If we want to have a cache, it should be well wrapped in an API with cache_get/cache_set.
> 
> We need to use some sort of cache in `rz_core_analysis_type_match()` because analysis stuff is repeated 2-4 times for every instruction. Well I can move some code in separate functions but `type_match()` and `rz_core_analysis_type_match()` will remain complicated.

That's alright, but at least we won't complicate it further.

> > can't you put in the cache opcodes that are not already in the cache so that you don't have to keep track of what is cached and what not?
> 
> I expect that all ops are pushed to the vector. But `rz_pvector_push()` can fail so we need to track what is cached to avoid memory leaks. Memory leak can also happen in case of cache miss in `type_match()`.

If something can't be put in the cache, skip the opcode or return from the function completely. If _push fails it's likely we used all memory (still need to see that happens, but anyway..), so probably it's not worth going forward anyway.

> > Users of the cache functions should not be that involved in dealing with indexes, etc.
> 
> Whole `type_match()` work is based on indexes. And what do you mean under `Users`? It is internal cache inside `rz_core_analysis_type_match()`.

I mean callers of the API.

thestr4ng3r:
Sorry to bother you so much but I agree with @ret2libc that since the type matching code is so complicated and fragile, we should really think twice about every new line added there.

>  * can we get rid of that `if (not cache) add to cache` and just move it inside the cache_seek function so that when an item is not found in the cache it is automatically added?

I think this is still an important point. At the moment there is still a lot of `if (!op || op->addr != instr_addr) ...` in the analysis code which should completely be handled by the cache.
Before I go on, to make sure I have understood this correctly: are these cache misses cases actually expected to happen in practice depending on the analyzed code or could they only happen in the case of memory exhaustion or a programming error on rizin side?

Assuming it is a real expected case, this whole logic should be done by the cache imo. Then making sure not to leak or double free is not trivial of course but here is my idea how I would approach this:
* The cache maintains its vector as normal so you have the seek forward/backwards functionality
* In addition, the cache keeps a hashset of pointers to all ops that are currently in the cache
* When you request an op from the cache, it first does the regular cache-retrieval if possible and if it fails, it allocates+fills a new op and returns that
* Whenever you don't need an op returned in the above point anymore, you call a dedicated free method of the cache, which will check the pointer against the hashset and free it if it is not contained, or otherwise just do nothing.

In any case, I think these new caching data structures need concrete documentation for their purpose and all their methods because right now it is really not obvious to see what purpose they have exactly.
Also we must make sure that all the relevant code paths are covered by tests, especially for the memory handling. (This should be up to date after a rebase I think: https://app.codecov.io/gh/rizinorg/rizin/compare/795/diff)

p371jah:
I hope this implementation is simple enough. Well it is slightly slower than previous (<5%) but anyway ok for now.

thestr4ng3r (approved):
That works for me too :-)

XVilka (approved):

XVilka:
Please don't do too many changes in the types code - I am working on these: https://github.com/rizinorg/rizin/projects/3 and it involves a lot of refactoring.

Issue #676, opened 2021-02-21
Long loading time when working with large executables


konatabrk:
### Work environment

| Questions                                            | Answers
|------------------------------------------------------|--------------------
| OS/arch/bits (mandatory)                             | Debian 5.9.1-1 x86_64
| File format of the file you reverse (mandatory)      | ELF
| Architecture/bits of the file (mandatory)            | x86/64
| `rizin -v` full output, **not truncated** (mandatory)  | rizin 0.2.0-git @ linux-x86-64 commit: 0d5e708da4d836fa25245420abb8ca4804075332, build: 2021-02-21__09:13:15

### Expected behavior

1. rizin loads the binary without noticeable time delay.
2. rizin doesn't crash with a segmentation fault when launching `ood` command with additional argv parameters

### Actual behavior

1. rizin freezes for 4-5 minutes when trying to load the binary
2. When trying to re-run the binary using `ood` command with additional argv parameters, rizin falls with a segmentation fault

### Steps to reproduce the behavior
Binary with the needed libraries are placed in the [attachment](https://drive.google.com/drive/folders/1zNBQC3hwSzkWSLSLX3XTlOoMmmul3icq?usp=sharing) as tar archive

### Additional Logs, screenshots, source code,  configuration dump, ...

More details in `readme.txt` file inside of the [attached archive](https://drive.google.com/drive/folders/1zNBQC3hwSzkWSLSLX3XTlOoMmmul3icq?usp=sharing)

I guess loading time depends on the size of the target binary file, but gdb loads it like in 10s and then after re-running it doesn't analyze/load binary again. In contrast, when using rizin, it loads and analyses the binary after each `ood` command which is unnecessary and time consuming.

Also, I need to notice that the binary is instrumented with AFL and ASAN at compilation time

- [rizin_testcase](https://drive.google.com/drive/folders/1zNBQC3hwSzkWSLSLX3XTlOoMmmul3icq?usp=sharing)

XVilka:
I also ran `aaa` on that binary under ASAN, there are some warnings that might be needed to fix:
```
[0x008e55a0]> aaa
[Invalid function name 'sym.asan.module_ctor' at 0x00992160
Invalid function name 'sym.asan.module_dtor' at 0x00992180
Invalid function name 'sym.asan.module_ctor' at 0x0113bee0
Invalid function name 'sym.asan.module_dtor' at 0x0113bf00
Invalid function name 'sym.asan.module_ctor' at 0x01176d90
[...]

PR #796, opened 2021-03-08
Dwarf Refactoring

thestr4ng3r:
**DO NOT SQUASH**

The individual commits were made conciously so you can look at them separately if you want to see the different steps in a clearer way.

This is a major refactor of some dwarf parts.
Along with some minor stuff, the overall changes in this pr are:
* make dwarf parsing local to an RzBinFile (could be later made even localer to RzBinObject)
* remove all printing and setting from `dwarf.c` so it really is a pure parser
* put all printing in `cdwarf.c` which is in core
* for dwarf line info (this is kind of a small language of instructions that when executed yields the line info in some registers), clearly split parsing of ops, running the ops and printing them in a readable way. It is however still possible to simultaneously parse and run them as part of the high-level loading process to avoid having to pull every op into memory.

This is also the foundation for #843 since now the accesses to this sdb are much more controlled.


XVilka:
Skimmed through and overall looks fine 👍

librz/bin/dwarf.c
```diff
@@ -1079,102 +1099,85 @@ static int parse_line_raw(const RzBin *a, const ut8 *obuf,
...
+		set->unit_length = unit_length;
+		set->is_64bit = is_64bit;
+		set->version = READ16(buf);
+		set->debug_info_offset = dwarf_read_offset(set->is_64bit, big_endian, &buf, buf_end);
+		set->address_size = READ8(buf);
+		set->segment_size = READ8(buf);
+		unit_length -= header_rest_size;
+
+		// align to 2*addr_size
+		size_t off = buf - start;
```
XVilka, line 1098: Maybe extract unto the microfunction in RzUtil's RzNum?

librz/include/rz_bin.h
```diff
@@ -295,7 +295,7 @@ struct rz_bin_file_t {
 	RzList *xtr_data;
 	Sdb *sdb;
 	Sdb *sdb_info;
-	Sdb *sdb_addrinfo;
+	RZ_DEPRECATE Sdb *sdb_addrinfo; //< deprecated to use for new code, should be refactored
```
XVilka, line 300: Just for the record, the issue was opened here: https://github.com/rizinorg/rizin/issues/843

XVilka:
You might also try to check with this file, since you are rewriting DWARF, it's the main cause of the slowness in that issue too: https://github.com/rizinorg/rizin/issues/676

XVilka:
Is this expected outcome or a bug (looks like a bug to me)?
```
[XX] db/cmd/cmd_pde pde ppc
RZ_NOPLUGINS=1 rizin -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -Qc 'e asm.comments=false
e asm.bytes=false
pde
' bins/elf/hello.ppc
-- stdout
@@ -3,7 +3,7 @@
 ;-- .text:
 ;-- _start:
 ;-- pc:
-0x10000308      mr    r9, r1                                           ; start.S:62
+0x10000308      mr    r9, r1                                           ; crtn.S:46
 0x1000030c      rlwinm r1, r1, 0, 0, 0x1b                              ; start.S:64
 0x10000310      li    r0, 0                                            ; start.S:69
 0x10000314      stwu  r1, -0x10(r1)                                    ; start.S:71
```

thestr4ng3r:
> You might also try to check with this file, since you are rewriting DWARF, it's the main cause of the slowness in that issue too: #676

This is a very good stress test for this indeed. It's a 600MB elf full of dwarf and it shows how bad the choice of sdb for `sdb_addrline` and the fact that the filenames are replicated in every `RzBinSourceRow` (previously `RzBinDwarfRow`) really are. This PR will not fix this, but it prepares everything such that this refactor can be done easily.

thestr4ng3r:
> Is this expected outcome or a bug (looks like a bug to me)?
> 
> ```
> [XX] db/cmd/cmd_pde pde ppc
> RZ_NOPLUGINS=1 rizin -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -Qc 'e asm.comments=false
> e asm.bytes=false
> pde
> ' bins/elf/hello.ppc
> -- stdout
> @@ -3,7 +3,7 @@
>  ;-- .text:
>  ;-- _start:
>  ;-- pc:
> -0x10000308      mr    r9, r1                                           ; start.S:62
> +0x10000308      mr    r9, r1                                           ; crtn.S:46
>  0x1000030c      rlwinm r1, r1, 0, 0, 0x1b                              ; start.S:64
>  0x10000310      li    r0, 0                                            ; start.S:69
>  0x10000314      stwu  r1, -0x10(r1)                                    ; start.S:71
> ```

Yes it's a bug. It was present before too, but now the loading order changed to a certain degree so it only started to show now.

librz/bin/dbginfo.c
```diff
@@ -5,6 +5,14 @@
 #include <rz_types.h>
 #include <rz_bin.h>
 
+RZ_API void rz_bin_source_row_free(RzBinSourceRow *row) {
+	if (!row) {
```
wargio, line 9: rz_return_if_false

librz/bin/dwarf.c
```diff
@@ -449,133 +468,125 @@ static const ut8 *parse_line_header_source(RzBinFile *bf, const ut8 *buf, const
...
 			}
-			hdr->file_names_count = count;
-			buf = tmp_buf;
-			count = 0;
 		}
+	} else {
+		include_dir = sdb_const_get(sdb_addrinfo, "DW_AT_comp_dir", 0);
+	}
+	if (!include_dir) {
+		include_dir = "./";
```
wargio, line 575: what if you have windows?

librz/bin/dwarf.c
```diff
@@ -449,133 +468,125 @@ static const ut8 *parse_line_header_source(RzBinFile *bf, const ut8 *buf, const
...
+	}
+	RzBinDwarfLineFileEntry *file = &header->file_names[file_index];
+	rz_return_val_if_fail(header && file, NULL);
+	const char *include_dir = NULL, *comp_dir = NULL;
+	if (file->id_idx > 0 && file->id_idx - 1 < header->include_dirs_count) {
+		include_dir = header->include_dirs[file->id_idx - 1];
+		if (include_dir && include_dir[0] != '/') {
+			comp_dir = sdb_const_get(sdb_addrinfo, "DW_AT_comp_dir", 0);
+			if (comp_dir) {
+				include_dir = rz_str_newf("%s/%s/", comp_dir, include_dir);
```
wargio, line 568: windows?

librz/bin/dwarf.c
```diff
@@ -635,324 +623,140 @@ static const ut8 *parse_line_header(
...
-	case DW_LNE_define_file:
-		filename = (const char *)buf;
-
-		if (mode == RZ_MODE_PRINT) {
-			print("define_file\n");
-			print("filename %s\n", filename);
+	}
+	case DW_LNE_define_file: {
+		size_t fn_len = rz_str_nlen((const char *)buf, buf_end - buf);
+		char *fn = malloc(fn_len + 1);
```
wargio, line 704: check on `fn`?

librz/bin/dwarf.c
```diff
@@ -1030,151 +953,143 @@ static int parse_line_raw(const RzBin *a, const ut8 *obuf,
...
+		}
+		buf += pad;
+		unit_length -= pad;
 
-	while ((buf - obuf) < len) {
-		ut64 adr, length;
-		if ((idx + 8) >= len) {
+		size_t arange_size = 2 * set->address_size;
+		set->aranges_count = unit_length / arange_size;
+		set->aranges = calloc(set->aranges_count, sizeof(RzBinDwarfARange));
```
wargio: RZ_NEWS0

wargio: also what would happen if `set->aranges_count` is zero?

librz/bin/dwarf.c
```diff
@@ -2191,193 +1955,128 @@ static ut8 *get_section_bytes(RzBin *bin, const char *sect_name, size_t *len) {
...
-			size_t i, j;
-			for (i = 0; i < info->count; i++) {
-				RzBinDwarfCompUnit *unit = &info->comp_units[i];
-				for (j = 0; j < unit->count; j++) {
-					RzBinDwarfDie *die = &unit->dies[j];
-					ht_up_insert(info->lookup_table, die->offset, die); // optimization for further processing}
-				}
+	ut64 len = section->size;
+	// what is this checking for?
+	if (len > (UT32_MAX >> 1) || len < 1) {
```
wargio: (answering to the comment) sanitize input, i guess?

librz/bin/dwarf.c
```diff
@@ -2191,193 +1955,128 @@ static ut8 *get_section_bytes(RzBin *bin, const char *sect_name, size_t *len) {
...
+		goto cave_debug_str_buf;
+	}
+	ut8 *buf = calloc(1, len);
```
wargio: RZ_NEW0 ?

librz/bin/dwarf.c
```diff
@@ -2191,193 +1955,128 @@ static ut8 *get_section_bytes(RzBin *bin, const char *sect_name, size_t *len) {
...
-				}
-				free(file);
-			}
-		}
-		ls_free(ls);
+	ut64 len = section->size;
+	if (len < 1) {
+		return NULL;
+	}
+	ut8 *buf = calloc(1, len + 1);
```
wargio: RZ_NEW0?

librz/bin/dwarf.c
```diff
@@ -2191,193 +1955,128 @@ static ut8 *get_section_bytes(RzBin *bin, const char *sect_name, size_t *len) {
...
+	RzBinSection *section = getsection(binfile, "debug_aranges");
+	if (!section) {
+		return NULL;
+	}
 
+	len = section->size;
+	if (len < 1 || len > ST32_MAX) {
+		return NULL;
+	}
+	buf = calloc(1, len);
```
wargio: RZ_NEW0

librz/bin/dwarf.c
```diff
@@ -2191,193 +1955,128 @@ static ut8 *get_section_bytes(RzBin *bin, const char *sect_name, size_t *len) {
...
-		if (!buf) {
-			goto cleanup;
-		}
-		if (!rz_buf_read_at(binfile->buf, section->paddr, buf, len)) {
-			free(buf);
-			goto cleanup;
+	RzBinSection *debug_str = debug_str = getsection(binfile, "debug_str");
+	if (debug_str) {
+		debug_str_len = debug_str->size;
+		debug_str_buf = calloc(1, debug_str_len + 1);
```
wargio: RZ_NEW0 ?

wargio (changes requested):

librz/bin/dwarf.c
thestr4ng3r: Dwarf standard does not seem to specify the exact format of paths so apparently it is indeed target-dependent. However the one Windows binary we have that also contains dwarf, which is `test/bins/pe/hello_world_not_stripped.exe`, contains path names like `/private/tmp/mingw-w64-20180808-82873-149eclj/mingw-w64-v5.0.4/mingw-w64-crt/build-i686/../crt//crtexe.c` with slashes so if we really encounter a file that has backslashes there it will probably be better to only use slashes in this place and later just treat slashes and backslashes equally in the result because we can't really detect that.

librz/bin/dwarf.c
thestr4ng3r: The comment was from old code, I think it was referring specifically to the `UT32_MAX` check which I don't understand either. I'll remove that and only check for `len` since that makes sense.

librz/bin/dwarf.c
XVilka: Just use `RZ_PATH_JOIN`

XVilka (approved):

librz/bin/dwarf.c
thestr4ng3r: `RZ_JOIN_2_PATHS` would be wrong here since it takes the separator of the host where rizin runs on rather than the one of the target binary. See also https://github.com/rizinorg/rizin/pull/796#discussion_r597545609

librz/bin/dwarf.c
XVilka: We need more samples then for MinGW-compiled files with DWARF then. But this is for another PR, imho.

librz/bin/dwarf.c
```diff
@@ -340,6 +333,44 @@ static const char *dwarf_unit_types[] = {
...
+
+RZ_API const char *rz_bin_dwarf_get_unit_type_name(ut64 unit_type) {
+	if (!unit_type || unit_type > DW_UT_split_type) {
+		return NULL;
+	}
+	return dwarf_unit_types[unit_type];
+}
+
+RZ_API const char *rz_bin_dwarf_get_lang_name(ut64 lang) {
+	if (lang >= sizeof(dwarf_langs) / sizeof(*dwarf_langs)) {
```
wargio: i belive there is a macro for this division. `RZ_ARRAY_SIZE` under rz_types.h

librz/bin/dwarf.c
```diff
@@ -340,6 +333,44 @@ static const char *dwarf_unit_types[] = {
...
 
+RZ_API const char *rz_bin_dwarf_get_tag_name(ut64 tag) {
+	if (tag >= DW_TAG_LAST) {
+		return NULL;
+	}
+	return dwarf_tag_name_encodings[tag];
+}
+
+RZ_API const char *rz_bin_dwarf_get_attr_name(ut64 attr_code) {
+	if (!((attr_code >= DW_AT_sibling && attr_code <= DW_AT_loclists_base) ||
```
ret2libc: how does this check work exactly? Can we extract it in a separate function with meaningful name?

librz/bin/dwarf.c
```diff
@@ -449,133 +468,125 @@ static const ut8 *parse_line_header_source(RzBinFile *bf, const ut8 *buf, const
...
-			if (count > 0) {
-				hdr->file_names = calloc(sizeof(file_entry), count);
-			} else {
-				hdr->file_names = NULL;
+RZ_API char *rz_bin_dwarf_line_header_get_full_file_path(Sdb *sdb_addrinfo, const RzBinDwarfLineHeader *header, ut64 file_index) {
+	if (file_index >= header->file_names_count) {
+		return NULL;
+	}
+	RzBinDwarfLineFileEntry *file = &header->file_names[file_index];
+	rz_return_val_if_fail(header && file, NULL);
```
wargio: doesn't `header->file_names` crashes before even call `rz_return_val_if_fail(header && file, NULL);` if `header` is NULL ?
i would place `rz_return_val_if_fail(header, NULL);` on the top and then just check with an `if (!file) ....`

librz/bin/dwarf.c
wargio: Please add a comment here then.

librz/bin/dwarf.c
ret2libc:
```suggestion
	if (lang >= RZ_ARRAY_SIZE(dwarf_langs)) {
```

wargio (changes requested):
please add a `rz_return[_val]_if_fail` at the beginning of all `rz_bin_dwarf_*` methods that can have a null ptr.

librz/bin/dwarf.c
```diff
@@ -449,133 +468,125 @@ static const ut8 *parse_line_header_source(RzBinFile *bf, const ut8 *buf, const
...
-		}
-		add_sdb_include_dir(sdb, str, i);
-		free(str);
-		i++;
+		rz_pvector_push(&incdirs, str);
 		buf += len + 1;
 	}
+	hdr->include_dirs_count = rz_pvector_len(&incdirs);
+	hdr->include_dirs = (char **)rz_pvector_flush(&incdirs);
+	rz_pvector_fini(&incdirs);
```
ret2libc, line 496: doesn't `flush` already `fini` the vector?

thestr4ng3r: From the view of only the vector itself, `flush` does a `clear`, which does remove everything that needs to be fini-d from the struct so technically the `fini` here is indeed not necessary. But this is rather an implementation detail and not part of the API. In fact, `clear` and `fini` are even more different in the sense that `clear` must always ensure that what you get out is still a valid vector with no elements, while `fini` may do anything.

librz/bin/dwarf.c
thestr4ng3r: It just checks whether it falls on an element in `dwarf_attr_encodings`. But actually, other than an upper bound check, this is unnecessary since undefined elements will automatically be NULL anyway. And even more, the array is very sparse so I think I will split it up.

wargio (approved):

PR #804, opened 2021-03-09
Java rewrite with LGPL license

wargio:

**Detailed description**

So new java core, new java bin, new java asm and new java analysis.

- Support till Java SE 15.
- New java core plugin
```
[0x00000120]> java?
Usage: java[?]   # Core plugin to visualize java class information
| java             # This help page
| javac[j]         # prints the class structure
| javaf[j]         # prints the class fields
| javai[j]         # prints the class interfaces
| javam[j]         # prints the class methods
| javap[j]         # prints the class constant pool
| javas[?]         # prints the class like a java source code
| javar[?] <index> # resolves the class constant pool value at a given index
```
- New java asm/disassemble plugin
- New java analysis plugin
- **Removed assembler** (i'll make a new PR later with it).

**Test plan**

Added tests for java (see related PR on testbin)

**Closing issues**

None


librz/analysis/p/analysis_java.c
```diff
@@ -1,346 +1,164 @@
...
-	ut8 result = (x & RZ_ANALYSIS_JAVA_CODE_OP) ? 1 : 0;
-	return result &&
-		((x & RZ_ANALYSIS_JAVA_CODEOP_LEAVE) == RZ_ANALYSIS_JAVA_CODEOP_LEAVE ||
-			(x & RZ_ANALYSIS_JAVA_CODEOP_RET) == RZ_ANALYSIS_JAVA_CODEOP_RET ||
-			(x & RZ_ANALYSIS_JAVA_CODEOP_JMP) == RZ_ANALYSIS_JAVA_CODEOP_JMP ||
-			(x & RZ_ANALYSIS_JAVA_CODEOP_SWITCH) == RZ_ANALYSIS_JAVA_CODEOP_SWITCH);
+static bool set_reg_profile(RzAnalysis *analysis) {
+	const char *p =
+		"=PC	pc\n"
+		"=SP	garbage\n"
```
XVilka, line 117: 🥇

librz/core/p/core_java.c
```diff
@@ -10,1659 +9,278 @@
...
+	return true;
 }
 
-// PLUGIN Definition Info
 RzCorePlugin rz_core_plugin_java = {
 	.name = "java",
 	.desc = "Suite of java commands, java help for more info",
-	.author = "RizinOrg",
-	.license = "Apache",
+	.license = "LGPL",
```
XVilka: Please set it with a version.

librz/analysis/p/analysis_java.c
```diff
@@ -1,346 +1,164 @@
...
+static bool java_fini(void *user) {
+	if (!user) {
+		return false;
 	}
-	return 0;
+	free(user);
+	return true;
 }
 
 RzAnalysisPlugin rz_analysis_plugin_java = {
```
XVilka, line 157: Does it handle JVM of different versions? Maybe we need to provide the available versions in `asm.cpu` then too.

XVilka (changes requested):

librz/include/rz_asm.h
```diff
@@ -132,7 +132,7 @@ typedef struct rz_asm_plugin_t {
 	void *user; // user data pointer
 	int bits;
 	int endian;
-	bool (*init)(void *user);
+	bool (*init)(void **user);
```
ret2libc: Given the PR is huge... why this change?

librz/include/rz_analysis.h
```diff
@@ -617,6 +617,7 @@ typedef struct rz_analysis_t {
 	int sleep; // analysis.sleep, sleep some usecs before analyzing more (avoid 100% cpu usages)
 	RzAnalysisCPPABI cpp_abi; // analysis.cpp.abi
 	void *user;
+	void *core;
```
ret2libc: Why this new pointer? Could you *point* me to where this is used?

librz/analysis/p/analysis_java.c
wargio: yes, the old versions, uses the same bytecode.

librz/include/rz_asm.h
wargio: because i need to save a context while the disassembler (or the analysis) runs. that user pointer was RzCore and makes no sense to have it there.

librz/include/rz_analysis.h
wargio: `void *user;` was used as `RzCore` for some weird core calls in `/core/` itself. 
Search `->analysis->user` in your current code.

XVilka:
Please also check how it replicates/corresponds to these use cases:
- http://deeso.github.io/posts/2014/04/radare-java-intro/
- http://deeso.github.io/posts/2014/04/radare-java-artifact-enumeration/
- https://github.com/radareorg/radareorg/blob/master/source/_files/radare_java.pdf

I think this is the file that was used in the slides: https://www.virustotal.com/gui/file/20fd2eec34cd1856a0c9e735d1914bf97d13162a869649afbcf5df3450a23b4c/detection

XVilka:
Not sure what caused this failure though:
```

[XX] db/tools/rz_bin rz-bin -D swift
RZ_NOPLUGINS=1 rizin -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -Qc '!rz-bin -D swift  _TFSSCfT21_builtinStringLiteralBp8byteSizeBw7isASCIIBi1__SS' bins/elf/libc.so.0
-- stdout

-- stderr
```
https://github.com/rizinorg/rizin/pull/804/checks?check_run_id=2090361351#step:16:503

wargio:
i dunno why that. i can't reproduce

XVilka:
It appears quite consistenly, and related to the fact that Swift demangler behaves differently on the MacOS: 
- https://github.com/rizinorg/rizin/blob/dev/librz/bin/mangling/swift-sd.c#L119
- https://github.com/rizinorg/rizin/blob/dev/librz/bin/mangling/swift-sd.c#L158

wargio:
yes, i had the same thought yesterday.

wargio:
@XVilka can you restart all the failed jobs? i can't

librz/include/rz_endian.h
```diff
@@ -114,6 +114,62 @@ static inline void rz_write_at_be64(void *dest, ut64 val, size_t offset) {
 	rz_write_be64(d, val);
 }
 
+static inline float rz_read_be_float(const void *src) {
```
XVilka: Could you please add unit tests for these float-reading functions?

XVilka (changes requested):
It's quite hard to review such a huge PR, especially for Java. But overall it looks good 👍 
Incredible job!

test/db/formats/mdmp
```diff
@@ -73,6 +73,7 @@ EOF
 RUN
 
 NAME=: test format definitions
+BROKEN=1
```
XVilka: Any particular reason why you marked it as broken? It is completely unrelated to the Java bytecode.

XVilka:
@wargio regarding the 3 first commits, maybe it's better to send them as a separate PR too?

XVilka:
Also I think it's good to go into 0.2.0.

test/db/formats/mdmp
wargio: on my system it was crashing randomly no reason

wargio:
> regarding the 3 first commits, maybe it's better to send them as a separate PR too?

Yes, it can be done

XVilka:
Also, please rebase due to the python plugin changes in the dev branch.

wargio:
Ready for review again 😄

XVilka (approved):
Looks good! 👍 💯

XVilka:
Should it be squashed or not?

wargio:
i think we can squash it

librz/analysis/p/analysis_java.c
```diff
@@ -1,346 +1,170 @@
...
+		"gpr	pc	    .32 0  0\n"
+		"gpr	garbage	.32 32 0\n";
+	return rz_reg_set_profile_string(analysis->reg, p);
 }
 
-static int java_op(RzAnalysis *analysis, RzAnalysisOp *op, ut64 addr, const ut8 *data, int len, RzAnalysisOpMask mask) {
-	/* get opcode size */
-	if (len < 1) {
-		op->type = RZ_ANALYSIS_OP_TYPE_ILL;
+static int archinfo(RzAnalysis *analysis, int query) {
```
ret2libc, line 131: this query should really be an enum value... Just thinking out loud.

librz/analysis/p/analysis_java.c
```diff
@@ -1,346 +1,170 @@
...
-	case RZ_ANALYSIS_JAVA_CODEOP_CALL: return conditional | RZ_ANALYSIS_OP_TYPE_CALL;
-	case RZ_ANALYSIS_JAVA_CODEOP_JMP: return conditional | RZ_ANALYSIS_OP_TYPE_JMP;
-	case RZ_ANALYSIS_JAVA_CODEOP_RET: return conditional | RZ_ANALYSIS_OP_TYPE_RET;
-	case RZ_ANALYSIS_JAVA_CODEOP_LEAVE: return RZ_ANALYSIS_OP_TYPE_LEAVE;
-	case RZ_ANALYSIS_JAVA_CODEOP_SWI: return RZ_ANALYSIS_OP_TYPE_SWI;
-	case RZ_ANALYSIS_JAVA_CODEOP_TRAP: return RZ_ANALYSIS_OP_TYPE_TRAP;
-	case RZ_ANALYSIS_JAVA_CODEOP_SWITCH: return RZ_ANALYSIS_OP_TYPE_SWITCH;
+	ut64 section = addr;
+	if (analysis->binb.bin) {
+		const RzBinSection *sec = analysis->binb.get_vsect_at(analysis->binb.bin, addr);
```
ret2libc, line 62: I think we could directly call `rz_bin_get_section_at`, because RzAnalysis already depends on RzBin.

librz/asm/p/asm_java.c
```diff
@@ -8,40 +7,131 @@
...
+	}
+
+	JavaVM vm = { 0 };
+	Bytecode bc = { 0 };
+
+	rz_strbuf_set(&op->buf_asm, "invalid");
+
+	ut64 section = a->pc;
+	if (a->binb.bin) {
+		const RzBinSection *sec = a->binb.get_vsect_at(a->binb.bin, a->pc);
```
ret2libc, line 73: just directly call the bin function.

librz/asm/p/asm_java.c
```diff
@@ -8,40 +7,131 @@
...
+			ctx->pc = a->pc;
+		} else if (bc.opcode == BYTECODE_AB_LOOKUPSWITCH) {
+			ctx->count = 0;
+			ctx->switchop = BYTECODE_AB_LOOKUPSWITCH;
+			ctx->ls = *((LookupSwitch *)bc.extra);
+			ctx->pc = a->pc;
+		}
+		bytecode_clean(&bc);
+	} else {
+		eprintf("[!] java_disassemble: jvm fetch failed.\n");
```
ret2libc: I would use RZ_LOG_ERROR instead, so you can filter the output or redirect to file, etc.

ret2libc: In general in the plugins, I mean.

librz/bin/p/bin_java.c
```diff
@@ -1,227 +1,259 @@
...
+static RzBinInfo *info(RzBinFile *bf) {
+	RzBinJavaClass *jclass = rz_bin_file_get_java_class(bf);
+	if (!jclass) {
+		return NULL;
+	}
+	RzBinInfo *binfo = RZ_NEW0(RzBinInfo);
+	if (!binfo) {
+		return NULL;
+	}
+	binfo->lang /*      */ = rz_bin_java_class_language(jclass);
```
ret2libc: I would avoid these comments... we don't use them anywhere.

librz/bin/p/bin_java.c
```diff
@@ -1,227 +1,259 @@
...
-	add_sdb_bin_obj(jvcname, bin);
-	bin->AllJavaBinObjs = DB;
-	free(jvcname);
 }
 
-static Sdb *get_sdb(RzBinFile *bf) {
-	RzBinObject *o = bf->o;
-	RzBinJavaObj *bin;
-	if (!o) {
+static RzList *classes(RzBinFile *bf) {
```
ret2libc, line 81: there is no way to get two classes in a single file, is there?

librz/bin/p/bin_java.c
```diff
@@ -1,227 +1,259 @@
...
 
-static RzList *fields(RzBinFile *bf) {
-	return NULL; // rz_bin_java_get_fields (bf->o->bin_obj);
+static int demangle_type(const char *str) {
+	return RZ_BIN_NM_JAVA;
 }
 
-static RzList *libs(RzBinFile *bf) {
-	return rz_bin_java_get_lib_names(bf->o->bin_obj);
+static char *enrich_asm(RzBinFile *bf, const char *asm_str, int asm_len) {
```
ret2libc, line 205: what is exactly the purpose of this new plugin method? Do we really need it? If it's going to modify the asm, isn't a RzParsePlugin better?

librz/core/p/core_java.c
```diff
@@ -10,1659 +9,321 @@
...
+	RzCmd *rcmd = core->rcmd;
+	RzCmdDesc *root_cd = rz_cmd_get_root(rcmd);
 	if (!root_cd) {
 		return false;
 	}
 
-	RzCmdDesc *cmd_java_cd = rz_cmd_desc_argv_new(cmd, root_cd, "java", rz_cmd_java_handler, &cmd_java_help);
-	rz_warn_if_fail(cmd_java_cd);
-	return cmd_java_cd != NULL;
+	RzCmdDesc *java = rz_cmd_desc_group_new(rcmd, root_cd, "java", name_handler(java), &name_help(java), &java_usage);
```
ret2libc: we don't usually have a command that shows nothing but its help, so I'd remove the handler of the `java` command. One has to do `java?`, for consistency, to get an help.

librz/core/p/core_java.c
```diff
@@ -10,1659 +9,321 @@
...
+		rz_warn_if_reached();
+		return false;
+	}
+
+	rz_cmd_desc_argv_modes_new_warn(rcmd, java, javac, RZ_OUTPUT_MODE_STANDARD | RZ_OUTPUT_MODE_JSON);
+	rz_cmd_desc_argv_modes_new_warn(rcmd, java, javaf, RZ_OUTPUT_MODE_STANDARD | RZ_OUTPUT_MODE_JSON);
+	rz_cmd_desc_argv_modes_new_warn(rcmd, java, javai, RZ_OUTPUT_MODE_STANDARD | RZ_OUTPUT_MODE_JSON);
+	rz_cmd_desc_argv_modes_new_warn(rcmd, java, javam, RZ_OUTPUT_MODE_STANDARD | RZ_OUTPUT_MODE_JSON);
+	rz_cmd_desc_argv_modes_new_warn(rcmd, java, javap, RZ_OUTPUT_MODE_STANDARD | RZ_OUTPUT_MODE_JSON);
+	rz_cmd_desc_argv_modes_new_warn(rcmd, java, javas, RZ_OUTPUT_MODE_STANDARD);
```
ret2libc: either you fix cmd_api so that STANDANDARD alone == regular command, or you use rz_cmd_desc_argv_new instead, because otherwise:
```
[0x00000339]> javas?
Usage: javas[?]   # prints the class like a java source code
| javas     # prints the class like a java source code
```

librz/core/p/core_java.c
```diff
@@ -10,1659 +9,321 @@
...
+	rz_cmd_desc_argv_modes_new_warn(rcmd, java, javar, RZ_OUTPUT_MODE_STANDARD);
```
ret2libc: same for this

test/db/analysis/java
```diff
@@ -1,189 +1,164 @@
...
-import java.lang.Object;
-import java.lang.String;
-import java.lang.Process;
-import java.io.BufferedReader;
-import java.lang.Exception;
-import java.lang.Runtime;
-import java.lang.Process;
-import java.lang.String;
-
-class JDWPExecCmd { // @0x0000
```
ret2libc, line 27: do these addresses still make sense? It seems you removed them.

ret2libc (commented):
```
$ ./build-java/binrz/rizin/rizin ./test/bins/java/crash2.class 
invalid constant pool tag: 24 at 0x14
WARNING: (../librz/bin/format/java/class_bin.c:89):java_class_parse: code should not be reached
WARNING: (../librz/bin/format/java/class_bin.c:1351):rz_bin_java_class_const_pool_as_imports: code should not be reached
WARNING: (../librz/bin/format/java/class_bin.c:1351):rz_bin_java_class_const_pool_as_imports: code should not be reached
WARNING: (../librz/bin/format/java/class_bin.c:1299):rz_bin_java_class_const_pool_as_symbols: code should not be reached
WARNING: (../librz/bin/format/java/class_bin.c:1299):rz_bin_java_class_const_pool_as_symbols: code should not be reached
WARNING: (../librz/bin/format/java/class_bin.c:1667):rz_bin_java_class_as_libraries: code should not be reached
WARNING: (../librz/bin/format/java/class_bin.c:1667):rz_bin_java_class_as_libraries: code should not be reached
WARNING: (../librz/bin/format/java/class_bin.c:386):rz_bin_java_class_name: code should not be reached
WARNING: (../librz/bin/format/java/class_bin.c:398):rz_bin_java_class_super: code should not be reached
WARNING: No calling convention defined for this file, analysis may be inaccurate.
 -- Use the '[' and ']' keys in visual mode to adjust the screen width

```

I think you are using rz_warn_if_reached in places where you should instead handle the problem, because the input comes from the user-provided file.

```
$ ./build-java/binrz/rizin/rizin ./test/bins/java/DemoTest.class 
WARNING: No calling convention defined for this file, analysis may be inaccurate.
 -- Select your architecture with: 'e asm.arch=<arch>' or rizin -a from the shell
[0x000001b9]> javar 0
WARNING: rz_bin_java_class_const_pool_resolve_index: assertion 'bin && index > 0' failed (line 560)
```

I haven't checked(and won't check) everything, but it seems well written overall :)

librz/bin/p/bin_java.c
wargio: no

librz/bin/p/bin_java.c
wargio: essentially, as explained on MM, on java you have a table with indices (from 0 to N) with all the references (like strings, definitions, const values, etc..) and when you load one of these, you have just the number of the index, not a file offset/address. so in order to enrich the asm to have `ldc constant_pool.11` you will have `ldc 1.5875f` for example.
This is just the cleanest way to fix this issue on asm and to avoid to have rzasm depending on rzbin

librz/core/p/core_java.c
wargio: actually i feel like is better to do this way, so when you type java it will prompt you the help message.

librz/core/p/core_java.c
wargio: probably i should.

test/db/analysis/java
wargio: no. if you want to know the addresses, please use the ic command not that.

librz/analysis/p/analysis_java.c
wargio: that api is meant to avoid analysis to depend on rzbin i think

librz/asm/p/asm_java.c
wargio: same here

wargio:
@ret2libc regarding the messed up java class binaries like (bins/java/crash2.class).
those binaries are not even vaild. i think is appropriate to show those errors, because i do sanitize all the risky inputs and when those checks fails, is correct to have those errors.
Essentially is like asking rizin to allocate 0xe0e0e0e0 sections with virtual size of 0x12345678 and physical of 0; ofc this has to fail, because it makes no sense at all.

wargio:
And yes, it is handled correctly... it just allows you to keep parsing the binary till it fails horribly hard (you know, just in case of some malware stuff happening, to prevent analysis of the binary itself)

librz/asm/p/asm_java.c
wargio: We decided to ignore these till we resolve the dependency nightmare

librz/analysis/p/analysis_java.c
wargio: We decided to ignore these till we resolve the dependency nightmare

librz/core/p/core_java.c
ret2libc: If we want this behaviour, let's do it for *all* commands in a separate PR. For example, see `zf`. `zf` does not exist, but it is the group name for `zfd` and `zfs`. If we want `zf` to get the help of the `zf` group, then we should have this in RzCmd to ensure consistency, not ad-hoc IMO. Otherwise we won't be consistent.

librz/bin/p/bin_java.c
```diff
@@ -1,227 +1,259 @@
...
+	RzBinJavaClass *jclass = rz_bin_file_get_java_class(bf);
+	if (!jclass) {
+		return NULL;
+	}
+	for (int i = 0; i < asm_len; ++i) {
+		if (asm_str[i] == 'c' && !strncmp(asm_str + i, "constant_pool.", 14)) {
```
ret2libc: let the compiler be smart for you
```suggestion
		if (!strncmp(asm_str + i, "constant_pool.", 14)) {
```

ret2libc: Also, put `"constant_pool."` in a var/define and use `strlen(var)` instead of `14`. The compiler will optimize away  the strlen anyway and I don't have to check that `constant_pool.` is really of 14 chars :D

librz/bin/p/bin_java.c
ret2libc: :( :( I hate these things... discussing on Mattermost right now, but I guess for now it's fine :( :(

librz/core/p/core_java.c
wargio: resolved like for zf. i wasn't aware of that way of having a group.

ret2libc (approved):
As I'm no expert in Java, I've mainly looked the changes to things outside java-format/asm/analysis you have done and everything looks ok. So, worst case, java is the only broken thing :D :D Good work!

PR #808, opened 2021-03-10
Refactoring RzAsm and RzAnalysis to allow plugin execution context.

wargio:
# DO NOT SQUASH ME

**Detailed description**

These 2 commits are needed by the #804

This PR aims to remove static structures from RzAsm and RzAnalysis

### DIFF
--- a/librz/analysis/analysis.c
+++ b/librz/analysis/analysis.c
@@ -133,7 +133,7 @@ RZ_API RzAnalysis *rz_analysis_new(void) {
 	analysis->leaddrs = NULL;
 	analysis->imports = rz_list_newf(free);
 	rz_analysis_set_bits(analysis, 32);
-	analysis->plugins = rz_list_newf((RzListFree)rz_analysis_plugin_free);
+	analysis->plugins = rz_list_newf(NULL);
 	if (analysis->plugins) {
 		for (i = 0; analysis_static_plugins[i]; i++) {
 			rz_analysis_add(analysis, analysis_static_plugins[i]);
@@ -142,9 +142,10 @@ RZ_API RzAnalysis *rz_analysis_new(void) {
 	return analysis;
 }
 
-RZ_API void rz_analysis_plugin_free(RzAnalysisPlugin *p) {
-	if (p && p->fini) {
-		p->fini(NULL);
+RZ_API void plugin_fini(RzAnalysis *analysis) {
+	RzAnalysisPlugin *p = analysis->cur;
+	if (p && p->fini && !p->fini(analysis->plugin_data)) {
+		RZ_LOG_ERROR("analysis plugin '%s' failed to terminate.\n", p->name);
 	}
 }
 
@@ -154,6 +155,9 @@ RZ_API RzAnalysis *rz_analysis_free(RzAnalysis *a) {
 	if (!a) {
 		return NULL;
 	}
+
+	plugin_fini(a);
+
 	rz_list_free(a->fcns);
 	ht_up_free(a->ht_addr_fun);
 	ht_pp_free(a->ht_name_fun);
@@ -185,15 +189,8 @@ RZ_API RzAnalysis *rz_analysis_free(RzAnalysis *a) {
 	return NULL;
 }
 
-RZ_API void rz_analysis_set_user_ptr(RzAnalysis *analysis, void *user) {
-	analysis->user = user;
-}
-
-RZ_API int rz_analysis_add(RzAnalysis *analysis, RzAnalysisPlugin *foo) {
-	if (foo->init) {
-		foo->init(analysis->user);
-	}
-	rz_list_append(analysis->plugins, foo);
+RZ_API int rz_analysis_add(RzAnalysis *analysis, RzAnalysisPlugin *p) {
+	rz_list_append(analysis->plugins, p);
 	return true;
 }
 
@@ -206,13 +203,12 @@ RZ_API bool rz_analysis_use(RzAnalysis *analysis, const char *name) {
 			if (!h->name || strcmp(h->name, name)) {
 				continue;
 			}
-#if 0
-			// regression happening here for asm.emu
-			if (analysis->cur && analysis->cur == h) {
-				return true;
-			}
-#endif
+			plugin_fini(analysis);
 			analysis->cur = h;
+			if (h && h->init && !h->init(&analysis->plugin_data)) {
+				RZ_LOG_ERROR("analysis plugin '%s' failed to initialize.\n", h->name);
+				return false;
+			}
 			rz_analysis_set_reg_profile(analysis);
 			return true;
 		}
--- a/librz/analysis/p/analysis_tms320.c
+++ b/librz/analysis/p/analysis_tms320.c
@@ -5,8 +5,6 @@
 #include "analysis_tms320c64x.c"
 #include "../../asm/arch/tms320/tms320_dasm.h"
 
-static tms320_dasm_t engine = { 0 };
-
 typedef int (*TMS_ANALYSIS_OP_FN)(RzAnalysis *analysis, RzAnalysisOp *op, ut64 addr, const ut8 *buf, int len);
 
 int tms320_c54x_op(RzAnalysis *analysis, RzAnalysisOp *op, ut64 addr, const ut8 *buf, int len);
@@ -23,10 +21,11 @@ int tms320_c54x_op(RzAnalysis *analysis, RzAnalysisOp *op, ut64 addr, const ut8 
 }
 
 int tms320_c55x_op(RzAnalysis *analysis, RzAnalysisOp *op, ut64 addr, const ut8 *buf, int len) {
-	const char *str = engine.syntax;
+	tms320_dasm_t *engine = (tms320_dasm_t *)analysis->plugin_data;
+	const char *str = engine->syntax;
 
 	op->delay = 0;
-	op->size = tms320_dasm(&engine, buf, len);
+	op->size = tms320_dasm(engine, buf, len);
 	op->type = RZ_ANALYSIS_OP_TYPE_NULL;
 
 	str = strstr(str, "||") ? str + 3 : str;
@@ -95,12 +94,22 @@ int tms320_op(RzAnalysis *analysis, RzAnalysisOp *op, ut64 addr, const ut8 *buf,
 	return aop(analysis, op, addr, buf, len);
 }
 
-static int tms320_init(void *unused) {
-	return tms320_dasm_init(&engine);
+static bool tms320_init(void **user) {
+	tms320_dasm_t *engine = RZ_NEW0(tms320_dasm_t);
+	if (!engine) {
+		return false;
+	}
+	tms320_dasm_init(engine);
+	*user = engine;
+	return true;
 }
 
-static int tms320_fini(void *unused) {
-	return tms320_dasm_fini(&engine);
+static bool tms320_fini(void *user) {
+	rz_return_val_if_fail(user, false);
+	tms320_dasm_t *engine = (tms320_dasm_t *)user;
+	tms320_dasm_fini(engine);
+	free(engine);
+	return true;
 }
 
 RzAnalysisPlugin rz_analysis_plugin_tms320 = {
--- a/librz/analysis/p/analysis_x86_cs.c
+++ b/librz/analysis/p/analysis_x86_cs.c
@@ -27,8 +27,6 @@ call = 4
 #define HAVE_CSGRP_PRIVILEGE 0
 #endif
 
-#define USE_ITERZ_API 0
-
 #if CS_API_MAJOR < 2
 #error Old Capstone not supported
 #endif
@@ -55,13 +53,17 @@ call = 4
 #define ARG1_AR   1
 #define ARG2_AR   2
 
+typedef struct x86_cs_context_t {
+	csh handle;
+	int omode;
+	cs_insn *insn;
+} X86CSContext;
+
 struct Getarg {
 	csh handle;
 	cs_insn *insn;
 	int bits;
 };
-
-static csh handle = 0;
 
 static void hidden_op(cs_insn *insn, cs_x86 *x, int mode) {
 	unsigned int id = insn->id;
@@ -107,7 +109,8 @@ static void hidden_op(cs_insn *insn, cs_x86 *x, int mode) {
 	}
 }
 
-static void opex(RzStrBuf *buf, cs_insn *insn, int mode) {
+static void opex(RzStrBuf *buf, X86CSContext *ctx, int mode) {
+	cs_insn *insn = ctx->insn;
 	int i;
 	PJ *pj = pj_new();
 	if (!pj) {
@@ -129,7 +132,7 @@ static void opex(RzStrBuf *buf, cs_insn *insn, int mode) {
 		switch (op->type) {
 		case X86_OP_REG:
 			pj_ks(pj, "type", "reg");
-			pj_ks(pj, "value", cs_reg_name(handle, op->reg));
+			pj_ks(pj, "value", cs_reg_name(ctx->handle, op->reg));
 			break;
 		case X86_OP_IMM:
 			pj_ks(pj, "type", "imm");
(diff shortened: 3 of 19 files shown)


librz/analysis/analysis.c
```diff
@@ -142,9 +142,10 @@ RZ_API RzAnalysis *rz_analysis_new(void) {
 	return analysis;
 }
 
-RZ_API void rz_analysis_plugin_free(RzAnalysisPlugin *p) {
-	if (p && p->fini) {
-		p->fini(NULL);
+RZ_API void plugin_fini(RzAnalysis *analysis) {
+	RzAnalysisPlugin *p = analysis->cur;
+	if (p && p->fini && !p->fini(analysis->user)) {
+		eprintf("analysis plugin '%s' failed to terminate.\n", p->name);
```
ret2libc: `RZ_LOG_ERROR`?

librz/analysis/analysis.c
```diff
@@ -154,6 +155,9 @@ RZ_API RzAnalysis *rz_analysis_free(RzAnalysis *a) {
 	if (!a) {
 		return NULL;
 	}
+
+	plugin_fini(a);
```
ret2libc, line 159: what about all other non-current analysis plugins? Shouldn't we `fini` them as well?

librz/include/rz_asm.h
```diff
@@ -132,7 +132,7 @@ typedef struct rz_asm_plugin_t {
 	void *user; // user data pointer
 	int bits;
 	int endian;
-	bool (*init)(void *user);
+	bool (*init)(void **user);
```
ret2libc, line 135: why `**user`?
If you need to return a pointer can't we just do:
```suggestion
	void *(*init)(void *user);
```
?

ret2libc (commented):
I haven't look at this in detail yet, but I think we should pay particular attention to the fact that this PR calls init/fini of each plugin every time you change the currently selected plugin. I wonder whether it is the right thing to do from a concept point of view and from the implementation point of view.

librz/analysis/analysis.c
wargio: only if they have static context, yes definitely

librz/include/rz_asm.h
wargio: because we want to handle errors.

ret2libc: Does this mean that you could have `*user` set but with an error? Or `*user` not set but without errors?

wargio: the arch might not need to return a data structure.

wargio:
> I haven't look at this in detail yet, but I think we should pay particular attention to the fact that this PR calls init/fini of each plugin every time you change the currently selected plugin. I wonder whether it is the right thing to do from a concept point of view and from the implementation point of view.

This is ok. there are only a bunch of archs that does this and it is the correct way of initializing the stuff, so people can actually use the library multiple times with different contexts

librz/asm/p/asm_arm_cs.c
```diff
@@ -294,10 +310,10 @@ RzAsmPlugin rz_asm_plugin_arm_cs = {
 	.bits = 16 | 32 | 64,
 	.endian = RZ_SYS_ENDIAN_LITTLE | RZ_SYS_ENDIAN_BIG,
 	.disassemble = &disassemble,
-	.mnemonics = mnemonics,
+	//.mnemonics = mnemonics,
```
XVilka: Just remove this?

librz/asm/p/asm_snes.c
```diff
@@ -49,6 +34,19 @@ static int dis(RzAsm *a, RzAsmOp *op, const ut8 *buf, int len) {
 	return dlen;
 }
 
+static bool snes_asm_init(void **user) {
+	*user = RZ_NEW0(struct snes_asm_flags);
+	return user != NULL;
+}
+
+static bool snes_asm_fini(void *user) {
+	if (!user) {
```
XVilka: I wonder if it makes sense to use `rz_return_val_if_fail(user, false)` to catch the programmer errors, since this shouldn't happen, I suppose?

librz/analysis/p/analysis_x86_cs.c
```diff
@@ -3118,45 +3123,53 @@ static int cs_len_prefix_opcode(uint8_t *item) {
...
-			handle = 0;
+			ctx->handle = 0;
 			return 0;
 		}
 	}
 	op->cycles = 1; // aprox
-	cs_option(handle, CS_OPT_DETAIL, CS_OPT_ON);
+	cs_option(ctx->handle, CS_OPT_DETAIL, CS_OPT_ON);
 	// capstone-next
 #if USE_ITERZ_API
```
XVilka: Is this even used `USE_ITERZ_API`? It looks like zero everywhere, just remove this code.

librz/analysis/p/analysis_x86_cs.c
```diff
@@ -3166,21 +3179,21 @@ static int analop(RzAnalysis *a, RzAnalysisOp *op, ut64 addr, const ut8 *buf, in
...
 			op->mnemonic = rz_str_newf("%s%s%s",
-				insn->mnemonic,
-				insn->op_str[0] ? " " : "",
-				insn->op_str);
+				ctx->insn->mnemonic,
+				ctx->insn->op_str[0] ? " " : "",
+				ctx->insn->op_str);
 		}
 		// int rs = a->bits / 8;
 		//const char *pc = (a->bits==16)?"ip": (a->bits==32)?"eip":"rip";
```
XVilka: Should be handled by the reg profile, IMHO, remove this too.

librz/analysis/p/analysis_x86_cs.c
```diff
@@ -3118,45 +3123,53 @@ static int cs_len_prefix_opcode(uint8_t *item) {
...
-#if USE_ITERZ_API
-	static
-#endif
-		cs_insn *insn = NULL;
-	int mode = (a->bits == 64) ? CS_MODE_64 : (a->bits == 32) ? CS_MODE_32
-		: (a->bits == 16)                                 ? CS_MODE_16
-								  : 0;
+	X86CSContext *ctx = (X86CSContext *)a->plugin_data;
+
+	int mode = 0;
```
XVilka: Extract this in a small inline function, I think.

XVilka (changes requested):

ret2libc:
> > I haven't look at this in detail yet, but I think we should pay particular attention to the fact that this PR calls init/fini of each plugin every time you change the currently selected plugin. I wonder whether it is the right thing to do from a concept point of view and from the implementation point of view.
> 
> This is ok. there are only a bunch of archs that does this and it is the correct way of initializing the stuff, so people can actually use the library multiple times with different contexts

The reason I said that is because of this scenario:
```
[0x00000000]> wx 00b5884206bf01300130042000bd
[0x00000000]> e asm.arch=arm asm.bits=16
[0x00000000]> pd 7
            0x00000000      push  {lr}
            0x00000002      cmp   r0, r1
            0x00000004      itte  eq
            0x00000006      addeq r0, 1
            0x00000008      addeq r0, 1
            0x0000000a      movne r0, 4
            0x0000000c      pop   {pc}
[0x00000000]> pd 2 @ 6
            0x00000006      addeq r0, 1
            0x00000008      addeq r0, 1
[0x00000000]> pd 2 @ 6 @a:x86
            0000:0006      add   word [bx + si], si
            0000:0008      add   word [bx + si], si
[0x00000000]> pd 2 @ 6
            0x00000006      adds  r0, 1
            0x00000008      adds  r0, 1
[0x00000000]> 
```
Some instructions in ARM depends on previous ones (in particular, the `it` instruction has some suffixes that changes how to interpret the following X instructions) and the arm plugin has a state to keep track of how to interpret them. As you can see, changing temporarily the plugin to x86 makes the arm plugin lose the state, so it doesn't know anymore how to interpret those instructions.

That said, it should not be a big problem and we can probably ignore it because anyway even right now, if you just go directly to address `6`, the asm plugin won't know how to interpret the adds instruction (because he never saw the `itte` instruction).

Just wanted to raise this.

XVilka:
Good idea, btw it might be a good idea to add your example as a test, even if broken.

librz/include/rz_asm.h
```diff
@@ -129,10 +130,9 @@ typedef struct rz_asm_plugin_t {
 	const char *cpus;
 	const char *desc;
 	const char *license;
-	void *user; // user data pointer
 	int bits;
 	int endian;
-	bool (*init)(void *user);
+	bool (*init)(void **user);
```
ret2libc, line 135: I think init should also allow to pass some arguments (e.g. some plugins may require some information from the outside to configure special stuff). @thestr4ng3r cc

librz/asm/p/asm_arm_cs.c
wargio: i thought i have reintroduced it. i must have forgotten to cherrypick that change.

librz/analysis/p/analysis_x86_cs.c
wargio: You mean to extract the context ?

librz/asm/p/asm_snes.c
wargio: > > > I haven't look at this in detail yet, but I think we should pay particular attention to the fact that this PR calls init/fini of each plugin every time you change the currently selected plugin. I wonder whether it is the right thing to do from a concept point of view and from the implementation point of view.
> > 
> > 
> > This is ok. there are only a bunch of archs that does this and it is the correct way of initializing the stuff, so people can actually use the library multiple times with different contexts
> 
> The reason I said that is because of this scenario:
> 
> ```
> [0x00000000]> wx 00b5884206bf01300130042000bd
> [0x00000000]> e asm.arch=arm asm.bits=16
> [0x00000000]> pd 7
>             0x00000000      push  {lr}
>             0x00000002      cmp   r0, r1
>             0x00000004      itte  eq
>             0x00000006      addeq r0, 1
>             0x00000008      addeq r0, 1
>             0x0000000a      movne r0, 4
>             0x0000000c      pop   {pc}
> [0x00000000]> pd 2 @ 6
>             0x00000006      addeq r0, 1
>             0x00000008      addeq r0, 1
> [0x00000000]> pd 2 @ 6 @a:x86
>             0000:0006      add   word [bx + si], si
>             0000:0008      add   word [bx + si], si
> [0x00000000]> pd 2 @ 6
>             0x00000006      adds  r0, 1
>             0x00000008      adds  r0, 1
> [0x00000000]> 
> ```
> 
> Some instructions in ARM depends on previous ones (in particular, the `it` instruction has some suffixes that changes how to interpret the following X instructions) and the arm plugin has a state to keep track of how to interpret them. As you can see, changing temporarily the plugin to x86 makes the arm plugin lose the state, so it doesn't know anymore how to interpret those instructions.
> 
> That said, it should not be a big problem and we can probably ignore it because anyway even right now, if you just go directly to address `6`, the asm plugin won't know how to interpret the adds instruction (because he never saw the `itte` instruction).
> 
> Just wanted to raise this.

I think this is correct behaviour: if you change ASM you need to change also the context. makes 0 sense to me to keep that context.

librz/analysis/p/analysis_x86_cs.c
XVilka: No, mode handling.

wargio:
@ret2libc I think this is correct behaviour: if you change arch you need to change also the context and you have to lose the previous one. makes 0 sense to me to keep that context.

librz/include/rz_asm.h
wargio: There are NO plugins that requires special stuff.
if they needed something i would have seen since rzcore was passed as "user".
The only plugin that actually do need something is java, which requires to have access to RzBin to resolve class names, constant variables, etc..

wargio:
@ret2libc if you tell me where i can find the temporary modifier for `@a:x86` i can prevent this.

librz/include/rz_asm.h
ret2libc: Right now there are not, but I remember discussing this with @thestr4ng3r not long ago. It was actually about IO and Bin, but the same logic applies i think.

ret2libc:
> @ret2libc I think this is correct behaviour: if you change arch you need to change also the context and you have to lose the previous one. makes 0 sense to me to keep that context.

> @ret2libc if you tell me where i can find the temporary modifier for `@a:x86` i can prevent this.

I'd prefer to avoid having special handling in `@a` (you would then have different behaviour if doing e.g. `@e:asm.arch` etc.). If we are ok with the behaviour, let just it be.

wargio:
> I'd prefer to avoid having special handling in `@a` (you would then have different behaviour if doing e.g. `@e:asm.arch` etc.). If we are ok with the behaviour, let just it be.

yes, but when you change the arch, you need to change also the analysis and etc..
therefore there is no reason at all the keep it in the asm/analysis plugin. so if you temporarily change the architecture, then you need a new context.

wargio:
@ret2libc and @XVilka i added the test that you wrote, so we can fix this properly later

XVilka (approved):

librz/analysis/p/analysis_tms320.c
```diff
@@ -95,12 +94,20 @@ int tms320_op(RzAnalysis *analysis, RzAnalysisOp *op, ut64 addr, const ut8 *buf,
 	return aop(analysis, op, addr, buf, len);
 }
 
-static int tms320_init(void *unused) {
-	return tms320_dasm_init(&engine);
+static bool tms320_init(void **user) {
+	tms320_dasm_t *engine = RZ_NEW0(tms320_dasm_t);
+	rz_return_val_if_fail(engine, false);
```
ret2libc: allocations should be checked with `if (!engine)`, not with rz_return_val.

librz/analysis/p/analysis_x86_cs.c
```diff
@@ -3272,16 +3224,18 @@ static int esil_x86_cs_init(RzAnalysisEsil *esil) {
 	return true;
 }
 
-static int init(void *p) {
-	handle = 0;
+static bool x86_init(void **user) {
+	X86CSContext *ctx = RZ_NEW0(X86CSContext);
+	rz_return_val_if_fail(ctx, false);
```
ret2libc: also here

librz/asm/p/asm_arm_cs.c
```diff
@@ -266,22 +272,63 @@ static int assemble(RzAsm *a, RzAsmOp *op, const char *buf) {
...
-	if (!ht_it) {
-		ht_it = ht_uu_new0();
-	}
-	if (!ht_itblock) {
-		ht_itblock = ht_uu_new0();
-	}
-	return 0;
+static bool arm_init(void **user) {
+	ArmCSContext *ctx = RZ_NEW0(ArmCSContext);
+	rz_return_val_if_fail(ctx, false);
```
ret2libc: also here

librz/asm/p/asm_tms320.c
```diff
@@ -83,17 +89,28 @@ static int tms320_disassemble(RzAsm *a, RzAsmOp *op, const ut8 *buf, int len) {
...
+	op->size = tms320_dasm(&ctx->engine, buf, len);
+	rz_asm_op_set_asm(op, ctx->engine.syntax);
 	return op->size;
 }
 
-static bool tms320_init(void *user) {
-	return tms320_dasm_init(&engine);
+static bool tms320_init(void **user) {
+	TmsContext *ctx = RZ_NEW0(TmsContext);
+	rz_return_val_if_fail(ctx, false);
```
ret2libc: again

wargio:
@ret2libc fixed all your points :D

Maijin (approved):

XVilka:
Ok, lets wait for the final look from @ret2libc and good to merge 👍

librz/asm/p/asm_snes.c
ret2libc: Why was this marked as resolved?

ret2libc (approved):
Please resolve properly the comment below, but overall LGTM!

wargio:
@ret2libc good catch. i messed up with the commits

PR #829, opened 2021-03-15
core/disasm: prevent double length when meta is string

### LINKED ISSUE #827: asm.size leads to wrong string length being displayed
**Environment information**
* Operating System: linuxmint 20.1
* Cutter version: 1.12.0
* File format: any
 * Arch: N/A
 * Type: N/A

**Describe the bug**
String length is rendered wrongly (some digits appended) depending on independent setting.
This is probably not the only attribute affected by this bug.

**To Reproduce**
1. Load shellcode `536f6d6520427574746572666c7900` with analysis
2. Disassembly > Line 0x0000 > Set as... > String... > Auto-detect
3. Menu Bar > Edit > Preferences > Disassembly > Style > Check "Show size of opcodes in disassembly (asm.size)"
4. See Error

**Expected behavior**
Length should not change depending on changed setting.

**Screenshots**
asm.size is off
image
asm.size is on
image

ret2libc:

**Detailed description**

Honestly... this fix is really bad, but I cannot think of anything else. It simply avoids re-printing the len with `ds_print_op_size` when a meta of type string was found.

**Test plan**

...

**Closing issues**

Closes #827


librz/core/disasm.c
```diff
@@ -5397,7 +5397,9 @@ RZ_API int rz_core_print_disasm(RzPrint *p, RzCore *core, ut64 addr, ut8 *buf, i
 				ds_print_core_vmode(ds, ds->asm_hint_pos);
 			}
 		}
-		ds_print_op_size(ds);
+		if (!mi_found || mi_type != RZ_META_TYPE_STRING) {
```
ITAYC0HEN: does it only happen with strings? what about structs? types? and other things that potentially have a length displayed?

ret2libc: I see also `data` meta display the length, but in that case it is showed in a nicer way. I'm not aware of other cases, are you?

ITAYC0HEN: I did not check. but I won't be surprised if other places have this duplication as well.

kazarmy:
> Honestly... this fix is really bad, but I cannot think of anything else. It simply avoids re-printing the len with `ds_print_op_size` when a meta of type string was found.

Technically speaking, meta items are not opcodes/instructions so `ds_print_op_size` should stay quiet if it encounters them.

kazarmy:
Oh I just saw 1ee3b4619e6de8617c44941f189c655f97f979d5 but all meta items that say "this is not code" should be ignored by `ds_print_op_size` as well (which should include the `data` meta item).

ret2libc:
> Oh I just saw [1ee3b46](https://github.com/rizinorg/rizin/commit/1ee3b4619e6de8617c44941f189c655f97f979d5) but all meta items that say "this is not code" should be ignored by `ds_print_op_size` as well (which should include the `data` meta item).

Done

librz/core/disasm.c
```diff
@@ -2844,21 +2844,34 @@ static void ds_print_offset(RDisasmState *ds) {
...
+	rz_pvector_foreach (metas, it) {
+		RzIntervalNode *node = *it;
+		RzAnalysisMetaItem *mi = node->data;
+		switch (mi->type) {
+		case RZ_META_TYPE_DATA:
+		case RZ_META_TYPE_STRING:
+		case RZ_META_TYPE_FORMAT:
+		case RZ_META_TYPE_MAGIC:
+		case RZ_META_TYPE_HIDE:
+		case RZ_META_TYPE_COMMENT:
```
kazarmy: Comments can be attached to code, right?

kazarmy (changes requested):

librz/core/disasm.c
```diff
@@ -2827,8 +2827,39 @@ static void ds_print_offset(RDisasmState *ds) {
...
```
ret2libc:
```suggestion
```

kazarmy (dismissed):

librz/core/disasm.c
```diff
@@ -2827,8 +2827,38 @@ static void ds_print_offset(RDisasmState *ds) {
...
+	rz_pvector_foreach (metas, it) {
+		RzIntervalNode *node = *it;
+		RzAnalysisMetaItem *mi = node->data;
+		switch (mi->type) {
+		case RZ_META_TYPE_DATA:
+		case RZ_META_TYPE_STRING:
+		case RZ_META_TYPE_FORMAT:
+		case RZ_META_TYPE_MAGIC:
+		case RZ_META_TYPE_HIDE:
+		case RZ_META_TYPE_RUN:
```
kazarmy, line 2854: Btw, what is meta type `run`?

librz/core/disasm.c
```diff
@@ -2827,8 +2827,38 @@ static void ds_print_offset(RDisasmState *ds) {
...
+		case RZ_META_TYPE_VARTYPE:
```
kazarmy: This is also actually a comment meta if I'm not mistaken.

kazarmy (changes requested):

librz/core/disasm.c
ret2libc:
```
			case RZ_META_TYPE_RUN:
				rz_core_cmd0(core, meta->str);
				break;
			default:
```
Some stuff...

librz/core/disasm.c
ret2libc:
```suggestion
```

librz/core/disasm.c
kazarmy: It's needed?

kazarmy (approved):

librz/core/disasm.c
ret2libc: ... Who knows :(

kazarmy: I have a feeling a corrupt project file is going to be lots of !!Fun!!.

XVilka: Who needs shellcode when you have this.

XVilka: I suggest to remove it.

XVilka:
This `RZ_META_TYPE_RUN` was added in 2016:
- https://github.com/radareorg/radare2/commit/b04e1f2c4ce9dc0ca2144dfe74ece005d1e14f75
- It was supposed to fix the following issue https://github.com/radareorg/radare2/issues/6060, I think it's a hack

When you marked some region as data with `Cd` you will get the same behavior as the one in the issue.

librz/core/disasm.c
ret2libc: I'll do separate PR.

ret2libc:
> When you marked some region as data with `Cd` you will get the same behavior as the one in the issue.

True, however... `Cr` seems to be much more powerful and flexible.You can in theory print whatever you want, not only basic stuff. We might want to keep it.

kazarmy:
> True, however... `Cr` seems to be much more powerful and flexible.You can in theory print whatever you want, not only basic stuff. We might want to keep it.

Yeah but `Cr` shouldn't be able to format the SSD while it's doing that ... maybe there should be a way to sandbox it so that it can't affect the outside world or the binary.

ret2libc:
> > True, however... `Cr` seems to be much more powerful and flexible.You can in theory print whatever you want, not only basic stuff. We might want to keep it.
> 
> Yeah but `Cr` shouldn't be able to format the SSD while it's doing that ... maybe there should be a way to sandbox it so that it can't affect the outside world or the binary.

The command executed by `Cr` is a command the user configured. It can format your SSD the same as any other command you execute could, no?

kazarmy:
Yes but I prefer that there is at least some protection, the same way that you don't do day-to-day work using `root` ... a typo/bug could be horrifying.

ret2libc:
> Yes but I prefer that there is at least some protection, the same way that you don't do day-to-day work using `root` ... a typo/bug could be horrifying.

You could say the same of any other command you type in the rizin shell, no? I mean, you could wrongly type `!rm -rf /` in your rizin shell :D how is `Cr` worse?

XVilka:
@ret2libc this meta tag (and corresponding command) is stored during the serialization process. So opening a random project file may trigger it on your computer: https://github.com/rizinorg/rizin/blob/dev/librz/analysis/serialize_analysis.c#L1392

ret2libc:
> @ret2libc this meta tag (and corresponding command) is stored during the serialization process. So opening a random project file may trigger it on your computer: https://github.com/rizinorg/rizin/blob/dev/librz/analysis/serialize_analysis.c#L1392

That, yes. I agree it's not awesome to have this risk... It might still be useful for "trusted" uses. Not sure though :)

PR #838, opened 2021-03-16
core/disasm: use bin section info instead of maps to show name/perm

ret2libc:

**Detailed description**

Without this patch the disassembly would almost never print that a section is writable, because the info were taken from IO instead of Bin. Moreover, it increases a bit the size of the section part, if present, because otherwise even the common name `.text` is truncated.

**Test plan**

...

**Closing issues**

...


ret2libc:
I think the openbsd job is being canceled :/

XVilka (approved):
Good catch and it was planned to that, AFAIR.

PR #839, opened 2021-03-16
TEST Dist split plugins meson

ret2libc:
TEST


XVilka:
@ret2libc lets close this until we decide what to do with that? Also, as I said before, it's probably better done after some of the refactorings we planned are complete.

ret2libc:
Please do not delete the branch.

PR #870, opened 2021-03-19
Remove slow and almost unused Source File Reading from Debug Info

thestr4ng3r:

**Detailed description**

When you loaded a binary with dwarf info, rizin would attempt to slurp every single source file at load time, however:
* This may be undesired if it is an untrusted binary because the binary can tell rizin any file paths it wants and rizin will read them from the host
*  It was only used in the `rz_cons_printf("\"CC %s:%d\"@0x%" PFMT64x "\n"...` code path which is not covered by tests, so it's probably irrelevant and in the json where I think it's also irrelevant.

Even worse, if loading for a file failed (which is basically always the case if you don't have the source), it would re-attempt to read the file on every single line number+column sample, whose number in the same ballpark as the number of instructions in the entire binary, creating a huge bottleneck.

Little comparison:
Before:
```
florian-macbook:rizin florian$ time rz-test -j1 test/db/cmd/dwarf
Running from /Users/florian/dev/rizin/test
Loaded 9 tests.
[9/9]                       9 OK         0 BR        0 XX        0 FX
Finished in 36 seconds.

real	0m36.465s
user	0m8.416s
sys	0m1.195s
```

After:
```
florian-macbook:rizin florian$ time rz-test -j1 test/db/cmd/dwarf
Running from /Users/florian/dev/rizin/test
Loaded 9 tests.
[9/9]                       9 OK         0 BR        0 XX        0 FX
Finished in 8 seconds.

real	0m8.047s
user	0m7.934s
sys	0m0.099s
```
Though this is not necessarily represantative since macOS has horribly slow io. I would expect the difference to be a bit less on Linux.

Also I can load the binary from #676 in about 1m instead of 2m30s but there is still more room for optimization.

### DIFF
--- a/librz/core/cbin.c
+++ b/librz/core/cbin.c
@@ -1108,41 +1108,6 @@ static int bin_info(RzCore *r, PJ *pj, int mode, ut64 laddr) {
 	return true;
 }
 
-typedef struct {
-	size_t *line_starts;
-	char *content;
-	size_t line_count;
-} FileLines;
-
-static void file_lines_free(FileLines *file) {
-	if (!file) {
-		return;
-	}
-	free(file->line_starts);
-	free(file->content);
-	free(file);
-}
-
-FileLines *read_file_lines(const char *path) {
-	FileLines *result = RZ_NEW0(FileLines);
-	if (!result) {
-		return result;
-	}
-	result->content = rz_file_slurp(path, NULL);
-	if (result->content) {
-		result->line_starts = rz_str_split_lines(result->content, &result->line_count);
-	}
-	if (!result->content || !result->line_starts) {
-		RZ_FREE(result);
-	}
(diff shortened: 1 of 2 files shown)


karliss:
Created #872 for discussing how the source file handling could later be implemented better, since a proper rewrite is out of scope for this PR.

Maijin:
Isn't that also a dup of `bin_source` (what's used in `iX`) ?

ret2libc (approved):

thestr4ng3r:
> Isn't that also a dup of `bin_source` (what's used in `iX`) ?

No, `bin_source` just shows the filenames without loading them. But in fact there is still a duplicate of this in `rz_bin_addr2text()`, which however seems a bit more sane than this.

PR #873, opened 2021-03-20
Add a Test for C++ Compilation

thestr4ng3r:
C++ is pickier than C in some cases (I expect CI to be red here at the moment) so making sure rizin headers build with it would be useful.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1,4 +1,4 @@
-project('rizin', 'c',
+project('rizin', ['c', 'cpp'],
   version: 'v0.2.0-git',
   license: 'LGPL3',
   meson_version: '>=0.55.0',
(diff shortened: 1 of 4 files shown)


meson.build
```diff
@@ -1,4 +1,4 @@
-project('rizin', 'c',
+project('rizin', ['c', 'cpp'],
```
thestr4ng3r, line 1: I do not like this since it adds a strict dependency on having a C++ compiler for the entire project. But without it, meson will not build anything in C++. It should depend on the `enable_cpp_test` option but we can't check that before the `project()` call. @ret2libc maybe you have an idea here?

ret2libc: If we create just a separate meson.build for the cpp tests?

XVilka: Agree, we could even put it in the `.github/workflows` since it's just for the CI.

XVilka: It's also possible to run `meson rewrite kwargs set project / language "cpp"` or something like that.

ret2libc:
See https://github.com/rizinorg/rizin/pull/1743

PR #887, opened 2021-03-21
Filter stdout warning in luac test

Heersin:

**Detailed description**

see https://travis-ci.com/github/rizinorg/rizin/jobs/492588616

This pr filters the stdout for ARMv8

**Addtional**
These warnings should be sent to stderr, but they were sent to stdout on ARMv8

### DIFF
--- a/librz/bin/format/luac/luac_bin.c
+++ b/librz/bin/format/luac/luac_bin.c
@@ -136,7 +136,10 @@ static void free_rz_symbol(RzBinSymbol *symbol) {
 }
 
 LuacBinInfo *luac_build_info(LuaProto *proto) {
-	rz_return_val_if_fail(proto, NULL);
+	if (proto == NULL) {
+		eprintf("Warning : No proto for building info\n");
+		return NULL;
+	}
 
 	LuacBinInfo *ret = RZ_NEW0(LuacBinInfo);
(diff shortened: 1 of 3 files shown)


wargio:
i'm not sure this is the right fix

Maijin:
Better change that in the code to print it on stderr, also what does "Nothing" mean, can you add more verbosity to that error message? If it's not useful, can you please remove it completely.

Heersin:
@Maijin @wargio  Yes, It's not the right fix. This test case is used to check whether the plugin can handle a broken file and report an error message (report the truncated offset of this file). And expect there are no output in stdout.

`WARNING : ..... assertion failed` is generated by `rz_return_if_fail`, but I don't know why It would be sent to stdout. Is the behavior of `H_LOG_` in `rz_assert` is different from archs and compiler options ?

I fix it in code to force it use `eprintf` when checking pointer.

XVilka:
It is fine, since `rz_return_if_fail()` and similar functions shouldn't be used in case of the real-world warning/error - it targets the developers' own errors rather than the unexpected situation in the data. Rizin aims to handle even botched and mangled data, so it should not spit warnings in vain.

XVilka (approved):

ret2libc (approved):

PR #897, opened 2021-03-24
FIX #334 Resolve data relocations in 32bit position-independent binaries

### LINKED ISSUE #334: Resolve data relocations in 32bit position-independent binaries
## Description
Modern distributions of GCC produce position-independent binaries in 32bit mode by default. It's implemented by calling a utility function like `__x86.get_pc_thunk.ax` which just returns the return address from the stack; then all the addressing is performed relatively to the obtained address. Rizin, however, doesn't recognize this pattern, so it's not easy to analyze even a simple helloworld statically:
```
┌ (fcn) main 60
│   int main (int argc, char **argv, char **envp);
│           ; var int var_8h @ ebp-0x8
│           ; arg int arg_4h @ esp+0x4
│           0x00001199      8d4c2404       lea ecx, [arg_4h]
│           0x0000119d      83e4f0         and esp, 0xfffffff0
│           0x000011a0      ff71fc         push dword [ecx - 4]
│           0x000011a3      55             push ebp
│           0x000011a4      89e5           mov ebp, esp
│           0x000011a6      53             push ebx
│           0x000011a7      51             push ecx
│           0x000011a8      e828000000     call sym.__x86.get_pc_thunk.ax
│           0x000011ad      052b2e0000     add eax, 0x2e2b             ; '+.'
│           0x000011b2      83ec0c         sub esp, 0xc
│           0x000011b5      8d9030e0ffff   lea edx, [eax - 0x1fd0]
│           0x000011bb      52             push edx                    ; const char *s
│           0x000011bc      89c3           mov ebx, eax
│           0x000011be      e86dfeffff     call sym.imp.puts           ; int puts(const char *s)
│           0x000011c3      83c410         add esp, 0x10
│           0x000011c6      b800000000     mov eax, 0
│           0x000011cb      8d65f8         lea esp, [var_8h]
│           0x000011ce      59             pop ecx
│           0x000011cf      5b             pop ebx
│           0x000011d0      5d             pop ebp
│           0x000011d1      8d61fc         lea esp, [ecx - 4]
└           0x000011d4      c3             ret
```
The value of edx prior to puts call is always str.Hello_World, which is hard to see without debugging the program.

## Describe the solution you'd like
The `*get_pc_thunk*` family of functions (or maybe the pattern of returning the return address) could be treated in a special way, letting the analyzer know that they're used for EIP-relative addressing. It would let the analyzer resolve EIP-relative data references.

## Additional context
Here's a screenshot of IDA, which recognizes the string literal (the offset of `str.Hello_World` is **0x2008**):

image

[The helloworld binary from the example](https://github.com/radare/radare2/files/2955726/hello.zip)

by @loskutov

Additional notes by @fabianMendez:

In android shared libraries (with PIC) references are calculated in a way which confuses radare2
Firsst, it sets a register to a base address (.got.plt VA). I've found two different methods:

### `call $+5 + pop reg (related to #11309)`

```
0x001b7d36      e800000000     call 0x1b7d3b
0x001b7d3b      5b             pop ebx
0x001b7d3c      81c30d4c7b01   add ebx, 0x17b4c0d
```

###  `call function which return $+5`
```
0x00000c86      e8b5ffffff     call fcn.mov_ebx_esp
0x00000c8b      81c3e1220000   add ebx, 0x22e1

; fcn.mov_ebx_esp ();
0x00000c40      8b1c24         mov ebx, dword [esp]
0x00000c43      c3             ret
```
Then it uses this register across the function to get references to global variables or strings:
```
0x00001025      8d83f2e4ffff   lea eax, [ebx - 0x1b0e]
0x0000102b      890424         mov dword [esp], eax
0x0000102e      e86dfaffff     call sym.imp.opendir
```

I'd like to get register base cross references in shared libraries with PIC enabled
```
0x00001025      8d83f2e4ffff   lea eax, str.proc_self_id
0x0000102b      890424         mov dword [esp], eax
0x0000102e      e86dfaffff     call sym.imp.opendir
```

If a library has PIC enabled Rizin should not make references like this:
```
; "_traitsIS6_EEEEEEvRNS0_11timer_queueIT_EERKNSB_9time_typeERNSC_14per_timer_dataEPNS0_7wait_opE"  │                         
0x001b7d54 6806000100     push 0x10006
```
```
[0x001b7d2b]> ps @ 0x10006
_traitsIS6_EEEEEEvRNS0_11timer_queueIT_EERKNSB_9time_typeERNSC_14per_timer_dataEPNS0_7wait_opE
```
I think should be assumed addresses are gonna be calculated in some way

Reference binaries:
[libnative-lib.zip](https://github.com/radare/radare2/files/2683213/libnative-lib.zip)
[libtermux.zip](https://github.com/radare/radare2/files/2683214/libtermux.zip)

See the file parsing at 
- `librz/bin/format/elf/`
- `librz/bin/p/bin_elf.c`

e888llll:

**Detailed description**

...

**Test plan**

...

**Closing issues**

...


librz/core/disasm.c
```diff
@@ -348,6 +348,12 @@ static void ds_end_line_highlight(RDisasmState *ds);
 static bool line_highlighted(RDisasmState *ds);
 static int ds_print_shortcut(RDisasmState *ds, ut64 addr, int pos);
 
+// add for str issue
```
XVilka: Please don't use global variables - Rizin is a library, theoretically can have multiple contexts to work in, it would break the logic.

XVilka (changes requested):

librz/core/disasm.c
XVilka: Also please write more meaningful commit message and PR description, see our
- [Contribution Guide](https://github.com/rizinorg/rizin/blob/dev/CONTRIBUTING.md)
- [Developers Guide](https://github.com/rizinorg/rizin/blob/dev/DEVELOPERS.md)

XVilka: Also there is no need for these `// add for str issue` comments.

e888llll: OK. I will modify it.

librz/core/disasm.c
```diff
@@ -1097,6 +1103,11 @@ static void ds_build_op_str(RDisasmState *ds, bool print_color) {
 		ds->opstr = strdup(ds->str);
 	}
 	ds->opstr = ds_sub_jumps(ds, ds->opstr);
+
+	// add for str issue
+	inst_bufferstr = (char *)malloc(sizeof(char) * (strlen(ds->opstr) + 1));
```
wargio: if you have to copy a string there are many way to do this, my suggestion is to use `strdup`

librz/core/disasm.c
```diff
@@ -348,6 +348,12 @@ static void ds_end_line_highlight(RDisasmState *ds);
 static bool line_highlighted(RDisasmState *ds);
 static int ds_print_shortcut(RDisasmState *ds, ut64 addr, int pos);
 
+// add for str issue
+int begin_mark = 0;
```
wargio: please avoid any global variable.

librz/core/disasm.c
```diff
@@ -1131,6 +1142,85 @@ static void ds_build_op_str(RDisasmState *ds, bool print_color) {
...
+				if (strcmp(s->name, ".got") == 0) {
+					got_beginaddr = s->vaddr;
+				}
+				if (strcmp(s->name, ".rodata") == 0) {
+					rodata_beginaddr = s->vaddr;
+					rodata_size = s->size;
+				}
+			}
+
+			char *numstr = (char *)malloc(sizeof(char) * 11);
```
wargio: use `RZ_NEW0` instead of malloc + memset

librz/core/disasm.c
```diff
@@ -1131,6 +1142,85 @@ static void ds_build_op_str(RDisasmState *ds, bool print_color) {
...
+			memset(numstr, 0, sizeof(char) * 11);
+			strncpy(numstr, markinst[1], 10);
+			sscanf(numstr, "%llx", &next_instaddr);
```
wargio: there is a parsing library for numbers in rizin. check [rz_num_get_input_value](https://github.com/rizinorg/rizin/blob/dev/librz/include/rz_util/rz_num.h#L76) (for example `rz_num_get_input_value(NULL, next_instaddr)`)

librz/core/disasm.c
```diff
@@ -1131,6 +1142,85 @@ static void ds_build_op_str(RDisasmState *ds, bool print_color) {
 				core->parser->subrel_addr = killme;
 			}
 		}
+
+		// add for str issue
+		if (begin_mark && markinst_num > 1 && strstr(ds->opstr, "0x")) {
```
wargio: you can use [rz_num_is_valid_input](https://github.com/rizinorg/rizin/blob/dev/librz/include/rz_util/rz_num.h#L75)

librz/core/disasm.c
```diff
@@ -1131,6 +1142,85 @@ static void ds_build_op_str(RDisasmState *ds, bool print_color) {
...
+				offset_got_str_2 = got_beginaddr - rodata_beginaddr;
+			} else {
+				offset_got_str_1 = rodata_beginaddr - got_beginaddr;
+				offset_got_str_2 = rodata_beginaddr - got_beginaddr + rodata_size;
+			}
+
+			// analyze the offset
+			bool find_offset_got_next = false;
+
+			char eip_reg[4] = "eax";
```
wargio: be sure this code runs only when x86/x64 is selected

librz/core/disasm.c
```diff
@@ -1131,6 +1142,85 @@ static void ds_build_op_str(RDisasmState *ds, bool print_color) {
...
+				if (str_addr <= offset_got_str_2 && str_addr >= offset_got_str_1) {
+					if (got_beginaddr < rodata_beginaddr) {
+						str_addr = got_beginaddr + str_addr;
+					} else {
+						str_addr = got_beginaddr - str_addr;
+					}
+
+					if (find_offset_got_next) {
+						RzFlagItem *str_flag = rz_flag_get_at(core->flags, str_addr, false);
+						ds->comment = (char *)malloc(sizeof(char) * (strlen(str_flag->name) + 1));
```
wargio: please use `ds->comment = rz_str_newf("; %s", str_flag->name)`

librz/core/disasm.c
```diff
@@ -1131,6 +1142,85 @@ static void ds_build_op_str(RDisasmState *ds, bool print_color) {
...
+				if (numstr) {
+					find_offset_got_next = true;
+					break;
+				}
+			}
+
+			if (find_offset_got_next) {
+				ut64 str_addr = 0;
+				numstr = strstr(ds->opstr, "0x");
+				sscanf(numstr, "%llx", &str_addr);
```
wargio: as above, same here.

librz/core/disasm.c
```diff
@@ -1131,6 +1142,85 @@ static void ds_build_op_str(RDisasmState *ds, bool print_color) {
...
+				}
+			}
+
+			char *numstr = (char *)malloc(sizeof(char) * 11);
+			memset(numstr, 0, sizeof(char) * 11);
+			strncpy(numstr, markinst[1], 10);
+			sscanf(numstr, "%llx", &next_instaddr);
+			ut64 offset_got_next = (got_beginaddr > next_instaddr) ? (got_beginaddr - next_instaddr) : (next_instaddr - got_beginaddr);
+			char offset_got_next_str[11] = { 0 };
+			snprintf(offset_got_next_str, 11, "0x%llx", offset_got_next);
```
wargio: use `snprintf(offset_got_next_str, sizeof(offset_got_next_str), "0x%" PFMT64x, offset_got_next);`

wargio (approved):

librz/core/disasm.c
wargio: you also don't seem to free it.

XVilka:
Regarding the core logic of this, I am not sure it should be in `disasm.c` at all. cc @thestr4ng3r @ret2libc

e888llll:
I think the problem needs to find two key offsets:
offset1 = .got beginning addr - "__x86.get_pc_thunk." next instruction addr
offset2 = .got beginning addr - const string addr
(The addr above based on the binary file, not the real addr in memory. And offset2 is a range because the string is unknown.)

The resolution mainly exists in disasm.c rz_core_print_disasm and ds_build_op_str.
1) rz_core_print_disasm analyzes every instruction (i)：If i contains "__x86.get_pc_thunk.", start to mark down the following instrucions.
2) Before ds_build_op_str colorize the opstr, analyze whether it can refer to a const string. Make the following judgements：
a. "__x86.get_pc_thunk." has showed up.
b. The instructions from "__x86.get_pc_thunk." to this opstr contain the offset1.
c. This opstr contains an addr and the addr is in the range of offset2.
Then it can guess the string.

The resolution may not be accurate(Dynamic execution is more accurate) but I think to a certain extent, it can resolve the problem.

librz/core/disasm.c
e888llll: Thanks for your advice. I have updated the code.

librz/core/disasm.c
```diff
@@ -1165,14 +1157,12 @@ static void ds_build_op_str(RDisasmState *ds, bool print_color) {
 				}
 			}
 
-			char *numstr = (char *)malloc(sizeof(char) * 11);
-			memset(numstr, 0, sizeof(char) * 11);
+			char *numstr = (char *)calloc(11, sizeof(char));
```
wargio: please use RZ_NEW0 or RZ_NEWS0

wargio: or just `char numstr[11] = {0};`

librz/core/disasm.c
```diff
@@ -1213,12 +1200,12 @@ static void ds_build_op_str(RDisasmState *ds, bool print_color) {
 
 					if (find_offset_got_next) {
 						RzFlagItem *str_flag = rz_flag_get_at(core->flags, str_addr, false);
-						ds->comment = (char *)malloc(sizeof(char) * (strlen(str_flag->name) + 1));
-						memset(ds->comment, 0, strlen(str_flag->name) + 1);
-						snprintf(ds->comment, strlen(str_flag->name) + 3, "; %s", str_flag->name);
+						ds->comment = rz_str_newf("; %s", str_flag->name);
```
wargio: i would check if str_flag is null.

wargio (changes requested):

XVilka:
My thoughts:
- This code should be limited to `asm.arch=x86` only
- Probably worth to split the code into parsing function and put it into `librz/core/cbin.c` or `librz/core/canalysis.c`
- Leave only displaying in the `librz/core/disasm.c`

librz/core/disasm.c
```diff
@@ -5469,7 +5551,21 @@ RZ_API int rz_core_print_disasm(RzPrint *p, RzCore *core, ut64 addr, ut8 *buf, i
 			}
 			ds_print_lines_right(ds);
 			ds_print_optype(ds);
-			ds_build_op_str(ds, true);
+			char *buildopstr_ret = ds_build_op_str(ds, true, markinst, markinst_num);
+
+			if (buildopstr_ret && rz_str_startswith(rz_config_get(core->config, "asm.arch"), "x86") && strcmp(rz_config_get(core->config, "asm.bits"), "32") == 0) {
```
XVilka, line 5556: In general we recommend to extract long conditions into the short inline functions.

ret2libc:
> Regarding the core logic of this, I am not sure it should be in `disasm.c` at all. cc @thestr4ng3r @ret2libc

I doubt it should be in disasm.c. There there should be only print-related logic, while this is a bit about analysis, since you first have to recognize the "get PC" pattern and then calculate/emulate the other addresses based on that.

XVilka:
@08A @thestr4ng3r should we close this and implement on the RzBin level? Could you please add a test for this, mark as broken for now?

ret2libc:
Ok for closing.

PR #901, opened 2021-03-24
Mention pointer modifiers in developer documentation

### LINKED ISSUE #863: Explain RZ_OWN, RZ_BORROW, etc in developers documentation
There are some conventions that are used in Rizin code but unexplained in DEVELOPERS.md.
They are:
- `RZ_OWN`
- `RZ_BORROW`
- `RZ_NULLABLE`
- `RZ_NONNULL`
- `RZ_INOUT`
- `RZ_OUT`
- `RZ_DEPRECATE`

They are defined in `librz/include/rz_types.h` and briefly explained there, but it's not enough for new developers.

chinggg:

**Detailed description**

Explain RZ_OWN, RZ_BORROW, etc(which I called `pointer modifiers`) in developers documentation.
Also add a missing leading `*` in an unordered list.

...

**Test plan**

...

**Closing issues**

closes #863 
...

If you think something in the document needs to be improved, feel free to request changes.


wargio (approved):

XVilka (approved):

DEVELOPERS.md
```diff
@@ -320,3 +318,77 @@ In particular, the SPDX header may look like:
...
+  // s was RZ_OWN, so main does not need to free it. s is now owned by `m`
+  // ... use m ....
+  // m was RZ_BORROW, so main is just borrowing it from `f`, and it does not have to free it.
+  my_file_free(f);
+  // f was created by main and never transferred to anything else, so main needs to free it.
+}
+```
+
+- You should use these modifiers consistently in both function definition and declaration.
+- We should use these modifiers when and only when it makes sense. For example, if your function returns `const char *`, the caller should not free it because of the `const`. So specifying `RZ_BORROW` in this case is probably redundant.
```
ret2libc:
```suggestion
- You should use these modifiers when and only when it makes sense. For example, if your function returns `const char *`, the caller should not free it because of the `const`. So specifying `RZ_BORROW` in this case is probably redundant.
```
For consistency with the sentence above

PR #928, opened 2021-03-29
Update rizin-shell-parser to tree-sitter 0.19.4

### LINKED ISSUE #755: Rizin segfaults with Tree-Sitter 0.19.x
### Work environment

| OS/arch/bits (mandatory)                             | Fedora 33 x86_64
| Architecture/bits of the file (mandatory)            |  x86/64
| `rizin -v` full output, **not truncated** (mandatory)  | https://github.com/rizinorg/rizin/commit/7423f10be0a4bc481044955a27ba0de43793a9da

### Expected behavior

No segfault.

### Actual behavior

```
rizin on  tree-sitter-0.19 [$] 
[i] ℤ gdb rizin                                                                                                                                                                                                                   12:24:21 
Reading symbols from rizin...
(gdb) run /bin/ls
Starting program: /home/user/.local/bin/rizin /bin/ls
Missing separate debuginfos, use: dnf debuginfo-install glibc-2.32-4.fc33.x86_64
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib64/libthread_db.so.1".

Program received signal SIGSEGV, Segmentation fault.
0x00007ffff669cfbc in ts_tree_root_node (self=0x0) at ../subprojects/tree-sitter-0.19.0/lib/src/./tree.c:43
43        return ts_node_new(self, &self->root, ts_subtree_padding(self->root), 0);
Missing separate debuginfos, use: dnf debuginfo-install bzip2-libs-1.0.8-4.fc33.x86_64 openssl-libs-1.1.1i-3.fc33.x86_64 xz-libs-5.2.5-4.fc33.x86_64 zlib-1.2.11-23.fc33.x86_64
(gdb) 
```

- Update subproject wrap to use Tree-Sitter 0.19.0: https://github.com/tree-sitter/tree-sitter/releases/tag/v0.19.0
- Build
- Run `rizin /bin/ls` or `rizin -v`

Latest Tree-Sitter release is https://github.com/tree-sitter/tree-sitter/releases/tag/v0.19.1

ret2libc:

**Detailed description**

It just re-generates the grammar with newer tree-sitter. Unfortunately generating the grammar in meson might be a bit too complicated as you need to run `npm install` and then tree-sitter generate... I'd leave it like this for now.

**Test plan**

...

**Closing issues**

Closes https://github.com/rizinorg/rizin/issues/755


XVilka:
```

Run find . -name "*.js" | grep -v "subprojects" | xargs npx prettier  --print-width 120 --check
Checking formatting...
[warn] shlr/rizin-shell-parser/bindings/node/index.js
[warn] Code style issues found in the above file(s). Forgot to run Prettier?
Error: Process completed with exit code 123.
```
Probably better to ignore this file in the CI linter.

karliss:
How about adding instructions in build documentation explaining how to regenerate the parser since anyone using use_sys_treesitter will likely need it.

Is npm install really needed or is it just one of the methods for installing tree-sitter generator which is itself written in rust? If it's just that then I would assume that in case of use_sys_treesitter everything should already be available.

ret2libc:
@karliss Are you referring to https://github.com/rizinorg/rizin/blob/dev/shlr/rizin-shell-parser/README.md ?

ret2libc:
> Is npm install really needed or is it just one of the methods for installing tree-sitter generator which is itself written in rust? If it's just that then I would assume that in case of use_sys_treesitter everything should already be available.

Yeah we might re-generate it only if use_sys_treesitter is true. I'd need to check what's provided in arch linux.

librz/core/cmd.c
```diff
@@ -6183,6 +6183,7 @@ static RzCmdStatus core_cmd_tsr2cmd(RzCore *core, const char *cstr, bool split_l
 	ts_parser_set_language(parser, (TSLanguage *)core->rcmd->language);
 
 	TSTree *tree = ts_parser_parse_string(parser, NULL, input, strlen(input));
+	rz_return_val_if_fail(tree, RZ_CMD_STATUS_INVALID);
```
wargio: shouldn't this be an if instead of that ?

ret2libc: it seems this happen only if something completely wrong is going on (e.g. in this case because the version used to generate the tree-sitter parser was incompatible with the one linked with rizin). However you are right as here it's probably better to have a `if`, but I will place a `rz_return_val_if_fail` on the return value of the call to `ts_parser_set_language`, because when that returns false there was a version mismatch and that's an error we might consider as a programmer's one and where i'd like to fail loudly and early.

sys/meson_tree_sitter_generate.py
```diff
@@ -0,0 +1,37 @@
...
+
+def main():
+    tree_sitter_exe, output_dir, grammar_js = parse()
+
+    grammar_js = os.path.abspath(grammar_js)
+    output_dir = os.path.abspath(output_dir)
+    tree_sitter_exe = os.path.abspath(tree_sitter_exe)
+
+    os.chdir(output_dir)
+    subprocess.run([tree_sitter_exe, 'generate', grammar_js], check=True)
```
karliss: Subprocess has `cwd` argument, no need to use os.chdir.

wargio (approved):

XVilka (approved):

Issue #329, opened 2021-01-07
Rizin report warning messages increasely when repeated run `aaa`


fangxlmr:
### Work environment

| Questions                                            | Answers
|------------------------------------------------------|--------------------
| OS/arch/bits (mandatory)                             | CentOS x86_64
| File format of the file you reverse (mandatory)      | ELF
| Architecture/bits of the file (mandatory)            | x86/64
| rizin -v full output, **not truncated** (mandatory)  | rizin 0.1.0-git 0 @ linux-x86-64 git.0.1.0-git commit: unknown build: 2021-01-07__13:06:23

### Expected behavior

Not so much warning when I run `aaa` command multiple times

### Actual behavior

Increasing warning messages after repeated running of `aaa`

### Steps to reproduce the behavior 
- open the binary
- run `aaa` repeatly

### Additional Logs, screenshots, source-code,  configuration dump, ...

```shell
~ rizin 001.make.elf.x86_64 
 -- Use /m to carve for known magic headers. speedup with search.
[0x00407f1c]> aaa
[x] Analyze all flags starting with sym. and entry0 (aa)
[x] Analyze function calls (aac)
[x] Analyze len bytes of instructions for references (aar)
[x] Check for vtables
[x] Type matching analysis for all functions (aaft)
[x] Propagate noreturn information
[x] Integrate dwarf function information.
[x] Use -AA or aaaa to perform additional experimental analysis.
[0x00407f1c]> aaa
[WARNING: rz_reg_get: assertion 'reg && name' failed (line 308)
WARNING: rz_reg_get: assertion 'reg && name' failed (line 308)
WARNING: rz_reg_get: assertion 'reg && name' failed (line 308)
WARNING: rz_reg_get: assertion 'reg && name' failed (line 308)
WARNING: rz_reg_get: assertion 'reg && name' failed (line 308)
WARNING: rz_reg_get: assertion 'reg && name' failed (line 308)
[x] Analyze all flags starting with sym. and entry0 (aa)
[x] Analyze function calls (aac)
[x] Analyze len bytes of instructions for references (aar)
[x] Check for vtables
[x] Type matching analysis for all functions (aaft)
[x] Propagate noreturn information
[x] Integrate dwarf function information.
[x] Use -AA or aaaa to perform additional experimental analysis.
[0x00407f1c]> aaa
[WARNING: rz_reg_get: assertion 'reg && name' failed (line 308)
[...]

PR #940, opened 2021-03-31
Fix undefined behaviour of the aaa analysis

ghost:

**Detailed description**

Rizin report warning messages increasingly when repeated run aaa.
This is due to the fact that analysis information aren't reset between
2 aaa analysis so the blocks analysis isn't deterministic.

The simple 'hack' is to force the fcns analysis data to be reset.

I add a simple test to verify that the second analysis isn't overriding
the first one.

I believe that if this ´solution´ is accepted, we should update the aaa help.

And the test suite will fail, because I didn´t add the binary to https://github.com/rizinorg/rizin-testbins. I would like to have some advice on this first.

**Test plan**

see  #329 and the new test

**Closing issues**

 #329

### DIFF
--- a/librz/core/canalysis.c
+++ b/librz/core/canalysis.c
@@ -4771,6 +4771,10 @@ static bool isSkippable(RzBinSymbol *s) {
 	return false;
 }
 
+static bool has_old_analysis_value(RzCore *core) {
+	return core->analysis->fcns->length > 0;
+}
+
 RZ_API int rz_core_analysis_all(RzCore *core) {
 	RzList *list;
 	RzListIter *iter;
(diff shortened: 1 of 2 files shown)


librz/core/canalysis.c
```diff
@@ -4771,6 +4771,10 @@ static bool isSkippable(RzBinSymbol *s) {
 	return false;
 }
 
+static bool has_old_analysis_value(RZ_BORROW RZ_NONNULL RzCore *core) {
```
XVilka: I am not sure `RZ_BORROW/RZ_NONNULL` here are necessary, it's very short and static function.

ghost:
@XVilka Do you think it is an acceptable solution/patch?

XVilka:
I don't think it's a right fix, but not 100% sure.

ret2libc:
What about functions defined by the user? I'm not sure either this is the right fix...

ghost:
~The bug come from the type matching analysis for all functions (aaft).~

rz_analysis_dwarf_integrate_functions introduce some side effect.

It introduce some side effect that impact (aa).

ghost:
see #945, I switch to a better src branch. Sorry.

PR #951, opened 2021-04-01
Link sdb and spp libs wholly to rz_util

p371jah:

**Detailed description**

Requires https://github.com/rizinorg/sdb/pull/25

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/librz/core/meson.build
+++ b/librz/core/meson.build
@@ -103,8 +103,9 @@ rz_core_deps = [
   rz_bin_dep,
   platform_deps,
   gdb_dep,
+  shell_parser_dep,
   lrt,
-  shell_parser_dep
+  mth
 ]
 
 if libuv_dep.found()
--- a/librz/lang/meson.build
+++ b/librz/lang/meson.build
@@ -11,10 +11,8 @@ rz_lang_sources = [
   #'p/spp.c',
 ]
 
-spp_inc = [platform_inc, include_directories('../../shlr/spp')]
-
 rz_lang = library('rz_lang', rz_lang_sources,
-  include_directories: [platform_inc, spp_inc],
+  include_directories: platform_inc,
   c_args: library_cflags,
   dependencies: [rz_util_dep, rz_cons_dep],
   install: true,
--- a/librz/util/meson.build
+++ b/librz/util/meson.build
@@ -83,7 +83,7 @@ rz_util_sources = [
   'subprocess.c',
 ]
 
-rz_util_deps = [ldl, lrt, mth, spp_dep, pth, utl, sdb_dep, zlib_dep, platform_deps]
+rz_util_deps = [ldl, lrt, mth, pth, utl, sdb_dep.as_link_whole(), spp_dep.as_link_whole(), zlib_dep, platform_deps]
 if host_machine.system().startswith('freebsd') or host_machine.system().startswith('netbsd') or host_machine.system().startswith('haiku')
   # backtrace_symbols_fd requires -lexecinfo
   rz_util_deps += [cc.find_library('execinfo', static: is_static_build)]
@@ -106,9 +106,14 @@ rz_util = library('rz_util', rz_util_sources,
   soversion: rizin_libversion
 )
 
-rz_util_dep = declare_dependency(link_with: rz_util,
-                                include_directories: platform_inc,
-                                dependencies: rz_util_deps)
+rz_util_dep = declare_dependency(
+  link_with: rz_util,
+  include_directories: platform_inc,
+  dependencies: [
+    sdb_dep.partial_dependency(includes: true),
+    spp_dep.partial_dependency(includes: true)
+  ]
+)
 
 pkgconfig_mod.generate(rz_util,
   subdirs: ['librz', 'librz/sdb'],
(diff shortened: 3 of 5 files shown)


thestr4ng3r:
OpenBSD:
```
librz/util/meson.build:88:0: ERROR: Unknown method "as_link_whole" in object.
```

p371jah:
`as_link_whole` was introduced in meson 0.56....

p371jah:
Well if it isn't allowed to change minimal required meson version, we can just add new dependency with `link_whole` to `sdb` project.

XVilka:
No, I think it's fine to use newer meson, we just could install it from `pip`.

XVilka:
@pelijah could you please just revert back to `as_link_whole` but add `pip install meson` to OpenBSD? I think this is a better way. cc @ret2libc @thestr4ng3r 
Eventually OpenBSD will update their meson version too.

p371jah:
> @pelijah could you please just revert back to `as_link_whole` but add `pip install meson` to OpenBSD?

For `spp_dep`? I think it would be better to use `link_whole` directly since `spp_dep` is declared in main `meson.build` file.

XVilka:
Ok, no problem then. I thought it's about SDB.

XVilka:
By the way, seems OpenBSD ports have updated Meson already: https://openports.se/devel/meson

ret2libc:
Ok fix the `WARNING: Project targeting '>=0.50.0' but tried to use feature introduced in '0.56.0': dep.as_link_whole.` please.

ret2libc:
Got this while compiling on my system:
```
/usr/bin/ld: librz/core/librz_core.so.0.2.0-git.p/cmd.c.o: in function `is_tcache_32':
/home/rschiron/projects/rizinorg/rizin/build-sdb-whole/../librz/core/linux_heap_glibc.c:130: undefined reference to `round'
```
I have no idea how it relates to your PR though...

p371jah:
> Ok fix the `WARNING: Project targeting '>=0.50.0' but tried to use feature introduced in '0.56.0': dep.as_link_whole.` please.

Requires to update minimal meson version in `sdb` too.

ret2libc:
> > Ok fix the `WARNING: Project targeting '>=0.50.0' but tried to use feature introduced in '0.56.0': dep.as_link_whole.` please.
> 
> Requires to update minimal meson version in `sdb` too.

If we do it for rizin it should be no problem doing it for sdb as well.

However we should be aware that if we update too much the min required meson version we risk that some distros won't be able to easily package rizin in their distros.

p371jah:
> However we should be aware that if we update too much the min required meson version we risk that some distros won't be able to easily package rizin in their distros.

Thats why I have proposed a workaround

ret2libc:
> > However we should be aware that if we update too much the min required meson version we risk that some distros won't be able to easily package rizin in their distros.
> 
> Thats why I have proposed a workaround

I agree with you. In my opinion we should wait a little bit to update the min required version to 0.56.0 . Ubuntu 20.10, for example, seems to still use 0.55.3, so it would be a problem there.

karliss:
Another issue with meson version increase is that 0.56 raises minimum python version from 3.5 to 3.6. And 3.5 is the version available on Ubuntu 16.04. So you can't even install it using pip unless you also install a newer python.

XVilka:
It's always possible to build Python 3.6 from sources, even on something old like Debian Wheezy: https://github.com/rizinorg/rizin/runs/2247159748#step:6:1

ret2libc:
> It's always possible to build Python 3.6 from sources, even on something old like Debian Wheezy: https://github.com/rizinorg/rizin/runs/2247159748#step:6:1

Yes but that's not something distros are going to do for sure. Packagers have to use the system provided python/meson to build packages usually, so it's better to use a workaround for link_whole IMO.

karliss:
My previous comment might be wrong, looking at the meson repo  and Cutter CI one more time indicates that 0.56.0 might be still compatible with python 3.5.  Cutter CI contains
```
 # 0.56.1 doesn't work with python 3.5 on Ubuntu 16.04
pip install meson==0.56.0
```
Looking at the commits it might be that 0.56.1 broke the compatibility and then there are commits in 0.56.2 removing some usage of python 3.6 features.   Of course I don't mind against ability to compile on 20.04 using system meson.

p371jah:
Requires https://github.com/rizinorg/sdb/pull/26

p371jah:
> 
> 
> Got this while compiling on my system:
> 
> ```
> /usr/bin/ld: librz/core/librz_core.so.0.2.0-git.p/cmd.c.o: in function `is_tcache_32':
> /home/rschiron/projects/rizinorg/rizin/build-sdb-whole/../librz/core/linux_heap_glibc.c:130: undefined reference to `round'
> ```
> 
> I have no idea how it relates to your PR though...

@ret2libc Please try again to verify my fix

ret2libc (approved):

XVilka:
I merged https://github.com/rizinorg/sdb/pull/25. Please update this PR with the new SDB commit.

XVilka (approved):

PR #961, opened 2021-04-03
Dwarf Fixes

thestr4ng3r:
**DO NOT SQUASH**

 

**Test plan**

Simply load the newly-added test bins of https://github.com/rizinorg/rizin-testbins/commit/69da6b9b0c15196cb32867370dd8e40ce6c0eacb and https://github.com/rizinorg/rizin-testbins/commit/017fa78f350e7046bf0c374502523e5a188e602e in rizin, before these fixes it segfaults on both.

### DIFF
--- a/librz/analysis/dwarf_process.c
+++ b/librz/analysis/dwarf_process.c
@@ -241,20 +241,25 @@ static st32 parse_array_type(Context *ctx, ut64 idx, RzStrBuf *strbuf) {
 }
 
 /**
- * @brief Recursively parses type entry of a certain offset into strbuf
+ * \brief Recursively parses type entry of a certain offset into strbuf
  *        saves type size into *size
  *
- * @param ctx
- * @param offset offset of the type entry
- * @param strbuf string to store the type into
- * @param size ptr to size of a type to fill up (can be NULL if unwanted)
- * @return st32 -1 if error else DW_TAG of the entry
+ * \param ctx
+ * \param offset offset of the type entry
+ * \param strbuf string to store the type into
+ * \param size ptr to size of a type to fill up (can be NULL if unwanted)
+ * \param set of visited die offsets, to prevent infinite recursion
+ * \return st32 -1 if error else DW_TAG of the entry
  *
  * TODO make cache for type entries, one type is usually referenced
  * multiple times which means it's parsed multiple times instead of once
  */
-static st32 parse_type(Context *ctx, const ut64 offset, RzStrBuf *strbuf, ut64 *size) {
-	rz_return_val_if_fail(strbuf, -1);
+static st32 parse_type(Context *ctx, const ut64 offset, RzStrBuf *strbuf, ut64 *size, RZ_NONNULL SetU *visited) {
+	rz_return_val_if_fail(strbuf && visited, -1);
+	if (set_u_contains(visited, offset)) {
+		return -1;
+	}
+	set_u_add(visited, offset);
 	RzBinDwarfDie *die = ht_up_find(ctx->die_map, offset, NULL);
 	if (!die) {
 		return -1;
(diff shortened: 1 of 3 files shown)


test/db/cmd/dwarf
```diff
@@ -1797,3 +1797,1265 @@ EXPECT=<<EOF
...
+    DW_AT_decl_file                DW_FORM_data1                 
+    DW_AT_decl_line                DW_FORM_data1                 
+    DW_AT_decl_column              DW_FORM_data1                 
+    DW_AT_type                     DW_FORM_ref4                  
+    DW_AT_external                 DW_FORM_flag                  
+    DW_AT_location                 DW_FORM_block1                
+   3      DW_TAG_base_type          [no children] (0x26)
+    DW_AT_byte_size                DW_FORM_data1                 
+    DW_AT_encoding                 DW_FORM_data1                 
+    DW_AT_name                     DW_FORM_data4
```
thestr4ng3r, line 1827: Binary patched here. Normally `DW_AT_name` would have form `strp` or `string`. `DW_FORM_data4` does not make much sense.
This causes the nonsensical
```
<0x42>: Abbrev Number: 3    (DW_TAG_base_type)
     DW_AT_byte_size           : 4
     DW_AT_encoding            : 5
     DW_AT_name                : 7630441
```
below (note that DW_AT_name has some random integer value now)

test/db/cmd/dwarf
```diff
@@ -1797,3 +1797,1265 @@ EXPECT=<<EOF
...
+     DW_AT_artificial          : 1
+<0xd6>: Abbrev Number: 0    (DW_TAG_null_entry)
+<0xd7>: Abbrev Number: 0    (DW_TAG_null_entry)
+<0xd8>: Abbrev Number: 10   (DW_TAG_const_type)
+     DW_AT_type                : <0x31>
+<0xdd>: Abbrev Number: 11   (DW_TAG_pointer_type)
+     DW_AT_byte_size           : 8
+     DW_AT_type                : <0x31>
+<0xe3>: Abbrev Number: 10   (DW_TAG_const_type)
+     DW_AT_type                : <0xe3>
```
thestr4ng3r, line 2325: Patched here. `0xe3` is a const type of `0xe3`.

librz/analysis/dwarf_process.c
```diff
@@ -316,51 +322,66 @@ static st32 parse_type(Context *ctx, const ut64 offset, RzStrBuf *strbuf, ut64 *
...
 		if (type_idx != -1) {
-			parse_type(ctx, die->attr_values[type_idx].reference, strbuf, size);
+			parse_type(ctx, die->attr_values[type_idx].reference, strbuf, size, visited);
 		}
 		rz_strbuf_append(strbuf, " &");
 		break;
 	default:
 		break;
 	}
+	set_u_delete(visited, offset);
```
ret2libc, line 364: so is `visited` just to avoid infinite recursion? what happens if you don't remove it from `visited`?

thestr4ng3r: Good question, in this case it won't do any harm to not remove it because there is only at most one recursive call in every execution of this function.
The most "correct" way though to do it though is to delete it here so `visited` is always exactly the offsets on the stack. This becomes important if maybe later structs for example are recursed into here too since structs can have multiple members.
I could remove this delete call and add a comment instead. What do you think?

ret2libc: Is it right to re-visit something that was already visited, even if not in the currently call stack?

thestr4ng3r: Yes. For example from the following struct:
```c
struct {
    A *member_a;
    A *member_b;
};
```
If you would not remove from the visited set every time you return from the function, as soon as you arrive at `member_b`, `A` would for example already be in the visited list from `member_a` and you would incorrectly detect an infinite recursion even though everything is fine.
But as said, such cases currently can't happen so it does not matter whether we delete or not.

thestr4ng3r: I think we should leave the `set_u_delete()` line in. It should become necessary as soon as we fully parse function types.

librz/analysis/dwarf_process.c
```diff
@@ -167,12 +167,13 @@ static char *get_die_name(const RzBinDwarfDie *die) {
...
 	st32 name_attr_idx = find_attr_idx(die, DW_AT_name);
 
-	if (name_attr_idx != -1 && die->attr_values[name_attr_idx].string.content) {
-		name = strdup(die->attr_values[name_attr_idx].string.content);
-	} else {
-		name = create_type_name_from_offset(die->offset);
+	if (name_attr_idx != -1) {
+		const char *s = rz_bin_dwarf_attr_value_get_string_content(&die->attr_values[name_attr_idx]);
+		if (s) {
+			name = strdup(s);
```
wargio: optional: please use `RZ_STR_DUP`

wargio (approved):

wargio (approved):

PR #962, opened 2021-04-03
Refactor VA Calculation, binsyms and Import Resolving

thestr4ng3r:
**DO NOT SQUASH**

**Detailed description**

The commits can be looked at separately.
* The first unclutters some of the of the virtual address calculation and it is now centered around the `RzBinObject`-local `rz_bin_object_get_vaddr()` function.
* The second converts binsym to use an actual enum type `RzBinSpecialSymbol` everywhere instead of int and also makes it `RzBinObject`-local.
* The third just removes `has_lit`
* The fourth removes an old hack where **GLOBAL** (!!!) sdb of "symbol name -> addr" was used during iterating over imports to resolve their addresses from their name. The same info is now available in a smaller hashtable in `RzBinObject`.

**Test plan**

This is quite well-covered by integration tests

### DIFF
--- a/librz/bin/bobj.c
+++ b/librz/bin/bobj.c
@@ -45,6 +45,7 @@ static void object_delete_items(RzBinObject *o) {
 	rz_list_free(o->sections);
 	rz_list_free(o->strings);
 	ht_up_free(o->strings_db);
+	ht_pp_free(o->import_name_symbols);
 	rz_list_free(o->symbols);
 	rz_list_free(o->classes);
 	ht_pp_free(o->classes_ht);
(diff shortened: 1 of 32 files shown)


librz/bin/bobj.c
```diff
@@ -449,6 +461,14 @@ RZ_API RBNode *rz_bin_object_patch_relocs(RzBinFile *bf, RzBinObject *o) {
 	return o->relocs;
 }
 
+RZ_API RzBinSymbol *rz_bin_object_get_symbol_of_import(RzBinObject *o, RzBinImport *imp) {
```
XVilka, line 468: Please add Doxygen too.

librz/bin/bobj.c
```diff
@@ -517,6 +537,48 @@ RZ_IPI void rz_bin_object_filter_strings(RzBinObject *bo) {
...
+ */
+RZ_API ut64 rz_bin_object_get_vaddr(RzBinObject *o, ut64 paddr, ut64 vaddr) {
+	rz_return_val_if_fail(o, UT64_MAX);
+
+	if (paddr == UT64_MAX) {
+		// everything we have is the vaddr
+		return vaddr;
+	}
+
+	/* hack to realign thumb symbols */
```
XVilka, line 563: I suggest to extract this Thumb hack into the separate function.

librz/core/cbin.c
```diff
@@ -2116,114 +2121,18 @@ static int bin_relocs(RzCore *r, PJ *pj, int mode, int va) {
 	return relocs_count >= 0;
 }
 
-#define MYDB 1
-/* this is a VERY VERY VERY hacky and bad workaround that needs proper refactoring in Rbin to use Sdb */
```
XVilka, line 2120: OMG WTF.

XVilka (changes requested):
Only minor nitpicks, refactoring looks good 👍

librz/core/cbin.c
thestr4ng3r: Yes.

librz/bin/bobj.c
thestr4ng3r: added.

librz/bin/bobj.c
thestr4ng3r: To be honest, I would rather not. Since this is quite a dirty hack, the goal should be to eventually remove it from there completely (not in this pr since it just moves code). By putting it in a separate functions we would obfuscate a bit how dirty it is when looking at the `rz_bin_object_get_vaddr()` function.

XVilka (approved):

PR #964, opened 2021-04-03
Separate Imports, Symbols, Classes and Resources Setting and Printing

thestr4ng3r:
**DO NOT SQUASH**

Follow-up to all the other recent bin prs. This shows how little is actually done when "applying" imports. Maybe we can find a better way to achieve the same thing later.

Before I could refactor the symbols I had to make some adjustments to `dmias` which can be used in debug mode to load the bin info of a shared library in the process. It should use the new applying functions, but not apply everything and not let it set the `bin->cur`.


test/db/cmd/cmd_i
```diff
@@ -1040,114 +1040,114 @@ CMDS=ic*
...
-"f method._JNIEnv.NewObject(_jclass*,__jmethodID*,_...) = 0x6074"
-"f method._JNIEnv.GetObjectClass(_jobject*) = 0x608e"
-"f method._JNIEnv.GetMethodID(_jclass*,_char_const*,_char_const*) = 0x6098"
-"f method._JNIEnv.CallObjectMethod(_jobject*,__jmethodID*,_...) = 0x60a4"
-"f method._JNIEnv.CallVoidMethod(_jobject*,__jmethodID*,_...) = 0x60c0"
-"f method._JNIEnv.GetStaticMethodID(_jclass*,_char_const*,_char_const*) = 0x60dc"
-"f method._JNIEnv.CallStaticObjectMethod(_jclass*,__jmethodID*,_...) = 0x60ea"
-"f method._JNIEnv.CallStaticIntMethod(_jclass*,__jmethodID*,_...) = 0x695e"
-"f method._JNIEnv.CallStaticVoidMethod(_jclass*,__jmethodID*,_...) = 0x697c"
+"f method._JNIEnv.FindClass_char_const = 0x6060"
```
thestr4ng3r, line 1043: These differences are because now `ic*` and the setting code for method-flags share the same code to build the flag name. Before there were some discrepancies with how the names were sanitized, so this is more correct.

librz/core/cbin.c
```diff
@@ -1049,6 +1075,363 @@ RZ_API bool rz_core_bin_apply_relocs(RzCore *core, RzBinFile *binfile, bool va_b
...
+		return sym->is_imported ? "sym.imp" : "sym";
+	}
+	return "sym";
+}
+
+#define MAXFLAG_LEN_DEFAULT 128
+
+static char *construct_symbol_flagname(const char *pfx, const char *libname, const char *symname, int len) {
+	char *r = rz_str_newf("%s.%s%s%s", pfx, libname ? libname : "", libname ? "_" : "", symname);
+	if (r) {
```
XVilka: reverse the condition please

XVilka (changes requested):

librz/core/cbin.c
thestr4ng3r: done

librz/core/cbin.c
```diff
@@ -327,14 +323,64 @@ RZ_API int rz_core_bin_set_by_name(RzCore *core, const char *name) {
...
+	}
+	if (mask & RZ_CORE_BIN_ACC_CLASSES) {
+		rz_core_bin_apply_classes(r, binfile);
+	}
+	if (mask & RZ_CORE_BIN_ACC_RESOURCES) {
+		rz_core_bin_apply_resources(r, binfile);
+	}
+	// ----
+
+	return true;
```
wargio, line 370: should we have a return value for failure in any of those methods called?

librz/core/cbin.c
```diff
@@ -327,14 +323,64 @@ RZ_API int rz_core_bin_set_by_name(RzCore *core, const char *name) {
 	return false;
 }
 
-RZ_API int rz_core_bin_apply_all_info(RzCore *r, RzBinFile *binfile) {
+RZ_API bool rz_core_bin_apply_info(RzCore *r, RzBinFile *binfile, int mask) {
+	rz_return_val_if_fail(r && binfile, false);
```
wargio: verify also that mask is non zero.

librz/core/cbin.c
```diff
@@ -327,14 +323,64 @@ RZ_API int rz_core_bin_set_by_name(RzCore *core, const char *name) {
 	return false;
 }
 
-RZ_API int rz_core_bin_apply_all_info(RzCore *r, RzBinFile *binfile) {
+RZ_API bool rz_core_bin_apply_info(RzCore *r, RzBinFile *binfile, int mask) {
```
wargio: i do not see any point in having mask as int. i would prefer to have it as `ut32` so we are sure it will go from 0 to UT32_MAX

librz/core/cbin.c
```diff
@@ -1043,10 +1069,367 @@ RZ_API bool rz_core_bin_apply_relocs(RzCore *core, RzBinFile *binfile, bool va_b
...
+	sn->demflag = NULL;
+	if (bin_demangle && sym->paddr) {
+		sn->demname = rz_bin_demangle(r->bin->cur, lang, sn->name, sym->vaddr, keep_lib);
+		if (sn->demname) {
+			sn->demflag = construct_symbol_flagname(pfx, sym->libname, sn->demname, -1);
+		}
+	}
+}
+
+static void snFini(SymName *sn) {
```
wargio: please rename this as `symname_free`

wargio (changes requested):

librz/core/cbin.c
thestr4ng3r: Imo, a failure in one of these sub-functions should not really mean a failure of the whole function since other info would still be loaded successfully.

wargio: agreed

librz/core/cbin.c
```diff
@@ -327,14 +323,60 @@ RZ_API int rz_core_bin_set_by_name(RzCore *core, const char *name) {
...
+	if (mask & RZ_CORE_BIN_ACC_DWARF) {
+		rz_core_bin_apply_dwarf(r, binfile);
+	}
+	if (mask & RZ_CORE_BIN_ACC_ENTRIES) {
+		rz_core_bin_apply_entry(r, binfile, va);
+	}
+	if (mask & RZ_CORE_BIN_ACC_SECTIONS) {
+		rz_core_bin_apply_sections(r, binfile, va);
+	}
+	if (mask & RZ_CORE_BIN_ACC_RELOCS && rz_config_get_i(r->config, "bin.relocs")) {
```
wargio: isn't `bin.relocs` a boolean ? if yes, then we should use `rz_config_get_b`

librz/core/cbin.c
```diff
@@ -1043,10 +1065,367 @@ RZ_API bool rz_core_bin_apply_relocs(RzCore *core, RzBinFile *binfile, bool va_b
...
+		if (!sym) {
+			continue;
+		}
+		ut64 addr = rva(o, sym->paddr, sym->vaddr, va ? VA_TRUE : VA_FALSE);
+		rz_meta_set(core->analysis, RZ_META_TYPE_DATA, addr, cdsz, NULL);
+	}
+	return true;
+}
+
+static const char *getPrefixFor(RzBinSymbol *sym) {
```
wargio: `getPrefixFor` -> `get_prefix_for`

librz/core/cbin.c
```diff
@@ -1043,10 +1065,367 @@ RZ_API bool rz_core_bin_apply_relocs(RzCore *core, RzBinFile *binfile, bool va_b
...
+RZ_API bool rz_core_bin_apply_symbols(RzCore *core, RzBinFile *binfile, bool va) {
+	rz_return_val_if_fail(core && binfile, NULL);
+	RzBinObject *o = binfile->o;
+	RzBinInfo *info = o ? o->info : NULL;
+	if (!info) {
+		return false;
+	}
+
+	bool is_arm = info && info->arch && !strncmp(info->arch, "arm", 3);
+	bool bin_demangle = rz_config_get_i(core->config, "bin.demangle");
```
wargio: `rz_config_get_b`

librz/core/cbin.c
```diff
@@ -1043,10 +1065,367 @@ RZ_API bool rz_core_bin_apply_relocs(RzCore *core, RzBinFile *binfile, bool va_b
...
+}
+
+RZ_API bool rz_core_bin_apply_classes(RzCore *core, RzBinFile *binfile) {
+	rz_return_val_if_fail(core && binfile, false);
+	RzBinObject *o = binfile->o;
+	RzList *cs = o ? o->classes : NULL;
+	if (!cs) {
+		return false;
+	}
+	if (!rz_config_get_i(core->config, "bin.classes")) {
```
wargio: `rz_config_get_b`

wargio (changes requested):

wargio (approved):

librz/core/cbin.c
```diff
@@ -1043,10 +1065,367 @@ RZ_API bool rz_core_bin_apply_relocs(RzCore *core, RzBinFile *binfile, bool va_b
...
+				if (core->bin->prefix) {
+					char *prname = rz_str_newf("%s.%s", core->bin->prefix, sn.methflag);
+					rz_name_filter(sn.methflag, -1, true);
+					free(sn.methflag);
+					sn.methflag = prname;
+				}
+				if (fi) {
+					rz_flag_item_set_realname(fi, sn.methname);
+					if ((fi->offset - core->flags->base) == addr) {
+						//		char *comment = fi->comment ? strdup (fi->comment) : NULL;
```
XVilka: Is this comment necessary? Maybe better remove it?

XVilka (approved):

librz/core/cbin.c
thestr4ng3r: comments are never necessary, removing.

PR #969, opened 2021-04-05
Refactor ELF PT_NOTE Parsing

thestr4ng3r:

**Detailed description**

This is about parsing `PT_NOTE` segment contents from core files (see for example https://www.gabriel.urdhr.fr/2015/05/29/core-file/)
Before, both `rz_bin_elf_grab_regstate()` and `get_nt_file_maps()`, would parse the segment on the fly (with a lot of code duplication) and return only the necessary information.
I will need to extract more fine-grained info from the regstate later so to avoid parsing it again, I refactored it to be parsed in the initial loading such that it is always available later.

**Test plan**

see the added test

### DIFF
--- a/librz/bin/format/elf/elf.c
+++ b/librz/bin/format/elf/elf.c
@@ -109,17 +109,25 @@ enum {
 	ARCH_LEN
 };
 
+/// Offset and length of the regstate inside a NT_PRSTATUS note for core files of a certain architecture
 typedef struct reginfo {
 	ut32 regsize;
+
+	/**
+	 * This delta is the offset into the actual data of an NT_PRSTATUS note
+	 * where the regstate of size regsize lies.
+	 * That is, it is the offset after the Elf_(Nhdr) and the variable-length string + optional padding
+	 * have already been skipped.
+	 */
 	ut32 regdelta;
 } reginfo_t;
 
 static reginfo_t reginf[ARCH_LEN] = {
-	{ 160, 0x5c },
-	{ 216, 0x84 },
-	{ 72, 0x5c },
-	{ 272, 0x84 },
-	{ 272, 0x84 }
+	{ 160, 0x48 },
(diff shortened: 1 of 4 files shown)


librz/bin/format/elf/elf.c
```diff
@@ -115,11 +115,11 @@ typedef struct reginfo {
 } reginfo_t;
 
 static reginfo_t reginf[ARCH_LEN] = {
-	{ 160, 0x5c },
-	{ 216, 0x84 },
-	{ 72, 0x5c },
-	{ 272, 0x84 },
-	{ 272, 0x84 }
+	{ 160, 0x48 },
```
XVilka, line 126: Where from this change comes?

thestr4ng3r: Before, these offsets were from the very start of the note, now they are from the start of the inner payload since that is where `parse_note_prstatus()` starts to read and it also makes more sense.

More specifically, these notes are:
3x32bit + a string of size defined in one of these 32bit values + optional padding + payload.
In the case of `NT_PRSTATUS` ones, the string is `CORE\0` (usually, not sure if always guaranteed to be this. The previous implementation relied on this length to be always the same, now it does not anymore), so you get
3*4 + 5 + 3 padding = 0x14 which is exactly the difference here.

XVilka: Could you please add this as a comment?

thestr4ng3r: This by itself doesn't make much sense as a comment because it primarily explains why it's different to before, but I will add something about the meaning of these values.

XVilka: Yes, that's what I meant.

thestr4ng3r: added

XVilka (approved):

PR #991, opened 2021-04-10
Coverity java

wargio:
# DO NOT SQUASH ME

**Detailed description**

...

**Test plan**

...

**Closing issues**

...


XVilka (approved):

librz/bin/format/java/class_attribute.c
```diff
@@ -5,6 +5,10 @@
 
 static ut8 *copy_buffer(RzBuffer *buf, st64 size, bool null_terminator) {
 	ut8 *buffer;
+	st64 buffer_size = rz_buf_size(buf);
```
ret2libc, line 8: I was just reading this code for the first time, so might be a bit unrelated to this PR but i'm writing this anyway, since you wrote this code.

I think it is a bit risky to use `RzBuf` API in this way because it partially defeats its purpose, which is to never have to load the entire buffer in memory. This function can be used to do that though, because the `size` argument comes from the `attribute_length`, which is read as-is from the file, so an attacker can put any value there and force you to allocate ~4GB for each attribute.

Is it really necessary to have these buffer allocated? I see this is used in `ATTRIBUTE_TYPE_SOURCEDEBUGEXTENSION`, `ATTRIBUTE_TYPE_UNKNOWN`, `ATTRIBUTE_TYPE_CODE` but from a quick look i couldn't find where these buffers are used. You probably know better.

Another thing: what's the purpose of the `null_terminator` var? Can't we just always allocate it? Just asking.

librz/bin/format/java/class_bin.c
```diff
@@ -713,10 +714,15 @@ RZ_API void rz_bin_java_class_as_source_code(RzBinJavaClass *bin, RzStrBuf *sb)
...
 
 					char *ptr = strchr(dem, '(');
-					*(ptr - 1) = 0;
-					rz_strbuf_append(sb, dem);
-					rz_strbuf_appendf(sb, "%s", tmp);
-					rz_strbuf_append(sb, ptr);
+					if (ptr) {
+						*(ptr - 1) = 0;
+						rz_strbuf_append(sb, dem);
+						rz_strbuf_appendf(sb, "%s", tmp);
```
ret2libc:
```suggestion
						rz_strbuf_append(sb, tmp);
```

librz/bin/format/java/class_bin.c
```diff
@@ -713,10 +714,15 @@ RZ_API void rz_bin_java_class_as_source_code(RzBinJavaClass *bin, RzStrBuf *sb)
...
+						rz_strbuf_append(sb, ptr);
+					} else {
+						rz_strbuf_append(sb, dem);
+						rz_strbuf_appendf(sb, "%s", tmp);
```
ret2libc:
```suggestion
						rz_strbuf_append(sb, tmp);
```

librz/bin/format/java/class_bin.c
```diff
@@ -713,10 +714,15 @@ RZ_API void rz_bin_java_class_as_source_code(RzBinJavaClass *bin, RzStrBuf *sb)
 					rz_str_replace_char(tmp, '/', '.');
 
 					char *ptr = strchr(dem, '(');
-					*(ptr - 1) = 0;
-					rz_strbuf_append(sb, dem);
-					rz_strbuf_appendf(sb, "%s", tmp);
-					rz_strbuf_append(sb, ptr);
+					if (ptr) {
+						*(ptr - 1) = 0;
+						rz_strbuf_append(sb, dem);
```
ret2libc, line 718: just do dem+tmp before the `if` and then add the `rz_strbuf_append(sb, ptr);` if ptr is not null.

librz/bin/format/java/class_bin.c
```diff
@@ -42,7 +42,8 @@ static char *java_class_constant_pool_stringify_at(RzBinJavaClass *bin, ut32 ind
 }
 
 static ut32 sanitize_size(st64 buffer_size, ut32 count, ut32 min_struct_size) {
-	ut64 memory_size = count * min_struct_size;
+	ut64 memory_size = count;
+	memory_size *= min_struct_size;
```
ret2libc: why this?

librz/bin/format/java/class_attribute.c
wargio: it really does allocate on file read? from the source code it doesn't seems to.

These extensions are used `ATTRIBUTE_TYPE_CODE` for the actual code, `ATTRIBUTE_TYPE_SOURCEDEBUGEXTENSION` for when you display all the details of the class.
`attribute_length` is always ut16, so whatever you place there is not that big on memory.

librz/bin/format/java/class_bin.c
wargio: No. i'm modifying some data. i can't. already tried and it failed. this is why i had to force push

librz/bin/format/java/class_bin.c
wargio: it's written in the vulnerability `Integer overflow` i'm multipying a ut32 with a  ut32 which can overflow. by copying the value i can make sure it won't

librz/bin/format/java/class_attribute.c
ret2libc: `ut32 attribute_length;` and `attribute_length = rz_buf_read_be32(buf);` so I think it is actually 32bit, not 16.

> it really does allocate on file read? from the source code it doesn't seems to.

what do you mean? Here you are doing a `malloc(size)` where `size` is attacker-controlled so an attacker can essentially make this code allocate as much as he wants (up to almost 4GB).

> These extensions are used ATTRIBUTE_TYPE_CODE for the actual code, ATTRIBUTE_TYPE_SOURCEDEBUGEXTENSION for when you display all the details of the class.

Where exactly? I could not find them. As said, I'm not familiar enough with the code, so they might be used somewhere.

librz/bin/format/java/class_bin.c
ret2libc: Ok got it on mattermost chat.

I suggest `ut64 memory_size = (ut64)count * min_struct_size` to make it clearer that the cast to ut64 is important. Otherwise it's easily overseen.

librz/bin/format/java/class_attribute.c
wargio: oh, sorry i got confused on the 32bit, but anyway it's impossible to allocate 4GB if i check it against the file size. the max size for a class file (including everything) is 2GB.

wargio: let's say that you even allocate 4GB (which is impossible with those changes), `if (!buffer || rz_buf_read(buf, buffer, size) < (size - 1)) {` will make sure to actually read 4GB. if it reads less it will drop the read and just say it failed to read the data

wargio: about the `null_terminator` is because `ATTRIBUTE_TYPE_SOURCEDEBUGEXTENSION` is a string that doesn't end with a zero terminator

wargio: ok, i forgot to add `ATTRIBUTE_TYPE_SOURCEDEBUGEXTENSION` into the strings output. this is why you don't find a xref on that

wargio:
the fail is due a 403

librz/bin/format/java/class_attribute.c
ret2libc: Shall we change:
```
if (!buffer || rz_buf_read(buf, buffer, size) < (size - 1)) {
```
to
```
if (!buffer || rz_buf_read(buf, buffer, size) != size) {
```
?


Anyway, my concerns as said were rather generic more than related to this specific PR. In general, it would be better if you avoid allocating all the memory in one shot but just keep using the RzBuf API.

Moreover, please double check that this buffer is really used and that `null_terminator` is really needed...

> about the null_terminator is because ATTRIBUTE_TYPE_SOURCEDEBUGEXTENSION is a string that doesn't end with a zero terminator

I see:
```
	attr->type = ATTRIBUTE_TYPE_SOURCEDEBUGEXTENSION;
	if (attr->attribute_length < 1) {
		return true;
	}
	st64 size = (st64)attr->attribute_length;
	attr->info = copy_buffer(buf, size, true);
```
so it seems like ATTRIBUTE_TYPE_SOURCEDEBUGEXTENSION wants the zero terminator.

wargio: yes it is needed the null terminator, and no, i can't change that size, because the last block of the class is a terminator that we do not have on a file, but it is added by the JVM which makes the size-1

ret2libc: Just always add the null-terminator, I think there is no real need to put it only sometimes... Always having one byte more probably won't hurt anyway.

wargio: ok. i'll patch this in another PR.

Issue #238, opened 2020-12-21
afls command family not working and aren't tested


Maijin:
### Work environment

n/a

### Actual behavior

| afls   same as aflsa
| aflsa  sort by address (same as afls)
| aflss  sort by size
| aflsn  sort by name
| aflsb  sort by number of basic blocks

Aren't working and aren't tested

### Expected behavior

| afls   same as aflsa
| aflsa  sort by address (same as afls)
| aflss  sort by size
| aflsn  sort by name
| aflsb  sort by number of basic blocks

Are working and tested or killed

XVilka:
In theory it should be subsumed/merged into `aflt` since the Table API is more suitable for sorting by the column.

Maijin:
Agree

fangxlmr:
@ReJimp  This issue is asigned to you as RSoC micro-task, so you could work on this from now on. Feel free to ask questions and send Pull Requests.

In case for anyone who is not aware about [RSoC](https://keenlab.tencent.com/zh/2021/03/04/RSoC-keenlab-2021/).

CC @XVilka

salutdamour:
I have found the targeted function rz_list_sort. I feel awkward to ask a simple question. Before I test and repair the afls command, shall I have to build the project? I have also installed the rizin by a easy way in my PC.

XVilka:
@ReJimp of course, see these documents for building and contributing:
- [BUILDING.md](https://github.com/rizinorg/rizin/blob/dev/BUILDING.md)
- [CONTRIBUTING.md](https://github.com/rizinorg/rizin/blob/dev/CONTRIBUTING.md)
- [DEVELOPERS.md](https://github.com/rizinorg/rizin/blob/dev/DEVELOPERS.md)

salutdamour:
@XVilka  Oh my god! I found I forgot to use git-clang-format 11 to format my code. It made PR run failed. What should I do to solve the problem.

ret2libc:
@pelijah did you remove the `afls` commands?

p371jah:
Yep, `afls` commands were removed since we have `aflt`.

Issue #613, opened 2021-02-12
Add ** (exponentiation/power) to ? math operations


caribpa:
**Is your feature request related to a problem? Please describe.**
The built-in math/logical calculation/conversion command `? <expression>` is quite handy. Unfortunately, in several occasions when calculating an offset, the fun is interrupted due to the missing exponentiation/power capabilities of `?`.

**Describe the solution you'd like**
Because `^` performs XOR, `**` can be used as symbol for exponentiation/power as it is what it's used in built-in bash and zsh math capabilities for this same purpose (zsh also accepts negative exponents):
```sh
$ rz --
> ?vi 3**6
# 729
> ? 2**-1 ~fvalue[1]
# 0.5
```

**Describe alternatives you've considered**
Currently bitwise left shift could be used with 2, but no shortcuts (other than multiplying the number, or parentheses operations, with itself multiple times) exist for the other numbers/parentheses operations 💔 

**Additional context**
🤷

majodu:
If possible could I take this issue?

ret2libc:
> If possible could I take this issue?

Sure! You can work on whatever you prefer ;)

ret2libc:
This issue is still clearly described, so I'm removing the stale flag in case some newcomers want to give it a shot since it should be quite easy to implement. Low prio however.

DhruvaG2000:
> Because ^ performs XOR, ** can be used as symbol for exponentiation/power

The challenge in using ``**`` as the new operator would be that the way operations are implemented currently is using a char input to a switch statement. If there's an easy work around for this please do let me know.

wargio:
you can switch to `*` and then check if the next char is another `*`

ret2libc:
> > Because ^ performs XOR, ** can be used as symbol for exponentiation/power
> 
> The challenge in using `**` as the new operator would be that the way operations are implemented currently is using a char input to a switch statement. If there's an easy work around for this please do let me know.

See for example how `RNCROR` `RNCSHR` and `RNCGT` are handled.

DhruvaG2000:
> See for example how RNCROR RNCSHR and RNCGT are handled.

Thanks @ret2libc this pointer helped alot. now `**` works 😄

PR #993, opened 2021-04-11
Fix afls command family

salutdamour:

**Detailed description**

issue #238 

| afls same as aflsa
| aflsa sort by address (same as afls)
| aflss sort by size
| aflsn sort by name
| aflsb sort by number of basic blocks

**Test plan**

Used the `bins/elf/ls` to test.

**Closing issues**

...


salutdamour:
Sorry. Still failed on clang-format. I have changed cmd_analysis.c on Ubuntu18, so I use the command `clang-format -style=file -i ~/rizin/librz/core/cmd_analysis.c` to format my code. However, it takes more problems compared with the last time.

wargio:
just run `./sys/clang-format.py` from the root folder of rizin

salutdamour:
> just run `./sys/clang-format.py` from the root folder of rizin

I only modified a C file.When I ran the clang-format.py, it changed 34 files. Is it normal?  I am afraid that I can’t run the clang-format test of PR.

XVilka:
@ReJimp it might be because you have too old clang version.

salutdamour:
@XVilka I think the file `clang-format.py` needs to be updated. I installed `clang-format-11` on Ubuntu20. When I run the script, it pointed out that I didn't install clang-format. When I used `sudo apt search clang-format
`, it showed that `clang-format` 's highest version is 10. `clang-format` on the 76th line and 78th line in  the `clang-format.py` replaced by `clang-format-11` is better.

wargio:
i suggest to add a symlink to `clang-format` since ubuntu calls it `clang-format-11`.
that said my clang-format version `clang-format version 11.1.0` does not produce that result.

wargio:
be sure to commit and push the changes.

librz/core/cmd_analysis.c
```diff
@@ -757,16 +757,16 @@ static int cmpname(const void *_a, const void *_b) {
 static int cmpsize(const void *a, const void *b) {
 	ut64 sa = (int)rz_analysis_function_linear_size((RzAnalysisFunction *)a);
 	ut64 sb = (int)rz_analysis_function_linear_size((RzAnalysisFunction *)b);
-	return (sa > sb) ? -1 : (sa < sb) ? 1
-					  : 0;
+	return (sa > sb) ? 1 : (sa < sb) ? 1
```
XVilka: I am not sure this change is right. Why it does return `1` in both cases? This is not how comparison callback works...

XVilka (changes requested):

XVilka:
Could you please add tests for this?

salutdamour:
Does it mean that show the test results or write a test in test/？What's more, I am also interested in the issue #613. Can I try solving it at the same time? If without your help, I may waste more time on the envrionment. Just when I felt I was on track, it was time for the RSOC review.

XVilka:
@ReJimp Yes, please add a few tests in `test/db/cmd/cmd_afl`.

XVilka:
@ReJimp sure, you can do that, take https://github.com/rizinorg/rizin/issues/613

librz/core/cmd_analysis.c
```diff
@@ -2908,22 +2908,27 @@ RZ_IPI int rz_cmd_analysis_fcn(void *data, const char *input) {
 			case 'a': // "aflsa"
 				core->analysis->fcns->sorted = false;
 				rz_list_sort(core->analysis->fcns, cmpaddr);
+				rz_core_analysis_fcn_list(core, NULL, "o");
```
ret2libc, line 2911: why `"o"`?

salutdamour: I think it can jump to the default branch in `rz_core_analysis_fcn_list` as same as the command `afl`.

ret2libc: Ok, good enough for now! That `o` does not have any particular meaning, but considering it's already used elsewhere it is fine for now, thanks!

ret2libc (commented):
@XVilka @thestr4ng3r @Maijin `afls` commands right now change the ordering of the function list, so if you do `aflsa`, then `afl` will show the new ordering. With this PR we are slightly changing the behaviour of these `afls` commands to also print the functions list immediately. Do we want to keep this side-effect behaviour?

What I mean is that if you do:
```
> afl
> aflsn
> afl
```
the two `afl` commands will show commands in a different order, just because `aflsn` was called in-between. WDYT? Maybe `aflsn` should not affect the ordering of the global function list.

XVilka:
@ret2libc yes, agree - definitely it **should not** change the global ordering.

salutdamour:
I think so. I will follow up on this issue.

XVilka:
I am closing this since there was no progress and updates for months. Feel free to reopen if you update the PR.

PR #997, opened 2021-04-11
Remove old ELF-Core-specific RzBinMap

thestr4ng3r:
**DO NOT SQUASH**

This removes the current `RzBinMap` structure, which until now was a very ELF-core specific thing. Specifically, it represented only the binary file mapping info out of an `NT_FILE` note as described here for example: https://www.gabriel.urdhr.fr/2015/05/29/core-file/#file-association

So what core would do would be the following:
* Create IO maps based on the `RzBinSections` (which in the case of elf are the segments), naming the map the same as the section/segment
* If the plugin reports it is a core file:
  * query all `RzBinMap`, iterate over them and if there is an io map with a matching address rename it to the file path of the `RzBinMap`.
  * Check if the stack pointer register value falls into an io map and call that `[stack]`.

After this pr, I will introduce a new `RzBinMap` struct that will tell core directly what to map where and how to call it without core having to use sections and patch around stuff. This has two requirements, which are the contents of this pr:
* Remove the old `RzBinMap` to free the name
* Move the fancy core-specific map-naming logic into the elf plugin (currently this is moved into the sections callback where it adds this `map_name` member to `RzBinSection` but this is only temporary and will be moved to a more fitting place once the new `RzBinMap` exists)

To make sure my changes don't break anything regarding core files, I have addes binaries and tests for all four supported architectures, out of which 3 were previously completely untested.


XVilka:
It is covered by `test/db/formats/elf/core` tests, right?

thestr4ng3r:
Yes, but only for Linux+x86_64. I want to add some more tests to it.

thestr4ng3r:
macOS failure is because of homebrew.

wargio (approved):

XVilka (approved):

PR #1008, opened 2021-04-13
Add RzBaseType listing API

XVilka:

**Detailed description**

- Renamed recently added functions like `rz_type_db_enums()`, `rz_type_db_structs()` into `rz_type_db_enum_names()` since they return only names of the corresponding types for listing and autocompletion purposes
- Added proper listing API: `rz_type_db_get_base_types()` and `rz_type_db_get_base_types_kind()`
- Removed broken Visual Types feature (`Vbt` command)

Removed Visual Types produced the following output which is largely useless:

image

After the complete "Types" refactoring project is finished it makes sense to implement such a navigation properly.

**Test plan**

CI is green


librz/type/base.c
```diff
@@ -277,6 +282,69 @@ RZ_API RzBaseType *rz_type_db_get_base_type(RzTypeDB *typedb, const char *name)
 	return base_type;
 }
 
+/**
+ * \brief Returns the list of all basic types of the chosen kind
+ *
+ * \param typedb Types Database instance
+ * \param kind Kind of the types to list
+ */
+RZ_API RZ_OWN RzList /* RzBaseType */ *rz_type_db_get_base_types_kind(RzTypeDB *typedb, RzBaseTypeKind kind) {
```
ret2libc: Would it make sense to have a `ANY` kind and just use one API instead of two that do the same thing?

ret2libc (approved):

librz/type/base.c
XVilka: It wouldn't make sense to add in `RzBaseTypeKind`.

XVilka: Also note, that the bodies of these functions will change when I migrate types storage to the hashtable.

thestr4ng3r: agree with @XVilka 
I would call this function `rz_type_db_get_base_types_of_kind()` though, so it's a bit more descriptive.

thestr4ng3r (commented):
overall lgtm

librz/type/base.c
XVilka: Renamed the function.

thestr4ng3r (approved):

Issue #815, opened 2021-03-12
Relocate automatically overlapping sections in object files.


XVilka:
Due to the structure of the ELF object files (`*.o`) before linking occurs, it's common to have the overlapping sections, e.g. `.text` and `.bss`. Unlike real ELF binaries, these sections can be harmlessly relocated with RzIO/RzBin to not interfere with disassembly and analysis.

```
[i] ℤ rizin unixterm.c-gcc-x64-O2.o                                    
[0x08000040]> px 10
- offset -   0 1  2 3  4 5  6 7  8 9  A B  C D  E F  0123456789ABCDEF
0x08000040  0100 0000 0100 0000 0000                 ..........
[0x08000040]> pd 10
            ;-- section..text:
            ;-- section..data:
            ;-- .text:
            ;-- .data:
            ;-- main::fileout:
            0x08000040      add   dword [rax], eax                     ; [02] -rw- section size 32 named .data
            0x08000042      add   byte [rax], al
            0x08000044      add   dword [rax], eax
            0x08000046      add   byte [rax], al
            0x08000048      add   byte [rax], al
            0x0800004a      add   byte [rax], al
            0x0800004c      add   byte [rax], al
            0x0800004e      add   byte [rax], al
            ;-- main::pfd:
            0x08000050      add   byte [rax], al                       ; main::pfd
            0x08000052      add   byte [rax], al
[0x08000040]> om=
0   0x8000000  █████―――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――0x8000040  r--   3     fmap.ehdr
1*  0x8000040  ――――████――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――0x8000060  r--   3     fmap..data
2   0x8000060  ―――――――████████████████████████████████████████████████████████████████████████████████―――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――0x8000460  rw-   4     mmap..bss
3   0x8000060  ―――――――█████――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――0x8000090  r--   3     fmap..rodata.str1.1
[...]

PR #1012, opened 2021-04-13
Add RzIOMap return to rz_io_open_at()

thestr4ng3r:

**Detailed description**

`rz_io_open_at()` opens a new `RzIODesc` (returned) and creates an `RzIOMap` at the same time. Often it is necessary to also access the `RzIOMap` afterwards and usually code would then query it by address, which works reliably because of how io priority works, but it's kind of dirty, also because the `RzIOMap` pointer is directly available at the end of the function, so I added this additional return arg.

**Test plan**

all tests pass

### DIFF
--- a/librz/core/cbin.c
+++ b/librz/core/cbin.c
@@ -653,33 +653,32 @@ static bool io_create_mem_map(RzIO *io, RzBinSection *sec, ut64 at) {
 	bool reused = false;
 	ut64 gap = sec->vsize - sec->size;
 	char *uri = rz_str_newf("null://%" PFMT64u, gap);
+	RzIOMap *iomap = NULL;
 	RzIODesc *desc = findReusableFile(io, uri, sec->perm);
 	if (desc) {
-		RzIOMap *map = rz_io_map_get(io, at);
-		if (!map && gap) {
-			rz_io_map_add_batch(io, desc->fd, desc->perm, 0LL, at, gap);
+		iomap = rz_io_map_get(io, at);
+		if (!iomap && gap) {
+			iomap = rz_io_map_add_batch(io, desc->fd, desc->perm, 0LL, at, gap);
 		}
 		reused = true;
 	}
 	if (!desc) {
-		desc = rz_io_open_at(io, uri, sec->perm, 0664, at);
+		desc = rz_io_open_at(io, uri, sec->perm, 0664, at, &iomap);
 	}
 	free(uri);
 	if (!desc) {
 		return false;
 	}
-	// this works, because new maps are always born on the top
-	RzIOMap *map = rz_io_map_get(io, at);
 	// check if the mapping failed
-	if (!map) {
+	if (!iomap) {
 		if (!reused) {
 			rz_io_desc_close(desc);
 		}
 		return false;
 	}
 	// let the section refere to the map as a memory-map
-	free(map->name);
-	map->name = sec->map_name && !sec->size ? strdup(sec->map_name) : rz_str_newf("mmap.%s", sec->name);
+	free(iomap->name);
+	iomap->name = sec->map_name && !sec->size ? strdup(sec->map_name) : rz_str_newf("mmap.%s", sec->name);
 	return true;
 }
 
--- a/librz/io/io.c
+++ b/librz/io/io.c
@@ -159,8 +159,18 @@ RZ_API RzIODesc *rz_io_open(RzIO *io, const char *uri, int perm, int mode) {
 	return desc;
 }
 
-/* opens a file and maps it to an offset specified by the "at"-parameter */
-RZ_API RzIODesc *rz_io_open_at(RzIO *io, const char *uri, int perm, int mode, ut64 at) {
+/**
+ * \brief Open a file and directly map it at the given offset
+ *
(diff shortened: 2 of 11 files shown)


librz/core/cbin.c
```diff
@@ -653,33 +653,29 @@ static bool io_create_mem_map(RzIO *io, RzBinSection *sec, ut64 at) {
...
 	ut64 gap = sec->vsize - sec->size;
 	char *uri = rz_str_newf("null://%" PFMT64u, gap);
+	RzIOMap *iomap = NULL;
 	RzIODesc *desc = findReusableFile(io, uri, sec->perm);
 	if (desc) {
-		RzIOMap *map = rz_io_map_get(io, at);
-		if (!map && gap) {
-			rz_io_map_add_batch(io, desc->fd, desc->perm, 0LL, at, gap);
-		}
+		iomap = rz_io_map_add_batch(io, desc->fd, desc->perm, 0LL, at, gap);
```
thestr4ng3r: Minor difference here, new maps are always created. It's the responsibility of the bin plugin to handle possible overlaps. Overlaps can also be desired and other io maps from completely different files can be here as well so we should not try to be smarter than we are.

librz/bin/p/bin_nes.c
```diff
@@ -101,13 +101,13 @@ static RzList *sections(RzBinFile *bf) {
 	ptr->name = strdup("ROM");
 	ptr->paddr = INES_HDR_SIZE;
 	ptr->size = ihdr.prg_page_count_16k * PRG_PAGE_SIZE;
+	bool mirror = ROM_START_ADDRESS + ptr->size <= ROM_MIRROR_ADDRESS; // not a 256bit ROM, mapper 0 mirrors the complete ROM in this case
 	ptr->vaddr = ROM_START_ADDRESS;
-	ptr->vsize = ROM_SIZE;
+	ptr->vsize = mirror ? ROM_MIRROR_ADDRESS - ROM_START_ADDRESS : ROM_SIZE; // make sure the ROM zero excess does not overlap the mirror
```
thestr4ng3r: Minor change in behavior here, for the change in `librz/core/cbin.c`
Before, the bin plugin would suggest a mapping like this:
```
 4 fd: 3 +0x00000010 0x00008000 - 0x0000bfff r-x fmap.ROM
 3 fd: 4 +0x00000000 0x0000c000 - 0x0000ffff r-x mmap.ROM
 2 fd: 3 +0x00000010 0x0000c000 - 0x0000ffff r-x fmap.ROM_MIRROR
 1 fd: 4 +0x00000000 0x00010000 - 0x00013fff r-x mmap.ROM_MIRROR
 ```
Before the `cbin.c` change, core would not map the `mmap.ROM` one, but now it does and by default it overlaps the `fmap.ROM_MIRROR` so the data there would be falsely zero. Now I made the bin plugin to not let the maps overlap.
An alternative would be to swap the priority but I think it makes more sense this way.

File: librz/io/io.c [hunk 3]
```diff
@@ -160,7 +160,7 @@ RZ_API RzIODesc *rz_io_open(RzIO *io, const char *uri, int perm, int mode) {
 }
 
 /* opens a file and maps it to an offset specified by the "at"-parameter */
```
ret2libc: Would you mind changing this with a better description of everything this function does?

librz/core/cbin.c
ret2libc: This is going to affect only the "bss" sections (or anyway, any sections where vsize > size), right? Do you have any real example where the old behaviour would fail? Just curious for future testing as well, because in theory I can see what you mean and why it could be a problem.

ret2libc (commented):
LGTM

librz/core/cbin.c
XVilka: @ret2libc see these files as an example. This is a common situation in raw object files: https://github.com/rizinorg/rizin/issues/815

thestr4ng3r: Yes, to be exact, it affects sections where `vsize > psize` in which case you get two mappings, one `fmap.<name>` (or none if `psize == 0`) and `mmap.<name>` with 0s.
The check here is for when an `mmap.` one is about to be created. Before, it was skipped if there was another map with a **lower priority** that contained the vaddr (only the **start address**) of the new mmap.

That means:
* The previous behavior would still overlap the .text section with 0s in this example, iff .bss has higher prio:
    ```
    [.bss                         ]
                     [.text                           ]
    ```

* The current behavior's overlaps are more or less unnoticeable if .text has higher prio than .bss. This is not the case for example in `eofarch.c-gcc-arm-O3.o` in the linked issue. In any case, we can not rely on this prio for object files since they are a special case anyway where there are no segments and we fall back to mapping sections, which are actually not meant to be mapped, so priority probably doesn't have much meaning there. (priority is taken from the order they appear in the binary, earlier is higher prio). So #815 is valid and should be fixed insided the elf plugin imo.

thestr4ng3r:
Actually these object files are a good example where the old behavior might still be necessary, even if it is a super-dirty hack. I will revert this change in this pr and grant it its own one so we only have clean refactoring in here.

thestr4ng3r:
All behavorial changes are removed from here and moved to #1021. Please review again.

XVilka (approved):

PR #1014, opened 2021-04-14
Fix UAFs and NULL DEREFs reported by Coverity

XVilka:

**Detailed description**

Just scoured through high-impact issues reported by Coverity.
Can be reviewed commit by commit.

*Note, that I marked UAF fixes with (UAF), OOB with (OOB) and NULL DEREFS in the commit message and they can be cherry-picked for 0.2.x patch release.*

**Test plan**

CI is green

### DIFF
--- a/librz/asm/asm.c
+++ b/librz/asm/asm.c
@@ -321,7 +321,7 @@ RZ_API bool rz_asm_use(RzAsm *a, const char *name) {
 		return true;
 	}
 	rz_list_foreach (a->plugins, iter, h) {
-		if (!strcmp(h->name, name) && h->arch) {
+		if (h->arch && h->name && !strcmp(h->name, name)) {
 			if (!a->cur || (a->cur && strcmp(a->cur->arch, h->arch))) {
 				plugin_fini(a);
 				char *rzprefix = rz_str_rz_prefix(RZ_SDB_OPCODES);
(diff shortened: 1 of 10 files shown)


librz/asm/asm.c
```diff
@@ -321,7 +321,7 @@ RZ_API bool rz_asm_use(RzAsm *a, const char *name) {
 		return true;
 	}
 	rz_list_foreach (a->plugins, iter, h) {
-		if (!strcmp(h->name, name) && h->arch) {
+		if (h && h->arch && h->name && !strcmp(h->name, name)) {
```
ret2libc:
```suggestion
		if (h->arch && h->name && !strcmp(h->name, name)) {
```
I think `h` should not be checked. If `h` is NULL there's definitely something "wronger" somewhere else.

librz/reg/profile.c
```diff
@@ -397,7 +397,6 @@ RZ_API char *rz_reg_parse_gdb_profile(const char *profile_file) {
 				str = rz_file_slurp(file, NULL);
 				free(file);
 			}
-			free(base);
```
ret2libc: ?? why?

librz/asm/asm.c
XVilka: We can add `rz_return_if_fail()` then here.

librz/reg/profile.c
XVilka: image

XVilka: Though you are probably right and it's false positive in this concrete case, I will remove this commit.

librz/asm/asm.c
ret2libc: I would just remove that completely. If you have doubts, have a look at the rz_list_append that append plugins to `a->plugins`.

XVilka: Removed NULL check for `h` here.

wargio (approved):

PR #1021, opened 2021-04-14
Remove hacky Overlap Check in Core IO Mapping from Bin

thestr4ng3r:

**Detailed description**
This affects sections where `vsize > psize` in which case you get two mappings, one `fmap.<name>` (or none if `psize == 0`) and `mmap.<name>` with 0s.
The check here is for when an `mmap.` one is about to be created. Before, it was skipped if there was another map with a **lower priority** that contained the vaddr (only the **start address**) of the new mmap, however **only** if there has been a "reusable" file found, meaning another `mmap` file of exactly the same size, which is present in `test/db/cmd/bug_3788` but extremely rare and specific, so this change should not change anything for almost all files.

That means:
* The previous behavior would still overlap the .text section with 0s in this example, iff .bss has higher prio:
    ```
    [.bss                         ]
                     [.text                           ]
    ```

* The current behavior's overlaps are more or less unnoticeable if .text has higher prio than .bss. This is not the case for example in `eofarch.c-gcc-arm-O3.o` in the linked issue. In any case, we can not rely on this prio for object files since they are a special case anyway where there are no segments and we fall back to mapping sections, which are actually not meant to be mapped, so priority probably doesn't have much meaning there. (priority is taken from the order they appear in the binary, earlier is higher prio). So #815 is valid and should be fixed insided the elf plugin imo.

**Test plan**

The only relevant case I am aware of is `test/db/cmd/bug_3788`, which fails without the changes in `bin_nes.c`.

### DIFF
--- a/librz/bin/p/bin_nes.c
+++ b/librz/bin/p/bin_nes.c
@@ -101,13 +101,13 @@ static RzList *sections(RzBinFile *bf) {
 	ptr->name = strdup("ROM");
 	ptr->paddr = INES_HDR_SIZE;
 	ptr->size = ihdr.prg_page_count_16k * PRG_PAGE_SIZE;
+	bool mirror = ROM_START_ADDRESS + ptr->size <= ROM_MIRROR_ADDRESS; // not a 256bit ROM, mapper 0 mirrors the complete ROM in this case
 	ptr->vaddr = ROM_START_ADDRESS;
-	ptr->vsize = ROM_SIZE;
+	ptr->vsize = mirror ? ROM_MIRROR_ADDRESS - ROM_START_ADDRESS : ROM_SIZE; // make sure the ROM zero excess does not overlap the mirror
 	ptr->perm = RZ_PERM_RX;
 	ptr->add = true;
 	rz_list_append(ret, ptr);
-	if (ROM_START_ADDRESS + ptr->size <= ROM_MIRROR_ADDRESS) {
-		// not a 256bit ROM, mapper 0 mirrors the complete ROM in this case
+	if (mirror) {
 		if (!(ptr = RZ_NEW0(RzBinSection))) {
 			return ret;
 		}
--- a/librz/core/cbin.c
+++ b/librz/core/cbin.c
@@ -656,10 +656,7 @@ static bool io_create_mem_map(RzIO *io, RzBinSection *sec, ut64 at) {
 	RzIOMap *iomap = NULL;
 	RzIODesc *desc = findReusableFile(io, uri, sec->perm);
 	if (desc) {
-		iomap = rz_io_map_get(io, at);
-		if (!iomap && gap) {
-			iomap = rz_io_map_add_batch(io, desc->fd, desc->perm, 0LL, at, gap);
-		}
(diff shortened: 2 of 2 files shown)


librz/bin/p/bin_nes.c
```diff
@@ -101,13 +101,13 @@ static RzList *sections(RzBinFile *bf) {
 	ptr->name = strdup("ROM");
 	ptr->paddr = INES_HDR_SIZE;
 	ptr->size = ihdr.prg_page_count_16k * PRG_PAGE_SIZE;
+	bool mirror = ROM_START_ADDRESS + ptr->size <= ROM_MIRROR_ADDRESS; // not a 256bit ROM, mapper 0 mirrors the complete ROM in this case
 	ptr->vaddr = ROM_START_ADDRESS;
-	ptr->vsize = ROM_SIZE;
+	ptr->vsize = mirror ? ROM_MIRROR_ADDRESS - ROM_START_ADDRESS : ROM_SIZE; // make sure the ROM zero excess does not overlap the mirror
```
thestr4ng3r, line 106: Minor change in behavior here, for the change in `librz/core/cbin.c`
Before, the bin plugin would suggest a mapping like this:
```
 4 fd: 3 +0x00000010 0x00008000 - 0x0000bfff r-x fmap.ROM
 3 fd: 4 +0x00000000 0x0000c000 - 0x0000ffff r-x mmap.ROM
 2 fd: 3 +0x00000010 0x0000c000 - 0x0000ffff r-x fmap.ROM_MIRROR
 1 fd: 4 +0x00000000 0x00010000 - 0x00013fff r-x mmap.ROM_MIRROR
 ```
Before the `cbin.c` change, core would not map the `mmap.ROM` one, but now it does and by default it overlaps the `fmap.ROM_MIRROR` so the data there would be falsely zero. Now I made the bin plugin to not let the maps overlap.
An alternative would be to swap the priority but I think it makes more sense this way.

File: librz/core/cbin.c [hunk 2]
```diff
@@ -656,10 +656,7 @@ static bool io_create_mem_map(RzIO *io, RzBinSection *sec, ut64 at) {
 	RzIOMap *iomap = NULL;
 	RzIODesc *desc = findReusableFile(io, uri, sec->perm);
 	if (desc) {
-		iomap = rz_io_map_get(io, at);
-		if (!iomap && gap) {
```
thestr4ng3r, line 660: removing this check is the significant change here.

wargio (approved):

XVilka (approved):

ret2libc:
> . (priority is taken from the order they appear in the binary, earlier is higher prio).

Is this right btw? I think the loader traverse sections/segments in the order they appear in the binary, so it potentially loads last  segments/sections on top of previously loaded one (thus, later segments have higher prio than first ones).

thestr4ng3r:
> > . (priority is taken from the order they appear in the binary, earlier is higher prio).
> 
> Is this right btw? I think the loader traverse sections/segments in the order they appear in the binary, so it potentially loads last segments/sections on top of previously loaded one (thus, later segments have higher prio than first ones).

I'm not sure exactly but in any case this is format-specific, so if a plugin needs it in one way or the other, it's its responsibility to serve the info in that way. Though from how the code looks it seems like this behavior was done on purpose.

PR #1045, opened 2021-04-18
Fix #1032 - Wrong use of wcstombs

### LINKED ISSUE #1032: Wrong use of `sz = wcstombs(NULL, buf, len);` in `librz/util/str.c`
### Work environment

| OS/arch/bits (mandatory)                             | Fedora 33 x86_64 
| `rizin -v` full output, **not truncated** (mandatory)  | https://github.com/rizinorg/rizin/commit/41192158ecf168eb18e831d225cdfb02829a40b9

### Expected behavior

No warnings

### Actual behavior

```
../librz/util/str.c: In function ‘rz_str_wc_to_mb_l’:
../librz/util/str.c:3519:7: warning: argument 1 is null but the corresponding size argument 3 range is [1, 2147483647] [-Wnonnull]
 3519 |  sz = wcstombs(NULL, buf, len);
      |       ^~~~~~~~~~~~~~~~~~~~~~~~
In file included from ../librz/include/rz_types.h:250,
                 from ../librz/util/str.c:4:
/usr/include/stdlib.h:937:15: note: in a call to function ‘wcstombs’ declared with attribute ‘write_only (1, 3)’
  937 | extern size_t wcstombs (char *__restrict __s,
      |               ^~~~~~~~
```

This call was introduced in https://github.com/rizinorg/rizin/commit/527ff71d4184a07d724f9a7c3a464a3b228ea408

wargio:

**Detailed description**

Tries to fix:
```
../librz/util/str.c: In function ‘rz_str_wc_to_mb_l’:
../librz/util/str.c:3519:7: warning: argument 1 is null but the corresponding size argument 3 range is [1, 2147483647] [-Wnonnull]
 3519 |  sz = wcstombs(NULL, buf, len);
      |       ^~~~~~~~~~~~~~~~~~~~~~~~
In file included from ../librz/include/rz_types.h:250,
                 from ../librz/util/str.c:4:
/usr/include/stdlib.h:937:15: note: in a call to function ‘wcstombs’ declared with attribute ‘write_only (1, 3)’
  937 | extern size_t wcstombs (char *__restrict __s,
      |               ^~~~~~~~

```

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/librz/util/str.c
+++ b/librz/util/str.c
@@ -3510,17 +3510,16 @@ err_r_str_mb_to_wc:
 
 RZ_API char *rz_str_wc_to_mb_l(const wchar_t *buf, int len) {
 	char *res_buf = NULL;
-	bool fail = true;
 	size_t sz;
 
 	if (!buf || len <= 0) {
 		return NULL;
 	}
-	sz = wcstombs(NULL, buf, len);
+	sz = wcstombs(NULL, buf, 0);
 	if (sz == (size_t)-1) {
 		goto err_r_str_wc_to_mb;
 	}
-	res_buf = (char *)calloc(1, (sz + 1) * sizeof(char));
+	res_buf = RZ_NEWS0(char, sz + 1);
 	if (!res_buf) {
 		goto err_r_str_wc_to_mb;
 	}
@@ -3528,12 +3527,11 @@ RZ_API char *rz_str_wc_to_mb_l(const wchar_t *buf, int len) {
 	if (sz == (size_t)-1) {
 		goto err_r_str_wc_to_mb;
 	}
-	fail = false;
-err_r_str_wc_to_mb:
-	if (fail) {
-		RZ_FREE(res_buf);
-	}
 	return res_buf;
+
+err_r_str_wc_to_mb:
+	free(res_buf);
+	return NULL;
 }
 
 RZ_API char *rz_str_wc_to_mb(const wchar_t *buf) {


librz/util/str.c
```diff
@@ -3510,30 +3510,31 @@ RZ_API wchar_t *rz_str_mb_to_wc_l(const char *buf, int len) {
...
 	size_t sz;
+	mbstate_t mbs;
+	const wchar_t *tmp = buf;
 
 	if (!buf || len <= 0) {
 		return NULL;
 	}
-	sz = wcstombs(NULL, buf, len);
+	mbrlen(NULL, 0, &mbs);
+	sz = wcsrtombs(NULL, &tmp, 0, &mbs);
```
karliss: I don't think that's a correct use of mbrlen. From what I understood by reading docs mbrlen requires an initialized mbstate_t structure. Also what's the point of using `wcsrtombs` instead of wcstombs if you are not using the restartability feature?

wargio: `A call to the function with a null pointer as pmb resets the shift state (and ignores parameter max).`
http://www.cplusplus.com/reference/cwchar/mbrlen/

wargio: `wcsrtombs` allows to pass a null pointer to get the size (like snprintf). The issue is that you can't use `wcstombs` with null ptr.

karliss: cplusplus.com is only place suggesting such use of mbrlen.  Cppreference and microsoft documentation doesn't mention such behavior and suggest to zero initialize the mbstate_t structure instead.

Why do you think wcstombs can't be used with null ptr for such use case? From what I understood GCC warning was because it was used with `dst=NULL` and `len > 0` at the same time not because using NULL in general.

karliss: On the other hand we might want to use wcsrtombs in both calls, because nonrestartable version internally uses a global mbstate_t variable making it not thread safe.

wargio: i actually tried to think about this. i wasn't sure because the references are mixed; but the behavior of wcsrtombs is standardized.
therefore i do not know. i can close this and let somebody else fix this.

ret2libc:
For better reference, I added the error msg in the PR description.

librz/util/str.c
wargio: @karliss after reading more, i decided to go your way.

XVilka (approved):
LGTM as is, though we might want thread safety too.

karliss (approved):

PR #1047, opened 2021-04-19
Refactor RzTypes to use structs instead of SDB + New Tree-Sitter based C Parser

### LINKED ISSUE #275: Replace TCC Header Parser by Tree-Sitter based one

### LINKED ISSUE #370: Add Conversion to String for RzParseCTypeType
`RzParseCTypeType` is the struct that represents a complex, combined type like `char *[42]`.
There exists code to parse such strings into the `RzParseCTypeType` representation, but not the other way around. The type should be formatted to a string in correct C syntax, as used in casts for example: `(char *[42])<someexpr>`.

~~This is not necessarily blocked, but should be done ideally after #369.~~ can be started now

### LINKED ISSUE #371: Refactor all places where CTypes are used as strings to RzParseCTypeType instead.
Historically, arbitrary types are represented throughout the code as strings, for example in a variable:
https://github.com/rizinorg/rizin/blob/a3a339b11b1e241ee95201d7b666ff291c8e2100/librz/include/rz_analysis.h#L788

Or, even more importantly, in the `RzAnalysisBaseType` struct for things like the typedef target type:
https://github.com/rizinorg/rizin/blob/a3a339b11b1e241ee95201d7b666ff291c8e2100/librz/include/rz_analysis.h#L249
as well as in struct members:
https://github.com/rizinorg/rizin/blob/a3a339b11b1e241ee95201d7b666ff291c8e2100/librz/include/rz_analysis.h#L215

This is problematic because the raw string in such cases could be a more complicated pointer/array/... type like `char *[3]` and the string is thus hard to deal with for things like type propagation, showing struct member accesses, etc. where deep inspection is needed.
So instead, the `RzParseCTypeType` struct should be used in these cases which provides exactly this semantic information.
But because there are also cases where the type must be a string, for example in projects or for the target type expression of a typedef in the database, this depends on #370 to be implemented first.

Warning: Do not confuse these possible complex ctypes with base types! Base types are the ones saved in the database (structs, typedefs, enums) that have a concrete name and they may be referred to by their name as a single string. But whenever there can be pointers, arrays, etc., this is a ctype, possibly built on top a base type.

### LINKED ISSUE #373: Re-think Function Types and Callable "Things"
There currently are something like function types in the types database, but they aren't really types, as you for example can't use them as a type for a variable.
Instead, the current "function types" are more like a database of signatures for imported functions, identified by a name, and used in things like type propagation, which dirtily operates directly on the sdb.
Having such info in general is necessary.

In addition to that, there are also real analyzed functions that define their signature depending on their `RzAnalysisVar`s. Afaik there is some conversion available to take the info from vars and put it into the db explained above, but such a conversion creates redundant info, must always be in sync and is thus highly error-prone and inconvenient.

However we also need actual function types to be used anywhere in structs, variables, etc. for use-cases like analyzing code using this:
```c
struct JNINativeInterface {
    void*       reserved0;
    void*       reserved1;
    void*       reserved2;
    void*       reserved3;
    jint        (*GetVersion)(JNIEnv *);
    jclass      (*DefineClass)(JNIEnv*, const char*, jobject, const jbyte*,
                        jsize);
    jclass      (*FindClass)(JNIEnv*, const char*);
    jmethodID   (*FromReflectedMethod)(JNIEnv*, jobject);
    jfieldID    (*FromReflectedField)(JNIEnv*, jobject);
    /* spec doesn't show jboolean parameter */
    jobject     (*ToReflectedMethod)(JNIEnv*, jclass, jmethodID, jboolean);
...
```

There should be some kind of unified notion of something that is `callable`. I could imagine this to be for example a struct `RzCallable` that contains the return type, argument types, optionally argument names and a calling convention.
This struct could then be created from any source imaginable like `RzAnalysisFunction` variables, a database of imports, a type, etc. and used in analysis algorithms without having to think about what the original source was.

Now there are some decisions to be made how to best approach this:
* Do we have a dedicated "function type" plus a "callable" and convert from one or the other or have only one and for example embed an `RzCallable` in our regular type structs?
* Related to the above question, is there something that has to be present in an arbitrary "callable" from analysis perspective that does not belong into a "function type" or vice-versa?
* For function types, are they `RzType`s or `RzBaseType`s? If they are base types, that would actually be a little bit similar to what is already present in the database, but it would be harder to handle in cases like the struct above where the types are define in-line.

## Current plan for implementing this

Function type is an `RzType`. This requires a few more changes for the imported function db but it's more flexible and closer to our final goal.
We shouldn't overengineer it so let's try to specify `RzCallable` and reuse it directly in `RzType`, refactoring this will be simple if we find out later it doesn't suffice.

Structure draft:
```c
typedef struct rz_callable_arg_t {
	RZ_NULLABLE char *name; // optional, function types in C can have names to their args but they don't have to
	RzType *type;
} RzCallableArg;

typedef struct rz_callable_t {
	RzType *ret;
	RzPVector /*<RzCallableArg>*/ args;
	RZ_NULLABLE const char *cc; // optional
} RzCallable;

typedef enum {
	...,
	RZ_TYPE_KIND_FUNCTION // or _CALLABLE
} RzTypeKind;

struct rz_type_t {
	RzTypeKind kind;
	union {
		...,
		RzCallable function; // or callable
	};
};
```

Steps:

- [ ] Add `RzCallable` describing anything that can be called as explained above. This should probably go into the new `RzTypes` module (~~so this issue is sort of blocked by #369~~ can be done now).
- [ ] Add a new `RzTypeKind` which just contains a single `RzCallable` and thus wraps it and makes it possible to use function types anywhere, including such inline definitions as above.
- [ ] Convert the old function databases. They should be considered a database of known library function callables, not actual function types. This refactored database should have an api that only works with `RzCallable` (not `RzType`) and no raw sdb. Concretely I could imagine a hashtable of `name -> RzCallable` in RzAnalysis. This step might be a bit bigger since there is probably a lot that accesses the database.

Further steps:
- [ ] Add an api to `RzAnalysisFunction` that constructs an `RzCallable` from its variables
- [ ] Add an api to `RzAnalysis` so you can query the callable at any given address, no matter whether it's a regular function or an import. Care has to be taken of indirect function calls. So the address of a function somewhere in memory is not the same as an actual call target. About this point I am not 100% sure yet how exactly to design this API, it depends on how the info is used in analysis.

### LINKED ISSUE #1249: Fix `rz_analysis_var_get_argnum` to support stack variables as well
Currently this function supports only the register arguments that often leads to the incorrect analysis results.

See `librz/analysis/var.c` for the implementation.

The issue requires a few unit tests to be written for all major architectures.

XVilka:
# SQUASH ME PLEASE

**Detailed description**

- Removed oldshell version of `t` (types) commands
- Removed `tk` command because types storage switched to the hashtable
- Added `tfc` command to show the calling convention of the function or set it
- Added `tsl` (RZ_OUTPUT_MODE_LONG) to show structure members' sizes and offsets
- Added `tul` (RZ_OUTPUT_MODE_LONG) to show union members' sizes
- Switched the types storage to the hashtable instead of SDB
- Added load and export from/to SDB for standard type libraries and serialization for Projects
- Changed all `char *type` occurences to the proper `RzType`
- Added `RzCallable` type for the function/class method types
- Changed all function types API to use the `RzCallable` API
- Switched the function types (`RzCallable`) storage to the hashtable instead of SDB
- Added a connection between `RzAnalysisFunction` and `RzCallable` - `rz_analysis_function_derive_type()` API function to derive new `RzCallable` type from the `RzAnalysisFunction` even if it doesn't exist in the RzTypeDB.
- Moved type links to the RzAnalysis where they belong
- Removed TCC and MPC-based parsers
- Added Tree-Sitter-based parser for C grammar (see `librz/type/parser/*`)
- Reworked type databases in `librz/analysis/d/` to be consistent with C standard, libc, platform-specific definitions
- Switched the type links from `char *` to `RzType`
- Switched the type links from SDB to the hashtable
- Added concept of the "type paths", e.g. `a.b[20].c` where `b` is a member of `a` (`a` is either structure or a union) and `c` is a member of `b`, and `b` is an array where we take 20-th element.
- Added concept of the typeclasses (a bit similar to Haskell) for numerical types: `None` (the most generic one), `Num` (includes `Integral` and `Floating`), `Integral` (includes `Signed Integral` and `Unsigned Integral`), `Floating`.
- Added concept of type equality (`rz_types_equal(a, b)` API function)
- Added Doxygen documentation to every API function in the `librz/type/*`.
- Updated the project format from **V2** to **V3**, added migration.

**Test plan**

CI is green

**Closing issues**

Fixes #371
Fixes #370
Fixes #373
Fixes #275
Fixes #1249
Partially addresses https://github.com/rizinorg/rizin/issues/1169, https://github.com/rizinorg/rizin/issues/1064


librz/analysis/dwarf_process.c
```diff
@@ -451,7 +451,10 @@ static RzTypeStructMember *parse_struct_member(Context *ctx, ut64 idx, RzTypeStr
 	}
 
 	result->name = name;
-	result->type = type;
+	result->type = rz_type_parse_string_single(ctx->analysis->typedb->parser, type, NULL);
```
thestr4ng3r, line 454: Just a comment since I am not too familiar with types in dward, I wonder what happens for dwarf of languages other than C like maybe Rust.
Eventually, we will want to press all of them into our C-like type representation since our "types" are really more like the raw memory representation of whatever types the high-level language uses.

librz/analysis/fcn.c
```diff
@@ -1704,14 +1704,18 @@ RZ_API char *rz_analysis_function_get_json(RzAnalysisFunction *function) {
 	RzAnalysis *a = function->analysis;
 	PJ *pj = a->coreb.pjWithEncoding(a->coreb.core);
 	unsigned int i;
-	const char *ret_type = rz_type_func_ret(a->typedb, function->name);
+	RzType *ret_type = rz_type_func_ret(a->typedb, function->name);
+	if (!ret_type) {
+		return NULL;
```
thestr4ng3r: maybe fallback to `void` here too instead of returning?

librz/analysis/var.c
```diff
@@ -46,7 +46,17 @@ RZ_API bool rz_analysis_function_rebase_vars(RzAnalysis *a, RzAnalysisFunction *
...
+		// Pointers don't shadow anything
+		return;
+	}
+	if (var->type->kind == RZ_TYPE_KIND_ARRAY) {
+		// TODO: support arrays
+		return;
+	}
+	// In other cases we check the type by name
+	const char *tname = var->type->identifier.name;
+	RzBaseType *btype = rz_type_db_get_base_type(var->fcn->analysis->typedb, tname);
```
thestr4ng3r: I think what we actually want to do is calculate the size of `var->type` (in its own API) and then just radically shadow all overlaps. Doesn't have to be in this pr, but maybe add a comment.

librz/analysis/var.c
```diff
@@ -117,21 +133,16 @@ RZ_API RzAnalysisVar *rz_analysis_function_set_var(RzAnalysisFunction *fcn, int
...
 
-RZ_API void rz_analysis_var_set_type(RzAnalysisVar *var, const char *type) {
-	char *nt = strdup(type);
-	if (!nt) {
-		return;
-	}
-	free(var->type);
-	var->type = nt;
+RZ_API void rz_analysis_var_set_type(RzAnalysisVar *var, RzType *type) {
+	var->type = type;
```
thestr4ng3r, line 171: Add `RZ_OWN` to the `RzType *type` argument to indicate that ownership is transferred.

librz/include/rz_type.h
```diff
@@ -24,11 +24,15 @@ typedef struct rz_type_target_t {
 	bool big_endian;
 } RzTypeTarget;
 
+typedef struct rz_type_parser_t RzTypeParser;
+
 typedef struct rz_type_db_t {
 	void *user;
-	Sdb *sdb_types;
-	Sdb *formats; // for `pf` formats
+	HtPP *types; // A hashtable of RzBaseType
```
thestr4ng3r: Better like this imo:
```suggestion
	HtPP /* <char *, RzBaseType *> */ *types; ///< name -> base type
```

librz/type/parser/c_cpp_parser.c
```diff
@@ -0,0 +1,251 @@
...
+	rz_type_parser_free(typedb->parser);
+	typedb->parser = rz_type_parser_new();
+}
+
+// Parses only single statement (the first one) and ignores everything else
+RZ_API RZ_OWN RzType *rz_type_parse_string_single(RzTypeParser *parser, const char *code, char **error_msg) {
+	// Create a parser.
+	TSParser *tsparser = ts_parser_new();
+	// Set the parser's language (C in this case)
+	ts_parser_set_language(tsparser, tree_sitter_c());
```
thestr4ng3r, line 295: Here we would use the second grammar for types only, right?

XVilka: Yes, this thing can be extended for more or slightly different grammars.

librz/analysis/dwarf_process.c
XVilka: A good point but for now there is no quick solution for this problem. I would postpone this, will open an issue for this to track.

librz/analysis/d/types.sdb.txt
```diff
@@ -1763,11 +1763,11 @@ func.quick_exit.args=1
 func.quick_exit.arg.0=int,status
 func.quick_exit.ret=void
 
-*aligned_alloc=func
```
XVilka, line 1766: @thestr4ng3r @ret2libc not sure why is this invalid function name was in our types SDB. I converted it to the normal name. Let me know if this was intentional.

thestr4ng3r: Maybe copypasta mistake from `void *aligned_alloc(...)`, anyway `aligned_alloc` is the correct name so this is fine 🔥🐶☕🔥

wargio:
`free(): double free detected in tcache 2`

XVilka:
@wargio yes, I know, it will happen because we need to parse also "partial" types, it's already in progress, I will push the updated version once it's done.

librz/type/parser/c_cpp_parser.c
XVilka: Fixed by using slightly patched grammar.

librz/analysis/d/types-atomic.sdb.txt
```diff
@@ -0,0 +1,126 @@
...
+float=type
+type.float=f
+type.float.size=32
+
+double=type
+type.double=F
+type.double.size=64
+
+long int=type
+<<<<<<< HEAD
```
ret2libc: ??

librz/analysis/serialize_analysis.c
```diff
@@ -827,7 +829,12 @@ RZ_API RZ_NULLABLE RzAnalysisVar *rz_serialize_analysis_var_load(RZ_NONNULL RzAn
 	if (!name || !type || kind == -1 || delta == ST64_MAX) {
 		goto beach;
 	}
-	ret = rz_analysis_function_set_var(fcn, delta, kind, type, 0, arg, name);
+	char *error_msg = NULL;
+	RzType *vartype = rz_type_parse_string_single(fcn->analysis->typedb->parser, type, &error_msg);
+	if (error_msg) {
+		eprintf("Fail to parse the function variable (\"%s\") type: %s\n", name, type);
```
ret2libc, line 840: does this fallthrough on purpose? Or shall it goto beach?

subprojects/tree-sitter-c.wrap
```diff
@@ -0,0 +1,5 @@
+[wrap-git]
+url = https://github.com/XVilka/tree-sitter-c
```
ret2libc: i guess this will go under rizinorg, right?

subprojects/packagefiles/tree-sitter-c/meson.build
```diff
@@ -0,0 +1,20 @@
+project('tree-sitter-c', version: 'f05e279aedde06a25801c3f2b2cc8ac17fac52ae')
```
ret2libc: I think it's the wrong `version`. Anyway, if we get this from our own repository, I think it's better to just keep the meson.build file there.

librz/type/parser/c_cpp_parser.c
```diff
@@ -0,0 +1,288 @@
...
+	// Create a parser.
+	TSParser *parser = ts_parser_new();
+	// Set the parser's language (C in this case)
+	ts_parser_set_language(parser, tree_sitter_c());
+
+	TSTree *tree = ts_parser_parse_string(parser, NULL, code, strlen(code));
+
+	// Get the root node of the syntax tree.
+	TSNode root_node = ts_tree_root_node(tree);
+	int root_node_child_count = ts_node_named_child_count(root_node);
```
ret2libc, line 138: When do you expect this to happen?

subprojects/packagefiles/tree-sitter-c/meson.build
XVilka: What do you mean by "wrong version"? It's the latest commit in https://github.com/XVilka/tree-sitter-c/commits/bare-types-support
Also, I will update it, since I plan to add more tests in the corpus. Also will move to `rizinorg/tree-sitter-c` for now.

librz/type/parser/c_cpp_parser.c
XVilka: If it will be just `(translation_unit)`, without any nodes.

subprojects/tree-sitter-c.wrap
XVilka: Yes.

librz/analysis/serialize_analysis.c
XVilka: It was on purpose for the development time, to not stop in case of error. I will add the `goto beach` when it's finished. Lets keep this open as a reminder.

librz/analysis/d/types-atomic.sdb.txt
XVilka: Whoops, missed this chunk during the rebase.

test/unit/test_pdb.c
```diff
@@ -149,13 +185,13 @@ bool test_pdb_tpi_cpp(void) {
...
-			char *type;
-			type_info->get_print_type(type_info, &type);
-			mu_assert_streq(type, "bitfield uint32_t : 1", "Incorrect bitfield print type");
+			//} else if (type->tpi_idx == 0x1062) {
+			//	mu_assert_eq(type_info->leaf_type, eLF_BITFIELD, "Incorrect data type");
+			//	SType *base_type = NULL;
+			//	type_info->get_base_type(type_info, (void **)&base_type);
+			//	char *type;
+			//	type_info->get_print_type(type_info, &type);
+			//	mu_assert_streq(type, "bitfield uint32_t : 1", "Incorrect bitfield print type");
```
XVilka, line 153: I commented this since the test was wrong - `bitfield uint32_t : 1` is a wrong syntax and handling bitfields in PDB should be done differently.

librz/analysis/serialize_analysis.c
XVilka: Fixed now. Thanks!

librz/analysis/dwarf_process.c
thestr4ng3r: Is there an issue yet?

librz/analysis/var.c
```diff
@@ -43,36 +63,31 @@ RZ_API bool rz_analysis_function_rebase_vars(RzAnalysis *a, RzAnalysisFunction *
...
+		}
+		if (strcmp(var->name, other->name) && other->kind == var->kind && other->delta > var->delta && other->delta < var->delta + varsize) {
+			rz_analysis_var_delete(other);
 		}
 	}
-	rz_type_base_type_free(btype);
 }
 
-RZ_API RzAnalysisVar *rz_analysis_function_set_var(RzAnalysisFunction *fcn, int delta, char kind, RZ_NULLABLE const char *type, int size, bool isarg, RZ_NONNULL const char *name) {
+RZ_API RzAnalysisVar *rz_analysis_function_set_var(RzAnalysisFunction *fcn, int delta, char kind, RZ_OWN RZ_NULLABLE RzType *type, int size, bool isarg, RZ_NONNULL const char *name) {
```
thestr4ng3r: I wonder if we really want the type here to be `RZ_OWN`, since `RzType` is rather a pass-by-value kind of type. So the alternative could be to make it `RZ_BORROW const` and clone inside `rz_analysis_function_set_var()`. That might have some minor performance implications but would make using this function more straightforward, since I saw there are a lot of places that have to clone the type before passing it in here.

librz/include/rz_analysis.h
```diff
@@ -1557,7 +1560,7 @@ RZ_API RzList *rz_analysis_get_fcns(RzAnalysis *analysis);
 
 /* var.c */
 RZ_API RZ_OWN char *rz_analysis_function_autoname_var(RzAnalysisFunction *fcn, char kind, const char *pfx, int ptr);
-RZ_API RZ_BORROW RzAnalysisVar *rz_analysis_function_set_var(RzAnalysisFunction *fcn, int delta, char kind, RZ_NULLABLE const char *type, int size, bool isarg, RZ_NONNULL const char *name);
+RZ_API RZ_BORROW RzAnalysisVar *rz_analysis_function_set_var(RzAnalysisFunction *fcn, int delta, char kind, RZ_NULLABLE RzType *type, int size, bool isarg, RZ_NONNULL const char *name);
```
thestr4ng3r: `type` is missing the `RZ_OWN` (or `RZ_BORROW` if you decide to change it) here

librz/analysis/fcn.c
```diff
@@ -440,7 +440,9 @@ static bool fcn_takeover_block_recursive_followthrough_cb(RzAnalysisBlock *block
 					: other_var->delta + (other_fcn->bp_off - our_fcn->bp_off);
 				RzAnalysisVar *our_var = rz_analysis_function_get_var(our_fcn, other_var->kind, actual_delta);
 				if (!our_var) {
-					our_var = rz_analysis_function_set_var(our_fcn, actual_delta, other_var->kind, other_var->type, 0, other_var->isarg, other_var->name);
+					// Here we clone the type to assign it to the new variable
+					RzType *cloned_type = rz_type_clone(other_var->type);
+					our_var = rz_analysis_function_set_var(our_fcn, actual_delta, other_var->kind, cloned_type, 0, other_var->isarg, other_var->name);
```
thestr4ng3r: Here is an example of the cloning that could be avoided by making `rz_analysis_function_set_var()` clone itself.

librz/analysis/serialize_analysis.c
```diff
@@ -827,7 +834,13 @@ RZ_API RZ_NULLABLE RzAnalysisVar *rz_serialize_analysis_var_load(RZ_NONNULL RzAn
...
+		goto beach;
```
thestr4ng3r, line 841: This case should not fail the entire loading of the variable, since there might be cases where a variable type was "valid" before but is not anymore now and you might lose some valuable vars from your project.

What it should do instead would be to try to create some stub type like either a generic integer or void one or `RZ_TYPE_KIND_IDENTIFIER` with the identifier being the broken string so we can reconstruct the variable as good as possible.
Then the warning should not be emitted with `eprintf()` but through the `RzSerializeResultInfo *` (see other `RZ_SERIALIZE_ERR()` usages), which you probably have to pass in from the outside first, which means that the user will see the warning explicitly as part of the loading process.

We should also add a test for this best-possible recovery logic.

librz/analysis/fcn.c
```diff
@@ -1720,9 +1726,10 @@ RZ_API char *rz_analysis_function_get_json(RzAnalysisFunction *function) {
 	for (i = 0; i < argc; i++) {
 		pj_o(pj);
 		const char *arg_name = rz_type_func_args_name(a->typedb, function->name, i);
-		const char *arg_type = rz_type_func_args_type(a->typedb, function->name, i);
+		RzType *arg_type = rz_type_func_args_type(a->typedb, function->name, i);
+		const char *arg_type_str = rz_type_as_string(a->typedb, arg_type);
```
thestr4ng3r: This `char *` should not be `const` and it is leaked. I also see a lot of other places where it is like this.

librz/core/cmd_type.c
```diff
@@ -93,28 +93,28 @@ static bool print_type_c(RzCore *core, const char *ctype) {
...
-	rz_types_typedef_print_c(core->analysis->typedb, NULL);
+	rz_core_types_typedef_print_c(core->analysis->typedb, NULL);
 	// List all enums in the C format without newlines
-	rz_types_enum_print_c_all(core->analysis->typedb, false);
+	rz_core_types_enum_print_c_all(core->analysis->typedb, false);
 }
 
 static RzCmdStatus type_format_print(RzCore *core, const char *type, ut64 address) {
-	char *fmt = rz_type_format(core->analysis->typedb, type);
+	const char *fmt = rz_type_format(core->analysis->typedb, type);
```
thestr4ng3r: Aren't these `fmt`s leaking?

librz/core/ctypes.c
```diff
@@ -178,269 +165,317 @@ RZ_IPI void rz_types_enum_print_c_all(RzTypeDB *typedb, bool multiline) {
...
-	if (!strcmp(v, "struct") && !rz_str_startswith(k, "typedef")) {
-		return true;
-	}
-	if (!strcmp(v, "typedef")) {
-		const char *typedef_key = sdb_fmt("typedef.%s", k);
-		const char *type = sdb_const_get(TDB, typedef_key, NULL);
-		if (type && rz_str_startswith(type, "struct")) {
-			return true;
+static void core_types_union_print(RzCore *core, RzBaseType *btype, RzOutputMode mode, PJ *pj) {
+	rz_return_if_fail(core && btype);
```
thestr4ng3r, line 174: Probably good to also `return_if_fail` for `btype->kind == RZ_BASE_TYPE_KIND_UNION`

librz/core/ctypes.c
```diff
@@ -178,269 +165,317 @@ RZ_IPI void rz_types_enum_print_c_all(RzTypeDB *typedb, bool multiline) {
...
 
-RZ_IPI void rz_types_union_print_sdb(RzTypeDB *typedb) {
-	Sdb *TDB = typedb->sdb_types;
-	SdbList *l = sdb_foreach_list_filter(TDB, sdb_if_union_cb, true);
-	rz_types_structured_print_sdb(l);
-	ls_free(l);
+// Structures
+
+static void core_types_struct_print(RzCore *core, RzBaseType *btype, RzOutputMode mode, PJ *pj) {
+	rz_return_if_fail(core && btype);
```
thestr4ng3r, line 303: same for struct

librz/core/ctypes.c
```diff
@@ -178,269 +165,317 @@ RZ_IPI void rz_types_enum_print_c_all(RzTypeDB *typedb, bool multiline) {
...
-		if (res) {
-			rz_cons_println(res);
-		} else {
-			return false;
-		}
-	} else {
-		eprintf("This is not an typedef\n");
-		return false;
+static void core_types_typedef_print(RzCore *core, RzBaseType *btype, RzOutputMode mode, PJ *pj) {
+	rz_return_if_fail(core && btype);
```
thestr4ng3r, line 437: same

librz/include/rz_type.h
```diff
@@ -95,22 +101,32 @@ typedef struct rz_base_type_t {
...
 } RzTypeIdentifierKind;
 
-typedef struct rz_type_t RzType;
+typedef struct rz_callable_arg_t {
+	RZ_NULLABLE char *name; // optional
+	RzType *type;
+} RzCallableArg;
+
+typedef struct rz_callable_at {
+	const char *name;
```
thestr4ng3r: Can we make this optional too?

librz/type/parser/c_cpp_parser.c
```diff
@@ -0,0 +1,355 @@
...
+	}
+	ut64 file_size = rz_file_size(path);
+	RZ_LOG_DEBUG("File size is %" PFMT64d " bytes, read %zu bytes\n", file_size, read_bytes);
+	return rz_type_parse_string_stateless(parser, source_code, error_msg);
+}
+
+/**
+ * \brief Parses the C types file creating the new parser state
+ *
+ * \param typedb RzTypeDB instance
```
thestr4ng3r, line 235: What does `creating the new parser state` mean? Also it should be made more clear, maybe even through the function names, that these functions actually also save the types in the `RzTypeDB` and thus mutate it, instead of just parsing.

thestr4ng3r (changes requested):

librz/include/rz_type.h
XVilka: Making this optional in general is fine. But for saving the type in the `RzTypeDB->callables` hashtable we require a name.

thestr4ng3r: Yeah, I think we should make the name required for these exact functions and document it, but in the structure have it optional.

librz/core/ctypes.c
XVilka: Good idea. Done.

librz/include/rz_type.h
XVilka: I made it optional.

librz/analysis/var.c
XVilka: You are right, moving the need to clone the type here simplified the code and function usage.

librz/type/parser/c_cpp_parser.c
XVilka: Creating new parser state means:
- Create new hashtable for base types
- Create new hashtable for callables
- Create new hashtable for forward declarations

XVilka:
Since we now generate `RzCallable` from `RzAnalysisFunction` but the latter has poor structure of the arguments, with them being unsorted, I added new (might be temporary) API to get the arguments count and the vector of arguments. Currently, to get the index of argument we should call `rz_analysis_var_get_argnum()` but it's unreliable since it doesn't support stack arguments:
```
[i] ℤ rizin test/bins/elf/analysis/pid_stripped                                                             20:03:35
 -- Reduce the delta where flag resolving by address is used with cfg.delta
[0x00401360]> e analysis.vars.stackname=true
[0x00401360]> s 0x4e2420
[0x004e2420]> af
[0x004e2420]> afs
fcn.004e2420 : 7 arguments total
fcn.004e2420 : arg "arg_8h" has wrong position: -1
fcn.004e2420 : inserting arg "arg5" at 4 position
fcn.004e2420 : inserting arg "arg4" at 3 position
fcn.004e2420 : inserting arg "arg6" at 5 position
fcn.004e2420 : inserting arg "arg1" at 0 position
fcn.004e2420 : inserting arg "arg2" at 1 position
fcn.004e2420 : inserting arg "arg3" at 2 position
../librz/analysis/fcn.c:2397:24: runtime error: member access within null pointer of type 'struct RzAnalysisVar'
AddressSanitizer:DEADLYSIGNAL
==1288355==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000010 (pc 0x7f271b15653e bp 0x7ffd4ea7f5c0 sp 0x7
ffd4ea7f570 T0)
==1288355==The signal is caused by a READ memory access.
==1288355==Hint: address points to the zero page.
    #0 0x7f271b15653e in rz_analysis_function_derive_type ../librz/analysis/fcn.c:2397
    #1 0x7f271b156b63 in rz_analysis_function_get_signature ../librz/analysis/fcn.c:1761
    #2 0x7f27173c9887 in rz_analysis_function_signature_handler ../librz/core/cmd_analysis.c:9149
    #3 0x7f27175872c7 in argv_call_cb ../librz/core/cmd_api.c:634
    #4 0x7f27175872c7 in call_cd ../librz/core/cmd_api.c:702
    #5 0x7f27175872c7 in rz_cmd_call_parsed_args ../librz/core/cmd_api.c:720
    #6 0x7f271755c98a in handle_ts_arged_stmt_internal ../librz/core/cmd.c:4631
    #7 0x7f271755c98a in handle_ts_arged_stmt ../librz/core/cmd.c:4579
    #8 0x7f2717373e84 in handle_ts_stmt ../librz/core/cmd.c:6063
    #9 0x7f271749333a in handle_ts_statements_internal ../librz/core/cmd.c:6120
    #10 0x7f271749333a in handle_ts_statements ../librz/core/cmd.c:6085
    #11 0x7f2717494ae8 in core_cmd_tsrzcmd ../librz/core/cmd.c:6222
    #12 0x7f2717459716 in rz_core_cmd ../librz/core/cmd.c:6269
    #13 0x7f271759e5f4 in rz_core_prompt_exec ../librz/core/core.c:2825
    #14 0x7f27175a0397 in rz_core_prompt_loop ../librz/core/core.c:2676
    #15 0x7f27225b641e in rz_main_rizin ../librz/main/rizin.c:1364
    #16 0x7f2721a24b74  (/lib64/libc.so.6+0x27b74)
    #17 0x4018fd in _start (/home/akochkov/.local/bin/rizin+0x4018fd)

AddressSanitizer can not provide additional info.
SUMMARY: AddressSanitizer: SEGV ../librz/analysis/fcn.c:2397 in rz_analysis_function_derive_type
==1288355==ABORTING
```

librz/analysis/serialize_analysis.c
XVilka: Please open an issue. I will address this once this PR is merged.

XVilka:
Please note, that some of the type propagation results were wrong before and somewhat wrong after this PR, but the real cause of that is the https://github.com/rizinorg/rizin/issues/1260

XVilka:
It's strange, looks like timeout feature of the `rz-test` doesn't work properly:
```
24/84 core_cmd                       OK              9.05s
25/84 core_bin                       TIMEOUT        30.27s   exit status 1
>>> PATH=C:/projects/rizin/build/librz/analysis;C:/projects/rizin/build/lib
```
But I set `rz-test -t 2880` in the `.appveyor.yml` specifically. cc @thestr4ng3r @kazarmy @GustavoLCR @Basstorm

XVilka:
Everything passes now, but one things remain - with this PR we cross the 1 hour boundary for the AppVeyor CI job. We need to either disable AppVeyor testing or find another solution. Ideas are welcome.

thestr4ng3r (approved):

wargio (approved):

wargio:
Please remember for the next iteration to use RZ_LOG_* instead of eprintf @XVilka

PR #1062, opened 2021-04-21
Introduce new RzBinMap and refactor all Bin Plugins for it

thestr4ng3r:
**PLEASE SQUASH ME**

**Detailed description**

This introduces the new `RzBinMap`, which is the source of information that a bin plugin gives for where to map which portions of the file. Previously this info was taken from `RzBinSection` (the `add` member meant which sections to map, which is now removed) but assuming that "sections" give the info for mapping is not really accurate for all formats (e.g. ELF) and results in some awkward sections emitted just for mapping some things.

So I went through every single bin plugin and updated it to now also implement the `maps` callback. Depending on how the format works, plugins have the following options for implementing this:
* Implement both `sections` and `maps` separately if they do not have much in common or `maps` is an easy to implement subset of `sections`.
* Implement `sections` and use `rz_bin_maps_of_file_sections()` as the `maps` callback, which just copies the section info to maps. This is useful when the format has a 1:1 correspondence between sections and mappings, so it doesn't have to be implemented twice.
* Implement `maps` and in `sections` pull in all maps using `rz_bin_sections_of_maps()` and add some more sections that should not be mapped. See docs of `rz_bin_sections_of_maps()` for more info about this.

Some notes on behavorial changes:

* ELF Core File Maps now also have `fmap.`/`mmap.` prefixes because they are now handled like any other mapping. Actually I think this is even better than before because you can see easier what is actually physical map and what is zeroed:
  old:
  ```
  21 fd: 3 +0x00001000 0x56621000 - 0x56621fff r-- /home/florian/dev/crash/crash-linux-x86
  20 fd: 6 +0x00000000 0x56622000 - 0x56622fff r-- /home/florian/dev/crash/crash-linux-x86
  ```

  new:
  ```
  21 fd: 3 +0x00001000 0x56621000 - 0x56621fff r-- fmap./home/florian/dev/crash/crash-linux-x86
  20 fd: 6 +0x00000000 0x56622000 - 0x56622fff r-- mmap./home/florian/dev/crash/crash-linux-x86
  ...
* Minidump maps are now called `memory.<addr>` or `memory64.<addr>` depending on the type instead of randomly numbering them:
  old:
  ```
  9 fd: 3 +0x000043ac 0x0398f9b8 - 0x0398ffff r-- fmap.Memory_Section
  8 fd: 3 +0x000049f4 0x77639dea - 0x77639ee9 r-- fmap.Memory_Section_1
  ```

  new:
  ```
  9 fd: 3 +0x000043ac 0x0398f9b8 - 0x0398ffff r-- fmap.memory.0x398f9b8
  8 fd: 3 +0x000049f4 0x77639dea - 0x77639ee9 r-- fmap.memory.0x77639dea
  ```


**Test plan**

run the tests

I have added tests whenever possible for each bin plugin where the concrete mapping functionality was not part of the test suite yet. For plugins that are too broken to add tests (xbe and sfc) or where we don't have testbins (kernelcache) I tested manually as good as possible. mbn is untested because I have no idea where to get such a bin but it's only using `rz_bin_maps_of_file_sections()` anyway so should be fine.

### DIFF
--- a/librz/bin/bin.c
+++ b/librz/bin/bin.c
@@ -1363,6 +1363,14 @@ RZ_API const char *rz_bin_get_meth_flag_string(ut64 flag, bool compact) {
 	}
 }
 
+RZ_IPI void rz_bin_map_free(RzBinMap *map) {
+	if (!map) {
+		return;
+	}
+	free(map->name);
+	free(map);
+}
+
 RZ_IPI RzBinSection *rz_bin_section_new(const char *name) {
 	RzBinSection *s = RZ_NEW0(RzBinSection);
 	if (s) {
--- a/librz/bin/bin.c
+++ b/librz/bin/bin.c
@@ -1394,7 +1394,7 @@ RZ_API RzList *rz_bin_maps_of_file_sections(RzBinFile *binfile) {
 	}
 	RzBinSection *sec;
 	RzListIter *it;
-	rz_list_foreach(sections, it, sec) {
+	rz_list_foreach (sections, it, sec) {
 		RzBinMap *map = RZ_NEW0(RzBinMap);
 		if (!map) {
 			goto hcf;
--- a/librz/bin/bin.c
+++ b/librz/bin/bin.c
@@ -1363,12 +1363,44 @@ RZ_API const char *rz_bin_get_meth_flag_string(ut64 flag, bool compact) {
 	}
 }
 
-RZ_IPI void rz_bin_map_free(RzBinMap *map) {
+RZ_API void rz_bin_map_free(RzBinMap *map) {
 	if (!map) {
 		return;
 	}
 	free(map->name);
 	free(map);
+}
+
+/**
+ * \brief Create a list of RzBinMap from RzBinSections
+ *
+ * Some binary formats have a 1:1 correspondence of mapping and
+ * their RzBinSections. This is not always the case (e.g. ELF)
+ * but if it is, plugins can use this function to generate
+ * mappings from sections.
+ */
+RZ_API RzList *rz_bin_maps_of_sections(RzList /*<RzBinSection>*/ *sections) {
+	rz_return_val_if_fail(sections, NULL);
+	RzList *r = rz_list_newf((RzListFree)rz_bin_map_free);
+	if (!r) {
+		return NULL;
+	}
+	RzBinSection *sec;
+	RzListIter *it;
+	rz_list_foreach(sections, it, sec) {
+		RzBinMap *map = RZ_NEW0(RzBinMap);
+		if (!map) {
+			return r;
+		}
+		map->name = sec->name ? strdup(sec->name) : NULL;
+		map->paddr = sec->paddr;
+		map->psize = sec->size;
+		map->vaddr = sec->vaddr;
+		map->vsize = sec->vsize;
+		map->perm = sec->perm;
+		rz_list_push(r, map);
+	}
+	return r;
 }
 
 RZ_IPI RzBinSection *rz_bin_section_new(const char *name) {
--- a/librz/bin/bin.c
+++ b/librz/bin/bin.c
@@ -1372,15 +1372,22 @@ RZ_API void rz_bin_map_free(RzBinMap *map) {
 }
 
 /**
- * \brief Create a list of RzBinMap from RzBinSections
+ * \brief Create a list of RzBinMap from RzBinSections queried from the given file
  *
  * Some binary formats have a 1:1 correspondence of mapping and
  * their RzBinSections. This is not always the case (e.g. ELF)
- * but if it is, plugins can use this function to generate
- * mappings from sections.
- */
-RZ_API RzList *rz_bin_maps_of_sections(RzList /*<RzBinSection>*/ *sections) {
-	rz_return_val_if_fail(sections, NULL);
+ * but if it is, plugins can use this function as their maps callback,
+ * which will generate mappings for sections.
+ * */
+RZ_API RzList *rz_bin_maps_of_file_sections(RzBinFile *binfile) {
+	rz_return_val_if_fail(binfile, NULL);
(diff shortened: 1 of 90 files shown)


librz/bin/bin.c
```diff
@@ -1363,6 +1363,90 @@ RZ_API const char *rz_bin_get_meth_flag_string(ut64 flag, bool compact) {
...
+	if (!binfile->o || !binfile->o->plugin || !binfile->o->plugin->sections) {
+		return NULL;
+	}
+	RzList *sections = binfile->o->plugin->sections(binfile);
+	if (!sections) {
+		return NULL;
+	}
+	RzList *r = rz_list_newf((RzListFree)rz_bin_map_free);
+	if (!r) {
+		return NULL;
```
wargio: you forgot to free sections

librz/bin/format/ne/ne.c
```diff
@@ -75,7 +75,7 @@ RzList *rz_bin_ne_get_segments(rz_bin_ne_obj_t *bin) {
 	if (!bin) {
 		return NULL;
 	}
-	RzList *segments = rz_list_newf(free);
+	RzList *segments = rz_list_newf((RzListFree)rz_bin_section_free);
```
wargio, line 78: Should we check for null pointer?

librz/bin/p/bin_ningba.c
```diff
@@ -67,7 +67,7 @@ static RzList *sections(RzBinFile *bf) {
 	RzList *ret = NULL;
 	RzBinSection *s = RZ_NEW0(RzBinSection);
```
wargio, line 68: no check on s?

librz/bin/p/bin_nro.c
```diff
@@ -85,17 +85,76 @@ static Sdb *get_sdb(RzBinFile *bf) {
...
+		ut32 sig0sz = rz_buf_read_le32_at(bf->buf, sig0 + 4);
+		map->name = strdup("sig0");
+		map->paddr = sig0;
+		map->psize = sig0sz;
+		map->vsize = sig0sz;
+		map->vaddr = sig0 + ba;
+		map->perm = RZ_PERM_R;
+		rz_list_append(ret, map);
+	} else {
+		eprintf("Invalid SIG0 address\n");
```
wargio: use `RZ_LOG_ERROR` when possible

librz/bin/p/bin_nro.c
```diff
@@ -125,70 +183,20 @@ static RzList *sections(RzBinFile *bf) {
 		ptr->paddr = mod0;
 		ptr->vaddr = mod0 + ba;
 		ptr->perm = RZ_PERM_R; // rw-
-		ptr->add = false;
 		rz_list_append(ret, ptr);
 	} else {
 		eprintf("Invalid MOD0 address\n");
```
wargio: same here: RZ_LOG_ERROR

wargio (changes requested):

librz/core/cbin.c
```diff
@@ -673,54 +669,89 @@ static bool io_create_mem_map(RzIO *io, RzBinSection *sec, ut64 at) {
 		}
 		return false;
 	}
-	// let the section refere to the map as a memory-map
-	free(iomap->name);
-	iomap->name = sec->map_name && !sec->size ? strdup(sec->map_name) : rz_str_newf("mmap.%s", sec->name);
+	// update the io map's name to refer to the bin map
+	if (map->name) {
```
ret2libc, line 673: Does it mean that maps can be without a name? And, is it ok to have `iomap` without a name?

librz/bin/p/bin_elf.inc
```diff
@@ -192,10 +192,117 @@ static RzBinElfNoteFile *note_file_for_load_segment(struct Elf_(rz_bin_elf_obj_t
...
+			for (size_t i = 0; !sections[i].last; i++) {
+				RzBinMap *map = RZ_NEW0(RzBinMap);
+				if (!map) {
+					break;
+				}
+				map->name = strdup(sections[i].name);
+				map->paddr = sections[i].offset;
+				map->psize = sections[i].type != SHT_NOBITS ? sections[i].size : 0;
+				map->vsize = sections[i].size;
+				map->vaddr = sections[i].rva;
```
ret2libc, line 261: Ok, I think this might be a good time to fix this.

If there is no PHDR, it means this is a relocatable file (.o) and in that case AFAIK the virtual address of the sections does not have any sense (all sections have address=0, i think). Essentially, we can place sections wherever we want, but we should not have them overlap at least. That's what the linker does in other words, as it finds a place for each section in the final file.

XVilka: @ret2libc I think it's better to do that in the separate PR since it might be a bit more code than that. We already have issue to track that: https://github.com/rizinorg/rizin/issues/815

librz/core/cbin.c
thestr4ng3r: Yes, I think we should allow names of both kinds of maps to be optional. I've added `RZ_NULLABLE`s now.

librz/bin/p/bin_elf.inc
thestr4ng3r: For reference: https://github.com/rizinorg/rizin/issues/815
I think so too. I wouldn't do this in this pr since it tries to not change behaviour unless strictly necessary, but after it is a good time. 
What we can do now with `RzBinMap` is also things like map things at a vaddr that we freely decide but still report the original 0 vaddr in the `RzBinSection` as it appears in the file. We might have to adjust some analysis/search code though in case something assumes that sections will always be mapped at the `RzBinSection` vaddr.

ret2libc: Ok ok makes sense! Awesome

XVilka (approved):

wargio (approved):

PR #1066, opened 2021-04-22
8051: Make Jump Targets in Higher Banks stay inside their Bank

thestr4ng3r:

**Detailed description**

8051 can only address 16bits so we take the higher ones from the address of the instruction itself so code can be relocated as long as it is aligned to 0x10000.

Practical use: when needing more code than one 16bit space can hold, 8051 can do bankswitching between multiple 16bit banks. Mapping all of these banks can be done in rizin by just mapping them one after each other. In such a case, you need the above behavior so higher banks than 0 have their jump targets correct.

### DIFF
--- a/librz/asm/arch/8051/8051_disas.c
+++ b/librz/asm/arch/8051/8051_disas.c
@@ -95,13 +95,13 @@ static char *rz_8051_disas(ut64 pc, const ut8 *buf, int len, int *olen) {
 		case 2:
 			if (len > 1) {
 				if (arg1 == A_OFFSET) {
-					disasm = rz_str_newf(name, arg_offset(pc, pc + 2, buf[1]));
+					disasm = rz_str_newf(name, (unsigned int)arg_offset(pc, pc + 2, buf[1]));
 				} else if (arg1 == A_ADDR11) {
-					disasm = rz_str_newf(name, arg_addr11(pc, pc + 2, buf));
+					disasm = rz_str_newf(name, (unsigned int)arg_addr11(pc, pc + 2, buf));
 				} else if ((arg1 == A_RI) || (arg1 == A_RN)) {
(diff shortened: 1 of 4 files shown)


librz/asm/arch/8051/8051_disas.c
```diff
@@ -95,13 +95,13 @@ static char *rz_8051_disas(ut64 pc, const ut8 *buf, int len, int *olen) {
 		case 2:
 			if (len > 1) {
 				if (arg1 == A_OFFSET) {
-					disasm = rz_str_newf(name, arg_offset(pc + 2, buf[1]));
+					disasm = rz_str_newf(name, (unsigned int)arg_offset(pc, pc + 2, buf[1]));
```
XVilka, line 98: What about using shorter types instead?

XVilka (changes requested):

librz/asm/arch/8051/8051_disas.c
thestr4ng3r: What do you mean by shorter types? We explicitly want to support more than 16bit here. This will still chop >32bit but I would like to avoid touching the format strings in there before reviewing and testing the entire architecture implementation.

XVilka: I meant `utXX`, since `strlen("ut32")` < `strlen("unsigned int")`. Impossibly pedantic nitpick, I know.

thestr4ng3r: Ah I see. No, we can't use `ut32` since the format strings use `%x` which is defined as (unsigned) `int`.

XVilka: Ok, no worries then.

XVilka (approved):

PR #746, opened 2021-03-01
Add file section type and more flags for the ELF sections information (`iS`) ##bin

PinkNoize:
## SQUASH ME

**Your checklist for this pull request**
- [X] I've read the [guidelines for contributing](https://github.com/rizinorg/rizin/blob/master/DEVELOPERS.md) to this repository
- [X] I made sure to follow the project's [coding style](https://github.com/rizinorg/rizin/blob/master/DEVELOPERS.md#code-style)
- [X] I've added tests that prove my fix is effective or that my feature works (if possible)
- [ ] I've updated the documentation and the [rizin book](https://github.com/rizinorg/book) with the relevant information (if needed)

**Detailed description**

Adds features to make `iS` more like readelf

**Test plan**

...

**Closing issues**

Fixes https://github.com/rizinorg/rizin/issues/404

PR #1074, opened 2021-04-23
Add alignment to `iSS` and `iS` commands where appropriate

### LINKED ISSUE #403: Add file offset and memory alignment for the ELF segments information ("iSS")
`iSS`/`iSSj` should contain more information about segments in case of ELF. Two most important missing parts are file offset and memory alignment:

```
[0x00407f1c]> iSS
[Segments]

nth paddr          size vaddr         vsize perm name
―――――――――――――――――――――――――――――――――――――――――――――――――――――
0   0x00000040    0x1f8 0x00400040    0x1f8 -r-x PHDR
1   0x00000238     0x1c 0x00400238     0x1c -r-- INTERP
2   0x00000000  0x31d8c 0x00400000  0x31d8c -r-x LOAD0
3   0x00031e08   0x13e8 0x00631e08   0x4c90 -rw- LOAD1
4   0x00031e18    0x1e0 0x00631e18    0x1e0 -rw- DYNAMIC
5   0x00000254     0x20 0x00400254     0x20 -r-- NOTE
6   0x0002cc68    0xacc 0x0042cc68    0xacc -r-- GNU_EH_FRAME
7   0x00000000      0x0 0x00000000      0x0 -rw- GNU_STACK
8   0x00031e08    0x1f8 0x00631e08    0x1f8 -r-- GNU_RELRO
9   0x00000000     0x40 0x00400000     0x40 -rw- ehdr
```
Compare that with the output of `readelf -lW`:
```
[i] ℤ readelf -lW 001.make.elf.x86_64                                                                                                                                                                                             16:10:08 

Elf file type is EXEC (Executable file)
Entry point 0x407f1c
There are 9 program headers, starting at offset 64

Program Headers:
  Type           Offset   VirtAddr           PhysAddr           FileSiz  MemSiz   Flg Align
  PHDR           0x000040 0x0000000000400040 0x0000000000400040 0x0001f8 0x0001f8 R E 0x8
  INTERP         0x000238 0x0000000000400238 0x0000000000400238 0x00001c 0x00001c R   0x1
      [Requesting program interpreter: /lib64/ld-linux-x86-64.so.2]
  LOAD           0x000000 0x0000000000400000 0x0000000000400000 0x031d8c 0x031d8c R E 0x200000
  LOAD           0x031e08 0x0000000000631e08 0x0000000000631e08 0x0013e8 0x004c90 RW  0x200000
  DYNAMIC        0x031e18 0x0000000000631e18 0x0000000000631e18 0x0001e0 0x0001e0 RW  0x8
  NOTE           0x000254 0x0000000000400254 0x0000000000400254 0x000020 0x000020 R   0x4
  GNU_EH_FRAME   0x02cc68 0x000000000042cc68 0x000000000042cc68 0x000acc 0x000acc R   0x4
  GNU_STACK      0x000000 0x0000000000000000 0x0000000000000000 0x000000 0x000000 RW  0x10
  GNU_RELRO      0x031e08 0x0000000000631e08 0x0000000000631e08 0x0001f8 0x0001f8 R   0x1

 Section to Segment mapping:
  Segment Sections...
   00     
   01     .interp 
   02     .interp .note.ABI-tag .hash .dynsym .dynstr .gnu.version .gnu.version_r .rela.dyn .rela.plt .init .plt .text .fini .rodata .eh_frame_hdr .eh_frame 
   03     .init_array .fini_array .dynamic .got .got.plt .data .bss 
   04     .dynamic 
   05     .note.ABI-tag 
   06     .eh_frame_hdr 
   07     
   08     .init_array .fini_array .dynamic .got 
```

[001.make.elf.x86_64.zip](https://github.com/rizinorg/rizin/files/5779729/001.make.elf.x86_64.zip)

chinggg:

**Detailed description**

image

Now I successfully print the alignment with `iSS` command, and the output is similar to `readelf` except an extra row of `ehdr`, which must be related to one unneccery line I added to function `static RzList *sections(RzBinFile *bf)`. It can be modifed later.

Since segments are specific to some formats, I use `grep` to find `is_segment`, it turns out that only `mach0.c` and `ne.c` have this variable, so I would only need to do similar changes to deal with these formats.

...

**Test plan**

Just as #746 

...

**Closing issues**

Fixes #403 

...

You can review my first commit, I will modify and force-push it until there are no errors (and fix the typo `iS`). Then I will start to deal with other formats.


chinggg:
It seems that the last line `ehdr` which `readelf` does not show has been there before I made this commit. Its alignment is always `0x0`, which may not be a problem.

But for the Mach-O and NE format, things are different. I have checked the their structs and founded `alignment` is not in the same place as it is in ELF format.

For [Mach-O format](https://github.com/rizinorg/rizin/blob/dev/librz/bin/format/mach0/mach0_defines.h#L617-L672), `alignment` is in `struct section` instead of `struct segment_command`.

For [NE format](https://github.com/rizinorg/rizin/blob/dev/librz/bin/format/ne/ne.h), `alignment` is in `struct ne_bin_obj_t` instead of [`NE_image_segment_entry`](https://github.com/rizinorg/rizin/blob/dev/librz/bin/format/ne/ne_specs.h#L63-L68).

```diff
@@ -2267,0 +2268 @@ RzList *MACH0_(get_segments)(RzBinFile *bf) {
+                       s->align = (ut64)bin->sects[i].align;
diff --git a/librz/bin/format/ne/ne.c b/librz/bin/format/ne/ne.c
index 8b2fb6e9a..1745e756a 100644
--- a/librz/bin/format/ne/ne.c
+++ b/librz/bin/format/ne/ne.c
@@ -94,0 +95 @@ RzList *rz_bin_ne_get_segments(rz_bin_ne_obj_t *bin) {
+               bs->align = (ut64)bin->alignment;
```
Making assignment like that will be completely wrong since `alignment` of the two formats are not in the context of `segment` at all.

Since sections of different formats are all transformed to `RzBinSection` type, to which I have added `ut64 align;`, the `iSS` command now prints `align` column for the non-ELF formats which also have `segment`, but the value is `0x0` since I cannot assign them with any value.

`readelf` does not work for non-ELF formats, so I have no idea what the correct output should be. And I don't know the concept of `alignment` in different formats very well, can I follow a similar pattern to print `align` as I have done with the ELF format? If they are not compatible with ELF format in `alignment`, then my task is nearly completed. Besides the modification of ELF test results, the only thing I need to consider is whether I should exclude the two format printing `alignment` or just keep the unmeaning `0x0` shown?
@XVilka @ret2libc @thestr4ng3r Can you give any suggestions? Thanks!

XVilka:
@chinggg for Mach-O it is better to probably print that alignment in `iS` command then. For the NE - it's probably better to just ignore it, at least for now.

chinggg:
> @chinggg for Mach-O it is better to probably print that alignment in `iS` command then. For the NE - it's probably better to just ignore it, at least for now.

So I will check if `align` exists for the segment then output the value, segments without `align` will keep the header but no values shown.

And to print section alignment for Mach-O, I have to add `align` column to `iS` command though only Mach-O has the value, which can be done by moving `rz_table_set_columnsf(table, "x", "align");` out of `if (print_segments)`.
Since all formats share the `RzBinSection` type, the `align` header will be there for all formats, which can be accepted though.

chinggg:
For ELF, the output will be like this.
```bash
❯ rizin bins/elf/analysis/x86-helloworld-gcc
 -- Show offsets in graphs with 'e graph.offset=true'
[0x08048300]> iSS
[Segments]

nth paddr        size vaddr       vsize perm name         align
―――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――
0   0x00000034  0x100 0x08048034  0x100 -r-x PHDR         0x4
1   0x00000134   0x13 0x08048134   0x13 -r-- INTERP       0x1
2   0x00000000  0x59c 0x08048000  0x59c -r-x LOAD0        0x1000
3   0x0000059c  0x118 0x0804959c  0x11c -rw- LOAD1        0x1000
4   0x000005a8   0xe8 0x080495a8   0xe8 -rw- DYNAMIC      0x4
5   0x00000148   0x44 0x08048148   0x44 -r-- NOTE         0x4
6   0x000004c0   0x2c 0x080484c0   0x2c -r-- GNU_EH_FRAME 0x4
7   0x00000000    0x0 0x00000000    0x0 -rw- GNU_STACK    0x4
8   0x00000000   0x34 0x08048000   0x34 -rw- ehdr


[0x08048300]> iS
[Sections]

nth paddr        size vaddr       vsize perm name               type       flags         align
――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――
0   0x00000000    0x0 0x00000000    0x0 ----                    NULL       
1   0x00000134   0x13 0x08048134   0x13 -r-- .interp            PROGBITS   alloc
2   0x00000148   0x20 0x08048148   0x20 -r-- .note.ABI-tag      NOTE       alloc
3   0x00000168   0x24 0x08048168   0x24 -r-- .note.gnu.build-id NOTE       alloc
...(omitted)...
```

For `Mach-O`, the output will be like this.
```bash
❯ rizin bins/mach0/hello-osx-x86_64
 -- Use '-e bin.strings=false' to disable automatic string search when loading the binary.
[0x100000ed0]> iSS
[Segments]

nth paddr         size vaddr         vsize perm name     align
――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――
0   0x00000000  0x1000 0x100000000  0x1000 -r-x TEXT
1   0x00001000  0x1000 0x100001000  0x1000 -rw- DATA
2   0x00002000  0x1000 0x100002000  0x1000 -r-- LINKEDIT


[0x100000ed0]> iS
[Sections]

nth paddr       size vaddr        vsize perm name                     type                     flags                               align
――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――
0   0x00000ed0  0x53 0x100000ed0   0x53 -r-x 0.__TEXT.__text          REGULAR                  PURE_INSTRUCTIONS,SOME_INSTRUCTIONS 0x2
1   0x00000f24   0xc 0x100000f24    0xc -r-x 1.__TEXT.__symbol_stub1  SYMBOL_STUBS             PURE_INSTRUCTIONS,SOME_INSTRUCTIONS 0x1
2   0x00000f30   0xc 0x100000f30    0xc -r-x 2.__TEXT.__cstring       CSTRING_LITERALS         
3   0x00000f3c  0x24 0x100000f3c   0x24 -r-x 3.__TEXT.__stub_helper   REGULAR                  PURE_INSTRUCTIONS,SOME_INSTRUCTIONS
4   0x00000f60  0x54 0x100000f60   0x54 -r-x 4.__TEXT.__unwind_info   REGULAR                                                      0x4
5   0x00000fb8  0x48 0x100000fb8   0x48 -r-x 5.__TEXT.__eh_frame      COALESCED                NO_TOC,STRIP_STATIC_SYMS            0x3
6   0x00001000  0x28 0x100001000   0x28 -rw- 6.__DATA.__program_vars  REGULAR                                                      0x5
7   0x00001028  0x10 0x100001028   0x10 -rw- 7.__DATA.__nl_symbol_ptr NON_LAZY_SYMBOL_POINTERS                                     0x2
8   0x00001038  0x10 0x100001038   0x10 -rw- 8.__DATA.__la_symbol_ptr LAZY_SYMBOL_POINTERS                                         0x2
9   0x00001048  0x20 0x100001048   0x20 -rw- 9.__DATA.__data          REGULAR                                                      0x3
```
@XVilka What's your opinion? It seems there is no way to avoid printing `align` header for all formats, the expected test results will be changed a lot.

XVilka:
What about checking for the file format specifically to print segment alignment only in case of ELF and section alignment only in case of MachO?

thestr4ng3r:
The `ehdr` "segment" has been removed recently since it was only needed for mapping but isn't a real segment, so that is fine.

Regarding formats that don't have alignment I would print the column only when there is at least one section/segment in the list with `align != 0`.

For mach-o, be aware that the align value is a power of 2, so the `RzBinSection` should contain `2^align` here to be consistent with elf:
https://web.archive.org/web/20090901205800/http://developer.apple.com/mac/library/documentation/DeveloperTools/Conceptual/MachORuntime/Reference/reference.html
> align
>
>    An integer specifying the section’s byte alignment. Specify this as a power of two; for example, a section with 8-byte alignment would have an align value of 3 (2 to the 3rd power equals 8).

XVilka (approved):

librz/core/cbin.c
```diff
@@ -2833,7 +2833,7 @@ static RzList *section_flag_to_rzlist(RzBin *bin, ut64 flag) {
 
 static int bin_sections(RzCore *r, PJ *pj, int mode, ut64 laddr, int va, ut64 at, const char *name, const char *chksum, bool print_segments) {
 	RzBinSection *section;
-	RzBinInfo *info = NULL;
```
chinggg: It seems that `info` has never been got from file before, that is weird. Is there any reason to keep it `NULL`?

librz/core/cbin.c
```diff
@@ -2916,6 +2918,12 @@ static int bin_sections(RzCore *r, PJ *pj, int mode, ut64 laddr, int va, ut64 at
 		if (plugin_flags_support && !print_segments) {
 			rz_table_set_columnsf(table, "s", "flags");
 		}
+		if (is_elf && print_segments) {
+			rz_table_set_columnsf(table, "x", "align");
+		}
+		if (is_mach0 && !print_segments) {
+			rz_table_set_columnsf(table, "x", "align");
+		}
```
chinggg: If you think using something like `print_align` to avoid duplicate line is better, I can make changes.

chinggg:
You can change the title of this Pull Request since it is more than `ELF alignment` and `iSS` command. 
In summary, there are two changes:
- ELF segment alignment for `iSS` command
- Mach-O section alignment for `iS` command

librz/core/cbin.c
XVilka: Good idea, please do.

librz/core/cbin.c
```diff
@@ -2848,6 +2848,8 @@ static int bin_sections(RzCore *r, PJ *pj, int mode, ut64 laddr, int va, ut64 at
 	const char *type = print_segments ? "segment" : "section";
 	bool plugin_type_support = plugin && plugin->section_type_to_string;
 	bool plugin_flags_support = plugin && plugin->section_flag_to_rzlist;
+	bool is_elf = info->rclass && !strcmp(info->rclass, "elf");
+	bool is_mach0 = info->rclass && !strcmp(info->rclass, "mach0");
```
thestr4ng3r: Hardcoding formats is a bad practice, imo just iterate through `sections` first and check if there is alignment info in one entry. Then you also don't need the `info` at all.

thestr4ng3r (changes requested):

librz/core/cbin.c
chinggg: Yes, I had thought about checking all the sections first but I wonder whether it will be more slow. Now I only know two kinds of formats having `align` info. I think file formats will be relatively stable so we are not likely to add new `align` frequently.

But to avoid hardcoding, I can iterate through `sections` first. And do you think I should keep the `info` to be `NULL`?

thestr4ng3r: Since iterating over all sections is necessary for printing anyway, the complexity will not change, so we don't need to worry much about performance here.

If `info` is always `NULL` and thus effectively unused, it can be removed completely. It might be a leftover after refactoring.

chinggg:
I have ran `rz-test` on my local machine and passed all related tests.

XVilka (approved):

ret2libc (approved):

wargio (approved):

PR #1079, opened 2021-04-26
Add RzBinVirtualFile

thestr4ng3r:
**DO NOT SQUASH**

Implementation of https://www.notion.so/rizin/Mapping-Data-Produced-by-RzBin-Plugins-f9ade70e855247a58d885441b9155ca8 For the full description read this.

**tl;dr**: This implements "virtual files" which are arbitrary buffers that can be exposed by bin plugins and can then be mapped into io to represent data that is not directly available in the raw binary, such as decompressed data, patched relocs, etc.

What exactly is contained in this pr:
* 09244da RzBinVirtualFile API/Interface in RzBin + the corresponding part in RzCore that takes care of mapping everything
* e661eeb 89e9ba0 c04e06f Rewritten ELF Relocs Patching in 3 steps, `io.cache` is not needed anymore as all patching happens in virtual files. Writing into a file (in the areas that are not patched) is still supported and tested. Patching relocs is enabled by default (no ugly warning anymore when opening an elf) and can be disabled using `bin.relocs`.


XVilka:
Could you please explain a bit what problem it solves in the PR description?

thestr4ng3r:
Yes, when it's less wip.

thestr4ng3r:
TODO: add some more docs

librz/core/cbin.c
```diff
@@ -694,8 +694,32 @@ static void add_map(RzCore *core, RZ_NULLABLE RzCoreFile *cf, RzBinMap *map, ut6
 		return;
 	}
 
-	// then we map the part of the section that comes from the physical file
-	char *map_name = map->name ? rz_str_newf("fmap.%s", map->name) : rz_str_newf("fmap.%d", fd);
+	const char *prefix = "fmap";
+
+	// open and use a different fd for virtual files
+	if (map->vfile_name) {
+		char *uri = rz_str_newf("vfile://%" PFMT32u "/%s", bf->id, map->vfile_name);
```
ret2libc, line 728: How will this work for projects? Is `bf->id` here ok for serializing/deserializing projects?
BTW, do we want to have projects considered immediately in each PR or as an afterthought? That is, shall we require you/contributors to immediately fix projects when they do these kind of changes or is it ok to have them fixed "later"?

librz/core/core.c
```diff
@@ -2476,11 +2486,13 @@ RZ_API bool rz_core_init(RzCore *core) {
 	/// XXX shouhld be using coreb
 	rz_parse_set_user_ptr(core->parser, core);
 	core->bin = rz_bin_new();
+	rz_event_hook(core->bin->event, RZ_EVENT_BIN_FILE_DEL, ev_binfiledel_cb, core);
```
ret2libc: :1st_place_medal:

librz/core/cvfile.c
```diff
@@ -0,0 +1,168 @@
...
+
+typedef struct {
+	RzBinFile *bf;
+	RzBinVirtualFile *vfile;
+	ut64 off;
+} VFileCtx;
+
+extern RzIOPlugin rz_core_io_plugin_vfile;
+
+static bool __check(RzIO *io, const char *pathname, bool many) {
```
ret2libc: I think we should not use `__` names...

librz/core/cvfile.c
```diff
@@ -0,0 +1,168 @@
...
+	return !strncmp(pathname, URI_SCHEME, URI_SCHEME_LEN);
+}
+
+static RzIODesc *__open(RzIO *io, const char *pathname, int rw, int mode) {
+	if (strncmp(pathname, URI_SCHEME, URI_SCHEME_LEN)) {
```
ret2libc: reuse `__check`?

librz/core/cvfile.c
```diff
@@ -0,0 +1,168 @@
...
+	if (!bf) {
+		RZ_LOG_ERROR("No bin file for id %" PFMT32u " from URI \"%s\"\n", bfid, pathname);
+		goto beach;
+	}
+	RzBinVirtualFile *vfile = NULL;
+	if (bf->o && bf->o->vfiles) {
+		RzListIter *it;
+		rz_list_foreach (bf->o->vfiles, it, vfile) {
+			if (!strcmp(vfile->name, filename)) {
+				goto found_file;
```
ret2libc: Abusing `goto` i'd say :D

Extract a `find_vfile` function and use that instead of multiple gotos.

librz/core/cbin.c
thestr4ng3r: In this case, since fine-grained bin/io serialization is not implemented at all yet, this does not need and updates in projects.

> How will this work for projects? Is bf->id here ok for serializing/deserializing projects?

In the way that I plan to implement bin/io serialization, yes this will be ok. I plan to make the ids and fds stable across loads. But even if they would not be stable (for example if we want to deserialize io/bin loadings on top of already loaded stuff, not sure if useful) then the loading code can reassign new ids as long as the correspondences are correctly represented in the serialization.

> BTW, do we want to have projects considered immediately in each PR or as an afterthought? That is, shall we require you/contributors to immediately fix projects when they do these kind of changes or is it ok to have them fixed "later"?

When projects are stable, they will have to be considered and updated/fixed immediately per commit (not per pr, if the pr is not squashed). Since we are in beta right now we can see it a little bit relaxed but should still update immediately whenever possible.

librz/core/cvfile.c
thestr4ng3r: The thing here is that right after this, I do `pathname + URI_SCHEME_LEN` so it relies on the fact that pathname is at least `URI_SCHEME_LEN` long. If I would use the check function here and the check function is updated later, this consequence may be forgotten.

ret2libc: mmm isn't this exactly what `__check` shall do? I mean, IIRC check is called before calling open, so keeping the condition only in check seems to be the best thing to do to avoid issues in case things are updated later.

librz/core/cvfile.c
thestr4ng3r:
```basic
10 print "fixed"
20 goto 10
```

librz/core/cvfile.c
thestr4ng3r: I think it's more of an implementation detail that `check` is always called before opening. I could imagine forcing another io plugin using C api but I think that's also a question of the IO API design, so it's debatable.

ret2libc: I didn't mean to rely on the fact that `vf_check` is already called, but that this check and the check in `vf_check` should be essentially the same I think.

thestr4ng3r: I've changed it to use `vf_check` now and added a comment in the `vf_check` body.

thestr4ng3r:
TODO (for me):
- [ ] Try to remove rva roundtrips
- [ ] Test `bin.relocs=false`
- [ ] remove now-unnecessary `io.cache` check

XVilka:
What happens if we want to patch the area where relocation patch already applies?

test/db/formats/elf/reloc
```diff
@@ -90,38 +88,50 @@ pd 1 @ 0x0800008c
...
+ 2 fd: 3 +0x00000920 0x08000920 - 0x08000aff --- fmap..symtab
+ 1 fd: 3 +0x00000000 0x08000000 - 0x0800003f r-- fmap.ehdr
+--
+            0x08000079      e800000000     call  strcmp
+            0x080000a2      e800000000     call  strcmp
 EOF
 RUN
 
 NAME=ELF: reloc flags
 FILE=bins/elf/analysis/filetime.c-clang-x64-O0.o
```
XVilka, line 131: Could you please add also the [rawmem.c-gcc-x64-O3.o](https://github.com/rizinorg/rz-ghidra/files/6120419/rawmem.c-gcc-x64-O3.o.zip) from https://github.com/rizinorg/rz-ghidra/issues/203 to tests?

test/db/formats/elf/elf-relro
```diff
@@ -12,25 +12,25 @@ CMDS=ir
...
-0x0000e260 0x0000e260 SET_64 free
-0x0000e268 0x0000e268 SET_64 AndroidBitmap_lockPixels
-0x0000e270 0x0000e270 SET_64 AndroidBitmap_getInfo
-0x0000e278 0x0000e278 SET_64 AndroidBitmap_unlockPixels
-0x0000e280 0x0000e280 SET_64 memcpy
-0x0000e288 0x0000e288 SET_64 memset
-0x0000e290 0x0000e290 SET_64 dlsym
+vaddr      paddr      target     type   name
+--------------------------------------------
+0x0000e210 0x0000e210 0x0000f740 SET_64 __cxa_finalize
```
XVilka, line 17: Good addition for the output :+1:

test/db/cmd/types
```diff
@@ -132,7 +132,17 @@ e asm.lines.bb=false
 aa
 s sym.Java_com_app_ndh_NDHActivity_print
 to bins/headers/jni.h
-tl JNINativeInterface 0x464c457f
+
+# This is actually quite wrong, regarding what this function actually does :-)
```
XVilka, line 137: I suggest to add `FIXME:` or `TODO:` here, we will address this during our RzTypes work soon.

XVilka: Or just even open an issue and assign it to me.

librz/bin/p/bin_elf.inc
```diff
@@ -299,391 +299,204 @@ static RzList *maps(RzBinFile *bf) {
...
-				ptr->name = strdup("DYNAMIC");
+/// size of the artificial reloc target vfile
+static ut64 reloc_targets_vfile_size(RzBinFile *bf, struct Elf_(rz_bin_elf_obj_t) * obj) {
+	if (!bf->o || !bf->o->opts.patch_relocs) {
+		return 0;
+	}
+	return obj->g_reloc_num * reloc_target_size(obj);
+}
+
+static void patch_reloc(struct Elf_(rz_bin_elf_obj_t) * obj, RzBinElfReloc *rel, ut64 S, ut64 B, ut64 L) {
```
XVilka, line 343: I wonder if it makes sense to move it into `librz/bin/format/elf` directory and call `elf_reloc.c` since it might grow quite big if we have to support all reloc types.

librz/bin/bobj.c
```diff
@@ -242,29 +243,32 @@ static RzList *classes_from_symbols(RzBinFile *bf) {
...
 	rz_return_val_if_fail(bf && plugin, NULL);
 	ut64 bytes_sz = rz_buf_size(bf->buf);
 	Sdb *sdb = bf->sdb;
 	RzBinObject *o = RZ_NEW0(RzBinObject);
 	if (!o) {
 		return NULL;
 	}
+	o->opts = *opts;
+	if (o->opts.loadaddr == UT64_MAX) {
+		// no loadaddr means 0 loadaddr
```
XVilka, line 256: Please note, that in `rz_core_fileopen()` and `rz_core_bin_load()` the `loadaddr == UT64_MAX` means autodetection, not the `0` loadaddr. Might be useful to point the distinction.

XVilka (changes requested):
Looks very good! I have a few minor suggestions though.

test/db/formats/elf/reloc
thestr4ng3r: Will add it. Are there any specific differences to the filetime one that I should test for especially?

thestr4ng3r:
> What happens if we want to patch the area where relocation patch already applies?

When you write in such an area, the write will be carried out below the patch, so it overwrites whatever is in the file there, but not our patch above. You will still see the patch after the write:

```
florian-macbook:test florian$ cp bins/elf/analysis/filetime.c-clang-x64-O0.o .tmp/elf-reloc-patchee
florian-macbook:test florian$ rz -w .tmp/elf-reloc-patchee
 -- Interpret rizin scripts with '. <path-to-script>'. Similar to the bash source alias command.
[0x08000040]> s 0x08000072
[0x08000072]> pd 3
            0x08000072      mov   ecx, reloc.target..rodata.str1.1     ; 0x8000636 ; "--help"; RELOC 32 .rodata.str1.1 @ 0x08000636
            0x08000077      mov   esi, ecx
            0x08000079      call  reloc.target.strcmp                  ; RELOC 32 strcmp
[0x08000072]> px 0x10
- offset -   0 1  2 3  4 5  6 7  8 9  A B  C D  E F  0123456789ABCDEF
0x08000072  b936 0600 0889 cee8 820e 0000 83f8 000f  .6..............
[0x08000072]> wx 42424242424242424242424242424242
[0x08000072]> px 0x10
- offset -   0 1  2 3  4 5  6 7  8 9  A B  C D  E F  0123456789ABCDEF
0x08000072  4236 0600 0842 4242 820e 0000 4242 4242  B6...BBB....BBBB
^D
florian-macbook:test florian$ rz -e bin.relocs=0 .tmp/elf-reloc-patchee
 -- Find hexpairs with '/x a0 cc 33'
[0x08000040]> px 0x10 @ 0x08000072
- offset -   0 1  2 3  4 5  6 7  8 9  A B  C D  E F  0123456789ABCDEF
0x08000072  4242 4242 4242 4242 4242 4242 4242 4242  BBBBBBBBBBBBBBBB
^D
florian-macbook:test florian$ rz -e bin.relocs=1 .tmp/elf-reloc-patchee
 -- Remember to maintain your ~/.rizin_history
[0x08000040]> px 0x10 @ 0x08000072
- offset -   0 1  2 3  4 5  6 7  8 9  A B  C D  E F  0123456789ABCDEF
0x08000072  4236 0600 0842 4242 820e 0000 4242 4242  B6...BBB....BBBB
```

That is the behavior of `RZ_BUF_SPARSE_WRITE_MODE_THROUGH` but other write modes can be implemented if desired.

librz/bin/bobj.c
thestr4ng3r: Thanks, that is a good pointer. I will re-check what the loadaddr means in detail here.

test/db/cmd/types
thestr4ng3r: #1132

librz/bin/bobj.c
thestr4ng3r: I can't see any autodetection of the laddr in these functions, from what I can tell it just gets passed into `rz_io_map_new()` (Except in `rz_core_file_open_many()` where it is used as the baddr somehow but that looks very wrong). Do you maybe mean the baddr autodetection?

test/db/formats/elf/reloc
thestr4ng3r: Oh there is a nice jumptable that is built on relocs, I will add a test for that.

librz/bin/p/bin_elf.inc
thestr4ng3r: That is a good idea, the `patch_relocs` and `patch_reloc` could indeed fit well in `librz/bin/format/elf`. Though I just tried doing this refactor and it turned out not entirely trivial because `reloc_targets_map_base()` uses `maps_unpatched()` so I think we should postpone it for after this pr.

librz/core/cvfile.c
```diff
@@ -0,0 +1,175 @@
+// SPDX-FileCopyrightText: 2021 Florian Märkl <info€florianmaerkl.de>
```
wargio: EUR?

thestr4ng3r: lol thanks

librz/bin/bobj.c
XVilka: I was referring exactly to this:
> Except in rz_core_file_open_many() where it is used as the baddr somehow but that looks very wrong.

Also this: https://github.com/rizinorg/rizin/blob/dev/librz/core/cfile.c#L1202:
```c
/* loadaddr is r2 -m (mapaddr) */
RZ_API RzCoreFile *rz_core_file_open(RzCore *r, const char *file, int flags, ut64 loadaddr) {
...
	//used by rz_core_bin_load otherwise won't load correctly
	//this should be argument of rz_core_bin_load <shrug>
	if (loadaddr != UT64_MAX) {
		rz_config_set_i(r->config, "bin.laddr", loadaddr);
	}
	rz_core_cmd0(r, "R!");
beach:
	r->times->file_open_time = rz_time_now_mono() - prev;
	return fh;
}
```

librz/bin/p/bin_elf.inc
```diff
@@ -808,302 +622,578 @@ static RzBinReloc *reloc_convert(struct Elf_(rz_bin_elf_obj_t) * bin, RzBinElfRe
...
+				}
+			}
 		}
-		break;
+		rz_list_append(ret, ptr);
+	}
+
+	// add entrypoint for jni libraries
+	// NOTE: this is slow, we shouldnt find for java constructors here
+	if (!(symbol = Elf_(rz_bin_elf_get_symbols)(obj))) {
```
XVilka, line 1041: Since you are the Java expert, @wargio please take a look at this particular piece of code.

XVilka (approved):
LGTM now.

librz/bin/p/bin_elf.inc
wargio: LGTM

librz/bin/bobj.c
thestr4ng3r: Ok, then as far as I can tell the comment in that specific place is correct. Still a lot to improve for laddr stuff probably.

librz/bin/p/bin_elf.inc
thestr4ng3r: Opened an issue: https://github.com/rizinorg/rizin/issues/1141

PR #1091, opened 2021-04-30
Track all related RzIOMaps and RzIODescs in RzCoreFile and other stories

thestr4ng3r:

**Detailed description**

`RzCoreFile` is for tracking any low-level state in `RzIO` and `RzBin` together, for example to correctly tear down all associated maps of a file when it is closed, but keeping mappings from other sources open.
Previously, it failed hard at that. For example when you opened an ELF file that had some zero-mmaps as an `RzCoreFile`, then loaded all the bin maps from it, then called `rz_core_file_close()` on it again, you would expect the created mappings including the zero one to be cleaned up again, but this was not the case:
```
florian-macbook:rizin florian$ rz test/bins/elf/true
 -- Move between your search hits in visual mode using the 'f' and 'F' keys
[0x00002110]> o
 3 * r-x 0x000087c0 test/bins/elf/true
 4 - rw- 0x00000198 null://408
[0x00002110]> om
 5 fd: 3 +0x00000000 0x00000000 - 0x00001167 r-- fmap.LOAD0
 4 fd: 3 +0x00002000 0x00002000 - 0x000050f0 r-x fmap.LOAD1
 3 fd: 3 +0x00006000 0x00006000 - 0x00007aff r-- fmap.LOAD2
 2 fd: 3 +0x00007c70 0x00008c70 - 0x0000907f r-- fmap.LOAD3
 1 fd: 4 +0x00000000 0x00009080 - 0x00009217 rw- mmap.LOAD3
[0x00002110]> o-3
[0x00002110]> o
 4 - rw- 0x00000198 null://408
[0x00002110]> om
 1 fd: 4 +0x00000000 0x00009080 - 0x00009217 rw- mmap.LOAD3
 ```
Here, `o-3` closes the core file as opposed to just the io file. If it was only the io file, this behavior would be correct but for the core file it's wrong.

Now the `RzCoreFile` tracks all additional files and maps that it creates during loading so they can be cleanly removed later again, and also without accidentally removing anything unrelated to that core file. The fact that the core file now holds pointers to IO maps and descs has the implication that when these low-level IO objects are deleted manually from elsewhere (which is possible and valid), then the core file must remove its references to them automatically, otherwise we will double-free.
This is why this also adds related events `RZ_EVENT_IO_DESC_CLOSE` and `RZ_EVENT_IO_MAP_DEL` so these cases can be tracked correctly.

In addition, 436871e8d6f1e48e5b26cadcacd97467bd2f18f1 removed a lot of old workarounds that were put on top of workarounds to avoid double-frees and stuff but were actually all not necessary if everything was handled right.
d26d25e35f5073dcaa88b30e0e4c446e5ae09abb also fixes a minor bug where the perms of zero-mappings were not set correctly.

**Test plan**

See the unit tests. I added tests for the new IO events, to make sure they are always emitted correctly so no free will be missed, and for the whole `RzCoreFile` stuff, making sure that it closes exactly what it should and does not double-free.

### DIFF
--- a/test/integration/test_open_analyse_save_load_project.c
+++ b/test/integration/test_open_analyse_save_load_project.c
@@ -44,15 +44,12 @@ bool test_open_analyse_save() {
 	mu_assert_eq(err, RZ_PROJECT_ERR_SUCCESS, "project save err");
 
 	// 5. Close the file
-	rz_core_file_free(file);
+	rz_core_file_close(file);
 	rz_core_free(core);
 
-	// 6. Open the file again
+	// 6. Create a new core
 	core = rz_core_new();
 	mu_assert_notnull(core, "new RzCore instance");
-	file = rz_core_file_open(core, fpath, RZ_PERM_R, loadaddr);
-	mu_assert_notnull(file, "open file");
-	rz_core_bin_load(core, fpath, loadaddr);
 
 	// 7. Load the previously saved project
 	RzSerializeResultInfo *res = rz_serialize_result_info_new();
@@ -69,7 +66,6 @@ bool test_open_analyse_save() {
 
 	// 10. Exit
 	rz_serialize_result_info_free(res);
-	rz_core_file_free(file);
 	rz_core_free(core);
 	mu_end;
 }
--- a/librz/bin/bfile.c
+++ b/librz/bin/bfile.c
@@ -672,20 +672,14 @@ RZ_API bool rz_bin_file_set_cur_by_name(RzBin *bin, const char *name) {
 	return rz_bin_file_set_cur_binfile(bin, bf);
 }
 
-RZ_API bool rz_bin_file_deref(RzBin *bin, RzBinFile *a) {
-	rz_return_val_if_fail(bin && a, false);
-	if (!rz_bin_cur_object(bin)) {
-		return false;
-	}
-	bin->cur = NULL;
-	return true;
-}
-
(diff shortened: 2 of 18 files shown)

### CODE REVIEW

File: test/integration/test_open_analyse_save_load_project.c [hunk 1]
```diff
@@ -44,15 +44,12 @@ bool test_open_analyse_save() {
...
 
 	// 5. Close the file
-	rz_core_file_free(file);
+	rz_core_file_close(file);
 	rz_core_free(core);
 
-	// 6. Open the file again
+	// 6. Create a new core
 	core = rz_core_new();
 	mu_assert_notnull(core, "new RzCore instance");
```
XVilka, line 52: Is this removal for a purpose? Why it was left as is on the first opening but you removed it after?

XVilka (approved):
Well thought and looks good. Nice to have even more event here 👍

test/integration/test_open_analyse_save_load_project.c
thestr4ng3r: The second time you are loading a project with the `load_bin_io = true` arg. This will reload the file automatically so loading the file manually the second time is not necessary and it would just be overwritten.

wargio (approved):
love it

wargio:
@thestr4ng3r i do not know if i can squash. so its up to you on how to merge this

thestr4ng3r:
Better not squash, the commits are self-contained.

PR #1097, opened 2021-05-02
Refactor RzBinFile Deletion API and add Events + other stuff

thestr4ng3r:
**DO NOT SQUASH**

**Detailed description**

This comes out of #1079. I would like to keep it as a draft for a while to be able to still change things if I notice that it's necessary. But feel free to review nonetheless! It's almost finished.

This contains multiple changes:

* Before, you would often see this pattern:
    ```c
    rz_bin_open(bin, "some file", ...);
    RzBinFile *bf = rz_bin_cur(bin);
    // do stuff with bf
    ```
    This relies on the newly opened file to be the current one afterwards but it is also kind of nonsense because the `rz_bin_open()` already knows the pointer so I changed it to just return it (`NULL` means failure):
    ```
    RzBinFile *bf = rz_bin_open(bin, "some file", ...);
    // do stuff with bf
    ```
* `rz_bin_file_delete()` now take an `RzBinFile *` instead of its id. This eliminates several roundtrips where for example the user entered an id, core would fetch the `RzBinFile *bf` for it, then call `rz_bin_file_delete(bf->id)`, then `rz_bin_file_delete()` would fetch the `RzBinFile *` again, etc.
* `RZ_EVENT_BIN_FILE_DEL` has been added for detecting file deletion
* `RzCoreFile` now also tracks `RzBinFile`s like in #1091 and they are closed when the core file is closed. Before, they leaked.

**Test plan**

unit tests

### DIFF
--- a/librz/include/rz_bin.h
+++ b/librz/include/rz_bin.h
@@ -320,6 +320,7 @@ struct rz_bin_t {
 	RZ_DEPRECATE RzBinFile *cur; ///< never use this in new code! Get a file from the binfiles list or track it yourself.
 	int narch;
 	void *user;
+	RzEvent *event;
 	/* preconfigured values */
 	int debase64;
 	int minstrlen;
@@ -783,6 +784,10 @@ typedef struct rz_bin_options_t {
 	const char *filename;
 } RzBinOptions;
 
+typedef struct rz_event_bin_file_del_t {
+	RzBinFile *bf;
+} RzEventBinFileDel;
+
 RZ_API RzBinImport *rz_bin_import_clone(RzBinImport *o);
 RZ_API const char *rz_bin_symbol_name(RzBinSymbol *s);
 typedef void (*RzBinSymbolCallback)(RzBinObject *obj, RzBinSymbol *symbol);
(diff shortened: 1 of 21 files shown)


librz/include/rz_bin.h
```diff
@@ -300,8 +300,8 @@ struct rz_bin_file_t {
 	struct rz_bin_xtr_plugin_t *curxtr;
 	// struct rz_bin_plugin_t *curplugin; // use o->plugin
 	RzList *xtr_data;
-	Sdb *sdb;
-	Sdb *sdb_info;
+	RZ_DEPRECATE Sdb *sdb;
```
XVilka: What is supposed to be used instead? Maybe add it as a comment?

test/db/cmd/cmd_open
```diff
@@ -410,36 +410,53 @@ RUN
 NAME=o-*
 FILE==
 CMDS=<<EOF
-o -
```
XVilka, line 413: I am surprised it was working before.

XVilka (changes requested):

test/db/cmd/cmd_open
thestr4ng3r: It tested that after `o-*` everything was deleted... but it never said that before it anything was loaded :-)

librz/include/rz_bin.h
thestr4ng3r: C

XVilka (approved):
Looks good as is already. Good job!

PR #1101, opened 2021-05-03
Add new RzBinRelocStorage and fix reloc source/target confusion

thestr4ng3r:
This was `RzBinReloc` before:
```c
typedef struct rz_bin_reloc_t {
	...
	ut64 vaddr;
	ut64 paddr;
	...
} RzBinReloc;
```

`paddr` was the raw file address like in other places.
However `vaddr` was a different beast:
* When relocs were not patched (i.e. by default this is determined by whether rizin is run with or without `io.cache`), then it was the corresponding vaddr to `paddr`. I refer to this as the **reloc vaddr**.
* When relocs **were** patched, it contained the vaddr of the **target**, i.e. the value patched into the reloc, which is has a completely different meaning, and the original vaddr was lost. I refer to this as the **target vaddr**.

Both vaddr values are important to have, so this pr changes it to this:

```c
typedef struct rz_bin_reloc_t {
	...
	ut64 vaddr; ///< the vaddr where the value should be patched into
	ut64 paddr; ///< the paddr where the value should be patched into
	ut64 target_vaddr; ///< the target address that the patched reloc points to
	...
} RzBinReloc;
```

This however has some consequences. Before, the relocs were stored in an `RBTree`, keyed by the value in the `vaddr` member. Having such a tree structure for the reloc addr is a good fit because we want to query for relocs contained within a given instruction, which a hashtable for example can't do.
But actually we need efficient access for the target addrs too, which before then you only got after patching the relocs, but lost the other efficient access, so I had to rewrite the data structure here:

```c
/// Efficient storage of relocations to query by address
struct rz_bin_reloc_storage_t {
	RzBinReloc **relocs; ///< all relocs, ordered by their vaddr
	size_t relocs_count;
	RzBinReloc **target_relocs; ///< all relocs that have a valid target_vaddr, ordered by their target_vaddr. size is target_relocs_count!
	size_t target_relocs_count;
}; // RzBinRelocStorage
```

I removed the `RBTree` and used two sorted arrays instead because `RBTree` is mostly useful for dynamically changing contents, but these are immutable after creating them once so we can just use arrays here, which are way more compact.
Now we can have both infos at the same time.

The behavior now works like this (when relocs are patched, which soon they will always be by default):
At the reloc addr you get a flag called `reloc.<name>`:
```
0x080000a3 4 reloc.strcmp.0
```
also, you can see in the disassembly a `RELOC` comment (note that the instruction addr is a bit lower than the reloc addr, the reloc is inside the instruction):
```
            0x080000a2      call  strcmp                               ; RELOC 32 strcmp
```

At the target that the patched reloc points to, which is usually just an artificially created stub, you get a `reloc.target.<name>` flag and a `RELOC TARGET` comment:
```
0x08000ef3 4 reloc.target.strcmp
```
```
            ;-- strcmp:
            0x08000ef3      .qword 0x0000000000000000                  ; RELOC TARGET 32 strcmp
```

To be able to test all of this, I also had to fix some other things along the way:
* When there were multiple relocs with the same function name, which is very common especially in unlinked objects, you would get only one `reloc.` flag because the others were subsequently overwritten. Now you get `reloc.<name>`, `reloc.<name>.0`, `reloc.<name>.1`, ...
* Both flags now get a correct realname. Before it was pretty much identical to the name including the `reloc.` prefix, which is exactly what realnames should not be.
* I moved the replacement of a jumptarget in disassembly by a reloc symbol name below replacing by a flag name so flags are prioritized. We might want to remove the reloc-specific workaround later when relocs are always patched.

Most test changes are consequences of these fixes.
The `NAME=ELF: reloc flags FILE=bins/elf/analysis/filetime.c-clang-x64-O0.o` test gives a good overview of the overall relocs behavior.


librz/bin/bobj.c
```diff
@@ -17,20 +17,143 @@ RZ_API void rz_bin_mem_free(void *data) {
...
+		}
+	}
+	rz_pvector_sort(&sorter, reloc_cmp);
+	ret->relocs_count = rz_pvector_len(&sorter);
+	ret->relocs = (RzBinReloc **)rz_pvector_flush(&sorter);
+	rz_pvector_fini(&sorter);
+	rz_pvector_sort(&target_sorter, reloc_target_cmp);
+	rz_pvector_shrink(&target_sorter);
+	ret->target_relocs_count = rz_pvector_len(&target_sorter);
+	ret->target_relocs = (RzBinReloc **)rz_pvector_flush(&target_sorter);
```
p371jah, line 91: But why?

thestr4ng3r: Why what exactly?

p371jah: Why not just use RzPVector?

thestr4ng3r: It's not necessary anymore here since the `RzBinRelocStorage` is meant to be immutable. `RzPVector` is primarily useful when you dynamically edit the array.

thestr4ng3r:
Looks like I need to optimize the quicksort before this becomes usable. It seems to exhaust the stack on OpenBSD on `test/bins/pe/exe2pe.exe`.

librz/core/cbin.c
```diff
@@ -931,10 +938,46 @@ static char *construct_reloc_name(RZ_NONNULL RzBinReloc *reloc, RZ_NULLABLE cons
 	return rz_strbuf_drain(buf);
 }
 
-static void set_bin_relocs(RzCore *r, RzBinReloc *reloc, ut64 addr, Sdb **db, char **sdb_module) {
+static void reloc_set_flag(RzCore *r, RzBinReloc *reloc, const char *prefix, ut64 flag_addr) {
 	int bin_demangle = rz_config_get_i(r->config, "bin.demangle");
 	bool keep_lib = rz_config_get_i(r->config, "bin.demangle.libs");
 	const char *lang = rz_config_get(r->config, "bin.lang");
+	char *reloc_name = construct_reloc_name(reloc, NULL);
+	if (!reloc_name || !*reloc_name) {
```
wargio: Use RZ_STR_ISEMPTY or IS_NULLSTR

librz/bin/bobj.c
```diff
@@ -17,20 +17,157 @@ RZ_API void rz_bin_mem_free(void *data) {
...
+}
+
+/// Get the reloc with the lowest vaddr that starts inside the given interval
+RZ_API RzBinReloc *rz_bin_reloc_storage_get_reloc_in(RzBinRelocStorage *storage, ut64 vaddr, ut64 size) {
+	rz_return_val_if_fail(storage && size >= 1, NULL);
+	if (!storage->relocs) {
+		return NULL;
+	}
+	size_t i;
+#define CMP(x, y) (x > y->vaddr ? 1 : (x < y->vaddr ? -1 : 0))
```
ghost: Create a static function.

```
static int reloc_vaddr_cmp(...)
```

librz/bin/bobj.c
```diff
@@ -17,20 +17,157 @@ RZ_API void rz_bin_mem_free(void *data) {
...
+}
+
+/// Get a reloc that points exactly to vaddr or NULL
+RZ_API RzBinReloc *rz_bin_reloc_storage_get_reloc_to(RzBinRelocStorage *storage, ut64 vaddr) {
+	rz_return_val_if_fail(storage, NULL);
+	if (!storage->target_relocs) {
+		return NULL;
+	}
+	size_t i;
+#define CMP(x, y) (x > y->target_vaddr ? 1 : (x < y->target_vaddr ? -1 : 0))
```
ghost: same

librz/core/cbin.c
```diff
@@ -931,10 +938,46 @@ static char *construct_reloc_name(RZ_NONNULL RzBinReloc *reloc, RZ_NULLABLE cons
...
+			snprintf(flagname, sizeof(flagname), "reloc.%s", demname);
+		}
+	}
+	rz_name_filter(flagname, 0, true);
+	RzFlagItem *existing = rz_flag_get(r->flags, flagname);
+	if (existing && existing->offset == flag_addr) {
+		// Mostly important for target flags.
+		// We don't want hundreds of reloc.target.<fcnname>.<xyz> flags at the same location
+		return;
+	}
```
ghost, line 969: An alternative could be `reloc.target.<fcnname>.[a-z]` or in some case `reloc.target.<fcnname>.[acd]`

wargio (approved):

wargio:
> Looks like I need to optimize the quicksort before this becomes usable. It seems to exhaust the stack on OpenBSD on `test/bins/pe/exe2pe.exe`.

you can check the file size vs amount of relocs and sanitize read values

librz/core/cbin.c
XVilka: I think IS_NULLSTR is deprecated, better to use RZ_STR_ISEMPTY

XVilka (approved):
Looks good. Did you notice any performance difference? Also, it might be useful to add `rawmem` binary into the tests too https://github.com/rizinorg/rz-ghidra/issues/203

librz/include/rz_bin.h
```diff
@@ -739,6 +761,10 @@ RZ_IPI void rz_bin_section_free(RzBinSection *bs);
 RZ_API void rz_bin_info_free(RzBinInfo *rb);
 RZ_API void rz_bin_import_free(RzBinImport *imp);
 RZ_API void rz_bin_symbol_free(RzBinSymbol *sym);
+static inline bool rz_bin_reloc_has_target(RzBinReloc *reloc) {
+	return reloc->target_vaddr && reloc->target_vaddr != UT64_MAX;
```
ret2libc, line 765: I know this is done in many places all around the code base, but... maybe we should add a `has_target_vaddr`?

librz/bin/bobj.c
```diff
@@ -17,20 +17,157 @@ RZ_API void rz_bin_mem_free(void *data) {
...
+	}
+	if (ar->target_vaddr < br->target_vaddr) {
+		return -1;
+	}
 	return 0;
 }
 
-static void reloc_free(RBNode *rbn, void *user) {
-	free(container_of(rbn, RzBinReloc, vrb));
+static int reloc_target_cmp(const void *a, const void *b) {
```
ret2libc, line 52: mmm these comparisons are becoming quite common and big.

What about a

```c
if (ar->target_vaddr != br->target_vaddr) {
  return CMP(ar->target_vaddr, br->target_vaddr);
} else if (ar->vaddr != br->vaddr) {
  return CMP(ar->vaddr, br->vaddr);
} else if (ar->paddr !+ br->paddr) {
  return CMP(ar->paddr, br->paddr);
} else {
  return CMP(ar->type, br->type);
}
```

librz/bin/bobj.c
```diff
@@ -17,20 +17,157 @@ RZ_API void rz_bin_mem_free(void *data) {
...
+	if (ar->type > br->type) {
+		return 1;
+	}
+	if (ar->type < br->type) {
+		return -1;
+	}
+	return 0;
+}
+
+RZ_API RzBinRelocStorage *rz_bin_reloc_storage_new(RZ_OWN RzList *relocs) {
```
ret2libc, line 64: I think this is not `RZ_OWN`, is it? If it was, `relocs` should be stored somewhere in RzBinRelocStorage and at some point it should be freed.

ret2libc: Oh, this is because the elements within `relocs` are moved in another vector... I see. I think however we should either remove the elements from the `relocs` list or free the whole `relocs` list within this function. Otherwise it's a bit unclear what exactly this function owns and what not.

librz/bin/bobj.c
```diff
@@ -17,20 +17,157 @@ RZ_API void rz_bin_mem_free(void *data) {
...
+	ret->target_relocs = (RzBinReloc **)rz_pvector_flush(&target_sorter);
+	rz_pvector_fini(&target_sorter);
+	return ret;
+}
+
+RZ_API void rz_bin_reloc_storage_free(RzBinRelocStorage *storage) {
+	if (!storage) {
+		return;
+	}
+	if (storage->relocs) {
```
ret2libc: this is probably unnecessary

librz/bin/bobj.c
```diff
@@ -17,20 +17,157 @@ RZ_API void rz_bin_mem_free(void *data) {
...
+	ret->relocs = (RzBinReloc **)rz_pvector_flush(&sorter);
+	rz_pvector_fini(&sorter);
+	rz_pvector_sort(&target_sorter, reloc_target_cmp);
+	ret->target_relocs_count = rz_pvector_len(&target_sorter);
+	ret->target_relocs = (RzBinReloc **)rz_pvector_flush(&target_sorter);
+	rz_pvector_fini(&target_sorter);
+	return ret;
+}
+
+RZ_API void rz_bin_reloc_storage_free(RzBinRelocStorage *storage) {
```
ret2libc, line 96: what about `target_relocs` array?

librz/bin/bobj.c
ret2libc: This kind of function -1/1/0 for numbers should really go in rz_util.h instead of being explicited everywhere (not just in this PR..)

thestr4ng3r:
> > Looks like I need to optimize the quicksort before this becomes usable. It seems to exhaust the stack on OpenBSD on `test/bins/pe/exe2pe.exe`.
> 
> you can check the file size vs amount of relocs and sanitize read values

Yeah it reads outside the file and then tries to parse about 200000 relocs in a 2560 byte file, here is the fix: https://github.com/rizinorg/rizin/pull/1103
But we should still be able to handle that high amount of relocs in theory so I will look into improving the sorting and testing it separately.

ret2libc (commented):
Overall it LGTM.

What I don't like too much (but can be fixed sometime later) is that when patching is enabled rizin shows `reloc.target.<name>` as if it was just another "data" of 8 bytes... IDA shows:
```
extern:0000000000201038 ; size_t strlen(const char *s)
extern:0000000000201038                 extrn strlen:near       ; DATA XREF: .got:strlen_ptr↑o
```
which is imho a bit clearer. Anyway!

librz/bin/bobj.c
thestr4ng3r: Yes, the list should be freed in here, that was the initial intention.

librz/bin/bobj.c
thestr4ng3r: I've implemented it similarly now, please check.

librz/core/cbin.c
thestr4ng3r: What do you mean exactly? Usually you have many relocs pointing to the same target. Then we would like to have a single flag there called just `reloc.target.<fcnname>` without a suffix. (Suffixes might get added if there are multiple imported functions with the same name but then consequently different patch targets)

librz/include/rz_bin.h
thestr4ng3r: Do you mean calling it `rz_bin_reloc_has_target_vaddr` instead of `rz_bin_reloc_has_target`? Or do you mean adding a boolean member `has_target_vaddr`?

librz/bin/bobj.c
thestr4ng3r: Thanks, lsan apparently tricked me there.

thestr4ng3r:
> What I don't like too much (but can be fixed sometime later) is that when patching is enabled rizin shows `reloc.target.<name>` as if it was just another "data" of 8 bytes...

I agree, specifically on the data of 8 bytes part, since data doesn't really make much sense there, as in a practical execution there would be code. It's possible this data-assigning is just a leftover from the confusion and was initially only intended for the reloc source addrs.
However until we have a better solution for displaying I think having data there is not too bad since otherwise you see whatever gets disassembled from `0x00` bytes, which might be even worse.

librz/core/cbin.c
ghost: simple case:
from
```
reloc.target.strcmp.0
reloc.target.strcmp.1
reloc.target.strcmp.2
reloc.target.strcmp.3
```
to
```
reloc.target.strcmp.[0-3]
```
complexe case (possible ?)
from
```
reloc.target.strcmp.0
reloc.target.strcmp.1
reloc.target.strcmp.3
```
to
```
reloc.target.strcmp.[013]
```

ghost (approved):

XVilka (approved):

librz/core/cbin.c
thestr4ng3r: Ah I see, that's a nice idea. But I think it would become too complicated here, in particular because the target flags than should also be perfectly in sync with the source flags.

wargio (approved):

librz/include/rz_bin.h
ret2libc: The latter. But this is really more of a general thing... In many places we check if addr != 0 && addr != UT64_MAX, but by doing so we implicitly assume that 0 and UT64_MAX are never valid values... I wonder if this assumption is always true.

thestr4ng3r: I am quite sure it is not always true. I think we might want to think about a ghidra-like scheme where an address is always space + 64bit offset. Then the space variable could be directly re-used to tell whether it's an invalid address instead of adding another bool.

ret2libc: Yeah probably as first thing we should refactor all ut64 used as addr to a typedef or struct `RzAddr` that can be implemented as necessary.. Anyway, challenge for another day :D

PR #1102, opened 2021-05-05
Convert `dts` commands to newshell

officialcjunior:

**Detailed description**

Converts commands under `dts`: `dts+`, `dts-`, `dtst`, `dtsf`, `dtsm` commands to newshell.

**Test plan**

Everything except `dtsm` has unit tests now. I'm not sure whether we can add test
for this one as it prints a lot of information which is dependent on the machine.

```
[0x00400565]> dtsm
Reading 4096 byte(s) from 0x00600000...
Reading 8192 byte(s) from 0x7fae8e85f000...
Reading 12288 byte(s) from 0x7fae8ea22000...
Reading 45056 byte(s) from 0x7fae8ea25000...
Reading 8192 byte(s) from 0x7fae8ea71000...
Reading 135168 byte(s) from 0x7ffc033f4000...
/home/cjunior/Downloads/rz/rizin/test/bins/elf/analysis/calls_x64: 22706addda4447a4a46b9a173f7f30a40f7c6f4287f2bf1b89490dbf97665c70
unk0: 604c8c108fbd51a88eafa4c63e29b0536c65a4abb74ca64434550c61c3cb41d5
/usr/lib/libc-2.33.so: c028ab380aec5b54dcba7245776a82623c6fbfd4f5d39576500d4f249d7681aa
unk1: ca1861200cf6788937e8e6a31f1b082aec9145fcc0f7e417f3258e3c46339f86
/usr/lib/ld-2.33.so: 52a808c284ba793f87955d3d6a537e50a178ad6ad781c127e6cd94f2bec9e719
[stack]: 612eb34cdb4f709fb53ed885d8cc55d9009a0260e46a480fd1ec6ad0fc7acd58
```
See the name of the handler as well, I had went with 
`rz_cmd_debug_list_trace_session_mmap_handler` fearing that the name might get a bit long. 

Putting it as draft to see if the CI finds something out.
**Closing issues**

None

### DIFF
--- a/librz/core/cmd_debug.c
+++ b/librz/core/cmd_debug.c
@@ -4265,6 +4265,57 @@ RZ_IPI RzCmdStatus rz_cmd_debug_step_until_flag_handler(RzCore *core, int argc, 
 	return RZ_CMD_STATUS_OK;
 }
 
+RZ_IPI RzCmdStatus rz_cmd_debug_start_trace_session_handler(RzCore *core, int argc, const char **argv) {
+	if (rz_debug_is_dead(core->dbg)) {
+		eprintf("Cannot start session outside of debug mode, run ood?\n");
+		return RZ_CMD_STATUS_ERROR;
+	}
+	if (core->dbg->session) {
+		eprintf("Session already started\n");
+		return RZ_CMD_STATUS_ERROR;
+	}
+	core->dbg->session = rz_debug_session_new();
+	rz_debug_add_checkpoint(core->dbg);
+
+	return RZ_CMD_STATUS_OK;
+}
+
+RZ_IPI RzCmdStatus rz_cmd_debug_stop_trace_session_handler(RzCore *core, int argc, const char **argv) {
+	if (!core->dbg->session) {
+		eprintf("No session started\n");
+		return RZ_CMD_STATUS_ERROR;
+	}
+	rz_debug_session_free(core->dbg->session);
+	core->dbg->session = NULL;
+	return RZ_CMD_STATUS_OK;
+}
+
+RZ_IPI RzCmdStatus rz_cmd_debug_save_trace_session_handler(RzCore *core, int argc, const char **argv) {
+	if (!core->dbg->session) {
+		eprintf("No session started\n");
+		return RZ_CMD_STATUS_ERROR;
+	}
+	rz_debug_session_save(core->dbg->session, argv[1]);
+	return RZ_CMD_STATUS_OK;
+}
+
+RZ_IPI RzCmdStatus rz_cmd_debug_load_trace_session_handler(RzCore *core, int argc, const char **argv) {
+	if (core->dbg->session) {
+		rz_debug_session_free(core->dbg->session);
+		core->dbg->session = NULL;
+	}
+	core->dbg->session = rz_debug_session_new();
+	rz_debug_session_load(core->dbg, argv[1]);
+	return RZ_CMD_STATUS_OK;
+}
+
+RZ_IPI RzCmdStatus rz_cmd_debug_list_trace_session_mmap_handler(RzCore *core, int argc, const char **argv) {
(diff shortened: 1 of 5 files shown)


officialcjunior:
Right, so, the failed test is for `dts-`, which can be fixed.
I'm not sure why the other one fails. I'm trying to figure out now. 😌

XVilka (approved):
Looks good as is 👍

librz/core/cmd_debug.c
```diff
@@ -4265,6 +4265,56 @@ RZ_IPI RzCmdStatus rz_cmd_debug_step_until_flag_handler(RzCore *core, int argc,
 	return RZ_CMD_STATUS_OK;
 }
 
+RZ_IPI RzCmdStatus rz_cmd_debug_start_trace_session_handler(RzCore *core, int argc, const char **argv) {
+	if (rz_debug_is_dead(core->dbg)) {
+		eprintf("Cannot start session outside of debug mode, run ood?\n");
+		return RZ_CMD_STATUS_ERROR;
+	}
+	if (core->dbg->session) {
```
XVilka, line 4273: In the future we might have a simple API something like this:
```c
RZ_API bool rz_debug_session_is_active(RzDebug *dbg)
```
Like we have no `rz_debug_is_dead()`. It will be very helpful also for use from non-C bindings of the API.

test/db/archos/linux-x64/dbg_dts
```diff
@@ -20,3 +20,28 @@ EXPECT=<<EOF
...
+dcu main
+dts+
+ds
+ds
+ds
+ds
+dts-
+ds
+ds
+dt
```
yossizap: `dt` is unrelated here, you will get the same result after removing `dts+` and `dts-` since it's a different type of tracing(in general one of these should be renamed to avoid confusion).

You could test `dts-` by attempting to step back(`dsb`) after stopping the session.

yossizap (changes requested):

test/db/archos/linux-x64/dbg_dts
officialcjunior: Thanks for the reply.

Okay, that makes sense, but what should I `EXPECT` over here?  Running `dsb` after stopping the session returns  `Session has not started`, but apparently rz-test can't catch that

yossizap: Maybe it's in `EXPECT_ERR`? Either way, you can print PC and see that you didn't actually step back.

yossizap (approved):

test/db/archos/linux-x64/dbg_dts
officialcjunior: Sweet! Thanks for the really quick reply!

librz/core/cmd_descs/cmd_debug.yaml
```diff
@@ -51,4 +51,31 @@ commands:
...
+    - name: dts-
+      summary: Stop trace session
+      cname: cmd_debug_stop_trace_session
+      args: []
+    - name: dtst
+      cname: cmd_debug_save_trace_session
+      summary: Save trace sessions to disk
+      args:
+        - name: dir
+          type: RZ_CMD_ARG_TYPE_STRING
```
ret2libc, line 71: I think this and all other dir args should be FILE type, so you can benefit from autocompletion... If we want to be even more precise we should probably define a new DIR type

officialcjunior: Yeah, I thought about that, but it looks like only these two commands have a directory as an argument.
```
[0x00000000]> ?*~dir
| dtst <dir>           # Save trace sessions to disk
| dtsf <dir>           # Load trace sessions to disk
| ls [<arg0> [<arg1> ...]] # List files and directories
| m                    # Make directories and move files
[0x00000000]>
```
So, I guess, we can go with `FILE` then?

PR #1112, opened 2021-05-08
Expose one instruction disassembly API

XVilka:

**Detailed description**

Rizin already had function to disassemble one instruction with some additional `pd`-like substitution but it was marked as static which is problematic for users of the API who is bound to run the disassembly process differently from how it's done in `axt`/`axtj` commands. This change exposes it as the new API:

```c
RZ_API RZ_OWN char *rz_core_disasm_instruction(RzCore *core, t64 addr, ut64 reladdr, RZ_NULLABLE RzAnalysisFunction *fcn, bool color);
```

**Test plan**

CI is green

### DIFF
--- a/librz/core/cmd_analysis.c
+++ b/librz/core/cmd_analysis.c
@@ -6129,60 +6129,6 @@ static void cmd_analysis_ucall_ref(RzCore *core, ut64 addr) {
 	}
 }
 
-static char *get_op_ireg(void *user, ut64 addr) {
-	RzCore *core = (RzCore *)user;
-	char *res = NULL;
-	RzAnalysisOp *op = rz_core_analysis_op(core, addr, 0);
-	if (op && op->ireg) {
-		res = strdup(op->ireg);
-	}
-	rz_analysis_op_free(op);
-	return res;
-}
-
-static char *get_buf_asm(RzCore *core, ut64 from, ut64 addr, RzAnalysisFunction *fcn, bool color) {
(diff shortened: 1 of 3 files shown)


ret2libc (commented):
Why do you need this? Isn't it enough to have an API that disassembles one instruction given an address and another API that gives you xrefs of a given address?

I'm not sure this is very much re-usable either..

XVilka:
@ret2libc it handles `sub.var` and colors.  In general, I wanted 1:1 replica of what `axt` shows

ret2libc:
> @ret2libc it handles `sub.var` and colors. In general, I wanted 1:1 replica of what `axt` shows

I think the purpose of an API should be to let the user choose what he wants, how to show that, etc. For example, such API would be useless for things like Cutter, where one may want to change color of disassembly based on cutter settings, etc. etc. 

That said, I'm not against *this* change.. I just think we should try to bloat the exposed API too much, if possible.

Also, i'm not sure what `fcn A function that is located in the Xref source or target (optional)` means/should do. Why is it necessary?

XVilka:
@ret2libc I checked other available APIs and there is seems no other API to get a similar disassembly with hints and vars when available.

XVilka:
@ret2libc I made this function more generic, updated the signature and Doxygen documentation, please take a look again

ret2libc (approved):

wargio (approved):

PR #1114, opened 2021-05-08
Tests, Fixes and Simplifications for ihex and Sparse RzBuffer

thestr4ng3r:
**DO NOT SQUASH**

**Detailed description**

This primarily adds tests for ihex and then some simplifications to sparse `RzBuffer`, which ihex is the only user of (soon to be more) so current sparse buffer use-cases are well-tested and we won't get any regressions there.

The tests intentionally still leave exact the outcoming ihex file contents open, because they contain overlaps and are thus somewhat broken. Since I will introduce some major changes to the sparse buffers that will also make fixing this a lot easier, this should be done after that. But as written above, the primary point is not to introduce regressions.

There is also a little fix in here, namely the change from `ut16 addh1 = rbs->to >> 16;` to `ut16 addh1 = (rbs->to - 1) >> 16;`. This prevents a write failure when a chunk ends at exactly an extended address boundary.

**Test plan**

unit tests

### DIFF
--- a/test/unit/test_io_ihex.c
+++ b/test/unit/test_io_ihex.c
@@ -1,0 +1,277 @@
+// SPDX-FileCopyrightText: 2021 Florian Märkl <info@florianmaerkl.de>
+// SPDX-License-Identifier: LGPL-3.0-only
+
+#include <rz_io.h>
+#include "minunit.h"
+
+static const char *ihex_sample =
+	":020000021000EC\r\n"
+	":10010000214601360121470136007EFE09D2190140\r\n"
+	":100110002146017EB7C20001FF5F16002148011988\r\n"
+	":020000022000DC\r\n"
+	":10010000194E79234623965778239EDA3F01B2CAC7\r\n"
+	":100110003F0156702B5E712B722B732146013421E7\r\n"
+	":00000001FF\r\n";
+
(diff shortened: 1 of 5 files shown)


test/unit/test_io_ihex.c
```diff
@@ -0,0 +1,277 @@
...
+	// write far out in the open space
+	mu_assert_eq(rz_io_desc_size(desc), 0x20120, "desc size");
+	r = rz_io_write_at(io, 0x40000, (const ut8 *)"Arecibo", 7);
+	mu_assert_true(r, "write success");
+	// writing outside also resizes
+	mu_assert_eq(rz_io_desc_size(desc), 0x40007, "desc size");
+	r = rz_io_read_at(io, 0x3ffff, buf, sizeof(buf));
+	mu_assert_true(r, "read success");
+	mu_assert_memeq(buf, (const ut8 *)"\xff"
+					  "Arecibo",
```
XVilka, line 117: Poor Arecibo...

XVilka (approved):

test/unit/test_io_ihex.c
thestr4ng3r: rip :-(

PR #1115, opened 2021-05-08
RzJson: support simple JSON paths

### LINKED ISSUE #1108: Support JSON Path in `librz/util/json_parser.c`
To ease the access of corresponding key and array number if available, something like `[0].methods[2].flags[0]` where first `[0]` will take first element of the JSON array, then key `"methods"`, which contains an array in turn, then take third element from that array, search for `"flags"` key, which has an array as a value, and take a first element of that array.

Files to change:
- `librz/util/json_parser.c`
- `librz/include/rz_util/rz_json.h`

herrgahr:
add rz_json_get_path, which allows querying by paths like `[0].sub.children[0].thing`

 

**Detailed description**

Support simple JSON paths like the one above. Could be extended to support fancy things like `things[*].whatever`, but I don't know if this is needed and what the idiomatic API for this would look like in rizin land.

**Test plan**

Tests were added in `test_json.c`so:
ninja test

**Closing issues**

closes #1108

### DIFF
--- a/librz/util/json_parser.c
+++ b/librz/util/json_parser.c
@@ -388,14 +388,20 @@ RZ_API RzJson *rz_json_parse(char *text) {
 	return js.children.first;
 }
 
-RZ_API const RzJson *rz_json_get(const RzJson *json, const char *key) {
+// getter with explicit size parameter, since in rz_json_get_path our key is
+// not zero-terminated.
+static const RzJson *rz_json_get_len(const RzJson *json, const char *key, size_t keysize) {
 	RzJson *js;
 	for (js = json->children.first; js; js = js->next) {
-		if (js->key && !strcmp(js->key, key)) {
+		if (js->key && !strncmp(js->key, key, keysize)) {
 			return js;
 		}
 	}
 	return NULL;
+}
+
+RZ_API const RzJson *rz_json_get(const RzJson *json, const char *key) {
+	return rz_json_get_len(json, key, strlen(key));
 }
 
 RZ_API const RzJson *rz_json_item(const RzJson *json, size_t idx) {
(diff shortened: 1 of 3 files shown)


librz/util/json_parser.c
```diff
@@ -407,3 +413,55 @@ RZ_API const RzJson *rz_json_item(const RzJson *json, size_t idx) {
...
+			}
+			path = key + keysize;
+			break;
+		default:
+			//TODO report error
+			RZ_JSON_REPORT_ERROR("JSON path: unexpected char", path - 1);
+			return NULL;
+		}
+	}
+	// unreachable code
```
herrgahr: Would `rz_warn_if_reached` be the proper thing to use here?

herrgahr: Nevermind, I changed the control flow to be more obvious, removing this dead code completely

wargio:
what if we want to iterate all arrays ?

herrgahr:
> what if we want to iterate all arrays ?

do you mean wildcard indices like `things[*].whatever`?

as a first shot, I did the simple thing but this could be done, too.

What would you imagine the API to be like?

For more than one wildcard like `things[*].children[*].whatever` it could be done recursively (with a callback?) or we'd have to allocate&keep state and require the user to clean up afterwards.

Is there any precedence for a similar API in the codebase?

XVilka:
The original issue and use case didn't require wildcards, so it's fine as is, I think.

wargio:
agree

XVilka (approved):

librz/util/json_parser.c
```diff
@@ -407,3 +413,53 @@ RZ_API const RzJson *rz_json_item(const RzJson *json, size_t idx) {
...
+			++path;
+			js = rz_json_item(js, index);
+			if (!js) {
+				return NULL;
+			}
+			break;
+		case '.':
+			key = path;
+			for (keysize = 0; key[keysize]; ++keysize) {
+				if (strchr(".[", key[keysize]))
```
wargio: parenthesis

wargio (changes requested):
LGTM

librz/util/json_parser.c
herrgahr: done

XVilka (approved):

PR #1119, opened 2021-05-10
Rewrite Sparse RzBuffer for Efficient Reading

thestr4ng3r:
**DO NOT SQUASH**

**Detailed description**

The Sparse RzBuffer, until now only used in ihex, will be used for relocs patching. Before this, it was an unordered linked list of data chunks, so reading from it would require linearly iterating through them and finding the right chunk, which is of course quite bad for performance. Moreover, these chunks could overlap in some cases, causing weird results in ihex.

The new implementation is a sorted flat vector of chunks that never overlap and as soon as two chunks would touch, they will be merged. That means read access is O(log n) with very few indirection. Writing is more expensive if writes happen in the middle because the new chunk has to be inserted in between.
An alternative would have been to use an `RBTree` but I think memory efficiency and read performance is more important here.

This also fixes a small bug regarding writing data spanning more than 2 16bit pages in ihex. This bug was present before too but only harder to hit because chunks were never merged.

**Test plan**

unit tests. There are tests for the sparse buf, including fuzzing ones and for ihex.

### DIFF
--- a/librz/io/p/io_ihex.c
+++ b/librz/io/p/io_ihex.c
@@ -68,14 +68,14 @@ static int fw04b(FILE *fd, ut16 eaddr) {
 
 //write contiguous block of data to file; ret 0 if ok
 //max 65535 bytes; assumes a 04 rec was written before
-static int fwblock(FILE *fd, ut8 *b, ut32 start_addr, ut16 size) {
+static int fwblock(FILE *fd, ut8 *b, ut32 start_addr, ut32 size) {
 	ut8 cks;
 	char linebuf[80];
 	ut16 last_addr;
 	int j;
 	ut32 i; //has to be bigger than size !
 
-	if (size < 1 || !fd || !b) {
+	if (size < 1 || size > 0x10000 || !fd || !b) {
 		return -1;
 	}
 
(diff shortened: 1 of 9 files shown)


librz/io/p/io_ihex.c
```diff
@@ -68,14 +68,14 @@ static int fw04b(FILE *fd, ut16 eaddr) {
...
-static int fwblock(FILE *fd, ut8 *b, ut32 start_addr, ut16 size) {
+static int fwblock(FILE *fd, ut8 *b, ut32 start_addr, ut32 size) {
 	ut8 cks;
 	char linebuf[80];
 	ut16 last_addr;
 	int j;
 	ut32 i; //has to be bigger than size !
 
-	if (size < 1 || !fd || !b) {
+	if (size < 1 || size > 0x10000 || !fd || !b) {
```
XVilka, line 78: Unnecessary for this PR, but maybe makes sense to extract this and 0xffff as this constant -1

librz/io/p/io_ihex.c
```diff
@@ -333,63 +334,53 @@ static bool ihex_write(RzIODesc *desc, Rihex *rih) {
...
-
-		if (addh0 != addh1) {
-			//we cross a 64k boundary, so write in two steps
-			if (addh0 != addh_cur) {
-				addh_cur = addh0;
+	for (size_t i = 0; i < chunks_count; i++) {
+		const RzBufferSparseChunk *rbs = &chunks[i];
+		ut64 from = rbs->from;
+		while (from >> 16 != rbs->to >> 16) {
+			//we cross a 64k boundary, so write in multiple steps
```
XVilka: space between `/`and text please

XVilka (approved):
Looks good. Just please address two minor nit

librz/io/p/io_ihex.c
XVilka: Ok, forget that.

XVilka (approved):

Issue #70, opened 2020-11-05
Rewrite/remove GPL-only code from core Rizin parts


XVilka:
- [x] `bin/mangling/cxx` - moved to [libdemangle](https://github.com/rizinorg/rz-libdemangle) https://github.com/rizinorg/rizin/pull/1656
- [ ] `shlr/qnx` - written by madprogrammer and defragger (can we ask them to relicense?)
- [ ] libopcodes in various places, see https://github.com/rizinorg/rizin/issues/209

To prevent the breakage it would be required to cover these pieces by tests a bit more than they are currently now.

```
[i] ℤ rg "SPDX-License-Identifier: GPL"                                                                                                                                                                                           17:53:38 
binrz/rz-gg/shellcodes.c.src
2:// SPDX-License-Identifier: GPL-2.0-or-later

librz/asm/arch/z80/z80asm.h
3:// SPDX-License-Identifier: GPL-3.0-or-later

librz/asm/arch/riscv/riscv.h
2:// SPDX-License-Identifier: GPL-3.0-or-later

librz/asm/arch/z80/z80asm.c
4:// SPDX-License-Identifier: GPL-3.0-or-later

librz/asm/arch/z80/expressions.c
3:// SPDX-License-Identifier: GPL-3.0-or-later

librz/asm/arch/riscv/riscv.c
2:// SPDX-License-Identifier: GPL-3.0-or-later

librz/asm/arch/riscv/riscv-opc.h
2:// SPDX-License-Identifier: GPL-3.0-or-later

librz/asm/arch/riscv/riscv-opc.c
2:// SPDX-License-Identifier: GPL-3.0-or-later

librz/asm/arch/vax/vax.h
2:// SPDX-License-Identifier: GPL-3.0-or-later

librz/asm/arch/vax/vax-dis.c
2:// SPDX-License-Identifier: GPL-3.0-or-later

librz/asm/arch/arm/aarch64/sysdep.h
2:// SPDX-License-Identifier: GPL-3.0-or-later

librz/asm/arch/arm/aarch64/aarch64-opc.h
2:// SPDX-License-Identifier: GPL-3.0-or-later

librz/asm/arch/arm/aarch64/aarch64.h
2:// SPDX-License-Identifier: GPL-3.0-or-later

librz/asm/arch/arm/aarch64/aarch64-tbl.h
2:// SPDX-License-Identifier: GPL-3.0-or-later

librz/asm/arch/arm/aarch64/aarch64-dis.h
[...]

PR #1126, opened 2021-05-13
Rewrite rz-diff and internal diffing library

wargio:
# squash me

**Detailed description**

This PR introduces the Ratcliff/Obershelp Pattern Recognition algorithm in rizin, which llows to diff 2 arrays of any kind.

Resources:
 - https://en.wikipedia.org/wiki/Gestalt_Pattern_Matching
 - https://github.com/chrismwendt/diff-gestalt/blob/master/src/Data/Algorithm/Diff/Gestalt.hs
 - https://github.com/python/cpython/blob/3.9/Lib/difflib.py

```
The code for diffing is quite simple, given 2 arrays containing
data, you calculate the longest sequences of data that matches
between the two inputs; to do that you need to create a map in 
which you will store all the hits found within an array:
- as key, each single element of one of the arrays.
- as value, a list of all the locations of which each element 
  appears within the array itself.

Once this map is created, you will need to find the longest
subsequence that can be found in both arrays by using the hit-map.
then you remove that subsequence from the area of search, and
search again for the 2nd longest subsequence (excluding the area
of the first subsequence).
Then you keep doing this, till all areas and all matches have been
found.

Now that you know all the matching areas, you can generate a series
of steps/operations which can transform the first array into the
second one, by removing the non matching areas in the 1st array
and inserting the missing areas from the 2nd array.

Example:
   array_a = [A,B,C,D,E,F,G,H,I]
   array_b = [Y,Z,B,C,D,L,Z,N,H,I]

1: create map of hits and their positions:
	- hit_map(array_b) = {
	  B: [2]
	  C: [3]
	  D: [4]
	  H: [8]
	  I: [9]
	  L: [5]
	  N: [7]
	  Y: [0]
	  Z: [1,6]
	}

2: find all matching areas using the hit-map:
 - match_0 = [B,C,D] from array_a[1] to array_a[3] and from array_b[3] to array_b[4]
 - match_1 = [H,I]   from array_a[1] to array_a[8] and from array_b[8] to array_b[9]

3: create the steps to convert array_a in array_b
 - remove [A] at 0
 - insert [Y,Z] at 0
 - keep   [B,C,D] at 1
 - remove [E,F,G] at 4
 - insert [L,Z,N] at 4
 - keep   [H,I] at 8
```
This PR also introduces a new rz-diff (example: https://asciinema.org/a/QkEiUhRJNNhryuH5lIgkSAGnz) with the following capabilities:
```
Usage: rz-diff [options] <file0> <file1>
  -a [arch] specify architecture plugin to use (x86, arm, ..)
  -b [bits] specify register size for arch (16 (thumb), 32, 64, ..)
  -d [algo] compute edit distance based on the choosen algorithm:
              myers | Eugene W. Myers' O(ND) algorithm (no substitution)
              leven | Levenshtein O(N^2) algorithm (with substitution)
  -h        this help message
  -j        json output
  -q        quite output
  -v        show version information
  -A        compare virtual and physical addresses
  -C        show colors
  -T        show timestamp information
  -0 [cmd]  command to execute for file0 when option -t 'commands' is given.
            if -1 is not set, it will execute the same cmd for file1.
  -1 [cmd]  command to execute for file1 when option -t 'commands' is given.
  -t [type] compute the difference between two files based on its type:
              bytes      | compares raw bytes in the files (only for small files)
              lines      | compares text files
              functions  | compares functions found in the files
              classes    | compares classes found in the files
              command    | compares command output returned when executed in both files
                         | requires -0 <cmd> and -1 <cmd> is optional
              entries    | compares entries found in the files
              fields     | compares fields found in the files
              imports    | compares imports found in the files
              libraries  | compares libraries found in the files
              sections   | compares sections found in the files
              strings    | compares strings found in the files
              symbols    | compares symbols found in the files
```

**Test plan**

I have fixed the rz-test binary to use the new code to diff the output.
Also i added unit test and tests for the rz-diff

**Closing issues**

- Resolves `librz/util/bdiff.c` point described in issue #70


binrz/rz-test/rz-test.c
```diff
@@ -609,6 +617,7 @@ static RzThreadFunctionRet worker_th(RzThread *th) {
 }
 
 static void print_diff(const char *actual, const char *expected, bool diffchar, const char *regexp) {
+	/*
```
XVilka: Just remove this completely?

XVilka:
Looks like it broke a bunch of unrelated tests:
```
[XX] C:\projects\rizin\test\db\cmd\types Basic union test
ANSICON=1 RZ_NOPLUGINS=1 rizin -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -Qc 'e asm.bytes=true
aa
td "union Books {char  title[50];char  author[50]; char  subject[100];};"
tu
tu Books
s main
tl Books 0x00177f88
afv~Books
s 0x000006d2
pd 1
' bins/elf/union_sample
-- stdout
-- stderr
[ ] Analyze all flags starting with sym. and entry0 (aa)[[x] Analyze all flags starting with sym. and entry0 (aa)
[XX] C:\projects\rizin\test\db\cmd\types tp with varname
ANSICON=1 RZ_NOPLUGINS=1 rizin -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -Qc 'td "struct Books {char  title[50];char  author[50]; char  subject[100];};"
aa
s main
aei
aeim
aecu 0x000006d5
?e
tp Books var_e0h
' bins/elf/struct_sample
-- stdout
-- stderr
[ ] Analyze all flags starting with sym. and entry0 (aa)[[x] Analyze all flags starting with sym. and entry0 (aa)
[XX] C:\projects\rizin\test\db\cmd\types to test.h
ANSICON=1 RZ_NOPLUGINS=1 rizin -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -Qc 'to bins/other/test.h
t~char,double,float,int,long,short,size,uid,void # basic universal types
ts~addr,date,dox,name
t addr
' =
-- stdout
```

librz/diff/diff.c
```diff
@@ -0,0 +1,827 @@
...
+	const void *a;
+	const void *b;
+	ut32 a_size;
+	ut32 b_size;
+	HtPP *b_hits;
+	MethodsInternal methods;
+};
+
+RZ_API ut32 rz_diff_hash_data(const ut8 *buffer, ut32 size) {
+	ut32 h = 5381;
```
XVilka, line 104: Please use `RZ_NULLABLE const ut8 *buffer` then

librz/diff/diff.c
```diff
@@ -0,0 +1,827 @@
...
+struct rz_diff_t {
+	const void *a;
+	const void *b;
+	ut32 a_size;
+	ut32 b_size;
+	HtPP *b_hits;
+	MethodsInternal methods;
+};
+
+RZ_API ut32 rz_diff_hash_data(const ut8 *buffer, ut32 size) {
```
XVilka: Also a Doxygen description, please

librz/diff/diff.c
```diff
@@ -0,0 +1,827 @@
...
+		if (!rz_list_append(list, NUM2PTR(i))) {
+			RZ_LOG_ERROR("rz_diff_set_b: cannot append index to list\n");
+			return false;
+		}
+	}
+
+	return true;
+}
+
+RZ_API RZ_OWN RzDiff *rz_diff_bytes_new(const ut8 *a, ut32 a_size, const ut8 *b, ut32 b_size, RzDiffIgnoreByte ignore) {
```
XVilka: Could `a` and `b` be NULL? Please mark them properly then (`RZ_NONNULL`)

librz/diff/diff.c
```diff
@@ -0,0 +1,827 @@
...
+		return NULL;
+	}
+	if (!set_b(diff, b, b_size)) {
+		rz_diff_free(diff);
+		return NULL;
+	}
+	return diff;
+}
+
+RZ_API RZ_OWN RzDiff *rz_diff_lines_new(const char *a, const char *b, RzDiffIgnoreLine ignore) {
```
XVilka: Same here, also Doxygen

librz/diff/diff.c
```diff
@@ -0,0 +1,827 @@
...
+		return NULL;
+	}
+	if (!set_b(diff, b_lines, rz_list_length(b_lines))) {
+		rz_diff_free(diff);
+		return NULL;
+	}
+	return diff;
+}
+
+RZ_API RZ_OWN RzDiff *rz_diff_generic_new(const void *a, ut32 a_size, const void *b, ut32 b_size, RzDiffMethods *methods) {
```
XVilka: And here, especially having the descriptive difference between those functions.

librz/diff/diff.c
```diff
@@ -0,0 +1,827 @@
...
+	return 0;
+}
+
+/**
+ * \brief generates a list of matching blocks
+ *
+ * Generates a list of matching blocks that are found in both inputs.
+ * If non are found it returns a match result with size of 0
+ * */
+RZ_API RzList *rz_diff_matches_new(RzDiff *diff) {
```
XVilka: Please also mark the ownership of the generated list.

librz/diff/diff.c
```diff
@@ -0,0 +1,827 @@
...
+	op->b_beg = b_beg;
+	op->b_end = b_end;
+}
+
+/**
+ * \brief Generates a list of steps needed to go from A to B
+ *
+ * Generates a list of opcodes that are needed to convert A to B.
+ * */
+RZ_API RzList *rz_diff_opcodes_new(RzDiff *diff) {
```
XVilka: Same here.

XVilka: Also, since you marked the type of the element in the list in some function below, please do it in the rest of the functions too.

librz/diff/lines_diff.c
```diff
@@ -0,0 +1,76 @@
...
+static void line_stringify(const char *a_elem, RzStrBuf *sb) {
+	rz_strbuf_set(sb, a_elem);
+}
+
+static void line_free(RzList *array) {
+	rz_list_free(array);
+}
+
+static const MethodsInternal methods_lines = {
+	.elem_at /*  */ = (RzDiffMethodElemAt)line_elem_at,
```
XVilka, line 70: Are those `/* */` necessary? Maybe better to just remove them?

XVilka (changes requested):
Overall, AMAZING job, as always. Apart from the fixes required for test, please address my comments, mostly style nitpicks.

librz/diff/lines_diff.c
wargio: tbh, i like having the data aligned, but it is not needed.

binrz/rz-test/rz-test.c
wargio: i actually forgot to fix this.

librz/diff/diff.c
```diff
@@ -0,0 +1,827 @@
+// SPDX-FileCopyrightText: 2021 RizinOrg <info@rizin.re>
+// SPDX-FileCopyrightText: 2021 deroad <wargio@libero.it>
+// SPDX-License-Identifier: LGPL-3.0-only
+
+/*
+ * Ratcliff/Obershelp Pattern Recognition algorithm applied to generic data
```
thestr4ng3r: It is possible to add documentation for a specific file in Doxygen: https://www.doxygen.nl/manual/commands.html#cmdfile
Maybe that would be appropritate here.

librz/main/rz-diff.c
```diff
@@ -1,1305 +1,1398 @@
...
-	if (as != bs) {
-		rz_cons_printf("...\n");
+
+	if (!rz_core_bin_update_arch_bits(core)) {
+		rz_diff_error("cannot set architecture with bits\n");
+		goto rz_diff_load_file_with_core_fail;
+	}
+
+	if (rz_list_empty(rz_bin_get_sections(core->bin))) {
+		rz_config_set_i(core->config, "io.va", false);
```
thestr4ng3r, line 458: What is the exact reason for this? I think you want to query for whether the file has maps, rather than sections. To avoid using the deprecated global `rz_bin_get_*` functions, take the binfile from the `cfile`.
The same goes for all the other bin functions used below, try to pass through an `RzBinFile *` or `RzBinObject *`.

thestr4ng3r (changes requested):

librz/main/rz-diff.c
wargio: how i do this for the maps?

thestr4ng3r: If you want to check if the binfile has bin-maps, check whether `cfile->binfiles[0]->o->maps` is not empty after loading. These are the maps that instruct core how to create io maps.
But then again, the question is what this is actually trying to achieve. I guess if there is no virtual mapping for the file, it should just use file file as-is. But iirc in such a case (which is when the bin-maps mentioned above are empty), core already maps the entire file as a single `RzIOMap` so switching `io.va` to false would change some internals but you would get the same results. So I think this needs some clarification/docs and tests for both cases.

librz/diff/distance.c
```diff
@@ -0,0 +1,105 @@
+// SPDX-FileCopyrightText: 2017 Fangrui Song <i@maskray.me>
+// SPDX-FileCopyrightText: 2016 NikolaiHampton <nikolaih@3583bytesready.net>
+// SPDX-License-Identifier: LGPL-3.0-only
+#include <rz_diff.h>
+#include <rz_util/rz_assert.h>
+
+// Eugene W. Myers' O(ND) diff algorithm
```
XVilka: Maybe convert this comment into Doxygen description too, since it's here anyway...

librz/diff/distance.c
```diff
@@ -0,0 +1,105 @@
...
+	//Clean up output on loop exit (purely aesthetic)
+	if (distance) {
+		*distance = di;
+	}
+	if (similarity) {
+		*similarity = length ? 1.0 - (double)di / length : 1.0;
+	}
+	return true;
+}
+
```
XVilka, line 60: Same here. While everyone and their dog knows what Levenstein distance is, some parameters explanation might be helpful for the API users.

librz/diff/unified_diff.c
```diff
@@ -0,0 +1,261 @@
...
+		if (len > 0 && p[len - 1] == '\n') {
+			newline = true;
+		}
+	}
+	pj_ks(pj, "value", rz_strbuf_get(&tmp));
+	pj_end(pj);
+	rz_strbuf_fini(&tmp);
+}
+
+RZ_API RZ_OWN char *rz_diff_unified_text(RZ_NONNULL RzDiff *diff, RZ_NULLABLE const char *from, RZ_NULLABLE const char *to, bool show_time, bool color) {
```
XVilka, line 135: Seems this is high-level API, would be **really** cool to have doxygen too ;)

librz/diff/unified_diff.c
```diff
@@ -0,0 +1,261 @@
...
+	rz_list_free(groups);
+	return rz_strbuf_drain(sb);
+
+rz_diff_unified_text_fail:
+	rz_strbuf_free(sb);
+	rz_list_free(groups);
+	return NULL;
+}
+
+RZ_API RZ_OWN PJ *rz_diff_unified_json(RZ_NONNULL RzDiff *diff, RZ_NULLABLE const char *from, RZ_NULLABLE const char *to, bool show_time) {
```
XVilka, line 203: Same here

librz/main/rz-diff.c
```diff
@@ -1,1305 +1,1454 @@
...
+		goto rz_diff_load_file_with_core_fail;
 	}
+
+	bfile = core_get_file(cfile);
+	if (!bfile) {
+		rz_diff_error("cannot get architecture with bits\n");
+		goto rz_diff_load_file_with_core_fail;
+	}
+
+	RzList *sections = rz_bin_maps_of_file_sections(bfile);
```
thestr4ng3r: `rz_bin_maps_of_file_sections()` is not correct here, it is a special function that bin plugins can use. Here, you should be able to use just the already populated `maps` member of `RzBinObject`.

librz/main/rz-diff.c
```diff
@@ -1,1305 +1,1454 @@
...
+		"              imports    | compares imports found in the files\n"
+		"              libraries  | compares libraries found in the files\n"
+		"              sections   | compares sections found in the files\n"
+		"              strings    | compares strings found in the files\n"
+		"              symbols    | compares symbols found in the files\n"
+		"");
+}
+
+static bool rz_diff_is_file(const char *file) {
+	//const char p = NULL;
```
XVilka: Remove this commented code please

librz/main/rz-diff.c
```diff
@@ -1,1305 +1,1454 @@
...
 }
 
-void print_bytes(const void *p, size_t len, bool big_endian) {
-	size_t i;
-	for (i = 0; i < len; i++) {
-		ut8 ch = ((ut8 *)p)[big_endian ? (len - i - 1) : i];
-		if (write(1, &ch, 1) != 1) {
-			break;
-		}
+/* This is terrible because can eat a lot of memory */
```
XVilka, line 298: Nice to know. Maybe add `FIXME:` so it will be easier to spot/find?

librz/main/rz-diff.c
```diff
@@ -1,1305 +1,1454 @@
...
 
-static void dump_cols_hexii(ut8 *a, int as, ut8 *b, int bs, int w) {
-	bool spacy = false;
-	ut32 sz = RZ_MIN(as, bs);
-	ut32 i, j;
-	int ctx = DUMP_CONTEXT;
-	int pad = 0;
-	if (!a || !b || as < 0 || bs < 0) {
-		return;
+static bool rz_diff_file_open(DiffFile *dfile, const char *filename) {
```
XVilka, line 481: probably worth to add `rz_return_val_if_fail()` here

thestr4ng3r (changes requested):
Unless I missed something, except for this, it lgtm now.

librz/main/rz-diff.c
```diff
@@ -1,1305 +1,1454 @@
...
+
+	rz_diff_free(diff);
+	rz_diff_file_close(&dfile_a);
+	rz_diff_file_close(&dfile_b);
+	free(a_buffer);
+	free(b_buffer);
+	return result;
+}
+
+RZ_API int rz_main_rz_diff(int argc, const char **argv) {
```
XVilka, line 1416: Maybe we should introduce `RZ_MAIN` to distinguish "main" functions from the rest of the API

XVilka (changes requested):

librz/main/rz-diff.c
wargio: that can't happen

librz/main/rz-diff.c
wargio: actually this is limited by a check on the file size

wargio: `MEGABYTE(5)` is max buffer for this

librz/main/rz-diff.c
wargio: i do not think this is needed.

librz/main/rz-diff.c
```diff
@@ -1,1305 +1,1448 @@
...
+		goto rz_diff_load_file_with_core_fail;
 	}
+
+	bfile = core_get_file(cfile);
+	if (!bfile) {
+		rz_diff_error("cannot get architecture with bits\n");
+		goto rz_diff_load_file_with_core_fail;
+	}
+
+	if (rz_list_empty(bfile->o->sections)) {
```
thestr4ng3r:
```suggestion
	if (rz_list_empty(bfile->o->maps)) {
```

XVilka (approved):
LGMT. Time to merge, if everyone else agrees 👍

thestr4ng3r (approved):

PR #1160, opened 2021-05-22
Deprecate RzSyscallPort and move the existing sysregs to SDB

officialcjunior:

**Detailed description**

`ioports.c` housed `RzSyscallPort`s, which were hardcoded definitions of sysregs which were iteratively searched and
served using `rz_syscall_get_io` and other APIs. These hardcoded values are moved to SDB in this PR.

This PR moves all of those definitions (only Atmega16's 8bit processor actually exists) to SDB, which follows a format like this:

```
SPH=reg
SPH.address=0x3e
SPH.comment=Stack higher bits SP8 SP10
```
Introduced `rz_sysreg_get()` to which you can pass on the type (`mmio/reg`) and the offset to get the port. This is used
where these ports are displayed as comments in `disasm.c`

The MMIO ports and registers are firstly filled into a new struct/module named `RzSysregItem` just after parsing and added to a hash table defined inside `RzSysregDB` where it's paired with the address or the offset where it is found. In `rz_sysreg_get` we are actually searching for the `RzSysregItem::comment` in this hashtable and returns when found. The name of the register is also saved as `RzSysregItem::name` for future endeavors (to be added as flags and such)

**Test plan**
~Although I could return the right values while testing it by passing the ports manually from disasm.c, it doesn't show up as comments yet somehow. There is a test for this, but it is failing now. I am working on it now.~

A representative test for the IO ports is added on one of the AVR binaries.
**Closing issues**

None


wargio (approved):
LOVE IT

librz/syscall/syscall.c
```diff
@@ -287,27 +277,22 @@ RZ_API RzList *rz_syscall_list(RzSyscall *s) {
...
-RZ_API const char *rz_syscall_get_io(RzSyscall *s, int ioport) {
+/**
+ * \brief Searches the sysregs SDB database inside librz/reg/d for the type (mmio/reg) and
+ * 		  returns its value as a string
+ *
+ * \param s reference to RzSyscall
+ * \param type reference to the type of key: mmio/reg
+ * \param port reference to the io port
+ */
+RZ_API const char *rz_sysreg_get(RzSyscall *s, const char *type, ut64 port) {
```
XVilka, line 413: Overall looks good, but I suggest to load SDB once, during the binary loading of the matching architecture, e.g. in the hashtable and query it instead during the analysis. This way, in case of many queries it will work much faster, especially for complex analysis loops.

XVilka (changes requested):

librz/syscall/syscall.c
```diff
@@ -84,6 +79,27 @@ static inline bool sysregs_reload_needed(RzSyscall *s, const char *arch, int bit
...
+ * 		  loads up the sysregs SDB database
+ *
+ * \param s reference to RzSyscall
+ * \param arch reference to value of asm.arch
+ */
+RZ_API bool rz_sysreg_set_arch(RzSyscall *s, const char *arch) {
+	char *dbName = rz_str_newf(RZ_JOIN_2_PATHS("reg", "%s-%s-%d"),
+		arch, s->cpu, s->bits);
+	if (dbName) {
+		if (!load_sdb(&s->db, dbName)) {
```
XVilka: I suggest to change the SDB usage here to the hashtable instead.

XVilka (changes requested):

librz/include/rz_syscall.h
```diff
@@ -24,10 +24,16 @@ typedef struct rz_syscall_item_t {
...
 
-typedef struct rz_syscall_port_t {
-	int port;
-	const char *name;
-} RzSyscallPort;
+typedef struct rz_sysreg_item_t {
+	char *type;
+	char *name;
+	char *comment;
+	ut64 address;
```
XVilka: The `address` is probably redundant, since it's the key in the hashtable.

librz/syscall/syscall.c
```diff
@@ -9,32 +9,65 @@
...
+ * \brief Frees an RzSyscall type
+ */
 RZ_API void rz_syscall_free(RzSyscall *s) {
 	if (s) {
 		if (s->refs > 0) {
 			s->refs--;
 			return;
 		}
-		sdb_free(s->srdb);
+		//	sdb_free(s->srdb);
```
XVilka: Just remove this

XVilka (changes requested):

librz/include/rz_syscall.h
officialcjunior: Actually, `address` is one of the values, right? I had changed the format of the SDB, resembling the one from types.

https://github.com/rizinorg/rizin/pull/1160/files#diff-cbadf4a774e67c5b7b27c4c6355eb978b19137269cdcb46b55789d0c7300a187R5-R7

officialcjunior: Hey, you are right about `address` being redundant. 

But, I'm not sure about how I'm going to parse it inside `get_sysreg_type` and bring it back to `sdb_load_sysregs`, which is where it's getting inserted into the hashtable.

officialcjunior: Combined the functionality of `get_sysreg_type` into `sdb_load_sysregs` and eliminated the redundant variable.

librz/syscall/syscall.c
```diff
@@ -43,6 +74,16 @@ RZ_API void rz_syscall_free(RzSyscall *s) {
 	}
 }
 
+/**
+ * \brief Frees an RzSysregItem type
+ */
+RZ_API void rz_sysreg_item_free(RZ_NONNULL RzSysregItem *s) {
```
XVilka: Usually,  in `_free()` functions we allow an argument to be NULL and just return if it is.

XVilka (changes requested):
So far looks good, just one minor nitpick.

XVilka (approved):

wargio (approved):

librz/syscall/syscall.c
```diff
@@ -287,27 +402,21 @@ RZ_API RzList *rz_syscall_list(RzSyscall *s) {
...
-	if (!s->db) {
+	if (!s->srdb) {
 		return NULL;
 	}
-	const char *key = sdb_fmt("%s,%" PFMT64d, type, num);
-	return sdb_const_get(s->db, key, 0);
+	RzSysregItem *item = ht_up_find(s->srdb->port, port, NULL);
+	if (item)
+		return item->comment;
+	return NULL;
```
officialcjunior, line 421: I missed to compare the `type` here. What I mean is, right now the `comment` gets returned whether the type is mmio or io.

I'll fix it one of the upcoming patches.

wargio: yes, remember that we are committing in dev, therefore is normal to do this.

PR #1161, opened 2021-05-22
Refactor mach0 Rebasing and Stripping to use RzBinVirtualFile

thestr4ng3r:
**DO NOT SQUASH**

**Detailed description**

"arm64e" Mach-O binaries, as found on arm macOS and later iOS, are not read directly but through a proxy that strips away some info so we get nice pointers. See also the comment in `mach0_rebase.c`
Previously, this was achieved by permanently swapping out the io plugin's read function pointer, but now that virtual files exist it can be implemented without such hacks.

**Test plan**

I've added dedicated tests for all kinds of reads that can appear in such a binary before the changes.


thestr4ng3r:
set to draft because dyldcache also needs tests for this

librz/bin/format/mach0/mach0_rebase.c
```diff
@@ -0,0 +1,179 @@
...
+ * In particular, we strip away additional info stored inside of pointers in the binary so we get
+ * the raw pointers out for convenient analysis.
+ */
+
+#include "mach0.h"
+
+#define IS_PTR_AUTH(x) ((x & (1ULL << 63)) != 0)
+#define IS_PTR_BIND(x) ((x & (1ULL << 62)) != 0)
+
+static void rebase_buffer(struct MACH0_(obj_t) * obj, ut64 off, ut8 *buf, int count) {
```
XVilka, line 22: Please check for `obj` and `buf` "nullness"

XVilka (changes requested):
Looks good, less "hackish" now.

wargio (approved):
i do not like the fact that there are struct around, but looks good.

librz/bin/format/mach0/mach0.c
```diff
@@ -2214,6 +2214,25 @@ static bool __isDataSection(RzBinSection *sect) {
 	return false;
 }
 
+RzList *MACH0_(get_virtual_files)(RzBinFile *bf) {
```
XVilka, line 2217: Dont you need to check here that `bf` is also not NULL?

librz/bin/format/mach0/mach0.c
```diff
@@ -2214,6 +2214,25 @@ static bool __isDataSection(RzBinSection *sect) {
...
+		vf->buf = MACH0_(new_rebasing_and_stripping_buf)(obj);
+		vf->buf_owned = true;
+		vf->name = strdup(MACH0_VFILE_NAME_REBASED_STRIPPED);
+		rz_list_push(ret, vf);
+	}
+	return ret;
+}
+
 RzList *MACH0_(get_maps)(RzBinFile *bf) {
 	struct MACH0_(obj_t) *bin = bf->o->bin_obj;
```
XVilka, line 2239: Same here

XVilka (approved):
Looks good.

librz/bin/format/mach0/mach0.c
thestr4ng3r: Doesn't hurt at least.

PR #1162, opened 2021-05-22
Set the `size` member of `RzBaseType` struct and notify the user if it's greater than the blocksize

### LINKED ISSUE #317: Using `tp` when sizeof(type) > block size results in zeros in values
### Work environment

| rizin -v full output, **not truncated** (mandatory)  | https://github.com/rizinorg/rizin/commit/06f23ea202cb31ba9b9591461c849dece10329a2

When using tp with a type that is greater than block size, all fields that are printed beyond the block boundary are silently set to 0.

### Expected behavior

I would have expected Rizin to have at least printed a warning that `sizeof(type)` is greater than the block boundary and therefore those fields have an incorrect value.

Another alternative is for rizin to temporarily set the block size to match that of the type being printed so all fields have the correct value.

### Actual behavior

Silently prints the wrong values
```
[0x00000000]> tp foobar
 a : 0x00000000 = 175704697
 b : 0x00000004 = 175704697
 c : 0x00000008 = 2681
 d : 0x0000000c = 0
[0x00000000]> 
```

Create a binary file:
```
$ yes | dd of=/tmp/test bs=1 count=1024
```
Load that file into rizin, set a block size smaller than a type and print:
```
rizin /tmp/test
[0x00000000]> b 10
[0x00000000]> "td struct foobar{uint32_t a; uint32_t b; uint32_t c; uint32_t d;}"
[0x00000000]> tp foobar
 a : 0x00000000 = 175704697
 b : 0x00000004 = 175704697
 c : 0x00000008 = 2681
 d : 0x0000000c = 0
[0x00000000]> 
```
If block size is increased correct values are returned:
```
[0x00000000]> b 16
[0x00000000]> tp foobar
 a : 0x00000000 = 175704697
 b : 0x00000004 = 175704697
 c : 0x00000008 = 175704697
 d : 0x0000000c = 175704697
[0x00000000]> 
```
by @hexagonal-sun

theopechli:

**Detailed description**

This PR sets the `size` member of the `RzBaseType` struct and if the user is using an inappropriate blocksize, the user is suggested to use the correct blocksize.

**Test plan**

**Closing issues**

closes #317

### DIFF
--- a/librz/type/base.c
+++ b/librz/type/base.c
@@ -81,6 +81,8 @@ static RzBaseType *get_enum_type(RzTypeDB *typedb, const char *sname) {
 	}
 	free(members);
 
+	base_type->size = base_type->enum_data.cases.len * base_type->enum_data.cases.elem_size;
+
 	return base_type;
 
 error:
@@ -144,6 +146,8 @@ static RzBaseType *get_struct_type(RzTypeDB *typedb, const char *sname) {
 	}
 	free(sdb_members);
 
+	base_type->size = base_type->struct_data.members.len * base_type->struct_data.members.elem_size;
+
 	return base_type;
 
 error:
@@ -198,6 +202,8 @@ static RzBaseType *get_union_type(RzTypeDB *typedb, const char *sname) {
 		sdb_aforeach_next(cur);
 	}
 	free(sdb_members);
+
+	base_type->size = base_type->union_data.members.len * base_type->union_data.members.elem_size;
 
 	return base_type;
 
--- a/librz/core/cmd_type.c
+++ b/librz/core/cmd_type.c
@@ -114,6 +114,11 @@ static void type_list_c_all_nl(RzCore *core) {
 }
 
 static RzCmdStatus type_format_print(RzCore *core, const char *type, ut64 address) {
+	RzBaseType *base_type = rz_type_db_get_base_type(core->analysis->typedb, type);
+	if (base_type->size > core->blocksize) {
+		rz_cons_printf(" -- Type '%s' is of greater size than blocksize. "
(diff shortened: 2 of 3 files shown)


librz/type/base.c
```diff
@@ -81,6 +81,8 @@ static RzBaseType *get_enum_type(RzTypeDB *typedb, const char *sname) {
 	}
 	free(members);
 
+	base_type->size = base_type->enum_data.cases.len * base_type->enum_data.cases.elem_size;
```
XVilka, line 84: It doesn't work like that. Enum size is either `uint32_t`  or `uint64_t` depending on how it's defined.

librz/type/base.c
```diff
@@ -199,6 +203,8 @@ static RzBaseType *get_union_type(RzTypeDB *typedb, const char *sname) {
 	}
 	free(sdb_members);
 
+	base_type->size = base_type->union_data.members.len * base_type->union_data.members.elem_size;
```
XVilka, line 206: Union size is also works differently, it's something like `RZ_MAX()`. See https://github.com/rizinorg/rizin/pull/1047/files#diff-f936f8d2d463f6bcab3c0c7bf452143da7ae6542ddb3bee7ae2e281cb948918dR660

XVilka (changes requested):

thestr4ng3r:
Thanks for the pr! These `size` members though are only meant to be meaningful for atomic base types, despite being a bit unfavourably placed. To get the size of any base type, there should instead be a function that calculates the whole size on the fly, which is basically `rz_type_db_get_bitsize()` but it should eventually be refactored to use the API.

Regarding #317, I think `pf` should be fixed to work with any block size currently set, so the user does not have to deal with this at all.

PR #1174, opened 2021-05-30
rz-test diff: Highlight in-line difference

kazarmy:

**Detailed description**

This pr uses the `diff-so-fancy` way (I think) to highlight in-line differences in a diff. The `diff-so-fancy` way basically scans the line from both left and right until it finds the first and last character that doesn't match. Highlighting only happens if:

1. The number of deleted lines is the number of inserted lines.
2. Either the first or last line character matches.

Sample output:

diff1
diff2

**Test plan**

Mangled test for output above (replace first test of `test/db/cmd/cmd_i` with this test and eyeball the output):

```
NAME=iS query
FILE=bins/elf/analysis/main
CMDS=iS size/sort:vaddr/cols/name
EXPECT=<<EOF
[Sections]

vaddr      name
---------------
0x00000000 
0x006006e0 .init_array
ax006006e8 .fini_arrar
0x006006f0 .jct
ax006008c8 .got
0x0040031e .gnu.version
0x004005b4 .fini
0x00600910 .bss
000600900 .data
004005c0 .rodata
0x00400348 .rela.dyn
0x004003a8 .ini
x00400260 .gnu.hash
0x00400200 .interpj
10x0040021c .note.ABI-tag
0x00400328 .gnu.version_r
0x00000000 .comment
0x0040023c .note.gnu.build-id
0x006008d0 .got.plt
0abc4005d0 .eh_frame_hdr
004002e0 .dynstr
0d004003d0 .plt
0x00400360 .rela.plt
0x00400280 .dynsym
0x00000000 .debug_abbrev
0x00000000 .debug_ranges
0x00000000 .debug_str
0x00600910 .bss
0x00000000 .debug_aranges
0x00400608 .eh_frame
0x000000000 .debug_line
0x00000000000 .shstrtab
000000000 .debug_info
0x00400410 .text
0x006006f8 .dynamic
0x00000 .strtab
0x0000000 .symtab

EOF
RUN
```

**Closing issues**

...

### DIFF
--- a/librz/diff/unified_diff.c
+++ b/librz/diff/unified_diff.c
@@ -4,13 +4,39 @@
 
 #include <rz_cons.h>
 
-#define Color_RANGE  Color_BBLUE
-#define Color_INSERT Color_BGREEN
-#define Color_DELETE Color_BRED
+#define Color_RANGE    Color_BBLUE
+#define Color_INSERT   Color_BGREEN
+#define Color_DELETE   Color_BRED
+#define Color_BGINSERT "\x1b[48;5;22m"
+#define Color_BGDELETE "\x1b[48;5;52m"
 
-#define FAST_MOD2(x, y)    ((x) & (y - 1))
-#define FAST_MOD64(x)      FAST_MOD2(x, 64)
-#define DIFF_COLOR(prefix) (prefix == '+' ? Color_INSERT : (prefix == '-' ? Color_DELETE : ""))
+#define FAST_MOD2(x, y)      ((x) & (y - 1))
+#define FAST_MOD64(x)        FAST_MOD2(x, 64)
+#define DIFF_COLOR(prefix)   (prefix == '+' ? Color_INSERT : (prefix == '-' ? Color_DELETE : ""))
+#define DIFF_BGCOLOR(prefix) (prefix == '+' ? Color_BGINSERT : (prefix == '-' ? Color_BGDELETE : ""))
+
+static inline ut32 count_newlines(RzDiff *diff, const void *array, st32 beg, st32 end) {
+	RzDiffMethodElemAt elem_at = diff->methods.elem_at;
+	RzDiffMethodStringify stringify = diff->methods.stringify;
+	int len = 0;
+	ut32 count = 0;
+	const char *p;
+	const void *elem;
+	RzStrBuf tmp;
+
+	for (st32 i = beg; i < end; ++i) {
+		rz_strbuf_init(&tmp);
+		elem = elem_at(array, i);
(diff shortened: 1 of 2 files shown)


XVilka (approved):
Looks good. Does it work on Windows too?
cc @wargio 

I also wonder if it's possible to autotest too.

kazarmy:
> I also wonder if it's possible to autotest too.

Probably possible by converting the escape character to `\0x1b`, but I'd like to do this in a different pr.

I probably will do some refactoring as well.

> Looks good. Does it work on Windows too?

Yes, even with `cmd`.

diff_cmd

librz/diff/unified_diff.c
```diff
@@ -176,12 +382,22 @@ RZ_API RZ_OWN char *rz_diff_unified_text(RZ_NONNULL RzDiff *diff, RZ_NULLABLE co
...
 			}
-			if (op->type == RZ_DIFF_OP_DELETE || op->type == RZ_DIFF_OP_REPLACE) {
+			if (op->type == RZ_DIFF_OP_DELETE) {
 				diff_unified_append_data(diff, diff->a, op->a_beg, op->a_end, sb, '-', color);
 			}
-			if (op->type == RZ_DIFF_OP_INSERT || op->type == RZ_DIFF_OP_REPLACE) {
+			if (op->type == RZ_DIFF_OP_INSERT) {
 				diff_unified_append_data(diff, diff->b, op->b_beg, op->b_end, sb, '+', color);
 			}
+			if (op->type == RZ_DIFF_OP_REPLACE) {
```
wargio, line 391: How about adding else? `if delete else if insert else if replace` ? we do not need to be if if if since that replace has its own scope

wargio (approved):

wargio:
@kazarmy you can easily add tests for this. just add a new test with the 2 files, then add a test as you would normally run but pipe the result to a file, and execute the test. then just copy the result into the test area. no encoding is needed.

librz/diff/unified_diff.c
kazarmy: Ok will do next.

kazarmy:
> ... copy the result into the test area

You mean copy the result into source code?

wargio:
> > ... copy the result into the test area
> 
> You mean copy the result into source code?

no. just copy it into the test.

librz/diff/unified_diff.c
kazarmy: Done in #1177.

kazarmy:
Oh ok #1177 ... I thought I had to do a unit test.

PR #1180, opened 2021-05-31
[WIP] Refactor: Deprecate rz_str_dup and use rz_str_new everywhere

### LINKED ISSUE #1176: Resolve the clash between `strdup()`, `rz_str_dup()`, and `rz_str_new()`.
We have 3 similar functions:
- `strdup()` from libc
- `rz_str_dup()`
- `rz_str_new()`

Some places use `strdup()` directly, while some - either `rz_str_dup()` or `rz_str_new()`. We should decide which one to choose and stick to the one, I think.

```c
// Returns a new heap-allocated copy of str.
// XXX what's the diff with rz_str_dup ?
RZ_API char *rz_str_new(const char *str) {
    return str ? strdup(str) : NULL;
}

// TODO: deprecate?
RZ_API char *rz_str_dup(char *ptr, const char *string) {
    char *str = rz_str_new(string);
    free(ptr); // in case ptr == string
    return str;
}
```

See `librz/util/str.c`

rbrishabh:
Fixes #1176

Hi @XVilka ! I'll take up this issue!  You can give an initial review when you have some time!
I'll add more commits ASAP!

Thanks :)

### DIFF
--- a/librz/util/str.c
+++ b/librz/util/str.c
@@ -699,7 +699,6 @@ RZ_API const char *rz_str_nstr(const char *s, const char *find, int slen) {
 }
 
 // Returns a new heap-allocated copy of str.
-// XXX what's the diff with rz_str_dup ?
 RZ_API char *rz_str_new(const char *str) {
 	return str ? strdup(str) : NULL;
 }
@@ -845,12 +844,6 @@ RZ_API char *rz_str_ndup(const char *ptr, int len) {
 	return out;
 }
 
-// TODO: deprecate?
-RZ_API char *rz_str_dup(char *ptr, const char *string) {
-	char *str = rz_str_new(string);
-	free(ptr); // in case ptr == string
-	return str;
-}
 
 RZ_API char *rz_str_prepend(char *ptr, const char *string) {
 	int slen, plen;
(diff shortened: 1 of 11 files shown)


librz/util/str.c
```diff
@@ -699,7 +699,6 @@ RZ_API const char *rz_str_nstr(const char *s, const char *find, int slen) {
 }
 
 // Returns a new heap-allocated copy of str.
-// XXX what's the diff with rz_str_dup ?
 RZ_API char *rz_str_new(const char *str) {
```
wargio, line 702: `rz_str_new` takes 1 argument but you just renamed all `rz_str_dup` to that but not changed the arguments.

wargio (changes requested):

rbrishabh:
yes, I see! adding more commits later today! thank you

librz/util/str.c
ret2libc: also, the old `rz_str_dup` was also freeing the `ptr` argument, which means that now all things will leak.

wargio:
Hello @rbrishabh do you plan on working on this or i can close this PR?

rbrishabh:
hi @wargio ! Due to some extra work, I'm not able to give enough time to this currently! I'm sorry!
I'll close this! :)

Issue #691, opened 2021-02-23
Add a helper script for quick finding of command handlers


thestr4ng3r:
Like `git grep "<cmdname>"` previously, but walking the yaml.
I'm not yet exactly sure what the best way will be to find the function in the C code once the name is known. But in the worst case a `git grep cmd_xyz_handler` should be useful enough already.

ret2libc:
Something like `git grep -nWG "^[^[:blank:]].*rz_tasks_break_handler(" *.c` could work.

Issue #489, opened 2021-01-29
Refactor output modes to use enum


XVilka:
Currently you can often see in the code the following pattern:
```c
RZ_API void rz_core_task_list(RzCore *core, int mode) {
    if (mode == 'j') {
         ...
    }
}
```
In some cases it looks like that:
```c
RZ_API int rz_core_disasm_pde(RzCore *core, int nb_opcodes, int mode) {
    if (mode == RZ_MODE_JSON) {
        ...
    }
}
```
Sometimes there is no "mode" word, and it looks like that:
```
RZ_API void rz_analysis_xrefs_list(RzAnalysis *analysis, int rad) {
    ...
    if (rad == 'j') {
        ...
    }
```

I recommend to search for `int mode` and `char mode` patterns in the code.

It prevents catching some errors at the compilation time and reduces the readability.
Should be refactored to use enum instead, across the whole Rizin code.

After everything ported, the following hardcoded values should be removed from `librz/include/rz_types.h`:
```c
#define RZ_MODE_PRINT     0x000
#define RZ_MODE_SET       0x002
#define RZ_MODE_SIMPLE    0x004
#define RZ_MODE_JSON      0x008
#define RZ_MODE_SIMPLEST  0x020
```

See, for example:
```
librz/util/range.c
314:int rz_range_list(RRange *rgs, int rad) {

librz/io/io_cache.c
93:RZ_API bool rz_io_cache_list(RzIO *io, int rad) {

librz/main/rz-bin.c
63:static RzOutputMode rad2outputmode(int rad) {
471:static int rabin_do_operation(RzBin *bin, const char *op, int rad, const char *output, const char *file) {
622:static void __listPlugins(RzBin *bin, const char *plugin_name, PJ *pj, int rad) {
656:    int rad = 0;

librz/core/core_private.h
123:RZ_IPI bool rz_core_debug_reg_list(RzCore *core, int type, int size, bool skip_covered, PJ *pj, int rad, const char *use_color);

librz/core/cdebug.c
300:RZ_IPI bool rz_core_debug_reg_list(RzCore *core, int type, int size, bool skip_covered, PJ *pj, int rad, const char *use_color) {

librz/core/cmd/cmd_flag.c
734:    int rad;
759:static void print_function_labels_for(RzAnalysisFunction *fcn, int rad, PJ *pj) {
772:static void print_function_labels(RzAnalysis *analysis, RzAnalysisFunction *fcn, int rad) {

librz/core/canalysis.c
[...]

PR #1182, opened 2021-06-01
Refactor `dm`

MalhotraPulak:

**Detailed description**
Plan:
- Refactor dm to use  RZ_API
- Refactor dmh to use RZ_API
- Make heap viewer in Cutter using the refactored commands

Currently:
- Add more tests for the commands
- Expose API to cutter

**Test plan**

CI is green

...

**Closing issues**

...


librz/core/cmd_debug.c
```diff
@@ -1443,24 +1443,145 @@ static bool get_bin_info(RzCore *core, const char *file, ut64 baseaddr, PJ *pj,
 	return true;
 }
 
+// dm
+RZ_IPI RzCmdStatus rz_cmd_debug_list_maps_handler(RzCore *core, int argc, const char **input, RzOutputMode mode) {
```
XVilka: I wonder if it's better to use `argv` instead of `input` as it's done in other places, to keep consistency. cc @ret2libc

XVilka (changes requested):

XVilka:
I suggest to port [that JSON PR](https://github.com/rizinorg/rizin/pull/1172) on top of this one once it's finished.

librz/core/cmd_debug.c
MalhotraPulak: yeah i will change that to `argv`

librz/core/cmd_debug.c
```diff
@@ -1443,370 +1443,618 @@ static bool get_bin_info(RzCore *core, const char *file, ut64 baseaddr, PJ *pj,
 	return true;
 }
 
-static int cmd_debug_map(RzCore *core, const char *input) {
+// dm
+RZ_IPI RzCmdStatus rz_cmd_debug_list_maps_handler(RzCore *core, int argc, const char **argv, RzOutputMode mode) {
+	if (argc == 1) {
+		char str[5] = "";
+		if (argc == 1) {
```
XVilka: It's already true here.

librz/core/cmd_debug.c
```diff
@@ -1443,370 +1443,618 @@ static bool get_bin_info(RzCore *core, const char *file, ut64 baseaddr, PJ *pj,
...
+			strcpy(str, argv[0] + 2);
```
XVilka: Are you sure `strcpy()` is necessary here? Maybe it will work without?

XVilka (changes requested):

librz/core/cmd_descs/cmd_debug.yaml
```diff
@@ -19,63 +19,176 @@ commands:
             cname: cmd_debug_step_until
             summary: Step until <addr>
             args:
-            - name: addr
-              type: RZ_CMD_ARG_TYPE_RZNUM
+              - name: addr
```
XVilka, line 22: Are you sure these formatting changes are necessary? I recommend to recheck using `yamllint`.

XVilka (changes requested):

librz/core/cmd_descs/cmd_debug.yaml
MalhotraPulak: My ide autoformatted it,  i will run it  through `yamllint`

librz/core/cmd_debug.c
MalhotraPulak: yeah it is not needed ill remove it

librz/core/cmd_debug.c
MalhotraPulak: yes i will remove the check

MalhotraPulak:
@XVilka I found `dmq`  and `dmq.` command in the codebase which were not mentioned in the help menu. I implemented that in the new shell. Can you tell what summary should I write for them?  
Also, in the help menu there are `dms` and `dms-` command mentioned but not implemented anywhere. Should I add them to new shell as commands that do nothing or leave them?

XVilka:
@PulakIIIT for
- `dms` - remove them
- `dmq` means `quiet`. Judging by the code it's the same as `dm` but with less verbose information. Just add ` RZ_OUTPUT_MODE_QUIET` in `dm` YAML and call it from here.

Moreover, I checked the code of `rz_debug_map_list()`. It is wrong nowadays because:
- Generally, when we have function ending in `_list()` it means it *returns* list of something, e.g. as `RzList *`. When the function *prints* something, we call it `_print()` instead.
- Generally, we prefer to avoid `RzCons` usage or printing outside of the `librz/core`.

Thus, I suggest to do the following:
- Add function `rz_debug_map_print()` in `librz/core/cdebug.c`
- Make function `rz_debug_map_list()` return `RzList *`
- Move all print helpers of old `rz_debug_map_list()` together with it into `librz/core/cdebug.c`
- Move `rz_debug_map_list_visual()` with helpers also into `librz/core/cdebug.c`
- Also I suggest to use `RzOutputMode` instead of `char` in that funciton like described in https://github.com/rizinorg/rizin/issues/489

MalhotraPulak:
@Xvilka I have moved the function `rz_debug_map_list` and related functions from `dmap.c` to `librz/core/cdebug.c` and renamed `rz_debug_map_list` to `rz_debug_map_print` since it mostly focuses on printing. I also refactored `rz_debug_map_print` to use `RzOutputMode`. Since I renamed that function at the moment there is no function  `rz_debug_map_list` which returns an `RzList *` as you mentioned in your first point. Is it necessary to have `rz_debug_map_list` in the codebase?

librz/core/cdebug.c
```diff
@@ -639,3 +639,270 @@ RZ_API RzCmdStatus rz_core_debug_plugins_print(RzCore *core, RzCmdStateOutput *s
...
+							    : "| %s0x%08" PFMT64x "%s %s %s\n";
+			dbg->cb_printf(fmtstr, color_prefix, map->addr_end, color_suffix,
+				rz_str_rwx_i(map->perm), map->name);
+			last = map->addr;
+		}
+	}
+}
+
+RZ_API void rz_debug_map_list_visual(RzDebug *dbg, ut64 addr, const char *input, int colors) {
+	if (dbg) {
```
XVilka: Please reverse the condition, so there will be less nesting.

librz/core/cdebug.c
```diff
@@ -639,3 +639,270 @@ RZ_API RzCmdStatus rz_core_debug_plugins_print(RzCore *core, RzCmdStateOutput *s
...
+		int i;
+		for (i = 0; i < 2; i++) { // Iterate over dbg::maps and dbg::maps_user
+			RzList *maps = (i == 0) ? dbg->maps : dbg->maps_user;
+			if (maps) {
```
XVilka: Same here.

XVilka (changes requested):

XVilka:
>  Since I renamed that function at the moment there is no function rz_debug_map_list which returns an RzList * as you mentioned in your first point. Is it necessary to have rz_debug_map_list in the codebase?

Please add, you will need it for Cutter anyway.

librz/core/cdebug.c
```diff
@@ -885,24 +885,31 @@ static void print_debug_maps_ascii_art(RzDebug *dbg, RzList *maps, ut64 addr, in
...
+					// print_debug_map_ascii_art (dbg, map);
 				}
 			}
+		} else { // "dm=" Show all maps with a graph
+			print_debug_maps_ascii_art(dbg, maps, addr, colors);
 		}
 	}
 }
+
+RZ_API RzList *rz_debug_map_list(RzDebug *dbg, bool user_map) {
```
XVilka: I recommend to move this function to `librz/debug/dmap.c` instead, it doesn't need to be in the core.
Please also add a short Doxygen description.

XVilka (changes requested):

XVilka:
Please also rebase on top of the latest dev to resolve the conflicts.

librz/core/linux_heap_glibc.c
```diff
@@ -1494,544 +2061,336 @@ static void GH(print_heap_segment)(RzCore *core, MallocState *main_arena,
...
+		} else if (!strcmp(input, "large")) {
+			bin_format = RZ_HEAP_BIN_LARGE;
+		}
+	}
+	GH(print_main_arena_bins)
+	(core, m_state, main_arena, global_max_fast, bin_format, json);
+	free(main_arena);
+	return RZ_CMD_STATUS_OK;
+}
+#ifndef HEAP32
```
ret2libc: why this condition? Where is `rz_cmd_heap_chunks_print_handler` defined if this is not present?

librz/core/linux_heap_glibc.c
```diff
@@ -1494,544 +2061,336 @@ static void GH(print_heap_segment)(RzCore *core, MallocState *main_arena,
...
+		return rz_cmd_heap_chunk_print_handler_32(core, argc, argv);
+	}
+}
 
-	MallocState *main_arena = RZ_NEW0(MallocState);
-	if (!main_arena) {
-		return false;
+RZ_IPI RzCmdStatus rz_cmd_heap_chunks_graph_handler(RzCore *core, int argc, const char **argv) {
+	// RZ_OUTPUT_MODE_LONG_JSON mode workaround for graph
+	if (core->rasm->bits == 64) {
```
ret2libc: can't you just call `rz_cmd_heap_chunks_print_handler`?

librz/core/linux_heap_glibc.c
```diff
@@ -1494,544 +2061,336 @@ static void GH(print_heap_segment)(RzCore *core, MallocState *main_arena,
...
+		}
+	}
+	GH(print_main_arena_bins)
+	(core, m_state, main_arena, global_max_fast, bin_format, json);
+	free(main_arena);
+	return RZ_CMD_STATUS_OK;
+}
+#ifndef HEAP32
+RZ_IPI RzCmdStatus rz_cmd_heap_chunks_print_handler(RzCore *core, int argc, const char **argv, RzOutputMode mode) {
+	if (core->rasm->bits == 64) {
```
ret2libc: i'd suggest to define a small macro to avoid all these `if (core->rasm->bits == 64) ...._64 else ..._32`.

librz/core/linux_heap_glibc.c
```diff
@@ -1406,84 +1840,217 @@ static void GH(print_heap_segment)(RzCore *core, MallocState *main_arena,
...
+			rz_cons_printf(", free)\n");
+		}
+		addr = (ut64)pos->GH(next);
+	}
+	rz_list_free(arenas_list);
+	free(main_arena);
+	return RZ_CMD_STATUS_OK;
+}
+
+RZ_IPI RzCmdStatus GH(rz_cmd_heap_chunks_print_handler)(RzCore *core, int argc, const char **argv, RzOutputMode mode) {
```
ret2libc: I'd have this function with `GHT m_state` argument instead of argc/argv. That woulld make calling them "easier" for other parts of the code.

librz/core/cmd_debug.c
```diff
@@ -223,33 +221,33 @@ static const char *help_msg_dko[] = {
...
-	"dms", "[?] <id> <mapaddr>", "Take memory snapshot",
-	"dms-", " <id> <mapaddr>", "Restore memory snapshot",
-	"dmS", " [addr|libname] [sectname]", "List sections of target lib",
-	"dmS*", " [addr|libname] [sectname]", "List sections of target lib in rizin commands",
-	"dmL", " address size", "Allocate <size> bytes at <address> and promote to huge page",
-	//"dm, " rw- esp 9K", "set 9KB of the stack as read+write (no exec)",
-	"TODO:", "", "map files in process memory. (dmf file @ [addr])",
-	NULL
-};
+//static const char *help_msg_dm[] = {
```
ret2libc: remember to just remove these if not used.

librz/core/cmd_descs/cmd_heap_glibc.yaml
```diff
@@ -0,0 +1,86 @@
+---
+name: cmd_heap_glibc
+commands:
+  - name: dmh
```
ret2libc, line 4: it would be great if you could use `type: RZ_CMD_DESC_TYPE_ARGV_STATE`. You can see `rz_analysis_function_blocks_list_handler`(and others) and read a bit the doc in `rz_cmd.h` on how that changes things.

librz/core/linux_heap_glibc.c
ret2libc: I think all these `dm` subcommands should have a nice check to see if debugging is enabled and provide a nice message to users if not.

librz/core/cmd_debug.c
```diff
@@ -1443,376 +1436,408 @@ static bool get_bin_info(RzCore *core, const char *file, ut64 baseaddr, PJ *pj,
...
-		addr = rz_num_math(core->num, input + 2);
+	default:
+		rz_core_cmd_help(core, help_msg_dmi);
+		break;
+	}
+	return RZ_CMD_STATUS_OK;
+}
+
+// dmp
+RZ_IPI RzCmdStatus rz_cmd_debug_dmp_handler(RzCore *core, int argc, const char **argv) {
```
ret2libc: i'd suggest a better name e.g. `rz_debug_memory_permission_handler`

librz/core/cmd_debug.c
```diff
@@ -1443,376 +1436,408 @@ static bool get_bin_info(RzCore *core, const char *file, ut64 baseaddr, PJ *pj,
...
+	RzListIter *iter;
+	RzDebugMap *map;
+	ut64 addr = 0, size = 0;
+	int perms;
+	if (argc == 4) { // dmp <addr> <size> <perms>
```
ret2libc: these checks are not in sync with the optional/non-optional arguments defined in the .yaml. In particular, I see `dmp [<addr> [<size> <perms>]]`, but if i do `dmp` it says `Wrong number of arguments passed to `dmp`, see its help with `dmp?``.

librz/core/cmd_descs/cmd_debug.yaml
```diff
@@ -19,63 +19,191 @@ commands:
...
+            type: RZ_CMD_ARG_TYPE_NUM
+            optional: true
+          - name: perms
+            type: RZ_CMD_ARG_TYPE_STRING
+            optional: false
+      - name: dmL
+        summary: Allocate <size> bytes at <address> and promote to huge page
+        cname: cmd_debug_dmL
+        args:
+          - name: addr
```
ret2libc: I would avoid all `addr` arguments, in general, because I think `@ <addr>` modifier is already good enough. So I would make most commands just assume that they have to act on the current address (the user can always use `@ addr` to change it.

librz/core/cmd_descs/cmd_heap_glibc.yaml
```diff
@@ -0,0 +1,86 @@
...
+    type: RZ_CMD_DESC_TYPE_OLDINPUT
+    args:
+      - name: bin_num|bin_num:malloc_state
+        type: RZ_CMD_ARG_TYPE_STRING
+        optional: true
+  - name: dmhc
+    cname: cmd_heap_chunk_print
+    summary: Get info about heap chunk at <addr>
+    args:
+      - name: addr
```
ret2libc: also here, just remove this argument and assume current offset.

librz/core/cmd_descs/cmd_debug.yaml
MalhotraPulak: Should I add "use `@ <addr>` modifier" in the summary of the commands which use that or we just assume the user knows that?

librz/core/linux_heap_glibc.c
MalhotraPulak: `linux_heap_glibc.c` file has a recursive declaration of itself in line 10 
```C
#define HEAP32 1
#include "linux_heap_glibc.c"
#undef HEAP32
```
If I do not wrap these functions in a conditional statement then they would be declared twice.

ret2libc: Move them to another file then... `cmd_linux_heap_glibc.c` or something like that. I think it's better.

librz/core/cmd_debug.c
MalhotraPulak: `dmp` commands can be run in two ways:  
- `dmp <addr> <size> <perms>`
- `dmp <perms> `
How should the yaml file look like for input like this?

librz/core/cmd_descs/cmd_debug.yaml
ret2libc: Let's just assume the user knows that. I think all commands should assume this behaviour and we should try to make it consistent everywhere, so users know they can use this pattern everywhere.

librz/core/linux_heap_glibc.c
MalhotraPulak: @ret2libc Is this check good for debugging is enabled or you would suggest something else?
```C
	if (!core || !core->dbg || core->dbg->pid == -1) {
		rz_cons_printf("Rizin not in debugging mode. Run ood?\n");
		return RZ_CMD_STATUS_ERROR;
	}
```

ret2libc: Seems like there is `rz_debug_is_dead` that can be used instead. cc @yossizap

librz/core/cmd_debug.c
ret2libc: let's change the way this command works IMHO. @XVilka opinions?

For example `dmp <perms> [<size>]` (again, I would not have `<addr>` in the arguments because that can be changed with `@ <addr>`). WDYT?

MalhotraPulak: Yeah, this seems like a nice way to deal with the case.

librz/core/cmd_descs/cmd_debug.yaml
```diff
@@ -19,63 +19,185 @@ commands:
...
+        modes:
+          - RZ_OUTPUT_MODE_STANDARD
+          - RZ_OUTPUT_MODE_RIZIN
+          - RZ_OUTPUT_MODE_JSON
+          - RZ_OUTPUT_MODE_QUIET
+        args: []
+      - name: dm+
+        cname: cmd_debug_allocate_maps
+        summary: >
+          Allocate <size> bytes at <offset>
```
ret2libc: Adjust the description, because `<offset>` does not make sense here.

librz/core/cmd_descs/cmd_debug.yaml
```diff
@@ -19,63 +19,185 @@ commands:
...
+              - RZ_OUTPUT_MODE_STANDARD
+              - RZ_OUTPUT_MODE_RIZIN
+              - RZ_OUTPUT_MODE_JSON
+            args: []
+          - name: dmm.
+            cname: cmd_debug_current_modules
+            summary: List memory map of current module
+            args: []
+      - name: dm-
+        summary: Deallocate memory map at <offset>
```
ret2libc: maybe "at current offset"

librz/core/cmd_descs/cmd_debug.yaml
```diff
@@ -19,63 +19,185 @@ commands:
...
+      - name: dml
+        summary: Load contents of file into current map region
+        cname: cmd_debug_dml
+        args:
+          - name: file
+            type: RZ_CMD_ARG_TYPE_FILE
+            optional: false
+      - name: dmp
+        summary: >
+          Change page at <offset> with <size>, protection <perms>
```
```suggestion
          Change page at current offset with <size>, protection <perms>
```

librz/core/cmd_descs/cmd_debug.yaml
```diff
@@ -19,63 +19,185 @@ commands:
...
+          / Change dbg.map permissions to <perms>
+        cname: debug_memory_permission
+        args:
+          - name: perms
+            type: RZ_CMD_ARG_TYPE_STRING
+            optional: false
```
ret2libc: no need for this AFAIK. it's implicit.

librz/core/cmd_descs/cmd_debug.yaml
```diff
@@ -19,63 +19,185 @@ commands:
...
+          - name: size
+            type: RZ_CMD_ARG_TYPE_NUM
+            optional: true
+      - name: dmL
+        summary: Allocate <size> bytes at <offset> and promote to huge page
```
```suggestion
        summary: Allocate <size> bytes at current offset and promote to huge page
```

librz/core/cmd_descs/cmd_debug.yaml
```diff
@@ -19,63 +19,185 @@ commands:
...
+        summary: >
+          Change page at <offset> with <size>, protection <perms>
+          / Change dbg.map permissions to <perms>
+        cname: debug_memory_permission
+        args:
+          - name: perms
+            type: RZ_CMD_ARG_TYPE_STRING
+            optional: false
+          - name: size
+            type: RZ_CMD_ARG_TYPE_NUM
```
```suggestion
            type: RZ_CMD_ARG_TYPE_RZNUM
```

librz/core/cmd_descs/cmd_debug.yaml
```diff
@@ -19,63 +19,185 @@ commands:
...
+            optional: true
+      - name: dmL
+        summary: Allocate <size> bytes at <offset> and promote to huge page
+        cname: cmd_debug_dmL
+        args:
+          - name: size
+            type: RZ_CMD_ARG_TYPE_NUM
```
```suggestion
            type: RZ_CMD_ARG_TYPE_RZNUM
```

librz/core/cmd_descs/cmd_heap_glibc.yaml
```diff
@@ -0,0 +1,84 @@
...
+      - name: malloc_state
+        type: RZ_CMD_ARG_TYPE_RZNUM
+        optional: true
+  - name: dmha
+    cname: cmd_arena_print
+    summary: List all the arenas
+    args: []
+  - name: dmhb
+    cname: cmd_heap_bins_list_print
+    summary: >
```
ret2libc, line 23: Maybe we should think about a shorter summary? You can also have the `description: ` field with a longer description.

librz/core/cmd_descs/cmd_heap_glibc.yaml
```diff
@@ -0,0 +1,84 @@
...
+      Display all parsed Double linked list
+      of main_arena's or a particular arena bins instance.
+      Use dmhbg command to get a graph of the linked list
+    type: RZ_CMD_DESC_TYPE_OLDINPUT
+    args:
```
ret2libc, line 27: if this is OLDINPUT type you don't need to specify the arguments.
Also, if you choose later to convert this to newshell, i would split `bin_num` and `malloc_state`.

librz/core/cmd_descs/cmd_heap_glibc.yaml
```diff
@@ -0,0 +1,84 @@
...
+      - name: bin_num|bin_num:malloc_state
+        type: RZ_CMD_ARG_TYPE_STRING
+        optional: true
+  - name: dmhc
+    cname: cmd_heap_chunk_print
+    summary: Get info about heap chunk at offset
```
ret2libc:
```suggestion
    summary: Get info about heap chunk at current offset
```

librz/core/cmd_descs/cmd_heap_glibc.yaml
```diff
@@ -0,0 +1,84 @@
...
+    cname: cmd_heap_arena_bins_print
+    summary: >
+      Display state of bins in an arena. <bin_type>
+      can be tcache/fast/unsorted/small/large
+    modes:
+      - RZ_OUTPUT_MODE_STANDARD
+      - RZ_OUTPUT_MODE_JSON
+    args:
+      - name: bin_type
+        type: RZ_CMD_ARG_TYPE_STRING
```
ret2libc: see CHOICE arg type. There are other examples of this type in .yaml files that you can check to see how to use it.

librz/core/cmd_descs/cmd_heap_glibc.yaml
```diff
@@ -0,0 +1,84 @@
...
+        optional: true
+  - name: dmhf
+    cname: cmd_heap_fastbins_print
+    type: RZ_CMD_DESC_TYPE_OLDINPUT
+    summary: >
+      Display all parsed fastbins of main_arena's or a particular
+      arena fastbinY instance
+    args:
```
ret2libc, line 54: OLDINPUT, thus args is not needed.

librz/include/rz_core.h
```diff
@@ -770,6 +771,22 @@ RZ_API char *rz_core_sysenv_begin(RzCore *core, const char *cmd);
 RZ_API void rz_core_sysenv_end(RzCore *core, const char *cmd);
 
 RZ_API void rz_core_recover_vars(RzCore *core, RzAnalysisFunction *fcn, bool argonly);
+
+/* linux_heap_glibc.c */
```
ret2libc, line 777: The fact that the malloc implementations have to stay in RzCore seems a bit wrong, but definitely not something to consider in this PR.

However, do we really need these API exported  in the public API? Can we put them as RZ_IPI in core_private.h? If they have to stay in the public API they should be at least prefixed with `rz_core_`.

XVilka: I agree that ideally they should be moved to `librz/debug`, except configuration handling and command handlers. But, indeed, this is out of the scope of this PR.

XVilka: Also agree on keeping these APIs private.

librz/core/cmd_descs/cmd_heap_glibc.yaml
MalhotraPulak: I used `args` to show them in help menu of the `dmh` command

librz/core/cdebug.c
```diff
@@ -639,3 +639,273 @@ RZ_API RzCmdStatus rz_core_debug_plugins_print(RzCore *core, RzCmdStateOutput *s
...
+			{
+				char *name = (map->name && *map->name)
+					? rz_str_newf("%s.%s", map->name, rz_str_rwx_i(map->perm))
+					: rz_str_newf("%08" PFMT64x ".%s", map->addr, rz_str_rwx_i(map->perm));
+				rz_name_filter(name, 0, true);
+				dbg->cb_printf("f map.%s 0x%08" PFMT64x " 0x%08" PFMT64x "\n",
+					name, map->addr_end - map->addr + 1, map->addr);
+				free(name);
+			} break;
+			case RZ_OUTPUT_MODE_QUIET: // "dmq"
```
XVilka, line 737: Please remove commented code.

librz/core/cdebug.c
```diff
@@ -639,3 +639,273 @@ RZ_API RzCmdStatus rz_core_debug_plugins_print(RzCore *core, RzCmdStateOutput *s
...
+		if (!maps) {
+			continue;
+		}
+		RzListIter *iter;
+		RzDebugMap *map;
+		if (input[1] == '.') { // "dm=." Only show map overlapping current offset
+			dbg->cb_printf("TODO:\n");
+			rz_list_foreach (maps, iter, map) {
+				if (addr >= map->addr && addr < map->addr_end) {
+					// print_debug_map_ascii_art (dbg, map);
```
XVilka: Why this part is commented?

librz/core/cdebug.c
```diff
@@ -639,3 +639,273 @@ RZ_API RzCmdStatus rz_core_debug_plugins_print(RzCore *core, RzCmdStateOutput *s
...
+	}
+}
+
+RZ_API void rz_debug_map_list_visual(RzDebug *dbg, ut64 addr, const char *input, int colors) {
+	if (!dbg) {
+		return;
+	}
+	int i;
+	for (i = 0; i < 2; i++) { // Iterate over dbg::maps and dbg::maps_user
+		//		RzList *maps = (i == 0) ? dbg->maps : dbg->maps_user;
```
XVilka: Please remove this unnecessary commented code

librz/core/cmd_debug.c
```diff
@@ -270,12 +240,12 @@ static const char *help_msg_dmm[] = {
 	NULL
 };
 
-static const char *help_msg_dmp[] = {
-	"Usage:", "dmp", " Change page permissions",
-	"dmp", " [addr] [size] [perms]", "Change permissions",
-	"dmp", " [perms]", "Change dbg.map permissions",
-	NULL
-};
+//static const char *help_msg_dmp[] = {
```
XVilka: Just remove this

librz/core/cmd_debug.c
```diff
@@ -1443,376 +1377,465 @@ static bool get_bin_info(RzCore *core, const char *file, ut64 baseaddr, PJ *pj,
...
+			    (libname != NULL && (strstr(map->name, libname))))) {
+			baddr = map->addr;
+			char *res;
+			const char *file = map->file ? map->file : map->name;
+			char *name = rz_str_escape((char *)rz_file_basename(file));
+			char *filesc = rz_str_escape(file);
+			/* TODO: do not spawn. use RzBin API */
+			if (sectname) {
+				char *sect = rz_str_escape(sectname);
+				res = rz_sys_cmd_strf("env RZ_BIN_PREFIX=\"%s\" rz-bin %s-B 0x%08" PFMT64x " -S \"%s\" | grep \"%s\"", name, mode, baddr, filesc, sect);
```
XVilka, line 1673: @thestr4ng3r could you please suggest a proper API as a substitute for this command?

librz/core/linux_heap_glibc.c
```diff
@@ -188,7 +188,14 @@ static void GH(update_arena_without_tc)(GH(RzHeap_MallocState) * cmain_arena, Ma
 	main_arena->GH(max_system_mem) = cmain_arena->max_system_mem;
 }
 
-static bool GH(update_main_arena)(RzCore *core, GHT m_arena, MallocState *main_arena) {
+/**
+ * @brief Store the MallocState struct of an arena with base address m_arena in main_arena
```
XVilka: In other places we use `\brief` syntax instead of `@brief`. Please use it consistently. See https://github.com/rizinorg/rizin/blob/dev/DEVELOPERS.md#documentation

librz/core/linux_heap_glibc.c
```diff
@@ -391,19 +398,26 @@ static void GH(print_arena_stats)(RzCore *core, GHT m_arena, MallocState *main_a
...
+ * @param m_arena Store the location of main arena at this integer pointer
+ * @return True if a main arena was found else False
+ */
+RZ_API bool GH(rz_resolve_main_arena)(RzCore *core, GHT *m_arena) {
 	rz_return_val_if_fail(core && core->dbg && core->dbg->maps, false);
 
-	if (core->dbg->main_arena_resolved) {
-		return true;
-	}
+	//	if (core->dbg->main_arena_resolved) {
```
XVilka: Why this was commented? Is this still necessary? If not, lets remove the commented code. cc @thestr4ng3r @ret2libc

librz/core/linux_heap_glibc.c
```diff
@@ -391,19 +398,26 @@ static void GH(print_arena_stats)(RzCore *core, GHT m_arena, MallocState *main_a
 	PRINT_GA("}\n\n");
 }
 
-static bool GH(rz_resolve_main_arena)(RzCore *core, GHT *m_arena) {
+/**
+ * @brief Store the base address of main arena at m_arena
```
XVilka: Use `\` vs `@`

librz/core/linux_heap_glibc.c
```diff
@@ -1025,278 +1083,656 @@ static void GH(tcache_print)(RzCore *core, GH(RTcache) * tcache) {
...
+				pj_end(pj);
+			}
 		}
 	}
 }
 
-static void GH(print_tcache_instance)(RzCore *core, GHT m_arena, MallocState *main_arena, bool main_thread_only) {
-	rz_return_if_fail(core && core->dbg && core->dbg->maps);
+/**
+ * @brief Get a list of RzTcache objects which contain information about tcache.
```
XVilka: Same here

librz/core/linux_heap_glibc.c
```diff
@@ -1025,278 +1083,656 @@ static void GH(tcache_print)(RzCore *core, GH(RTcache) * tcache) {
...
+		}
+		free(heap_info);
+		free(ms);
+	} else {
+		PRINT_RA("This address is not part of the arenas\n");
+	}
+}
+
+/**
+ * @brief Get list of chunks of <bin_num> bin from NBINS array of an arena.
```
XVilka: Same here

librz/core/linux_heap_glibc.c
```diff
@@ -1025,278 +1083,656 @@ static void GH(tcache_print)(RzCore *core, GH(RTcache) * tcache) {
...
+	if (json) {
+		pj_end(pj);
+		pj_end(pj);
+		rz_cons_println(pj_string(pj));
+		pj_free(pj);
+	}
+}
+
+/**
+ * @brief Get a list of MallocState structs for all the arenas
```
XVilka: Same here

librz/core/linux_heap_glibc.c
```diff
@@ -1025,278 +1083,656 @@ static void GH(tcache_print)(RzCore *core, GH(RTcache) * tcache) {
...
 			}
+			// thread arenas
+			rz_list_append(arena_list, ta);
+		}
+	}
+	return arena_list;
+}
+
+/**
+ * @brief Get a list of all the heap chunks in an arena. The chunks are in form of a struct RzHeapChunkListItem
```
XVilka: Same here

librz/debug/dmap.c
```diff
@@ -360,3 +88,13 @@ RZ_API RzList *rz_debug_map_list_new(void) {
 	list->free = (RzListFree)rz_debug_map_free;
 	return list;
 }
+
+/**
+ * @brief Get RzList* of memory maps for the process currently being debugged
```
XVilka: Same here

librz/include/rz_core.h
XVilka: @PulakIIIT don't forget about this one.

librz/include/rz_heap_glibc.h
```diff
@@ -278,6 +276,17 @@ typedef enum rz_heap_bin_type {
...
+	ut64 addr; /* Base addr of the chunk */
+	ut64 size; /* Size of the chunk */
+	char *status; /* Status of the chunk, allocated/free/corrupted */
+} RzHeapChunkListItem;
+
+typedef struct rz_arena_list_item {
+	ut64 addr;
+	char *type;
+	MallocState *arena;
+} RzArenaListItem;
```
XVilka: Add a new line after that.

XVilka (changes requested):
Overall looks good, there are some minor changes required though before being ready to merge.

librz/core/cdebug.c
MalhotraPulak: This is code for `dm=.` command and it was not fully implemented (therefore the `TODO` printf) even before the refactor. I can just remove this part since this code is not reachable after shifting this command to the new shell.

librz/core/linux_heap_glibc.c
MalhotraPulak: During the refactoring, I tried to get the code with this statement uncommented but I was facing some bugs in which `m_arena` was not be initialized properly. Due to the nature of the refactor `GHT m_arena` is declared multiple times so we cannot use this hack to save performance anymore.

librz/core/cdebug.c
XVilka: Yes, I think it's better to just remove that. If you think this is a feature is worth adding - just open an issue. But better to remove this dead code and TODO printing.

librz/include/rz_core.h
MalhotraPulak: @XVilka @ret2libc I added them as public APIs as in the future I will need some of the malloc-related APIs when I implement the heap viewer in Cutter

librz/core/cmd_debug.c
```diff
@@ -1443,376 +1370,465 @@ static bool get_bin_info(RzCore *core, const char *file, ut64 baseaddr, PJ *pj,
...
-					res = rz_sys_cmd_strf("env RZ_BIN_PREFIX=\"%s\" rz-bin %s-B 0x%08" PFMT64x " -S \"%s\" | grep \"%s\"", name, mode, baddr, filesc, sect);
-					free(sect);
-				} else {
-					res = rz_sys_cmd_strf("env RZ_BIN_PREFIX=\"%s\" rz-bin %s-B 0x%08" PFMT64x " -S \"%s\"", name, mode, baddr, filesc);
+				char *newfile = NULL;
+				if (!rz_file_exists(file)) {
+					newfile = rz_file_temp("memlib");
+					if (newfile) {
+						file = newfile;
+						rz_core_cmdf(core, "wtf %s 0x%" PFMT64x " @ 0x%" PFMT64x " 2> %s",
```
XVilka: Just use `rz_file_dump()` here instead.

librz/core/cmd_debug.c
```diff
@@ -1443,376 +1370,465 @@ static bool get_bin_info(RzCore *core, const char *file, ut64 baseaddr, PJ *pj,
...
-	case '.': // "dm."
-		rz_debug_map_list(core->dbg, addr, input);
-		break;
-	case 'm': // "dmm"
-		if (!strcmp(input + 1, ".*")) {
-			cmd_debug_modules(core, ':');
-		} else
-			cmd_debug_modules(core, input[1]);
+	case '\0': // "dmi" alias of "dmm"
+		rz_core_cmd(core, "dmm", 0);
```
XVilka: Please use API here too, if possible.

XVilka (changes requested):

librz/include/rz_core.h
XVilka: Then I think you need better names for them. Something like `rz_debug_heap_` or `rz_heap_`. Currently it is quite inconsistent.

librz/core/cmd_debug.c
```diff
@@ -1443,376 +1370,465 @@ static bool get_bin_info(RzCore *core, const char *file, ut64 baseaddr, PJ *pj,
...
+}
+
+// dmm
+RZ_IPI RzCmdStatus rz_cmd_debug_modules_handler(RzCore *core, int argc, const char **argv, RzOutputMode mode) {
+	if (rz_debug_is_dead(core->dbg)) {
+		rz_cons_println("Debugging is not enabled. Run ood?");
+		return RZ_CMD_STATUS_ERROR;
+	}
+	if (!strcmp(argv[0] + 3, ".*")) {
+		cmd_debug_modules(core, ':');
```
XVilka: Could you also please tear `cmd_debug_modules` function apart and use the API instead of passing characters/string further?

XVilka: I can see that is uses `int mode` and matching characters instead. I suggest to extract `?` to drop the help, it should be provided by the newshell anyway, and use `RzOutputMode mode` directly.

librz/include/rz_core.h
ret2libc: Not that it really matters for the current state of things with regards to the API, but IMO even if you need them in the future it would be better to make them private for now and switch them to public only as you really need them (with proper unit tests if time/will permits).

Anyway, I agree for better names!

librz/core/cmd_debug.c
MalhotraPulak: @XVilka In [this commit](https://github.com/rizinorg/rizin/pull/1182/commits/59691f818c7edd0ba4b6e7726c7ac8ec49b26fe7) commit I used `rz_core_dump()`. I was not able to come up with a testcase which reaches this codeblock so can you just test this once.

librz/core/cmd_descs/cmd_debug.yaml
```diff
@@ -121,8 +121,11 @@ commands:
             args: []
           - name: dmm.
             cname: cmd_debug_current_modules
+            modes:
+              - RZ_OUTPUT_MODE_STANDARD
+              - RZ_OUTPUT_MODE_RIZIN
             summary: List memory map of current module
-            args: []
+            args: [ ]
```
XVilka: Unnecessary space

librz/include/rz_core.h
MalhotraPulak: At the moment I have renamed the functions to be more consistent, in the following weeks I will be working on the heap viewer in Cutter so I will keep updating the Rizin API for heaps as I need and I will remove any API calls which are not required.

librz/core/cdebug.c
```diff
@@ -639,3 +639,255 @@ RZ_API RzCmdStatus rz_core_debug_plugins_print(RzCore *core, RzCmdStateOutput *s
...
+	}
+	pj_kn(pj, "addr", map->addr);
+	pj_kn(pj, "addr_end", map->addr_end);
+	pj_ks(pj, "type", map->user ? "u" : "s");
+	pj_ks(pj, "perm", rz_str_rwx_i(map->perm));
+	pj_end(pj);
+}
+
+/* Write the memory map header describing the line columns */
+static void print_debug_map_line_header(RzDebug *dbg) {
```
XVilka: Maybe remove this function since it's not implemented?

librz/core/cdebug.c
```diff
@@ -639,3 +639,255 @@ RZ_API RzCmdStatus rz_core_debug_plugins_print(RzCore *core, RzCmdStateOutput *s
...
+/* Write a single memory map line to the console */
+static void print_debug_map_line(RzDebug *dbg, RzDebugMap *map, ut64 addr, RzOutputMode mode) {
+	char humansz[8];
+	if (mode == RZ_OUTPUT_MODE_QUIET) { // "dmq"
+		char *name = (map->name && *map->name)
+			? rz_str_newf("%s.%s", map->name, rz_str_rwx_i(map->perm))
+			: rz_str_newf("%08" PFMT64x ".%s", map->addr, rz_str_rwx_i(map->perm));
+		rz_name_filter(name, 0, true);
+		rz_num_units(humansz, sizeof(humansz), map->addr_end - map->addr);
+		dbg->cb_printf("0x%016" PFMT64x " - 0x%016" PFMT64x " %6s %5s %s\n",
```
XVilka: Since this is located in `librz/core` there is no need to use `cb->printf` here, you could use RzCore and RzCons here freely (this is why we asked to move printing in here after all)/.

librz/core/cdebug.c
```diff
@@ -639,3 +639,255 @@ RZ_API RzCmdStatus rz_core_debug_plugins_print(RzCore *core, RzCmdStateOutput *s
...
+			case RZ_OUTPUT_MODE_JSON: // "dmj"
+				print_debug_map_json(map, pj);
+				break;
+			case RZ_OUTPUT_MODE_RIZIN: // "dm*"
+			{
+				char *name = (map->name && *map->name)
+					? rz_str_newf("%s.%s", map->name, rz_str_rwx_i(map->perm))
+					: rz_str_newf("%08" PFMT64x ".%s", map->addr, rz_str_rwx_i(map->perm));
+				rz_name_filter(name, 0, true);
+				dbg->cb_printf("f map.%s 0x%08" PFMT64x " 0x%08" PFMT64x "\n",
```
XVilka: Same here

librz/core/cdebug.c
```diff
@@ -639,3 +639,255 @@ RZ_API RzCmdStatus rz_core_debug_plugins_print(RzCore *core, RzCmdStateOutput *s
...
+				color_suffix = "";
+			}
+			if ((map->addr - last) > UT32_MAX) { // TODO: Comment what this is for
+				mul = findMinMax(maps, &min, &max, skip, width); //  Recalculate minmax
+			}
+			skip++;
+			fmtstr = dbg->bits & RZ_SYS_BITS_64 // Prefix formatting string (before bar)
+				? "map %4.8s %c %s0x%016" PFMT64x "%s |"
+				: "map %4.8s %c %s0x%08" PFMT64x "%s |";
+			dbg->cb_printf(fmtstr, humansz,
```
XVilka: And here

XVilka (changes requested):

XVilka (approved):
LGTM, except if there is a quick API to substitute that `rz-bin` spawn. Lets see what @ret2libc and @yossizap think.

librz/core/cmd_descs/cmd_debug.yaml
```diff
@@ -19,63 +19,186 @@ commands:
...
+            type: RZ_CMD_ARG_TYPE_FILE
+      - name: dtsm
+        cname: cmd_debug_list_trace_session_mmap
+        summary: List current memory map and hash
+        args: []
+  - name: dm
+    summary: Memory map commands
+    subcommands:
+      - name: dm
+        cname: cmd_debug_list_maps
```
ret2libc, line 86: use `type` `RZ_CMD_DESC_TYPE_ARGV_STATE` if possible.

librz/core/cmd_descs/cmd_debug.yaml
```diff
@@ -19,63 +19,186 @@ commands:
...
+          - RZ_OUTPUT_MODE_STANDARD
+          - RZ_OUTPUT_MODE_RIZIN
+          - RZ_OUTPUT_MODE_JSON
+          - RZ_OUTPUT_MODE_QUIET
+        args: []
+      - name: dm+
+        cname: cmd_debug_allocate_maps
+        summary: >
+          Allocate <size> bytes at current offset
+          (anywhere if offset is -1)
```
ret2libc: is this really true? It seems to me `addr` is passed to the `map_alloc` method of the dbg plugin, which can do whatever it wants. For debug_native plugin, addr is passed as first argument of `mmap`, which needs NULL to allocate "anywhere", not -1.

librz/core/cmd_descs/cmd_debug.yaml
```diff
@@ -19,63 +19,186 @@ commands:
...
+        cname: cmd_debug_map_current
+        summary: Show map name of current address
+        args: []
+      - name: dmm
+        summary: Module memory map commands
+        subcommands:
+          - name: dmm
+            summary: List modules (libraries, binaries loaded in memory)
+            cname: cmd_debug_modules
+            modes:
```
ret2libc, line 118: same, use ARGV_STATE if not too much work. Otherwise, it's ok.

librz/core/linux_heap_glibc.c
```diff
@@ -1406,632 +1836,473 @@ static void GH(print_heap_segment)(RzCore *core, MallocState *main_arena,
...
+		free(main_arena);
+		return RZ_CMD_STATUS_ERROR;
 	}
+	free(dup);
 	free(main_arena);
-	return true;
+	return RZ_CMD_STATUS_OK;
 }
+
+RZ_IPI RzCmdStatus GH(rz_cmd_heap_arena_bins_print_handler)(RzCore *core, int argc, const char **argv, RzOutputMode mode) {
```
ret2libc, line 2256:
```
[0x55a48dde7d80]> dmhd fastbin 
Tcache in Main Arena @  0x7f2f1eb42a00


Fast bins in Arena @ 0x7f2f1eb42a00
Fast_bin[01] [size: 0x20] Empty bin
Fast_bin[02] [size: 0x30] Empty bin
Fast_bin[03] [size: 0x40] Empty bin
Fast_bin[04] [size: 0x50] Empty bin
Fast_bin[05] [size: 0x60] Empty bin
Fast_bin[06] [size: 0x70] Empty bin
Fast_bin[07] [size: 0x80] Empty bin

Unsorted bin in Arena @ 0x7f2f1eb42a00
Found 0 chunks in unsorted bin

Small bins in Arena @ 0x7f2f1eb42a00
Found 0 chunks in 0 small bins

Large bins in Arena @ 0x7f2f1eb42a00
Found 0 chunks in 0 large bins
```
I would expect only fastbins. Works with other bin_type, but not with fastbins.

MalhotraPulak: Currently `dmhd fast` will show fastbins instead of `dmhd fastbin`, I can add support for both if you want

librz/core/cmd_descs/cmd_debug.yaml
MalhotraPulak: Oh, I am not sure about the exact working of this command, this help message was present before the refactor so I just used the same message. Should I remove the ` (anywhere if offset is -1)` part?

ret2libc: For now, yes.

librz/core/linux_heap_glibc.c
ret2libc: I think you used `fastbin` in the CHOICES of the argument. Either accept `fastbin` correctly in this command or change the CHOICES to use `fast` instead of `fastbin`.

MalhotraPulak: Yeah my bad, it should have been "fast", anyways I will change it so that it accepts both

XVilka: I think it's better to change CHOICES to `fast` and do not support `fastbin` to be consistent.

XVilka (approved):
Is this still "WIP"? Looks ready to me.

librz/core/cmd_debug.c
```diff
@@ -1257,84 +1211,75 @@ static int dump_maps(RzCore *core, int perm, const char *filename) {
...
+	RzList *list;
+	RzListIter *iter;
+	PJ *pj = state->d.pj;
+	RzOutputMode mode = state->mode;
+	if (mode == RZ_OUTPUT_MODE_JSON) {
 		if (!pj) {
 			return;
 		}
 		pj_a(pj);
 	}
```
ret2libc:
```suggestion
rz_cmd_state_output_array_start(state);
```

librz/core/cmd_debug.c
```diff
@@ -1257,84 +1211,75 @@ static int dump_maps(RzCore *core, int perm, const char *filename) {
...
+				filtered_name, map->addr);
+			rz_cons_printf("oba 0x%08" PFMT64x " %s\n", map->addr, escaped_path);
+			free(escaped_path);
+			free(filtered_name);
 		}
 	}
-beach:
-	if (mode == 'j') {
+	if (mode == RZ_OUTPUT_MODE_JSON) {
 		pj_end(pj);
```
ret2libc:
```suggestion
rz_cmd_state_output_array_end(state);
```

librz/core/cmd_debug.c
```diff
@@ -1443,376 +1357,462 @@ static bool get_bin_info(RzCore *core, const char *file, ut64 baseaddr, PJ *pj,
 	return true;
 }
 
-static int cmd_debug_map(RzCore *core, const char *input) {
+// dm
```
ret2libc, line 1353: I'm tempted to remove these comments... They will likely become wrong at some point. I'd prefer to focus on https://github.com/rizinorg/rizin/issues/691 to correctly find the right handler.

librz/core/cmd_debug.c
```diff
@@ -1443,376 +1357,462 @@ static bool get_bin_info(RzCore *core, const char *file, ut64 baseaddr, PJ *pj,
...
-			addr = rz_num_math(core->num, input + 1);
-			size = rz_num_math(core->num, p);
-			rz_debug_map_alloc(core->dbg, addr, size, true);
+		perms = rz_str_rwx(argv[1]);
+		if (addr != UT64_MAX && perms >= 0) {
+			rz_debug_map_protect(core->dbg, addr, (int)size, perms);
 		} else {
-			eprintf("Usage: dmL addr size\n");
-			return false;
+			return RZ_CMD_STATUS_WRONG_ARGS;
```
ret2libc:
```suggestion
			return RZ_CMD_STATUS_ERROR;
```
WRONG_ARGS usually means you passed the wrong number of args, but in this case it is something else.

librz/core/cmd_debug.c
```diff
@@ -1443,376 +1357,462 @@ static bool get_bin_info(RzCore *core, const char *file, ut64 baseaddr, PJ *pj,
...
+			addr = UT64_MAX;
+		}
+		if (!addr || addr == UT64_MAX) {
+			libname = argv[1];
+		}
+	}
+	rz_debug_map_sync(core->dbg); // update process memory maps
+	RzList *list = rz_debug_modules_list(core->dbg);
+	rz_list_foreach (list, iter, map) {
+		if ((!libname ||
```
ret2libc, line 1662: very very long `if`. Not sure if it comes from old code, but a separate function with a meaningful name or something else would be better

librz/core/cmd_debug.c
```diff
@@ -1443,376 +1357,462 @@ static bool get_bin_info(RzCore *core, const char *file, ut64 baseaddr, PJ *pj,
...
+#endif
+}
+
+// dmx
+RZ_IPI int rz_cmd_debug_heap_jemalloc(void *data, const char *input) {
+	RzCore *core = (RzCore *)data;
+	if (rz_debug_is_dead(core->dbg)) {
+		rz_cons_println("Debugging is not enabled. Run ood?");
+		return RZ_CMD_STATUS_ERROR;
+	}
```
ret2libc:
```suggestion
#define CMD_CHECK_DEBUG_DEAD(core) do { \
  if (rz_debug_is_dead(core->dbg)) { \
    rz_cons_println("Debugging is not enabled. Run ood?"); \
    return RZ_CMD_STATUS_ERROR; \
  }
} while(0)

	CMD_CHECK_DEBUD_DEAD(core);
```
better to avoid all the repetitions. This is probably the easy thing to reduce that.

ret2libc:
Overall looks good anyway! Great work ;)

librz/core/cmd_debug.c
MalhotraPulak: If I remove `pj_a(pj)` then JSON objects will not be inside a list, making it an invalid JSON

librz/core/cmd_debug.c
MalhotraPulak: Yeah this `if` comes from the old code indeed

librz/core/cmd_debug.c
ret2libc: `rz_cmd_state_output_array_start` is doing that.

MalhotraPulak: thanks! I missed that

ret2libc (approved):
It looks good to me, but since you are touching a lot of debug code (which is not tested too much) please do a round of manual testing with a debugging session. I did it already a couple of versions ago, but better be sure ;)

librz/include/rz_core.h
```diff
@@ -772,6 +773,22 @@ RZ_API char *rz_core_sysenv_begin(RzCore *core, const char *cmd);
...
+RZ_API bool rz_heap_resolve_main_arena_64(RzCore *core, ut64 *m_arena);
+RZ_API bool rz_heap_update_main_arena_64(RzCore *core, ut64 m_arena, MallocState *main_arena);
+RZ_API RzList *rz_heap_tcache_list_64(RzCore *core, ut64 m_arena, MallocState *main_arena, bool main_thread_only);
+RZ_API RzHeapChunk_32 *rz_heap_get_chunk_at_addr_32(RzCore *core, ut32 addr);
+RZ_API RzList *rz_heap_bin_content_list_32(RzCore *core, MallocState *main_arena, int bin_num);
+RZ_API RzList *rz_heap_arenas_list_32(RzCore *core, ut32 m_arena, MallocState *main_arena);
+RZ_API RzList *rz_heap_heap_chunks_list_32(RzCore *core, MallocState *main_arena, ut32 m_arena, ut32 m_state);
+RZ_API bool rz_heap_resolve_main_arena_32(RzCore *core, ut32 *m_arena);
+RZ_API bool rz_heap_update_main_arena_32(RzCore *core, ut32 m_arena, MallocState *main_arena);
+RZ_API RzList *rz_heap_tcache_list_32(RzCore *core, ut32 m_arena, MallocState *main_arena, bool main_thread_only);
```
yossizap, line 791: Since we are planning to use the API here for Cutter's HeapWidget frontend the _32/_64 variations in the API should be made internal and replaced with wrapper functions that will use a bool/bitness var to differentiate architectures.

yossizap (approved):
I am a bit late for this review so I'll just leave the comment regarding the API for a future PR. LGTM other than that.

librz/core/cmd_debug.c
wargio: this is not ok.

wargio: but we can change this in another PR

librz/include/rz_core.h
XVilka: @PulakIIIT I think this can be done in a separate PR. Let's merge this one.

PR #1200, opened 2021-06-09
Remove all PJEncodings

thestr4ng3r:
- [x] I've removed tests that prove my fix is effective or that my feature works (if possible)

**Detailed description**

These encodings set by `cfg.json` would influence how numbers and strings would be emitted to json, for example encoding all strings as base64. This however was merely a workaround for cases where the strings are invalid utf-8 (then they shouldn't be emitted as strings in json in the first place) or a target parser couldn't handle 64bit (then a different parser should be used).

Now all numbers in json are numbers and all strings are strings. In case the string really is invalid, this is technically a programming error in rizin, but pj now strips invalid chars away instead of using the out-of-spec `\x42` so it always emits valid json.


ret2libc:
> This however was merely a workaround for cases where the strings are invalid utf-8 (then they shouldn't be emitted as strings in json in the first place) or a target parser couldn't handle 64bit (then a different parser should be used).

However I saw that many standard json parsers are not able to correctly parse 64bit numbers... curl  and jq IIRC do not handle them well. I agree that it's their problem though :)

XVilka:
curl developer[s] is quite responsive to the issues, so might be enough to report the issue if it still happens.

XVilka:
Looks like it was a JavaScript's fault (what a wonderful language!):
- https://news.ycombinator.com/item?id=12858518
- https://stackoverflow.com/questions/209869/what-is-the-accepted-way-to-send-64-bit-values-over-json

XVilka:
```
71/84 str                                    FAIL            0.03s   killed by signal 6 SIGABRT
>>> MALLOC_PERTURB_=89 /home/build/rizin/build/test/unit/test_str
 ✀  
stderr:
test_str(29471) in free(): chunk canary corrupted 0x30f7408f300 0xb5@0xb5 (double free?)
```

wargio (approved):
THANKS

wargio:
yes, javascript supports up to 53 bits, because 64 bits values are actually doubles, i tested this a long long time ago. you need a library that splits 64 bits numbers to 32 bit ones.

thestr4ng3r:
Yes, json spec explicitly does not specify any bounds so what we emit is all according to spec.

test/db/cmd/cmd_i
```diff
@@ -4764,33 +4764,10 @@ RUN
 NAME=izjencoding
```
XVilka: Could you please also rename this test too?

test/db/cmd/cmd_ax
```diff
@@ -43,24 +43,22 @@ NAME=axjstrencoding
...
 s sym.imp.wprintf
 axj~{[1].name}
 EOF
 EXPECT=<<EOF
-c2VjdGlvbi4uaW5pdCsxNg==
+section..init+16
 EOF
 RUN
 
 NAME=axjnumencoding
```
XVilka: Please rename the test

test/db/cmd/cmd_ax
```diff
@@ -43,24 +43,22 @@ NAME=axjstrencoding
 FILE=bins/elf/analysis/hello-utf-16
```
XVilka, line 43: Please rename this test

XVilka (changes requested):

XVilka (approved):

PR #1206, opened 2021-06-10
Cleanup of RzLang

XVilka:

**Detailed description**

While sometimes it *might* be useful to have the ability to script Rizin in other languages, none of these languages are *a good choice* for scripting:
- Rust
- Vala
- Zip
- SPP

Moreover, the plugins had hardcoded strings of the code stubs in these languages which makes the update and testing too hard.

I think, instead we should aim to split the concept of scripting and using Rizin (librz) from various languages:
- Zig, Vala, Rust can integrate with existing C code just fine (as long as you provide comprehensive API) and use it directly, nothing special required
- For *scripting* provide the way to script inside the Rizin shell without such unnecessary glue. But for this better-suited languages should be used: Python (as the language of choice of the most reverse engineers), Lua, Julia, Haskell, etc
- Improving Swig bindings instead: https://github.com/rizinorg/rizin/pull/285

**Test plan**

CI is green.


wargio (approved):

ret2libc (approved):
I am ok with this

PR #1211, opened 2021-06-13
Fix the broken ASAN test by refactoring `is_cpu_valid()`

officialcjunior:
**Your checklist for this pull request**
- [x] I've read the [guidelines for contributing](https://github.com/rizinorg/rizin/blob/master/DEVELOPERS.md) to this repository
- [x] I made sure to follow the project's [coding style](https://github.com/rizinorg/rizin/blob/master/DEVELOPERS.md#code-style)
- [ ] I've added tests that prove my fix is effective or that my feature works (if possible)
- [ ] I've updated the documentation and the [rizin book](https://github.com/rizinorg/book) with the relevant information (if needed)

**Detailed description**

Fixes the broken ASAN test by removing the use of `RzPVector`.  Also frees `arch_cpu`

**Test plan**

- [x] CI is green

**Closing issues**

None

PR #1212, opened 2021-06-13
Fix RzCore typedef redefinition in rz_arch.h

thestr4ng3r:

**Detailed description**

Fixes this warning for `Apple clang version 12.0.5 (clang-1205.0.22.9)
Target: arm64-apple-darwin20.5.0`:
Bildschirmfoto 2021-06-13 um 14 20 48

### DIFF
--- a/librz/include/rz_arch.h
+++ b/librz/include/rz_arch.h
@@ -12,7 +12,7 @@
 extern "C" {
 #endif
 
-typedef struct rz_core_t RzCore;
+struct rz_core_t;
 
 typedef struct rz_arch_profile_t {
 	ut64 rom_size;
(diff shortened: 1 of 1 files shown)


librz/include/rz_arch.h
```diff
@@ -12,7 +12,7 @@
 extern "C" {
 #endif
 
-typedef struct rz_core_t RzCore;
+struct rz_core_t;
```
ret2libc, line 15: I probably missed the initial arch review, but why does arch needs rzcore? I think there shouldn't be any circular dependencies and arch should be usable even without core headers at all.

thestr4ng3r: @XVilka @wargio @officialcjunior

officialcjunior: It is just for this little API here:
https://www.github.com/rizinorg/rizin/tree/dev/librz%2Finclude%2Frz_arch.h#L43 -- `rz_arch_profile_add_flag_every_io`

which adds the IO/Extended IO registers from the CPU Profiles as flags.

I just thought it doesn't make sense to be put elsewhere. Sorry. 
I think it can be moved to where it should be.

XVilka (approved):

librz/include/rz_arch.h
XVilka: We probably could move that initialization in the RzCore instead.

officialcjunior: Alright.
I will fix that one up in https://github.com/rizinorg/rizin/pull/1211, then.

wargio: no is better to do this in another PR

wargio (approved):

PR #1214, opened 2021-06-14
Refactor add new interface to fetch data from elf ##refactor

ghost:

**Detailed description**

Add a new interface to fetch data from the elf:

* Elf_(rz_bin_elf_read_char)
* Elf_(rz_bin_elf_read_half)
* Elf_(rz_bin_elf_read_word)
* Elf_(rz_bin_elf_read_sword)
* Elf_(rz_bin_elf_read_xword)
* Elf_(rz_bin_elf_read_sxword)
* Elf_(rz_bin_elf_read_addr)
* Elf_(rz_bin_elf_read_off)
* Elf_(rz_bin_elf_read_section)
* Elf_(rz_bin_elf_read_versym)

**Problems**:

There is a code smell. Some struct use (s)word or (s)xword (depending of the endianess). So 2 functions
could be added to fix this. But I don't know how to named them...

In Elf_(rz_bin_elf_read_sword) and Elf_(rz_bin_elf_read_sxword), there is a undefined behavior when converting unsigned to signed. The problem is sword can have negative value. And I don't know how to the converting without casting.

The type of ut64 (st64) isn't equal to the type of uint64_t (int64_t) on my computer. It is why there is a cast before read_buffer_64 and read_buffer_32

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/librz/bin/format/elf/elf.c
+++ b/librz/bin/format/elf/elf.c
@@ -134,35 +134,32 @@ static HtUP *rel_cache_new(RzBinElfReloc *relocs, ut32 reloc_num) {
 	return rel_cache;
 }
 
-static void set_phdr_entry(ELFOBJ *bin, size_t phdr_entry_index, ut8 *entry) {
-	size_t pos = 0;
-
-	bin->phdr[phdr_entry_index].p_type = READ32(entry, pos);
+static bool set_phdr_entry(ELFOBJ *bin, Elf_(Phdr) * segment, ut64 offset) {
+	if (!Elf_(rz_bin_elf_read_word)(bin, &offset, &segment->p_type) ||
 #if RZ_BIN_ELF64
-	bin->phdr[phdr_entry_index].p_flags = READ32(entry, pos);
+		!Elf_(rz_bin_elf_read_word)(bin, &offset, &segment->p_flags) ||
 #endif
-	bin->phdr[phdr_entry_index].p_offset = RZ_BIN_ELF_READWORD(entry, pos);
-	bin->phdr[phdr_entry_index].p_vaddr = RZ_BIN_ELF_READWORD(entry, pos);
-	bin->phdr[phdr_entry_index].p_paddr = RZ_BIN_ELF_READWORD(entry, pos);
-	bin->phdr[phdr_entry_index].p_filesz = RZ_BIN_ELF_READWORD(entry, pos);
-	bin->phdr[phdr_entry_index].p_memsz = RZ_BIN_ELF_READWORD(entry, pos);
+		!Elf_(rz_bin_elf_read_off)(bin, &offset, &segment->p_offset) ||
+		!Elf_(rz_bin_elf_read_addr)(bin, &offset, &segment->p_vaddr) ||
+		!Elf_(rz_bin_elf_read_addr)(bin, &offset, &segment->p_paddr) ||
+		!Elf_(rz_bin_elf_read_word_xword)(bin, &offset, &segment->p_filesz) ||
+		!Elf_(rz_bin_elf_read_word_xword)(bin, &offset, &segment->p_memsz) ||
 #ifndef RZ_BIN_ELF64
-	bin->phdr[phdr_entry_index].p_flags = READ32(entry, pos);
+		!Elf_(rz_bin_elf_read_word)(bin, &offset, &segment->p_flags) ||
 #endif
-	bin->phdr[phdr_entry_index].p_align = RZ_BIN_ELF_READWORD(entry, pos);
-}
-
-static bool read_phdr_entry(ELFOBJ *bin, size_t phdr_entry_index) {
-	const size_t offset = bin->ehdr.e_phoff + phdr_entry_index * sizeof(Elf_(Phdr));
-	ut8 phdr[sizeof(Elf_(Phdr))] = { 0 };
-
-	if (rz_buf_read_at(bin->b, offset, phdr, sizeof(Elf_(Phdr))) < 0) {
-		bprintf("read (phdr)\n");
-		RZ_FREE(bin->phdr);
+		!Elf_(rz_bin_elf_read_word_xword)(bin, &offset, &segment->p_align)) {
 		return false;
 	}
 
-	set_phdr_entry(bin, phdr_entry_index, phdr);
+	return true;
+}
+
+static bool read_phdr_entry(ELFOBJ *bin, size_t phdr_entry_index) {
+	ut64 offset = bin->ehdr.e_phoff + phdr_entry_index * sizeof(Elf_(Phdr));
+
+	if (!set_phdr_entry(bin, bin->phdr + phdr_entry_index, offset)) {
+		return false;
+	}
 
 	return true;
 }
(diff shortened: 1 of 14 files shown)


librz/bin/format/elf/elf.c
```diff
@@ -417,29 +436,44 @@ static bool init_ehdr_ident(ELFOBJ *bin) {
...
-	bin->ehdr.e_flags = READ32(ehdr, i);
-	bin->ehdr.e_ehsize = READ16(ehdr, i);
-	bin->ehdr.e_phentsize = READ16(ehdr, i);
-	bin->ehdr.e_phnum = READ16(ehdr, i);
-	bin->ehdr.e_shentsize = READ16(ehdr, i);
-	bin->ehdr.e_shnum = READ16(ehdr, i);
-	bin->ehdr.e_shstrndx = READ16(ehdr, i);
+	ut64 offset = EI_NIDENT;
+
+	if (!Elf_(rz_bin_elf_read_half)(bin, &offset, &bin->ehdr.e_type) ||
```
XVilka, line 434: Note, that the previous code was more efficient since it was reading all of the header in the intermediate buffer, and then reading data from the buffer, since going all over the RzBin interface could be a bit slower - and if you have many read operations it multiplies. On the other hand, the RzBin has been changed significantly since by @thestr4ng3r, so it's unclear. I recommend you to compare the performance of your solution vs the old one, and if the new is slower - use the intermediate buffer as well.

ghost: Before

```
( repeat 10; do; rz-test db/formats/elf; done; )  105.47s user 11.56s system 373% cpu 31.366 total
```

After

```
( repeat 10; do; rz-test db/formats/elf; done; )  100.12s user 12.07s system 352% cpu 31.828 total
```

There is a slow down, but compare to the number of binary executed around 1480...

I didn't use a static buffer, because it is prone to bug.

librz/bin/format/elf/elf_misc.c
```diff
@@ -0,0 +1,153 @@
...
+		return false;
+	}
+
+	*offset += sizeof(ut32);
+	*result = tmp;
+
+	return true;
+}
+
+static st32 convert_to_complement_32(ut32 value) {
```
wargio: `static inline` and change name to `convert_to_two_complement_32`

librz/bin/format/elf/elf_misc.c
```diff
@@ -0,0 +1,153 @@
...
+		return false;
+	}
+
+	*offset += sizeof(ut64);
+	*result = tmp;
+
+	return true;
+}
+
+static st64 convert_to_complement_64(ut64 value) {
```
wargio: same here.

librz/bin/format/elf/elf.c
thestr4ng3r: I would expect lto to take care of most of this when enabled.

librz/bin/format/elf/elf_info.c
```diff
@@ -1124,6 +1141,8 @@ static Sdb *get_verdef_entry_sdb_aux(ELFOBJ *bin, Elf_(Verdef) * verdef_entry, s
 }
 
 static Sdb *get_verdef_entry_sdb(ELFOBJ *bin, Elf_(Verdef) * verdef_entry, size_t offset) {
+	Elf_(Verdaux) verdaux_entry;
```
thestr4ng3r: Better move these definitions to just before they are used for the first time. Here directly before `if (!get_verdaux_entry...` below.

librz/bin/format/elf/elf_info.c
```diff
@@ -1154,6 +1176,7 @@ static Sdb *get_version_info_gnu_verdef(ELFOBJ *bin) {
 
 	ut64 verdef_addr;
 	ut64 verdef_num;
+	Elf_(Verdef) verdef_entry;
```
thestr4ng3r: same

librz/bin/format/elf/elf_info.c
```diff
@@ -1232,6 +1258,8 @@ static Sdb *get_verneed_entry_sdb_aux(ELFOBJ *bin, Elf_(Verneed) verneed_entry,
 }
 
 static Sdb *get_verneed_entry_sdb(ELFOBJ *bin, Elf_(Verneed) verneed_entry, size_t offset) {
+	Elf_(Vernaux) vernaux_entry;
```
thestr4ng3r: same

librz/bin/format/elf/elf_info.c
```diff
@@ -1260,6 +1292,7 @@ static Sdb *get_verneed_entry_sdb(ELFOBJ *bin, Elf_(Verneed) verneed_entry, size
 static Sdb *get_version_info_gnu_verneed(ELFOBJ *bin) {
 	ut64 verneed_addr;
 	ut64 verneed_num;
+	Elf_(Verneed) verneed_entry;
```
thestr4ng3r: same

librz/bin/format/elf/elf_relocs.c
```diff
@@ -43,27 +43,34 @@ static ut64 get_size_rel_mode(Elf_(Xword) rel_mode) {
 }
 
 static bool read_reloc(ELFOBJ *bin, RzBinElfReloc *r, Elf_(Xword) rel_mode, ut64 vaddr) {
+	Elf_(Rela) reloc_info;
```
thestr4ng3r: same

thestr4ng3r (commented):
overall looks good

librz/bin/format/elf/elf.c
XVilka: What do you mean by "when enabled"?

XVilka (approved):

librz/bin/format/elf/elf.c
```diff
@@ -134,36 +134,41 @@ static HtUP *rel_cache_new(RzBinElfReloc *relocs, ut32 reloc_num) {
...
-#endif
-	bin->phdr[phdr_entry_index].p_offset = RZ_BIN_ELF_READWORD(entry, pos);
-	bin->phdr[phdr_entry_index].p_vaddr = RZ_BIN_ELF_READWORD(entry, pos);
-	bin->phdr[phdr_entry_index].p_paddr = RZ_BIN_ELF_READWORD(entry, pos);
-	bin->phdr[phdr_entry_index].p_filesz = RZ_BIN_ELF_READWORD(entry, pos);
-	bin->phdr[phdr_entry_index].p_memsz = RZ_BIN_ELF_READWORD(entry, pos);
-#ifndef RZ_BIN_ELF64
-	bin->phdr[phdr_entry_index].p_flags = READ32(entry, pos);
+	if (!Elf_(rz_bin_elf_read_word)(bin, &offset, &segment->p_type) ||
+		!Elf_(rz_bin_elf_read_word)(bin, &offset, &segment->p_flags) ||
```
ret2libc, line 140: can't you just ifdef this if this is the only difference between ELF64 and ELF32? Otherwise, create another static bool function that is defined in one case and empty in the other case.

librz/bin/format/elf/elf.c
```diff
@@ -249,33 +254,45 @@ static void init_shdr_sdb(ELFOBJ *bin) {
 	sdb_set(bin->kv, "elf_shdr.format", sdb_elf_shdr_format, 0);
 }
 
-static void set_shdr_entry(ELFOBJ *bin, size_t shdr_entry_index, ut8 *entry) {
-	size_t pos = 0;
+static bool set_shdr_entry(ELFOBJ *bin, Elf_(Shdr) * section, ut64 offset) {
+#if RZ_BIN_ELF64
```
ret2libc: same here... try to make the ifdef code as small/easily understandable as possible.

You could have:
```C
#if RZ_BIN_ELF64
static bool Elf_(rz_bin_elf_read_word_xword)(....) {
   return !Elf_(rz_bin_elf_read_xword)(bin, ....);
}
#else
static bool Elf_(rz_bin_elf_read_word_xword)(....) {
   return !Elf_(rz_bin_elf_read_word)(bin, ....);
}
```

and then in `set_shdr_entry` just use `rz_bin_elf_read_word_xword` instead of having two completely separate paths, one for ELF64 and one for ELF32.

librz/bin/format/elf/elf_relocs.c
```diff
@@ -48,22 +48,28 @@ static bool read_reloc(ELFOBJ *bin, RzBinElfReloc *r, Elf_(Xword) rel_mode, ut64
...
+	if (!Elf_(rz_bin_elf_read_addr)(bin, &offset, &reloc_info.rz_offset)) {
 		return false;
 	}
 
-	size_t i = 0;
-	Elf_(Rela) reloc_info;
-
-	reloc_info.rz_offset = RZ_BIN_ELF_READWORD(buf, i);
-	reloc_info.rz_info = RZ_BIN_ELF_READWORD(buf, i);
+#if RZ_BIN_ELF64
```
ret2libc: as said, if you create the function suggested above you can probably reuse it in many other places like this one.

librz/bin/format/elf/elf.c
ghost: Some field are not in the same order. (p_flags)

librz/bin/format/elf/elf_misc.c
```diff
@@ -0,0 +1,163 @@
...
+		return false;
+	}
+
+	*result = convert_to_two_complement_32(tmp);
+
+	return true;
+}
+
+static bool buffer_read_64(ELFOBJ *bin, ut64 *offset, ut64 *result) {
+	ut64 tmp = rz_buf_read_ble64_at(bin->b, *offset, bin->endian);
```
ret2libc, line 56: I suggest to use `rz_buf_read_at` and check  for the return value instead. This way, if any of the value you want to read can have UTXX_MAX as a valid value, it won't break.

ret2libc: Same for `buffer_read_32`

librz/bin/format/elf/elf_misc.c
```diff
@@ -0,0 +1,163 @@
...
+	if (tmp == UT64_MAX) {
+		return false;
+	}
+
+	*offset += sizeof(ut64);
```
ret2libc, line 61: from a look around I have not seen a case where `offset` is really needed to be updated. In particular, most uses of these new functions are just like "start from offset X and read some fields, one after the other". If this is the case, you could just use `rz_buf_seek` to move `bin->b` to the right offset and then use `rz_buf_read`-like functions instead of `rz_buf_read_at`. The difference is that `_at` functions perform a seek to get the current position, save it, seek to the specified address, perform the operation, seek back to the original address. Repeat for the next `_at` operation. All those seek are just useless probably.

librz/bin/format/elf/elf_misc.c
```diff
@@ -0,0 +1,163 @@
...
+}
+
+static bool buffer_read_64_signed(ELFOBJ *bin, ut64 *offset, st64 *result) {
+	ut64 tmp;
+
+	if (!buffer_read_64(bin, offset, &tmp)) {
+		return false;
+	}
+
+	*result = convert_to_two_complement_64(tmp);
```
ret2libc, line 74: Mmmm I'm not convinced of this. Why is it necessary? Can't you directly read in a st64 variable?

librz/bin/format/elf/elf.c
ret2libc: Ok, good point!

ret2libc: Still, I prefer the previous "format" where the common operations were outside of the ifdef and the special cases were in the ifdefs. Otherwise it is hard to compare the two things.

librz/bin/format/elf/elf_misc.c
ghost: The conversion between signed and unsigned is platform dependent. So Rizin should manually convert with this function in such case.

librz/bin/format/elf/elf_misc.c
ghost: Indeed, but those read functions shouldn't modify the bin->b state.

librz/bin/format/elf/elf_misc.c
ghost: Indeed, it could be a good idea. But the modification should be done in rz_buf. There should be a refactoring on the rz_buf interface.

librz/bin/format/elf/elf_misc.c
ret2libc: how is this done everywhere else? I think this is not a elf specific issue. @wargio @XVilka @thestr4ng3r

librz/bin/format/elf/elf_misc.c
ret2libc: Shouldn't they? Why?

Also, maybe it's jsut enough to restore the original seek after the global functions have finished.

ghost: Those functions should modify the state because, the state should be reset after each call and this is error prone behavior.

ret2libc: How so? bin->b is/should be an internal field of RzBin and it is somehow expected that the parsing functions read the file and move around it.

ret2libc: Anyway, it is fine as it is (and it was), so no need to worry now.

librz/bin/format/elf/elf_misc.c
XVilka: I am not 100% sure, it's either not handled at all, or just some band-aid in particular places.

thestr4ng3r: Afaik in other cases it is just casted, which as @08A pointed out is implementation-defined (not undefined) behavior. So to be 100% correct, the approach here is indeed the right one.

thestr4ng3r (approved):

XVilka (approved):
Looks good now. Ready to merge, in my opinion.

XVilka:
Unrelated to this PR, but spotted during the Windows build log inspection:
```
../librz/bin/format/elf/elf_symbols.c(824,19): warning: unused function '__strnlen' [-Wunused-function]
static inline int __strnlen(const char *str, int len) {
                  ^
1 warning generated
```
See https://ci.appveyor.com/project/rizinorg/rizin/builds/39635855/job/x5wdc7otqdydxyqt#L28945

@08A

PR #1230, opened 2021-06-20
Avoid logging stuff unless it's very important

wargio:

**Detailed description**

this should greenify the dev

### DIFF
--- a/librz/bin/format/elf/elf_segments.c
+++ b/librz/bin/format/elf/elf_segments.c
@@ -56,7 +56,7 @@ static bool get_elf_segment(ELFOBJ *bin, RzBinElfSegment *segment, ut64 offset, 
 
 	segment->is_valid = verify_phdr_entry(bin, &segment->data);
 	if (!segment->is_valid) {
-		rz_assert_log(RZ_LOGLVL_WARN, "Invalid segment %zu at 0x%" PFMT64x "\n", pos, offset);
+		RZ_LOG_INFO("Invalid segment %zu at 0x%" PFMT64x "\n", pos, offset);
 	}
 
 	return true;
@@ -107,7 +107,11 @@ static bool check_phdr_size(ELFOBJ *bin) {
 }
 
 RZ_BORROW RzBinElfSegment *Elf_(rz_bin_elf_get_segment_with_type)(RZ_NONNULL ELFOBJ *bin, Elf_(Word) type) {
-	rz_return_val_if_fail(bin && bin->segments, NULL);
+	rz_return_val_if_fail(bin, NULL);
+	if (!bin->segments) {
+		return NULL;
+	}
+
 	RzBinElfSegment *iter;
 
 	rz_bin_elf_foreach_segments(bin, iter) {


librz/bin/format/elf/elf_segments.c
```diff
@@ -56,7 +56,7 @@ static bool get_elf_segment(ELFOBJ *bin, RzBinElfSegment *segment, ut64 offset,
 
 	segment->is_valid = verify_phdr_entry(bin, &segment->data);
 	if (!segment->is_valid) {
-		rz_assert_log(RZ_LOGLVL_WARN, "Invalid segment %zu at 0x%" PFMT64x "\n", pos, offset);
+		RZ_LOG_INFO("Invalid segment %zu at 0x%" PFMT64x "\n", pos, offset);
```
ret2libc, line 59: @ret2libc note for me to check this.

wargio: this was already merged here: https://github.com/rizinorg/rizin/pull/1231

ret2libc: yes but why was this needed? It seemed alright to have it as a WARNING, no? What was the problem?

wargio: just tests failing because the ASAN tests have stderr redirected to stdout.

wargio: https://github.com/rizinorg/rizin/runs/2868460639
Plus that information is more at info level than warning, because otherwise you may get spammed by rizin.

ret2libc: @08A apparently the kernel is able to load tiny.elf even if p_offset + p_fsize is bigger than the file size, so I am not sure the second check done in `verify_phdr_entry` is correct.
```C
	Elf_(Off) end_off;
	if (!Elf_(rz_bin_elf_add_off)(&end_off, entry->p_offset, entry->p_filesz) || end_off > bin->size) {
		return false;
	}
```

wargio: you can point this out in the mm chat.

ghost: > [rizinorg/rizin/runs/2868460639](https://github.com/rizinorg/rizin/runs/2868460639)
> Plus that information is more at info level than warning, because otherwise you may get spammed by rizin.

What is the default log level? I classified those information as warning because the user should be notified that the loaded sections or segments seems to be invalid.

ghost: > @08A apparently the kernel is able to load tiny.elf even if p_offset + p_fsize is bigger than the file size, so I am not sure the second check done in `verify_phdr_entry` is correct.
> 
> ```c
> 	Elf_(Off) end_off;
> 	if (!Elf_(rz_bin_elf_add_off)(&end_off, entry->p_offset, entry->p_filesz) || end_off > bin->size) {
> 		return false;
> 	}
> ```

Tiny elf is a corner case. The verification check add more constraint than the kernel, because the idea is to notify the user when the segment or sections seems to be invalid.

ghost: > just tests failing because the ASAN tests have stderr redirected to stdout.

Shouldn't asan test be redirected to a file? Because some time warning just indicate important information that the user should watch.

ret2libc: Indeed I'm not sure why this was converted to INFO... I think converting the log to info just "hides" a problem.

@08A does it break anything if tiny.elf has `segment->is_valid=false`?

ghost: I test on tiny.elf, and it doesn't seems to break anything (tiny.elf doesn't have any symbols, relocs, ...).

wargio: because it is indeed an INFO log. we do not need to spam when it fails to load something.

ret2libc: @08A ok in this case it does not break anything, but i'm not sure it is alright to set `segment->isvalid` to false if the kernel is actually able to load such a segment. if the kernel can load it, it means the segment is valid IMO.

ghost: Tiny elf is a corner in maybe 90% of the case if the offset + filesize > bin->size the segments is invalid. The idea isn't to hold the absolute truth but to be good enough. In the future a configuration variable to disable the validity analysis could be created.

ret2libc: > in maybe 90% of the case if the offset + filesize > bin->size the segments is invalid

Are they really invalid? Maybe they are only if the `offset` is bigger than bin->size but not if offset + filesize > bin->size... Just guessing from the kernel behaviour I see.

ghost: I don't know. All those verification are just to notify the user that something seems wrong.

PR #1234, opened 2021-06-21
Refactor how sections are stored in ELFOBJ ##refactor

ghost:

**Detailed description**

* Fix some elf_notes problems (coverity)
* Add the flag is_valid to section information
* Refactor how section are stored
* Introduce a new misc function (rz_bin_elf_check_array)

The section loading use 2 steps because RzBinElfSection need shstrtab that need Elf_(Shdr).

PS: The next step is refactoring how strtab, dynstr, shstrab are stored

**Test plan**
`rz-test`

**Closing issues**

...

### DIFF
--- a/librz/bin/format/elf/elf.c
+++ b/librz/bin/format/elf/elf.c
@@ -180,7 +180,7 @@ static bool set_shstrtab(ELFOBJ *bin, Elf_(Off) offset) {
 
 	int res = rz_buf_read_at(bin->b, offset, (ut8 *)bin->shstrtab, bin->shstrtab_size);
 	if (res < 0) {
-		bprintf("read (shstrtab) at 0x%" PFMT64x "\n", (ut64)offset);
+		RZ_LOG_WARN("read (shstrtab) at 0x%" PFMT64x "\n", (ut64)offset);
 		RZ_FREE(bin->shstrtab);
 		return false;
 	}
@@ -240,7 +240,6 @@ static bool init_ehdr_ident(ELFOBJ *bin) {
 	memset(&bin->ehdr, 0, sizeof(Elf_(Ehdr)));
 
 	if (rz_buf_read_at(bin->b, 0, bin->ehdr.e_ident, EI_NIDENT) == -1) {
-		bprintf("read (magic)\n");
 		return false;
 	}
 
@@ -316,7 +315,6 @@ static bool set_dynstr(ELFOBJ *bin, RzBinElfSection *section) {
 	bin->dynstr = RZ_NEWS0(char, bin->dynstr_size + 1);
 
 	if (!bin->dynstr) {
-		bprintf("Cannot allocate memory for dynamic strings\n");
 		return false;
 	}
 
@@ -393,7 +391,7 @@ static bool rz_bin_elf_init_strtab(ELFOBJ *bin) {
 	ut64 strtab_size;
 
 	if (!Elf_(rz_bin_elf_get_dt_info)(bin, DT_STRTAB, &strtab_addr) || !Elf_(rz_bin_elf_get_dt_info)(bin, DT_STRSZ, &strtab_size)) {
-		bprintf("DT_STRTAB not found or invalid\n");
+		RZ_LOG_WARN("DT_STRTAB not found or invalid\n");
 		return false;
 	}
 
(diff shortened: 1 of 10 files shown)


XVilka:
Since we are removing `elf_write` here, what about removing it for all formats instead (in a separate PR)?

ret2libc (changes requested):
Why are we removing elf_write?

If we remove that, `rz-bin -O` wont' work anymore, I think. There are some interesting things there, like changing entrypoints, change section permission.

Do we plan to drop all these?

XVilka:
@ret2libc you can do this and much more with [LIEF](https://github.com/lief-project/LIEF). Our implementation is too basic, especially for resizing or moving sections. We could just create a plugin with LIEF in the future, IMHO.

wargio:
i agree with Xvilka. LIEF is much much better than this. i would avoid to implement this since we would need to regenerate the structures and that is a mess.
We can add later a plugin via rz-pm to do this or an extra utils for this stuff.

ret2libc:
I'm perfectly fine with removing the feature actually. But if we do, as @XVilka said, we should remove it for all plugins and we should also remove the flags/commands from rz-bin (`-O`) and rizin (`iO`).

ghost:
~Fix 2 test in travis CI~ and create 2 new failure in github CI.

ghost:
**2 failed tests**

I don't know why this 2 tests fail. On my computer it worked.

PS: I tested with a 32/64 bit binary

ghost:
I believe, I fixed the issue (wasn't using the appropriate function to warn).

ret2libc:
Are you going to remove the `write` method and the others bin_*_write plugins in a following PR?

librz/bin/format/elf/elf.c
```diff
@@ -156,133 +156,63 @@ static void init_shdr_sdb(ELFOBJ *bin) {
...
 	if (!bin->shstrtab) {
-		perror("malloc");
 		return false;
 	}
 
-	int res = rz_buf_read_at(bin->b, bin->shstrtab_section->sh_offset, (ut8 *)bin->shstrtab, bin->shstrtab_section->sh_size);
+	int res = rz_buf_read_at(bin->b, offset, (ut8 *)bin->shstrtab, bin->shstrtab_size);
 	if (res < 0) {
-		bprintf("read (shstrtab) at 0x%" PFMT64x "\n", (ut64)bin->shstrtab_section->sh_offset);
+		bprintf("read (shstrtab) at 0x%" PFMT64x "\n", (ut64)offset);
```
wargio: what is bprintf? and since this is a fail, please use RZ_LOG_ERR here

librz/bin/format/elf/elf.c
```diff
@@ -505,25 +421,29 @@ static bool rz_bin_elf_init(ELFOBJ *bin) {
...
 		}
 	} else {
-		if (!rz_bin_elf_init_shdr(bin)) {
-			bprintf("Cannot initialize section headers\n");
-		}
+		RzVector *sections = Elf_(rz_bin_elf_new_sections)(bin);
 
-		if (!rz_bin_elf_init_shstrtab(bin)) {
+		if (!rz_bin_elf_init_shstrtab(bin, sections)) {
 			bprintf("Cannot initialize strings table\n");
```
wargio: same here

librz/bin/format/elf/elf.c
```diff
@@ -505,25 +421,29 @@ static bool rz_bin_elf_init(ELFOBJ *bin) {
...
 
-		if (!rz_bin_elf_init_dynstr(bin) && !Elf_(rz_bin_elf_is_relocatable)(bin)) {
-			bprintf("Cannot initialize dynamic strings\n");
-		}
-
-		bin->baddr = Elf_(rz_bin_elf_get_baddr)(bin);
-
 		if (!Elf_(rz_bin_elf_is_relocatable)(bin) && !Elf_(rz_bin_elf_is_static)(bin)) {
 			if (!rz_bin_elf_init_dynamic_section(bin) || !rz_bin_elf_init_strtab(bin)) {
 				bprintf("Cannot initialize dynamic section\n");
```
wargio: same here.

librz/bin/format/elf/elf.c
```diff
@@ -505,25 +421,29 @@ static bool rz_bin_elf_init(ELFOBJ *bin) {
...
 			}
 		}
+
+		bin->baddr = Elf_(rz_bin_elf_get_baddr)(bin);
+
+		if (!rz_bin_elf_init_shdr(bin, sections)) {
+			bprintf("Cannot initialize section headers\n");
```
wargio: same here.

librz/bin/format/elf/elf.c
```diff
@@ -505,25 +421,29 @@ static bool rz_bin_elf_init(ELFOBJ *bin) {
...
+		}
+
+		rz_vector_free(sections);
+
+		if (!rz_bin_elf_init_dynstr(bin) && !Elf_(rz_bin_elf_is_relocatable)(bin)) {
+			bprintf("Cannot initialize dynamic strings\n");
```
wargio: same here

wargio (changes requested):

ghost:
> Are you going to remove the `write` method and the others bin_*_write plugins in a following PR?

~~Yes, It is the idea.~~ I will send a PR that remove every RzBinWrite.

librz/bin/format/elf/elf.c
ghost: It was a think I wanted to remove. It is a macro that printf if bin->verbose == true

ghost:
@wargio Should I remove bin->verbose option?

wargio:
> @wargio Should I remove bin->verbose option?

what is used for?

ghost:
> > @wargio Should I remove bin->verbose option?
> 
> what is used for?

In some place to print verbose output. There isn't a lot of use.

wargio:
> > > @wargio Should I remove bin->verbose option?
> > 
> > 
> > what is used for?
> 
> In some place to print verbose output. There isn't a lot of use.

Ok, remove the usage of this from the ELF code, and convert any eprintf/printf that uses that into the appropriate RZ_LOG_* counterpart.

Please check the output when loading some elf files, to see how verbose it is with the default log level (which is warning).

ghost:
> > > > @wargio Should I remove bin->verbose option?
> > > 
> > > 
> > > what is used for?
> > 
> > 
> > In some place to print verbose output. There isn't a lot of use.
> 
> Ok, remove the usage of this from the ELF code, and convert any eprintf/printf that uses that into the appropriate RZ_LOG_* counterpart.
> 
> Please check the output when loading some elf files, to see how verbose it is with the default log level (which is warning).

I remove the use of bprintf when the information was not important for an end user.

XVilka (approved):

wargio (approved):

ret2libc (approved):

PR #1235, opened 2021-06-22
Convert `!` commands to newshell

### LINKED ISSUE #1134: Update currently broken core_plugin_example.c to Newshell API
The example core plugin at https://github.com/rizinorg/rizin/blob/bdface85a24a52d1c4a8cf2daa00dc6099e5db51/librz/core/p/core_plugin_example.c still uses a `.call = rz_cmd_example_call` callback for injecting commands, which does not exist anymore.

For reference, here are some plugins that have been updated:
https://git.sr.ht/~thestr4ng3r/rz-commodore/tree/master/item/src/core_commodore.c
https://github.com/rizinorg/rz-ghidra/blob/dev/src/core_ghidra.cpp
https://github.com/rizinorg/jsdec/blob/master/p/core_pdd.c

It might also be nice to somehow embed this in our build system and add to the test suite in some way.

wargio:
# squash me

 

**Detailed description**

- ✅ `!                        list all historic commands`
  - Moved under `H`
- ✅ `!ls                      execute 'ls' in shell`
- ✅ `!echo $RZ_SIZE           display file size`
  - **Removed** `RZ_BYTES` and `RZ_BLOCK` because useless
- ❌ `!*rzp x                  run rizin command via rzpipe in current session`
  - **Removed**
- ✅ `!!                       save command history to hist file`
  - Moved under `H+`
- ✅ `!!ls~txt                 print output of 'ls' and grep for 'txt'`
- ❌ `!!!cmd [args|$type]      adds the autocomplete value`
  - **Removed**
- ❌ `!!!-cmd [args]           removes the autocomplete value`
  - **Removed**
- ✅ `!-                       clear history in current session`
  - **Moved under** `H-`
- ❌ `!-*                      clear and save empty history log`
  - **Removed**

**Issues**

- Fix https://github.com/rizinorg/rizin/issues/1134


librz/core/cfile.c
```diff
@@ -561,16 +529,14 @@ CURSOR cursor position(offset from curseek)
...
-
 	rz_sys_setenv("RZ_BIN_LANG", rz_config_get(core->config, "bin.lang"));
 	rz_sys_setenv("RZ_BIN_DEMANGLE", rz_config_get(core->config, "bin.demangle"));
 	rz_sys_setenv("RZ_ARCH", rz_config_get(core->config, "asm.arch"));
 	rz_sys_setenv("RZ_BITS", sdb_fmt("%" PFMT64u, rz_config_get_i(core->config, "asm.bits")));
-	rz_sys_setenv("RZ_COLOR", rz_config_get_i(core->config, "scr.color") ? "1" : "0");
-	rz_sys_setenv("RZ_DEBUG", rz_config_get_b(core->config, "cfg.debug") ? "1" : "0");
-	rz_sys_setenv("RZ_IOVA", rz_config_get_i(core->config, "io.va") ? "1" : "0");
+	rz_sys_setenv("RZ_COLOR", sdb_fmt("%" PFMT64u, rz_config_get_i(core->config, "scr.color")));
+	rz_sys_setenv("RZ_DEBUG", rz_config_get_b(core->config, "cfg.debug") ? "true" : "false");
```
XVilka: I suggest to just create some uber-simple RzUtil function to convert `bool` to string and use it here.

librz/core/p/core_plugin_example.c
```diff
@@ -1,40 +1,85 @@
+// SPDX-FileCopyrightText: 2021 somebody <somebody@domain.tld>
 // SPDX-License-Identifier: LGPL-3.0-only
 #if 0
```
XVilka: Can you convert this comment into Doxygen `\file` instead?

XVilka (changes requested):
Overall looks good.

librz/core/cmd_descs/cmd_system.yaml
```diff
@@ -0,0 +1,71 @@
+# SPDX-FileCopyrightText: 2021 RizinOrg <info@rizin.re>
+# SPDX-License-Identifier: LGPL-3.0-only
+---
+name: cmd_system3
```
ret2libc: why `3`?

librz/core/cmd_descs/cmd_system.yaml
```diff
@@ -0,0 +1,71 @@
...
+        flags: RZ_CMD_ARG_FLAG_ARRAY
+        optional: true
+    details:
+      - name: Examples
+        entries:
+          - text: "!!"
+            arg_str: "ls~txt"
+            comment: "executes the 'ls' command via system(3) and grep for 'txt'"
+  # These commands should me moved from ! to somewhere else.
+  - name: "!="
```
ret2libc: where did you get these from?

librz/core/cmd_descs/cmd_system.yaml
XVilka: Agree. Should be `42`.

librz/core/cmd_descs/cmd_system.yaml
wargio: It was a bit confusing having `!<cmd>` and `!<num>` under the same command,
therefore i moved the `show list history` and `execute history N` under `!=`

Btw, I added all my reasoning in my commit message; you can click on the 3 dots and read the body message
image

librz/core/cfile.c
wargio: tbh, i do not see any reason to have this. maybe a macro, but still, not really needed.

librz/core/cmd_descs/cmd_system.yaml
wargio: because is the cmd that calls system?

librz/core/cmd_system.c
```diff
@@ -0,0 +1,123 @@
...
+		RZ_LOG_ERROR("cannot find command with index %d.\n", index);
+		return RZ_CMD_STATUS_ERROR;
+	}
+
+	int ret = rz_core_cmd0(core, cmd);
+	return !ret ? RZ_CMD_STATUS_OK : RZ_CMD_STATUS_ERROR;
+}
+
+RZ_IPI RzCmdStatus rz_clear_history_handler(RzCore *core, int argc, const char **argv) {
+	if (argc != 1) {
```
XVilka: I think the arguments count now checked by the generic implementation and now unnecessary if they are specified as non-optional in YAML. @ret2libc might correct me on that.

wargio: if it is, i will remove this.

librz/core/cmd_descs/cmd_system.yaml
ret2libc: :+1:

librz/core/cmd_system.c
```diff
@@ -0,0 +1,123 @@
...
+	}
+
+	return rz_strbuf_drain(sb);
+}
+
+RZ_IPI RzCmdStatus rz_system_handler(RzCore *core, int argc, const char **argv) {
+	if (argc == 1) {
+		rz_line_hist_list();
+		return RZ_CMD_STATUS_OK;
+	}
```
ret2libc: Is this still needed?

librz/core/cmd_system.c
```diff
@@ -0,0 +1,123 @@
...
+	rz_core_sysenv_end(core);
+
+	return !ret ? RZ_CMD_STATUS_OK : RZ_CMD_STATUS_ERROR;
+}
+
+RZ_API RzCmdStatus rz_system_to_cons_handler(RzCore *core, int argc, const char **argv) {
+	if (argc < 2) {
+		return RZ_CMD_STATUS_WRONG_ARGS;
+	}
+
```
ret2libc: not necessary

librz/core/cmd_system.c
```diff
@@ -0,0 +1,123 @@
...
+	return !ret ? RZ_CMD_STATUS_OK : RZ_CMD_STATUS_ERROR;
+}
+
+RZ_IPI RzCmdStatus rz_list_or_exec_history_handler(RzCore *core, int argc, const char **argv) {
+	if (argc == 1) {
+		rz_line_hist_list();
+		return RZ_CMD_STATUS_OK;
+	} else if (argc > 2) {
+		return RZ_CMD_STATUS_WRONG_ARGS;
```
ret2libc: not necessary

librz/core/cmd_system.c
ret2libc: yes it is unnecessary.

librz/core/cmd_system.c
```diff
@@ -0,0 +1,123 @@
...
+
+	rz_line_hist_free();
+	return RZ_CMD_STATUS_OK;
+}
+
+RZ_IPI RzCmdStatus rz_history_save_handler(RzCore *core, int argc, const char **argv) {
+	if (argc != 1) {
+		return RZ_CMD_STATUS_WRONG_ARGS;
+	}
+
```
ret2libc: same

librz/core/cmd_system.c
```diff
@@ -0,0 +1,123 @@
...
+					s = p;
+				}
+			}
+			arg += s;
+		}
+		rz_strbuf_appendf(sb, "%s\"", arg);
+	}
+
+	return rz_strbuf_drain(sb);
+}
```
ret2libc: It remains to see whether this works well on windows... I remember that I wanted to switch all this `!` stuff to RzSubprocess API.

librz/core/cmd_system.c
wargio: i forgot to remove this.

librz/core/cmd_system.c
wargio: I can add a comment to refactor this later.

wargio:
```
E: Failed to fetch http://apt.llvm.org/focal/dists/llvm-toolchain-focal-12/main/binary-amd64/Packages  File has unexpected size (9437 != 9427). Mirror sync in progress? [IP: 151.101.250.49 443]
```

test/db/tools/rz
```diff
@@ -98,6 +98,7 @@ EOF
 RUN
 
 NAME=no-nl-at-eof script
+BROKEN=1
```
XVilka: Why?

XVilka (approved):
Overall is good to merge even as is. Just a couple minor things.

librz/core/cmd_descs/cmd_history.yaml
```diff
@@ -0,0 +1,29 @@
+# SPDX-FileCopyrightText: 2021 RizinOrg <info@rizin.re>
+# SPDX-License-Identifier: LGPL-3.0-only
+---
+name: cmd_history
+commands:
+  - name: "H"
+    cname: list_or_exec_history
```
ret2libc: i prefer `history_list_or_exec`, `history_clear` and `history_save`. we usually put the "topic" as first word

librz/core/cmd_history.c
```diff
@@ -0,0 +1,31 @@
+RZ_IPI RzCmdStatus rz_list_or_exec_history_handler(RzCore *core, int argc, const char **argv) {
+	if (argc == 1) {
+		rz_line_hist_list();
+		return RZ_CMD_STATUS_OK;
+	}
+
+	int index = atoi(argv[1]);
+	if (index < 1) {
+		RZ_LOG_ERROR("index must be a positive number.\n");
+		return RZ_CMD_STATUS_WRONG_ARGS;
```
ret2libc: Mmm maybe STATUS_ERROR even in this case? This is no different than passing a too big number.

librz/core/cmd_system.c
```diff
@@ -0,0 +1,92 @@
...
+				if (arg[p] == '\\' || arg[p] == '"') {
+					rz_strbuf_append_n(sb, arg + s, p - s);
+					rz_strbuf_append(sb, "\\");
+					s = p;
+				}
+			}
+			arg += s;
+		}
+		rz_strbuf_appendf(sb, "%s\"", arg);
+	}
```
ret2libc: not a big fan of all this logic actually... I *think* @thestr4ng3r already did something like this somewhere. Am I wrong Florian?

test/db/tools/rz
wargio: because it does work on my linux but not on the linux machine via the CI and i have no clue why

librz/core/cmd_system.c
```diff
@@ -0,0 +1,69 @@
...
+	};
+
+	RzSubprocess *proc = rz_subprocess_start_opt(&opt);
+	if (!proc) {
+		rz_subprocess_fini();
+		return false;
+	}
+
+	// wait max 1 min
+	RzSubprocessWaitReason reason = rz_subprocess_wait(proc, 60000);
```
ret2libc: Why the limit? If you do !vim you don't expect it to be closed after one min

XVilka:
Overall is fine, but could you please run it on more CI configurations, including older CentOS, Debian, Alpine by sending the PR from the `dist-` branch instead?

librz/core/cfile.c
```diff
@@ -492,19 +492,8 @@ RZ_API bool rz_core_file_resize_delta(RzCore *core, st64 delta) {
...
-	if (strstr(cmd, "RZ_BLOCK")) {
-		// remove temporary BLOCK file
-		char *f = rz_sys_getenv("RZ_BLOCK");
-		if (f) {
-			rz_file_rm(f);
-			rz_sys_setenv("RZ_BLOCK", NULL);
-			free(f);
-		}
-	}
+RZ_API void rz_core_sysenv_end(RzCore *core) {
```
XVilka, line 497: Since you modify this function anyway, could you please add a Doxygen description here?

librz/core/cfile.c
```diff
@@ -516,33 +505,12 @@ RZ_API void rz_core_sysenv_end(RzCore *core, const char *cmd) {
...
 
-#if DISCUSS
-EDITOR rz_sys_setenv("EDITOR", rz_config_get(core->config, "cfg.editor"));
-CURSOR cursor position(offset from curseek)
-	VERBOSE cfg.verbose
-#endif
-
-	RZ_API char *rz_core_sysenv_begin(RzCore *core, const char *cmd) {
-	char *f, *ret = cmd ? strdup(cmd) : NULL;
+RZ_API void rz_core_sysenv_begin(RzCore *core) {
```
XVilka, line 511: Sam (was) here

librz/core/cfile.c
XVilka: In fact there is already: `rz_str_bool()`. Moreover, if you just use `rz_config_get(core->config, "cfg.debug")` it should return "true" or "false" string - it checks if the variable is boolean internally already, see `librz/config/config.c`.

librz/core/cmd.c
```diff
@@ -2087,7 +2081,7 @@ static void rz_w32_cmd_pipe(RzCore *core, char *rizin_cmd, char *shell_cmd) {
 #undef __CLOSE_DUPPED_PIPES
 #endif
 
-RZ_API int rz_core_cmd_pipe(RzCore *core, char *rizin_cmd, char *shell_cmd) {
+RZ_API int rz_core_cmd_pipe_old(RzCore *core, char *rizin_cmd, char *shell_cmd) {
```
XVilka, line 1896: Could you please add Doxygen description here?

librz/core/cmd.c
```diff
@@ -2163,6 +2157,59 @@ RZ_API int rz_core_cmd_pipe(RzCore *core, char *rizin_cmd, char *shell_cmd) {
...
+	rz_subprocess_wait(proc, UT64_MAX);
+
+	output = rz_subprocess_out(proc);
+	rz_subprocess_free(proc);
+	rz_subprocess_fini();
+
+	return output;
+}
+
+RZ_API RzCmdStatus rz_core_cmd_pipe(RzCore *core, char *rizin_cmd, int argc, char **argv) {
```
XVilka: Since this is new API function, add Doxygen please.

XVilka (changes requested):

librz/core/cfile.c
wargio: that code is old. i will revert it back to the original code since i dropped due the new usage of the RzSubprocess

librz/core/cfile.c
wargio: dropping support for this.

librz/core/cfile.c
wargio: same. dropping support.

librz/core/cmd.c
wargio: not going to happen, since it is old code and likely being removed once the code does work.

librz/core/cmd.c
wargio: not sure if we should support this as an API.

XVilka: Probably not.

librz/include/rz_util/rz_subprocess.h
```diff
@@ -72,8 +72,8 @@ RZ_API void rz_subprocess_free(RzSubprocess *proc);
 RZ_API RzSubprocessWaitReason rz_subprocess_wait(RzSubprocess *proc, ut64 timeout_ms);
 RZ_API void rz_subprocess_kill(RzSubprocess *proc);
 RZ_API int rz_subprocess_ret(RzSubprocess *proc);
-RZ_API char *rz_subprocess_out(RzSubprocess *proc);
-RZ_API char *rz_subprocess_err(RzSubprocess *proc);
+RZ_API char *rz_subprocess_out(RzSubprocess *proc, int *length);
```
ret2libc: what about
```suggestion
RZ_API char *rz_subprocess_out(RzSubprocess *proc, size_t *length);
```
?

librz/util/subprocess.c
```diff
@@ -632,7 +620,7 @@ static char **create_child_env(const char *envvars[], const char *envvals[], siz
 	for (size_t i = 0; i < size; i++) {
 		new_env[i] = strdup(environ[i]);
 	}
-	for (size_t i = 0; i <= env_size; i++) {
+	for (size_t i = 0; i <= new_env_size; i++) {
```
ret2libc, line 639: nice catch... I guess it was wrong before.

test/db/tools/rz_bin
```diff
@@ -155,87 +155,87 @@ RUN
...
+CMDS=!!rz-bin -zz ${RZ_FILE} | grep "Password:"
 EXPECT=<<EOF
 86  0x00002619 0x00404019 10  11   .rdata  ascii   Password: 
 EOF
 RUN
 
 NAME=rz-bin -zzz pe
 FILE=bins/pe/ioli/w32/crackme0x00.exe
-CMDS=!rz-bin -zzz ${RZ_FILE} | grep "Password:"
+CMDS=!!rz-bin -zzz ${RZ_FILE} | grep "Password:"
```
ret2libc: I guess this kind of change is now required because the `|` is handled by the rz-shell, so you have to redirect the output/err to rzcons, so that it can be later sent to `grep`, right?

librz/core/cfile.c
```diff
@@ -561,16 +529,14 @@ CURSOR cursor position(offset from curseek)
...
 	rz_sys_setenv("RZ_CONFIG", config_sdb_path);
-
 	rz_sys_setenv("RZ_BIN_LANG", rz_config_get(core->config, "bin.lang"));
 	rz_sys_setenv("RZ_BIN_DEMANGLE", rz_config_get(core->config, "bin.demangle"));
 	rz_sys_setenv("RZ_ARCH", rz_config_get(core->config, "asm.arch"));
 	rz_sys_setenv("RZ_BITS", sdb_fmt("%" PFMT64u, rz_config_get_i(core->config, "asm.bits")));
-	rz_sys_setenv("RZ_COLOR", rz_config_get_i(core->config, "scr.color") ? "1" : "0");
-	rz_sys_setenv("RZ_DEBUG", rz_config_get_b(core->config, "cfg.debug") ? "1" : "0");
-	rz_sys_setenv("RZ_IOVA", rz_config_get_i(core->config, "io.va") ? "1" : "0");
+	rz_sys_setenv("RZ_COLOR", sdb_fmt("%" PFMT64u, rz_config_get_i(core->config, "scr.color")));
```
ret2libc: `rz_strf` please.

librz/core/cmd.c
```diff
@@ -2163,6 +2157,59 @@ RZ_API int rz_core_cmd_pipe(RzCore *core, char *rizin_cmd, char *shell_cmd) {
...
+	if (!rz_subprocess_init()) {
+		RZ_LOG_ERROR("Cannot initialize subprocess.\n");
+		return NULL;
+	}
+
+	RzSubprocessOpt opt = {
+		.file = argv[0],
+		.args = (const char **)&argv[1],
+		.args_size = argc - 1,
+		.envvars = NULL,
```
ret2libc: do we want to pass some vars to the subprocess? `RZ_FILE`, `RZ_OFFSET`, `RZ_ENDIAN`, etc.?

librz/core/cmd.c
ret2libc: please anyway a small description would be good IMHO.

librz/core/cmd.c
```diff
@@ -6007,17 +6054,22 @@ DEFINE_HANDLE_TS_FCN_AND_SYMBOL(html_enable_stmt) {
...
+	TSNode command_pipe = ts_node_named_child(node, 1);
+
+	char *rz_cmd = ts_node_sub_string(command_rizin, state->input);
+	if (!rz_cmd) {
+		return RZ_CMD_STATUS_INVALID;
+	}
+
+	RzCmdStatus res = RZ_CMD_STATUS_INVALID;
+	RzCmdParsedArgs *a = ts_node_handle_arg_prargs(state, node, command_pipe, 1, true);
+	if (a && a->argc > 1) {
```
ret2libc, line 5908: I would check `if (!a || a->argc <= 1) return INVALID`. Otherwise you just ignore the error, which I think is never good.

librz/core/cmd.c
```diff
@@ -6563,6 +6614,27 @@ RZ_API char *rz_core_cmd_str(RzCore *core, const char *cmd) {
 	return retstr;
 }
 
+RZ_API ut8 *rz_core_cmd_raw(RzCore *core, const char *cmd, int *length) {
+	const char *static_str;
+	ut8 *retstr = NULL;
+	rz_cons_push();
+	if (rz_core_cmd(core, cmd, 0) == -1) {
+		//eprintf ("Invalid command: %s\n", cmd);
```
ret2libc: `rz_cons_pop`

ret2libc (commented):
Awesome work @wargio , thanks a lot for moving this thing forward! I haven't tested it yet, I will do after some more changes, however things look nice so far!

librz/include/rz_util/rz_subprocess.h
wargio: i thought about this, but RzStrbuf does not use size_t.

librz/util/subprocess.c
wargio: i had a huge headache on this since asan was complaining but not telling me where due optimizations.

test/db/tools/rz_bin
wargio: Yes. because !! does fetch stdout, and ! does not

librz/core/cfile.c
wargio: i'll reset this back to the original code and even maybe remove this.

librz/core/cmd.c
wargio: hmm i do not know, because i expect the app to eat the stdin not the envvars.

ret2libc: according to `!??` programs are expected to find some env vars. It seems that in the past implementation that was all set with `rz_core_sysenv_begin`/`rz_core_sysenv_end`. It should be fine to keep this as-is, because current implementation of RzSubprocess just allows to add/replace env variables, but not to remove them. So by setting `.envvars = NULL`  we are really just keeping the parent env vars, which should be fine.

librz/util/subprocess.c
ret2libc: compile with `-buildtype=debug` if that was the problem.

librz/include/rz_util/rz_subprocess.h
ret2libc: :/ ok

librz/core/cmd.c
wargio: i saw that you added this in your PR, should i skip this change ? @ret2libc

librz/core/cmd.c
wargio: that is not true. `res` is `RZ_CMD_STATUS_INVALID` so if that `if` fails, it will return `RZ_CMD_STATUS_INVALID`

XVilka (approved):
LGTM. But please don't forget to test on `dist-` branch before merging.

XVilka:
94 broken tests: https://builds.sr.ht/~xvilka/job/534811#task-test-0

librz/core/cmd.c
ret2libc: I will just do a separate PR just with this ;) my PR will take some time to be done.

librz/core/cmd.c
ret2libc: Right!

librz/core/cfile.c
```diff
@@ -563,16 +533,14 @@ CURSOR cursor position(offset from curseek)
...
 	rz_sys_setenv("RZ_BIN_DEMANGLE", rz_config_get(core->config, "bin.demangle"));
 	rz_sys_setenv("RZ_ARCH", rz_config_get(core->config, "asm.arch"));
-	rz_sys_setenv("RZ_BITS", sdb_fmt("%" PFMT64u, rz_config_get_i(core->config, "asm.bits")));
-	rz_sys_setenv("RZ_COLOR", rz_config_get_i(core->config, "scr.color") ? "1" : "0");
-	rz_sys_setenv("RZ_DEBUG", rz_config_get_b(core->config, "cfg.debug") ? "1" : "0");
-	rz_sys_setenv("RZ_IOVA", rz_config_get_i(core->config, "io.va") ? "1" : "0");
+	rz_sys_setenv("RZ_BITS", rz_config_get(core->config, "asm.bits"));
+	rz_sys_setenv("RZ_COLOR", rz_config_get(core->config, "scr.color"));
+	rz_sys_setenv("RZ_DEBUG", rz_config_get(core->config, "cfg.debug"));
+	rz_sys_setenv("RZ_IOVA", rz_config_get(core->config, "io.va"));
```
ret2libc: This is not exactly the same as before.. For example, for booleans vars it would returne `true`/`false` instead of `1`/`0` as before. I think for env vars it is more common to have 1/0 ?

librz/core/cmd_history.c
```diff
@@ -0,0 +1,31 @@
+RZ_IPI RzCmdStatus rz_history_list_or_exec_handler(RzCore *core, int argc, const char **argv) {
+	if (argc == 1) {
+		rz_line_hist_list();
+		return RZ_CMD_STATUS_OK;
+	}
+
+	int index = atoi(argv[1]);
```
ret2libc, line 7: `rz_num`?

librz/core/cmd_system.c
```diff
@@ -0,0 +1,217 @@
...
+	const char *bin_lang = rz_config_get(core->config, "bin.lang");
+	const char *cfg_debug = rz_config_get(core->config, "cfg.debug");
+	const char *io_va = rz_config_get(core->config, "io.va");
+	const char *pdb_server = rz_config_get(core->config, "pdb.server");
+	const char *scr_color = rz_config_get(core->config, "scr.color");
+	const char *endian = rz_str_bool(core->rasm->big_endian);
+	char *cfg_path = config_path(core);
+	snprintf(file_size, sizeof(file_size), "%" PFMT64u, core->file ? rz_io_fd_size(core->io, core->file->fd) : 0);
+	snprintf(core_offset, sizeof(core_offset), "%" PFMT64u, core->offset);
+	snprintf(block_size, sizeof(block_size), "%u", core->blocksize);
```
ret2libc:
```suggestion
	rz_strf(file_size, "%" PFMT64u, core->file ? rz_io_fd_size(core->io, core->file->fd) : 0);
	rz_strf(core_offset, "%" PFMT64u, core->offset);
	rz_strf(block_size, "%u", core->blocksize);
```

librz/core/cmd_system.c
```diff
@@ -0,0 +1,217 @@
...
+		core_offset,
+		endian,
+		bin_demangle,
+		bin_lang,
+		pdb_server,
+		io_va,
+		scr_color,
+		block_size,
+		cfg_debug,
+		cfg_path ? cfg_path : ""
```
ret2libc:
```suggestion
		rz_str_get(cfg_path)
```

librz/core/cmd_system.c
```diff
@@ -0,0 +1,217 @@
...
+		return false;
+	}
+
+	if (!rz_subprocess_init()) {
+		RZ_LOG_ERROR("Cannot initialize subprocess.\n");
+		rz_list_free(alloc);
+		rz_file_rm(cfg_path);
+		free(cfg_path);
+		free(args);
+		return false;
```
ret2libc: put this at the end of the function and use `goto`, it will simplify errors handling and make sure things are done in the right order on each error.

librz/core/cmd_system.c
```diff
@@ -0,0 +1,217 @@
...
+	} else if (strlen(env) == len) {
+		return value;
+	}
+
+	if ((it = rz_list_find_ptr(alloc, (void *)arg))) {
+		string = it->data;
+	} else {
+		string = strdup(arg);
+	}
+
```
ret2libc: I don't see what advantages `alloc` array would bring. I would just go with `rz_str_replace` and remove `alloc`. Moreover, you can only find the same `arg` in `alloc` only if it's the same argv[i], so probably a list is not really necessary anyway.

ret2libc (commented):
Just looked at some parts for now! Doing another round in a bit.

librz/core/cmd.c
```diff
@@ -2163,6 +1925,66 @@ RZ_API int rz_core_cmd_pipe(RzCore *core, char *rizin_cmd, char *shell_cmd) {
...
+	return output;
+}
+
+/**
+ * \brief Executes a rizin command and pipes the result to the stdin when calling execvp
+ *
+ * Executes a rizin command and the raw bytes of the stdout is returned.
+ * The output is then used as stdin for execvp.
+ * The output of the execvp is then sent into RzCons.
+ * */
```
ret2libc: This should be better explained. In particular, it is not clear what is piped where. In this case, `rizin_cmd` is the first command and its output is piped as stdin of the command specified in (argc, argv).

librz/core/cmd.c
ret2libc: This is still to be done.

librz/core/cmd.c
```diff
@@ -6565,6 +6391,32 @@ RZ_API char *rz_core_cmd_str(RzCore *core, const char *cmd) {
 	return retstr;
 }
 
+/**
+ * \brief Executes a rizin command and returns the raw stdout and its length
+ *
+ * Executes a rizin command and returns the raw stdout and its length.
+ * */
+RZ_API ut8 *rz_core_cmd_raw(RzCore *core, const char *cmd, int *length) {
```
ret2libc, line 6474: Probably this can be unified with rz_core_cmd_str to avoid code dup.

librz/core/cfile.c
wargio: no. this better and this stuff is not used besides the user itself.

librz/core/cmd_history.c
wargio: no. i expect a number and i want a number. atoi is enough

librz/core/cmd_system.c
wargio: alloc, is just a simple idea where i place the allocated strings and just free them via rz_list so i am sure that i never forget to free an allocated value, so i do not have to care whether argv is allocated on the heap or not, using directly rz_str_replace is a super bad idea since you do not know where those arguments are really allocated.

librz/core/p/core_plugin_example.c
```diff
@@ -1,41 +1,93 @@
 // SPDX-FileCopyrightText: 2021 RizinOrg <info@rizin.re>
 // SPDX-License-Identifier: LGPL-3.0-only
-#if 0
-gcc -o core_plugin_example.so -O3 -std=c99 -Wall -fPIC `pkg-config --cflags --libs rz_core` core_plugin_example.c -shared
-mkdir -p ~/.config/rizin/plugins
-mv core_plugin_example.so ~/.config/rizin/plugins
-#endif
+
```
ret2libc, line 3: Awesome! Is this compiled anyhow while testing rizin? I was also wondering.. Aren't unit tests enough to showcase how to write a new command?

test/db/io/baddr
```diff
@@ -17,7 +17,7 @@ RUN
 
 NAME=3: io.baddr
 FILE=bins/elf/analysis/pie
-CMDS=!rz-bin -B 0xf00000 -re ${RZ_FILE} | grep 0x
+CMDS=!!rz-bin -B 0xf00000 -re ${RZ_FILE} | grep 0x
```
ret2libc: I think this should be kinda automatic or, at least, the default behaviour, without having to use `!!` all the time.
If we read the help of `|` it is `| <cmd> | <program> # Pipe output of command to program`, so by using `|` the user already express his will to pipe the output of one "command" to another program. 

Do you think it would make sense to detect the `pipe` and assume output should be redirected to rzcons?

librz/util/subprocess.c
```diff
@@ -36,13 +36,17 @@ static bool create_pipe_overlap(HANDLE *pipe_read, HANDLE *pipe_write, LPSECURIT
 		CloseHandle(*pipe_read);
 		return FALSE;
 	}
+
+	SetEnvironmentVariable(TEXT("RZ_PIPE_PATH"), name);
```
ret2libc, line 40: Why is this necessary? What does it do?

librz/core/p/core_plugin_example.c
wargio: yes

test/db/io/baddr
wargio: i disagree
you can't know if this is piped, or not

librz/util/subprocess.c
```diff
@@ -132,29 +136,29 @@ static LPWCH override_env(const char *envvars[], const char *envvals[], size_t e
...
+	} else {
+		snprintf(cmd_exe, sizeof(cmd_exe), "%s", opt->file);
+	}
 
 	char **argv = calloc(opt->args_size + 1, sizeof(char *));
 	if (!argv) {
 		return NULL;
 	}
-	argv[0] = (char *)opt->file;
+	argv[0] = ""; // a space is required to work correctly.
```
ret2libc, line 161: why the `cmd_exe` is not placed in `argv[0]`?

librz/core/p/core_plugin_example.c
ret2libc: Where is this compiled?

test/db/io/baddr
ret2libc: Why can't you? In `pipe_stmt` handler you know if a program is piped or not.

librz/core/cmd.c
```diff
@@ -6008,17 +5830,22 @@ DEFINE_HANDLE_TS_FCN_AND_SYMBOL(html_enable_stmt) {
...
-	char *second_str = ts_node_sub_string(second_cmd, state->input);
-	int value = state->core->num->value;
-	RzCmdStatus res = rz_cmd_int2status(rz_core_cmd_pipe(state->core, first_str, second_str));
-	state->core->num->value = value;
-	free(first_str);
-	free(second_str);
+	TSNode command_rizin = ts_node_named_child(node, 0);
+	TSNode command_pipe = ts_node_named_child(node, 1);
+
+	char *rz_cmd = ts_node_sub_string(command_rizin, state->input);
```
ret2libc: This is probably unnecessary. We can probably use `handle_ts_stmt` as done elsewhere to direclty work with the TSNode, so we avoid going again through the whole tree-sitter parsing and handling.

test/unit/test_rzpipe.c
```diff
@@ -11,7 +11,7 @@ static bool test_rzpipe(void) {
 	RzPipe *r = rzpipe_open(RIZIN_BUILD_PATH " -q0 =");
 	mu_assert("rzpipe can spawn", r);
 	char *hello = rzpipe_cmd(r, "?e hello world");
-	mu_assert_streq(hello, "hello world\n", "rzpipe hello world");
+	mu_assert_streq(hello, "[0x00000000]> hello world\n", "rzpipe hello world");
```
ret2libc: I think this change is wrong. The output of the `?e` command is just the string, not the prompt + string.

librz/core/cfile.c
ret2libc: How is this better exactly? Anyway, I'm ok with having `true`/`false` instead of 1/0, but please update the documentation because it says 1/0.

librz/util/subprocess.c
wargio: because we are calling `CreateProcessA` properly. check MS guide.

librz/util/subprocess.c
wargio: used by rz-pipe on windows.

test/db/io/baddr
wargio: yes, but what if my cmd is like 
```
[0x00000000] `aaa | bbb` | ccc
```

librz/util/subprocess.c
ret2libc: I see :/ Does linux needs it too?

librz/core/cfile.c
wargio: book needs to be updated anyway after this PR

ret2libc: I meant the rizin doc for `%??`.

By the way, I changed my mind. I think there is no need to do this change here and now. In particular, there are places that seem to assume boolean env vars as 1/0 and not true/false,  so I'll revert these.

librz/util/subprocess.c
wargio: as far i know, nope.

test/db/io/baddr
ret2libc: Not sure about it's better `!!<prog1> | <prog2>` or `!<prog1> | <prog2>`...

Choosing `!!<prog1> | <prog2>` means that essentially you never use `!<prog1> | <prog2>` and you have to type `!!` every time. Also, as a user you first do `!ls` and then you probably want to grep it and suddenly you have to do `!!ls` instead of just `!ls` if you want to pipe. On the other hand, this maintains the semantic of `!` and `!!`.

Choosing `!<prog1> | <prog2>` and make sure that `!<prog1>` automatically redirects stdout to rzcons whenever a pipe is used means that as a user it's easier to construct a command without really having to *know* about rzcons (which is a very internal thing after all..). However, it loses a bit the semantic of `!` vs `!!`.

WDYT @wargio  @XVilka @thestr4ng3r ?

librz/util/subprocess.c
```diff
@@ -132,29 +136,29 @@ static LPWCH override_env(const char *envvars[], const char *envvals[], size_t e
...
-	while (*str) {
-		if (str[0] == '\r' &&
-			!(str - start >= 4 && !strncmp(str - 4, RZ_CONS_CLEAR_SCREEN, 4))) {
-			memmove(str, str + 1, strlen(str + 1) + 1);
-			continue;
-		}
-		str++;
-	}
-}
-
```
ret2libc, line 146: @wargio I see this was removed from subprocess, but now old users of the subprocess API has to do this themselves. Did you remove this because you want to get the raw stdout/stderr of a program?

What do you think if we use these API?
```
RZ_API char *rz_subprocess_out_str(RzSubprocess *proc);
RZ_API char *rz_subprocess_err_str(RzSubprocess *proc);
RZ_API ut8 *rz_subprocess_out(RzSubprocess *proc, int *length);
RZ_API ut8 *rz_subprocess_err(RzSubprocess *proc, int *length);
```
The `_str` APIs apply `remove_cr` before returning the strings, so we don't need to replicate this function everywhere.

Together with this change we should also change a bit the `RzSubprocessOutput` structure. Something like:
```
typedef struct rz_process_output_t {
	char *out_str; // stdout
	char *err_str; // stderr
        ut8 *out, *err;
        int out_len, err_len;
	int ret; // exit code of the process
	bool timeout;
} RzSubprocessOutput;
```
Comments?

wargio: stdout needs to be raw. it's up to the user to know what there will be in the buffer.
this is the reason you can do `cat mybinfile | hexdump -C` for example

wargio: The remove_cr function shall not be there because the only one complaining about this is the testsuite, therefore that is a testsuite issue not an issue related to the pipe, nor subprocess

test/db/io/baddr
wargio: for me you can remove `!!` and merge them into `!` and do the autodetect for pipe/grep/etc..

librz/util/subprocess.c
ret2libc: I agree, that's why I kept the raw stdout in the proposal above. However, other times you just want the string and you want a "cleaned" string (e.g. by running remove_cr function), so I was thinking about providing that from RzSubprocess itself instead of having every user of the API definiting their own remove_cr function.

Also, no matter what we should probably add the `out_len`/`err_len` fields.

ret2libc: cc @thestr4ng3r you wrote that piece of code, right?

thestr4ng3r: This `remove_cr` implementation is super inefficient so I can only assume I wrote it to only be used as part of `rz-test`, where it doesn't really matter.
If another user of `RzSubprocess` needs this, the functionality should not be provided as part of the subprocess api but as just an independent `rz_str_` function that can be called.
So for me the move to `rz-test/run.c` as it is done here seems good.

ret2libc: Ok thanks for confirming @thestr4ng3r . Then I will just make sure to add `out_len`/`err_len` to `RzSubprocessOutput` and that's it.

test/db/io/baddr
XVilka: I think autodetection will provide the best user experience, indeed

librz/util/subprocess.c
XVilka: `""` is not a "space" though, it's an empty string.

test/db/cmd/feat_redirect
```diff
@@ -2,9 +2,9 @@ NAME="pb>x;!wc -c x"
 FILE=malloc://1024
 CMDS=<<EOF
 p8 10 > dump
-!wc -c dump|awk "{print \$1}"
+!!wc -c dump|awk "{print \$1}"
```
XVilka: Does this need to be updated too then?

test/db/cmd/cmd_system
```diff
@@ -16,14 +16,14 @@ RUN
 
 NAME=!| grep
 FILE==
-CMDS=!echo magic|grep bob
+CMDS=!!echo magic|grep bob
```
XVilka: And here

XVilka (changes requested):

test/db/cmd/feat_redirect
ret2libc: Yeah all these changes can be reverted back if we are ok with the auto detection.

librz/core/cmd_system.c
```diff
@@ -143,7 +143,7 @@ static bool system_exec(RzCore *core, int argc, const char **argv, char **output
 		.envvals = envvals,
 		.env_size = RZ_ARRAY_SIZE(envvars),
 		.stdin_pipe = RZ_SUBPROCESS_PIPE_NONE,
-		.stdout_pipe = output ? RZ_SUBPROCESS_PIPE_CREATE : RZ_SUBPROCESS_PIPE_NONE,
+		.stdout_pipe = force_output || core->is_pipe ? RZ_SUBPROCESS_PIPE_CREATE : RZ_SUBPROCESS_PIPE_NONE,
```
wargio: isn't easier to check for the output ptr?

ret2libc: I was not sure it is possible to do that way, because when you execute `!<cmd>` in some cases you want the output, in others you don't want it. So from rz_system_handler you have to always pass the `output` var.

wargio: yes, but the output needs to go somewhere and return it, so the output pointer MUST be non NULL.

wargio: i would move the core->is_pipe check one function above, then pass the pointer accordingly to the check.
i mean, this way also works.

XVilka:
```
clang-cl" "-Ibinrz/rz-test\rz-test.exe.p" "-I." "-I.." "-I..\librz\include" "-Isubprojects\sdb" "-I..\subprojects\sdb" "-Isubprojects\sdb\src" "-I..\subprojects\sdb\src" "-I..\shlr\spp" "/MD" "/nologo" "/showIncludes" "/W2" "/O2" "-fcommon" "-D__STDC__=1" "-D_CRT_DECLARE_NONSTDC_NAMES " "-D_CRT_SECURE_NO_WARNINGS" "-D_CRT_NONSTDC_NO_DEPRECATE" "-Werror=sizeof-pointer-memaccess" "/Fdbinrz/rz-test\rz-test.exe.p\run.c.pdb" /Fobinrz/rz-test/rz-test.exe.p/run.c.obj "/c" ../binrz/rz-test/run.c
../binrz/rz-test/run.c(12,10): error: 'unsigned char *' and 'char *' are not pointers to compatible types
                        !(str - start >= 4 && !strncmp(str - 4, RZ_CONS_CLEAR_SCREEN, 4))) {
                          ~~~ ^ ~~~~~
../binrz/rz-test/run.c(12,35): warning: passing 'unsigned char *' to parameter of type 'const char *' converts between pointers to integer types with different sign [-Wpointer-sign]
                        !(str - start >= 4 && !strncmp(str - 4, RZ_CONS_CLEAR_SCREEN, 4))) {
                                                       ^~~~~~~
C:\Program Files (x86)\Windows Kits\10\include\10.0.19041.0\ucrt\string.h(272,44): note: passing argument to parameter '_Str1' here
    _In_reads_or_z_(_MaxCount) char const* _Str1,
                                           ^
../binrz/rz-test/run.c(13,33): warning: passing 'unsigned char *' to parameter of type 'const char *' converts between pointers to integer types with different sign [-Wpointer-sign]
                        memmove(str, str + 1, strlen(str + 1) + 1);
                                                     ^~~~~~~
C:\Program Files (x86)\Windows Kits\10\include\10.0.19041.0\ucrt\string.h(216,24): note: passing argument to parameter '_Str' here
    _In_z_ char const* _Str
                       ^
2 warnings and 1 error generated.
[1510/1682] Linking target test/unit/auxiliary/subprocess-interactive.exe
```

XVilka:
```
"cl" "-Ibinrz/rz-test\rz-test.exe.p" "-I." "-I.." "-I..\librz\include" "-Isubprojects\sdb" "-I..\subprojects\sdb" "-Isubprojects\sdb\src" "-I..\subprojects\sdb\src" "-I..\shlr\spp" "/MT" "/nologo" "/showIncludes" "/W2" "/O2" "/Gw" "/Fdbinrz/rz-test\rz-test.exe.p\run.c.pdb" /Fobinrz/rz-test/rz-test.exe.p/run.c.obj "/c" ../binrz/rz-test/run.c
../binrz/rz-test/run.c(9): error C2146: syntax error: missing ')' before identifier 's'
../binrz/rz-test/run.c(9): error C2059: syntax error: ')'
[1436/1658] Compiling C object binrz/rz-test/rz-test.exe.p/rz-test.c.obj
[1437/1658] Compiling C object binrz/rz-test/rz-test.exe.p/load.c.obj
```

XVilka:
```
FAILED: test/unit/test_subprocess.exe.p/test_subprocess.c.obj 
"clang-cl" "-Itest/unit\test_subprocess.exe.p" "-I." "-I.." "-I..\librz\include" "-Itest/unit" "-I..\test\unit" "-Isubprojects\sdb" "-I..\subprojects\sdb" "-Isubprojects\sdb\src" "-I..\subprojects\sdb\src" "-I..\shlr\spp" "-Ishlr" "-I..\shlr" "-I..\librz\asm\arch\include" "-I..\librz\asm\arch" "-I..\librz\asm\arch\h8300" "-I..\librz\asm\arch\hexagon" "-I..\librz\asm\arch\msp430" "-I..\librz\asm\arch\rsp" "-I..\librz\asm\arch\mcore" "-I..\librz\asm\arch\v850" "-I..\librz\asm\arch\propeller" "-I..\librz\asm\arch\ebc" "-I..\librz\asm\arch\cr16" "-I..\librz\asm\arch\8051" "-I..\librz\asm\arch\v810" "-I..\librz\asm\arch\or1k" "-I..\librz\analysis\arch\gb" "-I..\librz\analysis\arch\hexagon" "-I..\librz\bin\format\elf" "-I..\librz\bin\mangling" "-I..\librz\bin\format" "-I..\shlr\heap\include\rz_windows" "-I..\librz\type\parser" "/MD" "/nologo" "/showIncludes" "/W2" "/O2" "-fcommon" "-D__STDC__=1" "-D_CRT_DECLARE_NONSTDC_NAMES " "-D_CRT_SECURE_NO_WARNINGS" "-D_CRT_NONSTDC_NO_DEPRECATE" "-Werror=sizeof-pointer-memaccess" "/Fdtest/unit\test_subprocess.exe.p\test_subprocess.c.pdb" /Fotest/unit/test_subprocess.exe.p/test_subprocess.c.obj "/c" ../test/unit/test_subprocess.c
../test/unit/test_subprocess.c(17,10): error: 'unsigned char *' and 'char *' are not pointers to compatible types
                        !(str - start >= 4 && !strncmp(str - 4, RZ_CONS_CLEAR_SCREEN, 4))) {
                          ~~~ ^ ~~~~~
../test/unit/test_subprocess.c(17,35): warning: passing 'unsigned char *' to parameter of type 'const char *' converts between pointers to integer types with different sign [-Wpointer-sign]
                        !(str - start >= 4 && !strncmp(str - 4, RZ_CONS_CLEAR_SCREEN, 4))) {
                                                       ^~~~~~~
C:\Program Files (x86)\Windows Kits\10\include\10.0.19041.0\ucrt\string.h(272,44): note: passing argument to parameter '_Str1' here
    _In_reads_or_z_(_MaxCount) char const* _Str1,
                                           ^
../test/unit/test_subprocess.c(18,33): warning: passing 'unsigned char *' to parameter of type 'const char *' converts between pointers to integer types with different sign [-Wpointer-sign]
                        memmove(str, str + 1, strlen(str + 1) + 1);
                                                     ^~~~~~~
C:\Program Files (x86)\Windows Kits\10\include\10.0.19041.0\ucrt\string.h(216,24): note: passing argument to parameter '_Str' here
    _In_z_ char const* _Str
                       ^
2 warnings and 1 error generated.
[1659/1682] Compiling C object test/unit/test_task.exe.p/test_task.c.obj
[1660/1682] Compiling C object test/unit/test_tree.exe.p/test_tree.c.obj
ninja: build stopped: subcommand failed
```

ret2libc (approved):
To me things are in good shape right now. Good work @wargio !! Please review this as well, in particular the latest changes I did.

XVilka (approved):

librz/core/p/core_plugin_example.c
wargio: on my machine.

PR #1255, opened 2021-06-30
Support self-unhooking of RzEvent Callbacks

thestr4ng3r:

**Detailed description**

Previously, when calling `rz_event_unhook()` inside an event callback itself, it would mess up the internal callback vectors while they are still being iterated over. I solved this by deferring the actual unhooking to after the send is complete if necessary.

There is one implementation detail:
* When two callbacks A and B are registered for the same event and
* B is after A in the vector and
* A unhooks B
* then B will still be still be called with this single event one last time, happening after A's request for unhook.

This has the following implications:
Pro: the fact whether B will be called on this specific event does not depend on the order in which the callbacks have been registered.
Con: If the unhook out of A comes from deleting/freeing something that B accesses, B might uaf depending on how it is implemented.

Since it's not completely clear whether not calling B in this case is "better" than this behavior and preventing the B call would need additional logic (which however can easily be implemented), I decided to leave this open for now and not specify any behavior for it yet, until we find a case where it is relevant.

**Test plan**

see the added unit test

### DIFF
--- a/librz/include/rz_util/rz_event.h
+++ b/librz/include/rz_util/rz_event.h
@@ -16,6 +16,7 @@ typedef struct rz_event_t {
 	bool incall;
 	HtUP *callbacks;
 	RzVector all_callbacks;
+	RzVector /*<RzEventCallbackHandle>*/ pending_unhook; ///< while inside of a call and a handle is unhooked, the unhook is deferred and saved here
 	int next_handle;
 } RzEvent;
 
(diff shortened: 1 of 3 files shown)


ret2libc (commented):
Seems fine to me. Did you consider just duplicating the structures and iterating over the dups instead of using the pending_unhook approach?

thestr4ng3r:
Yes. The reason I explicitly decided against it is that the `rz_event_send()` should be as lightweight as possible in the default case, which is when no unhooking happens, since events might be triggered extremely frequently, for example during analysis. Then dupping might add significant overhead.
With the `pending_unhook` approach, the only overhead that is added in the default case is a single check for the pending vector length.

wargio (approved):

ret2libc (approved):

PR #1269, opened 2021-07-05
Remove ELF_STRING_LENGTH from librz/bin/format/elf/*

### LINKED ISSUE #1268: Elf symbol name are truncated when >= 256
# This template is meant for bug reports, if you have a feature request, please be as descriptive as possible and delete the template

Make sure you are testing using the latest git version of rizin before submitting any issue.

*If you would like to report a bug, please fill the template below*

### Work environment

| OS/arch/bits (mandatory)                             | Archlinux x64
| Architecture/bits of the file (mandatory)            | x86/64
| `rizin -v` full output, **not truncated** (mandatory)  | rizin 0.3.0-git @ linux-x86-64 commit: c7395f56abb5bbe651c72546649abf52a3bd5eb2, build: 2021-07-04__17:09:53

### Expected behavior

Same result with long-symbol.elf
```
> readelf --symbols --wide long-symbol.elf | grep AAA
    28: 0000000000001139    15 FUNC    GLOBAL DEFAULT   14 AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA
```

PS: gcc seems to use a hard cap of 500 byte.

### Actual behavior

```
> rizin long-symbol.elf
 -- Add comments using the ';' key in visual mode or the 'CC' command from the rizin shell
[0x00001040]> is~AAA
28  0x00001139 0x00001139 GLOBAL FUNC   15       AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA
```

long-symbol.elf
```
#include <stdio.h>

int AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA(int a) {
	return a + 2;
}

int main(int argc, char **argv) {
	int v = AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA(argc);
	printf("%d\n", v);
	return 0;
}
```

gcc long-symbol.elf

[long-symbol.zip](https://github.com/rizinorg/rizin/files/6765417/long-symbol.zip)

ghost:
**Dont squash me**

 

**Detailed description**

* Remove the ELF_STRING_LENGTH
* RzBinElfSymbols are leaking (will be fixed in another PR)
* waiting https://github.com/rizinorg/rizin-testbins/pull/37
**Test plan**

```
> rizin long-symbols.elf
is~AAA
```

**Closing issues**

Close https://github.com/rizinorg/rizin/issues/1268


librz/bin/format/elf/elf_notes.c
```diff
@@ -75,22 +72,24 @@ static bool parse_note_file(RzBinElfNote *note, Elf_(Nhdr) * note_segment_header
...
-			break;
+			rz_vector_fini(&files);
+			return false;
 		}
 
-		char tmp[ELF_STRING_LENGTH] = { 0 };
 		ut64 string_offset = offset + strings_offset + strings_array_len;
-		if (rz_buf_read_at(bin->b, string_offset, (ut8 *)tmp, ELF_STRING_LENGTH - 1) < 0) {
-			break;
+		char *name = rz_buf_get_string(bin->b, string_offset);
```
ret2libc: This might be risky as it is now... Potentially we could end up reading the whole binary into memory if there is no NULL byte in the binary.

librz/bin/format/elf/elf_strtab.c
```diff
@@ -5,6 +5,16 @@
...
+RZ_BORROW const char *Elf_(rz_bin_elf_strtab_get)(RZ_NONNULL RzBinElfStrtab *strtab, ut64 index) {
+	rz_return_val_if_fail(strtab, false);
+
+	if (!Elf_(rz_bin_elf_strtab_has_index)(strtab, index)) {
+		return NULL;
+	}
+
+	return strtab->data + index;
+}
+
```
ret2libc, line 17: Nice that everything passes through this function and that the returned string is already RZ_BORROW. This could be a nice opportunity to reduce memory usage in the future. In particular we might return RzBinElfString data that is ref/unref as needed and `rz_bin_elf_strtab_get` could use a limited amount of space and read only things as needed, instead of reading everything in strtab->data.

Anyway, nice!

librz/bin/format/elf/elf_notes.c
XVilka: Can we limit the reading with the section size instead?

ret2libc: I think section size is wrong. There may be edge cases where things cross the section/segments I think, as long as they are consecutive in the virtual space.

ghost: There is no way to limit the string size. But there is a check at the beginning of the loop.

ret2libc: Oh, actually this is ELF notes... so we could in theory limit it with the size of the section, I think.

ghost: There is already the check at the beginning of the loop. What i was trying to say is that there isn't any hard cap or defined string size.

ret2libc: The check at the beginning of the loop does not prevent to read the whole bin.. It's not a huge deal, but I think we should have a `rz_buf_get_nstring(RzBuffer *b, ut64 addr, size_t max_size)`, where we would call this with max_size=note_section->size`

ghost: There isn't any function rz_buf_get_nstring right now. Should we wait, merge the PR or add the function on this PR?

ret2libc: As you prefer ;) Your call. As said, it's probably not a very big deal so far, so choose what you think is better.

ghost: I add the function and update elf_notes.c

wargio (approved):

XVilka (approved):

librz/util/buf.c
```diff
@@ -408,37 +408,44 @@ RZ_API bool rz_buf_append_buf_slice(RzBuffer *b, RzBuffer *a, ut64 offset, ut64
...
-			null_found = true;
-			break;
+#define GET_STRING_BUFFER_SIZE 32
+
+RZ_API char *rz_buf_get_nstring(RzBuffer *b, ut64 addr, size_t size) {
+	RzStrBuf *buf = rz_strbuf_new(NULL);
+
+	while (true) {
+		char tmp[GET_STRING_BUFFER_SIZE];
+		st64 r = rz_buf_read_at(b, addr, (ut8 *)tmp, RZ_ARRAY_SIZE(tmp));
```
ret2libc:
```suggestion
		st64 r = rz_buf_read_at(b, addr, (ut8 *)tmp, sizeof(tmp));
```

ret2libc: Also, here there should be a condition to choose between `sizeof(tmp)` and the remaining bytes to read. Otherwise I think you could end up with a longer string than what you requested.

librz/util/buf.c
```diff
@@ -408,37 +408,44 @@ RZ_API bool rz_buf_append_buf_slice(RzBuffer *b, RzBuffer *a, ut64 offset, ut64
...
 	}
-	return (char *)res;
+
+	char *result = rz_strbuf_drain(buf);
+	return result;
+}
+
+// return an heap-allocated string read from the RzBuffer b at address addr. The
+// length depends on the first '\0' found in the buffer. If there is no '\0' in
+// the buffer, there is no string, thus NULL is returned.
```
ret2libc: Use doxygen-style doc please.

librz/util/buf.c
```diff
@@ -408,37 +408,44 @@ RZ_API bool rz_buf_append_buf_slice(RzBuffer *b, RzBuffer *a, ut64 offset, ut64
...
-	st64 r = rz_buf_read_at(b, addr, res, MIN_RES_SZ);
-	bool null_found = false;
-	while (r > 0) {
-		const ut8 *needle = rz_mem_mem(res + sz, r, (ut8 *)"\x00", 1);
-		if (needle) {
-			null_found = true;
-			break;
+#define GET_STRING_BUFFER_SIZE 32
+
+RZ_API char *rz_buf_get_nstring(RzBuffer *b, ut64 addr, size_t size) {
```
ret2libc, line 424: Add doxygen-style doc

test/unit/test_buf.c
```diff
@@ -1055,6 +1055,26 @@ bool test_rz_buf_get_string_nothing(void) {
...
 }
 
+bool test_rz_buf_get_nstring(void) {
+	ut8 *ch = malloc(128);
+	memset(ch, 'A', 127);
+	ch[127] = '\0';
+	ch[111] = '\0';
+	RzBuffer *b = rz_buf_new_with_bytes(ch, 128);
+	char *s = rz_buf_get_nstring(b, 100, 10);
+	mu_assert_null(s, "there is no string with size < 100 (no null terminator)");
```
ret2libc:
```suggestion
	mu_assert_null(s, "there is no string with size < 10 (no null terminator)");
```

test/unit/test_buf.c
```diff
@@ -1055,6 +1055,26 @@ bool test_rz_buf_get_string_nothing(void) {
...
+	free(s);
+	s = rz_buf_get_nstring(b, 100, 11);
+	mu_assert_streq(s, (char *)ch + 100, "the string is the same");
+	free(s);
+	s = rz_buf_get_nstring(b, 0, 128);
+	mu_assert_streq(s, (char *)ch, "the string is empty");
```
ret2libc:
```suggestion
	mu_assert_streq(s, (char *)ch, "the full string is returned");
```

test/unit/test_buf.c
```diff
@@ -1055,6 +1055,26 @@ bool test_rz_buf_get_string_nothing(void) {
 	mu_end;
 }
 
+bool test_rz_buf_get_nstring(void) {
```
ret2libc, line 1058: add also a call to `rz_buf_get_nstring` with a size >= 32 but less than 64. You should check that the returned string is not longer than the expected size.

ret2libc (commented):
Nice! Thanks for also adding the unit test for rz_buf_get_nstring, very much appreciated.

Just few things, but it looks good to me!

test/unit/test_buf.c
ghost: I add a test, what do you think?

librz/util/buf.c
ghost: I add a doc

librz/util/buf.c
ghost: I add a doc

librz/util/buf.c
ghost: Why should I use sizeof(tmp) and not RZ_ARRAY_SIZE?

XVilka:
@officialcjunior looks like it's broken on static test, please take a look:
```
84/85 cpu_platform_profiles          FAIL            0.24s   killed by signal 11 SIGSEGV
>>> MALLOC_PERTURB_=51 /__w/rizin/rizin/rizin/build/test/integration/test_cpu_platform_profiles

85/85 open_analyse_save_load_project OK              1.73s

Summary of Failures:

84/85 cpu_platform_profiles          FAIL            0.24s   killed by signal 11 SIGSEGV


Ok:                 84  
Expected Fail:      0   
Fail:               1   
Unexpected Pass:    0   
Skipped:            0   
Timeout:            0   
```

librz/util/buf.c
ghost: > Also, here there should be a condition to choose between `sizeof(tmp)` and the remaining bytes to read. Otherwise I think you could end up with a longer string than what you requested.

It doesn't seems to break anything, the last test is testing the corner case.

test/unit/test_buf.c
```diff
@@ -1055,6 +1055,32 @@ bool test_rz_buf_get_string_nothing(void) {
...
+	memset(ch, 'A', 127);
+	ch[127] = '\0';
+	ch[111] = '\0';
+	RzBuffer *b = rz_buf_new_with_bytes(ch, 128);
+	char *s = rz_buf_get_nstring(b, 100, 10);
+	mu_assert_null(s, "there is no string with size < 10 (no null terminator)");
+	free(s);
+	s = rz_buf_get_nstring(b, 100, 11);
+	mu_assert_streq(s, (char *)ch + 100, "the string is the same");
+	free(s);
```
ret2libc:
```suggestion
	mu_assert_streq_free(s, (char *)ch + 100, "the string is the same");
```
You can do this if you want

test/unit/test_buf.c
ret2libc: I think you should actually test that `strlen(s) < x`, where `x` is the size you passed to `rz_buf_get_nstring`

librz/util/buf.c
ret2libc: RZ_ARRAY_SIZE gets the number of elements in an array... with chars it is the same as sizeof, but what you really want in this case is imho the all number of bytes. It seems conceptually better (to me at least).

test/unit/test_buf.c
ghost: Update

* test with string not found
* test with string found (strlen(s) == x)
* test with the full buffer
* test with string found (strlen(s) < x and strlen(s) == 32)
* test with string found (strlen(s) == x and strlen(s) == 32)

ghost:
rz_buf_get_nstring unit test was updated

test/unit/test_buf.c
ret2libc: mmm you didn't check that strlen(s) < x.

You could do:
```c
mu_assert_true(strlen(s) < x, "xxxx");
```

ghost: mu_assert_streq_free should assert this right?

ret2libc: In theory it checks that it's the same string pointed by the `expected` argument... I'm not 100% convinced this is the same as making sure that the returned string is of the maximum expected length.

ghost: I add `mu_assert_true`

ret2libc (approved):

test/unit/test_buf.c
ret2libc: I guess it is alright! Thanks!

PR #1270, opened 2021-07-05
Refactor notes storage elf file ##refactor

ghost:

**Detailed description**

* Add RzBinElfNotes
* Stored all note in RzVector

waiting https://github.com/rizinorg/rizin/pull/1269

**Test plan**

```
> rz-test
```

**Closing issues**

...

### DIFF
--- a/librz/bin/format/elf/elf.h
+++ b/librz/bin/format/elf/elf.h
@@ -35,6 +35,9 @@
 #define rz_bin_elf_foreach_relocs(bin, reloc) \
 	if (Elf_(rz_bin_elf_has_relocs)(bin)) \
 	rz_vector_foreach(bin->relocs, reloc)
+#define rz_bin_elf_foreach_notes_segment(bin, notes) \
+	if (Elf_(rz_bin_elf_has_notes)(bin)) \
+	rz_vector_foreach(bin->notes, notes)
 
 /// Information about the binary layout in a NT_PRSTATUS note for core files of a certain architecture and os
 typedef struct prstatus_layout_t {
@@ -128,32 +131,21 @@ typedef struct Elf_(rz_bin_elf_note_file_t) {
 RzBinElfNoteFile;
 
 /// Parsed PT_NOTE of type NT_PRSTATUS
-typedef struct Elf_(rz_bin_elf_note_prstatus_t) {
+typedef struct rz_bin_elf_note_prstatus_t {
 	size_t regstate_size;
 	ut8 *regstate;
 	// Hint: there is more info in NT_PRSTATUS notes that could be parsed if needed.
-}
-RzBinElfNotePrStatus;
+} RzBinElfNotePrStatus;
 
 /// A single PT_NOTE entry, parsed from an ElfW(Nhdr) and associated data.
 typedef struct Elf_(rz_bin_elf_note_t) {
 	Elf_(Word) type;
 	union {
-		struct {
-			size_t files_count;
-			RzBinElfNoteFile *files;
-		} file; //< for type == NT_FILE
+		RzBinElfNoteFile file; //< for type == NT_FILE
 		RzBinElfNotePrStatus prstatus; //< for type = NT_PRSTATUS
 	};
 }
 RzBinElfNote;
-
-/// A single parsed PT_NOTE segment
-typedef struct Elf_(rz_bin_elf_note_segment_t) {
-	size_t notes_count;
-	RzBinElfNote *notes;
-}
-RzBinElfNoteSegment;
 
 typedef struct rz_bin_elf_strtab RzBinElfStrtab;
 
@@ -186,7 +178,8 @@ struct Elf_(rz_bin_elf_obj_t) {
 	bool relocs_patched;
 	ut64 reloc_targets_map_base;
 
-	RzList /*<RzBinElfNoteSegment>*/ *note_segments;
+	// This is RzVector of note segment reprensented as RzVector<RzBinElfNote>
+	RzVector *notes; // RzVector<RzVector<RzBinElfNote>>
 
 	RzBinImport **imports_by_ord;
 	size_t imports_by_ord_size;
@@ -209,8 +202,6 @@ ut64 Elf_(rz_bin_elf_v2p_new)(RZ_NONNULL ELFOBJ *bin, ut64 vaddr);
 
 // elf_corefile.c
 
-RZ_BORROW RzBinElfPrStatusLayout *Elf_(rz_bin_elf_get_prstatus_layout)(RZ_NONNULL ELFOBJ *bin);
-RZ_BORROW const ut8 *Elf_(rz_bin_elf_grab_regstate)(RZ_NONNULL ELFOBJ *bin, RZ_NONNULL size_t *size);
 ut64 Elf_(rz_bin_elf_get_sp_val)(RZ_NONNULL ELFOBJ *bin);
 
 // elf_dynamic.c
@@ -253,7 +244,9 @@ ut64 Elf_(rz_bin_elf_get_main_offset)(RZ_NONNULL ELFOBJ *bin);
 
 // elf_notes.c
 
-bool Elf_(rz_bin_elf_init_notes)(RZ_NONNULL ELFOBJ *bin);
+RZ_BORROW RzBinElfPrStatusLayout *Elf_(rz_bin_elf_get_prstatus_layout)(RZ_NONNULL ELFOBJ *bin);
+RZ_OWN RzVector *Elf_(rz_bin_elf_notes_new)(RZ_NONNULL ELFOBJ *bin);
+bool Elf_(rz_bin_elf_has_notes)(RZ_NONNULL ELFOBJ *bin);
 
 // elf_misc.c
 
--- a/librz/bin/format/elf/elf.c
+++ b/librz/bin/format/elf/elf.c
@@ -324,14 +324,12 @@ static bool rz_bin_elf_init(ELFOBJ *bin) {
 	if (!rz_bin_elf_init_ehdr(bin)) {
 		return false;
 	}
+
 	if (!rz_bin_elf_init_phdr(bin) && !Elf_(rz_bin_elf_is_relocatable)(bin)) {
 		RZ_LOG_WARN("Cannot initialize program headers\n");
 	}
-	if (bin->ehdr.e_type == ET_CORE) {
-		if (!Elf_(rz_bin_elf_init_notes)(bin)) {
-			RZ_LOG_WARN("Cannot parse PT_NOTE segments\n");
-		}
-	} else {
+
+	if (bin->ehdr.e_type != ET_CORE) {
 		RzVector *sections = Elf_(rz_bin_elf_sections_new)(bin);
 
 		if (!rz_bin_elf_init_shstrtab(bin, sections)) {
@@ -354,6 +352,7 @@ static bool rz_bin_elf_init(ELFOBJ *bin) {
 	}
 
 	bin->relocs = Elf_(rz_bin_elf_relocs_new)(bin);
+	bin->notes = Elf_(rz_bin_elf_notes_new)(bin);
 
 	bin->boffset = Elf_(rz_bin_elf_get_boffset)(bin);
 
@@ -408,7 +407,7 @@ void Elf_(rz_bin_elf_free)(RZ_NONNULL ELFOBJ *bin) {
 
 	rz_vector_free(bin->relocs);
 
-	rz_list_free(bin->note_segments);
+	rz_vector_free(bin->notes);
 
 	if (bin->imports_by_ord) {
 		for (size_t i = 0; i < bin->imports_by_ord_size; i++) {
--- a/librz/bin/format/elf/elf_corefile.c
+++ b/librz/bin/format/elf/elf_corefile.c
@@ -6,85 +6,6 @@
 
 #include "elf.h"
 
-#define X86      0
-#define X86_64   1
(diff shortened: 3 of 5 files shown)


librz/bin/format/elf/elf.h
```diff
@@ -41,28 +41,28 @@ typedef struct prstatus_layout_t {
...
 	/**
-	 * This delta is the offset into the actual data of an NT_PRSTATUS note
-	 * where the regstate of size regsize lies.
-	 * That is, it is the offset after the Elf_(Nhdr) and the variable-length string + optional padding
-	 * have already been skipped.
-	 *
-	 * see size_t ELFLinuxPrStatus::GetSize(const lldb_private::ArchSpec &arch) in lldb source or similar
-	 * to determine values for this.
-	 */
+        * This delta is the offset into the actual data of an NT_PRSTATUS note
```
thestr4ng3r: why these spacing changes?

librz/bin/format/elf/elf.h
```diff
@@ -127,33 +126,12 @@ typedef struct Elf_(rz_bin_elf_note_file_t) {
 }
 RzBinElfNoteFile;
 
-/// Parsed PT_NOTE of type NT_PRSTATUS
```
thestr4ng3r, line 130: Please keep all these comments in the new data structures too.

librz/bin/format/elf/elf.h
```diff
@@ -127,33 +126,12 @@ typedef struct Elf_(rz_bin_elf_note_file_t) {
...
-			size_t files_count;
-			RzBinElfNoteFile *files;
-		} file; //< for type == NT_FILE
-		RzBinElfNotePrStatus prstatus; //< for type = NT_PRSTATUS
-	};
-}
-RzBinElfNote;
+typedef struct rz_bin_elf_note_prstatus_t {
+	size_t size;
+	ut8 *state;
```
thestr4ng3r: prstatus can have more than just the regstate, so this should stay `regstate_size`+`regstate` or similar.

thestr4ng3r (changes requested):
Summing up the change:
old:
```c
/// A single PT_NOTE entry, parsed from an ElfW(Nhdr) and associated data.
typedef struct Elf_(rz_bin_elf_note_t) {
	Elf_(Word) type;
	union {
		struct { ... } file; //< for type == NT_FILE
		RzBinElfNotePrStatus prstatus; //< for type = NT_PRSTATUS
	};
} RzBinElfNote;

/// A single parsed PT_NOTE segment
typedef struct Elf_(rz_bin_elf_note_segment_t) {
	size_t notes_count;
	RzBinElfNote *notes;
}
RzBinElfNoteSegment;

struct Elf_(rz_bin_elf_obj_t) {
	RzList /*<RzBinElfNoteSegment>*/ *note_segments;
};
```

new:
```c
typedef struct rz_bin_elf_notes_t {
	RzVector /*<RzBinElfNoteFile>*/ *files;
	RzVector /*<RzBinElfNotePrStatus>*/ *prstatus;
} RzBinElfNotes;

struct Elf_(rz_bin_elf_obj_t) {
	RzBinElfNotes *notes;
};
```

This change however is problematic since it loses both the association of which note  comes from which note section and the (interleaved) order of notes inside their section. Right now Rizin's core file handling is very crude and it only handles a single thread and even from that only a fraction of the available info.
But once multiple threads should be handled, the new layout would be a problem.
See this paragraph on why the order+section information is important: https://www.gabriel.urdhr.fr/2015/05/29/core-file/#thread-information

I would suggest to leave the overall data structure layout as it is, and only carry over the rather cosmetic details of this changeset like introducing a self-contained `RzBinElfNotes` class rather than storing the bare `note_segments` list in the elfobject directly.

As a sidenote, if you are interested in this topic, improving the core file handling regarding supporting multiple threads, more available info, more architectures and OSes (e.g. OpenBSD) is also still an overall open issue.

ghost:
> Summing up the change:
> old:
> 
> ```c
> /// A single PT_NOTE entry, parsed from an ElfW(Nhdr) and associated data.
> typedef struct Elf_(rz_bin_elf_note_t) {
> 	Elf_(Word) type;
> 	union {
> 		struct { ... } file; //< for type == NT_FILE
> 		RzBinElfNotePrStatus prstatus; //< for type = NT_PRSTATUS
> 	};
> } RzBinElfNote;
> 
> /// A single parsed PT_NOTE segment
> typedef struct Elf_(rz_bin_elf_note_segment_t) {
> 	size_t notes_count;
> 	RzBinElfNote *notes;
> }
> RzBinElfNoteSegment;
> 
> struct Elf_(rz_bin_elf_obj_t) {
> 	RzList /*<RzBinElfNoteSegment>*/ *note_segments;
> };
> ```
> 
> new:
> 
> ```c
> typedef struct rz_bin_elf_notes_t {
> 	RzVector /*<RzBinElfNoteFile>*/ *files;
> 	RzVector /*<RzBinElfNotePrStatus>*/ *prstatus;
> } RzBinElfNotes;
> 
> struct Elf_(rz_bin_elf_obj_t) {
> 	RzBinElfNotes *notes;
> };
> ```
> 
> This change however is problematic since it loses both the association of which note comes from which note section and the (interleaved) order of notes inside their section. Right now Rizin's core file handling is very crude and it only handles a single thread and even from that only a fraction of the available info.
> But once multiple threads should be handled, the new layout would be a problem.
> See this paragraph on why the order+section information is important: [gabriel.urdhr.fr/2015/05/29/core-file/#thread-information](https://www.gabriel.urdhr.fr/2015/05/29/core-file/#thread-information)
> 
> I would suggest to leave the overall data structure layout as it is, and only carry over the rather cosmetic details of this changeset like introducing a self-contained `RzBinElfNotes` class rather than storing the bare `note_segments` list in the elfobject directly.
> 
> As a sidenote, if you are interested in this topic, improving the core file handling regarding supporting multiple threads, more available info, more architectures and OSes (e.g. OpenBSD) is also still an overall open issue.

Yes the new data structure can't be used if Rizin was correctly parsing core file. But because core file isn't a short term objectives. I believe, we should keep the most basic implementation. Creating complex data structure just for hypothetical future is a case of premature optimization.

thestr4ng3r:
> Yes the new data structure can't be used if Rizin was correctly parsing core file. But because core file isn't a short term objectives. I believe, we should keep the most basic implementation. Creating complex data structure just for hypothetical future is a case of premature optimization.

Actually extending core files is sort of a mid term objective for me so having this info is required, in particular since it is already there. It's also not a premature optimization because it is not an optimization :-)

Do you have any particular use-cases right now where you would benefit from the refactoring here in terms of ease of use?

ghost:
> Actually extending core files is sort of a mid term objective for me so having this info is required, in particular since it is already there. It's also not a premature optimization because it is not an optimization :-)

The main advantages of the new implementation is the clear separation between both type of NT_NOTE (in the furture it could be the separation between CORE/NT_FILE, (CORE/NT_PRSTATUS, CORE/NT_FPREGSET,CORE/NT_X86_STATE,CORE/SIGINFO), ect...).

The side effect is a simplification when searching or iterating on both type (see librz/bin/p/bin_elf.inc).

PS: CORE/NT_PRSTATUS, CORE/NT_FPR... should be in the same struct.

librz/bin/format/elf/elf.h
```diff
@@ -253,7 +231,12 @@ ut64 Elf_(rz_bin_elf_get_main_offset)(RZ_NONNULL ELFOBJ *bin);
 
 // elf_notes.c
 
-bool Elf_(rz_bin_elf_init_notes)(RZ_NONNULL ELFOBJ *bin);
+RZ_BORROW RzBinElfPrStatusLayout *Elf_(rz_bin_elf_get_prstatus_layout)(RZ_NONNULL ELFOBJ *bin);
+RZ_BORROW RzVector *Elf_(rz_bin_elf_get_notes_file)(RZ_NONNULL ELFOBJ *bin);
+RZ_BORROW RzVector *Elf_(rz_bin_elf_get_notes_prstatus)(RZ_NONNULL ELFOBJ *bin);
```
ret2libc: do we really need a specific API just for prstatus? Maybe we can have a more generic API with `type` passed as argument, since there may be many other types of notes, right?

ret2libc:
I tend to agree with @thestr4ng3r that losing info which is already available might not be best. Though you guys seems to be more knowledgeable here.

thestr4ng3r:
> The main advantages of the new implementation is the clear separation between both type of NT_NOTE (in the furture it could be the separation between CORE/NT_FILE, (CORE/NT_PRSTATUS, CORE/NT_FPREGSET,CORE/NT_X86_STATE,CORE/SIGINFO), ect...).

I would argue that this has already been the case as `RzBinElfNote.type` (could be an enum too) clearly defines the contents of the union.

> The side effect is a simplification when searching or iterating on both type (see librz/bin/p/bin_elf.inc).

This is rather something I would design as an API on top, rather than modifying the data structures internally.

> PS: CORE/NT_PRSTATUS, CORE/NT_FPR... should be in the same struct.

This is an interesting point. It would essentially mean that this struct would not be `RzBinElfNotePrStatus` anymore but rather something like `RzBinElfThreadInfo`, containing all the notes describing one thread. So the data that is stored by us is not actually the notes but an abstraction over the information contained in the notes and then the parsing step would figure out the assignment depending on segmenting/order directly.

However even in that case, architecturally I would still only build this as an API/layer around the existing data structures keeping the segment/oder info instead of delegating too much logic to the initial parsing step building up these structures. That is why I am strongly for keeping the current data structures in place and rather trying to improve the API around them which can then easily do the assignment logic.

librz/bin/format/elf/elf.h
XVilka: Agree on that suggestion, especially since future GCC/Binutils might add more with time.

ghost: I modified the API

ghost:
> > The main advantages of the new implementation is the clear separation between both type of NT_NOTE (in the furture it could be the separation between CORE/NT_FILE, (CORE/NT_PRSTATUS, CORE/NT_FPREGSET,CORE/NT_X86_STATE,CORE/SIGINFO), ect...).
> 
> I would argue that this has already been the case as `RzBinElfNote.type` (could be an enum too) clearly defines the contents of the union.
> 
> > The side effect is a simplification when searching or iterating on both type (see librz/bin/p/bin_elf.inc).
> 
> This is rather something I would design as an API on top, rather than modifying the data structures internally.
> 
> > PS: CORE/NT_PRSTATUS, CORE/NT_FPR... should be in the same struct.
> 
> This is an interesting point. It would essentially mean that this struct would not be `RzBinElfNotePrStatus` anymore but rather something like `RzBinElfThreadInfo`, containing all the notes describing one thread. So the data that is stored by us is not actually the notes but an abstraction over the information contained in the notes and then the parsing step would figure out the assignment depending on segmenting/order directly.
> 
> However even in that case, architecturally I would still only build this as an API/layer around the existing data structures keeping the segment/oder info instead of delegating too much logic to the initial parsing step building up these structures. That is why I am strongly for keeping the current data structures in place and rather trying to improve the API around them which can then easily do the assignment logic.

I revert some changes, what do you think?

librz/bin/format/elf/elf.h
```diff
@@ -128,33 +131,21 @@ typedef struct Elf_(rz_bin_elf_note_file_t) {
...
 
 /// Parsed PT_NOTE of type NT_PRSTATUS
-typedef struct Elf_(rz_bin_elf_note_prstatus_t) {
+typedef struct rz_bin_elf_note_prstatus_t {
 	size_t regstate_size;
 	ut8 *regstate;
-	// Hint: there is more info in NT_PRSTATUS notes that could be parsed if needed.
-}
-RzBinElfNotePrStatus;
+} RzBinElfNotePrStatus;
```
thestr4ng3r, line 138:
```suggestion
	// Hint: there is more info in NT_PRSTATUS notes that could be parsed if needed.
} RzBinElfNotePrStatus;
```

librz/bin/format/elf/elf.h
```diff
@@ -128,33 +131,21 @@ typedef struct Elf_(rz_bin_elf_note_file_t) {
...
 
 /// A single PT_NOTE entry, parsed from an ElfW(Nhdr) and associated data.
 typedef struct Elf_(rz_bin_elf_note_t) {
 	Elf_(Word) type;
 	union {
-		struct {
-			size_t files_count;
-			RzBinElfNoteFile *files;
-		} file; //< for type == NT_FILE
+		RzBinElfNoteFile file;
```
thestr4ng3r:
```suggestion
		RzBinElfNoteFile file;  //< for type == NT_FILE
```

librz/bin/format/elf/elf.h
```diff
@@ -186,7 +177,7 @@ struct Elf_(rz_bin_elf_obj_t) {
 	bool relocs_patched;
 	ut64 reloc_targets_map_base;
 
-	RzList /*<RzBinElfNoteSegment>*/ *note_segments;
+	RzVector *notes; // RzVector<RzVector<RzBinElfNote>>
```
thestr4ng3r, line 182: Please add a documentation comment here clarifying that the elements of the outer `RzVector` each describe an individual segment.

thestr4ng3r (changes requested):
Apart from adding these docs, it lgtm now.

librz/bin/format/elf/elf.h
ghost: I add a comment

thestr4ng3r (approved):

Issue #1093, opened 2021-04-30
Integration Test test_open_analyse_save_load_project only works a single time


thestr4ng3r:
After running it once, the `.tmp` folder already exists and it fails:
![Bildschirmfoto 2021-04-30 um 09 21 48](https://user-images.githubusercontent.com/1460997/116662056-9631cf00-a995-11eb-8286-200f8412d8a6.png)

XVilka:
Fixed in https://github.com/rizinorg/rizin/commit/28faeff9ef261ff68922daea1e8b8b2ac33c0293

PR #1271, opened 2021-07-06
Add integration test for CPU and Platform profiles

officialcjunior:

**Detailed description**
* Add an integration test for CPU and Platform profiles.

* Return `NULL` if `target->platforms` is `NULL` in
`rz_arch_platform_target_new`

* Move the call of the API which adds the flags/comments for both
CPU and platform profiles to `rz_core_analysis_all()`.

**Test plan**

- [x] CI is green

**Closing issues**

None

### DIFF
--- a/test/integration/meson.build
+++ b/test/integration/meson.build
@@ -1,7 +1,8 @@
 if get_option('enable_tests')
   tests = [
     'open_analyse_save_load_project',
-    'bin_vfiles'
+    'bin_vfiles',
+    'cpu_platform_profiles'
   ]
 
   unit_test_env = environment()
--- a/test/integration/meson.build
+++ b/test/integration/meson.build
@@ -2,7 +2,7 @@ if get_option('enable_tests')
   tests = [
     'open_analyse_save_load_project',
     'bin_vfiles',
-    'cpu_platform_profiles'
+    'cpu_platform_profiles',
   ]
 
   unit_test_env = environment()
--- a/test/integration/test_cpu_platform_profiles.c
+++ b/test/integration/test_cpu_platform_profiles.c
@@ -5,7 +5,6 @@
 #include "../unit/minunit.h"
 
 bool test_cpu_profiles() {
-
 	// 1. Open the file
 	RzCore *core = rz_core_new();
 	mu_assert_notnull(core, "new RzCore instance");
@@ -61,7 +60,6 @@ bool test_cpu_profiles() {
 }
 
 bool test_platform_profiles() {
-
 	// 1. Open the file
 	RzCore *core = rz_core_new();
 	mu_assert_notnull(core, "new RzCore instance");
--- a/test/integration/test_cpu_platform_profiles.c
+++ b/test/integration/test_cpu_platform_profiles.c
@@ -29,7 +29,7 @@ bool test_cpu_profiles() {
 	mu_assert_eq(item->offset, 0x00000005, "Flag PORTB not found");
 
 	// 3. Save into the project
-	if (!rz_file_exists(".tmp/")) {
+	if (!rz_file_is_directory(".tmp/")) {
 		mu_assert_true(rz_sys_mkdir(".tmp/"), "create tmp directory");
 	}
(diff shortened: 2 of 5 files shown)


officialcjunior:
Right. The test I just introduced runs fine, and the other fails because the directory `.tmp/` where it stores the `.rzdb` is already there, I believe. 
Should I remove that check (at line 40) or get another directory?

XVilka:
But it should be fine even if it exists, I think. See also related https://github.com/rizinorg/rizin/issues/1093

test/integration/test_cpu_platform_profiles.c
```diff
@@ -0,0 +1,124 @@
+#include <rz_core.h>
+#include <rz_arch.h>
+#include <rz_project.h>
+
+#include "../unit/minunit.h"
+
+bool test_cpu_profiles() {
+
```
ret2libc:
```suggestion
```

test/integration/test_cpu_platform_profiles.c
```diff
@@ -0,0 +1,124 @@
...
+	mu_assert_eq(item->offset, 0x00000004, "Flag DDRB not found");
+	item = rz_flag_get(core->flags, "PORTB");
+	mu_assert_eq(item->offset, 0x00000005, "Flag PORTB not found");
+
+	rz_core_free(core);
+	mu_end;
+}
+
+bool test_platform_profiles() {
+
```
ret2libc:
```suggestion
```

test/integration/test_cpu_platform_profiles.c
```diff
@@ -0,0 +1,124 @@
+#include <rz_core.h>
```
ret2libc, line 1: @thestr4ng3r I probably missed the addition of this dir. How are we going to split things that need to go in test/integration and test/unit? I believe there are already some things in unit that are not that "unit". WDYT?

XVilka: Integration tests in this case are the ones that involve multiple parts/modules of Rizin, and usually represent some real use-case scenario.  They are good way to test possible use of the Rizin API.

test/integration/test_cpu_platform_profiles.c
officialcjunior: Thank you!

test/integration/test_open_analyse_save_load_project.c
```diff
@@ -37,7 +37,7 @@ bool test_open_analyse_save() {
 	size_t functions_count_expect = rz_list_length(functions);
 
 	// 4. Save into the project
-	if (!rz_file_exists(".tmp/")) {
+	if (!rz_file_is_directory(".tmp/")) {
```
wargio: is this folder even legal for windows? @ret2libc

XVilka: It is - test was passing on Windows

ret2libc: Yeah I think it is nowadays.. anyway, fixed on another PR.

wargio: You can remove this commit, since ret2libc has fixed this on another PR which is merged

officialcjunior: Rebased to `dev`.

XVilka (approved):

test/integration/meson.build
```diff
@@ -1,7 +1,8 @@
 if get_option('enable_tests')
   tests = [
     'open_analyse_save_load_project',
-    'bin_vfiles'
+    'bin_vfiles',
+    'cpu_platform_profiles'
```
ret2libc: Not important, but I think this should be valid in meson.build files. This way you don't have to modify the last line whenever you add a new entry to the array
```suggestion
    'cpu_platform_profiles',
```

ret2libc (approved):
Seems ok for what I understand.

On a different topic: I like the split in integration/unit tests! We could use the meson feature `suite: '` of `test` to split them and then run only unit by default.

Issue #277, opened 2020-12-27
Redesign of ESIL, or use of alternatives


ghost:
**Is your feature request related to a problem? Please describe.**
ESIL is imperfect (slow, no float support, no well tested).
In theory, if Rizin used a good IL, we could have a simplify analysis (as in ghidra).

"By modeling in this way, the analysis of different processors is put into a common framework, facilitating the development of retargetable analysis algorithms and applications" - ghidra

**Describe the solution you'd like**
Using alternative IL:
* p-code
* bil
* openreil
* ...

**Describe alternatives you've considered**
Modify the ESIL specification...

**Additional context**
https://ghidra.re/courses/languages/html/pcoderef.html

XVilka:
While I agree with the general idea, P-code is very old, OpenREIL as well (OpenREIL project is recent but REIL itself is very old). BIL is more modern but still substituted by the [Knowledge Base](http://binaryanalysisplatform.github.io/knowledge-intro-1). [Falcon](https://github.com/falconre/falcon) opted for the modification of **RREIL** (not to mistake it as *REIL* - it's completely different) instead. Also worth checking [Radeco middle layer](https://github.com/radareorg/radeco/tree/master/radeco-lib/src/middle) as well.
If we decide to switch to another IL/IR, it's better to keep in mind the current research.
See also my suggestion for Ghidra itself about describing the architecture/lifting rules using Datalog: https://github.com/NationalSecurityAgency/ghidra/issues/948

I am strongly in favor of something resembling Core Theory (something like an extension of SMT) and Knowledge Base, see more at BAP documentation:
- http://binaryanalysisplatform.github.io/bap/api/odoc/bap-knowledge/Bap_knowledge/Knowledge/index.html
- http://binaryanalysisplatform.github.io/bap/api/odoc/bap-core-theory/Bap_core_theory/index.html

XVilka:
This is what I had in mind if we modify the ESIL itself:
- [ ] Separate `RZ_API rz_analysis_esil_*` into `librz/include/rz_esil.h`
- [ ] Consider to isolate the ESIL handling into the separate library, like SIOL was
- [ ] Provide the better API to distinct between various ESIL operation kinds
- [ ] Provide the API for "un-RPN"-ing of the ESIL expressions, for printing purpose, maybe more readable representation, useful for debugging/etc
- [ ] Provide the way for easier distinction between memory accesses and register accesses in API
[...]

PR #1304, opened 2021-07-14
Remove ESIL CFG and DFG

XVilka:

**Detailed description**

In the light of the upcoming [new IL](https://github.com/rizinorg/rizin/issues/277) I propose the removal of the unused and barely tested ESIL CFG and DFG implementations.
It was used in:
- `aof` command, which had just one most basic test and not even in the help output
- `aeg` command which wasn't tested

It will also simplify the migration since less code would be required to update/modify.

Note, that having CFG and DFG is a cool feature and we should implement it on top of the new IL, cover with tests and reuse in the analysis loop. But I think it's easier to do from scratch rather than to modify the ESIL-tailored implementation.

**Test plan**

CI is green


thestr4ng3r (approved):
These indeed are very nice features but in light of the upcoming new il, it's highly unlikely they will be used anymore.

kazarmy (approved):

wargio (approved):

PR #1310, opened 2021-07-15
Update tef test

Basstorm:

**Detailed description**

The calculation of the hash index is `hash % ht.size`, so the index will change when the size changes. This may cause the order of output could be different on different OS/Arch because the types stored in the `SDB` are different.

**Test plan**

...

**Closing issues**

...


XVilka:
Could you please make a test instead that uses grep? For example,
- `tef 4~?` to count the number of output
- `tef4~FOO.BAR` to check the exact member

XVilka (approved):

Issue #158, opened 2020-12-10
Add Windows and Jemalloc heap parsing tests


XVilka:
Currently only glibc heap is tested with
- [rizin-testbins/heap/linux_glibc-2.30_x64.bin](https://github.com/rizinorg/rizin-testbins/blob/master/heap/linux_glibc-2.30_x64.bin)
- [test/db/archos/linux-x64/dbg_dmh](https://github.com/rizinorg/rizin/blob/dev/test/db/archos/linux-x64/dbg_dmh)

```
NAME=dmh/dmha with memory dump
FILE=bins/heap/linux_glibc-2.30_x64.bin
ARGS=-n
CMDS=<<EOF
#re-map arena and [heap]
om 3 0x7ffff7f8a000 0x898 0x0 rw- arena
om 3 0x555555559000 0x3200 0x898 rw- [heap]

dmha~?0x7ffff7f8a000
e dbg.glibc.tcache=0
dmh~?allocated
e dbg.glibc.tcache=1
dmh~?allocated
EOF
EXPECT=<<EOF
1
3
2
EOF
RUN
```

Note, that this static test probably should be moved under a separate category `heap`, since not really dependent from the operating system.

We need add this static tests for Windows heap parser and jemalloc too.

cc @GustavoLCR @meowmeowxw

meowmeowxw:
I'm the author of https://github.com/rizinorg/rizin/blob/dev/test/db/archos/linux-x64/dbg_dmht, every binary inside bins/elf/ has a patchef interpreter/linker. Example: 

```
$ patchelf --print-interpreter bins/elf/glibc-heap-2.27
bins/elf/ld-2.27.so
```

Now I'm particularly busy to continue working on this

XVilka:
@PulakIIIT regarding this. Currently heap parsing is running only on Linux. But the heap parsing is untied to the debugger. What about making existing heap parsing tests static? So they will be runned on all supported platforms - e.g. Linux glibc binary heap would be parsed even on Windows, MacOS, FreeBSD.

MalhotraPulak:
Can you elaborate more or provide some relevant links? I am a bit confused, I do not understand how Glibc binary can be run on a platform like MacOS.

XVilka:
@PulakIIIT idea is not to run these binaries. Idea is to:
1. Run the binary on Linux with desired configuration
2. Save the required memory maps in files
3. Add them in rizin-testbins
4. Open them as `om` 
[...]

PR #1328, opened 2021-07-19
Move windows heap to new shell

MalhotraPulak:

**Detailed description**
Goal of this PR is to add tests for windows heap and move it to new shell. Once this is done then I plan to create an Rizin API so this functionality can be exposed to Cutter. 
If anyone has any ideas how to write good tests for this code please share. 

Done
- Moved windows heap commands to new shell
- `dmw` ,`dmwb` `dmwbf` working

TODO
- Need to add atleast one test per command

...

**Test plan**

Need to write tests

...

**Closing issues**

Partially solves #158

...


MalhotraPulak:
I looked into writing tests for windows commands there is a problem that the command outputs are non-deterministic, both the addresses and the number of heap blocks. I can think of two ways to solve this but I am not sure how to implement either:
- Make tests like the ones for Linux heap with memory dump. I don't know how to create these memory dump files, but if we can do it on windows this might produce consistent output.
```
NAME=dmh/dmha with memory dump
FILE=bins/heap/linux_glibc-2.30_x64.bin
ARGS=-n
CMDS=<<EOF
#re-map arena and [heap]
om 3 0x7ffff7f8a000 0x898 0x0 rw- arena
om 3 0x555555559000 0x3200 0x898 rw- [heap]

dmha~?0x7ffff7f8a000
e dbg.glibc.tcache=0
dmh~?allocated
e dbg.glibc.tcache=1
dmh~?allocated
EOF
EXPECT=<<EOF
1
3
2
EOF
RUN
```
- The number of blocks (from `dmwb`) on my machine is non determinisitic but it is in range of 55 to 70 for every time i run a  specific binary. If this range remains constant across machines then we  can have a range based test specifying that any answer between 55-70 is okay.

XVilka:
@PulakIIIT yes, the idea is to use memory dump is the best way, it will help to test the parsing code on all systems as well. Maybe @GustavoLCR could help us with generating such dumps. It is also was suggested to use dumps here: https://github.com/rizinorg/rizin/issues/158

librz/core/windows_heap.c
```diff
@@ -1337,94 +1328,83 @@ static void w32_list_heaps_blocks(RzCore *core, const char format) {
...
+//};
+//
+//static const char *help_msg_block[] = {
+//	"Usage:", " dmhb[f|j]", " # Windows heap parsing commands",
+//	"dmwb [addr]", "", "List allocated heap blocks",
+//	"dmwbf", "", "Create flags for each allocated block",
+//	"dmwbj [addr]", "", "Print output in JSON format",
+//	NULL
+//};
+
```
wargio: delete this.

librz/core/windows_heap.c
```diff
@@ -1337,94 +1328,83 @@ static void w32_list_heaps_blocks(RzCore *core, const char format) {
...
+//	switch (input[0]) {
+//	case 'b': // dmhb
+//		cmd_debug_map_heap_block_win(core, input + 1);
+//		break;
+//	default:
+//		w32_list_heaps(core, input[0]);
+//		break;
+//	}
+//	return true;
+//}
```
wargio: delete this

librz/core/windows_heap.c
```diff
@@ -1337,94 +1328,83 @@ static void w32_list_heaps_blocks(RzCore *core, const char format) {
...
+//static const char *help_msg_block[] = {
+//	"Usage:", " dmhb[f|j]", " # Windows heap parsing commands",
+//	"dmwb [addr]", "", "List allocated heap blocks",
+//	"dmwbf", "", "Create flags for each allocated block",
+//	"dmwbj [addr]", "", "Print output in JSON format",
+//	NULL
+//};
+
+static void cmd_debug_map_heap_block_win(RzCore *core, const char *addr, RzOutputMode mode, bool flag) {
+	init_func();
```
wargio: What is this?

librz/core/windows_heap.c
```diff
@@ -1208,7 +1208,8 @@ static RzTable *__new_heapblock_tbl(void) {
 	return tbl;
 }
 
-static void w32_list_heaps(RzCore *core, const char format) {
+static void w32_list_heaps(RzCore *core, RzOutputMode mode) {
+	init_func();
```
wargio: what is this?

wargio (changes requested):

librz/core/windows_heap.c
MalhotraPulak: `init func` call was present in `cmd_debug_map_heap_win` function (line 1400 in old file)

XVilka:
I skimmed through code quickly, it heavily uses `ReadProcessMemory` which calls Windows API. In theory it could be changed to some middleware function that can either use `ReadProcessMemory` when ran in the debug mode, or read from buffer/file when load the memory dump instead.

librz/core/windows_heap.c
XVilka: It's a function that gets the pointers to the API. I suggest to rename it into something like `initialize_windows_ntdll_query_api_functions()`.

wargio: agree. that needs to be renamed. i thought this was something malicious.

librz/core/windows_heap.c
```diff
@@ -749,10 +749,10 @@ static bool GetSegmentHeapBlocks(RzDebug *dbg, HANDLE h_proc, PVOID heapBase, PH
...
 static PDEBUG_BUFFER GetHeapBlocks(DWORD pid, RzDebug *dbg) {
 	/*
-		TODO:
-			Break this behemoth
-			x86 vs x64 vs WOW64	(use dbg->bits or new structs or just a big union with both versions)
-	*/
+                TODO:
+                        Break this behemoth
+                        x86 vs x64 vs WOW64	(use dbg->bits or new structs or just a big union with both versions)
+        */
```
yossizap: Fix indent

librz/core/windows_heap.c
```diff
@@ -980,9 +980,9 @@ static PDEBUG_BUFFER GetHeapBlocks(DWORD pid, RzDebug *dbg) {
 
 static PHeapBlock GetSingleSegmentBlock(RzDebug *dbg, HANDLE h_proc, PSEGMENT_HEAP heapBase, WPARAM offset) {
 	/*
-	*	TODO:
-	*		- Backend (Is this needed?)
-	*/
+        *	TODO:
+        *		- Backend (Is this needed?)
+        */
```
yossizap: Fix indent

yossizap (changes requested):

XVilka (approved):

XVilka:
@GustavoLCR please take a look at this.

GustavoLCR (approved):

ret2libc:
> I looked into writing tests for windows commands there is a problem that the command outputs are non-deterministic, both the addresses and the number of heap blocks. I can think of two ways to solve this but I am not sure how to implement either:

There is also the chance to use `REGEXP_FILTER_OUT` in the tests. You have to be good with regexp though :) Grep around to see how its used. There is also some short explanation in https://github.com/rizinorg/rizin/blob/dev/test/README.md . Hope it helps.

librz/core/cmd_descs/cmd_debug.yaml
```diff
@@ -195,9 +195,28 @@ commands:
             optional: true
       - name: dmw
         summary: Windows heap commands
-        cname: cmd_debug_heap_windows
-        type: RZ_CMD_DESC_TYPE_OLDINPUT
-        args: []
+        subcommands:
```
ret2libc, line 256: please use `type: RZ_CMD_DESC_TYPE_ARGV_STATE` for all these subcommands... We should probably make it the default instead of `RZ_CMD_DESC_TYPE_ARGV` at some point...

librz/core/cmd_windows_heap.c
```diff
@@ -0,0 +1,35 @@
+#include <rz_core.h>
+
+RZ_IPI RzCmdStatus rz_cmd_debug_process_heaps_handler(RzCore *core, int argc, const char **argv, RzOutputMode mode) {
```
ret2libc: i'm tempted to recommend  having two parts in this file, one for windows and one for everything else.

```C
#if __WINDOWS__
RZ_IPI RzCmdStatus rz_cmd_debug_process_heaps_handler(RzCore *core, int argc, const char **argv, RzOutputMode mode) {
        ...
}
...
#else
RZ_IPI RzCmdStatus rz_cmd_debug_process_heaps_handler(RzCore *core, int argc, const char **argv, RzOutputMode mode) {
        eprintf("Windows heap parsing is not supported on this platform\n");
	return RZ_CMD_STATUS_ERROR;
}
...
#endif
```

yossizap (approved):

librz/core/cmd_windows_heap.c
wargio: it actually makes sense.

librz/core/cmd_descs/cmd_debug.yaml
ret2libc: Do you plan to address this?

ret2libc (approved):
LGTM. Just want to know though what's your plan with the comment about ARGV_STATE.

MalhotraPulak:
@ret2libc I think we can merge this for now, I will open another PR soon regarding windows heap API for Cutter, I will address the ARGV_STATE in that

ret2libc:
@wargio please re-review and auto-merge if you are ok

wargio (approved):

PR #1331, opened 2021-07-19
Add better warning messages librz/bin/format/elf/*

ghost:

**Detailed description**

* ~~I need help with one failing test (db/archos/linux-x64/dbg_oo ood check for ptrace errors)~~
* ~~I used RZ_LOG_WARN everywhere, not sure if it is a good idea.~~
* ~~There is 2 types of warning: memory and logic. Is there a specific function to report out of memory error (malloc, sdb, rz_str, ht_uu, rz_vector, etc).~~
* Warnings message should be review (grammar and logic)

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/librz/bin/format/elf/elf_dynamic.c
+++ b/librz/bin/format/elf/elf_dynamic.c
@@ -4,95 +4,9 @@
 
 #include "elf_dynamic.h"
 
-static bool get_dynamic_entry(ELFOBJ *bin, ut64 offset, Elf_(Dyn) * entry) {
-	if (!Elf_(rz_bin_elf_read_sword_sxword)(bin, &offset, &entry->d_tag)) {
-		return false;
-	}
-
-	if (!Elf_(rz_bin_elf_read_addr)(bin, &offset, &entry->d_un.d_ptr)) {
-		return false;
-	}
-
-	return true;
-}
-
-static void fill_dt_dynamic(ELFOBJ *bin, RzBinElfDtDynamic *ptr, ut64 offset, ut64 size) {
-	for (ut64 entry_offset = offset; entry_offset < offset + size; entry_offset += sizeof(Elf_(Dyn))) {
-		Elf_(Dyn) entry;
-
-		if (!get_dynamic_entry(bin, entry_offset, &entry)) {
-			break;
-		}
-
-		if (entry.d_tag == DT_NULL) {
-			break;
-		}
-
-		if (entry.d_tag == DT_NEEDED) {
-			rz_vector_push(ptr->dt_needed, &entry.d_un.d_val);
-		} else if (!ht_uu_insert(ptr->info, entry.d_tag, entry.d_un.d_val)) {
-			RZ_LOG_WARN("Dynamic tag %" PFMT64d " already handled\n", (ut64)entry.d_tag);
-		}
-	}
-}
-
-static bool init_dt_dynamic(ELFOBJ *bin, RzBinElfDtDynamic *ptr) {
-	RzBinElfSegment *segment = Elf_(rz_bin_elf_get_segment_with_type)(bin, PT_DYNAMIC);
-	if (!segment) {
-		return false;
-	}
-
-	ut64 size = segment->data.p_filesz;
-	ut64 offset = Elf_(rz_bin_elf_v2p_new)(bin, segment->data.p_vaddr);
-	if (offset == UT64_MAX || !size || offset + size > bin->size) {
-		return false;
-	}
-
-	fill_dt_dynamic(bin, ptr, offset, size);
-
-	return true;
-}
-
-RZ_BORROW RzVector *Elf_(rz_bin_elf_get_dt_needed)(RZ_NONNULL ELFOBJ *bin) {
-	rz_return_val_if_fail(bin && bin->dt_dynamic && bin->dt_dynamic->dt_needed, NULL);
-
-	return bin->dt_dynamic->dt_needed;
-}
-
-RZ_OWN RzBinElfDtDynamic *Elf_(rz_bin_elf_dt_dynamic_new)(RZ_NONNULL ELFOBJ *bin) {
-	RzBinElfDtDynamic *result = RZ_NEW(RzBinElfDtDynamic);
-	if (!result) {
-		return NULL;
-	}
-
-	result->info = ht_uu_new0();
-	if (!result->info) {
-		free(result);
-		return NULL;
-	}
-
-	result->dt_needed = rz_vector_new(sizeof(Elf_(Word)), NULL, NULL);
-	if (!result->dt_needed) {
-		ht_uu_free(result->info);
-		free(result);
-		return NULL;
-	}
-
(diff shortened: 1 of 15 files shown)


wargio:
RZ_LOG_WARN is ok

librz/bin/format/elf/elf_dynamic.c
```diff
@@ -16,80 +16,105 @@ static bool get_dynamic_entry(ELFOBJ *bin, ut64 offset, Elf_(Dyn) * entry) {
...
+		RZ_LOG_WARN("Dynamic section allocation failed.\n");
+		Elf_(rz_bin_elf_dt_dynamic_free)(result);
 		return NULL;
 	}
 
 	result->dt_needed = rz_vector_new(sizeof(Elf_(Word)), NULL, NULL);
 	if (!result->dt_needed) {
-		ht_uu_free(result->info);
-		free(result);
+		RZ_LOG_WARN("Dynamic section allocation failed.\n");
```
wargio: here use RZ_LOG_ERROR. if it is a fail, then is ERROR, like for example when you fail to alloc.
When is a problem, like you expect to have 255 max values but you have 1000, then is a warning.

wargio (changes requested):

librz/bin/format/elf/elf_dynamic.c
wargio: same for all failed allocs in this code.

ghost: Ty

librz/bin/format/elf/elf_sections.c
```diff
@@ -76,12 +76,12 @@ static bool create_section_from_phdr(ELFOBJ *bin, RzVector *result, const char *
 	section.size = sz;
 	section.name = strdup(name);
 	if (!section.name) {
-		RZ_LOG_WARN("Failed to allocate memory for the section name.")
+		RZ_LOG_WARN("Section name allocation railed.\n")
```
thestr4ng3r:
```suggestion
		RZ_LOG_WARN("Section name allocation failed.\n")
```

librz/bin/format/elf/elf_symbols.c
```diff
@@ -512,7 +515,19 @@ Elf_(Word) Elf_(rz_bin_elf_get_number_of_dynamic_symbols)(RZ_NONNULL ELFOBJ *bin
 		return result;
 	}
 
-	return get_number_of_symbols_from_heuristic(bin);
+	result = get_number_of_symbols_from_section(bin);
+	if (result) {
+		RZ_LOG_WARN("The number of dynamic symbols is deduced from section information.\n");
```
thestr4ng3r: This by itself doesn't say much when appearing to the user. Before the `result = get_number_of_symbols_from_section(bin);` line I would do something like
```c
RZ_LOG_WARN("Neither hash nor gnu_hash exist. Falling back to heuristics for deducing the number of dynamic symbols...\n");
```

And then in the end if all heuristics have failed:
```c
RZ_LOG_WARN("Failed to determine the number of dynamic symbols from heuristics.\n");
```
(These log strings are just examples, you will probably know better how to formulate them accurately here)

thestr4ng3r (commented):
> There is 2 types of warning: memory and logic. Is there a specific function to report out of memory error (malloc, sdb, rz_str, ht_uu, rz_vector, etc).

Almost all memory errors should simply be propagated outwards by returning `NULL` or whatever is appropriate in the specific case, without printing any warning, which is pretty much the standard way to do this in C. In particular, mallocs that allocate a static amount of memory like `rz_list_new()` shouldn't have any warnings at all.

A different case is when you allocate memory of a size given from the outside though, for example allocating an entire block of the a segment size that comes from the binary. Then it can make sense to print a warning to indicate there is something wrong with the binary or you just don't have enough ram to deal with that binary.

> Warnings message should be review (grammar and logic)

General comment for all `read failed` and `parsing failed`: Instead of e.g. `Note prstatus parsing failed.` I would write `Failed to parse NT_PRSTATUS note.`. This makes it a bit more comprehensible imo.

librz/bin/format/elf/elf_symbols.c
wargio: i agree on this one, but use `RZ_LOG_ERROR` for `RZ_LOG_WARN("Failed to determine the number of dynamic symbols from heuristics.\n");`.

ghost:
> > There is 2 types of warning: memory and logic. Is there a specific function to report out of memory error (malloc, sdb, rz_str, ht_uu, rz_vector, etc).
> 
> Almost all memory errors should simply be propagated outwards by returning `NULL` or whatever is appropriate in the specific case, without printing any warning, which is pretty much the standard way to do this in C. In particular, mallocs that allocate a static amount of memory like `rz_list_new()` shouldn't have any warnings at all.
> 
> A different case is when you allocate memory of a size given from the outside though, for example allocating an entire block of the a segment size that comes from the binary. Then it can make sense to print a warning to indicate there is something wrong with the binary or you just don't have enough ram to deal with that binary.
> 
> > Warnings message should be review (grammar and logic)
> 
> General comment for all `read failed` and `parsing failed`: Instead of e.g. `Note prstatus parsing failed.` I would write `Failed to parse NT_PRSTATUS note.`. This makes it a bit more comprehensible imo.

How far the propagation should go?

ghost:
~~I need help with db/archos/linux-x64/dbg_oo, I can't figure out how to fix the test because of `REGEXP_FILTER_ERR=([a-zA-Z-]+\s+)`.~~

librz/bin/format/elf/elf_symbols.c
ghost: Okay ty

XVilka:
```
[XX] db/formats/elf/main ELF: endbr-main-mov
RZ_NOPLUGINS=1 rizin -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -Qc '?v main' bins/elf/endbr-main
-- stderr
--- expected
+++ actual
@@ -1,3 +1,0 @@
-WARNING: Neither hash nor gnu_hash exist. Falling back to heuristics for deducing the number of dynamic symbols...
-WARNING: Neither hash nor gnu_hash exist. Falling back to heuristics for deducing the number of dynamic symbols...
-WARNING: Neither hash nor gnu_hash exist. Falling back to heuristics for deducing the number of dynamic symbols...
```

ghost:
@XVilka It is really weird.

* MacOs doesn't failed
* Freebsd doesn't failed
* Openbsd doesn't failed
* gcc-asan doesn't failed on a related test (some PE test failed)
* gcc and clang failed because no WARNING was printed.

In theory there shouldn't be any side effect because gcc-asan didn't failed. I am confused. I don't know where is the problem. I couldn't reproduce the failed test on my computer with and without asan.

XVilka (approved):
ASAN and Debian Wheezy failures are unrelated

wargio (approved):

Issue #1205, opened 2021-06-10
Add support for classes and methods in RzType


XVilka:
Currently RzType performs only C parsing.
We could extend it by adding C++ parser with tree-sitter-cpp.
Moreover, some of the types and the API should be extended or modified to accommodate classes properly.

- [ ] move the class data structures from RzAnalysis to the RzType module (and transition out of sdb while at it), that way all modules can access them. The class database could be made part of the RzTypeDB.
- [ ]  from parsing dwarf, pdf, etc. emit a list of types and a list of classes or similar
- [ ] for association between class+struct, I would add the association to the class data structure as for example the name of the struct. I would probably not add anything to RzBaseType structures because that is more low-level and should itself stay clean of class info.

We always have to keep in mind a few things though: C++ classes have a vtable member that is super important. Base classes are embedded structs and they also have a vtable pointer so if a class struct has a base, it contains the vtable pointer nested deeper. A class can have multiple bases at different offsets and then also multiple vtables.

You recognize for example a pointer to an object of a certain class by being a pointer to a structure that is referenced by a class in the database.

Actually I could imagine having some info like what member is a base class and what is the vtable ptr inside the RzBaseType (not all the class information itself though!). But I would rather build that around the low-level types instead of embedding it inside of them.

Basstorm:
There are some PDB tests marked as broken because we don't support classes yet.
https://github.com/rizinorg/rizin/pull/1517/commits/45ca9441e1740b4d17e0d545d73f881e8ae7be3d
See files:
- `test/db/formats/pdb` (search for `BROKEN=1` string)
- `librz/analysis/type_pdb.c` (`parse_regular_type()` function)

ret2libc:
I didn't see any work on this during this time, so I doubt it will be done for 0.4.0.
I'm removing the milestone for now until someone starts to actively looking at this.

cc @XVilka

XVilka:
After tree-sitter-cpp scanner was reimplemented in pure C, it can be used now for parsing C++ headers for class and related types information: https://github.com/tree-sitter/tree-sitter-cpp/commit/2e29f23abcb5f92ddffb22de8b292f09ed78db01

Issue #1240, opened 2021-06-24
Add proper bitfields support in RzType


XVilka:
Currently RzBaseType doesn't consider bitfields for structures, unions, enums. It should be recognized and processed properly.

See also https://github.com/rizinorg/rizin/issues/314

Depends on https://github.com/tree-sitter/tree-sitter-c/issues/42

A good description on bitfields layout on ARM is here: https://www.keil.com/support/man/docs/armcc/armcc_chr1360774870320.htm

XVilka:
There are some PDB changes required after added support for those

https://github.com/rizinorg/rizin/pull/1517/commits/45ca9441e1740b4d17e0d545d73f881e8ae7be3d
See files:
- `test/db/formats/pdb`
- `librz/analysis/type_pdb.c` (`parse_regular_type()` function)

Also add more tests in `test/unit/test_pdb.c`

PR #1350, opened 2021-07-23
Add class type for RzTypes

Basstorm:

**Detailed description**

When integrating `PDB` information into `RzTypes`, There is no class type for now to store.

**Test plan**

...

**Closing issues**

Partially addresses https://github.com/rizinorg/rizin/issues/1205

### DIFF
--- a/librz/include/rz_type.h
+++ b/librz/include/rz_type.h
@@ -74,6 +74,7 @@ typedef enum {
 	RZ_BASE_TYPE_KIND_ENUM,
 	RZ_BASE_TYPE_KIND_TYPEDEF,
 	RZ_BASE_TYPE_KIND_ATOMIC, // For real atomic base types
+	RZ_BASE_TYPE_KIND_CLASS
 } RzBaseTypeKind;
 
 typedef struct rz_type_enum_case_t {
@@ -85,19 +86,30 @@ typedef struct rz_type_struct_member_t {
 	char *name;
 	RzType *type;
 	size_t offset; // in bytes
-	size_t size; // in bits?
+	size_t size; // in bits
 } RzTypeStructMember;
+
+typedef struct rz_type_class_member_t {
+	char *name;
+	RzType *type;
+	size_t offset; // in bytes
+	size_t size; // in bits
+} RzTypeClassMember;
 
 typedef struct rz_type_union_member_t {
 	char *name;
 	RzType *type;
 	size_t offset; // in bytes
-	size_t size; // in bits?
+	size_t size; // in bits
 } RzTypeUnionMember;
 
 typedef struct rz_base_type_struct_t {
 	RzVector /*<RzTypeStructMember>*/ members;
 } RzBaseTypeStruct;
+
+typedef struct rz_base_type_class_t {
+	RzVector /*<RzTypeClassMember>*/ members;
+} RzBaseTypeClass;
 
 typedef struct rz_base_type_union_t {
 	RzVector /*<RzTypeUnionMember>*/ members;
@@ -117,6 +129,7 @@ typedef struct rz_base_type_t {
 		RzBaseTypeStruct struct_data;
 		RzBaseTypeEnum enum_data;
 		RzBaseTypeUnion union_data;
+		RzBaseTypeClass class_data;
 	};
 } RzBaseType;
 
@@ -134,7 +147,8 @@ typedef enum {
 	RZ_TYPE_IDENTIFIER_KIND_UNSPECIFIED,
 	RZ_TYPE_IDENTIFIER_KIND_STRUCT,
 	RZ_TYPE_IDENTIFIER_KIND_UNION,
-	RZ_TYPE_IDENTIFIER_KIND_ENUM
+	RZ_TYPE_IDENTIFIER_KIND_ENUM,
+	RZ_TYPE_IDENTIFIER_KIND_CLASS
 } RzTypeIdentifierKind;
 
 typedef struct rz_callable_arg_t {
--- a/librz/type/base.c
+++ b/librz/type/base.c
@@ -25,6 +25,13 @@ RZ_API void rz_type_base_union_member_free(void *e, void *user) {
 	free((char *)member->name);
 }
 
+RZ_API void rz_type_base_class_member_free(void *e, void *user) {
+	(void)user;
+	RzTypeClassMember *member = e;
+	rz_type_free(member->type);
+	free((char *)member->name);
+}
+
 /**
  * \brief Searches for the RzBaseType in the types database given the name
  *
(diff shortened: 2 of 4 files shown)


librz/include/rz_type.h
```diff
@@ -70,6 +70,7 @@ typedef enum {
 
 typedef enum {
 	RZ_BASE_TYPE_KIND_STRUCT,
+	RZ_BASE_TYPE_KIND_CLASS, // Just a "placeholder" for now.
```
XVilka: I think it's better to put it after ATOMIC one.

librz/include/rz_type.h
```diff
@@ -88,6 +89,13 @@ typedef struct rz_type_struct_member_t {
 	size_t size; // in bits?
 } RzTypeStructMember;
 
+typedef struct rz_type_class_member_t {
+	char *name;
+	RzType *type;
+	size_t offset; // in bytes
+	size_t size; // in bits?
```
XVilka: Remove `?` here.

librz/include/rz_type.h
```diff
@@ -88,6 +89,13 @@ typedef struct rz_type_struct_member_t {
 	size_t size; // in bits?
```
XVilka: And here while you are on it

librz/include/rz_type.h
```diff
@@ -115,6 +127,7 @@ typedef struct rz_base_type_t {
 	RzTypeAttribute attrs;
 	union {
 		RzBaseTypeStruct struct_data;
+		RzBaseTypeClass class_data;
 		RzBaseTypeEnum enum_data;
 		RzBaseTypeUnion union_data;
```
XVilka, line 131: Better to put it after `union_data`

librz/include/rz_type.h
```diff
@@ -133,6 +146,7 @@ typedef enum {
 typedef enum {
 	RZ_TYPE_IDENTIFIER_KIND_UNSPECIFIED,
 	RZ_TYPE_IDENTIFIER_KIND_STRUCT,
+	RZ_TYPE_IDENTIFIER_KIND_CLASS,
 	RZ_TYPE_IDENTIFIER_KIND_UNION,
 	RZ_TYPE_IDENTIFIER_KIND_ENUM
```
XVilka: Same here, after `KIND_ENUM`

librz/type/base.c
```diff
@@ -18,6 +18,13 @@ RZ_API void rz_type_base_struct_member_free(void *e, void *user) {
 	free((char *)member->name);
 }
 
+RZ_API void rz_type_base_class_member_free(void *e, void *user) {
```
XVilka: Move this function after `union_member_free()`

librz/type/base.c
```diff
@@ -117,6 +124,9 @@ RZ_API void rz_type_base_type_free(RzBaseType *type) {
 	case RZ_BASE_TYPE_KIND_STRUCT:
 		rz_vector_fini(&type->struct_data.members);
 		break;
+	case RZ_BASE_TYPE_KIND_CLASS:
```
XVilka: Same, at the end, after all C types

librz/type/base.c
```diff
@@ -147,6 +157,9 @@ RZ_API RZ_OWN RzBaseType *rz_type_base_type_new(RzBaseTypeKind kind) {
 	case RZ_BASE_TYPE_KIND_STRUCT:
 		rz_vector_init(&type->struct_data.members, sizeof(RzTypeStructMember), rz_type_base_struct_member_free, NULL);
 		break;
+	case RZ_BASE_TYPE_KIND_CLASS:
```
XVilka: And here

librz/type/base.c
```diff
@@ -193,6 +206,17 @@ RZ_API RZ_OWN char *rz_type_db_base_type_as_string(const RzTypeDB *typedb, RZ_NO
 		rz_strbuf_append(buf, " }");
 		break;
 	}
+	case RZ_BASE_TYPE_KIND_CLASS: {
```
XVilka: And here

XVilka (changes requested):

XVilka:
Can you add a test in `test/unit/test_types.c` also? We cannot parse them yet, but surely will be able to form them manually.
Also we need to think about serialization/deserialization. cc @thestr4ng3r

librz/include/rz_type.h
```diff
@@ -74,6 +74,7 @@ typedef enum {
 	RZ_BASE_TYPE_KIND_ENUM,
 	RZ_BASE_TYPE_KIND_TYPEDEF,
 	RZ_BASE_TYPE_KIND_ATOMIC, // For real atomic base types
+	RZ_BASE_TYPE_KIND_CLASS, // Just a "placeholder" for now.
```
XVilka: I don't think we need to mark it as a placeholder. Some basic features will work already.

XVilka (approved):
Looks good now. But please add a test.

wargio (approved):

test/unit/test_type.c
```diff
@@ -488,6 +488,146 @@ static bool test_struct_func_types(void) {
 	mu_end;
 }
 
+static bool test_form_class_type_maually(void) {
```
XVilka: `s/maually/manually/g`

librz/include/rz_type.h
```diff
@@ -85,20 +86,31 @@ typedef struct rz_type_struct_member_t {
 	char *name;
 	RzType *type;
 	size_t offset; // in bytes
-	size_t size; // in bits?
```
XVilka, line 88: I was actually wondering, if we switch `offset` measure unit from `bytes` to `bits` we will get the [bitfields support](https://github.com/rizinorg/rizin/issues/1240) for free. @Basstorm @thestr4ng3r @ret2libc @wargio what do you think?

XVilka (approved):

thestr4ng3r (commented):
I am not sure if this overall is a good idea. Classes and types are conceptually very different. Normally you will have a class and optionally an associated type (which then would be just a struct). For example you can also have a class that has no type at all, like a pure abstract one or one with only static members.
That is the idea behind analysis classes, which already exist and could be extended to refer to types too.

librz/include/rz_type.h
wargio: yes, but change the definition to ut64, so we are sure that is really an ut64

thestr4ng3r: sgtm

ret2libc:
> I am not sure if this overall is a good idea. Classes and types are conceptually very different. Normally you will have a class and optionally an associated type (which then would be just a struct). For example you can also have a class that has no type at all, like a pure abstract one or one with only static members.
That is the idea behind analysis classes, which already exist and could be extended to refer to types too.

@XVilka was this addressed anywhere? What is the decision here? Is this PR still necessary?

XVilka:
@ret2libc we will prepare the different PR with a different approach later, but please keep this open until we open the new one. There is some useful stuff in it.

XVilka:
We decided to make this a bit differently.

PR #1354, opened 2021-07-23
Add configuration variables (elf.load.sections, elf.checks.sections, elf.checks.segments)

ghost:

**Detailed description**

* Remove field init and fini from RzBinPlugin
* Add the field RzConfig *config to load buffer
* Add elf.load.sections
* Add elf.checks.segments
* Add elf.checks.sections

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/librz/bin/bin.c
+++ b/librz/bin/bin.c
@@ -109,6 +109,7 @@ RZ_API void rz_bin_options_init(RzBinOptions *opt, int fd, ut64 baseaddr, ut64 l
 	opt->obj_opts.baseaddr = baseaddr;
 	opt->obj_opts.loadaddr = loadaddr;
 	opt->obj_opts.patch_relocs = patch_relocs;
+	opt->obj_opts.elf_load_sections = true;
 	opt->fd = fd;
 	opt->rawstr = rawstr;
 }
--- a/librz/bin/bin.c
+++ b/librz/bin/bin.c
@@ -390,9 +390,6 @@ RZ_IPI RzBinXtrPlugin *rz_bin_get_xtrplugin_by_name(RzBin *bin, const char *name
 }
 
 static void rz_bin_plugin_free(RzBinPlugin *p) {
-	if (p && p->fini) {
-		p->fini(NULL);
-	}
 	RZ_FREE(p);
 }
 
@@ -402,9 +399,6 @@ RZ_API bool rz_bin_plugin_add(RzBin *bin, RzBinPlugin *foo) {
 
 	rz_return_val_if_fail(bin && foo, false);
 
-	if (foo->init) {
-		foo->init(bin->user);
-	}
 	rz_list_foreach (bin->plugins, it, plugin) {
 		if (!strcmp(plugin->name, foo->name)) {
 			return false;
--- a/librz/bin/bin.c
+++ b/librz/bin/bin.c
@@ -234,9 +234,17 @@ RZ_API RzBinFile *rz_bin_open(RzBin *bin, const char *file, RzBinOptions *opt) {
 
 RZ_API RzBinFile *rz_bin_reload(RzBin *bin, RzBinFile *bf, ut64 baseaddr) {
 	rz_return_val_if_fail(bin && bf, NULL);
+
 	bool patch_relocs = bf->o ? bf->o->opts.patch_relocs : false;
+	bool elf_load_sections = bf->o ? bf->o->opts.elf_load_sections : false;
+	bool elf_checks_sections = bf->o ? bf->o->opts.elf_checks_sections : false;
+	bool elf_checks_segments = bf->o ? bf->o->opts.elf_checks_segments : false;
+
 	RzBinOptions opt;
 	rz_bin_options_init(&opt, bf->fd, baseaddr, bf->loadaddr, patch_relocs, bin->rawstr);
+	opt.obj_opts.elf_load_sections = elf_load_sections;
+	opt.obj_opts.elf_checks_sections = elf_checks_sections;
+	opt.obj_opts.elf_checks_segments = elf_checks_segments;
 	opt.filename = bf->file;
 	rz_buf_seek(bf->buf, 0, RZ_BUF_SET);
 	RzBinFile *nbf = rz_bin_open_buf(bin, bf->buf, &opt);
--- a/librz/core/cbin.c
+++ b/librz/core/cbin.c
@@ -172,7 +172,20 @@ static ut64 rva(RzBinObject *o, ut64 paddr, ut64 vaddr, int va) {
 
 RZ_API void rz_core_bin_options_init(RzCore *core, RZ_OUT RzBinOptions *opts, int fd, ut64 baseaddr, ut64 loadaddr) {
 	rz_return_if_fail(core && opts);
-	rz_bin_options_init(opts, fd, baseaddr, loadaddr, rz_config_get_b(core->config, "bin.relocs"), core->bin->rawstr);
+
+	bool patch_relocs = rz_config_get_b(core->config, "bin.relocs");
+
+	rz_bin_options_init(
+		opts,
+		fd,
+		baseaddr,
+		loadaddr,
+		patch_relocs,
+		core->bin->rawstr);
+
+	opts->obj_opts.elf_load_sections = rz_config_get_b(core->config, "elf.load.sections");
+	opts->obj_opts.elf_checks_sections = rz_config_get_b(core->config, "elf.checks.sections");
+	opts->obj_opts.elf_checks_segments = rz_config_get_b(core->config, "elf.checks.segments");
 }
 
 RZ_API int rz_core_bin_set_by_fd(RzCore *core, ut64 bin_fd) {
--- a/librz/core/cfile.c
+++ b/librz/core/cfile.c
@@ -635,7 +635,7 @@ static int rz_core_file_do_load_for_debug(RzCore *r, ut64 baseaddr, const char *
 	int fd = cf ? cf->fd : -1;
 
 	RzBinOptions opt;
-	rz_bin_options_init(&opt, fd, baseaddr, UT64_MAX, false, rz_config_get_i(r->config, "bin.rawstr"));
+	rz_bin_options_init(&opt, fd, baseaddr, UT64_MAX, false, false);
 	opt.obj_opts.elf_load_sections = rz_config_get_b(r->config, "elf.load.sections");
 	opt.obj_opts.elf_checks_sections = rz_config_get_b(r->config, "elf.checks.sections");
 	opt.obj_opts.elf_checks_segments = rz_config_get_b(r->config, "elf.checks.segments");
@@ -644,6 +644,12 @@ static int rz_core_file_do_load_for_debug(RzCore *r, ut64 baseaddr, const char *
 	if (!binfile) {
 		eprintf("RzBinLoad: Cannot open %s\n", filenameuri);
 		if (rz_config_get_i(r->config, "bin.rawstr")) {
+			rz_bin_options_init(&opt, fd, baseaddr, UT64_MAX, false, true);
+			opt.obj_opts.elf_load_sections = rz_config_get_b(r->config, "elf.load.sections");
+			opt.obj_opts.elf_checks_sections = rz_config_get_b(r->config, "elf.checks.sections");
+			opt.obj_opts.elf_checks_segments = rz_config_get_b(r->config, "elf.checks.segments");
+			opt.xtr_idx = xtr_idx;
+
 			binfile = rz_bin_open(r->bin, filenameuri, &opt);
 			if (!binfile) {
 				return false;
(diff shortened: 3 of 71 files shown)


thestr4ng3r (changes requested):
The mechanism to disable loading certain kinds of information is good, but the problem currently is that RzBin should not depend on RzConfig.
The reason is that RzConfig is more of a generic config framework and what variables exist and what they mean is purely defined by the user of it, which in this case is core, so indirectly this makes bin depend on core or alternatively something that build an `RzConfig` object that mimics what RzCore normally does, which is quite fragile.

The right place to put loading config is in `RzBinObjectLoadOptions`, which already contains similar options. See for example https://github.com/rizinorg/rizin/blob/71bf245a9b47d7fdf2c64b8f58354e685d2e81c4/librz/bin/p/bin_elf.inc#L466 how this is used in elf for another option already.
It's a bit unfortunate that currently one has to go through the entire `bf->o->opts` chain. I do prefer the way you approached this by passing it to `load_buffer` but that's optional.

About how to put these three specific options in `RzBinObjectLoadOptions` exactly, I think `load.sections` could theoretically be generic and also apply to all other plugins (then bin shouldn't call the sections callback at all and thus also not fill the RzBinObject's section list). However there is also the problem that this would then also mean that elf segments (which are RzBinSections too) would not show up, so it probably still makes sense to have this elf-specific.
So as a starter, I would just add some more bools like `bool elf_load_sections`, etc. to `RzBinObjectLoadOptions` directly, with the easy possibility to refactor this. But maybe @ret2libc @XVilka @wargio also have some feedback for this.

XVilka:
I agree with @thestr4ng3r  on not linking `RzConfig` to `RzBin` and using additional structure field for options. It's very ELF specific so better not to introduce generic solution, just ELF one.

ghost:
The idea is to use `RzBinObjectLoadOptions` and populate elf_checks_segments, elf_load_sections and elf_checks_sections with `RzConfig` from core? So RzBin never use RzConfig. Am I right?

ghost:
Should I introduce a new struct in librz/bin/format/elf/?

```c
struct rz_bin_elf_load_options
    bool checks_sections;
    bool checks_segments;
    bool load_sections;
} RzBinElfLoadOptions;
```

And add this struct to RzBinObjectLoadOptions

ret2libc:
I would go with just adding `bool`s directly to `RzBinObjectLoadOptions` for now to avoid overengineering from the beginning.

ghost:
> I would go with just adding `bool`s directly to `RzBinObjectLoadOptions` for now to avoid overengineering from the beginning.

It is just that 4 bool has a parameter of function is error prone.

ret2libc:
> > I would go with just adding `bool`s directly to `RzBinObjectLoadOptions` for now to avoid overengineering from the beginning.
> 
> It is just that 4 bool has a parameter of function is error prone.

You don't need to pass them as parameters AFAIK. As @thestr4ng3r said you can access them with `bf->o->opts.XXX`

ghost:
> > > I would go with just adding `bool`s directly to `RzBinObjectLoadOptions` for now to avoid overengineering from the beginning.
> > 
> > 
> > It is just that 4 bool has a parameter of function is error prone.
> 
> You don't need to pass them as parameters AFAIK. As @thestr4ng3r said you can access them with `bf->o->opts.XXX`

Indeed, but I failed to explain my self. RzBinOptions is constructed with rz_bin_options_init and this is this function that has 4 bool has parameters.

ret2libc:
> > > > I would go with just adding `bool`s directly to `RzBinObjectLoadOptions` for now to avoid overengineering from the beginning.
> > > 
> > > 
> > > It is just that 4 bool has a parameter of function is error prone.
> > 
> > 
> > You don't need to pass them as parameters AFAIK. As @thestr4ng3r said you can access them with `bf->o->opts.XXX`
> 
> Indeed, but I failed to explain my self. RzBinOptions is constructed with rz_bin_options_init and this is this function that has 4 bool has parameters.

Ah that's not a problem. That function does not really set all fields within the structure, just few common ones. Each caller of `rz_bin_options_init` can set additional fields as necessary.

ghost:
I modified the interface load_buffer, because the old interface prevent me from using RzBinObjectLoadOPtions.

librz/bin/bin.c
```diff
@@ -104,11 +104,14 @@ RZ_API RzList *rz_bin_dump_strings(RzBinFile *bf, int min, int raw) {
 	return rz_bin_file_get_strings(bf, min, 1, raw);
 }
 
-RZ_API void rz_bin_options_init(RzBinOptions *opt, int fd, ut64 baseaddr, ut64 loadaddr, bool patch_relocs, int rawstr) {
+RZ_API void rz_bin_options_init(RzBinOptions *opt, int fd, ut64 baseaddr, ut64 loadaddr, bool patch_relocs, bool elf_load_sections, bool elf_checks_sections, bool elf_checks_segments, int rawstr) {
```
ret2libc: As I wrote in the comment I would not put these as parameters of this function, but just set them outside of it when necessary.

librz/core/cbin.c
```diff
@@ -172,7 +172,22 @@ static ut64 rva(RzBinObject *o, ut64 paddr, ut64 vaddr, int va) {
...
+	rz_bin_options_init(
+		opts,
+		fd,
+		baseaddr,
+		loadaddr,
+		patch_relocs,
+		elf_load_sections,
+		elf_checks_sections,
+		elf_checks_segments,
+		core->bin->rawstr);
```
ret2libc: instead of adding that many args, just set the elf_ fields after this call.

librz/core/cfile.c
```diff
@@ -658,15 +658,19 @@ static int rz_core_file_do_load_for_debug(RzCore *r, ut64 baseaddr, const char *
...
 	}
 #endif
+	bool elf_load_sections = rz_config_get_b(r->config, "elf.load.sections");
+	bool elf_checks_sections = rz_config_get_b(r->config, "elf.checks.sections");
+	bool elf_checks_segments = rz_config_get_b(r->config, "elf.checks.segments");
 	int fd = cf ? cf->fd : -1;
+
 	RzBinOptions opt;
-	rz_bin_options_init(&opt, fd, baseaddr, UT64_MAX, false, false);
+	rz_bin_options_init(&opt, fd, baseaddr, UT64_MAX, false, elf_load_sections, elf_checks_sections, elf_checks_segments, false);
```
ret2libc: try to use `rz_core_bin_options_init` instead.

librz/core/cfile.c
```diff
@@ -658,15 +658,19 @@ static int rz_core_file_do_load_for_debug(RzCore *r, ut64 baseaddr, const char *
...
 	opt.xtr_idx = xtr_idx;
 	RzBinFile *binfile = rz_bin_open(r->bin, filenameuri, &opt);
 	if (!binfile) {
 		eprintf("RzBinLoad: Cannot open %s\n", filenameuri);
 		if (rz_config_get_i(r->config, "bin.rawstr")) {
-			rz_bin_options_init(&opt, fd, baseaddr, UT64_MAX, false, true);
+			rz_bin_options_init(&opt, fd, baseaddr, UT64_MAX, false, elf_load_sections, elf_checks_sections, elf_checks_segments, true);
```
ret2libc: try to use `rz_core_bin_options_init` instead.

librz/core/linux_heap_glibc.c
```diff
@@ -46,8 +46,12 @@ static GHT GH(get_va_symbol)(RzCore *core, const char *path, const char *sym_nam
 	RzListIter *iter;
 	RzBinSymbol *s;
 
+	bool elf_load_sections = rz_config_get_b(core->config, "elf.load.sections");
+	bool elf_checks_sections = rz_config_get_b(core->config, "elf.checks.sections");
+	bool elf_checks_segments = rz_config_get_b(core->config, "elf.checks.segments");
+
 	RzBinOptions opt;
-	rz_bin_options_init(&opt, -1, 0, 0, false, false);
+	rz_bin_options_init(&opt, -1, 0, 0, false, elf_load_sections, elf_checks_sections, elf_checks_segments, false);
```
ret2libc: try to use `rz_core_bin_options_init` instead.

librz/core/linux_heap_jemalloc.c
```diff
@@ -39,8 +39,12 @@ static GHT GH(je_get_va_symbol)(const char *path, const char *symname) {
 		return GHT_MAX;
 	}
 
+	bool elf_load_sections = rz_config_get_b(core->config, "elf.load.sections");
+	bool elf_checks_sections = rz_config_get_b(core->config, "elf.checks.sections");
+	bool elf_checks_segments = rz_config_get_b(core->config, "elf.checks.segments");
+
 	RzBinOptions opt;
-	rz_bin_options_init(&opt, -1, 0, 0, false, false);
+	rz_bin_options_init(&opt, -1, 0, 0, false, elf_load_sections, elf_checks_sections, elf_checks_segments, false);
```
ret2libc: try to use `rz_core_bin_options_init` instead.

librz/main/rz-bin.c
```diff
@@ -1045,8 +1045,12 @@ RZ_API int rz_main_rz_bin(int argc, const char **argv) {
 	rz_bin_force_plugin(bin, forcebin);
 	rz_bin_load_filter(bin, action);
 
+	bool elf_load_sections = rz_config_get_b(core.config, "elf.load.sections");
+	bool elf_checks_sections = rz_config_get_b(core.config, "elf.checks.sections");
+	bool elf_checks_segments = rz_config_get_b(core.config, "elf.checks.segments");
+
 	RzBinOptions bo;
-	rz_bin_options_init(&bo, fd, baddr, laddr, false, rawstr);
+	rz_bin_options_init(&bo, fd, baddr, laddr, false, elf_load_sections, elf_checks_sections, elf_checks_segments, rawstr);
```
ret2libc: try to use `rz_core_bin_options_init` instead.

ret2libc (commented):
The change to `load_buffer` API seems reasonable...

ghost:
@ret2libc There is some unresolved comment when `rz_core_bin_options_init` couldn't be used.

librz/core/cfile.c
```diff
@@ -659,15 +659,17 @@ static int rz_core_file_do_load_for_debug(RzCore *r, ut64 baseaddr, const char *
 	}
 #endif
 	int fd = cf ? cf->fd : -1;
+
 	RzBinOptions opt;
-	rz_bin_options_init(&opt, fd, baseaddr, UT64_MAX, false, false);
+	rz_bin_options_init(&opt, fd, baseaddr, UT64_MAX, false, rz_config_get_i(r->config, "bin.rawstr"));
+	opt.obj_opts.elf_load_sections = rz_config_get_b(r->config, "elf.load.sections");
+	opt.obj_opts.elf_checks_sections = rz_config_get_b(r->config, "elf.checks.sections");
+	opt.obj_opts.elf_checks_segments = rz_config_get_b(r->config, "elf.checks.segments");
```
ret2libc: This seems different from previous behaviour. Was it tested?

test/unit/test_dwarf.c
```diff
@@ -82,6 +82,7 @@ bool test_dwarf3_c_basic(void) { // this should work for dwarf2 aswell
 	rz_io_bind(io, &bin->iob);
 
 	RzBinOptions opt = { 0 };
+	opt.obj_opts.elf_load_sections = true;
```
ret2libc: I see you have to do this all the time, so maybe it would make sense to have a variable that express the opposite of this, so it is false by default. Like... `elf_noload_sections`. WDYT?

librz/core/cfile.c
ghost: Yes, in linux64/*. I believe.

test/unit/test_dwarf.c
ghost: Negative sentence variable is a code smell. I don't know if those could be rewrite with rz_bin_init_options (eld_load_sections will be set to true in this function)

ret2libc: right, better to set it to true directly in rz_bin_init_options.

ghost: The PR has been udpated, WDYT?

XVilka (approved):

librz/core/cfile.c
```diff
@@ -665,15 +665,17 @@ static int rz_core_file_do_load_for_debug(RzCore *r, ut64 baseaddr, const char *
 	}
 #endif
 	int fd = cf ? cf->fd : -1;
+
 	RzBinOptions opt;
-	rz_bin_options_init(&opt, fd, baseaddr, UT64_MAX, false, false);
+	rz_bin_options_init(&opt, fd, baseaddr, UT64_MAX, false, rz_config_get_i(r->config, "bin.rawstr"));
```
ret2libc:
```suggestion
	rz_bin_options_init(&opt, fd, baseaddr, UT64_MAX, false, false);
```
I would keep old behaviour (and also few lines below), unless this is carefully tested and we make sure the reason for doing as it was it not right/relevant any more.

ghost: I revert the behavior.

ret2libc (approved):

ret2libc:
Mark as ready if so. Thanks!

thestr4ng3r (approved):

ghost:
@ret2libc Can I switched to rebase (and not squash). I used atomic commit to simplify debug.

Issue #1341, opened 2021-07-21
Fix PE: corkami 65535sects.exe - section list, entrypoint, open and analyze with ASAN build


XVilka:
It is unnecessary slow. Could be and should be optimized. It fails on our CI
```
[XX] db/formats/pe/65535sects PE: corkami 65535sects.exe - section list, entrypoint, open and analyze
RZ_NOPLUGINS=1 rizin -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -Qc 'aa
om~?
s
pi 1
q!
' bins/pe/65535sects.exe
-- stdout
--- expected
+++ actual
@@ -1,3 +1,0 @@
-65536
-0x291120
-mov edi, 0x7027aff9

-- stderr
[ ] Analyze all flags starting with sym. and entry0 (aa)
[
[x] Analyze all flags starting with sym. and entry0 (aa)

-- exit status: -1
```

p371jah:
I guess it's just killed by timeout.

wargio:
this behaviour changed with the new code for the Type Analysis.

ret2libc:
Not sure of the status of this... Seems it worked in some cases recently. Let's see.

wargio:
if you check the phy size and ignore any 0 sizes when allocating the PE data structure, you will notice that the test will be super fast.
now the issue is that we should rule out any invalid data structure that is not a known used PE section.

wargio:
is this fixed? dev is green

ret2libc:
It seems so, but let's wait few days and then we can close this and probably also another one.

ret2libc:
Seems green lately!

PR #1364, opened 2021-07-27
Fix memleaks in librz/type/*

XVilka:
# SQUASH ME, SQUEEZE ME, SQUELCH ME

**Detailed description**

Imported from https://github.com/rizinorg/rizin/pull/1363 to trigger more CI jobs

**Test plan**

CI is green

### DIFF
--- a/librz/type/parser/c_cpp_parser.c
+++ b/librz/type/parser/c_cpp_parser.c
@@ -372,7 +372,9 @@ RZ_API RZ_OWN RzType *rz_type_parse_string_single(RzTypeParser *parser, const ch
 	ts_tree_delete(tree);
 	ts_parser_delete(tsparser);
 	free(patched_code);
-	return tpair ? tpair->type : NULL;
+	RzType *ret = tpair ? tpair->type : NULL;
+	free(tpair);
+	return ret;
 }
 
 /**
--- a/librz/type/parser/types_parser.c
+++ b/librz/type/parser/types_parser.c
@@ -197,9 +197,14 @@ int parse_sole_type_name(CParserState *state, TSNode node, const char *text, Par
 	// Do allow forward-looking definitions we just add the type into the forward hashtable
 	if (c_parser_forward_definition_store(state, real_type)) {
 		parser_debug(state, "Added forward definition of type: \"%s\"\n", real_type);
+		rz_type_base_type_free((*tpair)->btype);
+		(*tpair)->btype = NULL;
 		free(real_type);
 		return 0;
 	}
+	rz_type_free((*tpair)->type);
+	rz_type_base_type_free((*tpair)->btype);
+	RZ_FREE(*tpair);
 	free(real_type);
 	return -1;
 }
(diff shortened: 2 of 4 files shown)


librz/type/parser/c_cpp_parser.c
```diff
@@ -372,7 +372,9 @@ RZ_API RZ_OWN RzType *rz_type_parse_string_single(RzTypeParser *parser, const ch
 	ts_tree_delete(tree);
 	ts_parser_delete(tsparser);
 	free(patched_code);
-	return tpair ? tpair->type : NULL;
+	RzType *ret = tpair ? tpair->type : NULL;
+	free(tpair);
```
ret2libc, line 376: Is it enough to use `free` on tpair? What about `tpair->base`? Is that still allocated?

librz/type/parser/types_parser.c
```diff
@@ -370,6 +381,7 @@ int parse_struct_node(CParserState *state, TSNode node, const char *text, Parser
 		TSNode first_leaf = ts_node_named_child(child, 0);
 		if (ts_node_is_null(first_leaf)) {
 			node_malformed_error(state, child, text, "field_declaration");
+			free(name);
```
ret2libc: I think all these `free(name)` mean we should switch to `goto err` instead.

librz/type/serialize_functions.c
```diff
@@ -60,6 +60,7 @@ static RzCallable *get_callable_type(RzTypeDB *typedb, Sdb *sdb, const char *nam
 		goto error;
 	}
 
+	char arg_name[32];
```
ret2libc: i'd place this inside the `for` loop as well.

librz/type/parser/c_cpp_parser.c
p371jah: tpair->btype is stored in a hashtable.

XVilka:
ASAN failure is unrelated, it's https://github.com/rizinorg/rizin/issues/1341

librz/type/parser/c_cpp_parser.c
XVilka: Yes, in general `RzType` are allocated and owned by the user, `RzBaseType` are stored in the hashtable and only borrowed (except edit/remove cases).

ret2libc: Ok!

librz/type/serialize_functions.c
XVilka: Why? There is a need of only one static buffer so it can be allocated on stack only once (depending on the compiler of course).

librz/type/parser/types_parser.c
XVilka: Done.

librz/type/serialize_functions.c
ret2libc: Because that's compiler's work. They are smart enough to allocate that just one time and not on every loop iteration. It's better to define things in the smallest scope needed (e.g. otherwise you could also put `char *values` at line 65 here, or other things...).

When I see `char arg_name[32]` here I don't know where this var is used exactly unless I read the whole function after this line. If I see this same definition inside the loop I know for sure that nothing else outside of the loop can use this buffer.

XVilka: OK, moved then.

ret2libc (approved):

PR #1369, opened 2021-07-28
Reorganize map skyline segments

GustavoLCR:

**Detailed description**

* Segments are added in the order that they show up, so that the map skyline is properly formed.
* Improves section load times on PE:
   - PE sections are organized in ascending order. Before, by adding sections from the highest `vaddr` to the lowest, the new map would be added to the start of the skyline array, hitting an expensive `memmove()` over the entire array every time. As this reverses the order in which maps are added, new maps are added to the end of the array with a simple memcpy() to the end of the buffer.

**Test plan**

Tests pass

**Closing issues**

None


XVilka:
@GustavoLCR does this solve also this problem? https://github.com/rizinorg/rizin/issues/815

XVilka:
```


The output from the failed tests:

23/85 core_bin                       FAIL            0.79s   exit status 1
--- command ---
06:03:10 /home/runner/work/rizin/rizin/build/test/unit/test_core_bin
--- stdout ---
test_map ERR
[XX] Fail at line 116: io map name: expected mmap.vfile map with zeroes, got fmap.direct map.

test_cfile_close OK
test_cfile_close_multiple OK
test_cfile_close_manual_maps OK
test_cfile_close_manual_fd OK
test_cfile_close_manual_vfile_fd OK
test_cfile_close_manual_vfile_map ERR
[XX] Fail at line 504: vfile desc name: expected vfile://0/vfile0, got vfile://0/vfile1.
```

ret2libc:
If you recreate the PR with a `asan-` in the branch name it will also run the asan tests and we could directly check if by any chance it fixes the issue there related to 65535sects.

librz/core/cbin.c
```diff
@@ -790,16 +790,14 @@ RZ_API bool rz_core_bin_apply_maps(RzCore *core, RzBinFile *binfile, bool va) {
 
 	RzListIter *it;
 	RzBinMap *map;
-	// reverse because of the way io handles maps
```
ret2libc, line 806: Do you know what did this mean? I guess there was a reason to handling things in reverse order... is that not relevant anymore? Maybe adding a bit of explanation of why this is not relevant anymore would help even for future reference.

GustavoLCR:
> @GustavoLCR does this solve also this problem? #815

@XVilka I dont think so

librz/core/cbin.c
GustavoLCR: @ret2libc In the past RzIO would go through the maps array backwards to construct the skyline vector. This isn't true anymore

ret2libc: wasn't the reason for that to assign the proper priorities to each maps? I have fuzzy memories about that.

ghost (approved):

librz/core/cbin.c
XVilka: I think @thestr4ng3r knows best of this part's changes.

XVilka (approved):

XVilka:
Looks like it broke a lot of tests:  https://builds.sr.ht/~xvilka/job/554747

p371jah:
I guess you need to change this line https://github.com/rizinorg/rizin/blob/58433a72accfee45dce6085f49d3f7b84a56aa05/librz/core/cmd_open.c#L398

librz/core/cbin.c
```diff
@@ -803,16 +803,14 @@ RZ_API bool rz_core_bin_apply_maps(RzCore *core, RzBinFile *binfile, bool va) {
 
 	RzListIter *it;
 	RzBinMap *map;
```
p371jah, line 805: BTW since number of maps is known we can reserve memory for them in the skyline vector to avoid reallocations.

XVilka (approved):

librz/core/cbin.c
thestr4ng3r: This is my last knowledge of the priorities:

https://github.com/rizinorg/rizin/blob/5f5c0d66dfdc01db1af782bb024c2873fdd27a1a/librz/include/rz_io.h#L73

Which I think makes sense, to specify the priorities by their position in the vector. Whether it's reverse or not is then merely convention (and maybe slightly relevant for insertion/removal performance).

> This isn't true anymore

Was this changed at some point?

GustavoLCR: @thestr4ng3r I checked again and confirmed that new maps are pushed to the end of the vector as they are added, and they have higher priority, this never changed, what changed was an implementation detail of how the skyline is created and that should be unrelated to this.
Why this code bothered to go through maps in reverse order, I don't know.
This would only work if it was a requirement that `RzBinObject->maps` always had the highest priority maps as the first elements, but there isn't anywhere where that that is specified, so this never was a guarantee anyway.

XVilka: Good. Lets see if @thestr4ng3r approves and lets merge this. 👍

wargio:
why this PR is called `Fix loading overlapped ELF segments` when it only touches Java and Dex ?

ret2libc (approved):
Seems ok for me, but let's also get @thestr4ng3r 's approval.

thestr4ng3r (approved):

Issue #1294, opened 2021-07-12
Migrate ESIL and Debug trace from SDB to hashtable


XVilka:
Currently ESIL trace used heavily in deep levels of analysis and during the type propagation stage. Every instructions produces SDB KV, analysis go over them multiple times recursively (depending on the code path). It slows down the analysis process significantly.
Storing everything in the hashtable should improve the situation. Curiously, part of the ESIL tracing already uses HT, but some part - uses SDB.

See:
- `librz/analysis/esil_trace.c`

```
[i] ℤ rg "esil->trace"                                                                                                                                                                                                            15:00:38 
librz/core/cmd_debug.c
4594:                                           sdb_free(core->analysis->esil->trace->db);
4595:                                           core->analysis->esil->trace->db = sdb_new0();
4608:                                   char *s = sdb_querys(core->analysis->esil->trace->db,

librz/core/analysis_tp.c
16:     *et = core->analysis->esil->trace;
18:     core->analysis->esil->trace = rz_analysis_esil_trace_new(core->analysis->esil);
32:     return (core->dbg->trace && core->analysis->esil->trace);
39:     rz_analysis_esil_trace_free(core->analysis->esil->trace);
40:     core->analysis->esil->trace = et;
298:    Sdb *trace = core->analysis->esil->trace->db;
763:    Sdb *etracedb = core->analysis->esil->trace->db;
823:                    int loop_count = sdb_num_get(analysis->esil->trace->db, sdb_fmt("0x%" PFMT64x ".count", addr), 0);
827:                    sdb_num_set(analysis->esil->trace->db, sdb_fmt("0x%" PFMT64x ".count", addr), loop_count + 1, 0);
833:                    Sdb *trace = analysis->esil->trace->db;

librz/core/cmd_analysis.c
4142:   rz_return_val_if_fail(core->analysis->esil && core->analysis->esil->trace, -1);
4144:   if (esil->trace->idx > 0) {
4145:           rz_analysis_esil_trace_restore(esil, esil->trace->idx - 1);
4152:   rz_return_val_if_fail(core->analysis->esil && core->analysis->esil->trace, false);
4154:   if (esil->trace->idx == 0) {
4159:   RzVector *vreg = ht_up_find(esil->trace->registers, ripc->offset | (ripc->arena << 16), NULL);
4170:           if (reg->idx >= esil->trace->idx) {
5257:                           if (esil->trace) {
5261:                           esil->trace = rz_analysis_esil_trace_new(esil);
[...]

PR #1376, opened 2021-07-29
Migrate etrace sdb to hashtable 

Heersin:

**Detailed description**
Use ht_pp to replace `sdb` in etrace

**Test plan**
Should pass the old esil tests

**Closing issues**
#1294


Heersin:
```C
	// TODO : migrate dtrace from sdb to ht
	RzDebugTrace *dtrace = core->dbg->trace;
	opt = dtrace->ht->opt;
	ht_pp_free(dtrace->ht);
	dtrace->ht = ht_pp_new_size(fcn->ninstr, opt.dupvalue, opt.freefn, opt.calcsizeV);
	dtrace->ht->opt = opt;
```
As far as I see the debug trace has already migrated to `ht`, is it still a TODO work?

XVilka:
I think this TODO is outdated.

XVilka:
Some tests are broken after that.

Heersin:
> Some tests are broken after that.

Yeah, some type info in the output are lost.

librz/core/cmd_debug.c
```diff
@@ -4437,15 +4442,16 @@ RZ_IPI int rz_cmd_debug(void *data, const char *input) {
 				rz_analysis_esil_trace_show(
 					core->analysis->esil, idx);
 			} break;
+			// FIXME : should we mark this deprecated ?
```
XVilka: I think we should remove this, yes

librz/analysis/esil/esil_trace.c
```diff
@@ -18,6 +18,84 @@ static void htup_vector_free(HtUPKv *kv) {
 	rz_vector_free(kv->value);
 }
 
+static void htpp_string_free(HtPPKv *kv) {
+	free(kv->key);
+	free(kv->value);
+}
+
+RZ_API void ht_db_array_add(HtPP *db, const char *key, const char *val) {
```
XVilka: I think it's better to move those to the `librz/util/` instead.

XVilka (changes requested):

ret2libc (commented):
Why do you want to still work on esil if we are going to remove it? Where do you need it?

librz/analysis/esil/esil_trace.c
ret2libc: Also, for new RZ_API functions please provide doxygen style doc (and unit tests, if possible).
And... why do we want this? This seems to mimic old sdb behaviour which was really bad. If you need an array, instead of just concatenating string use an HtPP <char *, RzVector *>. Using strings everywhere is not good.

p371jah:
Since sdb uses HtPP in almost the same way I don't see the need in these changes.

XVilka:
@ret2libc we will just swap ESIL with RZIL in the same code after

librz/analysis/esil/esil_trace.c
```diff
@@ -18,6 +18,111 @@ static void htup_vector_free(HtUPKv *kv) {
...
 
+RZ_API RzPVector *ht_db_get(HtPP *db, const char *key) {
+	return ht_pp_find(db, key, NULL);
+}
+
+static void htpp_vector_free(HtPPKv *kv) {
+	rz_vector_free(kv->value);
+}
+
+RZ_API void ht_db_array_add(HtPP *db, const char *key, const char *val) {
```
ret2libc: does this need to be a RZ_API ? If so, you should name it in a better way because otherwise it may be confused as a generic ht function, however it is not.

Heersin: Actually it's only used for two files now. It's not really necessary to be an RZ_API, I will make them static after migrate the `string-based 'array'` to `RzPvector`

Heersin: Or If we really need something like it in other places. I will try to move it to `util` as Xvilka said. What do you think ?

librz/include/rz_analysis.h
```diff
@@ -1024,7 +1024,8 @@ typedef struct rz_analysis_esil_trace_t {
 	ut64 stack_size;
 	ut8 *stack_data;
 	//TODO remove `db` and reuse info above
-	Sdb *db;
+	// Sdb *db;
```
XVilka: Remove `Sdb *db` here too.

XVilka (changes requested):
-

librz/analysis/esil/esil_trace.c
```diff
@@ -126,8 +292,8 @@ static int trace_hook_reg_read(RzAnalysisEsil *esil, const char *name, ut64 *res
 	if (ret) {
 		ut64 val = *res;
 		//eprintf ("[ESIL] REG READ %s 0x%08"PFMT64x"\n", name, val);
-		sdb_array_add(DB, KEY("reg.read"), name, 0);
-		sdb_num_set(DB, KEYREG("reg.read", name), val, 0);
+		rz_analysis_esil_trace_db_array_add(DB, KEY("reg.read"), name);
```
XVilka: I wonder, what about creating 4 different hashtables for reading/writing memory, reading/writing registers? It should speed up the access as well. cc @pelijah @ret2libc

librz/analysis/esil/esil_trace.c
```diff
@@ -18,6 +18,172 @@ static void htup_vector_free(HtUPKv *kv) {
...
+ * @return vec RzPVector of char *, contains the address of read/write
+ */
+RZ_API RzPVector *rz_analysis_esil_trace_db_get(HtPP *db, const char *key) {
+	RzPVector *vec = ht_pp_find(db, key, NULL);
+
+	if (!vec) {
+		return NULL;
+	}
+
+	return vec;
```
ret2libc:
```suggestion
	return ht_pp_find(db, key, NULL);
```

librz/analysis/esil/esil_trace.c
```diff
@@ -18,6 +18,172 @@ static void htup_vector_free(HtUPKv *kv) {
...
+ */
+RZ_API void rz_analysis_esil_trace_db_array_add(HtPP *db, const char *key, const char *val) {
+	RzPVector *array = rz_analysis_esil_trace_db_get(db, key);
+	if (!array) {
+		// no corresponding value, create one in hash table
+		rz_analysis_esil_trace_db_set(db, key, val);
+		return;
+	}
+
+	if (rz_analysis_esil_trace_db_array_contains(db, key, val)) {
```
ret2libc: This means that `array` is retrieve, again, from the HT, when we already have it.

librz/analysis/esil/esil_trace.c
```diff
@@ -18,6 +18,172 @@ static void htup_vector_free(HtUPKv *kv) {
...
+ * @return val const char* address/reg/idx
+ */
+RZ_API const char *rz_analysis_esil_trace_db_const_get(HtPP *db, const char *key) {
+	RzPVector *array = rz_analysis_esil_trace_db_get(db, key);
+	if (!array) {
+		return NULL;
+	}
+
+	char *array_string = ht_db_array_to_string(array);
+	return array_string;
```
ret2libc:
```suggestion
	return ht_db_array_to_string(array);
```

ret2libc (commented):
Honestly I am not sure why we want to do this. ESIL is not where we are going, so why going through all this effort when what we had worked already for what it was supposed to do? That said, I think the whole "ESIL trace" thing should be a little be redesigned to understand well what you want to get and how.

Seeing
```
0.addr=0x4d8d
0.mem.write=0x177fd0
0.mem.write.data.0x177fd0=0000000000000000
0.reg.read=rsp,rbx
0.reg.read.rbx=0
0.reg.read.rsp=0x177fd8
0.reg.write=rsp
0.reg.write.rsp=0x177fd0
1.addr=0x4d8e
1.reg.read=rsp
1.reg.read.rsp=0x177fd0
1.reg.write=rsp,of,sf,zf,pf,cf,af
1.reg.write.af=0
1.reg.write.cf=0
1.reg.write.of=0
1.reg.write.pf=0
1.reg.write.rsp=0x177f68
1.reg.write.sf=0
1.reg.write.zf=0
2.addr=0x4d92
2.mem.read=0
2.mem.read.data.0x0=7f454c4602010100
2.reg.read=rsi
2.reg.read.rsi=0
2.reg.write=r12
2.reg.write.r12=0x10102464c457f
idx=2
```

As output does not seem like the best to me, but rather an internal dump (indeed it is the SDB dump, more or less...). Converting SDB to Ht just to keep it working almost the same as SDB is not that useful imho. All those convertions with atoi, etc. are wrong by default imho. It means we are just still working with strings.

I propose to just ignore this improvement to ESIL, unless you really need it for something else.



Otherwise, as @XVilka said in a comment we should really re-think how we use the ESIL trace date and what it is used for. Do we need multiple hashtables (or data structures, in general) to store different kind of information? The problem with SDB is that it's all strings and often enough a single instance was used to store different kind of info by just changing a bit the key string.

I do not have a full and clear pictures of ESIL trace, but:
1) instead of having 0.addr, 0.mem.write, 0.reg, etc. shall we just have a vector/array of "something" that group all those info? Let's call it RzEsilTraceInstruction (or something else..)
2) what do we want to capture for each instruction being traced? 1) address of the instr 2) memory written to and corresponding value 3) memory read from and corresponding value 4) reg... etc. etc.

So maybe we want something like:
```
typedef struct {
  ut64 addr;
  bool is_read; // or just use an enum
  ut64 value; // either written to or read from
} RzEsilTraceMemOp;

typedef struct {
  ut64 addr;
  RzVector<RzEsilTraceMemOp> write_ops;
  RzVector<RzEsilTraceMemOp> read_ops;
  RzVector<RzEsilTraceRegOp> write_reg_ops;
  RzVector<RzEsilTraceRegOp> read_reg_ops;
} RzEsilTraceInstruction;

typedef struct {
   ....
   RzVector<RzEsilTraceInstruction> instructions;
} RzAnalysisEsilTrace;
```

Again, the above is really just a very quick example and maybe it does not make 100% sense, but I think this is what we want if we want to swith to something better than plain SDB with everything stored as string in a single SDB instance. Given all the above, shall we really go down this root right now with ESIL?

cc @thestr4ng3r

XVilka:
It's not an improvement for ESIL, it's an improvement for *any* IL emulation trace, which is used in the analysis anyway. So the same code will be used for the new Rizin IL as well.

XVilka:
One more thing, instead of calling these new structures `RzEsilTraceMemOp` please use more generic name like `RzILTraceMemOp` or something like that.

Heersin:
> Converting SDB to Ht just to keep it working almost the same as SDB is not that useful imho
> Otherwise, as @XVilka said in a comment we should really re-think how we use the ESIL trace date and what it is used for. Do we need multiple hashtables (or data structures, in general) to store different kind of information? The problem with SDB is that it's all strings and often enough a single instance was used to store different kind of info by just changing a bit the key string.

Yes. At first I though it's a simple migration from SDB to Ht according to #1294 (Because there is no need to have such a SDB abstraction layer in `trace`). And I don't really have a clear picture about `ESIL trace`, especially about the part it interacts with `rz_debug`. So I just mimic the interfaces provided by SDB.

As far as I know, the info collected by `trace` used in two places : 
1. In `debug`, it traces the version of registers and memory, and used to restore the status to the previous version (step back) if needed.
2. In `type`, `analysis_tp` use the info to propagate types. 

Currently, `trace` relies on multiple callbacks and sdb heavily (and lacks of document to describe some callbacks also), which makes it difficult to locate bug and integrate with new IL.  @ret2libc I agree with your suggestion that `trace` should be redesigned.  the values of `mem.read`/`mem.write`/`reg.read`/`reg.write`... may have different structures in hash table. Splitting one single ht to 4 looks good to me and we can discard the string representation in SDB. 

If we want to redesign it, the most import thing we should clarify is the demands of `trace`. **What information should `trace` provide for analysis ?**

XVilka:
I think the proposed structure of using vectors looks ok to implement.

ret2libc:
As you said, traces are used in debug and type propagation. Type propagation queries for `<idx>.mem.write`, `<idx>.reg.read`, `<idx>.reg.read.<regname>`, `<idx>.reg.write`, and probably few other similar things. I think they are all things that can be easily expressed in data structures like the ones mentioned above.

Now... That seems to be quite a big refactoring and it remains to see whether we want to dedicate a big chunk of time to it or not. @XVilka

XVilka:
@ret2libc It's necessary anyway, retrofitting RZIL in this current ESIL trace code is PITA.

Heersin:
Working on refactoring now.

librz/analysis/il_trace.c
```diff
@@ -0,0 +1,199 @@
+// SPDX-FileCopyrightText: 2021 Heersin <teablearcher@gmail.com>
+// SPDX-License-Identifier: LGPL-3.0-only
+
+/**
+ * \file il_trace.c
+ * \brief new rizin il trace implementation
+ *
+ * provide operations to new il trace structure to record
```
XVilka: In comments it's better to refer to `il` as `IL`, it's an abbreviation.

librz/analysis/esil/esil_trace.c
```diff
@@ -298,132 +341,44 @@ RZ_API void rz_analysis_esil_trace_restore(RzAnalysisEsil *esil, int idx) {
...
-					p("wx %s @ %s\n", str2, addr);
-				} else {
-					eprintf("Invalid entry in reg.read\n");
-				}
-				ptr = next;
-			} while (next);
-		}
+	RzILTraceInstruction *instruction = rz_analysis_esil_get_instruction_trace(esil->trace, idx);
+	if (!instruction) {
+		printf("Invalid trace id : %d\n", idx);
```
XVilka: Please use either `eprintf()` or, better, `RZ_LOG_ERROR()`.

librz/analysis/esil/esil_trace.c
```diff
@@ -175,11 +199,22 @@ static int trace_hook_mem_write(RzAnalysisEsil *esil, ut64 addr, const ut8 *buf,
...
+	ut64 val = strtol(hexbuf, NULL, 16);
+
+	// Trace memory read behavior
+	RzILTraceMemOp *mem_write = RZ_NEW0(RzILTraceMemOp);
+	mem_write->value = val;
+	mem_write->behavior = TRACE_WRITE;
+	mem_write->addr = addr;
+	esil_add_mem_trace(esil->trace, mem_write);
+
+	// clean the hex buffer
```
XVilka: Unnecessary comment, imho.

XVilka (changes requested):

librz/analysis/esil/esil_trace.c
```diff
@@ -74,7 +81,10 @@ RZ_API void rz_analysis_esil_trace_free(RzAnalysisEsilTrace *trace) {
 			rz_reg_arena_free(trace->arena[i]);
 		}
 		free(trace->stack_data);
-		sdb_free(trace->db);
+		if (trace->instructions) {
+			rz_pvector_free(trace->instructions);
+			trace->instructions = NULL;
+		}
```
ret2libc:
```suggestion
		rz_pvector_free(trace->instructions);
		trace->instructions = NULL;
```

librz/include/rz_analysis.h
```diff
@@ -1435,7 +1435,54 @@ RZ_API void rz_analysis_esil_interrupts_fini(RzAnalysisEsil *esil);
 RZ_API void rz_analysis_esil_mem_ro(RzAnalysisEsil *esil, int mem_readonly);
 RZ_API void rz_analysis_esil_stats(RzAnalysisEsil *esil, int enable);
 
-/* trace */
+/* new trace implementation */
+typedef enum {
+	TRACE_READ, ///< read
+	TRACE_WRITE ///< write
+} TRACE_OPERATION;
```
ret2libc:
```suggestion
} RzILTraceOpType;
```

librz/analysis/il_trace.c
```diff
@@ -0,0 +1,208 @@
...
+	// TODO : handle error
+	return instruction_trace;
+}
+
+/**
+ * clean an il trace
+ * \param instruction RzILTraceInstruction, trace to be cleaned
+ */
+RZ_API void rz_analysis_il_trace_instruction_free(RzILTraceInstruction *instruction) {
+	if (instruction->write_reg_ops) {
```
ret2libc: there is no need for all these `if (...) free(...)`. Just free.

librz/analysis/il_trace.c
```diff
@@ -0,0 +1,208 @@
...
+	instruction_trace->write_reg_ops = rz_pvector_new((RzPVectorFree)free_reg_op);
+	// TODO : handle error
+	return instruction_trace;
+}
+
+/**
+ * clean an il trace
+ * \param instruction RzILTraceInstruction, trace to be cleaned
+ */
+RZ_API void rz_analysis_il_trace_instruction_free(RzILTraceInstruction *instruction) {
```
ret2libc, line 54: check if `instruction == NULL` and exit early if so. That is the common behaviour of `free` functions.

librz/analysis/il_trace.c
```diff
@@ -0,0 +1,208 @@
...
+		rz_pvector_free(instruction->write_mem_ops);
+		instruction->write_mem_ops = NULL;
+	}
+
+	if (instruction->read_mem_ops) {
+		rz_pvector_free(instruction->read_mem_ops);
+		instruction->read_mem_ops = NULL;
+	}
+
+	RZ_FREE(instruction);
```
ret2libc:
```suggestion
	free(instruction);
```
RZ_FREE is not wrong, but it is unnecessary..

librz/analysis/il_trace.c
```diff
@@ -0,0 +1,208 @@
...
+}
+
+/**
+ * clean an il trace
+ * \param instruction RzILTraceInstruction, trace to be cleaned
+ */
+RZ_API void rz_analysis_il_trace_instruction_free(RzILTraceInstruction *instruction) {
+	if (instruction->write_reg_ops) {
+		rz_pvector_free(instruction->write_reg_ops);
+		instruction->write_reg_ops = NULL;
```
ret2libc: also all these `instruction->xxx = nULL` are unnecessary.

librz/analysis/il_trace.c
```diff
@@ -0,0 +1,208 @@
...
+
+/**
+ * add memory change to an instruction trace
+ * \param trace RzILTraceInstruction *, trace of instruction which triggers a memory change
+ * \param mem RzILTraceMemOp *, info of memory change
+ */
+RZ_API void rz_analysis_il_trace_add_mem(RzILTraceInstruction *trace, RzILTraceMemOp *mem) {
+	if (!trace || !mem) {
+		return;
+	}
```
ret2libc:
```suggestion
	rz_return_if_fail(trace && mem);
```

librz/include/rz_analysis.h
```diff
@@ -1435,7 +1435,54 @@ RZ_API void rz_analysis_esil_interrupts_fini(RzAnalysisEsil *esil);
 RZ_API void rz_analysis_esil_mem_ro(RzAnalysisEsil *esil, int mem_readonly);
 RZ_API void rz_analysis_esil_stats(RzAnalysisEsil *esil, int enable);
 
-/* trace */
+/* new trace implementation */
+typedef enum {
+	TRACE_READ, ///< read
```
ret2libc:
```suggestion
	RZ_IL_TRACE_OP_READ, ///< read
```

librz/analysis/il_trace.c
```diff
@@ -0,0 +1,208 @@
...
+
+	int is_write = mem->behavior;
```
ret2libc:
```suggestion
	int is_write = mem->behavior == TRACE_WRITE;
```

librz/include/rz_analysis.h
```diff
@@ -1435,7 +1435,54 @@ RZ_API void rz_analysis_esil_interrupts_fini(RzAnalysisEsil *esil);
...
+
+/* Independent Trace Functions */
+RZ_API RzILTraceInstruction *rz_analysis_il_trace_instruction_new(ut64 addr);
+RZ_API void rz_analysis_il_trace_instruction_free(RzILTraceInstruction *instruction);
+RZ_API void rz_analysis_il_trace_add_mem(RzILTraceInstruction *trace, RzILTraceMemOp *mem);
+RZ_API void rz_analysis_il_trace_add_reg(RzILTraceInstruction *trace, RzILTraceRegOp *reg);
+RZ_API RzILTraceMemOp *rz_analysis_il_get_mem_op_trace(RzILTraceInstruction *trace, ut64 addr, bool is_write);
+RZ_API RzILTraceRegOp *rz_analysis_il_get_reg_op_trace(RzILTraceInstruction *trace, const char *regname, bool is_write);
+RZ_API bool rz_analysis_il_mem_trace_contains(RzILTraceInstruction *trace, ut64 addr, bool is_write);
+RZ_API bool rz_analysis_il_reg_trace_contains(RzILTraceInstruction *trace, const char *regname, bool is_write);
```
ret2libc:
```suggestion
RZ_API RzILTraceMemOp *rz_analysis_il_get_mem_op_trace(RzILTraceInstruction *trace, ut64 addr, RzILTraceOpType op_type);
RZ_API RzILTraceRegOp *rz_analysis_il_get_reg_op_trace(RzILTraceInstruction *trace, const char *regname, RzILTraceOpType op_type);
RZ_API bool rz_analysis_il_mem_trace_contains(RzILTraceInstruction *trace, ut64 addr, RzILTraceOpType op_type);
RZ_API bool rz_analysis_il_reg_trace_contains(RzILTraceInstruction *trace, const char *regname, RzILTraceOpType op_type);
```

librz/analysis/il_trace.c
```diff
@@ -0,0 +1,208 @@
...
+**/
+
+#include <rz_util.h>
+#include <rz_analysis.h>
+
+static void free_reg_op(RzILTraceRegOp *reg_op) {
+	if (reg_op) {
+		if (reg_op->reg_name) {
+			free(reg_op->reg_name);
+		}
```
ret2libc, line 23: it is never necessary to do `if (x) free (x)`. If it is necessary, it means the `free` function needs to gracefully handle the case when the arg is NULL.

librz/analysis/il_trace.c
```diff
@@ -0,0 +1,208 @@
...
+
+/**
+ * add register change to an instruction trace
+ * \param trace RzILTraceInstruction *, trace of instruction which triggers a register change
+ * \param mem RzILTraceRegOp *, info of register change
+ */
+RZ_API void rz_analysis_il_trace_add_reg(RzILTraceInstruction *trace, RzILTraceRegOp *reg) {
+	if (!trace || !reg) {
+		return;
+	}
```
ret2libc: use rz_return_if_fail

ret2libc: also in all other places where it makes sense.

librz/analysis/il_trace.c
```diff
@@ -0,0 +1,208 @@
...
+	if (!trace || !mem) {
+		return;
+	}
+
+	int is_write = mem->behavior;
+	if (rz_analysis_il_mem_trace_contains(trace, mem->addr, is_write)) {
+		return;
+	}
+
+	if (is_write) {
```
ret2libc: it might make sense to use something like:
```c
switch (op_type) {
case RZ_IL_TRACE_OP_READ:
  rz_pvector_push(trace->read_mem_ops, mem);
  trace->stats |= TRACE_INS_HAS_MEM_R;
case RZ_IL_TRACE_OP_WRITE:
  rz_pvector_push(trace->write_mem_ops, mem);
  trace->stats |= TRACE_INS_HAS_MEM_W;
default:
  rz_warn_if_reached();
  break;
}
```

This way if you ever add some other values to the op_type enum you will immediately see you have to touch this code.

ret2libc: Same for the other places ofc.

librz/include/rz_analysis.h
```diff
@@ -1435,7 +1435,54 @@ RZ_API void rz_analysis_esil_interrupts_fini(RzAnalysisEsil *esil);
...
+/* new trace implementation */
+typedef enum {
+	TRACE_READ, ///< read
+	TRACE_WRITE ///< write
+} TRACE_OPERATION;
+
+typedef struct {
+	ut64 addr; ///< memory address
+	TRACE_OPERATION behavior; ///< read or write, see TRACE_OPERATION enums
+	ut64 value; ///< data either written to or read from
```
ret2libc, line 1447: it might be wrong to assume ut64. For example, instructions that work on XMM/YMM registers in x86_64 might access 128bits in one shot or more probably. Maybe we should have `ut8 value[32]`.

librz/analysis/esil/esil_trace.c
```diff
@@ -175,11 +199,21 @@ static int trace_hook_mem_write(RzAnalysisEsil *esil, ut64 addr, const ut8 *buf,
...
 	int ret = 0;
 	char *hexbuf = malloc((1 + len) * 3);
-	sdb_array_add_num(DB, KEY("mem.write"), addr, 0);
+
+	// convert data to ut64
+	// FIXME : a better way to convert between them or change the argument
 	rz_hex_bin2str(buf, len, hexbuf);
-	sdb_set(DB, KEYAT("mem.write.data", addr), hexbuf, 0);
-	//eprintf ("[ESIL] MEM WRITE 0x%08"PFMT64x" %s\n", addr, hexbuf);
+	ut64 val = strtol(hexbuf, NULL, 16);
```
ret2libc, line 204: see the comment about the `value` field of RzILTraceMemOp. It should probably be possible to just copy the buf data into RzILTraceMemOp->value.

librz/core/analysis_tp.c
```diff
@@ -801,26 +841,41 @@ RZ_API void rz_core_analysis_type_match(RzCore *core, RzAnalysisFunction *fcn) {
...
 			if (rz_analysis_op_nonlinear(aop->type)) { // skip the instr
 				rz_reg_set_value(core->dbg->reg, r, addr + aop->size);
 			} else {
 				rz_core_esil_step(core, UT64_MAX, NULL, NULL, false);
 			}
-			Sdb *trace = analysis->esil->trace->db;
-			ctx.cur_idx = sdb_num_get(trace, "idx", 0);
+
+			// CHECK_ME : previous idx ?
+			RzPVector *ins_traces = analysis->esil->trace->instructions;
```
ret2libc, line 878:
```suggestion
			RzPVector *ins_traces = etrace->instructions;
```

librz/analysis/esil/esil_trace.c
```diff
@@ -193,6 +227,16 @@ static int trace_hook_mem_write(RzAnalysisEsil *esil, ut64 addr, const ut8 *buf,
 	return ret;
 }
 
+/**
+ * Get instruction trace from esil trace by index
+ * \param etrace RzAnalysisEsilTrace *, esil trace
+ * \param idx int, index of instruction
+ * \return RzILTraceInstruction *, instruction trace at index
+ */
+RZ_API RzILTraceInstruction *rz_analysis_esil_get_instruction_trace(RzAnalysisEsilTrace *etrace, int idx) {
```
ret2libc, line 234: Add checks on `idx`. Is it within the bounds of the vector?

librz/core/analysis_tp.c
```diff
@@ -564,6 +589,12 @@ void propagate_types_among_used_variables(RzCore *core, HtUP *op_cache, Sdb *tra
 	bool userfnc = false;
 	bool prop = false;
 	ut32 type = aop->type & RZ_ANALYSIS_OP_TYPE_MASK;
+
+	RzILTraceInstruction *cur_instr_trace = NULL;
```
ret2libc, line 602: I would just pass `RzEsilTrace` to this function instead of RzPVector and then use `esil_trace_get_instruction(etrace, idx);` to get the right instruction. By doing so all the proper checks are in that function and not here.

ret2libc: It also allow us to easily change the implementation of how traced instructions are stored without having to touch again all this code.

librz/core/analysis_tp.c
```diff
@@ -564,6 +589,12 @@ void propagate_types_among_used_variables(RzCore *core, HtUP *op_cache, Sdb *tra
...
+	if (ctx->cur_idx < rz_pvector_len(instr_traces)) {
+		cur_instr_trace = rz_pvector_at(instr_traces, ctx->cur_idx);
+	}
```
ret2libc, line 605: Can we even continue with the rest of the function if we `cur_instr_trace` is nuLL?

ret2libc (commented):
Much much better IMHO! Thanks a lot for your work.

librz/analysis/esil/esil_trace.c
```diff
@@ -124,21 +134,26 @@ static int trace_hook_reg_read(RzAnalysisEsil *esil, const char *name, ut64 *res
...
 
 static int trace_hook_reg_write(RzAnalysisEsil *esil, const char *name, ut64 *val) {
 	int ret = 0;
-	//eprintf ("[ESIL] REG WRITE %s 0x%08"PFMT64x"\n", name, *val);
-	sdb_array_add(DB, KEY("reg.write"), name, 0);
-	sdb_num_set(DB, KEYREG("reg.write", name), *val, 0);
+
+	// add reg write to trace
+	RzILTraceRegOp *reg_write = RZ_NEW0(RzILTraceRegOp);
+	reg_write->reg_name = strdup(name);
```
p371jah, line 150: IMO it would be better to use RzStrConstPool here

librz/include/rz_analysis.h
```diff
@@ -1435,7 +1435,54 @@ RZ_API void rz_analysis_esil_interrupts_fini(RzAnalysisEsil *esil);
...
+
+typedef struct {
+	ut64 addr; ///< Address of instruction
+	ut32 stats; ///< Has write/read to reg/mem ? see TRACE_INS_HAS_* enums
+
+	RzPVector *write_mem_ops; ///< Vector<RzILTraceRegOp>
+	RzPVector *read_mem_ops; ///< Vector<RzILTraceRegOp>
+
+	RzPVector *write_reg_ops; ///< Vector<RzILTraceRegOp>
+	RzPVector *read_reg_ops; ///< Vector<RzILTraceRegOp>
```
p371jah, line 1471: @ret2libc too much memory allocations, don't you think? `RzILTraceInstruction` -> `RzPVector` -> `RzILTraceRegOp`/`RzILTraceMemOp`

librz/analysis/esil/esil_trace.c
Heersin: It makes sense to find a better way handle const strings. Some const string from the argument are freed so I have to dump them manually. Where can I find the usage of ConstPool ?

librz/core/analysis_tp.c
Heersin: Yeah, there is a similar get function, but I don't use it sometimes.  I will update them to prevent getting element directly from the vector.

librz/analysis/esil/esil_trace.c
Heersin: Yes, or it will report an assertion fail from the vector when it exceed the bounds.

librz/core/analysis_tp.c
Heersin: The old SDB code will return a NULL when there are no corresponding key. The function seems cover that situation. I will check them again and debug.

librz/include/rz_analysis.h
Heersin: About this `value` field, I checked the original code. Though it store the buffer data into SDB, it still read from it convert to ut64 type by using `sdb_get_num`. I think we can use the buffer and keep the logic of returning a `ut64` value when get from mem_op_trace for now, what do you think ?

Heersin:
Close this pr and create a new one to trigger ASAN build

PR #1381, opened 2021-07-29
Use a multi-stage Dockerfile to improve readability

ret2libc:

**Detailed description**

* Add a .dockerignore file to avoid useless re-building
* Remove rz-ghidra from container, at least for now. It does not work
  right now and ideally rz-ghidra should be compiled in the stage 0 and
  moved into /tmp/rizin-install, however rz-pm currently cannot easily
  do this. Things are too messy on the rz-pm side right now, so better
  to review once plugins will be easier to handle.

**Test plan**

...

**Closing issues**

...


Dockerfile
```diff
@@ -72,47 +62,34 @@ RUN apt-get update && \
...
+
+
+FROM debian:10
 ENV RZ_ARM64_AS=${with_arm64_as:+aarch64-linux-gnu-as}
 ENV RZ_ARM32_AS=${with_arm32_as:+arm-linux-gnueabi-as}
 ENV RZ_PPC_AS=${with_ppc_as:+powerpc64le-linux-gnu-as}
 
-# Create non-root user
 RUN useradd -m rizin
+RUN apt-get update && apt-get install -y --no-install-recommends ca-certificates git python3
```
XVilka, line 84: Wouldn't you also need a C/C++ toolchain for `rz-pm` plugins?

Dockerfile
```diff
@@ -72,47 +62,34 @@ RUN apt-get update && \
...
+
+WORKDIR /tmp
+RUN git clone -b "$RZ_PIPE_PY_VERSION" https://github.com/rizinorg/rz-pipe
+RUN pip3 install --root=/tmp/rizin-install ./rz-pipe/python
+
+WORKDIR /tmp/rizin
+RUN meson --prefix=/usr /tmp/build
+RUN meson compile -C /tmp/build
+RUN meson install --destdir /tmp/rizin-install -C /tmp/build
+
```
XVilka, line 77: Unnecessary empty line

XVilka (changes requested):

Dockerfile
ret2libc: Yes but i wouldn't install it by default and leave that to the user or another layer of images based on this. To avoid big images. Otherwise i should also keep cmake, ninja, meson and other things

Dockerfile
ret2libc, line 77:
```suggestion
```

XVilka (approved):

PR #1400, opened 2021-08-04
Allow referring to structs/unions without specifiers

thestr4ng3r:

**Detailed description**

Used for example in rz-ghidra tests: https://github.com/rizinorg/rz-ghidra/blob/40ea3766d1bb283e255e31b9e7c3a61885dd3cb0/test/db/extras/ghidra#L2007

**Test plan**

See unit tests

Or:
Load some struct like `struct Stuff { int a; };`
Assign it to a variable like `afvb var_name "Stuff *"` as opposed to `afvb var_name "struct Stuff *"`

### DIFF
--- a/test/unit/test_type.c
+++ b/test/unit/test_type.c
@@ -499,7 +499,7 @@ static bool test_struct_identifier_without_specifier(void) {
 	int r = rz_type_parse_string(typedb, "struct bla { int a; };", &error_msg);
 	mu_assert_eq(r, 0, "parse struct definition");
 
-	// After definin a struct `struct bla` we also want to be able to refer to
+	// After defining a struct `struct bla` we also want to be able to refer to
 	// it by just `bla` rather than `struct bla`
 
 	RzType *ttype = rz_type_parse_string_single(typedb->parser, "bla *", &error_msg);
@@ -527,7 +527,7 @@ static bool test_union_identifier_without_specifier(void) {
 	int r = rz_type_parse_string(typedb, "union bla { int a; };", &error_msg);
 	mu_assert_eq(r, 0, "parse union definition");
 
-	// After definin a union `union bla` we also want to be able to refer to
+	// After defining a union `union bla` we also want to be able to refer to
 	// it by just `bla` rather than `union bla`
 
 	RzType *ttype = rz_type_parse_string_single(typedb->parser, "bla *", &error_msg);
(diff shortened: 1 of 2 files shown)


ret2libc (commented):
Shall we define a typedef every time we add a struct definition instead?

test/unit/test_type.c
```diff
@@ -488,6 +488,62 @@ static bool test_struct_func_types(void) {
...
+	mu_assert_notnull(typedb, "Couldn't create new RzTypeDB");
+	mu_assert_notnull(typedb->types, "Couldn't create new types hashtable");
+	const char *dir_prefix = rz_sys_prefix(NULL);
+	rz_type_db_init(typedb, dir_prefix, "x86", 64, "linux");
+
+	char *error_msg = NULL;
+	int r = rz_type_parse_string(typedb, "struct bla { int a; };", &error_msg);
+	mu_assert_eq(r, 0, "parse struct definition");
+
+	// After definin a struct `struct bla` we also want to be able to refer to
```
XVilka: A typo

XVilka (approved):
LGTM, though I did it on purpose initially, to denote the difference. But this would work as well.

test/unit/test_type.c
```diff
@@ -488,6 +488,62 @@ static bool test_struct_func_types(void) {
...
+	mu_assert_notnull(typedb, "Couldn't create new RzTypeDB");
+	mu_assert_notnull(typedb->types, "Couldn't create new types hashtable");
+	const char *dir_prefix = rz_sys_prefix(NULL);
+	rz_type_db_init(typedb, dir_prefix, "x86", 64, "linux");
+
+	char *error_msg = NULL;
+	int r = rz_type_parse_string(typedb, "union bla { int a; };", &error_msg);
+	mu_assert_eq(r, 0, "parse union definition");
+
+	// After definin a union `union bla` we also want to be able to refer to
```
XVilka: A typo

wargio (approved):

thestr4ng3r:
> Shall we define a typedef every time we add a struct definition instead?

> LGTM, though I did it on purpose initially, to denote the difference. But this would work as well.

The main point at the moment is that base types are uniquely identified by their name, so a typedef to the same name currently wouldn't work anyway, and that is also the reason why previously you would get a weird error in the mentioned cases when it tries to forward-define the type name but can't because it's already taken.

This concept is a bit different to C and may initially cause issues with some headers that do `typedef struct thing {} thing;` but keeping names unique has a lot of benefits in how the types code can be kept cleaner so I think it's good like this.

Issue #1185, opened 2021-06-02
Uncouple ARM's Thumb mode detection from `asm.bits`


XVilka:
Currently the manual and automatic Thumb detection involves switching the value of `asm.bits` which affects other parts of the analysis, how type length is interpreted, and so on. Ideally, the autodetection and manual switch *should not* setting `asm.bits` and triggering corresponding callbacks.

PR #1403, opened 2021-08-05
Refactor elf arm symbols (librz/bin/format/elf/* and librz/bin/p/bin_elf.inc)

ghost:
**Don't squash** atomic commit where used.

 

**Detailed description**
Arm symbols fix is done before the conversion to RzBinSymbol.

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/librz/bin/format/elf/elf.c
+++ b/librz/bin/format/elf/elf.c
@@ -419,6 +419,7 @@ static bool init(ELFOBJ *bin, RzBinObjectLoadOptions *options) {
 	bin->notes = Elf_(rz_bin_elf_notes_new)(bin);
 
 	bin->symbols = Elf_(rz_bin_elf_symbols_new)(bin);
+	bin->bits = Elf_(rz_bin_elf_get_bits)(bin);
 	bin->imports = Elf_(rz_bin_elf_analyse_imports)(bin);
 
 	return true;
--- a/librz/bin/format/elf/elf.c
+++ b/librz/bin/format/elf/elf.c
@@ -128,6 +128,11 @@ static bool init_phdr_aux(ELFOBJ *bin, RzBinObjectLoadOptions *options) {
 }
 
 static void init_phdr(ELFOBJ *bin, RzBinObjectLoadOptions *options) {
+	if (!bin->ehdr.e_phnum) {
+		RZ_LOG_WARN("There is no program header.\n");
+		return;
+	}
+
 	if (!init_phdr_aux(bin, options)) {
 		RZ_LOG_WARN("Failed to initialize program header.\n");
 	}
--- a/librz/bin/format/elf/elf.c
+++ b/librz/bin/format/elf/elf.c
@@ -168,7 +168,7 @@ static bool init_ehdr_ident(ELFOBJ *bin) {
 		return false;
 	}
 
-	bin->endian = bin->ehdr.e_ident[EI_DATA] == ELFDATA2MSB ? 1 : 0;
+	bin->big_endian = bin->ehdr.e_ident[EI_DATA] == ELFDATA2MSB;
 
 	return true;
 }
--- a/librz/bin/format/elf/elf_arm.c
+++ b/librz/bin/format/elf/elf_arm.c
@@ -1,0 +1,57 @@
+// SPDX-FileCopyrightText: 2021 08A <08A@riseup.net>
+// SPDX-FileCopyrightText: 2021 RizinOrg <info@rizin.re>
+// SPDX-License-Identifier: LGPL-3.0-only
+
+#include "elf.h"
+
+static bool start_a_sequence_of_instruction(RzBinSymbol *symbol) {
+	return strlen(symbol->name) > 3 && rz_str_startswith(symbol->name, "$a.");
+}
+
+static bool start_a_sequence_of_thumb_instruction(RzBinSymbol *symbol) {
+	return strlen(symbol->name) > 3 && rz_str_startswith(symbol->name, "$t.");
+}
+
+static bool start_a_sequence_of_data(RzBinSymbol *symbol) {
+	return strlen(symbol->name) > 3 && rz_str_startswith(symbol->name, "$d.");
+}
+
+bool Elf_(rz_bin_elf_is_arm_binary_supporting_thumb)(RZ_NONNULL ELFOBJ *bin) {
+	rz_return_val_if_fail(bin, false);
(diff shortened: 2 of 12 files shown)


librz/bin/format/elf/elf_arm.c
```diff
@@ -0,0 +1,80 @@
...
+
+	*bits = Elf_(rz_bin_elf_get_bits)(bin);
+
+	if (*bits == 64) {
+		return;
+	}
+
+	*bits = 32;
+
+	if (*paddr == UT64_MAX) {
```
wargio: Tbh, i do not like that you are testing specifically for this value. you should instead testing if is outside the text section.

librz/bin/format/elf/elf_arm.c
```diff
@@ -0,0 +1,80 @@
...
+void Elf_(rz_bin_elf_fix_arm_object)(RZ_NONNULL ELFOBJ *bin, RZ_NONNULL ut64 *paddr, RZ_NONNULL ut64 *vaddr, RZ_NONNULL int *bits) {
+	rz_return_if_fail(bin && paddr && vaddr && bits);
+
+	*bits = Elf_(rz_bin_elf_get_bits)(bin);
+
+	if (*bits == 64) {
+		return;
+	}
+
+	*bits = 32;
```
XVilka, line 36: I wonder if it makes sense to add a link here to https://github.com/rizinorg/rizin/issues/1185

XVilka (changes requested):

librz/bin/format/elf/elf_arm.c
ghost: The functions used the "thumb" keyword, so it will be easy to track in the codebase.

I agree about the problem with arm thumb. Indeed, the output of `is` is in contradiction with `readelf --symbols` when there is a thumb symbol. I believe that the thumb refactoring should be done after the refactoring of address representation is Rizin (@ret2libc).

librz/bin/format/elf/elf_arm.c
ghost: I don't agree with the idea of checking if the value is in the text section. Indeed, this check should have been done before the call of the function. The only thing this line should check is that the address is valid. And right now the only way to do this in Rizin is to check the max value.

XVilka (approved):

librz/bin/format/elf/elf_arm.c
wargio: hmm yes. probably in a section, but i do understand your point.

wargio (approved):

test/db/analysis/msp430
```diff
@@ -1,7 +1,7 @@
 NAME=aFr on ELF
 FILE=bins/elf/msp430.elf
 CMDS=<<EOF
-s main
+s loc.main
```
ret2libc: Why this?

ret2libc (changes requested):

test/db/analysis/msp430
ghost: The old behavior was buggy. Indeed, to detect the entry, the function was returning the offset without checking if it was a vaddr or a paddr. So the `s main` was jumping on a vaddr not a paddr (for this binary). I removed the `is_vaddr` so no entry is detected (msp430 is a “broken”).

ret2libc: This is strange. I'd still expect to have `main` and not `loc.main`, because there is even a `main` symbol in the binary.

ghost: Okay, I will fix this. And in the next PR, I will try to remove the “hack”.

ghost:
I have updated the PR.

`bin->bits` is loaded 2 times because symbols are used to deduct the binary bits (thumb symbols).
This is an egg-and-chicken problem, but the first `bin->bits` value doesn't have too much impact (`bin->bits` < 16 is ignored).

ghost:
There is an alternative to this ugly reload. I can add a function `Elf_(rz_bin_elf_update_symbols_bits)`. That will be use after the first bin->bits.

```
	bin->symbols = Elf_(rz_bin_elf_symbols_new)(bin);
	bin->bits = Elf_(rz_bin_elf_get_bits)(bin);
    Elf_(rz_bin_elf_update_symbols_bits)(bin, bin->symbols);
	bin->imports = Elf_(rz_bin_elf_analyse_imports)(bin);
```

PS: This problem could be solved with #1185.

ret2libc:
> I have updated the PR.
> 
> `bin->bits` is loaded 2 times because symbols are used to deduct the binary bits (thumb symbols).
> This is an egg-and-chicken problem, but the first `bin->bits` value doesn't have too much impact (`bin->bits` < 16 is ignored).

It seems to be computed even 3 times... Is it right? Why do you need bin->bits for creating the ELF symbols?

ghost:
> > I have updated the PR.
> > `bin->bits` is loaded 2 times because symbols are used to deduct the binary bits (thumb symbols).
> > This is an egg-and-chicken problem, but the first `bin->bits` value doesn't have too much impact (`bin->bits` < 16 is ignored).
> 
> It seems to be computed even 3 times... Is it right? Why do you need bin->bits for creating the ELF symbols?

There was a typo (3 -> 2). `RzBinElfSymbol` need `bin->bits` to fix arm thumb address. I moved the thumb addresses process in librz/bin/format/elf/*

ghost:
I refactor where symbols are “fixed” (thumb).

XVilka (approved):

librz/bin/format/elf/elf.c
```diff
@@ -163,7 +168,7 @@ static bool init_ehdr_ident(ELFOBJ *bin) {
 		return false;
 	}
 
-	bin->endian = bin->ehdr.e_ident[EI_DATA] == ELFDATA2MSB ? 1 : 0;
+	bin->big_endian = bin->ehdr.e_ident[EI_DATA] == ELFDATA2MSB ? true : false;
```
ret2libc:
```suggestion
	bin->big_endian = bin->ehdr.e_ident[EI_DATA] == ELFDATA2MSB;
```

thestr4ng3r (approved):

ret2libc (changes requested):
Just to prevent merging. There are some manual tests that may deserve a better look before merging.

librz/bin/format/elf/elf.c
ghost: Shame on me.

ret2libc (approved):

PR #1405, opened 2021-08-05
Convert `av` command to newshell

Basstorm:

**Detailed description**

Convert the `av` command group from `oldshell` to `newshell` directly. This is for the next global variables concept. I will put the related commands to the `av` group like `avg` which means 'analysis-variable-global' since it does not belong to the function.

**Test plan**

CI is green

**Closing issues**

...

### DIFF
--- a/librz/analysis/rtti.c
+++ b/librz/analysis/rtti.c
@@ -14,8 +14,8 @@ RZ_API char *rz_analysis_rtti_demangle_class_name(RzAnalysis *analysis, const ch
 	return rz_analysis_rtti_itanium_demangle_class_name(&context, name);
 }
 
-RZ_API void rz_analysis_rtti_print_at_vtable(RzAnalysis *analysis, ut64 addr, int mode) {
-	bool use_json = mode == 'j';
+RZ_API void rz_analysis_rtti_print_at_vtable(RzAnalysis *analysis, ut64 addr, RzOutputMode mode) {
+	bool use_json = mode == RZ_OUTPUT_MODE_JSON;
 	if (use_json) {
 		rz_cons_print("[");
 	}
@@ -33,11 +33,11 @@ RZ_API void rz_analysis_rtti_print_at_vtable(RzAnalysis *analysis, ut64 addr, in
 	}
 }
 
-RZ_API void rz_analysis_rtti_print_all(RzAnalysis *analysis, int mode) {
+RZ_API void rz_analysis_rtti_print_all(RzAnalysis *analysis, RzOutputMode mode) {
 	RVTableContext context;
 	rz_analysis_vtable_begin(analysis, &context);
 
-	bool use_json = mode == 'j';
+	bool use_json = mode == RZ_OUTPUT_MODE_JSON;
 	if (use_json) {
 		rz_cons_print("[");
 	}
--- a/librz/analysis/rtti_itanium.c
+++ b/librz/analysis/rtti_itanium.c
@@ -6,6 +6,7 @@
 #include <rz_core.h>
 #include <rz_flag.h>
 #include <rz_cons.h>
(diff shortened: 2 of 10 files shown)


librz/analysis/rtti.c
```diff
@@ -14,30 +14,30 @@ RZ_API char *rz_analysis_rtti_demangle_class_name(RzAnalysis *analysis, const ch
 	return rz_analysis_rtti_itanium_demangle_class_name(&context, name);
 }
 
-RZ_API void rz_analysis_rtti_print_at_vtable(RzAnalysis *analysis, ut64 addr, int mode) {
-	bool use_json = mode == 'j';
+RZ_API void rz_analysis_rtti_print_at_vtable(RzAnalysis *analysis, ut64 addr, RzCmdStateOutput *state) {
+	bool use_json = state->mode == RZ_OUTPUT_MODE_JSON;
 	if (use_json) {
 		rz_cons_print("[");
 	}
```
ret2libc:
```suggestion
	rz_cmd_state_output_array_start(state);
```

librz/analysis/rtti.c
```diff
@@ -14,30 +14,30 @@ RZ_API char *rz_analysis_rtti_demangle_class_name(RzAnalysis *analysis, const ch
...
-		rz_analysis_rtti_msvc_print_at_vtable(&context, addr, mode, false);
+		rz_analysis_rtti_msvc_print_at_vtable(&context, addr, state, false);
 	} else {
-		rz_analysis_rtti_itanium_print_at_vtable(&context, addr, mode);
+		rz_analysis_rtti_itanium_print_at_vtable(&context, addr, state);
 	}
 
 	if (use_json) {
 		rz_cons_print("]\n");
 	}
```
ret2libc:
```suggestion
	rz_cmd_state_output_array_end(state);
```

librz/analysis/rtti.c
```diff
@@ -14,30 +14,30 @@ RZ_API char *rz_analysis_rtti_demangle_class_name(RzAnalysis *analysis, const ch
...
-RZ_API void rz_analysis_rtti_print_all(RzAnalysis *analysis, int mode) {
+RZ_API void rz_analysis_rtti_print_all(RzAnalysis *analysis, RzCmdStateOutput *state) {
 	RVTableContext context;
 	rz_analysis_vtable_begin(analysis, &context);
 
-	bool use_json = mode == 'j';
+	bool use_json = state->mode == 'j';
 	if (use_json) {
 		rz_cons_print("[");
 	}
```
ret2libc:
```suggestion
	rz_cmd_state_output_array_start(state);
```

ret2libc: Similarly, use _end instead of printing `]` at the end of the function.

Basstorm: For this PR, I don't want to change the other codes that are not related to the `newshell`...

Basstorm: Because it's just a converting.

XVilka: It is related to the newshell. It helps us to avoid passing `PJ *pj` everywhere. And it's a small change, see this self-contained example: https://github.com/rizinorg/rizin/blob/dev/librz/core/chash.c

Basstorm: I have changed the command type to `RZ_CMD_DESC_TYPE_ARGV_MODES`, and it works well. So there is no `RzCmdStateOutput` to use. Also, if use `RzCmdStateOutput` directly like the code I wrote before, it will break the test because the current print function is printing out by order, but the `RzCmdStateOutput` is to print out at the end, thus I need to change the logic of the current print function.

ret2libc (approved):

Issue #333, opened 2021-01-07
Rizin should have the concept of Global Variables


ITAYC0HEN:
**Is your feature request related to a problem? Please describe.**
Currently, Rizin does not have the concept of Global Variables, this means that one can't set an address as a global variable.
Not only that, but Rizin also doesn't detect global variables and label them automatically upon analysis. Thus, you can see addresses referenced in the disassembly, while they are clearly globals.

The current workaround that the user can do is to assign a flag in an address of a global variable. In addition, users can also apply structs to these addresses.

Worth to mention, this is highly important for decompilers that rely on global variables and their types.

**Describe the solution you'd like**
1. Let users the option to define a global variable and its type
2. Automatically detect and define global variables and their type upon analysis

ret2libc:
> 1. Let users the option to define a global variable and its type

One one side it would be cool to have a command like `afvg` to define `global variables`, on the other side `afv` means `analysis function variables` and globals are not really part of any function.

karliss:
Considering potential approaches does rizin need a new separate mechanism for it or just use the information which is already there and maybe add some commands for more easily managing it. Is there any additional information beside name, offset and type that needs to be stored? There is already a bit of mess due to having multiple mechanisms for naming functions. Having two different mechanisms for defining type at address would also be confusing.

Parts that could use global variable type information:
* Automatic assignment from debug symbols
* type analysis
    * propagating types obtained from accessing global variable fields
    * defining new global variables with correct type when a global variable has pointer field pointing to other structure
    * variable type deduction from function argument types
* recognizing field access in disassembly
* type synchronization with decompiler

XVilka:
Was fixed and supported in both Rizin and Cutter.

PR #1407, opened 2021-08-06
New global variable concept

Basstorm:

**Detailed description**

```
Usage: avg[jadnt]   # Global variables
| avg[j] [<var_name>]             # show global variables
| avga <var_name> <addr> <type>   # add global variable manually
| avgd <addr>                     # delete the global variable at the addr
| avgn <old_var_name> <new_var_name> # rename the global variable
| avgt <var_name> <type>          # change the global variable type
```

**Test plan**

New cmd and unit tests.

**Closing issues**

Partially addresses https://github.com/rizinorg/rizin/issues/333

### DIFF
--- a/librz/analysis/analysis.c
+++ b/librz/analysis/analysis.c
@@ -143,6 +143,7 @@ RZ_API RzAnalysis *rz_analysis_new(void) {
 			rz_analysis_add(analysis, analysis_static_plugins[i]);
 		}
 	}
+	analysis->ht_global_var = ht_pp_new0();
 	return analysis;
 }
 
@@ -194,6 +195,8 @@ RZ_API RzAnalysis *rz_analysis_free(RzAnalysis *a) {
 	free(a->last_disasm_reg);
 	rz_list_free(a->imports);
 	rz_str_constpool_fini(&a->constpool);
+	rz_analysis_var_global_delete_all(a);
+	ht_pp_free(a->ht_global_var);
 	free(a);
 	return NULL;
 }
--- a/librz/analysis/analysis.c
+++ b/librz/analysis/analysis.c
@@ -78,6 +78,11 @@ static void rz_meta_item_free(void *_item) {
 	}
 }
 
+static void global_kv_free(HtPPKv *kv) {
+	free(kv->key);
+	rz_analysis_var_global_free(kv->value);
+}
+
 RZ_API RzAnalysis *rz_analysis_new(void) {
 	int i;
 	RzAnalysis *analysis = RZ_NEW0(RzAnalysis);
@@ -143,7 +148,7 @@ RZ_API RzAnalysis *rz_analysis_new(void) {
 			rz_analysis_add(analysis, analysis_static_plugins[i]);
 		}
 	}
-	analysis->ht_global_var = ht_pp_new0();
+	analysis->ht_global_var = ht_pp_new(NULL, global_kv_free, NULL);
 	analysis->global_var_tree = NULL;
 	return analysis;
 }
--- a/librz/analysis/analysis.c
+++ b/librz/analysis/analysis.c
@@ -195,7 +195,6 @@ RZ_API RzAnalysis *rz_analysis_free(RzAnalysis *a) {
 	free(a->last_disasm_reg);
 	rz_list_free(a->imports);
 	rz_str_constpool_fini(&a->constpool);
-	rz_analysis_var_global_delete_all(a);
 	ht_pp_free(a->ht_global_var);
 	free(a);
 	return NULL;
--- a/librz/analysis/analysis.c
+++ b/librz/analysis/analysis.c
@@ -144,6 +144,7 @@ RZ_API RzAnalysis *rz_analysis_new(void) {
 		}
 	}
 	analysis->ht_global_var = ht_pp_new0();
+	analysis->global_var_tree = NULL;
 	return analysis;
 }
 
--- a/librz/analysis/serialize_analysis.c
+++ b/librz/analysis/serialize_analysis.c
@@ -2,9 +2,11 @@
 // SPDX-License-Identifier: LGPL-3.0-only
 
 #include <rz_util/rz_serialize.h>
+#include <rz_util/rz_num.h>
 #include <rz_vector.h>
 #include <rz_type.h>
 #include <rz_analysis.h>
+#include <rz_core.h>
 
 #include <errno.h>
 
@@ -866,33 +868,28 @@ beach:
 
 RZ_API void rz_serialize_analysis_global_var_save(RZ_NONNULL Sdb *db, RZ_NONNULL RzAnalysis *anal) {
 	rz_return_if_fail(db && anal);
-	RzList *l = rz_analysis_var_global_get_all(anal);
-	if (!l) {
-		return;
-	}
-	RzListIter *it;
-	RzAnalysisVarGlobal *var;
-	RzStrBuf key;
-	char *vartype;
+
 	PJ *j = pj_new();
 	if (!j) {
 		return;
 	}
-	rz_list_foreach (l, it, var) {
-		rz_strbuf_init(&key);
-		rz_strbuf_setf(&key, "0x%" PFMT64x, var->addr);
-		pj_o(j);
+	RBIter it;
+	RzAnalysisVarGlobal *var;
+	char *vartype;
+	rz_rbtree_foreach (anal->global_var_tree, it, var, RzAnalysisVarGlobal, rb) {
 		vartype = rz_type_as_string(anal->typedb, var->type);
 		if (!vartype) {
 			eprintf("Global variable \"%s\" has undefined type\n", var->name);
 			pj_free(j);
 			return;
 		}
+		char addr[32];
+		rz_strf(addr, "0x%" PFMT64x, var->addr);
 		pj_o(j);
 		pj_ks(j, "name", var->name);
-		pj_kn(j, "addr", var->addr);
+		pj_ks(j, "addr", addr);
 		pj_ks(j, "type", vartype);
-		pj_kn(j, "size", var->size);
+		//We do not need to serialize size here because it can be got from type
 		if (!rz_vector_empty(&var->constraints)) {
 			pj_ka(j, "constrs");
 			RzTypeConstraint *constr;
@@ -903,18 +900,17 @@ RZ_API void rz_serialize_analysis_global_var_save(RZ_NONNULL Sdb *db, RZ_NONNULL
 			pj_end(j);
 		}
 		pj_end(j);
-		sdb_set(db, rz_strbuf_get(&key), pj_string(j), 0);
+
+		sdb_set(db, addr, pj_string(j), 0);
 		pj_reset(j);
 	}
 	pj_free(j);
-	rz_list_free(l);
 }
 
 enum {
 	GLOBAL_VAR_FIELD_NAME,
 	GLOBAL_VAR_FIELD_ADDR,
 	GLOBAL_VAR_FIELD_TYPE,
-	GLOBAL_VAR_FIELD_SIZE,
 	GLOBAL_VAR_FIELD_CONSTRS
 };
 
@@ -926,7 +922,6 @@ RZ_API RzSerializeAnalGlobalVarParser rz_serialize_analysis_global_var_parser_ne
 	rz_key_parser_add(parser, "name", GLOBAL_VAR_FIELD_NAME);
 	rz_key_parser_add(parser, "addr", GLOBAL_VAR_FIELD_ADDR);
 	rz_key_parser_add(parser, "type", GLOBAL_VAR_FIELD_TYPE);
-	rz_key_parser_add(parser, "size", GLOBAL_VAR_FIELD_SIZE);
 	rz_key_parser_add(parser, "constrs", GLOBAL_VAR_FIELD_CONSTRS);
 	return parser;
 }
@@ -955,8 +950,8 @@ static bool global_var_load_cb(void *user, const char *k, const char *v) {
 
 	const char *name = NULL;
 	const char *type = NULL;
+	const char *addr_s = NULL;
 	ut64 addr = 0;
-	ut64 size = 0;
 	RzVector constraints;
 	rz_vector_init(&constraints, sizeof(RzTypeConstraint), NULL, NULL);
 
@@ -970,24 +965,16 @@ static bool global_var_load_cb(void *user, const char *k, const char *v) {
 			name = child->str_value;
 			break;
 		case GLOBAL_VAR_FIELD_ADDR:
-			if (child->type != RZ_JSON_INTEGER) {
-				eprintf("addr nop\n");
+			if (child->type != RZ_JSON_STRING) {
 				break;
 			}
-			addr = child->num.s_value;
+			addr_s = child->str_value;
 			break;
 		case GLOBAL_VAR_FIELD_TYPE:
 			if (child->type != RZ_JSON_STRING) {
 				break;
 			}
 			type = child->str_value;
-			break;
-		case GLOBAL_VAR_FIELD_SIZE:
-			if (child->type != RZ_JSON_INTEGER) {
-				eprintf("size nop\n");
-				break;
-			}
-			size = child->num.s_value;
 			break;
 		case VAR_FIELD_CONSTRS: {
 			if (child->type != RZ_JSON_ARRAY) {
@@ -1031,12 +1018,14 @@ static bool global_var_load_cb(void *user, const char *k, const char *v) {
 		eprintf("Fail to parse the function variable (\"%s\") type: %s\n", name, type);
 		goto beach;
 	}
+	RzCore *core = ctx->analysis->core;
+	addr = rz_num_math(core->num, addr_s);
 	glob = rz_analysis_var_global_new(name, addr);
 	if (!glob) {
 		goto beach;
 	}
 	rz_analysis_var_global_set_type(glob, vartype);
-	glob->size = size;
+	glob->size = rz_type_db_get_bitsize(ctx->analysis->typedb, vartype) / 8;
 
 	RzTypeConstraint *constr;
 	rz_vector_foreach(&constraints, constr) {
--- a/librz/analysis/serialize_analysis.c
+++ b/librz/analysis/serialize_analysis.c
@@ -889,7 +889,6 @@ RZ_API void rz_serialize_analysis_global_var_save(RZ_NONNULL Sdb *db, RZ_NONNULL
 		pj_ks(j, "name", var->name);
 		pj_ks(j, "addr", addr);
 		pj_ks(j, "type", vartype);
-		//We do not need to serialize size here because it can be got from type
 		if (!rz_vector_empty(&var->constraints)) {
 			pj_ka(j, "constrs");
 			RzTypeConstraint *constr;
@@ -1025,7 +1024,6 @@ static bool global_var_load_cb(void *user, const char *k, const char *v) {
 		goto beach;
 	}
 	rz_analysis_var_global_set_type(glob, vartype);
-	glob->size = rz_type_db_get_bitsize(ctx->analysis->typedb, vartype) / 8;
 
 	RzTypeConstraint *constr;
 	rz_vector_foreach(&constraints, constr) {
--- a/librz/analysis/serialize_analysis.c
+++ b/librz/analysis/serialize_analysis.c
@@ -864,6 +864,212 @@ beach:
 	return ret;
 }
 
+RZ_API void rz_serialize_analysis_global_var_save(RZ_NONNULL Sdb *db, RZ_NONNULL RzAnalysis *anal) {
+	rz_return_if_fail(db && anal);
+	RzList *l = rz_analysis_var_global_get_all(anal);
+	if (!l) {
+		return;
+	}
+	RzListIter *it;
+	RzAnalysisVarGlobal *var;
+	RzStrBuf key;
+	char *vartype;
+	PJ *j = pj_new();
+	if (!j) {
+		return;
+	}
+	rz_list_foreach (l, it, var) {
+		rz_strbuf_init(&key);
+		rz_strbuf_setf(&key, "0x%" PFMT64x, var->addr);
+		pj_o(j);
+		vartype = rz_type_as_string(anal->typedb, var->type);
+		if (!vartype) {
+			eprintf("Global variable \"%s\" has undefined type\n", var->name);
+			pj_free(j);
+			return;
+		}
+		pj_o(j);
+		pj_ks(j, "name", var->name);
+		pj_kn(j, "addr", var->addr);
+		pj_ks(j, "type", vartype);
+		pj_kn(j, "size", var->size);
+		if (!rz_vector_empty(&var->constraints)) {
+			pj_ka(j, "constrs");
+			RzTypeConstraint *constr;
+			rz_vector_foreach(&var->constraints, constr) {
+				pj_i(j, (int)constr->cond);
+				pj_n(j, constr->val);
+			}
+			pj_end(j);
+		}
+		pj_end(j);
+		sdb_set(db, rz_strbuf_get(&key), pj_string(j), 0);
+		pj_reset(j);
+	}
+	pj_free(j);
+	rz_list_free(l);
+}
+
+enum {
+	GLOBAL_VAR_FIELD_NAME,
+	GLOBAL_VAR_FIELD_ADDR,
+	GLOBAL_VAR_FIELD_TYPE,
+	GLOBAL_VAR_FIELD_SIZE,
+	GLOBAL_VAR_FIELD_CONSTRS
+};
+
+RZ_API RzSerializeAnalGlobalVarParser rz_serialize_analysis_global_var_parser_new(void) {
+	RzSerializeAnalDiffParser parser = rz_key_parser_new();
+	if (!parser) {
+		return NULL;
+	}
+	rz_key_parser_add(parser, "name", GLOBAL_VAR_FIELD_NAME);
+	rz_key_parser_add(parser, "addr", GLOBAL_VAR_FIELD_ADDR);
+	rz_key_parser_add(parser, "type", GLOBAL_VAR_FIELD_TYPE);
+	rz_key_parser_add(parser, "size", GLOBAL_VAR_FIELD_SIZE);
+	rz_key_parser_add(parser, "constrs", GLOBAL_VAR_FIELD_CONSTRS);
+	return parser;
+}
+
+RZ_API void rz_serialize_analysis_global_var_parser_free(RzSerializeAnalGlobalVarParser parser) {
+	rz_key_parser_free(parser);
+}
+
+typedef struct {
+	RzAnalysis *analysis;
+	RzKeyParser *parser;
+} GlobalVarCtx;
+
+static bool global_var_load_cb(void *user, const char *k, const char *v) {
+	GlobalVarCtx *ctx = user;
+
+	char *json_str = strdup(v);
+	if (!json_str) {
+		return true;
+	}
+	RzJson *json = rz_json_parse(json_str);
+	if (!json || json->type != RZ_JSON_OBJECT) {
+		free(json_str);
+		return false;
+	}
+
+	const char *name = NULL;
+	const char *type = NULL;
+	ut64 addr = 0;
+	ut64 size = 0;
+	RzVector constraints;
+	rz_vector_init(&constraints, sizeof(RzTypeConstraint), NULL, NULL);
+
+	RzAnalysisVarGlobal *glob = NULL;
+
+	RZ_KEY_PARSER_JSON(ctx->parser, json, child, {
+		case GLOBAL_VAR_FIELD_NAME:
+			if (child->type != RZ_JSON_STRING) {
+				break;
+			}
+			name = child->str_value;
+			break;
+		case GLOBAL_VAR_FIELD_ADDR:
+			if (child->type != RZ_JSON_INTEGER) {
+				eprintf("addr nop\n");
+				break;
+			}
+			addr = child->num.s_value;
+			break;
+		case GLOBAL_VAR_FIELD_TYPE:
+			if (child->type != RZ_JSON_STRING) {
+				break;
+			}
+			type = child->str_value;
+			break;
+		case GLOBAL_VAR_FIELD_SIZE:
+			if (child->type != RZ_JSON_INTEGER) {
+				eprintf("size nop\n");
+				break;
+			}
+			size = child->num.s_value;
+			break;
+		case VAR_FIELD_CONSTRS: {
+			if (child->type != RZ_JSON_ARRAY) {
+				break;
+			}
+			RzJson *baby;
+			for (baby = child->children.first; baby; baby = baby->next) {
+				if (baby->type != RZ_JSON_INTEGER) {
+					break;
+				}
+				RzJson *sibling = baby->next;
+				if (!sibling || sibling->type != RZ_JSON_INTEGER) {
+					break;
+				}
+				RzTypeConstraint constr;
+				constr.cond = (RzTypeCond)baby->num.s_value;
+				constr.val = sibling->num.u_value;
+				if (constr.cond < RZ_TYPE_COND_AL || constr.cond > RZ_TYPE_COND_LS) {
+					baby = sibling;
+					continue;
+				}
+				rz_vector_push(&constraints, &constr);
+				baby = sibling;
+			}
+			break;
+		}
+		default:
+			break;
+	})
+
+	if (!name || !type) {
+		goto beach;
+	}
+	char *error_msg = NULL;
+	RzTypeParser *parser = rz_type_parser_new();
+	if (!parser) {
+		goto beach;
+	}
+	RzType *vartype = rz_type_parse_string_single(parser, type, &error_msg);
+	if (error_msg) {
+		eprintf("Fail to parse the function variable (\"%s\") type: %s\n", name, type);
+		goto beach;
+	}
+	glob = rz_analysis_var_global_new(name, addr);
+	if (!glob) {
+		goto beach;
+	}
+	rz_analysis_var_global_set_type(glob, vartype);
+	glob->size = size;
+
+	RzTypeConstraint *constr;
+	rz_vector_foreach(&constraints, constr) {
+		rz_analysis_var_global_add_constraint(glob, constr);
+	}
+	rz_type_parser_free(parser);
+	return rz_analysis_var_global_add(ctx->analysis, glob);
+
+beach:
+	rz_vector_fini(&constraints);
+	return false;
+}
+
+RZ_API bool rz_serialize_analysis_global_var_load(RZ_NONNULL Sdb *db, RZ_NONNULL RzAnalysis *analysis, RZ_NULLABLE RzSerializeResultInfo *res) {
+	GlobalVarCtx ctx = {
+		.analysis = analysis,
+		.parser = rz_serialize_analysis_global_var_parser_new(),
+	};
+	bool ret;
+	if (!ctx.parser) {
+		RZ_SERIALIZE_ERR(res, "parser init failed");
+		ret = false;
+		goto beach;
+	}
+	ret = sdb_foreach(db, global_var_load_cb, &ctx);
+	if (!ret) {
+		RZ_SERIALIZE_ERR(res, "functions parsing failed");
+	}
+beach:
+	rz_key_parser_free(ctx.parser);
+	return ret;
+}
+
 static bool store_label_cb(void *j, const ut64 k, const void *v) {
 	pj_kn(j, v, k);
 	return true;
@@ -2049,6 +2255,7 @@ RZ_API void rz_serialize_analysis_save(RZ_NONNULL Sdb *db, RZ_NONNULL RzAnalysis
 	rz_serialize_analysis_imports_save(sdb_ns(db, "imports", true), analysis);
 	rz_serialize_analysis_pin_save(sdb_ns(db, "pins", true), analysis);
 	rz_serialize_analysis_cc_save(sdb_ns(db, "cc", true), analysis);
+	rz_serialize_analysis_global_var_save(sdb_ns(db, "vars", true), analysis);
 }
 
 RZ_API bool rz_serialize_analysis_load(RZ_NONNULL Sdb *db, RZ_NONNULL RzAnalysis *analysis, RZ_NULLABLE RzSerializeResultInfo *res) {
@@ -2095,6 +2302,7 @@ RZ_API bool rz_serialize_analysis_load(RZ_NONNULL Sdb *db, RZ_NONNULL RzAnalysis
 	SUB("imports", rz_serialize_analysis_imports_load(subdb, analysis, res));
 	SUB("pins", rz_serialize_analysis_pin_load(subdb, analysis, res));
 	SUB("cc", rz_serialize_analysis_cc_load(subdb, analysis, res));
+	SUB("vars", rz_serialize_analysis_global_var_load(subdb, analysis, res));
 
 	ret = true;
 beach:
--- a/librz/analysis/var_global.c
+++ b/librz/analysis/var_global.c
@@ -12,7 +12,7 @@
  * \param comment variable comment
  * \return RzAnalysisVarGlobal *
  */
-RZ_API RZ_OWN RzAnalysisVarGlobal *rz_analysis_var_global_new(RZ_NONNULL const char *name, ut64 addr, const char *comment) {
+RZ_API RZ_OWN RzAnalysisVarGlobal *rz_analysis_var_global_new(RZ_NONNULL const char *name, ut64 addr, RZ_NULLABLE const char *comment) {
 	rz_return_val_if_fail(name && addr, NULL);
 	RzAnalysisVarGlobal *glob = RZ_NEW0(RzAnalysisVarGlobal);
 	if (!glob) {
@@ -36,7 +36,7 @@ RZ_API RZ_OWN RzAnalysisVarGlobal *rz_analysis_var_global_new(RZ_NONNULL const c
 RZ_API RZ_OWN bool rz_analysis_var_global_add(RzAnalysis *analysis, RzAnalysisVarGlobal *global_var) {
 	rz_return_val_if_fail(analysis && global_var, false);
 	if (rz_analysis_var_global_get_byaddr(analysis, global_var->addr)) {
-		eprintf("Global variable at 0x%" PFMT64x " is already exist!\n", global_var->addr);
+		RZ_LOG_ERROR("Global variable at 0x%" PFMT64x " is already exist!\n", global_var->addr);
 		return false;
 	}
 	return ht_pp_insert(analysis->ht_global_var, global_var->name, global_var);
@@ -72,8 +72,8 @@ RZ_API bool rz_analysis_var_global_delete_byname(RzAnalysis *analysis, const cha
 	rz_return_val_if_fail(analysis && name, false);
 	RzAnalysisVarGlobal *glob = rz_analysis_var_global_get_byname(analysis, name);
 	if (!glob) {
-		eprintf("No such global variable!\n");
-		return true;
+		RZ_LOG_ERROR("No such global variable!\n");
+		return false;
 	}
 	rz_analysis_var_global_free(glob);
 	return ht_pp_delete(analysis->ht_global_var, name);
@@ -90,12 +90,14 @@ RZ_API bool rz_analysis_var_global_delete_byaddr(RzAnalysis *analysis, ut64 addr
 	rz_return_val_if_fail(analysis, false);
 	RzAnalysisVarGlobal *glob = rz_analysis_var_global_get_byaddr(analysis, addr);
 	if (!glob) {
-		eprintf("No such global variable!\n");
-		return true;
+		RZ_LOG_ERROR("No such global variable!\n");
+		return false;
 	}
 	bool deleted = ht_pp_delete(analysis->ht_global_var, glob->name);
-	rz_analysis_var_global_free(glob);
-	return deleted ? true : false;
+	if (deleted) {
+		rz_analysis_var_global_free(glob);
+	}
+	return deleted;
 }
 
 /**
@@ -107,10 +109,7 @@ RZ_API bool rz_analysis_var_global_delete_byaddr(RzAnalysis *analysis, ut64 addr
  */
 RZ_API RZ_BORROW RzAnalysisVarGlobal *rz_analysis_var_global_get_byname(RzAnalysis *analysis, const char *name) {
 	rz_return_val_if_fail(analysis && name, NULL);
-	RzAnalysisVarGlobal *glob;
-	bool found;
-	glob = ht_pp_find(analysis->ht_global_var, name, &found);
-	return found ? glob : NULL;
+	return (RzAnalysisVarGlobal *)ht_pp_find(analysis->ht_global_var, name, NULL);
 }
 
 struct list_addr {
@@ -137,6 +136,9 @@ static bool global_var_collect_addr_cb(void *user, const void *k, const void *v)
 RZ_API RZ_BORROW RzAnalysisVarGlobal *rz_analysis_var_global_get_byaddr(RzAnalysis *analysis, ut64 addr) {
 	rz_return_val_if_fail(analysis, NULL);
 	RzList *list = rz_list_new();
+	if (!list) {
+		return NULL;
+	}
 	struct list_addr l = { list, addr };
 	ht_pp_foreach(analysis->ht_global_var, global_var_collect_addr_cb, &l);
 	if (rz_list_length(list) != 1) {
@@ -164,6 +166,9 @@ static bool global_var_collect_cb(void *user, const void *k, const void *v) {
 RZ_API RZ_OWN RzList *rz_analysis_var_global_get_all(RzAnalysis *analysis) {
 	rz_return_val_if_fail(analysis, NULL);
 	RzList *globals = rz_list_new();
+	if (!globals) {
+		return NULL;
+	}
 	ht_pp_foreach(analysis->ht_global_var, global_var_collect_cb, globals);
 	return globals;
 }
@@ -180,7 +185,7 @@ RZ_API bool rz_analysis_var_global_rename(RzAnalysis *analysis, const char *old_
 	rz_return_val_if_fail(analysis && old_name && newname, false);
 	RzAnalysisVarGlobal *glob = rz_analysis_var_global_get_byname(analysis, old_name);
 	if (!glob) {
-		eprintf("No such global variable!\n");
+		RZ_LOG_ERROR("No such global variable!\n");
 		return false;
 	}
 	RZ_FREE(glob->name);
@@ -200,6 +205,7 @@ RZ_API bool rz_analysis_var_global_set_comment(RzAnalysis *analysis, const char 
 	rz_return_val_if_fail(analysis && name && comment, false);
 	RzAnalysisVarGlobal *glob = rz_analysis_var_global_get_byname(analysis, name);
 	if (!glob) {
+		RZ_LOG_ERROR("No such global variable!\n");
 		return false;
 	}
 	RZ_FREE(glob->comment);
@@ -219,8 +225,8 @@ RZ_API void rz_analysis_var_global_set_type(RzAnalysisVarGlobal *glob, RzType *t
 	glob->type = type;
 }
 
-static st64 var_access_cmp(ut64 x, char *y) {
-	return x - (ut64)((RzAnalysisVarGlobal *)y)->addr;
+static st64 var_access_cmp(st64 x, char *y) {
+	return x - (st64)((RzAnalysisVarGlobal *)y)->addr;
 }
 
 /**
@@ -247,6 +253,9 @@ RZ_API void rz_analysis_var_global_set_access(RzAnalysis *analysis, RzAnalysisVa
 	}
 	if (!acc || acc->offset != offset) {
 		acc = rz_vector_insert(&glob->accesses, index, NULL);
+		if (!acc) {
+			return;
+		}
 		acc->offset = offset;
 		acc->type = 0;
 	}
@@ -272,7 +281,7 @@ RZ_API void rz_analysis_var_global_remove_access_at(RzAnalysisVarGlobal *glob, u
 		return;
 	}
 	RzAnalysisVarAccess *acc = rz_vector_index_ptr(&glob->accesses, index);
-	if (acc->offset == offset) {
+	if (acc && acc->offset == offset) {
 		rz_vector_remove_at(&glob->accesses, index, NULL);
 	}
 }
@@ -366,6 +375,9 @@ RZ_API void rz_analysis_var_global_list_show(RzAnalysis *analysis, RzCmdStateOut
 	RzAnalysisVarGlobal *glob = NULL;
 	if (name) {
 		global_vars = rz_list_new();
+		if (!global_vars) {
+			return;
+		}
 		glob = rz_analysis_var_global_get_byname(analysis, name);
 		if (!glob) {
 			return;
@@ -380,6 +392,7 @@ RZ_API void rz_analysis_var_global_list_show(RzAnalysis *analysis, RzCmdStateOut
 	char *comment = NULL;
 	bool json = state->mode == RZ_OUTPUT_MODE_JSON;
 	PJ *pj = json ? state->d.pj : NULL;
+	// to use rz_cmd_state_output_array_start we need to set RzCore as the dependency of RzAnalysis, which is impossible
 	if (json) {
 		pj_a(pj);
 	}
@@ -396,7 +409,7 @@ RZ_API void rz_analysis_var_global_list_show(RzAnalysis *analysis, RzCmdStateOut
 		}
 		switch (state->mode) {
 		case RZ_OUTPUT_MODE_STANDARD:
-			analysis->cb_printf("global %s %s @ 0x%" PFMT64x " %s\n",
+			analysis->cb_printf("global %s %s @ 0x%" PFMT64x " ;%s\n",
 				var_type, glob->name,
 				glob->addr, comment);
 			break;
--- a/librz/analysis/var_global.c
+++ b/librz/analysis/var_global.c
@@ -184,7 +184,7 @@ RZ_API RZ_BORROW RzAnalysisVarGlobal *rz_analysis_var_global_get_byaddr_in(RzAna
 		return NULL;
 	}
 	ut64 size = rz_type_db_get_bitsize(analysis->typedb, var->type) / 8;
-	if (addr > (var->addr + size - 1) || addr < var->addr) {
+	if (addr >= var->addr + size) {
 		return NULL;
 	}
 	return var;
--- a/librz/analysis/var_global.c
+++ b/librz/analysis/var_global.c
@@ -50,9 +50,6 @@ RZ_API RZ_OWN bool rz_analysis_var_global_add(RzAnalysis *analysis, RZ_NONNULL R
 	if (!ht_pp_insert(analysis->ht_global_var, global_var->name, global_var)) {
 		return false;
 	}
-	if (global_var->type) {
-		global_var->size = rz_type_db_get_bitsize(analysis->typedb, global_var->type) / 8;
-	}
 	if (!rz_rbtree_aug_insert(&analysis->global_var_tree, &global_var->addr, &global_var->rb, global_var_node_cmp, NULL, NULL)) {
 		return false;
 	}
@@ -186,7 +183,8 @@ RZ_API RZ_BORROW RzAnalysisVarGlobal *rz_analysis_var_global_get_byaddr_in(RzAna
 	if (!var) {
 		return NULL;
 	}
-	if (addr > (var->addr + var->size - 1) || addr < var->addr) {
+	ut64 size = rz_type_db_get_bitsize(analysis->typedb, var->type) / 8;
+	if (addr > (var->addr + size - 1) || addr < var->addr) {
 		return NULL;
 	}
 	return var;
--- a/librz/analysis/var_global.c
+++ b/librz/analysis/var_global.c
@@ -26,6 +26,23 @@ RZ_API RZ_OWN RzAnalysisVarGlobal *rz_analysis_var_global_new(RZ_NONNULL const c
 	return glob;
 }
 
+typedef struct {
+	RBNode rb;
+	ut64 addr;
+	ut64 size;
+} GlobalVarNode;
+
+int global_var_node_cmp(const void *incoming, const RBNode *in_tree, void *user) {
+	ut64 ia = *(ut64 *)incoming;
+	ut64 ta = container_of(in_tree, const GlobalVarNode, rb)->addr;
+	if (ia < ta) {
+		return -1;
+	} else if (ia > ta) {
+		return 1;
+	}
+	return 0;
+}
+
 /**
  * \brief Add the global variable into hashtable
  * 
@@ -36,10 +53,25 @@ RZ_API RZ_OWN RzAnalysisVarGlobal *rz_analysis_var_global_new(RZ_NONNULL const c
 RZ_API RZ_OWN bool rz_analysis_var_global_add(RzAnalysis *analysis, RzAnalysisVarGlobal *global_var) {
 	rz_return_val_if_fail(analysis && global_var, false);
 	if (rz_analysis_var_global_get_byaddr(analysis, global_var->addr)) {
-		RZ_LOG_ERROR("Global variable at 0x%" PFMT64x " is already exist!\n", global_var->addr);
+		RZ_LOG_ERROR("Global variable %s at 0x%" PFMT64x " already exists!\n", global_var->name, global_var->addr);
+		return false;
+	} else if (ht_pp_find(analysis->ht_global_var, global_var->name, NULL)) {
+		RZ_LOG_ERROR("Global variable %s already exists!\n", global_var->name);
 		return false;
 	}
-	return ht_pp_insert(analysis->ht_global_var, global_var->name, global_var);
+	if (!ht_pp_insert(analysis->ht_global_var, global_var->name, global_var)) {
+		return false;
+	}
+	GlobalVarNode *node = RZ_NEW0(GlobalVarNode);
+	node->addr = global_var->addr;
+	node->size = 0;
+	if (global_var->type) {
+		node->size = rz_type_db_get_bitsize(analysis->typedb, global_var->type) / 8;
+	}
+	if (!rz_rbtree_aug_insert(&analysis->global_var_tree, &node->addr, &node->rb, global_var_node_cmp, NULL, NULL)) {
+		return false;
+	}
+	return true;
 }
 
 /**
@@ -76,6 +108,7 @@ RZ_API bool rz_analysis_var_global_delete_byname(RzAnalysis *analysis, const cha
 		return false;
 	}
 	rz_analysis_var_global_free(glob);
+	rz_rbtree_delete(&analysis->global_var_tree, &glob->addr, global_var_node_cmp, NULL, NULL, NULL);
 	return ht_pp_delete(analysis->ht_global_var, name);
 }
 
@@ -95,6 +128,7 @@ RZ_API bool rz_analysis_var_global_delete_byaddr(RzAnalysis *analysis, ut64 addr
 	}
 	bool deleted = ht_pp_delete(analysis->ht_global_var, glob->name);
 	if (deleted) {
+		rz_rbtree_delete(&analysis->global_var_tree, &glob->addr, global_var_node_cmp, NULL, NULL, NULL);
 		rz_analysis_var_global_free(glob);
 	}
 	return deleted;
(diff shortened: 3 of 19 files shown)


librz/analysis/var_global.c
```diff
@@ -0,0 +1,311 @@
+// SPDX-FileCopyrightText: 2021 Basstorm <basstorm@nyist.edu.cn>
+// SPDX-License-Identifier: LGPL-3.0-only
+
+#include <rz_analysis.h>
+#include <rz_list.h>
+
+RZ_API RzAnalysisVarGlobal *rz_analysis_var_global_new(char *name, ut64 addr, char *comment) {
```
XVilka: Please don't forget to denote ownership with `RZ_OWN`/`RZ_BORROW`.

librz/analysis/var_global.c
```diff
@@ -0,0 +1,311 @@
...
+	}
+	glob->name = strdup(name);
+	glob->addr = addr;
+	if (comment) {
+		glob->comment = strdup(comment);
+	}
+	return glob;
+}
+
+RZ_API bool rz_analysis_var_global_add(RzAnalysis *analysis, RzAnalysisVarGlobal *global_var) {
```
XVilka: Same here

librz/analysis/var_global.c
```diff
@@ -0,0 +1,311 @@
...
+	rz_return_val_if_fail(analysis && global_var, false);
+	if (rz_analysis_var_global_get_byaddr(analysis, global_var->addr)) {
+		eprintf("Global variable at 0x%" PFMT64x " is already exist!\n", global_var->addr);
+		return false;
+	}
+	return ht_pp_insert(analysis->ht_global_var, global_var->name, global_var);
+}
+
+RZ_API void rz_analysis_var_global_free(RzAnalysisVarGlobal *glob) {
+	rz_return_if_fail(glob);
```
XVilka: Usually `*_free()` functions do not error/warning on NULL pointer, just exit.

librz/analysis/var_global.c
```diff
@@ -0,0 +1,311 @@
...
+		return true;
+	}
+	bool deleted = ht_pp_delete(analysis->ht_global_var, glob->name);
+	rz_analysis_var_global_free(glob);
+	return deleted ? true : false;
+}
+
+RZ_API void rz_analysis_var_global_delete_all(RzAnalysis *analysis) {
+	rz_return_if_fail(analysis);
+	RzList *globals = rz_analysis_var_global_get_all(analysis);
```
XVilka: Probably better to just call `ht_pp_free()` here? Should be way faster

librz/analysis/var_global.c
```diff
@@ -0,0 +1,311 @@
...
+}
+
+static bool global_var_collect_cb(void *user, const void *k, const void *v) {
+	RzList *l = user;
+	RzAnalysisVarGlobal *glob = (RzAnalysisVarGlobal *)v;
+	rz_list_append(l, glob);
+	return true;
+}
+
+RZ_API RzList *rz_analysis_var_global_get_all(RzAnalysis *analysis) {
```
XVilka: Also don't forget about ownership, especially in case of `RzList`.

XVilka (changes requested):

XVilka:
@karliss also suggested to automatically create globals based on the information from symbols, DWARF or PDB. It might be a good target for the next PR.

Basstorm:
> @karliss also suggested to automatically create globals based on the information from symbols, DWARF or PDB. It might be a good target for the next PR.

Yeah, It's in the plan.

librz/analysis/var_global.c
ghost: and the RZ_NONNULL

ghost (changes requested):

ghost:
Is there any way to add test for these modifications?

Basstorm:
> Is there any way to add test for these modifications?

Yes, but I haven't done it.

librz/core/cmd_analysis.c
```diff
@@ -8088,6 +8088,111 @@ RZ_IPI RzCmdStatus rz_analysis_rtti_demangle_class_name_handler(RzCore *core, in
...
+
+RZ_IPI RzCmdStatus rz_analysis_global_variable_add_handler(RzCore *core, int argc, const char **argv) {
+	if (argc < 4) {
+		return RZ_CMD_STATUS_ERROR;
+	}
+	char *var_name = (char *)argv[1];
+	ut64 addr = rz_num_math(core->num, (char *)argv[2]);
+	char *type = (char *)argv[3];
+	char *comment = NULL;
+	if (argc == 5) {
```
XVilka: What happens if there are 6 arguments? It's probably add a check for `argc > 5` at the beginning too.

Basstorm: Yes, I agree

librz/analysis/var_global.c
```diff
@@ -0,0 +1,423 @@
...
+RZ_API RZ_OWN RzAnalysisVarGlobal *rz_analysis_var_global_new(RZ_NONNULL char *name, ut64 addr, char *comment) {
+	rz_return_val_if_fail(name && addr, NULL);
+	RzAnalysisVarGlobal *glob = RZ_NEW0(RzAnalysisVarGlobal);
+	if (!glob) {
+		return NULL;
+	}
+	glob->name = strdup(name);
+	glob->addr = addr;
+	if (comment) {
+		glob->comment = strdup(comment);
```
XVilka: `strdup()` on NULL will return an error. I recommend to either declare it `RZ_NONNULL` or just handle this case with `?:` operator. Also, if you use `strdup()` I recommend to define the argument as `const char*` instead.

librz/analysis/var_global.c
```diff
@@ -0,0 +1,423 @@
...
+ * \param glob Global variable instance
+ * \return void
+ */
+RZ_API void rz_analysis_var_global_free(RzAnalysisVarGlobal *glob) {
+	if (!glob) {
+		return;
+	}
+
+	RZ_FREE(glob->name);
+	if (glob->comment) {
```
XVilka: It should work just fine with `comment` being NULL, not a big deal.

librz/analysis/var_global.c
```diff
@@ -0,0 +1,423 @@
...
+}
+
+/**
+ * \brief Get the instance of global variable by its name
+ * 
+ * \param analysis RzAnalysis
+ * \param name Global variable name
+ * \return RzAnalysisVarGlobal *
+ */
+RZ_API RZ_BORROW RzAnalysisVarGlobal *rz_analysis_var_global_get_byname(RzAnalysis *analysis, char *name) {
```
XVilka: same here, `const char *`

librz/analysis/var_global.c
```diff
@@ -0,0 +1,423 @@
...
+}
+
+/**
+ * \brief Delete and free the global variable by its name
+ * 
+ * \param analysis RzAnalysis
+ * \param name Global Variable name
+ * \return true if success
+ */
+RZ_API bool rz_analysis_var_global_delete_byname(RzAnalysis *analysis, char *name) {
```
XVilka: `const char*`

librz/analysis/var_global.c
```diff
@@ -0,0 +1,423 @@
...
+
+/**
+ * \brief Rename the global variable
+ * 
+ * \param analysis RzAnalysis
+ * \param old_name The old name of the global variable
+ * \param newname The new name of the global variable
+ * \return true if succeed
+ */
+RZ_API bool rz_analysis_var_global_rename(RzAnalysis *analysis, char *old_name, char *newname) {
```
XVilka: `const char*`

librz/analysis/var_global.c
```diff
@@ -0,0 +1,423 @@
...
+
+/**
+ * \brief Set the comment of the global variable
+ * 
+ * \param analysis RzAnalysis
+ * \param name The name of the global variable to set
+ * \param comment The comment to set
+ * \return true if succeed
+ */
+RZ_API bool rz_analysis_var_global_set_comment(RzAnalysis *analysis, char *name, char *comment) {
```
XVilka: `const char*`

librz/analysis/var_global.c
```diff
@@ -0,0 +1,423 @@
...
+	rz_return_val_if_fail(analysis && name && comment, false);
+	RzAnalysisVarGlobal *glob = rz_analysis_var_global_get_byname(analysis, name);
+	if (!glob) {
+		return false;
+	}
+	if (glob->comment) {
```
XVilka: same here, just free it.

librz/analysis/var_global.c
```diff
@@ -0,0 +1,423 @@
...
+
+/**
+ * \brief Print out the global variables
+ * 
+ * \param analysis RzAnalysis
+ * \param state RzCmdStateOutput *
+ * \param name Which one to print
+ * \return void
+ */
+RZ_API void rz_analysis_var_global_list_show(RzAnalysis *analysis, RzCmdStateOutput *state, char *name) {
```
XVilka: `const char*`. Also mark it as `RZ_NULLABLE` then

XVilka (changes requested):

librz/analysis/var_global.c
XVilka: Judging by the function logic it should be `RZ_NULLABLE const char *comment`

librz/include/rz_analysis.h
```diff
@@ -737,6 +738,15 @@ typedef struct rz_analysis_var_t {
 	int argnum;
 } RzAnalysisVar;
 
+typedef struct rz_analysis_var_global_t {
+	char *name; // name of the variable
+	ut64 addr; // address of the global variable
+	RzType *type; // type of the variable
+	RzVector /*<RzAnalysisVarAccess>*/ accesses; // ordered by offset, touch this only through API or expect uaf
```
thestr4ng3r: Is this actually needed? Normally accesses to a global variable would just be represented by xrefs. Special care must be taken when xrefs do not go exactly to the variable's address but at another one that still intersects that variable's type, but that can be solved too.

librz/include/rz_analysis.h
```diff
@@ -623,6 +623,7 @@ typedef struct rz_analysis_t {
 	RzList *leaddrs;
 	RzArchTarget *arch_target;
 	RzArchPlatformTarget *platform_target;
+	HtPP *ht_global_var; // global variables
```
thestr4ng3r, line 626: In addition to the name-keyed hashtable, we will need a (non-augmented) RBtree of vars keyed by their address since variables have a certain size. Then querying for a variable that contains a certain address becomes possible, which will be quite important in some cases.

librz/include/rz_analysis.h
```diff
@@ -737,6 +738,15 @@ typedef struct rz_analysis_var_t {
 	int argnum;
 } RzAnalysisVar;
 
+typedef struct rz_analysis_var_global_t {
+	char *name; // name of the variable
+	ut64 addr; // address of the global variable
+	RzType *type; // type of the variable
+	RzVector /*<RzAnalysisVarAccess>*/ accesses; // ordered by offset, touch this only through API or expect uaf
+	char *comment;
```
thestr4ng3r: Is having comments here really something that makes sense? What is the advantage over putting a regular comment at the same address?

XVilka: @thestr4ng3r it will be shown everywhere the variable accessed. for example.

thestr4ng3r: This could be implemented on top of regular comments too.

librz/include/rz_analysis.h
Basstorm: Yes, this is a great point

librz/include/rz_analysis.h
```diff
@@ -1573,6 +1583,26 @@ RZ_API char *rz_analysis_fcn_format_sig(RZ_NONNULL RzAnalysis *analysis, RZ_NONN
...
+// Global vars
+RZ_API RzAnalysisVarGlobal *rz_analysis_var_global_new(char *name, ut64 addr, char *comment);
+RZ_API bool rz_analysis_var_global_add(RzAnalysis *analysis, RzAnalysisVarGlobal *global_var);
+RZ_API void rz_analysis_var_global_free(RzAnalysisVarGlobal *glob);
+RZ_API bool rz_analysis_var_global_delete_byname(RzAnalysis *analysis, char *name);
+RZ_API bool rz_analysis_var_global_delete_byaddr(RzAnalysis *analysis, ut64 addr);
+RZ_API void rz_analysis_var_global_delete_all(RzAnalysis *analysis);
+RZ_API RzAnalysisVarGlobal *rz_analysis_var_global_get_byname(RzAnalysis *analysis, char *name);
+RZ_API RzAnalysisVarGlobal *rz_analysis_var_global_get_byaddr(RzAnalysis *analysis, ut64 addr);
+RZ_API RzList *rz_analysis_var_global_get_all(RzAnalysis *analysis);
```
MalhotraPulak: Add RZ_BORROW/RZ_OWN to the APIs?

librz/include/rz_analysis.h
XVilka: Please note, that size might change depending on the type set, thus during the "set new type" operation or type propagation we would need to update the RBTree as well.

wargio:
regarding `avga` and `avgd` i think we should set the command as `avga <var_name> <type> [<comment>] @ <address|flag>` and `avgd @ <address|flag>`

librz/analysis/var_global.c
wargio: i agree with you XVilka

librz/analysis/var_global.c
```diff
@@ -0,0 +1,419 @@
...
+ * 
+ * \param analysis RzAnalysis
+ * \param name Global Variable name
+ * \return true if succeed
+ */
+RZ_API bool rz_analysis_var_global_delete_byname(RzAnalysis *analysis, const char *name) {
+	rz_return_val_if_fail(analysis && name, false);
+	RzAnalysisVarGlobal *glob = rz_analysis_var_global_get_byname(analysis, name);
+	if (!glob) {
+		eprintf("No such global variable!\n");
```
wargio: use `RZ_LOG_ERROR` instead of `eprintf`

librz/analysis/var_global.c
```diff
@@ -0,0 +1,419 @@
...
+		return true;
```
wargio: return false, since we failed.

librz/analysis/var_global.c
```diff
@@ -0,0 +1,419 @@
...
+ * 
+ * \param analysis RzAnalysis
+ * \param addr Global Variable address
+ * \return true if succeed
+ */
+RZ_API bool rz_analysis_var_global_delete_byaddr(RzAnalysis *analysis, ut64 addr) {
+	rz_return_val_if_fail(analysis, false);
+	RzAnalysisVarGlobal *glob = rz_analysis_var_global_get_byaddr(analysis, addr);
+	if (!glob) {
+		eprintf("No such global variable!\n");
```
wargio: use `RZ_LOG_ERROR` instead of `eprintf`

librz/analysis/var_global.c
```diff
@@ -0,0 +1,419 @@
...
+		return true;
```
wargio: return false.

librz/analysis/var_global.c
```diff
@@ -0,0 +1,419 @@
...
+	}
+	bool deleted = ht_pp_delete(analysis->ht_global_var, glob->name);
+	rz_analysis_var_global_free(glob);
+	return deleted ? true : false;
```
wargio: useless compare. `return deleted` directly since deleted is already a boolean

librz/analysis/var_global.c
```diff
@@ -0,0 +1,419 @@
...
+ * \param analysis RzAnalysis
+ * \param name Global variable name
+ * \return RzAnalysisVarGlobal *
+ */
+RZ_API RZ_BORROW RzAnalysisVarGlobal *rz_analysis_var_global_get_byname(RzAnalysis *analysis, const char *name) {
+	rz_return_val_if_fail(analysis && name, NULL);
+	RzAnalysisVarGlobal *glob;
+	bool found;
+	glob = ht_pp_find(analysis->ht_global_var, name, &found);
+	return found ? glob : NULL;
```
wargio:
```suggestion
	return (RzAnalysisVarGlobal *)ht_pp_find(analysis->ht_global_var, name, NULL);
```
No need to check for that boolean, since it will return a NULL when fail, and even if you find a value and is NULL, then the returned pointer is still NULL.

librz/analysis/var_global.c
```diff
@@ -0,0 +1,419 @@
...
+/**
+ * \brief Same as rz_analysis_var_global_get_byname but by its address
+ * 
+ * \param analysis RzAnalysis
+ * \param addr Global variable address
+ * \return RzAnalysisVarGlobal *
+ */
+RZ_API RZ_BORROW RzAnalysisVarGlobal *rz_analysis_var_global_get_byaddr(RzAnalysis *analysis, ut64 addr) {
+	rz_return_val_if_fail(analysis, NULL);
+	RzList *list = rz_list_new();
```
wargio: check if list is NULL and fail if it is.

librz/analysis/var_global.c
```diff
@@ -0,0 +1,419 @@
...
+
+/**
+ * \brief Get all of the added global variables
+ * 
+ * \param analysis RzAnalysis
+ * \return RzList *
+ */
+RZ_API RZ_OWN RzList *rz_analysis_var_global_get_all(RzAnalysis *analysis) {
+	rz_return_val_if_fail(analysis, NULL);
+	RzList *globals = rz_list_new();
```
wargio, line 208: check for allocation failure (i.e. if NULL).

librz/analysis/var_global.c
```diff
@@ -0,0 +1,419 @@
...
+ * \param analysis RzAnalysis
+ * \param old_name The old name of the global variable
+ * \param newname The new name of the global variable
+ * \return true if succeed
+ */
+RZ_API bool rz_analysis_var_global_rename(RzAnalysis *analysis, const char *old_name, RZ_NONNULL const char *newname) {
+	rz_return_val_if_fail(analysis && old_name && newname, false);
+	RzAnalysisVarGlobal *glob = rz_analysis_var_global_get_byname(analysis, old_name);
+	if (!glob) {
+		eprintf("No such global variable!\n");
```
wargio: use `RZ_LOG_ERROR` instead of `eprintf`

librz/analysis/var_global.c
```diff
@@ -0,0 +1,419 @@
...
+ * \return void
+ */
+RZ_API void rz_analysis_var_global_set_type(RzAnalysisVarGlobal *glob, RzType *type) {
+	rz_return_if_fail(glob && type);
+	glob->type = type;
+}
+
+static st64 var_access_cmp(ut64 x, char *y) {
+	return x - (ut64)((RzAnalysisVarGlobal *)y)->addr;
+}
```
wargio: you are doing unsigned 64bit subtraction and returning a signed 64bit value. are you sure about this?

librz/analysis/var_global.c
```diff
@@ -0,0 +1,419 @@
...
+	}
+	if (!acc || acc->offset != offset) {
+		acc = rz_vector_insert(&glob->accesses, index, NULL);
+		acc->offset = offset;
+		acc->type = 0;
+	}
+
+	acc->type |= (ut8)access_type;
+	acc->stackptr = stackptr;
+	acc->reg = rz_str_constpool_get(&analysis->constpool, reg);
```
wargio: this code contains a NULL dereference bug.
you should add an if to just check if `acc` is NULL, and if it is return.

librz/analysis/var_global.c
```diff
@@ -0,0 +1,419 @@
...
+RZ_API void rz_analysis_var_global_remove_access_at(RzAnalysisVarGlobal *glob, ut64 address) {
+	rz_return_if_fail(glob);
+	st64 offset = (st64)address - (st64)glob->addr;
+	size_t index;
+	rz_vector_lower_bound(&glob->accesses, offset, index, var_access_cmp);
+	if (index >= glob->accesses.len) {
+		return;
+	}
+	RzAnalysisVarAccess *acc = rz_vector_index_ptr(&glob->accesses, index);
+	if (acc->offset == offset) {
```
wargio: can `acc` be NULL ? if yes, apply the suggestion.
```suggestion
	if (acc && acc->offset == offset) {
```

librz/analysis/var_global.c
```diff
@@ -0,0 +1,419 @@
...
+			high = true;
+			break;
+		case RZ_TYPE_COND_GT:
+			rz_strbuf_appendf(&sb, "> 0x%" PFMT64x "", constr->val);
+			high = true;
+			break;
+		default:
+			break;
+		}
+		if (low && high && i != n - 1) {
```
wargio, line 303: at the first loop `low = false` and `high = false` and changes based on the cases, but then what happens if `default` is hit after the 2nd loop?
Is this the correct behavior in case of bad data? or this cannot happen?

librz/analysis/var_global.c
```diff
@@ -0,0 +1,419 @@
...
+ * \param state RzCmdStateOutput *
+ * \param name Which one to print
+ * \return void
+ */
+RZ_API void rz_analysis_var_global_list_show(RzAnalysis *analysis, RzCmdStateOutput *state, RZ_NULLABLE const char *name) {
+	rz_return_if_fail(analysis && state);
+	RzList *global_vars = NULL;
+	RzAnalysisVarGlobal *glob = NULL;
+	if (name) {
+		global_vars = rz_list_new();
```
wargio: check allocation failure (i.e. if is NULL)

librz/analysis/var_global.c
```diff
@@ -0,0 +1,419 @@
...
+	}
+
+	RzListIter *it = NULL;
+	char *var_type = NULL;
+	char *comment = NULL;
+	bool json = state->mode == RZ_OUTPUT_MODE_JSON;
+	PJ *pj = json ? state->d.pj : NULL;
+	if (json) {
+		pj_a(pj);
+	}
```
wargio: @ret2libc i remember there was a function for this, right?

librz/core/cmd_analysis.c
```diff
@@ -8088,6 +8088,111 @@ RZ_IPI RzCmdStatus rz_analysis_rtti_demangle_class_name_handler(RzCore *core, in
...
+RZ_IPI RzCmdStatus rz_analysis_global_variable_delete_handler(RzCore *core, int argc, const char **argv) {
+	if (argc == 2) {
+		ut64 addr = rz_num_math(core->num, (char *)argv[1]);
+		if (rz_analysis_var_global_delete_byaddr(core->analysis, addr)) {
+			return RZ_CMD_STATUS_OK;
+		} else {
+			return RZ_CMD_STATUS_ERROR;
+		}
+	}
+	return RZ_CMD_STATUS_WRONG_ARGS;
```
wargio:
```suggestion
	if (argc != 2) {
		return RZ_CMD_STATUS_WRONG_ARGS;
	}
	ut64 addr = rz_num_math(core->num, (char *)argv[1]);
	if (!rz_analysis_var_global_delete_byaddr(core->analysis, addr)) {
		return RZ_CMD_STATUS_ERROR;
	}
	return RZ_CMD_STATUS_OK;
```

librz/core/cmd_analysis.c
```diff
@@ -8088,6 +8088,111 @@ RZ_IPI RzCmdStatus rz_analysis_rtti_demangle_class_name_handler(RzCore *core, in
...
+	if (argc == 3) {
+		char *oldname = (char *)argv[1];
+		char *newname = (char *)argv[2];
+		if (rz_analysis_var_global_rename(core->analysis, oldname, newname)) {
+			return RZ_CMD_STATUS_OK;
+		} else {
+			return RZ_CMD_STATUS_ERROR;
+		}
+	}
+	return RZ_CMD_STATUS_WRONG_ARGS;
```
wargio:
```suggestion
	if (argc != 3) {
		return RZ_CMD_STATUS_WRONG_ARGS;
	}
	char *oldname = (char *)argv[1];
	char *newname = (char *)argv[2];
	if (!rz_analysis_var_global_rename(core->analysis, oldname, newname)) {
		return RZ_CMD_STATUS_ERROR;
	}
	return RZ_CMD_STATUS_OK;
```

librz/core/cmd_analysis.c
```diff
@@ -8088,6 +8088,111 @@ RZ_IPI RzCmdStatus rz_analysis_rtti_demangle_class_name_handler(RzCore *core, in
...
+		RzAnalysisVarGlobal *glob = rz_analysis_var_global_get_byname(core->analysis, name);
+		if (!glob) {
+			eprintf("No such global variable!\n");
+			return RZ_CMD_STATUS_ERROR;
+		}
+		RzTypeParser *parser = rz_type_parser_new();
+		char *errmsg = NULL;
+		RzType *typ = rz_type_parse_string_single(parser, type, &errmsg);
+		if (errmsg) {
+			eprintf("%s : Error parsing type: \"%s\" message:\n%s\n", __FUNCTION__, type, errmsg);
```
wargio: RZ_LOG_ERR here

librz/core/cmd_analysis.c
```diff
@@ -8088,6 +8088,111 @@ RZ_IPI RzCmdStatus rz_analysis_rtti_demangle_class_name_handler(RzCore *core, in
...
+			return RZ_CMD_STATUS_OK;
+		} else {
+			return RZ_CMD_STATUS_ERROR;
+		}
+	}
+	return RZ_CMD_STATUS_WRONG_ARGS;
+}
+
+RZ_IPI RzCmdStatus rz_analysis_global_variable_retype_handler(RzCore *core, int argc, const char **argv) {
+	if (argc == 3) {
```
wargio: simplify as i did above.

librz/core/cmd_analysis.c
```diff
@@ -8088,6 +8088,111 @@ RZ_IPI RzCmdStatus rz_analysis_rtti_demangle_class_name_handler(RzCore *core, in
...
+		}
+		rz_type_free(glob->type);
+		rz_analysis_var_global_set_type(glob, typ);
+		return RZ_CMD_STATUS_OK;
+	}
+	return RZ_CMD_STATUS_WRONG_ARGS;
+}
+
+RZ_IPI RzCmdStatus rz_analysis_global_variable_recomment_handler(RzCore *core, int argc, const char **argv) {
+	if (argc == 3) {
```
wargio: same here.

librz/core/cmd_analysis.c
```diff
@@ -8088,6 +8088,111 @@ RZ_IPI RzCmdStatus rz_analysis_rtti_demangle_class_name_handler(RzCore *core, in
...
+		char *name = (char *)argv[1];
+		char *comment = (char *)argv[2];
+		RzAnalysisVarGlobal *glob = rz_analysis_var_global_get_byname(core->analysis, name);
+		if (!glob) {
+			eprintf("No such global variable!\n");
```
wargio: RZ_LOG_ERR

wargio (changes requested):

librz/analysis/var_global.c
Basstorm: yeah, it's a stupid mistake.

librz/analysis/var_global.c
Basstorm: This function is copy-paste from `var.c`, I did not check it .. 😢

librz/include/rz_analysis.h
thestr4ng3r: Actually we don't have to update it. The RBTree (when not augmented, which is fine here) just saves the origin address. Global variables must not overlap. But even if they do we can handle it gracefully and emit warnings.
Then to get the variable that contains a given address, we find the highest variable whose address is less than or equal to the given one. This is the only one that could potentially contain this address so all we have to do is do one interval check against it and we are done.

librz/analysis/var_global.c
Basstorm: Most of the jump conditions are comparing the size of the arguments, but there are some special conditions like `JC/JO/JS`, they don't affect the result of constraints because the constraint is the value of the variable.

Basstorm:
> regarding `avga` and `avgd` i think we should set the command as `avga <var_name> <type> [<comment>] @ <address|flag>` and `avgd @ <address|flag>`

Currently, the syntax of newshell doesn't support `addr|flag`. When using them at the same time (defined as `RZ_CMD_ARG_TYPE_STRING`), the parser will report an error `invalid address 'xx.xx'`.

XVilka:
Please don't forget to add a few command integration tests in `test/db/cmd`

librz/include/rz_analysis.h
```diff
@@ -737,6 +738,15 @@ typedef struct rz_analysis_var_t {
 	int argnum;
 } RzAnalysisVar;
 
+typedef struct rz_analysis_var_global_t {
```
ret2libc, line 745: please use doxygen-style doc. And add a simple description for the whole structure as well.

e.g. Use `///<` instead of `//` for struct members.

librz/include/rz_analysis.h
ret2libc: I tend to agree with @thestr4ng3r . This struct should be about one thing and do it well.

librz/core/cmd_descs/cmd_analysis.yaml
```diff
@@ -394,6 +394,68 @@ commands:
...
+          - name: avg
+            summary: show global variables
+            type: RZ_CMD_DESC_TYPE_ARGV_STATE
+            cname: analysis_print_global_variable
+            modes:
+              - RZ_OUTPUT_MODE_STANDARD
+              - RZ_OUTPUT_MODE_JSON
+            args:
+              - name: var_name
+                type: RZ_CMD_ARG_TYPE_STRING
```
ret2libc, line 409: let's add right now a new type in RzCmdArgType (e.g. `RZ_CMD_ARG_TYPE_GLOB_VAR`) and use it wherever a global variable name is used as argument. So when we want to implement autocompletion it will be easy to do. Also, it makes it clearer from the beginning that these are not just regular strings.

librz/core/cmd_analysis.c
```diff
@@ -8088,6 +8088,115 @@ RZ_IPI RzCmdStatus rz_analysis_rtti_demangle_class_name_handler(RzCore *core, in
 	return RZ_CMD_STATUS_ERROR;
 }
 
+RZ_IPI RzCmdStatus rz_analysis_print_global_variable_handler(RzCore *core, int argc, const char **argv, RzCmdStateOutput *state) {
+	char *var_name = NULL;
+	if (argc == 2) {
+		var_name = (char *)argv[1];
```
ret2libc: why the cast to (char *)? (const char *) should be fine, I think.

librz/core/cmd_analysis.c
```diff
@@ -8088,6 +8088,115 @@ RZ_IPI RzCmdStatus rz_analysis_rtti_demangle_class_name_handler(RzCore *core, in
...
+	}
+	rz_analysis_var_global_list_show(core->analysis, state, var_name);
+	return RZ_CMD_STATUS_OK;
+}
+
+RZ_IPI RzCmdStatus rz_analysis_global_variable_add_handler(RzCore *core, int argc, const char **argv) {
+	if (argc < 4 || argc > 5) {
+		return RZ_CMD_STATUS_WRONG_ARGS;
+	}
+	char *var_name = (char *)argv[1];
```
ret2libc: Again, I think these `char *` casts are not necessary.

librz/core/cmd_analysis.c
```diff
@@ -8088,6 +8088,115 @@ RZ_IPI RzCmdStatus rz_analysis_rtti_demangle_class_name_handler(RzCore *core, in
...
+		var_name = (char *)argv[1];
+	}
+	rz_analysis_var_global_list_show(core->analysis, state, var_name);
+	return RZ_CMD_STATUS_OK;
+}
+
+RZ_IPI RzCmdStatus rz_analysis_global_variable_add_handler(RzCore *core, int argc, const char **argv) {
+	if (argc < 4 || argc > 5) {
+		return RZ_CMD_STATUS_WRONG_ARGS;
+	}
```
ret2libc: unnecessary, the number of arguments are already checked by rz-shell.

librz/core/cmd_analysis.c
```diff
@@ -8088,6 +8088,115 @@ RZ_IPI RzCmdStatus rz_analysis_rtti_demangle_class_name_handler(RzCore *core, in
...
+		RZ_LOG_ERROR("%s : Error parsing type: \"%s\" message:\n%s\n", __FUNCTION__, type, errmsg);
+		free(errmsg);
+		return RZ_CMD_STATUS_ERROR;
+	}
+	RzAnalysisVarGlobal *glob = rz_analysis_var_global_new(var_name, addr, comment);
+	if (!glob) {
+		return RZ_CMD_STATUS_ERROR;
+	}
+	rz_analysis_var_global_set_type(glob, typ);
+	if (rz_analysis_var_global_add(core->analysis, glob)) {
```
ret2libc: please do the opposite
```suggestion
	if (!rz_analysis_var_global_add(core->analysis, glob)) {
	    ...
```

For consistency. All the indentend code in the function is "error code", so this looks a bit out of place.

librz/core/cmd_analysis.c
```diff
@@ -8088,6 +8088,115 @@ RZ_IPI RzCmdStatus rz_analysis_rtti_demangle_class_name_handler(RzCore *core, in
...
+		return RZ_CMD_STATUS_OK;
+	}
+	rz_analysis_var_global_free(glob);
+	return RZ_CMD_STATUS_ERROR;
+}
+
+RZ_IPI RzCmdStatus rz_analysis_global_variable_delete_handler(RzCore *core, int argc, const char **argv) {
+	if (argc != 2) {
+		return RZ_CMD_STATUS_WRONG_ARGS;
+	}
```
ret2libc: no need for this

librz/core/cmd_analysis.c
```diff
@@ -8088,6 +8088,115 @@ RZ_IPI RzCmdStatus rz_analysis_rtti_demangle_class_name_handler(RzCore *core, in
...
+	ut64 addr = rz_num_math(core->num, (char *)argv[1]);
```
ret2libc:
```suggestion
	ut64 addr = rz_num_math(core->num, argv[1]);
```

librz/core/cmd_analysis.c
```diff
@@ -8088,6 +8088,115 @@ RZ_IPI RzCmdStatus rz_analysis_rtti_demangle_class_name_handler(RzCore *core, in
...
+	if (!rz_analysis_var_global_delete_byaddr(core->analysis, addr)) {
+		return RZ_CMD_STATUS_ERROR;
+	}
+	return RZ_CMD_STATUS_OK;
+}
+
+RZ_IPI RzCmdStatus rz_analysis_global_variable_rename_handler(RzCore *core, int argc, const char **argv) {
+	if (argc != 3) {
+		return RZ_CMD_STATUS_WRONG_ARGS;
+	}
```
ret2libc: no need for this and also all the others.

librz/core/cmd_analysis.c
```diff
@@ -8088,6 +8088,115 @@ RZ_IPI RzCmdStatus rz_analysis_rtti_demangle_class_name_handler(RzCore *core, in
...
+	char *oldname = (char *)argv[1];
```
ret2libc: also all these (char *) casts are unnecessary.

librz/core/cmd_descs/cmd_analysis.yaml
```diff
@@ -394,6 +394,68 @@ commands:
...
+            summary: add global variable manually
+            type: RZ_CMD_DESC_TYPE_ARGV
+            cname: analysis_global_variable_add
+            args:
+              - name: var_name
+                type: RZ_CMD_ARG_TYPE_STRING
+              - name: addr
+                type: RZ_CMD_ARG_TYPE_RZNUM
+              - name: type
+                type: RZ_CMD_ARG_TYPE_STRING
```
ret2libc: @XVilka shall we use `RZ_CMD_ARG_TYPE_ANY_TYPE` here?

librz/core/cmd_descs/cmd_analysis.yaml
```diff
@@ -394,6 +394,68 @@ commands:
...
+          - name: avgt
+            summary: change the global variable type
+            type: RZ_CMD_DESC_TYPE_ARGV
+            cname: analysis_global_variable_retype
+            args:
+              - name: var_name
+                type: RZ_CMD_ARG_TYPE_STRING
+              - name: type
+                type: RZ_CMD_ARG_TYPE_STRING
```
ret2libc: also here, RZ_CMD_ARG_TYPE_ANY_TYPE ?

librz/analysis/var_global.c
```diff
@@ -0,0 +1,432 @@
...
+/**
+ * \brief Create a new instance of global variable
+ * 
+ * \param name variable name
+ * \param addr variable address
+ * \param comment variable comment
+ * \return RzAnalysisVarGlobal *
+ */
+RZ_API RZ_OWN RzAnalysisVarGlobal *rz_analysis_var_global_new(RZ_NONNULL const char *name, ut64 addr, RZ_NULLABLE const char *comment) {
+	rz_return_val_if_fail(name && addr, NULL);
```
ret2libc:
```suggestion
	rz_return_val_if_fail(name, NULL);
```
Can't we have a global at 0 ?

librz/analysis/var_global.c
```diff
@@ -0,0 +1,432 @@
...
+ * \brief Add the global variable into hashtable
+ * 
+ * \param analysis RzAnalysis
+ * \param global_var Global variable instance
+ * \return true if succeed
+ */
+RZ_API RZ_OWN bool rz_analysis_var_global_add(RzAnalysis *analysis, RzAnalysisVarGlobal *global_var) {
+	rz_return_val_if_fail(analysis && global_var, false);
+	if (rz_analysis_var_global_get_byaddr(analysis, global_var->addr)) {
+		RZ_LOG_ERROR("Global variable at 0x%" PFMT64x " is already exist!\n", global_var->addr);
```
ret2libc:
```suggestion
		RZ_LOG_ERROR("Global variable at 0x%" PFMT64x " already exists!\n", global_var->addr);
```

librz/analysis/var_global.c
```diff
@@ -0,0 +1,432 @@
...
+ * 
+ * \param analysis RzAnalysis
+ * \param name Global Variable name
+ * \return true if succeed
+ */
+RZ_API bool rz_analysis_var_global_delete_byname(RzAnalysis *analysis, const char *name) {
+	rz_return_val_if_fail(analysis && name, false);
+	RzAnalysisVarGlobal *glob = rz_analysis_var_global_get_byname(analysis, name);
+	if (!glob) {
+		RZ_LOG_ERROR("No such global variable!\n");
```
ret2libc: I don't think this is necessary... I think we should not print anything in such cases and it should be the caller of the API to choose what to write, if write at all something.

librz/analysis/var_global.c
```diff
@@ -0,0 +1,432 @@
...
+		return false;
+	}
+	rz_analysis_var_global_free(glob);
+	return ht_pp_delete(analysis->ht_global_var, name);
```
ret2libc: just do:
```suggestion
	return ht_pp_delete(analysis->ht_global_var, name);
```

librz/analysis/analysis.c
```diff
@@ -143,6 +143,7 @@ RZ_API RzAnalysis *rz_analysis_new(void) {
 			rz_analysis_add(analysis, analysis_static_plugins[i]);
 		}
 	}
+	analysis->ht_global_var = ht_pp_new0();
```
ret2libc: this should be `ht_pp_new(NULL, global_kv_free, NULL);`.

And `global_kv_free` should probably free the `kv->key` and also use `rz_analysis_var_global_free` on the `kv->value`. That way whenever you delete an item from the Ht, it will also automatically free the global variable.

ret2libc (commented):
It would be awesome to have some unit tests for the new global API. Those will help develop the API and will work as basic examples on how to use the API and also as regular tests.

librz/include/rz_analysis.h
```diff
@@ -1573,6 +1584,25 @@ RZ_API char *rz_analysis_fcn_format_sig(RZ_NONNULL RzAnalysis *analysis, RZ_NONN
...
+RZ_API RZ_BORROW RzAnalysisVarGlobal *rz_analysis_var_global_get_byaddr(RzAnalysis *analysis, ut64 addr);
+RZ_API RZ_OWN RzList *rz_analysis_var_global_get_all(RzAnalysis *analysis);
+RZ_API bool rz_analysis_var_global_rename(RzAnalysis *analysis, const char *old_name, RZ_NONNULL const char *newname);
+RZ_API bool rz_analysis_var_global_set_comment(RzAnalysis *analysis, const char *name, RZ_NONNULL const char *comment);
+RZ_API void rz_analysis_var_global_set_type(RzAnalysisVarGlobal *glob, RzType *type);
+RZ_API void rz_analysis_var_global_set_access(RzAnalysis *analysis, RzAnalysisVarGlobal *glob, const char *reg, ut64 access_addr, int access_type, st64 stackptr);
+RZ_API void rz_analysis_var_global_remove_access_at(RzAnalysisVarGlobal *glob, ut64 address);
+RZ_API void rz_analysis_var_global_clear_accesses(RzAnalysisVarGlobal *glob);
+RZ_API void rz_analysis_var_global_add_constraint(RzAnalysisVarGlobal *glob, RzTypeConstraint *constraint);
+RZ_API char *rz_analysis_var_global_get_constraints_readable(RzAnalysisVarGlobal *glob);
```
XVilka: Is this `RZ_OWN`?

librz/include/rz_analysis.h
Basstorm: Done, please take a look. @thestr4ng3r  @XVilka

librz/analysis/var_global.c
XVilka: Yes, surely we can.

librz/analysis/var_global.c
```diff
@@ -0,0 +1,480 @@
...
+	if (!ht_pp_insert(analysis->ht_global_var, global_var->name, global_var)) {
+		return false;
+	}
+	GlobalVarNode *node = RZ_NEW0(GlobalVarNode);
+	node->addr = global_var->addr;
+	node->size = 0;
+	if (global_var->type) {
+		node->size = rz_type_db_get_bitsize(analysis->typedb, global_var->type) / 8;
+	}
+	if (!rz_rbtree_aug_insert(&analysis->global_var_tree, &node->addr, &node->rb, global_var_node_cmp, NULL, NULL)) {
```
XVilka: Do we need an augmented RBTree? cc @thestr4ng3r

librz/analysis/var_global.c
Basstorm: We will need the address to delete RBTree here.

wargio:
> > regarding `avga` and `avgd` i think we should set the command as `avga <var_name> <type> [<comment>] @ <address|flag>` and `avgd @ <address|flag>`
> 
> Currently, the syntax of newshell doesn't support `addr|flag`. When using them at the same time (defined as `RZ_CMD_ARG_TYPE_STRING`), the parser will report an error `invalid address 'xx.xx'`.

what i meant that you can use `avgd @ 0x1234567` or `avgd @ some.flag` and what you just need to do is get the current address withing the command, since `@` is interpreted as temporary seek automatically

librz/analysis/var_global.c
```diff
@@ -0,0 +1,370 @@
...
+		rz_list_append(global_vars, glob);
+	} else {
+		global_vars = rz_analysis_var_global_get_all(analysis);
+	}
+
+	RzListIter *it = NULL;
+	char *var_type = NULL;
+	bool json = state->mode == RZ_OUTPUT_MODE_JSON;
+	PJ *pj = json ? state->d.pj : NULL;
+	// to use rz_cmd_state_output_array_start we need to set RzCore as the dependency of RzAnalysis, which is impossible
```
XVilka: To solve this, we could just move this particular function to `librz/core/canalysis.c` instead. We tend to have printing functions there.

librz/core/cmd_analysis.c
```diff
@@ -8088,6 +8088,84 @@ RZ_IPI RzCmdStatus rz_analysis_rtti_demangle_class_name_handler(RzCore *core, in
 	return RZ_CMD_STATUS_ERROR;
 }
 
+RZ_IPI RzCmdStatus rz_analysis_print_global_variable_handler(RzCore *core, int argc, const char **argv, RzCmdStateOutput *state) {
+	char *var_name = NULL;
```
XVilka: Can be solved by defining `var_name` as `const char*` instead.

librz/core/cmd_descs/cmd_analysis.yaml
XVilka: Yes, please.

librz/core/cmd_descs/cmd_analysis.yaml
XVilka: Yes.

librz/analysis/var_global.c
Basstorm: This could be done in another PR.

librz/core/cmd_analysis.c
Basstorm: Ah, I forgot this..

XVilka:
> what i meant that you can use `avgd @ 0x1234567` or `avgd @ some.flag` and what you just need to do is get the current address withing the command, since `@` is interpreted as temporary seek automatically

I suggest to make it in the following PR, with other enhancements, e.g. better analysis integration.

librz/analysis/var_global.c
```diff
@@ -0,0 +1,370 @@
...
+}
+
+/**
+ * \brief Delete and free the global variable by its name
+ * 
+ * \param analysis RzAnalysis
+ * \param name Global Variable name
+ * \return true if succeed
+ */
+RZ_API bool rz_analysis_var_global_delete_byname(RzAnalysis *analysis, const char *name) {
```
XVilka: Please add `RZ_NONNULL const char *name`

librz/analysis/var_global.c
```diff
@@ -0,0 +1,370 @@
...
+}
+
+/**
+ * \brief Add the global variable into hashtable
+ * 
+ * \param analysis RzAnalysis
+ * \param global_var Global variable instance
+ * \return true if succeed
+ */
+RZ_API RZ_OWN bool rz_analysis_var_global_add(RzAnalysis *analysis, RzAnalysisVarGlobal *global_var) {
```
XVilka: `RZ_NONNULL` for `global_var`

librz/analysis/var_global.c
```diff
@@ -0,0 +1,370 @@
...
+}
+
+/**
+ * \brief Get the instance of global variable by its name
+ * 
+ * \param analysis RzAnalysis
+ * \param name Global variable name
+ * \return RzAnalysisVarGlobal *
+ */
+RZ_API RZ_BORROW RzAnalysisVarGlobal *rz_analysis_var_global_get_byname(RzAnalysis *analysis, const char *name) {
```
XVilka: `RZ_NONNULL` for `name`.

librz/analysis/var_global.c
```diff
@@ -0,0 +1,370 @@
...
+
+/**
+ * \brief Rename the global variable
+ * 
+ * \param analysis RzAnalysis
+ * \param old_name The old name of the global variable
+ * \param newname The new name of the global variable
+ * \return true if succeed
+ */
+RZ_API bool rz_analysis_var_global_rename(RzAnalysis *analysis, const char *old_name, RZ_NONNULL const char *newname) {
```
XVilka: `RZ_NONNULL` for `old_name`

librz/analysis/var_global.c
```diff
@@ -0,0 +1,370 @@
...
+}
+
+/**
+ * \brief Set the type of the global variable
+ * 
+ * \param glob Global variable instance
+ * \param type The type to set. RzType*
+ * \return void
+ */
+RZ_API void rz_analysis_var_global_set_type(RzAnalysisVarGlobal *glob, RzType *type) {
```
XVilka: `RZ_NONNULL` and `RZ_BORROW` for `type`.

librz/include/rz_analysis.h
```diff
@@ -1573,6 +1585,21 @@ RZ_API char *rz_analysis_fcn_format_sig(RZ_NONNULL RzAnalysis *analysis, RZ_NONN
 
 RZ_API void rz_analysis_fcn_vars_add_types(RzAnalysis *analysis, RZ_NONNULL RzAnalysisFunction *fcn);
 
+// Global vars
+RZ_API RZ_OWN RzAnalysisVarGlobal *rz_analysis_var_global_new(RZ_NONNULL const char *name, ut64 addr);
+RZ_API RZ_OWN bool rz_analysis_var_global_add(RzAnalysis *analysis, RzAnalysisVarGlobal *global_var);
+RZ_API void rz_analysis_var_global_free(RzAnalysisVarGlobal *glob);
+RZ_API bool rz_analysis_var_global_delete_byname(RzAnalysis *analysis, const char *name);
+RZ_API bool rz_analysis_var_global_delete_byaddr(RzAnalysis *analysis, ut64 addr);
+RZ_API RZ_BORROW RzAnalysisVarGlobal *rz_analysis_var_global_get_byname(RzAnalysis *analysis, const char *name);
```
XVilka: And synchronize those markers with prototypes here.

XVilka (changes requested):

XVilka:
OK, I think the only missing thing now is [de]serialization. These global vars should be serialized for storing in projects, see `librz/analysis/serialize_analysis.c`. I suggest to add new section called `/vars`. Since there were no global vars before, I think no project migration required, but we still should up the project version nevertheless, since there will be the new section. cc @thestr4ng3r

librz/analysis/var_global.c
```diff
@@ -0,0 +1,370 @@
...
+ */
+RZ_API RZ_OWN bool rz_analysis_var_global_add(RzAnalysis *analysis, RzAnalysisVarGlobal *global_var) {
+	rz_return_val_if_fail(analysis && global_var, false);
+	if (rz_analysis_var_global_get_byaddr(analysis, global_var->addr)) {
+		RZ_LOG_ERROR("Global variable %s at 0x%" PFMT64x " already exists!\n", global_var->name, global_var->addr);
+		return false;
+	} else if (ht_pp_find(analysis->ht_global_var, global_var->name, NULL)) {
+		RZ_LOG_ERROR("Global variable %s already exists!\n", global_var->name);
+		return false;
+	}
```
ret2libc: this can be avoided because `ht_pp_insert` will already check if the element exists or not.

librz/analysis/var_global.c
```diff
@@ -0,0 +1,370 @@
...
+ * \return true if succeed
+ */
+RZ_API bool rz_analysis_var_global_delete_byname(RzAnalysis *analysis, const char *name) {
+	rz_return_val_if_fail(analysis && name, false);
+	RzAnalysisVarGlobal *glob = rz_analysis_var_global_get_byname(analysis, name);
+	if (!glob) {
+		return false;
+	}
+	rz_rbtree_delete(&analysis->global_var_tree, &glob->addr, global_var_node_cmp, NULL, NULL, NULL);
+	return ht_pp_delete(analysis->ht_global_var, name);
```
ret2libc: I suggest to delete things in the same way you are doing in rz_analysis_var_global_delete_byaddr.

librz/analysis/var_global.c
```diff
@@ -0,0 +1,370 @@
...
+	}
+	glob->name = strdup(name);
+	glob->addr = addr;
+	return glob;
+}
+
+typedef struct {
+	RBNode rb;
+	ut64 addr;
+	ut64 size;
```
thestr4ng3r: You can just put a `RzAnalysisVarGlobal *` pointer in here instead of addr+size.
In fact, as another alternative to be even more memory-efficient, you can also put the `RBNode` inside the `RzAnalysisVarGlobal` itself and store them in the tree directly without the need for the additional `GlobalVarNode` struct. See for example `RzAnalysisArchHintRecord`/`arch_hints` as an example.

librz/analysis/var_global.c
thestr4ng3r: No, augmentation is only useful when overlaps are allowed, but for global variables this wouldn't make much sense. Instead of `rz_rbtree_aug_insert`, `rz_rbtree_insert` can be used here (but it's just an alias).

librz/analysis/var_global.c
XVilka: Why? This way it's shorter and just one call.

librz/analysis/var_global.c
wargio: next PR

librz/analysis/var_global.c
Basstorm: Yeah, we've got the global variable here, no need to call `rz_analysis_var_global_delete_byaddr` to get again.

librz/analysis/var_global.c
```diff
@@ -0,0 +1,335 @@
...
+	return (RzAnalysisVarGlobal *)ht_pp_find(analysis->ht_global_var, name, NULL);
+}
+
+struct list_addr {
+	RzList /* <RzAnalysisVarGlobal> */ *list;
+	ut64 addr;
+};
+
+/**
+ * \brief Same as rz_analysis_var_global_get_byname but by its address
```
ret2libc:
```suggestion
 * \brief Get the instance of global variable by its address
```

librz/analysis/var_global.c
```diff
@@ -0,0 +1,335 @@
...
+ * 
+ * \param analysis RzAnalysis
+ * \param addr Global variable address
+ * \return RzAnalysisVarGlobal *
+ */
+RZ_API RZ_BORROW RzAnalysisVarGlobal *rz_analysis_var_global_get_byaddr(RzAnalysis *analysis, ut64 addr) {
+	rz_return_val_if_fail(analysis, NULL);
+	RBIter it;
+	RzAnalysisVarGlobal *node, *tmp = NULL;
+	rz_rbtree_foreach (analysis->global_var_tree, it, node, RzAnalysisVarGlobal, rb) {
```
ret2libc: I think this should not be rz_rbtree_foreach, right? Otherwise you lose all the advantage of using an Rbtree if you always have to iterate over all elements. something like rz_rbtree_lower_bound_forward should probably be used. cc @thestr4ng3r

librz/analysis/var_global.c
```diff
@@ -0,0 +1,335 @@
...
+/**
+ * \brief Set the type of the global variable
+ * 
+ * \param glob Global variable instance
+ * \param type The type to set. RzType*
+ * \return void
+ */
+RZ_API void rz_analysis_var_global_set_type(RzAnalysisVarGlobal *glob, RZ_NONNULL RZ_BORROW RzType *type) {
+	rz_return_if_fail(glob && type);
+	glob->type = type;
```
ret2libc, line 245: shall you also free the old glob->type before assigning a new one?

librz/analysis/var_global.c
```diff
@@ -0,0 +1,335 @@
...
+
+/**
+ * \brief Print out the global variables
+ * 
+ * \param analysis RzAnalysis
+ * \param state RzCmdStateOutput *
+ * \param name Which one to print
+ * \return void
+ */
+RZ_API void rz_analysis_var_global_list_show(RzAnalysis *analysis, RzCmdStateOutput *state, RZ_NULLABLE const char *name) {
```
ret2libc: I think this whole function should just be moved to canalysis.c. No reason for not doing it here and now I think. It should just be a simple move and you can directly use rz_cons_printf instead of callbacks.

test/unit/test_analysis_global_var.c
```diff
@@ -0,0 +1,115 @@
...
+	mu_assert_notnull(glob, "get global var by addr");
+	mu_assert_streq(glob->name, "bar", "global var name");
+
+	bool deleted = rz_analysis_var_global_delete_byaddr(analysis, 0x1337);
+	mu_assert_true(deleted, "delete global var");
+	glob = NULL;
+	glob = rz_analysis_var_global_get_byaddr(analysis, 0x1337);
+	mu_assert_null(glob, "get deleted global var");
+
+	// re add
```
ret2libc, line 59: just a note, no need to change it now or at all, but I think it is usually better to have smaller unit tests that just do few things instead of a single one that does many things together.

XVilka:
You forgot to update this test:
```
 [XX] db/cmd/project migrate info
RZ_NOPLUGINS=1 rizin -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -Qc 'Po prj/v1-noreturn.rzdb
' --
-- stderr
--- expected
+++ actual
@@ -1,3 +1,4 @@
 Detailed project load info:
   project migrated from version 1 to 2.
   project migrated from version 2 to 3.
+  project migrated from version 3 to 4.
```

test/db/cmd/cmd_avg
```diff
@@ -0,0 +1,77 @@
...
+avgn foo bar
+avg
+EOF
+EXPECT=<<EOF
+global int foo @ 0x100
+global int bar @ 0x100
+EOF
+RUN
+
+NAME=avgt # retype global variable
```
XVilka, line 65: I suggest to add a couple more tests for more complex types - structures, unions, enums, function pointers.
Also, add a types of pointer and array.

XVilka (changes requested):
Overall looks complete. Lets do one more review sweep, and should be ready to merge.

Basstorm:
Thank you all for pointing out so many problems and suggestions!

librz/analysis/serialize_analysis.c
```diff
@@ -864,6 +864,212 @@ RZ_API RZ_NULLABLE RzAnalysisVar *rz_serialize_analysis_var_load(RZ_NONNULL RzAn
 	return ret;
 }
 
+RZ_API void rz_serialize_analysis_global_var_save(RZ_NONNULL Sdb *db, RZ_NONNULL RzAnalysis *anal) {
```
ret2libc, line 869: cc @thestr4ng3r

librz/analysis/serialize_analysis.c
```diff
@@ -864,6 +864,212 @@ RZ_API RZ_NULLABLE RzAnalysisVar *rz_serialize_analysis_var_load(RZ_NONNULL RzAn
...
+			pj_ka(j, "constrs");
+			RzTypeConstraint *constr;
+			rz_vector_foreach(&var->constraints, constr) {
+				pj_i(j, (int)constr->cond);
+				pj_n(j, constr->val);
+			}
+			pj_end(j);
+		}
+		pj_end(j);
+		sdb_set(db, rz_strbuf_get(&key), pj_string(j), 0);
```
ret2libc: you can use something like:
```
char tmpbuf[32];
sdb_set(db, rz_strf(tmpbuf, "0x%" PFMT64x, var->addr), pj_string(j), 0);
```

librz/core/project_migrate.c
```diff
@@ -131,11 +131,22 @@ RZ_API bool rz_project_migrate_v2_v3(RzProject *prj, RzSerializeResultInfo *res)
 	return true;
 }
 
+RZ_API bool rz_project_migrate_v3_v4(RzProject *prj, RzSerializeResultInfo *res) {
```
ret2libc, line 160: cc @thestr4ng3r is it really necessary to have a migration to add a new field?

Basstorm: Otherwise, it will fail at the unit test. Error: `RZ_PROJECT_ERR_INVALID_CONTENTS `

thestr4ng3r: Yes. In fact I think we should consider removing type links in this pr too, and make the migration convert all type links to global variables with some kind of auto-generated names.
Otherwise the next migration that does typelinks+globals => globals will be trickier because it has to take into account that an auto-generated name might already exist.

librz/include/rz_analysis.h
```diff
@@ -737,6 +739,18 @@ typedef struct rz_analysis_var_t {
...
 
+/**
+ * \brief Global variables
+ */
+typedef struct rz_analysis_var_global_t {
+	RBNode rb; ///< RBTree node for address management
+	char *name; ///< name of the variable
+	ut64 addr; ///< address of the global variable
+	RzType *type; ///< type of the variable
+	ut64 size; ///< size of the variable
```
thestr4ng3r: The size is uniquely defined by the type already, so this field should not exist.

librz/analysis/var_global.c
thestr4ng3r: It would be `rz_rbtree_upper_bound()` here. Also it is necessary to disambiguate what "by its address" means. Usually we have two functions:
* `rz_<...>_get_<x>_at` which gives you the thing that starts exactly at the given address. (use `rz_rbtree_find()`)
* `rz_<...>_get_<x>_in` which gives you the thing that contains the given address. (use `rz_rbtree_upper_bound()` + a single check against the size of the type)

I would recommend to do this here too.

librz/analysis/serialize_analysis.c
```diff
@@ -864,6 +864,212 @@ RZ_API RZ_NULLABLE RzAnalysisVar *rz_serialize_analysis_var_load(RZ_NONNULL RzAn
 	return ret;
 }
 
+RZ_API void rz_serialize_analysis_global_var_save(RZ_NONNULL Sdb *db, RZ_NONNULL RzAnalysis *anal) {
+	rz_return_if_fail(db && anal);
+	RzList *l = rz_analysis_var_global_get_all(anal);
```
thestr4ng3r: Better use `rz_rbtree_foreach` instead of copying everything to a list.

thestr4ng3r (changes requested):

librz/include/rz_analysis.h
Basstorm: `RzType` does not have a size field.

librz/core/project_migrate.c
Basstorm: Done, please take a look.

XVilka:
```
./librz/core/project_migrate.c: In function ‘v3_v4_types_foreach_cb’:
../librz/core/project_migrate.c:148:9: error: passing argument 1 of ‘pj_ks’ from incompatible pointer type [-Werror=incompatible-pointer-types]
  148 |   pj_ks(k, "addr", k);
      |         ^
      |         |
      |         const char *
In file included from ../librz/include/rz_util/rz_json.h:9,
                 from ../librz/include/rz_util/rz_serialize.h:7,
                 from ../librz/include/rz_project.h:7,
                 from ../librz/core/project_migrate.c:4:
../librz/include/rz_util/pj.h:48:12: note: expected ‘PJ *’ {aka ‘struct pj_t *’} but argument is of type ‘const char *’
   48 | RZ_API PJ *pj_ks(PJ *j, const char *k, const char *v);
      |            ^~~~~
```

Basstorm:
> ```
> ./librz/core/project_migrate.c: In function ‘v3_v4_types_foreach_cb’:
> ../librz/core/project_migrate.c:148:9: error: passing argument 1 of ‘pj_ks’ from incompatible pointer type [-Werror=incompatible-pointer-types]
>   148 |   pj_ks(k, "addr", k);
>       |         ^
>       |         |
>       |         const char *
> In file included from ../librz/include/rz_util/rz_json.h:9,
>                  from ../librz/include/rz_util/rz_serialize.h:7,
>                  from ../librz/include/rz_project.h:7,
>                  from ../librz/core/project_migrate.c:4:
> ../librz/include/rz_util/pj.h:48:12: note: expected ‘PJ *’ {aka ‘struct pj_t *’} but argument is of type ‘const char *’
>    48 | RZ_API PJ *pj_ks(PJ *j, const char *k, const char *v);
>       |            ^~~~~
> cc1: all warnings being treated as errors
> ```

I am waiting for my local asan build.

librz/include/rz_analysis.h
XVilka: There is `rz_type_size()` function.

librz/core/project_migrate.c
XVilka: If we remove typelinks, should we remove the `tl` commands to then? Along with the `librz/analysis/typelink.c`? There is more than that. `librz/core/analysis_tp.c` uses linked types during the type propagation, thus all APIs should be changed as well. It will require too many changes, IMHO. I wonder if it make sense to postpone that change. cc @thestr4ng3r

XVilka (changes requested):
Overall is good to go, I have doubts about typelink changes though, see the comment.

XVilka:
```
[i] ℤ rg rz_analysis_type_link                                                                                                                                                                                                    12:53:55 
librz/include/rz_analysis.h
1624:RZ_API bool rz_analysis_type_link_exists(RzAnalysis *analysis, ut64 addr);
1625:RZ_API RZ_BORROW RzType *rz_analysis_type_link_at(RzAnalysis *analysis, ut64 addr);
1629:RZ_API bool rz_analysis_type_link_offset(RzAnalysis *analysis, RZ_BORROW RzType *type, ut64 addr);
1630:RZ_API RZ_OWN RzList /* RzType */ *rz_analysis_type_links(RzAnalysis *analysis);
1631:RZ_API RZ_OWN RzList /* RzTypePath */ *rz_analysis_type_links_by_offset(RzAnalysis *analysis, ut64 offset);

librz/core/ctypes.c
983:    RzType *link = rz_analysis_type_link_at(core->analysis, addr);

librz/core/disasm.c
5326:           RzType *link_type = rz_analysis_type_link_at(core->analysis, ds->addr + idx);

librz/analysis/typelink.c
16:RZ_API bool rz_analysis_type_link_exists(RzAnalysis *analysis, ut64 addr) {
31:RZ_API RZ_BORROW RzType *rz_analysis_type_link_at(RzAnalysis *analysis, ut64 addr) {
96:RZ_API RZ_OWN RzList /* RzType */ *rz_analysis_type_links(RzAnalysis *analysis) {
142:RZ_API RZ_OWN RzList /* RzTypePath */ *rz_analysis_type_links_by_offset(RzAnalysis *analysis, ut64 offset) {
```
These typelinks are used in `librz/core/ctypes.c` in `resolve_type_links()` and subsequently in `rz_core_link_stroff()` functions. which is then used in the analysis for the `aat` command:
```
[i] ℤ rg rz_core_link_stroff                                                                                                                                                                                                      13:03:12 
librz/include/rz_core.h
526:RZ_API void rz_core_link_stroff(RzCore *core, RzAnalysisFunction *fcn);

librz/core/ctypes.c
761:RZ_API void rz_core_link_stroff(RzCore *core, RzAnalysisFunction *fcn) {
977:            rz_core_link_stroff(core, fcn);

librz/core/cmd_analysis.c
7936:                           rz_core_link_stroff(core, fcn);
7949:                           rz_core_link_stroff(core, fcn);
```

librz/include/rz_analysis.h
Basstorm: Yeah, I use `rz_type_db_get_bitsize` to set the size field. I think this is reasonable.

XVilka: Please note, that `rz_type_db_get_bitsize()` recalculates the new size depending on the chosen `asm.arch`, `asm.bits`, `asm.cpu`, `asm.os`, etc

librz/core/project_migrate.c
thestr4ng3r: Postponing is a fine option too, yes.

librz/analysis/var_global.c
```diff
@@ -0,0 +1,310 @@
...
+	RBNode *node = rz_rbtree_upper_bound(analysis->global_var_tree, &addr, global_var_node_cmp, NULL);
+	if (!node) {
+		return NULL;
+	}
+	RzAnalysisVarGlobal *var = (RzAnalysisVarGlobal *)container_of(node, RzAnalysisVarGlobal, rb);
+	if (!var) {
+		return NULL;
+	}
+	ut64 size = rz_type_db_get_bitsize(analysis->typedb, var->type) / 8;
+	if (addr > (var->addr + size - 1) || addr < var->addr) {
```
thestr4ng3r: This check is correct but it can be simplified a bit:
```suggestion
	if (addr >= var->addr + size) {
```
`addr < var->addr` is already ensured by `rz_rbtree_upper_bound()`.

librz/core/project_migrate.c
```diff
@@ -131,11 +132,59 @@ RZ_API bool rz_project_migrate_v2_v3(RzProject *prj, RzSerializeResultInfo *res)
...
+		return false;
+	}
+	Sdb *typelinks_db = sdb_ns(analysis_db, "typelinks", true);
+	sdb_foreach(typelinks_db, v3_v4_types_foreach_cb, &ctx);
+	RzListIter *it;
+	char *s;
+	rz_list_foreach (ctx.moved_keys, it, s) {
+		sdb_unset(typelinks_db, s, 0);
+	}
+	rz_list_free(ctx.moved_keys);
```
thestr4ng3r, line 183: If we decide to postpone removing typelinks completely in this pr (which seems to be the case), then the migration should also not move the typelinks yet and it should be reverted to how it was before (just creating the global vars) namespace.

But this code you added here looks good already so in my opinion you can just enclose it in `#if 0` so we can reuse it in the next iteration.

librz/core/cmd_analysis.c
```diff
@@ -8088,6 +8088,75 @@ RZ_IPI RzCmdStatus rz_analysis_rtti_demangle_class_name_handler(RzCore *core, in
...
+	rz_analysis_var_global_list_show(core->analysis, state, argv[1]);
+	return RZ_CMD_STATUS_OK;
+}
+
+RZ_IPI RzCmdStatus rz_analysis_global_variable_add_handler(RzCore *core, int argc, const char **argv) {
+	const char *var_name = argv[1];
+	ut64 addr = rz_num_math(core->num, argv[2]);
+	const char *type = argv[3];
+
+	RzTypeParser *parser = rz_type_parser_new();
```
XVilka: I think it's better to use `core->analysis->parser` instead, to be able to reuse all currently loaded types

XVilka (changes requested):

XVilka (approved):

test/unit/test_project_migrate.c
```diff
@@ -190,7 +194,6 @@ int all_tests() {
 	mu_run_test(test_migrate_v2_v3);
 	mu_run_test(test_load_v1_noreturn);
 	mu_run_test(test_load_v1_noreturn_empty);
-	mu_run_test(test_load_v2_typelink);
```
XVilka, line 193: I guess it shouldn't be removed then?

XVilka (approved):

thestr4ng3r (approved):

wargio (approved):

librz/core/cmd_descs/cmd_analysis.yaml
DMaroo: This has been implemented now in #1577.

PR #1408, opened 2021-08-06
socket process priority change now possible for haiku since beta3 rel…

devnexen:
…ease.

 

### DIFF
--- a/librz/socket/run.c
+++ b/librz/socket/run.c
@@ -1217,7 +1217,7 @@ RZ_API int rz_run_start(RzRunProfile *p) {
 #endif
 
 		if (p->_nice) {
-#if __UNIX__ && !defined(__HAIKU__)
+#if HAVE_NICE
 			if (nice(p->_nice) == -1) {
 				return 1;
 			}
--- a/librz/include/rz_types.h
+++ b/librz/include/rz_types.h
@@ -349,11 +349,6 @@ static inline void *rz_new_copy(int size, void *data) {
 #include <dirent.h>
 #include <unistd.h>
 #include <sys/time.h>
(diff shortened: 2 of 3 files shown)


ret2libc (approved):
Thanks!

ret2libc:
@devnexen not related to this, but could you check if everything works on Haiku if we remove the #ifdef block at https://github.com/rizinorg/rizin/blob/dev/librz/include/rz_types.h#L352 ? We would like to move away from all this "if windows or osx or haiku or ..." stuff and let just meson check which functions/headers/macros are available and use them accordingly.

wargio (approved):

librz/socket/run.c
```diff
@@ -1217,7 +1217,7 @@ RZ_API int rz_run_start(RzRunProfile *p) {
 #endif
 
 		if (p->_nice) {
-#if __UNIX__ && !defined(__HAIKU__)
```
yossizap, line 1220: Is it possible to define whether this functionality is available with meson's [`compiler.run`](https://mesonbuild.com/Compiler-properties.html) during compile time? Or with some version define from a system header. Seeing that this is only available in a beta build and some people may still use older versions.

devnexen: you re kind of forced to update to beta3 via the package manager.

ret2libc: We want to do that in the future. In meson.build you can check if nice function exists on the host target and then have a define HAVE_NICE If it is.

devnexen: sorry I misread the line commented, yes sure doing it.

yossizap (approved):

PR #1415, opened 2021-08-09
Converting ph commands

wargio:

**Detailed description**

ph commands has been converted.

i added `phl` for listing all algorithms

### DIFF
--- a/librz/core/cmd_descs/cmd_print.yaml
+++ b/librz/core/cmd_descs/cmd_print.yaml
@@ -53,14 +53,11 @@ commands:
     cname: cmd_print_default
     subcommands:
       - name: ph
-        summary: Prints a hash/message digest or entropy
+        summary: Prints a hash/message digest or entropy (use @! to change the block size)
         cname: cmd_print_msg_digest
         args:
           - name: algo
             type: RZ_CMD_ARG_TYPE_STRING
-          - name: length
-            type: RZ_CMD_ARG_TYPE_NUM
-            optional: true
       - name: phl
         summary: Lists all the supported algorithms
         cname: cmd_print_msg_digest_algo_list
--- a/librz/core/cmd_descs/cmd_print.yaml
+++ b/librz/core/cmd_descs/cmd_print.yaml
@@ -48,6 +48,29 @@ commands:
             type: RZ_CMD_ARG_TYPE_NUM
           - name: h
             type: RZ_CMD_ARG_TYPE_NUM
+  - name: ph
+    summary: Print hash/message digest or entropy
+    cname: cmd_print_default
+    subcommands:
+      - name: ph
+        summary: Prints a hash/message digest or entropy
+        cname: cmd_print_msg_digest
+        args:
+          - name: algo
+            type: RZ_CMD_ARG_TYPE_STRING
+          - name: length
+            type: RZ_CMD_ARG_TYPE_NUM
+            optional: true
+      - name: phl
+        summary: Lists all the supported algorithms
+        cname: cmd_print_msg_digest_algo_list
+        type: RZ_CMD_DESC_TYPE_ARGV_STATE
+        modes:
+          - RZ_OUTPUT_MODE_STANDARD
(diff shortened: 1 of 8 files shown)


librz/core/cmd_descs/cmd_print.yaml
```diff
@@ -48,6 +48,29 @@ commands:
...
+    cname: cmd_print_default
+    subcommands:
+      - name: ph
+        summary: Prints a hash/message digest or entropy
+        cname: cmd_print_msg_digest
+        args:
+          - name: algo
+            type: RZ_CMD_ARG_TYPE_STRING
+          - name: length
+            type: RZ_CMD_ARG_TYPE_NUM
```
ret2libc:
```suggestion
            type: RZ_CMD_ARG_TYPE_RZNUM
```

librz/core/cmd_print.c
```diff
@@ -3207,59 +3207,46 @@ static void disasm_strings(RzCore *core, const char *input, RzAnalysisFunction *
...
-			} else {
-				handle_msg_digest(plugin->name, core->block, len);
-			}
-			handled_cmd = true;
-			break;
+
+	ut32 old_blocksize = core->blocksize;
+	ut32 new_blocksize = old_blocksize;
+
+	if (RZ_STR_ISNOTEMPTY(argv[1])) {
```
ret2libc: argv[1] cannot be null, so you can avoid this `if`.

librz/core/cmd_print.c
```diff
@@ -3207,59 +3207,46 @@ static void disasm_strings(RzCore *core, const char *input, RzAnalysisFunction *
...
+		plugin = rz_msg_digest_plugin_by_name(argv[1]);
+	}
+
+	if (!plugin) {
+		RZ_LOG_ERROR("algorithm '%s' does not exists\n", argv[1]);
+		return RZ_CMD_STATUS_INVALID;
```
ret2libc:
```suggestion
		return RZ_CMD_STATUS_ERROR;
```
I think this is better. _INVALID is used mainly for shell/parser/grammar errors.

librz/core/cmd_print.c
```diff
@@ -3207,59 +3207,46 @@ static void disasm_strings(RzCore *core, const char *input, RzAnalysisFunction *
...
+	}
+
+	if (argc == 3) {
+		if (!rz_num_is_valid_input(core->num, argv[2])) {
+			return RZ_CMD_STATUS_INVALID;
```
ret2libc: ERROR as well.

librz/core/cmd_print.c
```diff
@@ -3207,59 +3207,46 @@ static void disasm_strings(RzCore *core, const char *input, RzAnalysisFunction *
...
 		}
+		new_blocksize = rz_num_math(core->num, argv[2]);
+
+		if (!rz_core_block_size(core, new_blocksize)) {
+			rz_core_block_size(core, old_blocksize);
```
ret2libc: this is likely not necessary if the previous call has failed.

librz/core/cmd_descs/cmd_print.yaml
```diff
@@ -48,6 +48,29 @@ commands:
...
+            optional: true
```
ret2libc: In theory we could eliminate these kind of things and just rely on `@!`, which works for every command. If we use that approach we can assume all commands work on core->offset and with core->blocksize, but you can always use `@` and `@!` to change those. Just an idea anyway...

librz/core/cmd_print.c
wargio: i think i made a mistake. that was supposed to be the number.

librz/core/cmd_descs/cmd_print.yaml
wargio: i agree. i thought about that too.. but usability may sink..

ret2libc (approved):

XVilka (approved):

PR #1446, opened 2021-08-13
Add `rz_deflate*` functions and save projects in compressed format

### LINKED ISSUE #1417: Save project in compressed format, to save disk space.
Since default gzip will compress 8.1M rzdb to 1.1M, it will save a lot of disk space.

Rizin has `rz_inflate_*()` functions for decompression of the Zlib data. It might be useful to add the `rz_deflate_*()` functions and use them in the compressing the projects.

See:
- `librz/include/rz_util/rz_file.h`
- `librz/util/file.c`
- `librz/core/project.h`

DMaroo:
# SQUASH ME

**Detailed description**

Currently, rizin and Cutter projects are stored without any compression. With this PR, I have added compression features which will allow to store the project files in a compressed format, thus taking up less memory space.

**Test plan**

The integration tests and the unit tests are passing. The CI is green.

**Closing issues**

Fixes #1417.


librz/util/zip.c
```diff
@@ -107,4 +106,83 @@ RZ_API ut8 *rz_inflatew(const ut8 *src, int srcLen, int *srcConsumed, int *dstLe
 	inflateEnd(&stream);
 	free(dst);
 	return NULL;
-}
\ No newline at end of file
+}
+
+/**
+ * @brief deflate uncompressed data to zlib or gzipped, use MAX_WBITS as the window size logarithm.
```
XVilka: Please use `\brief` Doxygen syntax

librz/util/zip.c
```diff
@@ -107,4 +106,83 @@ RZ_API ut8 *rz_inflatew(const ut8 *src, int srcLen, int *srcConsumed, int *dstLe
...
+ * @see rz_deflatew()
+ */
+RZ_API ut8 *rz_deflate(const ut8 *src, int srcLen, int *srcConsumed, int *dstLen) {
```
XVilka: Please add `RZ_NONNULL` for `src`, also `rz_return_val_if_fail()`

librz/util/zip.c
```diff
@@ -107,4 +106,83 @@ RZ_API ut8 *rz_inflatew(const ut8 *src, int srcLen, int *srcConsumed, int *dstLe
...
+ * \brief compress/deflate data to zlib or gzip.
+ * \param src source uncompressed bytes
+ * \param srcLen source bytes length
+ * \param srcConsumed consumed source bytes length
+ * \param dstLen compressed bytes length
+ * \param controls the size of the history buffer (or “window size”), and what header and trailer format is expected.
+ * \return ptr to compressed
+ */
+RZ_API ut8 *rz_deflatew(const ut8 *src, int srcLen, int *srcConsumed, int *dstLen, int wbits) {
+	int err = 0;
```
XVilka, line 137: Same here

XVilka (changes requested):

librz/include/rz_util/rz_file.h
```diff
@@ -45,6 +45,8 @@ RZ_API char *rz_file_path_unix_to_local(const char *path);
 RZ_API char *rz_file_binsh(void);
 RZ_API ut8 *rz_inflatew(const ut8 *src, int srcLen, int *srcConsumed, int *dstLen, int wbits);
 RZ_API ut8 *rz_inflate(const ut8 *src, int srcLen, int *srcConsumed, int *dstLen);
+RZ_API ut8 *rz_deflatew(const ut8 *src, int srcLen, int *srcConsumed, int *dstLen, int wbits);
```
XVilka: When you add those `RZ_*` markers in the function definition, please add them in header files as well. Would be awesome to have them for the `rz_inflate*()` functions as well, btw.

librz/util/zip.c
```diff
@@ -24,31 +24,33 @@ static const char *gzerr(int n) {
...
 
 /**
- * @brief inflate zlib compressed or gzipped. The input must be a raw stream with no header or trailer.
- * @see rz_inflatew()
+ * \brief inflate zlib compressed or gzipped. The input must be a raw stream with no header or trailer.
+ * \see rz_inflatew()
  */
-RZ_API ut8 *rz_inflate_ignore_header(const ut8 *src, int srcLen, int *srcConsumed, int *dstLen) {
+RZ_API ut8 *rz_inflate_ignore_header(RZ_NONNULL const ut8 *src, int srcLen, int *srcConsumed, int *dstLen) {
 	return rz_inflatew(src, srcLen, srcConsumed, dstLen, -MAX_WBITS);
```
XVilka, line 43: Also `rz_return_val_if_fail()`

librz/util/zip.c
```diff
@@ -107,4 +108,83 @@ RZ_API ut8 *rz_inflatew(const ut8 *src, int srcLen, int *srcConsumed, int *dstLe
...
-}
\ No newline at end of file
+}
+
+/**
+ * \brief deflate uncompressed data to zlib or gzipped, use MAX_WBITS as the window size logarithm.
+ * \see rz_deflatew()
+ */
+RZ_API ut8 *rz_deflate(RZ_NONNULL const ut8 *src, int srcLen, int *srcConsumed, int *dstLen) {
+	return rz_deflatew(src, srcLen, srcConsumed, dstLen, MAX_WBITS + 32);
```
XVilka: And here

XVilka (changes requested):

librz/util/zip.c
DMaroo: I have added `rz_return_val_if_fail(...)` only in `rz_deflatew(...)` and `rz_inflatew(...)`, since these both are the functions which are indirectly called by other functions (like `rz_deflate(...)`). Putting checks in all the functions would be redundant. I hope this is fine. Thank you.

librz/include/rz_util/rz_file.h
DMaroo: Oh, I am sorry. My bad. Thank you. :)

librz/util/zip.c
DMaroo: But wouldn't that be redundant, since the non-null check is present in `rz_inflatew(...)`?

librz/util/zip.c
DMaroo: But wouldn't that be redundant, since the non-null check is present in `rz_deflatew(...)` (same reasoning as above)?

librz/util/zip.c
XVilka: These checks appear only in certain "debug" builds, during the release, packaging, etc they evaluate to nothing. See https://github.com/rizinorg/rizin/blob/dev/DEVELOPERS.md#c

DMaroo: Alright, makes sense. I'll add the checks to make it consistent.

DMaroo:
Thank you for the review. I'll now go forward with implementing compression in while saving a project.

ret2libc (commented):
Hi! Please add unit tests for these two functions. Cc @thestr4ng3r

librz/util/file.c
```diff
@@ -1183,6 +1183,32 @@ RZ_API bool rz_file_copy(const char *src, const char *dst) {
 #endif
 }
 
+RZ_API bool rz_file_gzip(const char *src) {
```
XVilka: When you add new API functions  please add Doxygen as well.

librz/util/file.c
```diff
@@ -1183,6 +1183,32 @@ RZ_API bool rz_file_copy(const char *src, const char *dst) {
...
+	/* TODO: implement gzip for Windows */
+#if __WINDOWS__
+	return true;
+#else
+	char *src2 = rz_str_replace(strdup(src), "'", "\'", 1);
+	int rc = rz_sys_cmdf("gzip -f '%s'", src2);
```
XVilka: The whole function is wrong. Why do you need to call `gzip` if you already have `zlib`? Just use these `inflate`/`deflate` functions.

librz/util/file.c
```diff
@@ -1183,6 +1183,32 @@ RZ_API bool rz_file_copy(const char *src, const char *dst) {
...
+
+	free(src2);
+	return rc == 0;
+#endif
+}
+
+RZ_API bool rz_file_gunzip(const char *src) {
```
XVilka: Same here

XVilka (changes requested):

librz/util/file.c
DMaroo: Yes. @thestr4ng3r informed me to use `rz_deflate` and `rz_inflate` functions. I'll be refactoring it now. I'll squash the commits in the end (to keep the commit history clean).

DMaroo:
The `rz_deflate` doesn't seem to be working correctly. Please feel free to suggest any improvements.

librz/util/file.c
```diff
@@ -1257,3 +1257,40 @@ RZ_API char *rz_file_path_join(const char *s1, const char *s2) {
...
 	return rz_str_newf("%s%s%s", s1, sep, s2);
 }
+
+/**
+ * \brief zip the contents of src and store in dst
+ * \param src source file (string containing filename)
+ * \param dst destination file (string containing filename)
+ * \return true, if successful; false otherwise
+ */
+RZ_API bool rz_file_zip(const char *src, const char *dst) {
```
XVilka: It's not ZIP. ZIP is a different format. I suggest to keep it as `rz_file_deflate()` and `rz_file_inflate()` instead.

librz/util/file.c
```diff
@@ -1257,3 +1257,40 @@ RZ_API char *rz_file_path_join(const char *s1, const char *s2) {
...
+	size_t len, comp_len;
+	char *content = rz_file_slurp(src, &len);
+	char *comp_content = rz_deflate(content, len, NULL, &comp_len);
```
XVilka: Check for NULL in both cases

librz/util/file.c
```diff
@@ -1257,3 +1257,40 @@ RZ_API char *rz_file_path_join(const char *s1, const char *s2) {
...
+/**
+ * \brief unzip the contents of src and store in dst
+ * \param src source file (string containing filename)
+ * \param dst destination file (string containing filename)
+ * \return true, if successful; false otherwise
+ */
+RZ_API bool rz_file_unzip(const char *src, const char *dst) {
+	size_t len, decomp_len;
+	char *content = rz_file_slurp(src, &len);
+	char *decomp_content = rz_inflate(content, len, NULL, &decomp_len);
```
XVilka: Same here - check for NULL in both cases.

XVilka (changes requested):

librz/util/file.c
```diff
@@ -1257,3 +1257,53 @@ RZ_API char *rz_file_path_join(const char *s1, const char *s2) {
...
+RZ_API bool rz_file_deflate(const char *src, const char *dst) {
+	size_t len, comp_len;
+	char *content = rz_file_slurp(src, &len);
+	if (!content) {
+		return false;
+	}
+
+	char *comp_content = rz_deflate(content, len, NULL, &comp_len);
+	if (!comp_content) {
+		return false;
```
XVilka: Don't forget `free(content)` here.

librz/util/file.c
```diff
@@ -1257,3 +1257,53 @@ RZ_API char *rz_file_path_join(const char *s1, const char *s2) {
...
+ */
+RZ_API bool rz_file_inflate(const char *src, const char *dst) {
+	size_t len, decomp_len;
+	char *content = rz_file_slurp(src, &len);
+	if (!content) {
+		return false;
+	}
+
+	char *decomp_content = rz_inflate(content, len, NULL, &decomp_len);
+	if (!decomp_content) {
```
XVilka: Don't forget `free(content)` here.

librz/util/file.c
```diff
@@ -1257,3 +1257,53 @@ RZ_API char *rz_file_path_join(const char *s1, const char *s2) {
...
 	return rz_str_newf("%s%s%s", s1, sep, s2);
 }
+
+/**
+ * \brief zip the contents of src and store in dst
+ * \param src source file (string containing filename)
+ * \param dst destination file (string containing filename)
+ * \return true, if successful; false otherwise
+ */
+RZ_API bool rz_file_deflate(const char *src, const char *dst) {
```
XVilka: Please mark `src` and `dst` as `RZ_NONNULL` and add `rz_return_val_if_fail(src && dst, false)`

librz/util/file.c
```diff
@@ -1257,3 +1257,53 @@ RZ_API char *rz_file_path_join(const char *s1, const char *s2) {
...
+	return result;
+}
+
+/**
+ * \brief unzip the contents of src and store in dst
+ * \param src source file (string containing filename)
+ * \param dst destination file (string containing filename)
+ * \return true, if successful; false otherwise
+ */
+RZ_API bool rz_file_inflate(const char *src, const char *dst) {
```
XVilka: Please mark `src` and `dst` as `RZ_NONNULL` and add `rz_return_val_if_fail(src && dst, false)`

librz/include/rz_util/rz_file.h
```diff
@@ -67,6 +69,8 @@ RZ_API char *rz_file_tmpdir(void);
 RZ_API char *rz_file_readlink(const char *path);
 RZ_API bool rz_file_copy(const char *src, const char *dst);
 RZ_API RzList *rz_file_globsearch(const char *globbed_path, int maxdepth);
+RZ_API bool rz_file_deflate(const char *src, const char *dst);
```
XVilka: Also `RZ_NONNULL` here for both `src` and `dst`.

XVilka (changes requested):

librz/core/project.c
```diff
@@ -24,8 +24,9 @@ RZ_API RZ_NONNULL const char *rz_project_err_message(RzProjectErr err) {
 		return "invalid content encountered";
 	case RZ_PROJECT_ERR_MIGRATION_FAILED:
 		return "migration failed";
-	case RZ_PROJECT_ERR_UNKNOWN:
-		break;
+	case RZ_PROJECT_ERR_ZIP:
```
XVilka: Since we do not use ZIP, maybe make it more generic? `RZ_PROJECT_ERR_COMPRESSION_FAILED`? Also it's better to separate compression from decompression errors.

XVilka (changes requested):

librz/core/project.c
```diff
@@ -24,8 +24,9 @@ RZ_API RZ_NONNULL const char *rz_project_err_message(RzProjectErr err) {
 		return "invalid content encountered";
 	case RZ_PROJECT_ERR_MIGRATION_FAILED:
 		return "migration failed";
-	case RZ_PROJECT_ERR_UNKNOWN:
-		break;
+	case RZ_PROJECT_ERR_COMPRESSION_FAILED:
+		return "project file compression failed";
+	case RZ_PROJECT_ERR_UNKNOWN: break;
```
ret2libc: style

librz/core/project.c
```diff
@@ -38,22 +39,33 @@ RZ_API RzProjectErr rz_project_save(RzCore *core, RzProject *prj, const char *fi
 }
 
 RZ_API RzProjectErr rz_project_save_file(RzCore *core, const char *file) {
+	char *tmp_file = rz_file_temp("svprj");
```
ret2libc: I think it's better/safer to use `rz_file_mkstemp`. `rz_file_temp` uses `mktemp`, which is not good and probably we should deprecate this function...

librz/core/project.c
```diff
@@ -63,10 +75,19 @@ RZ_API RzProject *rz_project_load_file_raw(const char *file) {
 	if (!prj) {
 		return NULL;
 	}
-	if (!sdb_text_load(prj, file)) {
+
+	char *tmp_file = rz_file_temp("ldprj");
```
ret2libc: same here. mkstemp.

test/unit/test_zip.c
```diff
@@ -0,0 +1,67 @@
+#include <rz_util.h>
```
XVilka: I think the test should be called `test_inflate_deflate.c` not `test_zip`.

test/unit/meson.build
```diff
@@ -91,7 +91,8 @@ if get_option('enable_tests')
     'uleb128',
     'unum',
     'util',
-    'vector'
+    'vector',
+	'zip'
```
XVilka: Broken indentation.

test/unit/test_zip.c
```diff
@@ -0,0 +1,67 @@
...
+	{ "1234567890abcdefghijklmnopqrstuvwxyz\n", "\x1f\x8b\x08\x08\x00\x00\x00\x00\x00\x03\x31\x33\x00\x33\x34\x32\x36\x31\x35\x33\xb7\xb0\x34\x48\x4c\x4a\x4e\x49\x4d\x4b\xcf\xc8\xcc\xca\xce\xc9\xcd\xcb\x2f\x28\x2c\x2a\x2e\x29\x2d\x2b\xaf\xa8\xac\xe2\x02\x00\x9f\x79\xdc\xdb\x25\x00\x00\x00" }
+};
+
+bool test_rz_inflate(void) {
+	int len0 = strlen(test_cases[0].inflated);
+	unsigned char *inflated0 = rz_inflate((unsigned char *)test_cases[0].deflated, 287, NULL, NULL);
```
XVilka: I recommend to check also `inflated0` for NULL with `mu_assert_nonnull()`

test/unit/test_zip.c
```diff
@@ -0,0 +1,67 @@
...
+	char *null_terminated0 = malloc(sizeof(char) * (len0 + 1));
+	strncpy(null_terminated0, (char *)inflated0, len0);
+	null_terminated0[len0] = '\0';
+	free(inflated0);
+	mu_assert_streq(null_terminated0, test_cases[0].inflated, "rz_inflate failed for test 0");
+	free(null_terminated0);
+
+	int len1 = strlen(test_cases[1].inflated);
+	unsigned char *inflated1 = rz_inflate((unsigned char *)test_cases[1].deflated, 640, NULL, NULL);
```
XVilka: Same

test/unit/test_zip.c
```diff
@@ -0,0 +1,67 @@
...
+	char *null_terminated1 = malloc(sizeof(char) * (len1 + 1));
+	strncpy(null_terminated1, (char *)inflated1, len1);
+	null_terminated1[len1] = '\0';
+	free(inflated1);
+	mu_assert_streq(null_terminated1, test_cases[1].inflated, "rz_inflate failed for test 1");
+	free(null_terminated1);
+
+	int len2 = strlen(test_cases[2].inflated);
+	unsigned char *inflated2 = rz_inflate((unsigned char *)test_cases[2].deflated, 60, NULL, NULL);
```
XVilka: Same

test/unit/test_zip.c
```diff
@@ -0,0 +1,67 @@
...
+	null_terminated2[len2] = '\0';
+	free(inflated2);
+	mu_assert_streq(null_terminated2, test_cases[2].inflated, "rz_inflate failed for test 2");
+	free(null_terminated2);
+
+	mu_end;
+}
+
+bool test_rz_deflate(void) {
+	unsigned char *deflated0 = rz_deflate((unsigned char *)test_cases[0].inflated, strlen(test_cases[0].inflated), NULL, NULL);
```
XVilka: I recommend to check first for NULL with `mu_assert_nonnull`

test/unit/test_zip.c
```diff
@@ -0,0 +1,67 @@
...
+	mu_assert_streq((char *)deflated0, test_cases[0].deflated, "rz_deflate failed for 0");
+	free(deflated0);
+
+	unsigned char *deflated1 = rz_deflate((unsigned char *)test_cases[1].inflated, strlen(test_cases[1].inflated), NULL, NULL);
```
XVilka: Same

test/unit/test_zip.c
```diff
@@ -0,0 +1,67 @@
...
+	mu_assert_streq((char *)deflated1, test_cases[1].deflated, "rz_deflate failed for 1");
+	free(deflated1);
+
+	unsigned char *deflated2 = rz_deflate((unsigned char *)test_cases[2].inflated, strlen(test_cases[2].inflated), NULL, NULL);
```
XVilka: Same

XVilka (changes requested):

librz/core/project.c
```diff
@@ -38,22 +39,33 @@ RZ_API RzProjectErr rz_project_save(RzCore *core, RzProject *prj, const char *fi
...
 	}
+
+	if (err == RZ_PROJECT_ERR_SUCCESS) {
+		if (!rz_file_deflate(tmp_file, file)) {
+			free(tmp_file);
+			return RZ_PROJECT_ERR_COMPRESSION_FAILED;
+		}
+	}
+	free(tmp_file);
+
```
ret2libc:
```suggestion
	if (!sdb_text_save(prj, tmp_file, true)) {
		err = RZ_PROJECT_ERR_FILE;
	}
	sdb_free(prj);
	if (err != RZ_PROJECT_ERR_SUCCESS) {
		goto tmp_file_err;
	}
		
	if (!rz_file_deflate(tmp_file, file)) {
		err = RZ_PROJECT_ERR_COMPRESSION_FAILED;
		goto tmp_file_err;
	}
	rz_config_set(core->config, "prj.file", file);
tmp_file_err:
	free(tmp_file);
```

librz/core/project.c
```diff
@@ -63,10 +75,19 @@ RZ_API RzProject *rz_project_load_file_raw(const char *file) {
...
+		return NULL;
+	}
+
+	if (!sdb_text_load(prj, tmp_file)) {
 		sdb_free(prj);
+		free(tmp_file);
 		return NULL;
 	}
+	free(tmp_file);
+
```
ret2libc, line 121:
```suggestion
```

test/unit/meson.build
ret2libc: indentation

test/unit/test_zip.c
```diff
@@ -0,0 +1,67 @@
...
+
+bool test_rz_inflate(void) {
+	int len0 = strlen(test_cases[0].inflated);
+	unsigned char *inflated0 = rz_inflate((unsigned char *)test_cases[0].deflated, 287, NULL, NULL);
+	char *null_terminated0 = malloc(sizeof(char) * (len0 + 1));
+	strncpy(null_terminated0, (char *)inflated0, len0);
+	null_terminated0[len0] = '\0';
+	free(inflated0);
+	mu_assert_streq(null_terminated0, test_cases[0].inflated, "rz_inflate failed for test 0");
+	free(null_terminated0);
```
ret2libc: can we put this in a loop and just iterate over test_cases?

test/unit/test_zip.c
```diff
@@ -0,0 +1,67 @@
...
+	mu_assert_streq(null_terminated2, test_cases[2].inflated, "rz_inflate failed for test 2");
+	free(null_terminated2);
+
+	mu_end;
+}
+
+bool test_rz_deflate(void) {
+	unsigned char *deflated0 = rz_deflate((unsigned char *)test_cases[0].inflated, strlen(test_cases[0].inflated), NULL, NULL);
+	mu_assert_streq((char *)deflated0, test_cases[0].deflated, "rz_deflate failed for 0");
+	free(deflated0);
```
ret2libc: same with loop ?

XVilka:
Also please rebase on top of the latest `dev`.

librz/util/file.c
```diff
@@ -1257,3 +1257,61 @@ RZ_API char *rz_file_path_join(const char *s1, const char *s2) {
...
 	return rz_str_newf("%s%s%s", s1, sep, s2);
 }
+
+/**
+ * \brief zip the contents of src and store in dst
+ * \param src source file (string containing filename)
+ * \param dst destination file (string containing filename)
+ * \return true, if successful; false otherwise
+ */
+RZ_API bool rz_file_deflate(RZ_NONNULL const char *src, RZ_NONNULL const char *dst) {
```
ret2libc, line 1262: I think the implementation of this function is good enough as a first implementaiton, but it requires loading the whole file in memory which imo is not ideal. I believe the zlib library has a stream API by default, so you can call inflate/deflate by providing more data as necessary. If that is so, we could use `rz_buf_new_file` to create a `RzBuffer` from the `src` file and then have a `rz_deflate_buf` function that uses `rz_buf_read` as necessary to read data from the buffer. In this way we don't need to load the whole file in memory. What do you think?

ret2libc (commented):
Nice feature to have!

librz/util/file.c
DMaroo: It doesn't work as expected, as of now. I will first need to fix that. After that, I can definitely work on using streaming behaviour instead of storing the whole file in memory.

librz/core/project.c
```diff
@@ -38,22 +42,40 @@ RZ_API RzProjectErr rz_project_save(RzCore *core, RzProject *prj, const char *fi
 }
 
 RZ_API RzProjectErr rz_project_save_file(RzCore *core, const char *file) {
+	char *tmp_file;
+	int err_mkstemp = rz_file_mkstemp("svprj", &tmp_file);
```
ret2libc: please note that in theory the security guarantees of mkstemp come from the fact that you use the file descriptor returned by it, that is `err_mkstemp` in this case. However, we would need to change few APIs to actually use directly the descriptor instead of taking a file, so... Please close at least the file descriptor returned by this function.

librz/core/project.c
```diff
@@ -63,10 +85,30 @@ RZ_API RzProject *rz_project_load_file_raw(const char *file) {
 	if (!prj) {
 		return NULL;
 	}
-	if (!sdb_text_load(prj, file)) {
+
+	char *tmp_file;
+	int err_mkstemp = rz_file_mkstemp("ldprj", &tmp_file);
```
ret2libc: same, at least close the file descriptor.

librz/include/rz_project.h
```diff
@@ -24,6 +24,8 @@ typedef enum rz_project_err {
 	RZ_PROJECT_ERR_NEWER_VERSION,
 	RZ_PROJECT_ERR_INVALID_CONTENTS,
 	RZ_PROJECT_ERR_MIGRATION_FAILED,
+	RZ_PROJECT_ERR_COMPRESSION_FAILED,
+	RZ_PROJECT_ERR_MKSTEMP_FAILED,
```
ret2libc: I don't think it makes too much sense to go that detailed in RzProjectErr about the reason of the failing... Whether we use mkstemp or something else is pretty much an internal detail. Right @thestr4ng3r ?

librz/core/project.c
```diff
@@ -38,22 +42,43 @@ RZ_API RzProjectErr rz_project_save(RzCore *core, RzProject *prj, const char *fi
 }
 
 RZ_API RzProjectErr rz_project_save_file(RzCore *core, const char *file) {
+	char *tmp_file;
+	int mkstemp_fd = rz_file_mkstemp("svprj", &tmp_file);
+
+	if (mkstemp_fd == -1 || !tmp_file) {
+		close(mkstemp_fd);
```
ret2libc, line 47: if you have to do this both in the if and outside, just put it before the `if` one time.

librz/core/project.c
```diff
@@ -63,10 +88,37 @@ RZ_API RzProject *rz_project_load_file_raw(const char *file) {
...
 		return NULL;
 	}
-	if (!sdb_text_load(prj, file)) {
-		sdb_free(prj);
+
+	char *tmp_file;
+	int mkstemp_fd = rz_file_mkstemp("ldprj", &tmp_file);
+
+	if (mkstemp_fd == -1 || !tmp_file) {
+		close(mkstemp_fd);
```
ret2libc: same

librz/util/file.c
DMaroo: I have implemented this in my last commit. Please take a look. However, this breaks `rz_file_inflate`. I don't know of any way to have streaming behaviour in `rz_file_inflate`, and just using the current implementation gives `RZ_ERR_INVALID_CONTENTS` error. However, when I manually decompress (using `gzip`) the file generated by `rz_file_inflate` and open that with rizin, it works perfectly fine.

librz/include/rz_project.h
XVilka: Yes, I suggest to remove MKSTEMP error

librz/core/project.c
```diff
@@ -38,22 +42,42 @@ RZ_API RzProjectErr rz_project_save(RzCore *core, RzProject *prj, const char *fi
 }
 
 RZ_API RzProjectErr rz_project_save_file(RzCore *core, const char *file) {
+	char *tmp_file;
+	int mkstemp_fd = rz_file_mkstemp("svprj", &tmp_file);
+	close(mkstemp_fd);
+
+	if (mkstemp_fd == -1 || !tmp_file) {
+		return RZ_PROJECT_ERR_MKSTEMP_FAILED;
```
XVilka: We could return ` RZ_PROJECT_ERR_FILE` here instead.

XVilka (changes requested):

XVilka:
Project loading tests still do not work:
```
87/88 cpu_platform_profiles          FAIL            0.48s   exit status 1
88/88 open_analyse_save_load_project FAIL            1.88s   exit status 1


The output from the failed tests:

55/88 project_migrate                FAIL            0.35s   killed by signal 11 SIGSEGV
--- command ---
04:42:21 /home/runner/work/rizin/rizin/build/test/unit/test_project_migrate
-------

87/88 cpu_platform_profiles          FAIL            0.48s   exit status 1
--- command ---
04:42:23 /home/runner/work/rizin/rizin/build/test/integration/test_cpu_platform_profiles
--- stdout ---
test_cpu_profiles ERR
[XX] Fail at line 53: project load err: expected 0, got 5.

test_platform_profiles ERR
[XX] Fail at line 104: project load err: expected 0, got 5.

--- stderr ---
Cannot find base type "uid_t"
Cannot find base type "pid_t"
Cannot find base type "uid_t"
Cannot find base type "uid_t"
Cannot find base type "gid_t"
Cannot find base type "pid_t"
WARNING: Neither hash nor gnu_hash exist. Falling back to heuristics for deducing the number of dynamic symbols...
WARNING: Neither hash nor gnu_hash exist. Falling back to heuristics for deducing the number of dynamic symbols...
WARNING: Neither hash nor gnu_hash exist. Falling back to heuristics for deducing the number of dynamic symbols...
-------

88/88 open_analyse_save_load_project FAIL            1.88s   exit status 1
--- command ---
04:42:23 /home/runner/work/rizin/rizin/build/test/integration/test_open_analyse_save_load_project
--- stdout ---
test_open_analyse_save ERR
[XX] Fail at line 58: project load err: expected 0, got 5.

--- stderr ---
[ ] Analyze function calls (aac)
[
[x] Analyze function calls (aac)
[ ] Analyze len bytes of instructions for references (aar)
[
[x] Analyze len bytes of instructions for references (aar)
[ ] Check for classes
[
[x] Check for classes
[ ] Type matching analysis for all functions (aaft)
[
[x] Type matching analysis for all functions (aaft)
[ ] Propagate noreturn information
[
[x] Propagate noreturn information
[ ] Use -AA or aaaa to perform additional experimental analysis.
[
[x] Use -AA or aaaa to perform additional experimental analysis.
functions count = 146
-------


Summary of Failures:

55/88 project_migrate                FAIL            0.35s   killed by signal 11 SIGSEGV
87/88 cpu_platform_profiles          FAIL            0.48s   exit status 1
88/88 open_analyse_save_load_project FAIL            1.88s   exit status 1


Ok:                 85  
Expected Fail:      0   
Fail:               3   
Unexpected Pass:    0   
Skipped:            0   
Timeout:            0   
```

DMaroo:
Yes, I am aware that the tests are failing. I will have a look at that once I finish implementing `rz_file_inflate`.

librz/util/file.c
```diff
@@ -1257,3 +1257,92 @@ RZ_API char *rz_file_path_join(const char *s1, const char *s2) {
...
+	rz_return_val_if_fail(src && dst, false);
+
+	RzBuffer *src_buf = rz_buf_new_file(src, O_RDONLY, 0);
+	RzBuffer *dst_buf = rz_buf_new_file(dst, O_WRONLY | O_CREAT, 0644);
+
+	if (!(src_buf && dst_buf)) {
+		free(src_buf);
+		free(dst_buf);
```
ret2libc:
```suggestion
		rz_buf_free(src_buf);
		rz_buf_free(dst_buf);
```

librz/util/file.c
ret2libc: Mmm I'm not sure it works that way. I think you should add `rz_deflate_buf` and in that function you should call `deflateInit2` only one time and then `deflate` multiple times.

Right not it seems to me you just call `rz_deflate` with multiple chunks, but `rz_deflate` essentially starts from scratch with every call. I hope it's clear what I mean :)

ret2libc: I propose to add the APIs:
```C
RZ_API RzBuffer *rz_inflate_buf(RzBuffer *src);
RZ_API RzBuffer *rz_deflate_buf(RzBuffer *src);
```

DMaroo:
A few questions:
 1. `rz_deflate` and `rz_deflatew` are no longer being used anywhere. Should I remove them, or let them be for future convenience?
 2. Should I write any tests for `rz_buf_deflate`?

DMaroo:
I have now done the majority of work for this PR. What remains is only cleaning out unused stuff (if required) and making the CI green. I am now converting this draft to a ready for review PR.

librz/util/zip.c
```diff
@@ -266,3 +266,79 @@ RZ_API bool rz_deflatew_buf(RZ_NONNULL RzBuffer *src, RZ_NONNULL RzBuffer *dst,
...
+
+	stream.zalloc = Z_NULL;
+	stream.zfree = Z_NULL;
+	stream.opaque = Z_NULL;
+
+	if (inflateInit2(&stream, wbits) != Z_OK) {
+		return false;
+	}
+
+	int comp_factor = 1032; // maximum compression ratio
```
DMaroo, line 306: As of now, I just prepare for the highest possible compression ratio and assign memory accordingly, but this forces us not choose a large enough `block_size`. So, we either need to go with a small `block_size` (right now, I am using `1 << 13 = 8 KB` (https://github.com/rizinorg/rizin/pull/1446/commits/b3fa0d5630f65a0deab3d1d08c75e34e845485aa#diff-b6e47a6b965ef83d50289bc0568bb5f82a4e47d5fe654cc6be44a2c85c7b0ce8R1309), which is kind of small for large files, but still amounts to 9 MB memory) and subsequently more number of inflate loops, or go with a larger block size, thus larger memory usage. If there's any better way to handle this, please feel free to suggest.

DMaroo:
Also, I haven't profiled this. So I am not sure what is the performance effect on saving huge projects. It would be nice if someone could profile this.

librz/util/zip.c
```diff
@@ -107,4 +110,235 @@ RZ_API ut8 *rz_inflatew(const ut8 *src, int srcLen, int *srcConsumed, int *dstLe
...
+ * \brief deflate data contained in a RzBuffer using zlib
+ * \param src source buffer
+ * \param dst destination buffer
+ * \param block_size block sizes to use while deflating data
+ * \param src_consumed consumed source buffer length
+ * \param dst_len deflated buffer length
+ * \param wbits the size of the history buffer (or “window size”), and what header and trailer format is expected.
+ * \return true if successful; false otherwise
+ */
+RZ_API bool rz_deflatew_buf(RZ_NONNULL RzBuffer *src, RZ_NONNULL RzBuffer *dst, ut64 block_size, ut8 *src_consumed, ut8 *dst_len, int wbits) {
```
ret2libc: no need for dst_len, because RzBuffer already has the method rz_buf_size to compute the size.

librz/util/zip.c
```diff
@@ -107,4 +110,235 @@ RZ_API ut8 *rz_inflatew(const ut8 *src, int srcLen, int *srcConsumed, int *dstLe
...
+	rz_return_val_if_fail(src && dst, false);
```
ret2libc, line 212: add:
```
rz_return_val_if_fail(block_size != 0, false);
```

ret2libc (commented):
It would be great to have unit tests for rz_inflate/deflate_buf functions, yes!

librz/util/zip.c
DMaroo: Wouldn't it be consistent to have this parameter (compared to `rz_deflate`)? Just asking.

DMaroo:
@ret2libc Would adding tests for `rz_file_inflate` and `rz_file_deflate` suffice (since they are only a wrapper around `rz_inflate_buf` and `rz_deflate_buf`, and it would be easier to write and manage tests for `rz_file*` functions, I think)?

DMaroo:
I am unable to figure out why the `project_migrate` test is failing.

librz/core/project.c
```diff
@@ -63,10 +85,41 @@ RZ_API RzProject *rz_project_load_file_raw(const char *file) {
...
+			prj = NULL;
+			goto return_goto;
+		}
+	} else {
+		if (!sdb_text_load(prj, file)) {
+			sdb_free(prj);
+			prj = NULL;
+		}
+		goto return_goto;
+	}
```
ret2libc: this code is dupped. Rewrite the code so that you just have to switch the argument of `sdb_text_load` instead of rewriting the whole piece of code.

librz/util/zip.c
```diff
@@ -107,4 +109,223 @@ RZ_API ut8 *rz_inflatew(const ut8 *src, int srcLen, int *srcConsumed, int *dstLe
...
+/**
+ * \brief compress/deflate data to zlib or gzip
+ * \param src source uncompressed bytes
+ * \param srcLen source bytes length
+ * \param srcConsumed consumed source bytes length
+ * \param dstLen compressed bytes length
+ * \param wbits the size of the history buffer (or “window size”), and what header and trailer format is expected.
+ * \return ptr to compressed
+ */
+RZ_API ut8 *rz_deflatew(RZ_NONNULL const ut8 *src, int srcLen, int *srcConsumed, int *dstLen, int wbits) {
```
ret2libc, line 133: remove this and rz_inflate and rz_inflatew and rz_deflate. Let's just use the rz_(de|in)flate_buf versions which do not require to read the whole file in mem.

test/unit/test_inflate_deflate.c
```diff
@@ -0,0 +1,78 @@
...
+	{ "1234567890abcdefghijklmnopqrstuvwxyz\n", "\x1f\x8b\x08\x00\x00\x00\x00\x00\x00\x03\x33\x34\x32\x36\x31\x35\x33\xb7\xb0\x34\x48\x4c\x4a\x4e\x49\x4d\x4b\xcf\xc8\xcc\xca\xce\xc9\xcd\xcb\x2f\x28\x2c\x2a\x2e\x29\x2d\x2b\xaf\xa8\xac\xe2\x02\x00\x9f\x79\xdc\xdb\x25\x00\x00\x00" }
+};
+
+int deflated_lengths[] = { 284, 637, 57 };
```
ret2libc: How did you compute these sizes? Can we get them with `sizeof` or `strlen` instead?

librz/util/file.c
```diff
@@ -1252,3 +1252,80 @@ RZ_API char *rz_file_path_join(const char *s1, const char *s2) {
...
+
+	rz_buf_free(src_buf);
+	rz_buf_free(dst_buf);
+	return true;
+
+error_route:
+	rz_buf_free(src_buf);
+	rz_buf_free(dst_buf);
+	return false;
+}
```
ret2libc: remove this and place `error_route` above. Add a `bool res` variable which is set to false before the goto. This way we don't have to duplicate the code and it's less opportunities for us to forget freeing something.

librz/util/file.c
```diff
@@ -1252,3 +1252,80 @@ RZ_API char *rz_file_path_join(const char *s1, const char *s2) {
...
```
ret2libc: same

test/unit/test_inflate_deflate.c
DMaroo: `strlen` would stop counting at null bytes. `sizeof` should work fine. I'll try.

librz/util/zip.c
DMaroo: That'll require a few changes in `analysis/flirt.c`, `core/cmd_print.c`, `io/p/io_gzip.c` and `util/file.c` in the `librz` (al the places where `rz_inflate` is being used).

wargio:
> A few questions:
> 
> 1. `rz_deflate` and `rz_deflatew` are no longer being used anywhere. Should I remove them, or let them be for future convenience?
> 2. Should I write any tests for `rz_buf_deflate`?

@XVilka we need to test this on flirt signatures, before merging

test/unit/test_inflate_deflate.c
DMaroo: I cannot seem to use `sizeof` since it would require an array, and I cannot use `char inflated[]; char deflated[];` in the struct.

librz/util/zip.c
ret2libc: Didn't you say that some functions were never used? What are those functions never used?

librz/util/zip.c
```diff
@@ -107,4 +109,223 @@ RZ_API ut8 *rz_inflatew(const ut8 *src, int srcLen, int *srcConsumed, int *dstLe
...
+		stream.next_out = dst_tmpbuf;
+		stream.avail_out = block_size;
+		stream.total_out = 0;
+
+		if (src_readlen < block_size) {
+			flush = Z_FINISH;
+		}
+		err = deflate(&stream, flush);
+		if (err < 0) {
+			RZ_LOG_ERROR("deflate error: %d %s\n", err, gzerr(-err));
```
ret2libc, line 246: shall we `break` if there's an error?

librz/util/zip.c
```diff
@@ -107,4 +109,223 @@ RZ_API ut8 *rz_inflatew(const ut8 *src, int srcLen, int *srcConsumed, int *dstLe
...
+		stream.next_out = dst_tmpbuf;
+		stream.avail_out = comp_factor * block_size;
+		stream.total_out = 0;
+
+		if (src_readlen < block_size) {
+			flush = Z_FINISH;
+		}
+		err = inflate(&stream, flush);
+		if (err < 0) {
+			RZ_LOG_ERROR("inflate error: %d %s\n", err, gzerr(-err));
```
ret2libc, line 322: shall we `break`?

test/unit/test_inflate_deflate.c
```diff
@@ -0,0 +1,99 @@
...
+	{ "1234567890abcdefghijklmnopqrstuvwxyz\n", "\x1f\x8b\x08\x00\x00\x00\x00\x00\x00\x03\x33\x34\x32\x36\x31\x35\x33\xb7\xb0\x34\x48\x4c\x4a\x4e\x49\x4d\x4b\xcf\xc8\xcc\xca\xce\xc9\xcd\xcb\x2f\x28\x2c\x2a\x2e\x29\x2d\x2b\xaf\xa8\xac\xe2\x02\x00\x9f\x79\xdc\xdb\x25\x00\x00\x00" }
+};
+
+int deflated_lengths[3] = { 284, 637, 57 };
+int inflated_lengths[3];
+
```
ret2libc: just add a `deflated_length` field in the test_cases above.

test/unit/test_inflate_deflate.c
```diff
@@ -0,0 +1,99 @@
...
+void init_lengths()
+{
+	for (int i = 0; i < 3; i++)
+	{
+		inflated_lengths[i] = strlen(test_cases[i].inflated);
+	}
+}
+
```
ret2libc: i think there's no need to compute the inflated_lengths, given they are strings. Just use `mu_assert_streq` in the tests and you are automatically testing also the length of the strings.

ret2libc (changes requested):
Just few minor things, but I think code looks good! Good job and thanks :)

librz/util/zip.c
DMaroo: Yes, I think we should. zlib specifies negative return codes for `deflate`/`inflate` as errors. I can't think of a case where continuing after getting an error would be helpful.

test/unit/test_inflate_deflate.c
DMaroo: We will still need the `inflated_lengths` to pass them to `rz_deflate` and `rz_buf_new_with_bytes`. However, I'll change `mu_assert_memeq` to `mu_assert_streq` for the inflated strings.

ret2libc: You can still just use `strlen(test_case[i].inflated_string)` instead.

DMaroo: Okay, but that felt cumbersome to me, which is why I removed it in the first place. I'll update it now, though.

librz/util/zip.c
```diff
@@ -107,4 +109,231 @@ RZ_API ut8 *rz_inflatew(const ut8 *src, int srcLen, int *srcConsumed, int *dstLe
...
+			inflateEnd(&stream);
```
ret2libc: as in other places, please use `goto`. Otherwise it becomes too easy to forget about stuff that needs to be freed. Do the same in rz_deflatew_buf

ret2libc (changes requested):

ret2libc (approved):

librz/util/file.c
```diff
@@ -1252,3 +1252,80 @@ RZ_API char *rz_file_path_join(const char *s1, const char *s2) {
...
+
+/**
+ * \brief check whether a file is a deflated (gzip) file
+ * \param src source file (string containing filename)
+ * \return true, if src is a deflated (gzip) file; false otherwise
+ */
+RZ_API bool rz_file_is_deflated(RZ_NONNULL const char *src) {
+	rz_return_val_if_fail(src, false);
+
+	unsigned char *header = (unsigned char *)rz_file_slurp_range(src, 0, 3, NULL);
```
XVilka, line 1329: Please check the returned length. If `NULL` or less than 3 bytes returned.

librz/util/zip.c
XVilka: @ret2libc @DMaroo wait. Lets keep them. FLIRT is tricky to test, and it *does require* wbits option. It has the raw data without the header.

XVilka (changes requested):
Good job! One minor thing and it's good to go.

librz/util/zip.c
XVilka: Ah, nevermind, they indeed seem to be unused:
```
[i] ℤ rg "rz_inflatew"                                                                                                                                                                                                            17:33:25 
librz/util/zip.c
28: * @see rz_inflatew()
31:     return rz_inflatew(src, srcLen, srcConsumed, dstLen, MAX_WBITS + 32);
36: * @see rz_inflatew()
39:     return rz_inflatew(src, srcLen, srcConsumed, dstLen, -MAX_WBITS);
51:RZ_API ut8 *rz_inflatew(const ut8 *src, int srcLen, int *srcConsumed, int *dstLen, int wbits) {

librz/include/rz_util/rz_file.h
46:RZ_API ut8 *rz_inflatew(const ut8 *src, int srcLen, int *srcConsumed, int *dstLen, int wbits);
rizin on  dev [$] 
[i] ℤ rg "rz_deflatew"   
```
Kill them.

XVilka: The `rz_inflate` is being used:
```
[i] ℤ rg "rz_inflate("                                                                                                                                                                                                             13:17:35 
librz/util/file.c
502:    ut8 *out = rz_inflate(in, (int)sz, NULL, outlen);

librz/io/p/io_gzip.c
169:            mal->buf = rz_inflate(data, (int)len, NULL, size);

librz/core/cmd_print.c
6183:                           out = rz_inflate(block, core->blocksize, &inConsumed, &outlen);
6191:                           out = rz_inflate(block, core->blocksize, NULL, &outlen);
6198:                           out = rz_inflate(block, core->blocksize, NULL, &outlen);

librz/analysis/flirt.c
1395:                   if (!(decompressed_buf = rz_inflate(buf, size, NULL, &decompressed_size))) {
```

XVilka (approved):

thestr4ng3r (changes requested):
Compression should be optional and disabled by default. I suggest adding a config var `prj.compress`.

DMaroo:
> Compression should be optional and disabled by default. I suggest adding a config var `prj.compress`.

Alright, makes sense. But why disable it by default. Shouldn't it be enabled by default, given that it is not particularly buggy or slow, and makes the files quite smaller?

UPD: I have pushed a commit which does what you suggested (except it decompresses by default, which can be trivially changed later if needed). It works well when changing options, and no duplicates are created (as I understand from my manual testing).

XVilka (approved):

librz/core/cconfig.c
```diff
@@ -3146,6 +3146,7 @@ RZ_API int rz_core_config_init(RzCore *core) {
 
 	/* prj */
 	SETPREF("prj.file", "", "Path of the currently opened project");
+	SETBPREF("prj.decompress", "true", "Decompress the project file while saving");
```
XVilka: I think it should be `prj.compress` instead, and `Compress the project file` instead

XVilka (changes requested):

librz/core/cmd_project.c
```diff
@@ -26,13 +26,14 @@ RZ_IPI RzCmdStatus rz_project_save_handler(RzCore *core, int argc, const char **
...
 			eprintf("There is no project file associated with the current session yet.\n"
-				"Specify the file explitily as `Ps <file.rzdb>` or set it manually with `e prj.file=<project-path>`.\n");
+				"Specify the file explicitly as `Ps <file.rzdb>` or set it manually with `e prj.file=<project-path>`.\n");
 			return RZ_CMD_STATUS_ERROR;
 		}
 	} else { // argc == 2 checked by the shell
 		file = argv[1];
 	}
-	RzProjectErr err = rz_project_save_file(core, file);
+	bool decompress = rz_config_get_b(core->config, "prj.compress");
```
XVilka: Should be `compress` as well then.

XVilka (changes requested):

librz/core/cmd_project.c
XVilka: See also other places with the same name problem.

XVilka (approved):

librz/core/project.c
```diff
@@ -37,7 +39,20 @@ RZ_API RzProjectErr rz_project_save(RzCore *core, RzProject *prj, const char *fi
 	return RZ_PROJECT_ERR_SUCCESS;
 }
 
-RZ_API RzProjectErr rz_project_save_file(RzCore *core, const char *file) {
+RZ_API RzProjectErr rz_project_save_file(RzCore *core, const char *file, bool compress) {
+
```
ret2libc:
```suggestion
```

librz/core/project.c
```diff
@@ -37,7 +39,20 @@ RZ_API RzProjectErr rz_project_save(RzCore *core, RzProject *prj, const char *fi
...
+	char *tmp_file;
```
ret2libc:
```suggestion
	char *tmp_file = NULL;
```

librz/core/project.c
```diff
@@ -47,13 +62,32 @@ RZ_API RzProjectErr rz_project_save_file(RzCore *core, const char *file) {
...
+	if (err != RZ_PROJECT_ERR_SUCCESS) {
+		goto tmp_file_err;
+	}
+
+	if (!rz_file_deflate(tmp_file, file)) {
+		err = RZ_PROJECT_ERR_COMPRESSION_FAILED;
+		goto tmp_file_err;
+	}
+
+	rz_config_set(core->config, "prj.file", file);
```
ret2libc:
```suggestion
	if (err != RZ_PROJECT_ERR_SUCCESS) {
		goto tmp_file_err;
	}
	
	if (compress && !rz_file_deflate(tmp_file, file)) {
		err = RZ_PROJECT_ERR_COMPRESSION_FAILED;
		goto tmp_file_err;
	}

	rz_config_set(core->config, "prj.file", file);
```
something like this.

ret2libc (commented):
I don't see particular problems with compressing by default. Actually, on a project of a chrome binary analysis it saved quite a lot (something like ~1.4GB vs less than 200MB).

librz/core/project.c
DMaroo: Thanks. This makes it much simpler.

XVilka (approved):

PR #1459, opened 2021-08-16
Fix some of the RzType memory leaks

XVilka:

**Detailed description**

Fixed some of the memory leaks reported by Coverity

**Test plan**

CI is green

### DIFF
--- a/librz/analysis/fcn.c
+++ b/librz/analysis/fcn.c
@@ -2441,7 +2441,7 @@ RZ_API RZ_OWN RzPVector *rz_analysis_function_args(RzAnalysis *a, RzAnalysisFunc
 			// where each var is assigned at its argnum
 			if (argnum >= rz_pvector_len(args)) {
 				if (!rz_pvector_reserve(args, argnum + 1)) {
-					return args;
+					goto cleanup;
 				}
 				while (argnum >= rz_pvector_len(args)) {
 					rz_pvector_push(args, NULL);
@@ -2451,6 +2451,7 @@ RZ_API RZ_OWN RzPVector *rz_analysis_function_args(RzAnalysis *a, RzAnalysisFunc
 			fcn->argnum++;
 		}
 	}
+cleanup:
 	rz_pvector_free(tmp);
 	return args;
 }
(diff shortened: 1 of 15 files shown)


librz/analysis/fcn.c
```diff
@@ -2441,6 +2441,7 @@ RZ_API RZ_OWN RzPVector *rz_analysis_function_args(RzAnalysis *a, RzAnalysisFunc
 			// where each var is assigned at its argnum
 			if (argnum >= rz_pvector_len(args)) {
 				if (!rz_pvector_reserve(args, argnum + 1)) {
+					rz_pvector_free(tmp);
```
ret2libc: this is ok, though probably `goto err` with err being at line 2454 would be less error prone

librz/analysis/var.c
```diff
@@ -104,6 +105,7 @@ static void resolve_var_overlaps(RzAnalysisVar *var) {
 			rz_analysis_var_delete(other);
 		}
 	}
+	rz_pvector_free(cloned_vars);
```
ret2libc, line 108: If var->fcn->vars is defined with a free function, this will free the original items of the vector as well, I think. Right @thestr4ng3r ?

XVilka: It doesn't, otherwise, it would fail on my tests. I am sure this code was called multiple times during the `aaa`/`aaft` tests

librz/analysis/fcn.c
XVilka: I plan to do another sweep for memory leaks. I can change this tomorrow then. If there are no objections regarding the semantics, should be fine to merge then.

XVilka: Nevermind, fixed.

librz/core/cmd_type.c
```diff
@@ -117,6 +117,7 @@ static RzCmdStatus type_format_print(RzCore *core, const char *type, ut64 addres
 	char *fmt = rz_type_format(core->analysis->typedb, type);
 	if (RZ_STR_ISEMPTY(fmt)) {
 		eprintf("Cannot find type %s\n", type);
```
wargio, line 119: Please @XVilka in the next PR change these to RZ_LOG_XXX

wargio (approved):

librz/analysis/var.c
ret2libc: I'm not 100% convinced... also, `RzPVector *cloned_vars = (RzPVector *)rz_vector_clone((RzVector *)&var->fcn->vars);` seems wrong, because rz_vector_clone returns a RzVector. Why does it need the cast? Why casting a PVector to a Vector?

XVilka: I copied the code from the one written by the @thestr4ng3r

PR #1488, opened 2021-08-19
Use PJ in `config_print_node()`

officialcjunior:

**Detailed description**

This is PR is to use PJ in `config_print_node`.
* Rename rz_config_list to rz_core_config_print_all
* Use RzOutputMode in rz_core_config_print_all
* Move all of these functions to `cconfig.c` and remove `cb_printf` from `RzConfig`

There are many things left to do. Here's a list of things that we are definitely going to cover:
- [x] Use `RzOutputMode` for handling the JSON and verbose stuff.
- [x] The function is being called at two other places. Needs to find out which those commands are and fix them.
- [x] Maybe remove the`cb_printf` usage by moving `rz_config_list()` from `librz/config/config.c` to `librz/core/cconfig.c` and rename it to `rz_config_print()` instead.
- [x] Add tests

**Test plan**

Tests for all commands under `el` has been added.

**Closing issues**

...


librz/config/config.c
```diff
@@ -174,7 +178,7 @@ RZ_API void rz_config_list(RzConfig *cfg, const char *str, int rad) {
 	case 0:
 		rz_list_foreach (cfg->nodes, iter, node) {
 			if (!str || (str && (!strncmp(str, node->name, len)))) {
-				config_print_node(cfg, node, pfx, sfx, verbose, json);
+			config_print_node(cfg, pfx, sfx, verbose, false);
```
p371jah: Something wrong with that line

librz/config/config.c
```diff
@@ -67,44 +67,49 @@ static void config_print_value_json(RzConfig *cfg, RzConfigNode *node) {
...
 				}
-				cfg->cb_printf("]");
+				pj_end(pj);
 			}
-			cfg->cb_printf("}");
+			pj_end(pj);
+			cfg->cb_printf("%s\n", pj_string (pj));
+			pj_free (pj);			
 		} else {
 			cfg->cb_printf("\"%s\":", node->name);
```
p371jah: `pj` leaks here

officialcjunior:
I believe that checks like [these](https://github.com/rizinorg/rizin/blob/dev/librz/config/config.c#L183) is what's making passing things as argument to `e` work. See the test that is broken now.

I'm not sure how to handle that now. I had made the iteration through `cfg->nodes` happen inside `config_print_node()` to make the JSON part straightforward.

librz/core/cconfig.c
```diff
@@ -1731,6 +1731,244 @@ RZ_API bool rz_core_esil_cmd(RzAnalysisEsil *esil, const char *cmd, ut64 a1, ut6
...
+			rz_cons_printf("]");
+		}
+		rz_cons_println("");
+	} else if (mode == RZ_OUTPUT_MODE_QUIET) {
+		rz_cons_printf("%s%s = %s%s\n", pfx,
+			node->name, node->value, sfx);
+	}
+}
+
+RZ_API void rz_config_list(RzConfig *cfg, const char *str, int rad) {
```
XVilka: I suggest to rename this function as well, since we moved it from outside the `librz/config`. `rz_core_config_print_all()` or something like that. Please also use `RzOutputMode mode` instead of `int rad`.

librz/include/rz_config.h
```diff
@@ -76,7 +75,7 @@ RZ_API RzConfig *rz_config_new(void *user);
 RZ_API RzConfig *rz_config_clone(RzConfig *cfg);
 RZ_API void rz_config_free(RzConfig *cfg);
 RZ_API void rz_config_lock(RzConfig *cfg, int l);
-RZ_API bool rz_config_eval(RzConfig *cfg, const char *str, bool many);
+RZ_API bool rz_core_config_eval_and_print(RzConfig *cfg, const char *str, bool many);
```
XVilka: It should be in the `librz/include/rz_core.h` then.

XVilka: Also it should get the `RzCore` instead of the `RzConfig`

XVilka (changes requested):

librz/core/cconfig.c
officialcjunior: White using `RzOutputMode`, what is the equivalent output mode for the case 's', which lists the subspaces?
https://github.com/rizinorg/rizin/blob/dev/librz/config/config.c#L191

Everything except `RZ_OUTPUT_MODE_TABLE` and `RZ_OUTPUT_MODE_SDB` are used - see `rz_eval_list_handler`

librz/include/rz_config.h
```diff
@@ -89,7 +87,7 @@ RZ_API bool rz_config_get_b(RzConfig *cfg, const char *name);
 RZ_API const char *rz_config_get(RzConfig *cfg, const char *name);
 RZ_API const char *rz_config_desc(RzConfig *cfg, const char *name, const char *desc);
 RZ_API const char *rz_config_node_desc(RzConfigNode *node, const char *desc);
-RZ_API void rz_config_list(RzConfig *cfg, const char *str, int rad);
+RZ_API void rz_core_config_print_all(RzConfig *cfg, const char *str, int rad);
```
XVilka: You forgot to move this to `librz/inlude/rz_core.h`

XVilka (changes requested):
@officialcjunior looks like `es` corresponds to the `RZ_OUTPUT_MODE_SHORT` which is not implemented. I suggest to add it into the `RzOutputMode` enum. cc @ret2libc @wargio what do you think?

ret2libc:
> @officialcjunior looks like `es` corresponds to the `RZ_OUTPUT_MODE_SHORT` which is not implemented. I suggest to add it into the `RzOutputMode` enum. cc @ret2libc @wargio what do you think?

Mmm I think `es` is just a separate command.

```
[0x00006b60]> es?
Usage: es [<key>]   # List all config variable spaces or sub-keys/sub-spaces if a <key> is provided
```

It lists config variable spaces.

XVilka:
@ret2libc good idea, lets just make it a separate command @officialcjunior

ret2libc:
> @ret2libc good idea, lets just make it a separate command @officialcjunior

I'm not sure I understand @XVilka . `es` is *already* a separate command.
```
[0x00006b60]> e?
Usage: e[?]   # List/get/set config evaluable vars
| e <key>[=<val|?>] [<key>[=<val|?>] ...]] # Get/Set value of config variable <key>
| el[j*qlJ] [<key>]      # List config variables with their descriptions
| e-                     # Reset config variables
| e! <key>               # Invert the boolean value of config variable <key>
| ec[?]                  # Set color for given key (prompt, offset, ...)
| ee <key>               # Open editor to change the value of config variable <key>
| er <key>               # Set config variable <key> as read-only
| es [<key>]             # List all config variable spaces or sub-keys/sub-spaces if a <key> is provided
| et <key>               # Show type of given config variable <key>
| env [<varname>[=<varvalue>]] # Get/set environment variables
```

XVilka:
@ret2libc I meant separate function handler, sorry.

librz/core/cconfig.c
```diff
@@ -1731,6 +1731,201 @@ RZ_API bool rz_core_esil_cmd(RzAnalysisEsil *esil, const char *cmd, ut64 a1, ut6
...
+			str = NULL;
+			len = 0;
+		}
+	}
+
+	switch (mode) {
+	case RZ_OUTPUT_MODE_RIZIN:
+		pfx = "\"e ";
+		sfx = "\"";
+	/* fallthrou */
```
XVilka: `fallthrough`

XVilka (changes requested):
There is one broken test also:
```
[XX] db/cmd/cmd_eval e-dot
RZ_NOPLUGINS=1 rizin -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -Qc 'e zoom.' =
-- stdout
--- expected
+++ actual
@@ -1,5 +1,0 @@
-zoom.byte = h
-zoom.from = 0
-zoom.in = io.map
-zoom.maxsz = 512
-zoom.to = 0
```

librz/core/cmd_eval.c
```diff
@@ -502,14 +502,14 @@ RZ_IPI int rz_cmd_eval(void *data, const char *input) {
 	RzCore *core = (RzCore *)data;
 	switch (input[0]) {
 	case '\0': // "e"
-		rz_config_list(core->config, NULL, 0);
```
XVilka, line 505: @officialcjunior @ret2libc I think we can safely remove this `rz_cmd_eval()` function since the `e` commands were completely ported to the newshell, and there is no `RZ_CMD_DESC_OLDINPUT` for `e` commands in `librz/core/cmd_descs/cmd_descs.yaml`.

officialcjunior: Yes, I think it's about time that function can be removed. 👍🏽 
Will wait for what ret2libc has to say.

officialcjunior: Removed.

test/db/cmd/cmd_eval
```diff
@@ -106,3 +106,33 @@ scr.highlight = "e cmd.prompt =  "e scr.highlight = cmd.prompt ="e scr.rainbow =
...
+elq zoom
+EOF
+EXPECT=<<EOF
+{"zoom.byte":"h","zoom.from":0,"zoom.in":"io.map","zoom.maxsz":512,"zoom.to":0}
+[{"name":"zoom.byte","value":"h","type":"str","desc":"Zoom callback to calculate each byte (See pz? for help)","ro":false},{"name":"zoom.from","value":0,"type":"int","desc":"Zoom start address","ro":false},{"name":"zoom.in","value":"io.map","type":"str","desc":"Specify  boundaries for zoom","ro":false,"options":["raw","block","bin.section","bin.sections","bin.sections.rwx","bin.sections.r","bin.sections.rw","bin.sections.rx","bin.sections.wx","bin.sections.x","io.map","io.maps","io.maps.rwx","io.maps.r","io.maps.rw","io.maps.rx","io.maps.wx","io.maps.x","dbg.stack","dbg.heap","dbg.map","dbg.maps","dbg.maps.rwx","dbg.maps.r","dbg.maps.rw","dbg.maps.rx","dbg.maps.wx","dbg.maps.x","analysis.fcn","analysis.bb"]},{"name":"zoom.maxsz","value":512,"type":"int","desc":"Zoom max size of block","ro":false},{"name":"zoom.to","value":0,"type":"int","desc":"Zoom end address","ro":false}]
+"e zoom.byte = h"
+"e zoom.from = 0"
+"e zoom.in = io.map"
+"e zoom.maxsz = 512"
+"e zoom.to = 0"
```
ret2libc: These should be:
```suggestion
e zoom.byte=h
e zoom.from=0
e zoom.in=io.map
e zoom.maxsz=512
e zoom.to=0
```

The string after the `=` should probably be the result of `rz_cmd_escape_arg` with `RZ_CMD_ESCAPE_ONE_ARG`.

librz/core/cconfig.c
```diff
@@ -1737,6 +1737,219 @@ RZ_API bool rz_core_esil_cmd(RzAnalysisEsil *esil, const char *cmd, ut64 a1, ut6
...
+			len--;
+		}
+		if (len > 0 && str[0] == ' ') {
+			str++;
+			len--;
+		}
+		if (strlen(str) == 0) {
+			str = NULL;
+			len = 0;
+		}
```
ret2libc: This should not be necessary.

Trimming the argument, dealing with `j` suffix or whatnot should be done by the parser/caller.

librz/core/cconfig.c
```diff
@@ -1737,6 +1737,219 @@ RZ_API bool rz_core_esil_cmd(RzAnalysisEsil *esil, const char *cmd, ut64 a1, ut6
...
+	}
+
+	switch (mode) {
+	case RZ_OUTPUT_MODE_RIZIN:
+		pfx = "\"e ";
+		sfx = "\"";
```
ret2libc: This needs to be fixed, otherwise it won't work. If you wrap a command in `"` it goes through the old handlers.

librz/core/cconfig.c
```diff
@@ -1737,6 +1737,219 @@ RZ_API bool rz_core_esil_cmd(RzAnalysisEsil *esil, const char *cmd, ut64 a1, ut6
...
+		rz_list_foreach (cfg->nodes, iter, node) {
+			if (!str || (str && (!strncmp(str, node->name, len)))) {
+				config_print_node(cfg, node, pfx, sfx, NULL, RZ_OUTPUT_MODE_QUIET);
+			}
+		}
+		break;
+	case RZ_OUTPUT_MODE_STANDARD:
+		rz_list_foreach (cfg->nodes, iter, node) {
+			if (!str || (str && (!strncmp(str, node->name, len)))) {
+				if (!str || !strncmp(str, node->name, len)) {
```
ret2libc: ?

librz/core/cconfig.c
```diff
@@ -1737,6 +1737,219 @@ RZ_API bool rz_core_esil_cmd(RzAnalysisEsil *esil, const char *cmd, ut64 a1, ut6
...
+			str++;
+			len--;
+		}
+		if (strlen(str) == 0) {
+			str = NULL;
+			len = 0;
+		}
+	}
+
+	switch (mode) {
```
ret2libc: I suggest to move the rz_list_foreach outside the switch, remove the switch and just call `config_print_node` with `mode`. Inside `config_print_node` print the right thing based on the `mode`.

librz/core/cmd_eval.c
```diff
@@ -678,23 +520,23 @@ RZ_IPI RzCmdStatus rz_eval_list_handler(RzCore *core, int argc, const char **arg
 	const char *arg = argc > 1 ? argv[1] : "";
 	switch (mode) {
 	case RZ_OUTPUT_MODE_STANDARD:
-		rz_config_list(core->config, arg, 2);
+		rz_core_config_print_all(core->config, arg, RZ_OUTPUT_MODE_STANDARD);
```
ret2libc: In theory all this switch should not be necessary... You just want do to `rz_core_config_print_all(core->config, arg, mode);`

ret2libc: Also, it may make sense to use RZ_CMD_DESC_TYPE_ARGV_STATE for this handler and pass `state` around instead of just `mode`.

librz/core/cmd_eval.c
```diff
@@ -741,7 +583,43 @@ RZ_IPI RzCmdStatus rz_eval_readonly_handler(RzCore *core, int argc, const char *
 
 RZ_IPI RzCmdStatus rz_eval_spaces_handler(RzCore *core, int argc, const char **argv) {
 	const char *arg = argc > 1 ? argv[1] : "";
-	rz_config_list(core->config, arg, 's');
+	RzConfigNode *node;
+	RzListIter *iter;
+	arg = rz_str_trim_head_ro(arg);
```
ret2libc: there's no need for this. The argument is trimmed by the parser, so you just need to take it without further processing.

librz/core/cmd_eval.c
```diff
@@ -741,7 +583,43 @@ RZ_IPI RzCmdStatus rz_eval_readonly_handler(RzCore *core, int argc, const char *
...
+	if (arg && *arg) {
+		rz_list_foreach (core->config->nodes, iter, node) {
```
ret2libc: I think you can find a way to avoid this code dupped below. You can probably have just one loop that works for both cases.

librz/core/cconfig.c
```diff
@@ -1737,6 +1737,219 @@ RZ_API bool rz_core_esil_cmd(RzAnalysisEsil *esil, const char *cmd, ut64 a1, ut6
...
+
+	if (*s == '-') {
+		rz_config_rm(core->config, s + 1);
+		free(s);
+		return false;
+	}
+	if (many) {
+		// space separated list of k=v k=v,..
+		// if you want to use spaces go for base64 or e.
+		RzList *list = rz_str_split_list(s, ",", 0);
```
ret2libc: is it separated by `,` or spaces? the comment says spaces.

ret2libc (commented):
Can you remove `cb_printf` field from `RzConfig`?

I've done one review. I will do another after some of the feedback has been addressed! Thanks for this, very useful!

librz/core/cconfig.c
```diff
@@ -1737,6 +1737,219 @@ RZ_API bool rz_core_esil_cmd(RzAnalysisEsil *esil, const char *cmd, ut64 a1, ut6
...
+		break;
+	case RZ_OUTPUT_MODE_LONG:
+		rz_list_foreach (cfg->nodes, iter, node) {
+			if (!str || (str && (!strncmp(str, node->name, len)))) {
+				config_print_node(cfg, node, pfx, sfx, pj, RZ_OUTPUT_MODE_LONG);
+			}
+		}
+		break;
+	case RZ_OUTPUT_MODE_LONG_JSON:
+		if (!pj) {
```
XVilka: As @ret2libc mentioned, if you use the `state` instead you will not need to check & initialize PJ

librz/core/cconfig.c
```diff
@@ -1737,6 +1737,219 @@ RZ_API bool rz_core_esil_cmd(RzAnalysisEsil *esil, const char *cmd, ut64 a1, ut6
 	return false;
 }
 
+static void __evalString(RzConfig *cfg, char *name) {
```
XVilka: I suggest to rename the function to be `like_that` and without the preceding `__`

XVilka (changes requested):

librz/core/cconfig.c
officialcjunior: We'll go with `eval_string`, then?

XVilka: Yes.

librz/core/cmd_eval.c
XVilka: @officialcjunior you forgot this one.

librz/core/cconfig.c
```diff
@@ -1737,6 +1737,180 @@ RZ_API bool rz_core_esil_cmd(RzAnalysisEsil *esil, const char *cmd, ut64 a1, ut6
...
+	}
+
+	if (RZ_STR_ISNOTEMPTY(str)) {
+		str = rz_str_trim_head_ro(str);
+		len = strlen(str);
+		if (len > 0 && str[0] == ' ') {
+			str++;
+			len--;
+		}
+	}
```
ret2libc: As said, this is not needed/should not be needed.

test/db/cmd/cmd_eval
```diff
@@ -81,6 +81,7 @@ RUN
 
 NAME=multiple assignments
 FILE==
+ARGS=-ecfg.oldshell=true
```
ret2libc: Remove this and fix the CMDS.. `e scr.highlight=" scr.highlight.grep  =scr.rainbow =true"` instead of the first line should work.

librz/core/cmd_eval.c
```diff
@@ -741,7 +566,35 @@ RZ_IPI RzCmdStatus rz_eval_readonly_handler(RzCore *core, int argc, const char *
 
 RZ_IPI RzCmdStatus rz_eval_spaces_handler(RzCore *core, int argc, const char **argv) {
```
ret2libc, line 558: this does not work for e.g. `es asm.sub`.

librz/core/cconfig.c
```diff
@@ -1737,6 +1737,180 @@ RZ_API bool rz_core_esil_cmd(RzAnalysisEsil *esil, const char *cmd, ut64 a1, ut6
...
+		free(s);
+		return true;
+	}
+	eval_string(core->config, s);
+	free(s);
+	return true;
+}
+
+static void config_print_node(RzConfig *cfg, RzConfigNode *node, RzCmdStateOutput *state) {
+	rz_return_if_fail(cfg);
```
ret2libc:
```suggestion
	rz_return_if_fail(cfg && node && state);
```

librz/core/cconfig.c
```diff
@@ -1737,6 +1737,180 @@ RZ_API bool rz_core_esil_cmd(RzAnalysisEsil *esil, const char *cmd, ut64 a1, ut6
...
+				}
+				rz_cons_printf("%s", option);
+			}
+			rz_cons_printf("]");
+		}
+		rz_cons_println("");
+	} else if (mode == RZ_OUTPUT_MODE_QUIET) {
+		rz_cons_printf("%s=%s\n", node->name, node->value);
+	} else if (mode == RZ_OUTPUT_MODE_RIZIN) {
+		rz_cons_printf("e %s=%s\n", node->name, node->value);
```
ret2libc:
```suggestion
		rz_cons_printf("e %s=%s\n", node->name, rz_cmd_escape_arg(node->value, RZ_CMD_ESCAPE_ONE_ARG));
```

librz/core/cconfig.c
```diff
@@ -1737,6 +1737,180 @@ RZ_API bool rz_core_esil_cmd(RzAnalysisEsil *esil, const char *cmd, ut64 a1, ut6
...
+static void config_print_node(RzConfig *cfg, RzConfigNode *node, RzCmdStateOutput *state) {
+	rz_return_if_fail(cfg);
+	char *option;
+	bool isFirst;
+	RzOutputMode mode = state->mode;
+	PJ *pj = state->d.pj;
+	RzListIter *iter;
+	char *es = NULL;
+
+	if (mode == RZ_OUTPUT_MODE_JSON) {
```
ret2libc: use a switch case:
```
switch (state->mode) {
case RZ_OUTPUT_MODE_JSON:
    ....
  break;
case RZ_OUTPUT_MODE_LONG_JSON:
  ....
  break;
default:
   rz_warn_if_reached();
  break;
```

librz/core/cconfig.c
```diff
@@ -1737,6 +1737,180 @@ RZ_API bool rz_core_esil_cmd(RzAnalysisEsil *esil, const char *cmd, ut64 a1, ut6
...
+		if (rz_str_isnumber(node->value)) {
+			pj_kn(pj, node->name, rz_num_math(NULL, node->value));
+			return;
+		} else if (rz_str_is_bool(node->value)) {
+			pj_kb(pj, node->name, (node->value));
+			return;
+		}
+		pj_ks(pj, node->name, node->value);
```
ret2libc: shall this be `} else { pj_ks (pj, node->name, node->value)`?

librz/core/cconfig.c
```diff
@@ -1737,6 +1737,180 @@ RZ_API bool rz_core_esil_cmd(RzAnalysisEsil *esil, const char *cmd, ut64 a1, ut6
...
+
+	rz_list_foreach (cfg->nodes, iter, node) {
+		if (!str || (!strncmp(str, node->name, len))) {
+			config_print_node(cfg, node, state);
+		}
+	}
+
+	if (mode == RZ_OUTPUT_MODE_LONG_JSON || mode == RZ_OUTPUT_MODE_JSON) {
+		pj_end(pj);
+		rz_cons_println(pj_string(pj));
```
ret2libc: this is not going to be needed after i push a fix for RzCmd.

librz/core/cmd_eval.c
```diff
@@ -498,142 +473,13 @@ RZ_IPI int rz_eval_color(void *data, const char *input) {
...
 RZ_IPI RzCmdStatus rz_eval_getset_handler(RzCore *core, int argc, const char **argv) {
 	if (argc == 1) {
-		rz_config_list(core->config, NULL, 0);
+		RzCmdStateOutput state = { 0 };
+		rz_cmd_state_output_init(&state, RZ_OUTPUT_MODE_QUIET);
+		rz_core_config_print_all(core->config, NULL, &state);
+		rz_cmd_state_output_print(&state);
+		rz_cmd_state_output_fini(&state);
 		return RZ_CMD_STATUS_OK;
 	}
```
ret2libc: actually this code is never reached, because `e` must have at least one argument... I guess you can remove it, because this is anyway covered by `elq`.

librz/core/cconfig.c
```diff
@@ -1737,6 +1737,180 @@ RZ_API bool rz_core_esil_cmd(RzAnalysisEsil *esil, const char *cmd, ut64 a1, ut6
...
+			if (v) {
+				rz_cons_printf("%s\n", v);
+			} else {
+				rz_cons_printf("Invalid config key %s\n", name);
+			}
+		}
+	}
+}
+
+RZ_API bool rz_core_config_eval_and_print(RzCore *core, const char *str, bool many) {
```
ret2libc: `many` is not required anymore. it was used before, but in newshell it's not needed. you can remove the argument and the path depending on it.

librz/core/cconfig.c
```diff
@@ -1737,6 +1737,180 @@ RZ_API bool rz_core_esil_cmd(RzAnalysisEsil *esil, const char *cmd, ut64 a1, ut6
...
+	rz_return_val_if_fail(core && str, false);
+
+	char *s = rz_str_trim_dup(str);
```
ret2libc: this is not needed. let the caller do the trimming, if necessary (but it should not be).

librz/core/cconfig.c
```diff
@@ -1737,6 +1737,180 @@ RZ_API bool rz_core_esil_cmd(RzAnalysisEsil *esil, const char *cmd, ut64 a1, ut6
...
+
+	if (!*s || !strcmp(s, "help")) {
+		RzCmdStateOutput state = { 0 };
+		rz_cmd_state_output_init(&state, RZ_OUTPUT_MODE_QUIET);
+		rz_core_config_print_all(core->config, NULL, &state);
+		rz_cmd_state_output_print(&state);
+		rz_cmd_state_output_fini(&state);
+		free(s);
+		return false;
+	}
```
ret2libc: IMHO we can remove this thing. There are other ways to get the list of evals. @XVilka cc

librz/core/cconfig.c
```diff
@@ -1737,6 +1737,180 @@ RZ_API bool rz_core_esil_cmd(RzAnalysisEsil *esil, const char *cmd, ut64 a1, ut6
...
+
+	if (*s == '-') {
+		rz_config_rm(core->config, s + 1);
+		free(s);
+		return false;
+	}
```
ret2libc: I'd remove even this in favour of a cleaner API. I don't think anyway users should be allowed to remove config variables.... In r2 a user can do `e -asm.var` and remove the asm.var variable... I don't think that's useful nor should be supported.
cc @XVilka WDYT?

librz/core/cconfig.c
ret2libc: Actually...  I'm starting to doubt the usefulness of the whole function. @XVilka it seems to be used only in rizin.c.. Do you see anything useful that can be done there that cannot be done with `rz_config_get`+`rz_cons_printf`?

XVilka: I think we can indeed remove it completely it seems.

librz/core/cconfig.c
XVilka: Yes, it's safe to remove it.

librz/core/cconfig.c
officialcjunior: This seems to be doing something. When I remove this, passing keys as arguments to these commands doesn't work; it just prints the whole thing like no arguments were passed.

librz/core/cconfig.c
XVilka: @officialcjunior can be removed after the rebase on top of the latest "dev"

librz/core/cconfig.c
officialcjunior: Looks like this broke a gazillion of tests. Should we make this a part of another PR or something?

librz/core/cmd_eval.c
officialcjunior: What should `es asm.sub` return, actually?
It doesn't seem to be returning anything even on `dev`?

librz/core/cconfig.c
officialcjunior: Now, we handle `RZ_OUTPUT_MODE_RIZIN` separately in `config_print_node()`

librz/core/cconfig.c
XVilka: @officialcjunior the breakage looks like because `-escr.color=0` didn't work in `binrz/rz-test/run.c`. Just inserting space should fix the problem.

librz/config/config.c
```diff
@@ -545,3 +545,13 @@ RZ_API bool rz_config_unserialize(RZ_NONNULL RzConfig *config, RZ_NONNULL Sdb *d
 	sdb_foreach(db, load_config_cb, config);
 	return true;
 }
+
+RZ_API bool rz_config_eval(RzConfig *cfg, const char *str) {
+	char *s = rz_str_trim_dup(str);
```
ret2libc: you should `free` this.

librz/config/config.c
```diff
@@ -545,3 +545,13 @@ RZ_API bool rz_config_unserialize(RZ_NONNULL RzConfig *config, RZ_NONNULL Sdb *d
...
+	char *eq = strchr(s, '=');
+	if (eq) {
+		*eq++ = 0;
+		rz_config_set(cfg, s, eq);
+	}
```
ret2libc:
```suggestion
	if (!eq) {
		return false;
	}
	*eq++ = 0;
	rz_config_set(cfg, s, eq);
```

XVilka:
```
+rz_config_set: variable 'e scr.interactive' not found
+rz_config_set: variable 'e scr.interactive' not found
+rz_config_set: variable 'e scr.interactive' not found
+rz_config_set: variable 'e scr.interactive' not found
+rz_config_set: variable 'e scr.interactive' not found
+rz_config_set: variable 'e scr.interactive' not found
```
I tries to set this variable.

librz/core/cconfig.c
ret2libc: This is wrong anyway. If something breaks, it's likely because the caller should do the trimming itself.

librz/core/cconfig.c
```diff
@@ -1737,6 +1737,123 @@ RZ_API bool rz_core_esil_cmd(RzAnalysisEsil *esil, const char *cmd, ut64 a1, ut6
...
+			}
+			rz_cons_printf("]");
+		}
+		rz_cons_println("");
+		break;
+	case RZ_OUTPUT_MODE_QUIET:
+		rz_cons_printf("%s=%s\n", node->name, node->value);
+		break;
+	case RZ_OUTPUT_MODE_RIZIN:
+		rz_cons_printf("e %s=%s\n", node->name, rz_cmd_escape_arg(node->value, RZ_CMD_ESCAPE_ONE_ARG));
```
ret2libc: rz_cmd_escape_arg should be freed

librz/core/cmd_eval.c
ret2libc: It's broken also on dev indeed. `es asm.sub` should return the subkeys of `asm.sub`, the same way as `es zoom` returns the subkeys of zoom.

Anyway, let's ignore this thing for now and do it another time.

librz/core/cconfig.c
officialcjunior: Good catch, thanks.

So, I looked around and it looks like this is the way?
```suggestion c
		char *es = rz_cmd_escape_arg(node->value, RZ_CMD_ESCAPE_ONE_ARG);
		rz_cons_printf("e %s=%s\n", node->name, es);
		free(es);
```

librz/main/rizin.c
```diff
@@ -1294,8 +1294,8 @@ RZ_API int rz_main_rizin(int argc, const char **argv) {
 		rz_core_block_read(r);
 
 		rz_list_foreach (evals, iter, cmdn) {
-			rz_config_eval(r->config, cmdn, false);
-			rz_cons_flush();
+			rz_config_eval(r->config, cmdn);
+			rz_cons_flush();
```
XVilka: Trailing whitespace?

XVilka (changes requested):

officialcjunior:
Two tests left: https://builds.sr.ht/~xvilka/job/583422#task-test-14

```bash
[XX] db/cmd/regexp "/e /t\wst\d\d\d\s\w\w/i" - oldshell
RZ_NOPLUGINS=1 rizin -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -Qc 'w "test123 ab"
w "Test123 ab" @444
?e
e search.in=block
b 777
cfg.oldshell=true
"/e /t\wst\d\d\d\s\w\w/i"
' malloc://1024
-- stdout
--- expected
+++ actual
@@ --1,5 +-1,5 @@
 
-0x00000001 hit0_0 ""test123 ab""
-0x000001bd hit0_1 ""Test123 ab""
+0x00000000 hit0_0 "test123 ab"
+0x000001bc hit0_1 "Test123 ab"

-- stderr
Please. use 'cf [file]'
Searching 10 bytes in [0x0-0x309]
```

and

```bash
[XX] db/cmd/cmd_eval multiple assignments
RZ_NOPLUGINS=1 rizin -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -Qc 'e scr.highlight=" scr.highlight.grep =scr.rainbow =true"
e scr.highlight
e scr.highlight.grep
e scr.rainbow
' =
-- stdout
--- expected
+++ actual
@@ -1,3 +1,3 @@
-scr.highlight.grep  =scr.rainbow =true
+ scr.highlight.grep =scr.rainbow =true
 false
 false
```
I believe that the second one can be fixed according to the output?

librz/core/cconfig.c
ret2libc: yep!

XVilka:
For the first test the new output looks better, no? It's because in the newshell `w "bla foo"` interpreted differently from the oldshell. You can also move the `cfg.oldshell=true` before the `w` command to restore the old output, I think.

XVilka:
To be honest, I wonder if we should this horrendous syntax `e scr.highlight=" scr.highlight.grep =scr.rainbow =true"` completely?

ret2libc:
For both you can just fix the output, yes. New version seems better.

librz/config/config.c
```diff
@@ -831,3 +545,15 @@ RZ_API bool rz_config_unserialize(RZ_NONNULL RzConfig *config, RZ_NONNULL Sdb *d
 	sdb_foreach(db, load_config_cb, config);
 	return true;
 }
+
+RZ_API bool rz_config_eval(RzConfig *cfg, const char *str) {
```
XVilka: Please add `RZ_NONNULL`, `rz_return_val_if_fail(str, false)`, Doxygen.

librz/core/cconfig.c
```diff
@@ -1737,6 +1737,125 @@ RZ_API bool rz_core_esil_cmd(RzAnalysisEsil *esil, const char *cmd, ut64 a1, ut6
...
+	PJ *pj = state->d.pj;
+	RzOutputMode mode = state->mode;
+
+	if (mode == RZ_OUTPUT_MODE_LONG_JSON) {
+		pj_a(pj);
+	} else if (mode == RZ_OUTPUT_MODE_JSON) {
+		pj_o(pj);
+	}
+
+	if (RZ_STR_ISNOTEMPTY(str)) {
```
XVilka: Have you tried to remove this? How does it affect the tests?

XVilka (changes requested):

librz/core/cconfig.c
officialcjunior: Yes. On a closer look, I realized that (as @ret2libc foretold), it looks like the trimming is already done properly, but the length needs to be calculated on `strncmp` below. I tried to use `strcmp`instead but that didn't seem to work - but I was a bit busy while trying it out so I will try this again after some time and let you know.

librz/core/cconfig.c
officialcjunior: There has been some improvements. Please see https://github.com/rizinorg/rizin/pull/1488#discussion_r704276782

librz/core/cconfig.c
```diff
@@ -1737,6 +1737,121 @@ RZ_API bool rz_core_esil_cmd(RzAnalysisEsil *esil, const char *cmd, ut64 a1, ut6
...
+
+RZ_API void rz_core_config_print_all(RzConfig *cfg, const char *str, RzCmdStateOutput *state) {
+	rz_return_if_fail(cfg);
+	RzConfigNode *node;
+	RzListIter *iter;
+	int len = 0;
+	PJ *pj = state->d.pj;
+	RzOutputMode mode = state->mode;
+
+	if (mode == RZ_OUTPUT_MODE_LONG_JSON) {
```
XVilka, line 1839: @ret2libc is this still required? I believe you added it as a `rz_cmd_state_output_array_start()` (and the footer `rz_cmd_state_output_array_end()`). Is the use `pj_o()` for "short JSON" really required?

ret2libc: yes in this case it is needed because there's a difference between `elj` and `elJ`. `elj` just returns a json object like:
```json
{
  "cfg.oldshell": true,
  "asm.var.sub": false,
}
```
while `elJ` returns:
```json
[
  {
    "name": "cfg.oldshell",
   "value": true,
   "type": "boolean",
  "option": ....
   "description: "...."
  },
  {
    "name": "asm.var.sub",.....
  }
]
```

ret2libc: So in one case you have an array, in the other you have an object.

librz/core/cconfig.c
ret2libc: I think this should not be needed either. If `str` is the string `a` that means that len will be 0, right? In that case you'd run strncmp with a length of 0... That's wrong. Remove this and if something fails, check what is calling this function with the wrong values, they should be fixed, not this function IMO.

ret2libc (changes requested):

librz/core/cconfig.c
officialcjunior: Just passing the length of the string without those check- `if (!str || (!strncmp(str, node->name, strlen(str)))) {` seems to work.
Can I push that in? Isn't that alright?

ret2libc: use rz_str_startswith. That should work as well and it's better.

librz/core/cconfig.c
```diff
@@ -1737,6 +1737,115 @@ RZ_API bool rz_core_esil_cmd(RzAnalysisEsil *esil, const char *cmd, ut64 a1, ut6
...
+		rz_cons_printf("%20s: %s\n", node->name,
+			node->desc ? node->desc : "");
+		break;
+	default:
+		rz_warn_if_reached();
+		break;
+	}
+}
+
+RZ_API void rz_core_config_print_all(RzConfig *cfg, const char *str, RzCmdStateOutput *state) {
```
XVilka, line 1832: Please add Doxygen here

XVilka (approved):
Looks good now, just please add Doxygen. @ret2libc it makes sense to import this to the 0.3.0, what do you think?

librz/config/config.c
```diff
@@ -831,3 +545,23 @@ RZ_API bool rz_config_unserialize(RZ_NONNULL RzConfig *config, RZ_NONNULL Sdb *d
...
+ * \param cfg reference to RzConfig
+ * \param str reference the configuration variable string (eg, 'asm.arch=x86')
+ */
+RZ_API bool rz_config_eval(RZ_NONNULL RzConfig *cfg, RZ_NONNULL const char *str) {
+	rz_return_val_if_fail(str, false);
+
+	char *name = rz_str_trim_dup(str);
+	char *value = strchr(name, '=');
+	if (!value) {
+		return false;
```
ret2libc, line 562:
```suggestion
		free(name);
		return false;
```

test/db/cmd/regexp
```diff
@@ -30,9 +30,9 @@ EOF
 RUN
 
 NAME="/e /t\wst\d\d\d\s\w\w/i" - oldshell
-ARGS=-ecfg.oldshell=true
 FILE=malloc://1024
 CMDS=<<EOF
+cfg.oldshell=true
```
ret2libc, line 35: Why this?

XVilka: @ret2libc it was in the old test afaik.

XVilka: The test itself called "oldshell"

ret2libc (approved):

PR #1500, opened 2021-08-22
Add forward-search in shell (#571)

### LINKED ISSUE #571: Ctrl+S - Forward-search
When using `Ctrl+R` to use the backward-search you may accidentally skip the history result you wanted and you are forced to cancel the search and start from the beginning.

Pressing `Ctrl+S` performs forward-search (like emacs/readline), allowing selecting the history result you just skipped.

💔 

**Additional context**
As an emacs user, this issue annoys me every time 🙂

By the way, there's something mapped in `Ctrl+S` that seems to cancel the search and move the cursor one character back when pressed out of the search-mode, what is it supposed to do?

borzacchiello:

**Detailed description**

The PR adds a forward-search to the rizin shell, allowing selecting the history skipped using the backward search (Ctrl-R).

I added it to `Ctrl-G` since `Ctrl-S` was already mapped to [_backspace_](https://github.com/rizinorg/rizin/blob/41b4da831eef5bcaeb980415b64f26df366d8045/librz/cons/dietline.c#L1508) command. I choose `Ctrl-G` only because isn't mapped and `G` is close to `R` on the keyboard, but maybe `Ctrl-S` can be remapped instead. Let me know what you prefer, and I will modify the commit accordingly!

**Test plan**

If I run the commands:
```
[0x000bfe24]> aa
[0x000bfe24]> ab
[0x000bfe24]> ac
[0x000bfe24]> ad
```
And I hit `Ctrl-R`, I write `a`, and I hit `Ctrl-R` two times, I get:
```
(reverse-i-search (a)): ab
```
Using the _forward-search_ I can go back to the skipped commands. E.g., hitting `Ctrl-G` one time I get:
```
(forward-i-search (a)): ac
```

**Closing issues**

closes #571

### DIFF
--- a/librz/cons/dietline.c
+++ b/librz/cons/dietline.c
@@ -1509,7 +1509,9 @@ RZ_API const char *rz_line_readline_cb(RzLineReadCallback cb, void *user) {
 			break;
 		case 19: // ^S -- forward-search
 			if (gcomp) {
-				gcomp_idx--;
+				if (gcomp_idx > 0) {
+					gcomp_idx--;
+				}
 				gcomp_is_rev = false;
 			} else {
 				__move_cursor_left();
@@ -2000,17 +2002,12 @@ RZ_API const char *rz_line_readline_cb(RzLineReadCallback cb, void *user) {
 						if (i == 0) {
 							if (gcomp_is_rev) {
 								gcomp_idx--;
-							} else {
-								gcomp_idx++;
 							}
 						}
 					}
 				}
-				if (gcomp_is_rev) {
-					printf("\r (reverse-i-search (%s)): %s\r", I.buffer.data, gcomp_line);
-				} else {
-					printf("\r (forward-i-search (%s)): %s\r", I.buffer.data, gcomp_line);
-				}
+				const char *prompt = gcomp_is_rev ? "reverse-i-search" : "forward-i-search";
+				printf("\r (%s (%s)): %s\r", prompt, I.buffer.data, gcomp_line);
 			} else {
 				__print_prompt();
 			}
--- a/librz/cons/dietline.c
+++ b/librz/cons/dietline.c
@@ -1332,6 +1332,7 @@ RZ_API const char *rz_line_readline_cb(RzLineReadCallback cb, void *user) {
 	static int gcomp_idx = 0;
 	static bool yank_flag = 0;
 	static int gcomp = 0;
+	static int gcomp_is_rev = true;
 	char buf[10];
 #if USE_UTF8
 	int utflen;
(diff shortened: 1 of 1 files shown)


XVilka (approved):

ret2libc (changes requested):
Hi and thanks for your contribution!! This is indeed a nice feature to have.

Since you mentioned ^S is already taken, I was looking at its code and found out:
```C
		case 19: // ^S -- backspace
			if (gcomp) {
				gcomp--;
			} else {
				__move_cursor_left();
			}
			break;
```

As you can see the cursor is moved to the left only if `!gcomp`. `gcomp` is the variable that says if we are searching the history or not. If you look at ^R you see that when gcomp is active, `gcomp_idx` is incremented to move the search backward. So `gcomp_idx` should be decremented to move the search forward.

I think `^S` was already written with the idea of having forward-search, however the code incorrectly decrements `gcomp` instead of `gcomp_idx`. I think fixing that should be enough to have properly working forward-search.

borzacchiello:
> Hi and thanks for your contribution!! This is indeed a nice feature to have.
> 
> Since you mentioned ^S is already taken, I was looking at its code and found out:
> 
> ```c
> 		case 19: // ^S -- backspace
> 			if (gcomp) {
> 				gcomp--;
> 			} else {
> 				__move_cursor_left();
> 			}
> 			break;
> ```
> 
> As you can see the cursor is moved to the left only if `!gcomp`. `gcomp` is the variable that says if we are searching the history or not. If you look at ^R you see that when gcomp is active, `gcomp_idx` is incremented to move the search backward. So `gcomp_idx` should be decremented to move the search forward.
> 
> I think `^S` was already written with the idea of having forward-search, however the code incorrectly decrements `gcomp` instead of `gcomp_idx`. I think fixing that should be enough to have properly working forward-search.

Done! I modified the commit fixing `^S` and deleting `^G`.

XVilka:
@ret2libc why 0.4.0? It doesn't break backward compatibility and basically a bugfix.

XVilka (approved):

librz/cons/dietline.c
```diff
@@ -1995,11 +1998,19 @@ RZ_API const char *rz_line_readline_cb(RzLineReadCallback cb, void *user) {
...
+							}
 						}
 					}
 				}
-				printf("\r (reverse-i-search (%s)): %s\r", I.buffer.data, gcomp_line);
+				if (gcomp_is_rev) {
+					printf("\r (reverse-i-search (%s)): %s\r", I.buffer.data, gcomp_line);
+				} else {
+					printf("\r (forward-i-search (%s)): %s\r", I.buffer.data, gcomp_line);
+				}
```
ret2libc:
```suggestion
				const char *prompt = gcomp_is_rev ? "reverse-i-search" : "forward-i-search";
				printf("\r (%s (%s)): %s\r", prompt, I.buffer.data, gcomp_line);
```

ret2libc (changes requested):
If I press many times ^S i think the index keeps increasing, thus I have to press many times ^R to go back to the previous search element. I think the index should stop increasing when you reached the last element you can search.

librz/cons/dietline.c
borzacchiello: Done

borzacchiello:
> If I press many times ^S i think the index keeps increasing, thus I have to press many times ^R to go back to the previous search element. I think the index should stop increasing when you reached the last element you can search.

Now it should be fine. The problem was that `gcomp_idx` became negative

ret2libc (approved):
Great!! Thanks a lot for this ;)

PR #1526, opened 2021-08-25
Fix #1507 - Ignore sigpipe

### LINKED ISSUE #1507: pipe to head/tail crashes rizin because of SIGPIPE
### Work environment

| OS/arch/bits (mandatory)                             | Fedora 33 x64
| File format of the file you reverse (mandatory)      | Any
| `rizin -v` full output, **not truncated** (mandatory)  | rizin 0.3.0-git @ linux-x86-64 commit: 3b0d11eebb3af4634741c7a22f70b6c8d4e174f5, build: 2021-08-23__18:10:22

### Expected behavior
Rizin should continue to run.

### Actual behavior
Rizin is closed without saying anything after the data piped through `head` is printed.

```
$ rizin /bin/ls
> pd 1000 | head
```
maybe run `pd 1000 | head` few times if it doesn't work immediately.

From a quick look it seems that the `write` in `rz_subprocess_stdin_write` causes a SIGPIPE to be sent to rizin process, which terminates because it does not handle it. The SIGPIPE is likely caused by head/tail commands that close their ends of the pipe after enough data has been read.

We should probably handle the SIGPIPE signal appropriately in `RzSubprocess`.

wargio:

**Detailed description**

Ignores the sigpipe signal on *nix systems (as it was done in the old code).

Closes #1507

### DIFF
--- a/librz/util/subprocess.c
+++ b/librz/util/subprocess.c
@@ -758,7 +758,6 @@ RZ_API RzSubprocess *rz_subprocess_start_opt(RzSubprocessOpt *opt) {
 	// because we can't use functions that lock after fork
 	char **child_env = create_child_env(opt->envvars, opt->envvals, opt->env_size);
 
-	rz_sys_signal(SIGPIPE, SIG_IGN);
 	proc->pid = rz_sys_fork();
 	if (proc->pid == -1) {
 		// fail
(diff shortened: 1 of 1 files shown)


ret2libc (commented):
Is there a better way to do this? Like... can we ignore the sigpipe only if the broken pipe is with a process created with subprocess? I should investigate a bit... do you know?

wargio:
i looked this up and there is no better way. on osx/bsd you can do this by setting an option on the pipe but that is not available on linux.
A portable solution outside this unfortunately does not exists.

ret2libc:
Only interesting things I could find are https://riptutorial.com/posix/example/17424/handle-sigpipe-generated-by-write---in-a-thread-safe-manner and http://www.microhowto.info/howto/ignore_sigpipe_without_affecting_other_threads_in_a_process.html .

They have the advantage of ignoring the SIGPIPE only on the current thread and not on everything. Also, we should probably ignore the SIGPIPE just before writing and restoring it after doing the write.

ret2libc:
At the least, I think we should ignore sigpipe just before the write and restore it after it. WDYT?

ret2libc (approved):

PR #1552, opened 2021-08-30
Use `rz_read_{le,be}XX` API in TI TMS320 code

XVilka:

**Detailed description**

Removed some unnecessary endian functions from both `librz/asm/arch/tms320` and `librz/include/rz_endian.h`

Note, that previously used `ut8p_bw` in OMF and SNES code was misnamed - I suppose `bw` means `big endian word`, while in fact it's little endian.

Please check carefully if I didn't mix things up - I often tend to mistaken endianness and right/left between each other.

**Test plan**

CI is green


wargio (approved):

librz/include/rz_endian.h
```diff
@@ -65,6 +65,23 @@ static inline void rz_write_at_be16(void *dest, ut16 val, size_t offset) {
 	rz_write_be16(d, val);
 }
 
+static inline ut32 rz_read_be24(const void *src) {
+	const ut8 *up = (const ut8 *)src + 2;
+	return rz_read_be16(src) | rz_read_be8(up);
```
ret2libc: I'm not sure about this...

It should probably be something like:
```
rz_read_be8(src) << 16 | rz_read_be8(src+1) << 8 | rz_read_be8(src+2)
```
no?

XVilka: You are right, yes.

librz/asm/arch/tms320/tms320_dasm.c
```diff
@@ -10,11 +10,32 @@
...
+}
+
+static inline ut32 be24(ut32 v) {
+	ut32 value = v;
+#if __BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__
+	ut8 *pv = (void *)&v;
+	value = (pv[0] << 16) | (pv[1] << 8) | pv[2];
+#endif
+	return value;
+}
```
ret2libc, line 37: are these still needed?

XVilka: Yes, since they convert `ut32` that is read from the bitfield. Using just the `rz_read_*()` API would require more significant refactoring. Thus I stopped on this middle-ground solution.

ret2libc: Actually... I'm not sure what you mean. Isn't `be24` the same thing as `rz_read_be24`?

XVilka: It isn't - just see the input values. `rz_read_be24()` reads memory, this `be24` *converts* the host-endianned *integer*, which is in turn was read from the bitfield with `field_value()` macros. Of course it can be converted to read corresponding memory directly, but its *a lot of work*.

ret2libc: I'd need to think a bit more about this... Anyway, tms320.... Ok :D

ret2libc (approved):

PR #1574, opened 2021-09-02
Add serialization functionality for breakpoints

### LINKED ISSUE #707: Serialize Breakpoints to Projects
**Environment information**
* Operating System: Windows 10 Pro (20H2) 19042.804
* Cutter version: 1.12.0
* Obtained from:
  - [] Built from source
  - [x ] Downloaded release from Cutter website or GitHub 
  - [ ] Distribution repository
* File format: .exe

**Describe the bug**
Breakpoints are not saved.

I placed some breakpoints and saved my project, I loaded my project again, but I didn't find my saved breakpoints.

**To Reproduce**

1. Set a break point.
2. Save your project.
3. Close the program.
4. Load your saved project again.
5. Open breakpoint tab.

**Expected behavior**

Breakpoints must be loaded when I load my project again.

**Screenshots**

**Additional context**

I don't know If this is a bug or this feature not available yet.

### LINKED ISSUE #2735: Breakpoints are not saved
**Environment information**
* Operating System: windows 10
* Cutter version: 2.0.2
* File format: pe
 * Arch: x86
 * Type: EXEC (Executable file)

**Describe the bug**
Breakpoints are not saved to the project.

**To Reproduce**
1. Add a breakpoint.
2. Save the project.
3. Close the project.
4. Re open the same project.

**Expected behavior**
To have all breakpoint in the breakpoint window.

**What acutely happens**
No breakpoints in the breakpoint window.

DMaroo:

**Detailed description**

* Breakpoints now serialized and saved in project files
* You get them back where they were when you reopen a project

**Test plan**

Add tests once the serialization is approved and reviewed.

**Closing issues**

Closes #707.
Closes rizinorg/cutter#2735.


librz/bp/serialize_bp.c
```diff
@@ -0,0 +1,146 @@
...
+	rz_return_if_fail(db && bp);
+
+	RzListIter *iter;
+	RzBreakpointItem *bp_item;
+	rz_list_foreach (bp->bps, iter, bp_item) {
+		PJ *j = pj_new();
+		if (!j) {
+			return;
+		}
+		pj_o(j);
```
ret2libc, line 23: @thestr4ng3r is it preferred to save as json or as sdb directly?

DMaroo: My motivation for doing this was basically the code from `serialize_flag.c`. Much of the code in this file is similar to that of serializing flag zones.

Also, my 2 cents: Serializing directly as sdb will probably write every attribute of every breakpoint as a new key value pair. When we first serialize to JSON and then store as sdb, the data looks prettier and less cluttered.

librz/bp/serialize_bp.c
```diff
@@ -0,0 +1,146 @@
...
+
+	const RzJson *child;
+	RzBreakpointItem *bp_item = RZ_NEW0(RzBreakpointItem);
+	if (!bp_item) {
+		goto return_goto;
+	}
+	bp_item->name = strdup(k);
+
+	for (child = json->children.first; child; child = child->next) {
+		if (child->type == RZ_JSON_INTEGER) {
```
ret2libc: I saw other code using `RZ_KEY_PARSER_JSON`, is that possible here as well?

DMaroo: I'll have a look. Thanks.

librz/bp/serialize_bp.c
thestr4ng3r: As a rule of thumb, an sdb entry should correspond to a single "object", whatever that means in the specific case. Attributes should be contained in the value as json.
So here, creating one sdb entry per breakpoint with all info inside of it as json is the best approach in my opinion.
You just have to make sure `bp_item->name` really is a unique identifier for the bp, I don't know the breakpoint code well enough to tell whether it is without checking.

librz/bp/serialize_bp.c
DMaroo: https://github.com/rizinorg/rizin/blob/dev/librz/analysis/serialize_analysis.c#L637-L664

Looks like we can also make a key parser to do this job, but I don't exactly know how helpful it will be. @thestr4ng3r Your thoughts, please.

librz/bp/serialize_bp.c
DMaroo: Yes. Names aren't a unique identifier for breakpoints. I'll use addresses instead.

librz/bp/serialize_bp.c
thestr4ng3r: Yes, the key parser should be used here. It's both more concise and faster since it avoids the `strcmp` chaining.

test/unit/test_serialize_bp.c
```diff
@@ -0,0 +1,81 @@
...
+#include "minunit.h"
+
+bool test_bp_save_load() {
+	Sdb *bp_sdb = sdb_new0();
+	mu_assert_notnull(bp_sdb, "failed to create sdb");
+	RzDebug *debug_save = rz_debug_new(0);
+	mu_assert_notnull(debug_save, "failed to create debug instance");
+
+	RzBreakpointItem *item_save = rz_debug_bp_add(debug_save, 0x1337, 0, true, 0, "doom", 42);
+	mu_assert_notnull(item_save, "failed to create bp item");
```
DMaroo, line 16: Tests are failing here. I think I am creating a breakpoint at a memory location non-existent in the current debug instance, however, I am not sure how to resolve this. Will require some help, @thestr4ng3r.

librz/bp/serialize_bp.c
```diff
@@ -0,0 +1,149 @@
...
+
+		pj_ki(j, "size", bp_item->size);
+		pj_kb(j, "swstep", bp_item->swstep);
+		pj_ki(j, "togglehits", bp_item->togglehits);
+		pj_ki(j, "trace", bp_item->trace);
+		pj_end(j);
+
+		char key[19];
+		snprintf(key, 19, "0x%" PFMT64x, bp_item->addr);
+		sdb_set(db, key, pj_string(j), 0);
```
ret2libc:
```suggestion
		sdb_set(db, rz_strf(key, "0x%" PFMT64x, bp_item->addr), pj_string(j), 0);
```

DMaroo:
The breakpoint loading doesn't seem to work. Everything looks correct and as expected in the `.rzdb` project file, but loading fails when trying to load the breakpoints. I suspect the error to be somewhere in the following lines: 

https://github.com/rizinorg/rizin/pull/1574/files#diff-e10dc9ab50cb92db60c5eaf3dd32a456141636e478a2e6b320f25d3c28892e08R82-R84

DMaroo:
Breakpoint loading doesn't seem to work. I seem to be making some mistake here. From my analysis, the `bp_load_cb` callback function never gets called, and therefore `false` is returned by `rz_serialize_bp_load`. However, I am clearly able to see the saved breakpoints in the `.rzdb` project file. I suppose I must be missing to add or set something, but I am unable to figure it out.
cc @thestr4ng3r.

librz/bp/serialize_bp.c
```diff
@@ -0,0 +1,308 @@
+// SPDX-FileCopyrightText: 2021 DMaroo <dhruvmaru007@gmail.com>
+// SPDX-License-Identifier: LGPL-3.0-only
+
+#include <rz_util/rz_serialize.h>
+#include <rz_bp.h>
+
+RZ_API void rz_serialize_bp_save(RZ_NONNULL Sdb *db, RZ_NONNULL RzBreakpoint *bp) {
```
XVilka, line 13: Please add Doxygen

librz/debug/serialize_debug.c
```diff
@@ -0,0 +1,24 @@
+// SPDX-FileCopyrightText: 2021 DMaroo <dhruvmaru007@gmail.com>
+// SPDX-License-Identifier: LGPL-3.0-only
+
+#include <rz_debug.h>
+#include <rz_bp.h>
+#include <rz_util/rz_serialize.h>
+
+RZ_API void rz_serialize_debug_save(RZ_NONNULL Sdb *db, RZ_NONNULL RzDebug *dbg) {
```
XVilka, line 14: Doxygen please

librz/debug/serialize_debug.c
```diff
@@ -0,0 +1,24 @@
...
+	rz_return_if_fail(db && dbg);
+	rz_serialize_bp_save(sdb_ns(db, "breakpoints", true), dbg->bp);
+}
+
+RZ_API bool rz_serialize_debug_load(RZ_NONNULL Sdb *db, RZ_NONNULL RzDebug *dbg, RZ_NULLABLE RzSerializeResultInfo *res) {
```
XVilka, line 27: Also Doxygen

librz/bp/serialize_bp.c
```diff
@@ -0,0 +1,308 @@
...
+	ret = true;
+
+beach:
+	rz_json_free(json);
+heaven:
+	free(json_str);
+	return ret;
+}
+
+RZ_API bool rz_serialize_bp_load(RZ_NONNULL Sdb *db, RZ_NONNULL RzBreakpoint *bp, RZ_NULLABLE RzSerializeResultInfo *res) {
```
XVilka, line 302: Doxygen

librz/bp/serialize_bp.c
```diff
@@ -0,0 +1,308 @@
...
+	BP_FIELD_ENABLED,
+	BP_FIELD_TOGGLEHITS,
+	BP_FIELD_HITS,
+	BP_FIELD_PIDS,
+	BP_FIELD_DATA,
+	BP_FIELD_COND,
+	BP_FIELD_EXPR
+};
+
+RZ_API RzSerializeBpParser rz_serialize_bp_parser_new(void) {
```
XVilka, line 90: Doxygen

XVilka (changes requested):

librz/core/project_migrate.c
```diff
@@ -238,13 +238,30 @@ RZ_API bool rz_project_migrate_v4_v5(RzProject *prj, RzSerializeResultInfo *res)
 	return true;
 }
 
+// --
+// Migration 5 -> 6
+//
+// Changes from <commit-hash>
```
DMaroo, line 244: Will need to fill this commit hash once this PR gets merged into dev.

librz/core/project_migrate.c
```diff
@@ -238,13 +238,30 @@ RZ_API bool rz_project_migrate_v4_v5(RzProject *prj, RzSerializeResultInfo *res)
...
+//	Added serialization functionality for debug (only for breakpoints as of now)
```
XVilka, line 245: Please also mention *exact* namespace where it goes.

XVilka (approved):
Overall looks good. 👍

ret2libc (changes requested):
Just to block merging. I'd like to test it a bit before approving. Also, if @thestr4ng3r can give a look it would be great since he's the project expert :D

thestr4ng3r (approved):

DMaroo:
> Just to block merging. I'd like to test it a bit before approving. Also, if @thestr4ng3r can give a look it would be great since he's the project expert :D

Have you tested it yet? Also, I could help in the testing, so please let me know in case there is need for some more testing.

DMaroo:
Can this be merged? Thank you.

librz/bp/serialize_bp.c
```diff
@@ -0,0 +1,326 @@
...
+
+	RzBreakpointItem *bp_item = NULL;
+	if (bp_item_temp.hw) {
+		bp_item = rz_bp_add_hw(ctx->bp, bp_item_temp.addr, bp_item_temp.size, bp_item_temp.perm);
+	} else {
+		bp_item = rz_bp_add_sw(ctx->bp, bp_item_temp.addr, bp_item_temp.size, bp_item_temp.perm);
+	}
+	if (!bp_item) {
+		goto beach;
+	}
```
ret2libc, line 256: I was testing this on Windows with:
```
λ rizin-install\bin\rizin.exe -p rizin.rzdb -d rizin-install\bin\rizin.exe
DebugActiveProcess: (0x5) Access is denied.
DebugActiveProcess: (0x5) Access is denied.
WARNING: Breakpoint won't be placed since it's not in a valid map.
You can bypass this check by setting dbg.bpinmaps to false.
Spawned new process with pid 22532, tid = 21028
```
The breakpoints seem to work then, but I do wonder whether the restoring of the debug info happens at the right time... @thestr4ng3r idea? Is it alright as a first step?

Also, what about PIE? This is not going to work, is it?

ret2libc:
> Can this be merged? Thank you.

Sorry for the delay, we were a bit busy finishing some bugs for the upcoming release. I hope to have this merged very soon!! Thanks for your patience

librz/bp/serialize_bp.c
DMaroo: Thanks for testing on Windows. I'll check whether this works for PIE or not.

DMaroo: The serialization works because (it seems) that base addresses are (maybe) stored in the project files. This seems to work for both PIE and no-PIE. I'll make sure that this is what is indeed happening (by inspecting the code), and the breakpoints serialization working is not just a coincidence.

ret2libc: It is probably alright that it doesn't work, because IO/Bin is not properly serialized yet... let's see.

thestr4ng3r: Right now, before IO/Bin serialization is implemented, the only available information is a filename. No base address or anything similar. But even with IO and Bin, that still won't give away much info about mapping in a debug process. That is a whole different story.

The idea right now would be to never load or save a project in debug mode, but only in static mode and then switch to debug with `ood`. Then the regular rebasing that happens should take care of relocating the breakpoints to wherever the file is statically loaded by default.

ret2libc (approved):

thestr4ng3r (approved):

PR #1594, opened 2021-09-04
Fix #469 by checking the value of `analysis.jmp.cref`

### LINKED ISSUE #469: `aar` ignores `analysis.jmp.cref` value
### Work environment

| `rizin -v` full output, **not truncated** (mandatory)  | 6814610

Option `analysis.jmp.cref` is set to false by default but `aar` ignores its value and anyway creates xrefs for conditional jumps.

See `librz/core/cmd_analysis.c` for how `aar` command is implemented.

theopechli:

**Detailed description**

This issue was that during `aaa`, the config variable `analysis.jmp.cref` is ignored and, as such, `aar` is run despite its value.

Note that its default value is `false`, which may be undesirable.

**Test plan**

I can add tests tomorrow, if needed.

**Closing issues**

closes #469


librz/core/canalysis.c
```diff
@@ -6667,12 +6667,14 @@ RZ_API bool rz_core_analysis_everything(RzCore *core, bool experimental, char *d
 		}
 	}
 
-	oldstr = rz_print_rowlog(core->print, "Analyze len bytes of instructions for references (aar)");
-	(void)rz_core_analysis_refs(core, ""); // "aar"
-	rz_print_rowlog_done(core->print, oldstr);
-	rz_core_task_yield(&core->tasks);
-	if (rz_cons_is_breaked()) {
-		return false;
+	if (rz_config_get_b(core->config, "analysis.jmp.cref")) {
```
wargio, line 6670: Maybe i'm missing something here. why this flag?

wargio:
not sure if this is the right way.. let's wait for the others and especially @pelijah

p371jah:
Well I have no idea how it should be done.

XVilka:
I checked some tests output with the `analysis.jmp.cref=true` by default and the output looks better, in fact. I am not sure it should be `false` by default. cc @wargio @thestr4ng3r @ret2libc

test/db/tools/rz_diff
```diff
@@ -299,7 +299,7 @@ RUN
 
 NAME=rz-diff graphs comparison with addresses JSON
 FILE==
-CMDS=!rz-diff -C -j -t graphs -0 fcn.00401ae0 -1 fcn.00401af0 bins/other/rz-diff/true bins/other/rz-diff/false
+CMDS=!rz-diff -e analysis.jmp.cref=true -C -j -t graphs -0 fcn.00401ae0 -1 fcn.00401af0 bins/other/rz-diff/true bins/other/rz-diff/false
```
wargio, line 302: there is no `-e` option on rz-diff

ret2libc (changes requested):
I think @wargio is right. Your PR run or does not run `rz_core_analysis_refs` based on the `analysis.jmp.cref` variable. However, `analysis.jmp.cref` says `analysis.jmp.cref: Create references for conditional jumps`, while `rz_core_analysis_refs` does much more than just analyzing refs for conditional jumps, it actually defines refs for calls, unconditional branches, data, etc.

I think you should look at `rz_core_analysis_refs` to see what to modifies and in particular at `rz_core_analysis_search_xrefs`. Please let us know if you need additional help!

test/db/tools/rz_diff
theopechli: I've been fooled by `man rz-diff`

wargio: yeah. that one is old

theopechli:
The only test that's failing is `db/tools/rz_diff rz-diff graphs comparison with addresses JSON`, because `analysis.jmp.cref=false` by default and rz-diff has no option to configure variables, despite what `man rz-diff` prints.

Two obvious solutions are:
1. Set `analysis.jmp.cref` to true
2. Mark that test as broken

I don't know if I missed something else in this PR.

wargio:
> The only test that's failing is `db/tools/rz_diff rz-diff graphs comparison with addresses JSON`, because `analysis.jmp.cref=false` by default and rz-diff has no option to configure variables, despite what `man rz-diff` prints.
> 
> Two obvious solutions are:
> 
> 1. Set `analysis.jmp.cref` to true
> 2. Mark that test as broken
> 
> I don't know if I missed something else in this PR.

i will try to add `-e` option to rz-diff today so you can rebase and it should work.

ret2libc:
@wargio @theopechli what's the status of this? I think the patch looks good, why is it failing? Can we workaround the problem so that `-e` in rz-diff can be done at a later time?

wargio:
no. i need to patch rz-diff and add the -e flag. i will try to add it tonight.

ret2libc:
@wargio ping

wargio:
i'm so sorry for the delay. i have added the option. once it is merged, we can update this

wargio (approved):

ret2libc (approved):

PR #1603, opened 2021-09-06
debug: generate coredump on netbsd.

devnexen:

### DIFF
--- a/librz/core/cmd_debug.c
+++ b/librz/core/cmd_debug.c
@@ -4823,7 +4823,7 @@ RZ_IPI int rz_cmd_debug(void *data, const char *input) {
 			rz_file_rm(corefile);
 			RzBuffer *dst = rz_buf_new_file(corefile, O_RDWR | O_CREAT, 0644);
 			if (dst) {
-				if (!core->dbg->cur->gcore(core->dbg, dst)) {
+				if (!core->dbg->cur->gcore(core->dbg, corefile, dst)) {
 					eprintf("dg: coredump failed\n");
 				}
 				rz_buf_free(dst);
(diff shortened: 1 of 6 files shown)


ret2libc (changes requested):
I think it's better to place a `#if` in rz_debug_native plugin, in the gcore method.

devnexen:
would be ideal but it is to avoid changing the plugin api calls (ie here not using the buffer file type as we do not use the file descriptor).

librz/core/cmd_debug.c
```diff
@@ -4830,6 +4831,11 @@ RZ_IPI int rz_cmd_debug(void *data, const char *input) {
 			} else {
 				perror("rz_buf_new_file");
 			}
+#else
+			if (ptrace(PT_DUMPCORE, core->dbg->pid, corefile, strlen(corefile)) == -1) {
+				eprintf("dg: coredump failed\n");
```
wargio: RZ_LOG_ERROR

wargio (changes requested):

ret2libc:
> would be ideal but it is to avoid changing the plugin api calls (ie here not using the buffer file type as we do not use the file descriptor).

That's a good point indeed. I'd still prefer having all the logic wrapped around in the right place though.
I see 2 options:
1) add a new `rz_buf_get_name` API to rz_buf that gets the name of a RzBuffer (for file-type RzBuf it would be the filename, for other things it can be less meaningful things....). Then use rz_buf_get_name inside gcore debug method to retrieve the filename and use it on NetBSD
2) change the debug plugin API to accept filename + RzBuffer and use the filename for netbsd. 

Probably the 2) is just 2 lines of change considering that only two plugins implement gcore. 1) would definitely be more involved. What do you think?

XVilka (approved):

ret2libc (changes requested):
I think there's also a win dbg plugin, can you double check?

yossizap (approved):

ret2libc (approved):

PR #1605, opened 2021-09-06
Replace 'obr' with 'obR' and allow reloc patching more than once 

Rot127:

**Detailed description**

`obr` should rebase the currently selected binary file to a specified address. At the moment it basically only sets the base address. All flags of the flag spaces and their offsets stay untouched.

This pull request simply maps the 'obr' command to
```command
> oba <current baddr> <new baddr>  # Load currently loaded RzBinFile to new baddr 
> ob-X # Delete currently selceted RzBinFile
```

Is it fine to have `obr` simply mapped to other commands?


ret2libc:
> Is it fine to have `obr` simply mapped to other commands?

Hi and thanks for your contribution! We are actually trying to avoid having commands that call other commands, because this results in very hard to debug/maintain code, considering that if `oba` or `ob-` changes syntax or arguments you have a very hard time finding all the occurrences that need to be changed. Instead, we should use the underlying API as much as possible.

If `obr` does not work I'd say there's definitely something in `rz_core_bin_rebase` that needs to be fixed (or maybe some other APIs that need to be called to propagate the updated data, I'm not sure). If you are willing to give it a try we can guide you!

librz/core/cmd_open.c
```diff
@@ -338,6 +334,18 @@ static void cmd_open_bin(RzCore *core, const char *input) {
...
 		rz_core_block_read(core);
 		break;
+	case 'r':; // "obr"
+		ut64 baddr = rz_bin_get_baddr(core->bin);
+		ut64 addr = rz_num_math(core->num, input + 2); // mapaddr
+		ut32 cur_bf_id = core->bin->cur->id;
+		char cmd[64] = "";
+
+		sprintf(cmd, "oba 0x%llx 0x%llx", baddr, addr);
+		rz_core_cmd0(core, cmd);
```
wargio: no need to use sprintf since there is a rz_core_cmdf

Rot127:
> If you are willing to give it a try we can guide you!

That would be great. I had a pretty hard time to understand the code. Especially because the ELF parsing code has a lot of macros which can not be tracked easily. That's also why I chose the lazy fix with the commands.

> If obr does not work I'd say there's definitely something in rz_core_bin_rebase that needs to be fixed.

I still think that the easiest solution is to simply open the RzBinFile at the given base address and close the old one. This will guarantee that all analysis steps will be applied.
If you agree with that I can push another version to this pull request.

Besides, is there somewhere a talk or document which describes the rizin <-> plugin architecture on a high level? I get pretty quickly lost in the source unfortunately and have a hard time finding out how the bin plugin interacts with the core.

ret2libc:
> > If you are willing to give it a try we can guide you!
> 
> That would be great. I had a pretty hard time to understand the code. Especially because the ELF parsing code has a lot of macros which can not be tracked easily. That's also why I chose the lazy fix with the commands.
> 
> > If obr does not work I'd say there's definitely something in rz_core_bin_rebase that needs to be fixed.
> 
> I still think that the easiest solution is to simply open the RzBinFile at the given base address and close the old one. This will guarantee that all analysis steps will be applied.
> If you agree with that I can push another version to this pull request.

It seems `obr` was added in https://github.com/radareorg/radare2/commit/40bc4cb751dddd21f4ab53e97a82b3ff984d5dfe and it was never tested (no coverage according to https://codecov.io/gh/rizinorg/rizin/src/ecb5d4472e05e1e945f12faef0deba8106bbbedd/librz/core/cmd_open.c ). So... As far as I know it could have been always broken (it seems there are also some out-of-bounds accesses detected by ASAN when using that command). Thus I'm ok to completely remove/replace `rz_core_bin_rebase` however I'm not sure about the open+close approach, as that would mean you operate on a new object instead that on the same one rebased... There could be other unseen problems with that as well. @thestr4ng3r might know more. 

That said, i'm noticing so many problems already with this `obr` command that likely having it as `oba`+`ob-` would be better than what we have now. cc @XVilka @wargio 

> Besides, is there somewhere a talk or document which describes the rizin <-> plugin architecture on a high level? I get pretty quickly lost in the source unfortunately and have a hard time finding out how the bin plugin interacts with the core.

Unfortunately, no. Documentation is, at best, lacking in many places :( For next release (0.4.0) I'm planning on focusing on rz-pm and plugins in general, so I will probably try to come up with some simple examples for various plugins, but we do not have anything at the moment. I suggest you configure a nice editor to navigate the source code (for example, vscode can easily solve all those `Elf_` macros that are in the elf code).

wargio:
think obr was used by java before my refactoring. i would say tho that is kinda not needed because if you need to reopen & rebase is better to close the session and restart it with different params, but anyway, i do agree with you, ret2libc.

Rot127:
>  i would say tho that is kinda not needed because if you need to reopen & rebase is better

I agree with you on that. Though the other changes regarding the reloc patching need to stay in this case
Using the `-m <base address>` option wouldn't patch the relocs with the given base address otherwise.

If no one disagrees I would simply remove the `obr` command. Is this fine?

XVilka:
Not sure about removal. Just recently @ret2libc also moved `ib` to the `obR`, it's somehow related to the `obr` in semantics, I think.
If we remove one, it's better to remove also the another.

ret2libc:
> Not sure about removal. Just recently @ret2libc also moved `ib` to the `obR`, it's somehow related to the `obr` in semantics, I think.
> If we remove one, it's better to remove also the another.

I think it's the opposite actually. If we kill one, we should ensure the other works. And from a very quick test it seems to me that `obR` works better than `obr`. So maybe we can indeed kill `obr` and update the description of `obR` to be `Reload the current buffer at another base address`. We could even go a bit further as to move the current `obR` to `obr` (i guess if we remove what is now `obr` there's no reason to keep the `R` instead of `r`). WDYT?

XVilka:
Yes, if remove `obr` and leave `obR` then the rename of `obR` to `obr` is the obvious choice. 👍

Rot127:
Technically this pull request has two changes.

1. The 'obr' command thing
2. Enabling reloc patching more than once.

Do you wish to have separated pull request for those?

librz/core/cmd_open.c
```diff
@@ -327,10 +326,6 @@ static void cmd_open_bin(RzCore *core, const char *input) {
 		break;
 	}
 	case 'r': // "obr"
-		rz_core_bin_rebase(core, rz_num_math(core->num, input + 3));
```
XVilka, line 330: The function `rz_core_bin_rebase` was used only here, I suggest to remove it completely. Moreover, I suggest to move `core_bin_reload()` function to the `librz/core/cbin.c` and making it an API - `RZ_API bool rz_core_bin_reload()`.

librz/core/cmd_open.c
```diff
@@ -327,10 +326,6 @@ static void cmd_open_bin(RzCore *core, const char *input) {
...
-		rz_core_bin_apply_all_info(core, rz_bin_cur(core->bin));
-		break;
-	case 'R': // "obR"
 		// XXX: this will reload the bin using the buffer.
 		// An assumption is made that assumes there is an underlying
 		// plugin that will be used to load the bin (e.g. malloc://)
```
XVilka, line 314: Remove this `TODO` below then

test/db/analysis/java
```diff
@@ -169,15 +169,15 @@ e asm.lines.bb=false
...
 oba 0xa00
 pdf
-obR
+obr
```
XVilka, line 180: Please add a test with some binary where `obr` has numerical argument.

XVilka (changes requested):

librz/core/cmd_open.c
XVilka: Better to the `rz_core_file_reload()` and into `librz/core/cfile.c`, since it operates on the file, not `RzBin`.

librz/core/cbin.c
```diff
@@ -368,6 +368,23 @@ RZ_API bool rz_core_bin_apply_all_info(RzCore *r, RzBinFile *binfile) {
 	return true;
 }
 
+RZ_API bool rz_core_file_reload(RzCore *r, const char *file, ut64 baseaddr) {
```
XVilka, line 371: Could you please add a Doxygen for this function? Also, I think it's better to call the first parameter as `RzCore *core` for clarity. 
- https://github.com/rizinorg/rizin/blob/dev/DEVELOPERS.md#documentation

XVilka (changes requested):
Overall looks good, please see my comment, also a test is required.
Moreover, I suggest to squash all changes in two commits - one is the ELF relocation changes, and the second is the `obr` thing.

librz/core/cbin.c
ret2libc, line 371: This function should also be added to rz_core.h. Also, given it's a "reload" there should be no `file` argument (it is not even used) because it is assumed the current file is reloaded, i guess. Maybe, instead of passing the filename you could directly pass `RzBinFile` and move:
```
RzCoreFile *cf = rz_core_file_cur(r);
	if (!cf) {
		return false;
	}
	RzBinFile *obf = rz_bin_file_find_by_fd(r->bin, cf->fd);
	if (!obf) {
		return false;
	}
```
to the caller of this API. This is slightly better from an API point of view, in my opinion, because it allows you to reuse this API to also reload binaries that are not the current one.

Also, I think the new RzBinFile should be returned by this function instead of bool.

ret2libc (changes requested):
As you probably discovered there is also something wrong with patch relocations...

Rot127:
Thanks for all your feedback!
I will take a look at it during the next week.

Rot127:
Since the issue is also discussed here: https://github.com/rizinorg/rizin/issues/1651 I will close this one.
Rebasing seems to need a rework and should be separated from the relocation patching problem.

PR #1338, opened 2021-07-20
Hexagon update

Rot127:
**Your checklist for this pull request**
- [x] I've read the [guidelines for contributing](https://github.com/rizinorg/rizin/blob/master/DEVELOPERS.md) to this repository
- [x] I made sure to follow the project's [coding style](https://github.com/rizinorg/rizin/blob/master/DEVELOPERS.md#code-style)
- [x] I've added tests that prove my fix is effective or that my feature works (if possible)
- [ ] I've updated the documentation and the [rizin book](https://github.com/rizinorg/book) with the relevant information (if needed)

**Detailed description**

This is an update of the hexagon `asm` and `analysis` plugin. It used the LLVM source to generate the plugin code.

This pull request only synchronizes the source code with the `rz-hexagon` repo.
For more details please refer to the pull request over there: https://github.com/rizinorg/rz-hexagon/pull/4

**Test plan**

**asm tests**
The tests for the `asm` plugin are updated as well, since some of the old ones where incorrect.
The `asm` tests in this pull requests were written and disassembled with the Hexagon IDE tools.

**analysis tests**
Those tests got not updated (unfortunately my dev time is up for this month) and fail at the moment. I confirmed the correct working of the hardware loops, jumps and calls manually.

**Closing issues**
closes https://github.com/rizinorg/rz-hexagon/issues/13

None

Issue #1287, opened 2021-07-09
Add vector and control register type


Rot127:
**Is your feature request related to a problem? Please describe.**

The Hexagon architecture has vector registers (for Hexagon HVX) and control registers (hardware loop counts, frame limits etc.). The register types in [reg.h](https://github.com/rizinorg/rizin/blob/dev/librz/include/rz_reg.h) have non of those two.

**Describe the solution you'd like**
Add the vector and control register as type to `reg.h` and `reg.c`.

**Describe alternatives you've considered**
Non really. Maybe reusing the x86 special types?

**Additional context**
I rework the Hexagon disassembler at the moment and have trouble to decide on a register type for the register profile.
If this solution is fine, I will create a pull request.

**Edit**: There are also Guest registers and Predicate registers.
Basically I would like to know, whether a pull request with X many new register types would be accepted.

Issue #1141, opened 2021-05-17
Split off Relocs Code from bin_elf.inc


thestr4ng3r:
From https://github.com/rizinorg/rizin/pull/1079#discussion_r632270049

Conceptually `librz/bin/p/bin_elf.inc` should implement the side towards the RzBin plugin interface, while standalong ELF parsing should mostly be in `librz/bin/format/elf` instead.
Relocs handling and patching is quite a chunk of code and it is currently implemented in `librz/bin/p/bin_elf.inc`, but a lot of it falls more in the category of ELF parsing so it would fit better in `librz/bin/format/elf`, in a new file `elf_reloc.c`.
What is a bit tricky here is that reloc patching somewhat depends on the generated final maps to know where to point relocs to (`reloc_targets_map_base()`).

XVilka:
There are hundreds of the relocations missing as well, we should add all supported by the Binutils.

PR #1614, opened 2021-09-07
Update Hexagon asm and analysis plugins

### LINKED ISSUE #13: RzAnalysisOp.val won't get set
The analysis code does not set the `RzAnalysisOp.val` member. This is especially problematic, because the `val` member is used when one uses `/ai` to search for immediate operands.

It should be set to the immediate value of the instruction.

### LINKED ISSUE #15: Disassemble 0x00000000 to <unknown>; not to a valid dublex instruction.
At the moment the disassembler maps `0x0000000` to `R0 = memw(R0+#0x0) ; R0 = memw(R0+#0x0)`.

It should be `<unknown>` or `<invalid>`.

### LINKED ISSUE #17: Omit packet prefix if only one instruction is disassembled
In case only one instruction is disassembled the packet prefix should be omitted.

```
[0x00005000]> pi 1
\ jump 0x1209c
```
should become
```
[0x00005000]> pi 1
jump 0x1209c
```

To do this we have to track somehow whether a whole block is disassembled or only one instruction.
Also see the discussion here: https://github.com/rizinorg/rizin/pull/1338#issuecomment-898234777

### LINKED ISSUE #18: Track instruction packet membership in more detail
Hexagon instructions can belong to instruction packets. To mark those packet the assembly gets prefixed (with `/,\,|`) like that:

```c
; two instruction packets.
/ immext(##0x1c00)
| R11 = ##loc.start_pc
| immext(##0x1c00)
\ R12 = ##loc.start_sp
/ R15 = asl(R15,R2)
\ memw(R13+R2<<#0x2) = R5
```

Here is how the plugin assigns the correct packet prefix to the assembly:

- The last instruction in a packet has certain bits set. (See reference manual v67 Chapter 10.5).
  - Prefix it with `\`, set `new_pkt_starts = true`
- A new packet starts (because the previous instruction was the last instruction in its packet.)
  - The next instruction is prefixed with a `/`
- Each following instruction is prefixed wit a `|`, if it is not the last instr. nor the first one.

The logic is implemented in [hex_set_pkt_info()](https://github.com/rizinorg/rizin/blob/3df40c999560e6126a8006197cd88572fd01f0fb/librz/asm/arch/hexagon/hexagon.c#L626).

Unfortunately this causes some problems.
Assume the correctly disassembled instruction packet:

```c
> s 0x00005170
> pi 4
0x00005170      /     immext(##0x1c00)    ;  #1 in packet
0x00005174      |     R11 = ##loc.start_  ;  #2 in packet
0x00005178      |     immext(##0x1c00)    ;  #3 in packet
0x0000517c      \     R12 = ##loc.start_  ;  #4 in packet
``` 
now do:

```c
> s 0x00005174
> pi 1
/     R11 = ##loc.start_  // Is marked as #1, but is #2 in packet
```

**Bug**: `R11 = ##loc.start_` should be prefixed with a `|` but has a `/` because the last disassembled instruction (`R12 = ##loc.start_ ` at `0x0000517c`) was marked as the last packet in the packet.

To solve this the plugin somehow needs to peek into the previous instruction in memory. This way it could check, whether it is indeed the first one.

### LINKED ISSUE #19: Simpler search for negative immediates
Searching for negative immediate values via `/ai` only works if they are written as unsigned int.
E.g.: to search for `-0x8` type: `/ai 0xfffffff8`

It would be convenient to simply type `/ai -0x8`.

### LINKED ISSUE #21: Easy way to import undocumented and missing instructions
LLVM has not all instructions defined. E.g.: all system/Monitor instructions are missing (chapter 11.9.2 in the programmers reference manual v67)

It is also possible that there are some undocumented ones we would like to include.

Maybe add a `.json` file which has all those instructions?

### LINKED ISSUE #22: Add Doxygen documentation to functions

### LINKED ISSUE #24: Track the state and context of the asm and analysis plugin.
The Hexagon instructions have two properties which depend on the context they are in:

**1.** An instructions immediate value can be extended, if the previous instruction is an `immext()` instruction.
Example:
```asm
s 0x00005174
pi 1
0x00005174      ?     R0 = ##0x10  ; Immediate = 0x10

s 0x00005170
pi 2
0x00005170      ?     immext(##0x100)  ; 0x100 is added to the next immediate
0x00005174      ?     R0 = ##0x110  ; Immediate = 0x110
```

**2.** The position of an instruction in an instruction packet depends on the prior instruction.

Rizin marks the position of an instruction in a packet like this:

```asm
0x0000516c      [     R1 = ##0x77         ;  #1 and last in packet
0x00005170      /     immext(##0x1c00)    ;  #1 in packet
0x00005174      |     R11 = ##loc.start_  ;  #2 in packet
0x00005178      |     immext(##0x1c00)    ;  #3 in packet
0x0000517c      \     R12 = ##loc.start_  ;  #4 in packet
```

The position is determined by the parsing bits in the instruction opcode.

```asm
0x0000516c      [     R1 = ##0x77         ;  Parsing bits = 0b11 => Last
0x00005170      /     immext(##0x1c00)    ;  Parsing bits = 0b01 => Not last. But first in new packet because prev. instr. was last
0x00005174      |     R11 = ##loc.start_  ;  Parsing bits = 0b01 => Not last
0x00005178      |     immext(##0x1c00)    ;  Parsing bits = 0b01 => Not last
0x0000517c      \     R12 = ##loc.start_  ;  Parsing bits = 0b11 => Last
```

In both cases the asm plugin needs to know the three previous instructions in the binary to print the correct assembly.
_But this is a problem in `rizin`._

Rizin disassembles an instruction by simply giving four bytes to the `disassemble()` function. The `disassemble()` function has no way to peek into the previous instructions in the binary.
The same happens during analysis. The analysis function (`hex_op6_op()`) will get four bytes which it passes to `disassemble()` and adds some attributes afterwards.

**Problem**

At the moment _static variables_ track the new packet start (`new_pkt_starts`), the previous address (`previous_addr`) and the last `constant_extender`.

Though this becomes a problem if the analysis and disassembly function are *not* called sequentially. The analysis function can be called on offset `0xffffee`, the disassembly function on offset `0x0c`.

The _static variables_ will hold inconsistent values as a consequence.
(Inconsistent static variable content was also the reason for the failing tests in https://github.com/rizinorg/rizin/pull/1614).

**Possible solutions I can think of**

- Add the ability to peek into the previous three instructions in the binary (is against the rizin architecture).
- Add some kind of struct to the plugins which track the last instructions disassembled by the `analysis` and `asm` plugin separately (needs more memory and quite a lot of `memcpy()`).
- Peek into the buffer given to `disassemble()`, not into the binary. Problem with instructions üackets which span over multiple block sizes.

Rot127:
Replaces https://github.com/rizinorg/rizin/pull/1338

**Detailed description**

Updates Hexagon `analysis` and `asm` plugin.

**Supports**
- Add `asm` support for v5-v68 and HVX instructions
  - Recognize instruction packets.
- Basic analysis support
  - set instruction type (return, call, jumps hardware loop etc.)
  - Enables search for immediate operands.
- Patching of relocs 

**Not yet implemented**
- ~~System registers~~
- ~~[System/Monitor instructions](https://github.com/rizinorg/rz-hexagon/issues/21)~~
- [Can not analyse >1024bit vector registers](https://github.com/rizinorg/rizin/issues/1287)
- No ESIL support
- [~~No tests yet for HVX instructions and hardware loops.~~](https://github.com/rizinorg/rz-hexagon/issues/8)
- Test files for reloc patching

**Test plan**

See test files.

**Closing issues**

closes https://github.com/rizinorg/rz-hexagon/issues/13
closes https://github.com/rizinorg/rz-hexagon/issues/15
closes https://github.com/rizinorg/rz-hexagon/issues/17
closes https://github.com/rizinorg/rz-hexagon/issues/18
closes https://github.com/rizinorg/rz-hexagon/issues/19
closes https://github.com/rizinorg/rz-hexagon/issues/21
closes https://github.com/rizinorg/rz-hexagon/issues/22
closes https://github.com/rizinorg/rz-hexagon/issues/24


Rot127:
I have absolutely no idea what the cause of the failing test is. I just created a VM and built rizin as the ubuntu-tcc-test Action does it. Basically copied and pasted the commands to build TinyCC and meson.

All tests succeeded. No error.

The only difference is, that I use a debian-10 in the VM. No Ubuntu. Unfortunately I can't install a Ubuntu VM at the moment and my enthusiasm debugging it is meanwhile really low also.

Do you have any advice how to proceed?

Just in case I'm a blockhead. The commands I compiled rizin with TinyCC:

```
git clone https://repo.or.cz/tinycc.git
cd tinycc
#git checkout mob
# Well-tested commit
git checkout 675046bd59bc6977bb2016c7d2e115ace8a6ae6c
chmod +x ./configure
./configure --prefix=/usr
make
sudo make install
sudo apt-get --assume-yes install python3-wheel python3-setuptools python3-pip
sudo python3 -m pip install ninja
git clone https://github.com/ret2libc/meson.git
cd meson
git checkout tiny-cc
python3 setup.py build
sudo python3 setup.py install
cd ~
git clone https://github.com/Rot127/rizin.git
cd ~/rizin/
git checkout Hexagon
env CC=tcc meson --prefix=/usr build
ninja -C build && sudo ninja -C build install
mkdir -p test/bins/elf/analysis/
cd test/bins/elf/analysis/
wget https://raw.githubusercontent.com/rizinorg/rizin-testbins/master/elf/analysis/hexagon-hello-loop
cd ~/rizin
rz-test test/db/analysis/hexagon 

Running from /home/user/rizin/test
Loaded 3 tests.
Skipping json tests because jq is not available.
[3/3]                       3 OK         0 BR        0 XX        0 FX
Finished in 0 seconds.
```

Rot127:
So I ran the tests on an Ubuntu 20.04.3 LTS VM and they succeeded. It must be something with Github Actions than.

wargio:
> So I ran the tests on an Ubuntu 20.04.3 LTS VM and they succeeded. It must be something with Github Actions than.

also with tinycc?

Rot127:
Oh, I was stupid. Sorry.
Ran `meson --prefix=/usr build` instead of `env CC=tcc meson --prefix=/usr build`.
Now the tests fail.

Will debug it once I have more time.

wargio (approved):

librz/bin/p/bin_elf.inc
```diff
@@ -387,13 +387,323 @@ static ut64 reloc_targets_vfile_size(RzBinFile *bf, ELFOBJ *obj) {
...
+			//	val = (G);
+			//	break;
+			//case R_HEX_LD_GOT_16:
+			//	word = 4;
+			//	val = (G);
+			//	break;
+			//case R_HEX_LD_GOT_32_6_X:
+			//	word = 4;
+			//	val = (G) >> 6;
+			//	break;
```
ret2libc: remove if not needed.

ret2libc (changes requested):
I haven't really followed here... but if @wargio things is good, I'm ok to merge this as it is anyway quite confined to hexagon code. Please address just this thing first.

librz/analysis/p/analysis_hexagon.c
```diff
@@ -1,6 +1,12 @@
-// SPDX-FileCopyrightText: 2018 xvilka <anton.kochkov@gmail.com>
+// SPDX-FileCopyrightText: 2021 Rot127 <unisono@quyllur.org>
+//
```
XVilka: Remove this empty line please.

librz/analysis/arch/hexagon/hexagon_analysis.h
```diff
@@ -1,4 +1,10 @@
-// SPDX-FileCopyrightText: 2018-2021 Anton Kochkov <anton.kochkov@gmail.com>
+// SPDX-FileCopyrightText: 2021 Rot127 <unisono@quyllur.org>
+//
```
XVilka: Remove this empty line please

librz/asm/arch/hexagon/hexagon.c
```diff
@@ -1,237 +1,987 @@
-// SPDX-FileCopyrightText: 2018-2021 Anton Kochkov <anton.kochkov@gmail.com>
+// SPDX-FileCopyrightText: 2021 Rot127 <unisono@quyllur.org>
+//
```
XVilka: Remove this empty line please

librz/asm/arch/hexagon/hexagon.c
```diff
@@ -1,237 +1,987 @@
...
-void hex_op_extend_off(HexOp *op, int offset) {
-	if ((constant_extender != 1) && (op->type == HEX_OP_TYPE_IMM)) {
-		op->op.imm = (op->op.imm) >> offset;
-		hex_op_extend(op);
+static inline bool imm_is_scaled(const HexOpAttr attr) {
+	return (attr & HEX_OP_IMM_SCALED);
+}
+
+/**
+ * @brief Applies the last constant extender to the immediate value of the given HexOp.
```
XVilka: Please use `\brief` syntax everywhere: https://github.com/rizinorg/rizin/blob/dev/DEVELOPERS.md#documentation

librz/asm/arch/hexagon/hexagon.h
```diff
@@ -1,9 +1,16 @@
-// SPDX-FileCopyrightText: 2018-2021 Anton Kochkov <anton.kochkov@gmail.com>
+// SPDX-FileCopyrightText: 2021 Rot127 <unisono@quyllur.org>
+//
```
XVilka: Remove this empty line please

librz/asm/arch/hexagon/hexagon.h
```diff
@@ -46,237 +51,521 @@ typedef enum {
...
+		st64 imm;
 	} op;
 	ut8 attr;
+	ut8 shift;
 } HexOp;
 
 typedef struct {
 	int instruction;
 	ut32 mask;
-	HexPred predicate; // predicate set if set
```
XVilka, line 66: Why are predicates no longer in use? They are pretty common in the code.

librz/bin/p/bin_elf.inc
```diff
@@ -387,13 +387,323 @@ static ut64 reloc_targets_vfile_size(RzBinFile *bf, ELFOBJ *obj) {
...
+static void patch_reloc(struct Elf_(rz_bin_elf_obj_t) * obj, RzBinElfReloc *rel, ut64 S, ut64 B, ut64 L, ut64 GOT) {
 	ut16 e_machine = obj->ehdr.e_machine;
-	ut64 val;
+	ut64 val = 0;
 	ut64 A = rel->addend, P = rel->vaddr;
 	ut64 patch_addr = rel->paddr != UT64_MAX ? rel->paddr : Elf_(rz_bin_elf_v2p_new)(obj, rel->vaddr);
 	ut8 buf[8];
 	switch (e_machine) {
+	case EM_QDSP6:;
+		ut8 word = 0;
```
XVilka: Can you please extract the reloc patching for Hexagon in a separate function? We are working on improving and refactoring relocations patching in ELF and it will make that easier in the future:
- https://github.com/rizinorg/rizin/pull/1699
- https://github.com/rizinorg/rizin/issues/1141

librz/bin/p/bin_elf.inc
```diff
@@ -387,13 +387,323 @@ static ut64 reloc_targets_vfile_size(RzBinFile *bf, ELFOBJ *obj) {
...
+		ut64 bitmask = UT64_MAX; // Mask for bits in the opcode which get patched.
+		switch (rel->type) {
+		default:
+			eprintf("Patching for reloc type %d not implemented.", rel->type);
```
XVilka: Use `RZ_LOG_ERROR()`

XVilka (changes requested):
Overall looks good. What bothers me - why you don't use predicates? They are pretty important.
Also, for relocations - could you please add a test specifically for relocations? Like it's done in https://github.com/rizinorg/rizin/pull/1699/files#diff-815fd301b119045a3c4aef6ecbb05174a1d997c1c43cf2b65d4a461ea57bd7b4R114
So for every different type of relocation you implemented, if it's in the test binary - print bytes, print instruction. It's recommended to cross-check your results against `readelf`, `objdump`, IDA Pro, Binary Ninja, Ghidra.

Good job!

librz/asm/arch/hexagon/hexagon.h
Rot127: Oh, I quite frankly forgot about them. Will implement as soon as possible.

Rot127:
@XVilka 

> Good job!

Thank you. I really appreciate the work you guys put into reviewing this.

> What bothers me - why you don't use predicates? They are pretty important.

I quite frankly forgot about them. Will start with the implementation today.

Can you give me a hint where rizin uses this information? I can't find a member in `RzAnalysisOp` which represents predicates properly. But maybe my understanding is limited.

> It's recommended to cross-check your results against readelf, objdump, IDA Pro, Binary Ninja, Ghidra.

| Tool | comment |
|-------|--------------|
| objdump |  I checked the disassembly and analysis test against the SDKs `objdump`. And actively advice it in the test documentation in the rz-hexagon docs. 
| IDA Pro | I can't effort a license for IDA Pro, therefore not test against it. That's the reason why I started developing the Hexagon plugin for rizin in the first place. |
| Binary Ninja | Same like IDA Pro. I can't effort a license and the demos can't load the [Hexagon plugin](https://github.com/google/binja-hexagon). |
| Ghidra | [Ghidra does not support the Hexagon architecture yet](https://github.com/NationalSecurityAgency/ghidra/issues/1595). 

The problem for me is, that tests implementations need a lot of time which I currently not have (e.g. compiling HVX code with the SDK is buggy, building a ELF binary with all implemented relocations is quite hard, as I just start learning about relocations details etc.).
At the moment I can, unfortunately, only promise that I will add them gradually once I use it more myself. For me it is absolutely fine if this PR is not merged yet for this reason. The plugin will be under development at least for the next 5-6 months anyways.

Rot127:
Almost forgot. Testing the reloc patching is quite hard at the moment since the rebasing is so broken.

XVilka:
@Rot127 do you mean this one? https://github.com/rizinorg/rizin/issues/1661
cc @ret2libc

Rot127:
> @Rot127 do you mean this one? #1661

Yes. It does correctly patch if the binary is loaded with `rizin -m <baddr>`. But because of #1661 and building a binary which uses all implemented relocations is quite time consuming I postponed the tests.

librz/bin/p/bin_elf.inc
```diff
@@ -387,13 +387,235 @@ static ut64 reloc_targets_vfile_size(RzBinFile *bf, ELFOBJ *obj) {
 	return Elf_(rz_bin_elf_get_relocs_count)(obj) * reloc_target_size(obj);
 }
 
-static void patch_reloc(struct Elf_(rz_bin_elf_obj_t) * obj, RzBinElfReloc *rel, ut64 S, ut64 B, ut64 L) {
+static void patch_reloc_hexagon(RZ_INOUT RzBuffer *buf_patched, const int rel_type, const ut64 patch_addr, const ut64 S, const ut64 B, const ut64 L, const ut64 GOT, const ut64 A, const ut64 P) {
```
XVilka: Instead of passing so many parameters I suggest just create some context structure to group them into one or a few structures and pass them instead.

XVilka (changes requested):

librz/include/rz_bin.h
```diff
@@ -744,6 +744,20 @@ struct rz_bin_reloc_storage_t {
 	size_t target_relocs_count;
 }; // RzBinRelocStorage
 
+typedef struct rz_bin_reloc_formular_symbols_t {
```
XVilka: I don't think it should be in the `rz_bin.h`. Could you please move it back to the `bin_elf.inc`? It's ELF-specific, not needed for anything else.

XVilka (changes requested):

XVilka:
Please also rebase on top of the latest `dev` to resolve the conflicts.

librz/asm/arch/hexagon/hexagon.h
Rot127: Fixed with https://github.com/rizinorg/rizin/pull/1614/commits/1435be081c4e3265aa820706993112eb78670499

Rot127: Although https://github.com/rizinorg/rz-hexagon/issues/29 still persists.

Rot127:
`TinyCC build` test fails because of https://github.com/rizinorg/rz-hexagon/issues/24.
Besides, `tcc-builds` fail on real world binaries. The static variables get always incorrectly set. But I can't find the issue.

It seems to me that this is a pretty annoying `tcc` bug. Do you have any advice how to proceed with this? Guess it can't be merged if it does not work with `tcc`?

XVilka:
@Rot127 usually TinyCC devs fix the error when reported provided you can minimize the reproducer. See their mailing list:
- https://lists.nongnu.org/mailman/listinfo/tinycc-devel
- https://lists.nongnu.org/archive/html/tinycc-devel/

Then, once they fixed we can bump the used commit in [.github/workflows/tcc.yml:49](https://github.com/rizinorg/rizin/blob/dev/.github/workflows/tcc.yml#L49).

Rot127:
https://github.com/rizinorg/rizin/pull/1614/commits/52d6deb1645f3682dd941c31ac6e4180c2862aed changes the architecture of the plugin pretty substantially.
It closes https://github.com/rizinorg/rz-hexagon/issues/24 and is hopefully useful for the future (if RzArch is introduced). 

An opcode is now reversed by only one function (in hexagon_disas.c).
The flow graph would look a bit like that:

```
                                   <--------- writes RzAsmOp ---------------------------------------+
rz_do_disas()-----+                                                                                  |
                  |----> reverse_opcode() -----> |add empty insn to global state| -----> do_disas_and_analysis(insn)
rz_do_analysis()--+                                                                                  |
                                  <------- writes RzAnalysisOp--------------------------------------+
```

Also added in this commit:
- Fix: `Rs.new` registers get recognized correctly. 
- Fix: Jump offsets are now added to the packet address, not the address of the instruction itself (which was incorrect).
- Simple jumps are now marked as block end.

XVilka:
@wargio @ret2libc @thestr4ng3r please take a look in regard of keeping the same state between analysis and disassembly plugins.

wargio:
Damn.. we really need RzArch

Rot127:
Is it a problem that the `hexagon_disas.c` is so large that gcc needs >8GB of RAM to build rizin?
My VM fails to build it, if it has 8GB assigned. Need to add a swap file with additional 4GB.

Building with tcc obviously works fine. Besides the issues with the tests.

XVilka:
Yes, this is a problem and also a bug in GCC. I recommend reporting it to https://gcc.gnu.org/bugzilla/
Meanwhile, the solution would be to split this file. BTW, how big it is?

Rot127:
> BTW, how big it is?

It is 4.4M and will get 30% to 50% bigger if we add more info to each instruction in the future (RZIL, memory access info etc.). Splitting it now should make each file ~0.14M big

> Meanwhile, the solution would be to split this file.

This pull request gets bigger every day. Three months ago I just wanted to add the new v68 instructions :smiley:  
Anyways. Will split them.

thestr4ng3r:
> It is 4.4M

That is quite large, even for generated code. Do you think it would be possible to optimize the code generator, for example splitting the individual instruction in different categories that will then use the same code for certein instruction properties and can thus avoid code duplication.

Rot127:
> for example splitting the individual instruction in different categories

Not without reworking it. The meaning of each bit at position i differs from opcode to opcode (sometimes pretty significantly). There is certainly a logic to it, but it isn't publicly documented. So it probably would need some reversing and al lot more testing.

That's why the generator simply iterates over all instruction defintions of the LLVM and sets the struct members accordingly.

But: [this](https://github.com/n-o-o-n/idp_hexagon) IDA plugin implements the logic. But it very much looks like the person who develops it has access too proprietary Qualcomm documentation.

thestr4ng3r:
Hmm I see. What about, instead of generating code for each property, generating a lookup table in C for some properties? That would shift a bit of the work to the runtime, but probably also compress the entire thing a bit.
By the way, what is the code size after the file has been compiled?

Rot127:
> By the way, what is the code size after the file has been compiled?

`ls -l` for `./build/librz/analysis/librz_analysis.a.p/.._asm_arch_hexagon_hexagon_disas.c.o` gives me:
| compiler | target | size |
|---|----|----| 
| tcc | x86_64 |  5.3M |
| gcc | x86_64 | 1.9M |
| clang | aarch64 | 1.4M |

Rot127:
> What about, instead of generating code for each property, generating a lookup table in C for some properties?

Mh, I will ponder about this a bit. But for now I will remove unnecessary function calls from it (~3 per instruction). This should drive the memory usage of gcc down quite a bit.

@thestr4ng3r Just in case you have time and want to take a quick look at it yourself, in `librz/asm/arch/hexagon/hexagon_disas.c @ line 56088` is an instruction which is pretty complex and has a lot of assignments. Maybe you see something right away.

Rot127:
It's weird. But I no longer get the problem with >4gb memory usage of gcc (top shows something around 16MB).
So if you like, you can review it.

XVilka:
There are a lot of unrelated changes. Thus, I propose:
1. Squash all commits into one - this way rebasing should be easier
2. Rebase on top of the latest `dev`
3. Fix `clang-format` warning:
```
Ubuntu clang-format version 13.0.1-++20211110063012+9dc7d6d5e326-1~exp1~20211110183538.22
librz/asm/arch/hexagon/hexagon_disas.c:18:26: error: code should be clang-formatted [-Wclang-format-violations]
#include "hexagon_arch.h"
                         ^
```

XVilka (approved):
Looks good to merge. Please use the better commit name when merging.

wargio (approved):

XVilka:
Merge if https://github.com/rizinorg/rizin/pull/2033 is green (excluding the known debug test ASAN failure)

XVilka:
Fails to build on Windows AppVeyor:
```
FAILED: librz/asm/librz_asm.a.p/arch_hexagon_hexagon_arch.c.obj 
"cl" "-Ilibrz/asm\librz_asm.a.p" "-I." "-I.." "-I..\librz\include" "-Ishlr" "-I..\shlr" "-I..\librz\asm\arch\include" "-I..\librz\asm\arch" "-I..\librz\asm\arch\h8300" "-I..\librz\asm\arch\hexagon" "-I..\librz\asm\arch\msp430" "-I..\librz\asm\arch\rsp" "-I..\librz\asm\arch\mcore" "-I..\librz\asm\arch\v850" "-I..\librz\asm\arch\propeller" "-I..\librz\asm\arch\ebc" "-I..\librz\asm\arch\cr16" "-I..\librz\asm\arch\8051" "-I..\librz\asm\arch\v810" "-I..\librz\asm\arch\or1k" "-Isubprojects\sdb" "-I..\subprojects\sdb" "-Isubprojects\sdb\src" "-I..\subprojects\sdb\src" "-I..\librz\bin\format" "-I..\subprojects\capstone-bundled\include" "-I..\subprojects\capstone-bundled\include\capstone" "-I..\shlr\spp" "/MT" "/nologo" "/showIncludes" "/utf-8" "/W2" "/O2" "/Gw" "-DRZ_PLUGIN_INCORE=1" "/Fdlibrz/asm\librz_asm.a.p\arch_hexagon_hexagon_arch.c.pdb" /Folibrz/asm/librz_asm.a.p/arch_hexagon_hexagon_arch.c.obj "/c" ../librz/asm/arch/hexagon/hexagon_arch.c
../librz/asm/arch/hexagon/hexagon_arch.c(613): error C2375: 'hex_extend_op': redefinition; different linkage
c:\projects\rizin\librz\asm\arch\hexagon\hexagon.h(550): note: see declaration of 'hex_extend_op'
[1116/1727] Compiling C object librz/asm/librz_asm.a.p/arch_h8300_h8300_disas.c.obj
[1117/1727] Compiling C object librz/asm/librz_asm.a.p/arch_hexagon_hexagon_disas.c.obj
```
https://ci.appveyor.com/project/rizinorg/rizin/builds/41680793/job/6m3ufxlgdbrqrbi0

XVilka:
Failing test with ASAN:
```


[XX] db/analysis/hexagon hexagon jumps
RZ_NOPLUGINS=1 rizin -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -Qc 'b 0x100000
e analysis.jmp.cref=true
s sym.main
aaa
afx
' bins/elf/analysis/hexagon-hello-loop
-- stdout
--- expected
+++ actual
@@ -5,3 +5,7 @@
 C 0x0000513c -> 0x000050f8 [   call sym.pWorld
 c 0x00005140 -> 0x00005144 [   jump 0x5144
 c 0x00005150 -> 0x00005128 ?   jump 0x5128
+WARNING: rz_reg_set_value: assertion 'reg && item' failed (line 145)
+WARNING: rz_reg_set_value: assertion 'reg && item' failed (line 145)
+WARNING: rz_reg_set_value: assertion 'reg && item' failed (line 145)
+WARNING: rz_reg_set_value: assertion 'reg && item' failed (line 145)

-- stderr
Cannot determine entrypoint, using 0x00005000.
[ ] Analyze all flags starting with sym. and entry0 (aa)
```

Rot127:
https://github.com/rizinorg/rizin/pull/1614/commits/6942f6c9083f5f2347f697b95d93ae09f10b0a88 should fix the Windows build.

XVilka:
The other PR was green.

Issue #1592, opened 2021-09-04
ELF gnu_hash not correctly detected


ret2libc:
### Work environment

| Questions                                            | Answers
|------------------------------------------------------|--------------------
| OS/arch/bits (mandatory)                             | Fedora 33
| File format of the file you reverse (mandatory)      | ELF
| Architecture/bits of the file (mandatory)            | x86/64
| `rizin -v` full output, **not truncated** (mandatory)  |  rizin 0.3.0-git @ linux-x86-64 commit: aed6854bdfaa34ca56ffc35334a5006113fc84b1, build: 2021-09-04__12:05:20

### Expected behavior
No warnings for a regularly compiled binary.
[test.zip](https://github.com/rizinorg/rizin/files/7109582/test.zip)

### Actual behavior
```
WARNING: Neither hash nor gnu_hash exist. Falling back to heuristics for deducing the number of dynamic symbols...
WARNING: Neither hash nor gnu_hash exist. Falling back to heuristics for deducing the number of dynamic symbols...
WARNING: Neither hash nor gnu_hash exist. Falling back to heuristics for deducing the number of dynamic symbols...
```

### Steps to reproduce the behavior
Compile the following program:
```
#include <stdio.h>

int main(int argc, char **argv) {
	printf("%s\n", argv[1]);
	return 0;
}
```
Then
```
$ rizin /tmp/test
```

ret2libc:
This happens when there are no defined symbols. gnu creates this "ill formed" table so it avoids doing any operation on it at all. Thus when this happens, relying on the gnu hash table to determine the number of entries in dynsym cannot be done. We can instead rely on relocations and find out what is the highest symbol found there.

ret2libc:
I think this can be implemented correctly and with no rush after 0.3.0, maybe added for 0.3.1 if there will be. We probably just need to make the error messages less noisy or clearer for now. After all, the situation is *at least better* than before. cc @XVilka @wargio WDYT?

wargio:
ok for me

XVilka:
Sounds good to me.

XVilka:
Can we just remove the warning for now?

ret2libc:
Yes, that warning needs to be improved for sure

ret2libc:
I didn't see any work on this during this time, so I doubt it will be done for 0.4.0.
I'm removing the 0.4.0 milestone for now until someone starts to actively looking at this.

cc @XVilka

ret2libc:
[...]

PR #1615, opened 2021-09-07
Use RZ_LOG_INFO instead of RZ_LOG_WARN for hash/gnu_hash sections

ret2libc:
Right now we don't correctly handle some ELF cases where the gnu_hash
section is ill-formed to allow the LD loader to skip it quickly (e.g.
when no symbols are defined in the binary). In such cases, we should not
bombard the user with warnings that may confuse them, considering the
situation has just improved compared to the last release.

However, we want in the future to properly count dynamic symbol entries
by using relocations. Until then though, the users don't want to see
warnings even for regular binaries.

Refer to #1592

### DIFF
--- a/librz/bin/format/elf/elf_symbols.c
+++ b/librz/bin/format/elf/elf_symbols.c
@@ -377,7 +377,7 @@ Elf_(Word) Elf_(rz_bin_elf_get_number_of_dynamic_symbols)(RZ_NONNULL ELFOBJ *bin
 		return result;
 	}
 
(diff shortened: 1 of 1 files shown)


XVilka (approved):

ghost:
I would have agree with this if the default log level was warning. But in release mode the log level is error. So I don't understand the down grade. I mean that this message should be displayed in debug mode by default.

ret2libc:
You are right @08A , I didn't notice that! Thanks for the review. No need for this anymore ;)

PR #1616, opened 2021-09-07
Add `RZ_BUILD_DEBUG` compile time definition for debug logging

### LINKED ISSUE #1612: RZ_LOG_DEBUG should not be compiled in releases mode
While debugging/developing you may want to add a lot of logs that can be enabled with `e log.level=X` where X is a number for 0 to 6. However, when we release Rizin we likely do not want those DEBUG prints to ever be printed, saving in some cases a lot of useless printing and calculations.

When buildtype==release, RZ_LOG_DEBUG and RZ_LOG_SILLY should be defined to nothing, thus they will be compiled out entirely from release builds.

None.

**Additional context**
None.

DMaroo:

**Detailed description**

`RZ_LOG_DEBUG` and `RZ_LOG_SILLY` are now only compiled when `RZ_BUILD_DEBUG` is set/defined.

**Test plan**

Green CI.

**Closing issues**

Closes #1612.

### DIFF
--- a/librz/type/parser/c_cpp_parser.c
+++ b/librz/type/parser/c_cpp_parser.c
@@ -230,6 +230,7 @@ RZ_API int rz_type_parse_file_stateless(RzTypeParser *parser, const char *path, 
 		return -1;
 	}
 	ut64 file_size = rz_file_size(path);
+	(void)file_size;
 	RZ_LOG_DEBUG("File size is %" PFMT64d " bytes, read %zu bytes\n", file_size, read_bytes);
 	int result = rz_type_parse_string_stateless(parser, source_code, error_msg);
 	free(source_code);
@@ -252,6 +253,7 @@ RZ_API int rz_type_parse_file(RzTypeDB *typedb, const char *path, const char *di
 		return -1;
 	}
 	ut64 file_size = rz_file_size(path);
+	(void)file_size;
 	RZ_LOG_DEBUG("File size is %" PFMT64d " bytes, read %zu bytes\n", file_size, read_bytes);
 	int result = rz_type_parse_string(typedb, source_code, error_msg);
 	free(source_code);
--- a/librz/type/parser/c_cpp_parser.c
+++ b/librz/type/parser/c_cpp_parser.c
@@ -229,9 +229,7 @@ RZ_API int rz_type_parse_file_stateless(RzTypeParser *parser, const char *path, 
 		free(source_code);
 		return -1;
 	}
-	ut64 file_size = rz_file_size(path);
-	(void)file_size;
-	RZ_LOG_DEBUG("File size is %" PFMT64d " bytes, read %zu bytes\n", file_size, read_bytes);
+	RZ_LOG_DEBUG("File size is %" PFMT64d " bytes, read %zu bytes\n", rz_file_size(path), read_bytes);
 	int result = rz_type_parse_string_stateless(parser, source_code, error_msg);
 	free(source_code);
 	return result;
@@ -252,9 +250,7 @@ RZ_API int rz_type_parse_file(RzTypeDB *typedb, const char *path, const char *di
 		free(source_code);
 		return -1;
 	}
-	ut64 file_size = rz_file_size(path);
-	(void)file_size;
-	RZ_LOG_DEBUG("File size is %" PFMT64d " bytes, read %zu bytes\n", file_size, read_bytes);
+	RZ_LOG_DEBUG("File size is %" PFMT64d " bytes, read %zu bytes\n", rz_file_size(path), read_bytes);
 	int result = rz_type_parse_string(typedb, source_code, error_msg);
 	free(source_code);
 	return result;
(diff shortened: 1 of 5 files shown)


DMaroo:
Do I need to add tests for this?

ret2libc (changes requested):
Instead of adding a new variable I'd suggest to use NDEBUG define.

You can do #ifdef NDEBUG and use add `default_options: ['b_ndebug=if-release']` in the `project` part of the main meson.build file.

ret2libc:
> Do I need to add tests for this?

It should not be needed. We already have several CI jobs that use both debug and release buildtypes.

DMaroo:
> Instead of adding a new variable I'd suggest to use NDEBUG define.
> 
> You can do #ifdef NDEBUG and use add `default_options: ['b_ndebug=if-release']` in the `project` part of the main meson.build file.

Since we are using our own logging functions, should we avoid using `NDEBUG`, since defining/un-defining it will affect other things as well, like assert macros? Just asking, I don't know which of the ways is preferred.

ret2libc:
> > Instead of adding a new variable I'd suggest to use NDEBUG define.
> > You can do #ifdef NDEBUG and use add `default_options: ['b_ndebug=if-release']` in the `project` part of the main meson.build file.
> 
> Since we are using our own logging functions, should we avoid using `NDEBUG`, since defining/un-defining it will affect other things as well, like assert macros? Just asking, I don't know which of the ways is preferred.

Yeah maybe you are right... We could also re-use the RZ_CHECKS_LEVEL define that we already have, which is 1 on release builds and 2 otherwise.

```
/*
 * RZ_CHECKS_LEVEL determines the behaviour of the rz_return_* set of functions.
 *
 * 0: completely disable every function and make them like no-operation
 * 1: silently enable checks. Check expressions and do return, but do not log anything
 * 2: enable checks and logging (DEFAULT)
 * 3: transform them into real assertion
 */
```
In that case, we should make RZ_LOG_DEBUG/SILLY no-op if RZ_CHECKS_LEVEL <= 1.

Or we could just use the additional define you did :D Not really so sure anymore...

@kazarmy @XVilka ?

DMaroo:
Also, since we are using `-Werror` in the Linux gcc build, therefore, warnings such as `unused variable` are leading to failure (https://github.com/rizinorg/rizin/pull/1616/checks?check_run_id=3533582589). This is introduced because `RZ_DEBUG_LOG` becomes a no-op, and therefore, some of the variables are used nowhere. I think we can just do some sort of `(void)fmtstr` and same for rest of the arguments to avoid the warning.

ret2libc:
> Also, since we are using `-Werror` in the Linux gcc build, therefore, warnings such as `unused variable` are leading to failure (https://github.com/rizinorg/rizin/pull/1616/checks?check_run_id=3533582589). This is introduced because `RZ_DEBUG_LOG` becomes a no-op, and therefore, some of the variables are used nowhere. I think we can just do some sort of `(void)fmtstr` and same for rest of the arguments to avoid the warning.

Yeah that should be easily fixed. Only few cases, it seems so far.

librz/type/parser/c_cpp_parser.c
```diff
@@ -230,6 +230,7 @@ RZ_API int rz_type_parse_file_stateless(RzTypeParser *parser, const char *path,
 		return -1;
 	}
 	ut64 file_size = rz_file_size(path);
+	(void)file_size;
 	RZ_LOG_DEBUG("File size is %" PFMT64d " bytes, read %zu bytes\n", file_size, read_bytes);
```
wargio:
```suggestion
	RZ_LOG_DEBUG("File size is %" PFMT64d " bytes, read %zu bytes\n", rz_file_size(path), read_bytes);
```

librz/type/parser/c_cpp_parser.c
```diff
@@ -252,6 +253,7 @@ RZ_API int rz_type_parse_file(RzTypeDB *typedb, const char *path, const char *di
...
```
wargio:
```suggestion
	RZ_LOG_DEBUG("File size is %" PFMT64d " bytes, read %zu bytes\n", rz_file_size(path), read_bytes);
```

wargio (commented):
How about this way?

meson.build
```diff
@@ -298,6 +298,10 @@ if checks_level == 9999
   endif
 endif
 
+if get_option('buildtype').startswith('debug')
+  add_project_arguments('-DRZ_BUILD_DEBUG', language : 'c')
+endif
```
ret2libc: Let's put this thing in rz_userconf.h.in and set it similar to the other variables there. That way people using the API can also see if rizin was compiled in debug mode or not.

librz/type/parser/c_cpp_parser.c
DMaroo: Makes sense to me. Thanks.

XVilka (approved):

XVilka:
I think it would be worth to cherry-pick it into 0.3.0 as well, to reduce the size of the binaries (slightly).

ret2libc (approved):

PR #1634, opened 2021-09-08
Fix function pointers and enums printing in complex types

ret2libc:


ret2libc:
@XVilka i think the test will fail. I was trying to create the "pretty_print" function and i noticed that the code for `tc` command handles array, pointers and callable in a different way, while the code of `rz_type_db_base_type_as_string` does not.... Are you sure we really need a separate function for pretty print? It seems to me that `rz_type_db_base_type_as_string` is just called in a few places.

XVilka:
Expected:
```
struct bla { int a; wchar_t (*func)(int a, const char *b); }
```
Got
```
struct bla { int a; wchar_t (*func)(int a, const char * b); func;  }
```

librz/type/function.c
```diff
@@ -428,7 +485,13 @@ RZ_API RZ_OWN char *rz_type_callable_ptr_as_string(const RzTypeDB *typedb, RZ_NO
...
 RZ_API RZ_OWN char *rz_type_callable_as_string(const RzTypeDB *typedb, RZ_NONNULL const RzCallable *callable) {
 	rz_return_val_if_fail(typedb && callable, NULL);
-	return callable_as_string(typedb, callable, false);
+	RzStrBuf *buf = rz_strbuf_new("");
+	if (!callable_as_string(buf, typedb, callable, 0)) {
+		rz_strbuf_free(buf);
+		return NULL;
+	}
+	char *result = rz_strbuf_drain(buf);
+	return result;
```
ret2libc:
```suggestion
	return rz_strbuf_drain(buf);
```

librz/type/function.c
```diff
@@ -413,11 +463,18 @@ RZ_API RZ_OWN char *rz_type_callable_ptr_as_string(const RzTypeDB *typedb, RZ_NO
...
+	if (!callable) {
+		return NULL;
+	}
+	RzStrBuf *buf = rz_strbuf_new("");
+	if (!callable_as_string(buf, typedb, callable, ptr_depth)) {
+		rz_strbuf_free(buf);
+		return NULL;
 	}
+	char *result = rz_strbuf_drain(buf);
+	return result;
```
ret2libc:
```suggestion
	return rz_strbuf_drain(buf);
```

librz/type/function.c
```diff
@@ -374,17 +388,55 @@ RZ_API bool rz_type_is_callable_ptr(RZ_NONNULL const RzType *type) {
 	return false;
 }
 
-static RZ_OWN char *callable_as_string(const RzTypeDB *typedb, RZ_NONNULL const RzCallable *callable, bool is_ptr) {
-	rz_return_val_if_fail(typedb && callable, NULL);
+/**
+ * \brief Checks if the RzType is the nested pointer to the RzCallable
```
ret2libc, line 389: what does it mean? can you make an example of what kind of "thing" would return true and what would return false?

librz/type/function.c
```diff
@@ -359,10 +359,24 @@ RZ_API bool rz_type_func_ret_set(RzTypeDB *typedb, const char *name, RZ_OWN RZ_N
...
  * \param typedb Types Database instance
- * \param type RzCallable
+ * \param type RzType
+ */
+RZ_API bool rz_type_is_callable(RZ_NONNULL const RzType *type) {
+	rz_return_val_if_fail(type, false);
+	if (type->kind == RZ_TYPE_KIND_CALLABLE) {
+		return true;
+	}
+	return false;
```
ret2libc:
```suggestion
	return type->kind == RZ_TYPE_KIND_CALLABLE;
```

XVilka (approved):

PR #1637, opened 2021-09-09
Implement pretty printing API for types [WIP]

### LINKED ISSUE #1624: Provide `rz_type_as_string_pretty` API for Cutter and others to pretty-print a type
Existing API like `rz_type_db_base_type_as_string` just print the type in one line, even though i'd like a nice representation similar to what's done in `tc <type>`.

Provide a new API that allows you to specify whether you want multiline and maybe other kind of pretty-printing stuff.

Just use `tc <type>`, but that's not great.

**Additional context**
I'm trying to convert the TypesWidget in Cutter to use types API and found this part is missing in terms of API.

DMaroo:
Started implementing pretty printing functions. Only implemented for structs as of now. Want to confirm whether this is the right way to go before proceeding to implement the complete function.

**Detailed description**

Allow for a pretty printing API for types to be used by Cutter.

**Test plan**

Tests not written yet.

**Closing issues**

Will close #1624 when complete.


librz/include/rz_type.h
```diff
@@ -243,7 +243,8 @@ RZ_API bool rz_type_db_delete_base_type(RzTypeDB *typedb, RZ_NONNULL RzBaseType
 RZ_API RZ_OWN RzList /* RzBaseType */ *rz_type_db_get_base_types_of_kind(const RzTypeDB *typedb, RzBaseTypeKind kind);
 RZ_API RZ_OWN RzList /* RzBaseType */ *rz_type_db_get_base_types(const RzTypeDB *typedb);
 
-RZ_API RZ_OWN char *rz_type_db_base_type_as_string(const RzTypeDB *typedb, RZ_NONNULL const RzBaseType *type);
+RZ_API RZ_OWN char *rz_type_db_base_type_as_string(RZ_NONNULL const RzTypeDB *typedb, RZ_NONNULL const RzBaseType *type);
+RZ_API RZ_OWN char *rz_type_db_base_type_as_string_pretty(RZ_NONNULL const RzTypeDB *typedb, RZ_NONNULL const RzBaseType *type, bool multiline, bool unfold);
```
ret2libc: I would go for a int that can contain some flags instead of several bool.

So you'd call it like `rz_type_db_base_type_as_string_pretty(....., RZ_TYPE_PRINT_MULTILINE | RZ_TYPE_PRINT_FOLD)` or something like this.

librz/type/base.c
```diff
@@ -237,3 +237,54 @@ RZ_API RZ_OWN char *rz_type_db_base_type_as_string(const RzTypeDB *typedb, RZ_NO
 	char *bufstr = rz_strbuf_drain(buf);
 	return bufstr;
 }
+
+char *rz_type_db_base_type_as_string_pretty_helper(RZ_NONNULL const RzTypeDB *typedb, RZ_NONNULL const RzBaseType *type, bool multiline, bool unfold, int indent_level) {
```
ret2libc: static

ret2libc (commented):
I think part of this effort should be to refactor some of the printing in the types module. Also, you should probably sync with @XVilka because he's already touching some of this code.

librz/include/rz_type.h
XVilka: Agree on this proposal, much cleaner.

DMaroo:
I feel the current code can (probably) be improved from the point of view of performance, readability and maintainability. So please suggest any sort of such changes, especially, I am looking for ways to simplify the control flow of the current function definition. Maybe I could reuse some other functions and make this simpler? I don't know for sure.

Also, it would be better that any extra functionality/customizability in terms of options is added now, instead of refactoring it later (I suppose), since it will be easier to do it now than later, I suppose. Please suggest if any.

librz/include/rz_type.h
```diff
@@ -243,7 +251,8 @@ RZ_API bool rz_type_db_delete_base_type(RzTypeDB *typedb, RZ_NONNULL RzBaseType
 RZ_API RZ_OWN RzList /* RzBaseType */ *rz_type_db_get_base_types_of_kind(const RzTypeDB *typedb, RzBaseTypeKind kind);
 RZ_API RZ_OWN RzList /* RzBaseType */ *rz_type_db_get_base_types(const RzTypeDB *typedb);
 
-RZ_API RZ_OWN char *rz_type_db_base_type_as_string(const RzTypeDB *typedb, RZ_NONNULL const RzBaseType *type);
+RZ_API RZ_OWN char *rz_type_db_base_type_as_string(RZ_NONNULL const RzTypeDB *typedb, RZ_NONNULL const RzBaseType *type);
+RZ_API RZ_OWN char *rz_type_db_base_type_as_string_pretty(RZ_NONNULL const RzTypeDB *typedb, RZ_NONNULL const RzBaseType *type, RzTypePrintOpts opts);
```
ret2libc:
```suggestion
RZ_API RZ_OWN char *rz_type_db_base_type_as_string_pretty(RZ_NONNULL const RzTypeDB *typedb, RZ_NONNULL const RzBaseType *type, int opts);
```
This should be an int, because you could pass several values of `RzTypePrintOpts` as a single option.

librz/include/rz_type.h
```diff
@@ -197,6 +197,14 @@ typedef enum {
 	RZ_TYPE_COND_LS ///< Unsigned lower or same     Less than or equal
 } RzTypeCond;
 
+/**
+ * \brief Pretty printing options for types
+ */
+typedef enum {
+	RZ_TYPE_PRINT_MULTILINE,
+	RZ_TYPE_PRINT_UNFOLD
```
ret2libc:
```suggestion
	RZ_TYPE_PRINT_MULTILINE = 1 << 0,
	RZ_TYPE_PRINT_UNFOLD = 1 << 1,
```
Otherwise they can't be easily mixed with `|`

ret2libc (commented):
I guess you know, but writing anyway: ideally rz_type_db_base_type_as_string_pretty should be used by `tc` command handler . Also, IF possible other `_as_string` functions should rely on the same code.

librz/include/rz_type.h
DMaroo: I already did this. Just haven't committed yet :)

XVilka:
See https://github.com/rizinorg/rizin/pull/1645 first, it unifies all `*_as_string()` functions.

DMaroo:
Thanks for that PR. Using the new API will make this code much more simpler and easier to read/maintain.

librz/include/rz_type.h
XVilka: @DMaroo no need for `RZ_NONNULL` for `typedb` parameter. It's obvious that almost any function in the `rz_type_*` namespace expects nonnull type database, except the `*_new()` one.

XVilka (changes requested):

ret2libc (commented):
BTW, make sure to unit test this ;) Actually, would be awesome to have them first and discuss the expected results there instead of going with the impl first and then realize things are not as expected.

ret2libc:
? Why closing?

DMaroo:
I am sorry, I messed up. I did a git reset to remove the old code (since it would have been easier to rewrite the functions using the new API than to modify/update existing code, and some merge conflicts), and then force pushed, and GitHub automatically closed the PR. No worries, I suppose. I'll open another PR.

PR #1641, opened 2021-09-09
Fixes for macOS 10.5/ppc64

thestr4ng3r:
**DO NOT SQUASH**

There are more issues, but this at least fixes the build with `-Ddebugger=false` on Leopard/ppc.

Tested on:
- [x] 10.5/ppc
- [x] 10.5/x86
- [x] 10.6/x86
- [x] 11.6/arm64

### DIFF
--- a/binrz/rizin/macos_sign.sh
+++ b/binrz/rizin/macos_sign.sh
@@ -13,4 +13,16 @@ if [ "$IS_INSTALL" = "true" ] ; then
     SRC="${MESON_INSTALL_DESTDIR_PREFIX}/${SRC}"
 fi
 
+# The --entitlements arg was only added in Mac OS X 10.6 Snow Leopard / Xcode 3.2 / security_systemkeychain-36515.
+# Moreover, our build system adds load mach-o load commands that the ancient codesign does not recognize either so let's skip it.
+# uname -r is
+#   9.<something> for 10.5
+#   10.<something> for 10.6
+#   ...
+OSVER=$(uname -r)
+if [ "${OSVER%%.*}" -lt 10 ]; then
+	echo "Detected Mac OS X < 10.6, skipping codesign."
+	exit 0
+fi
+
 codesign --entitlements "${ENTITLEMENT}" --force -s - "${SRC}"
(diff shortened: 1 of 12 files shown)


meson.build
```diff
@@ -349,7 +349,7 @@ if not cc.has_function('clock_gettime', prefix: '#include <time.h>') and cc.has_
 endif
 
 code = '#include <unistd.h>\nextern char **environ;\nint main() { char **env = environ; }'
-have_environ = cc.links(code, name: 'have extern char **environ')
+have_environ = false # cc.links(code, name: 'have extern char **environ')
```
XVilka: This doesn't look right

binrz/rizin/macos_sign.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
```
XVilka: Is this possible to do without bash? Especially, considering that later MacOS versions switched to the Zsh as the default one.

librz/util/sys.c
```diff
@@ -1246,7 +1246,9 @@ RZ_API char **rz_sys_get_environ(void) {
 
 RZ_API void rz_sys_set_environ(char **e) {
 	env = e;
-#if HAVE_ENVIRON
+#if __APPLE__ && !HAVE_ENVIRON
```
XVilka, line 1249: @ret2libc please take a look, maybe it's better to move to Meson checks as well?

XVilka (changes requested):

librz/util/sys.c
thestr4ng3r: Could be done I think, but `__NSGetEnviron()` is both macOS/iOS-specific and always available there so `__APPLE__` would be the right indicator without needing to check compilation.

binrz/rizin/macos_sign.sh
thestr4ng3r: The relevant part is
```
OSVER=$(uname -r)
if [ "${OSVER%%.*}" -lt 10 ]; then
```
I think this should still work with sh. At least shellcheck does not complain. So I will just have to revert the shebang.

XVilka (approved):
LGTM, but could you please send from the `dist-` branch instead? Just to be sure it doesn't break any of the existing CI builds.

PR #1645, opened 2021-09-10
Add tests for `rz_type*_as_string()` functions

XVilka:

**Detailed description**

Added unit tests specifically for 3 different "type as string" functions in various cases.

**Test plan**

CI is green.


test/unit/test_type.c
```diff
@@ -455,6 +455,63 @@ static bool test_const_types(void) {
 	mu_end;
 }
 
+static char *array = "int a[65][5][0]";
+static char *array_exp1 = "int [65][5][0]";
+static char *array_ptr = "int * const *a[][][][9]";
+static char *array_ptr_exp1 = "int * const *[0][0][0][9]";
+static char *array_ptr_exp2 = "int * const *a[0][0][0][9]";
```
ret2libc, line 462: Maybe stupid question from me, but is `a[][][][9]` really equal to `a[0][0][0][9]`?

test/unit/test_type.c
```diff
@@ -455,6 +455,63 @@ static bool test_const_types(void) {
...
+	ttype = rz_type_parse_string_single(typedb->parser, array_ptr, &error_msg);
+	mu_assert_notnull(ttype, "array type parse successfull");
+	mu_assert_true(ttype->kind == RZ_TYPE_KIND_ARRAY, "is array");
+
+	array_str1 = rz_type_as_string(typedb, ttype);
+	array_str2 = rz_type_declaration_as_string(typedb, ttype);
+	array_str3 = rz_type_identifier_declaration_as_string(typedb, ttype, "a");
+	mu_assert_streq_free(array_str1, array_ptr_exp1, "rz_type_as_string");
+	mu_assert_streq_free(array_str2, array_ptr_exp1, "rz_type_declaration_as_string");
+	mu_assert_streq_free(array_str3, array_ptr_exp2, "rz_type_identifier_declaration_as_string");
```
ret2libc, line 497: I think that the "pretty print" function developed by @DMaroo should be able to print, with different options, all these different types of output. What do you think?

However, I'm not sure I do see the difference between `rz_type_as_string` and `rz_type_identifier_declaration_as_string`. Could you explain a bit?

Also, the doc for those functions should be very explicit about what they output and maybe even provide some examples IMO.

XVilka: I explained this before already:
- `rz_type_as_string()` shows pure type without identifier without unfolding it: `struct bla *[2]`
- `rz_type_declaration_as_string()` shows pure type without identifier and unfolds it: `struct bla { int a; char foo; } *[2]`
- `rz_type_identifier_declaration_as_string()` shows type with identifier: `struct bla *boo[2]`. Moreover, it also expands the anonymous structures inline. They all are used in different parts of the code and serve different purpose. We need them all.

test/unit/test_type.c
XVilka: It's VLA, we need some value in the array size anyway, we set 0 when not specified in `[]` during the parsing. Then yes, they are equivalent.

test/unit/test_type.c
ret2libc: Ok yeah it's clearer with `struct_array_ptr` test. Please make sure those things are properly captured in the doxygen doc. And by looking at that example actually it seems that `rz_type_as_string` does not unfold the struct, while `rz_type_declaration_as_string` does. Right?

So, to wrap up this is what I understand:
- rz_type_as_string == nothing special
- rz_type_declaration_as_string == UNFOLD
- rz_type_identifier_declaration_as_string == UNFOLD + IDENTIFIER
- `tc` output == MULTILINE (no UNFOLD and no IDENTIFIER, right?)

Another question: do we need to differentiate between UNFOLD and UNFOLD_INNER (meaning that it unfolds also the inner structures, if present)? I'm trying to gather all the requirements to have a single pretty-print function for @DMaroo .

XVilka: Using `rz_type_identifier_declaration_as_string()` is very important since we need to put all pointer things *before* identifier and array things *after* identifier. Thus it's required to pass the identifier *inside* the function during the string formation, it's crucial - we cannot simple concatenate "typestr + identifier".

XVilka: @ret2libc @DMaroo we likely need to have 2 different unfolds as well:
- `UNFOLD_ALL` (unfold all structures/unions that are not typedefs)
- `UNFOLD_ANONYMOUS` (only anonymous structures/unions)
Apart from that right, yes

librz/type/type.c
```diff
@@ -1029,8 +1042,8 @@ RZ_API RZ_OWN char *rz_type_identifier_declaration_as_string(const RzTypeDB *typ
  * \brief Returns the type C identifier
  *
  * In case of the compound types it returns the name of identifier
- * For example, for "char **ptr" it will return "ptr",
- * for "const int **arr[56][76]" it will return "arr"
+ * For example, for "char **ptr" it will return "char",
+ * for "const int **arr[56][76]" it will return "int"
```
ret2libc, line 1046: It seems there's something wrong here.

ret2libc (approved):
Fix that doc comment, then you are good to go!

librz/type/type.c
XVilka: It isn't. It *was*.

XVilka: The identifier means the *type identifier* aka name, not the variable name.

librz/type/type.c
```diff
@@ -787,17 +787,21 @@ RZ_API ut64 rz_type_db_get_bitsize(const RzTypeDB *typedb, RZ_NONNULL RzType *ty
 	return 0;
 }
 
-static char *type_as_string(const RzTypeDB *typedb, RZ_NONNULL const RzType *type, bool prev_ptr, bool is_first) {
-	rz_return_val_if_fail(typedb && type, NULL);
+struct HelperBufs {
+	RzStrBuf *arraybuf;
+	RzStrBuf *ptrbuf;
+};
```
wargio, line 793: Please use typedef and place this on top of the file, so we always have in a .c or .h file

1. macros
2. types
2. definitions
3. functions

XVilka: There is no need to use typedef in this case, I used `struct` on purpose, to show the true essence of this parameter - it's just a pair of buffers. And it's located before 3 functions it's used in. It is unnecessary everywhere else and shouldn't be used outside those.

ret2libc: I agree no typedefs necessary when the thing is local to a file

test/unit/test_type.c
DMaroo: (For the pretty print function) How should cases of self-reference (linked list) or circular references be resolved while unfolding a struct? Should the unfolding be avoided if the type in question is same as the root type?

PR #1646, opened 2021-09-10
Fix #1642 - remove `toe` command

### LINKED ISSUE #1642: toe does not load existing type in editor and it does not change it afterward
| OS/arch/bits (mandatory)                             | Fedora 33
| `rizin -v` full output, **not truncated** (mandatory)  | rizin 0.3.0-git

### Expected behavior
Current c output of the given type is automatically added in the buffer and the new value is correctly parsed.

### Actual behavior
Vim is executed, but `tm` structure is not loaded in the vim buffer. Also, when you save, there's a parser error even if the data is right.

```
$ rizin 
> toe tm
```

Few things I noticed:
1) use rz_cmd_escape_arg when executing `tc` command in `rz_types_open_editor`
2) use `rz_core_editor(core, NULL, str)` instead of `*.h`. It seems that somehow that fixes the first problem, that is that data is not loaded in the vim buffer.
3) even though i tried the step 2), something failed while parsing the type back from vim. I'm not sure why.

XVilka:

**Detailed description**

Removed the `toe` command.

**Test plan**

CI is green

**Closing issues**

Closes https://github.com/rizinorg/rizin/issues/1642

### DIFF
--- a/librz/core/ctypes.c
+++ b/librz/core/ctypes.c
@@ -999,7 +999,7 @@ RZ_IPI bool rz_types_open_file(RzCore *core, const char *path) {
 		}
 	}
 	if (!strcmp(path, "-")) {
-		char *tmp = rz_core_editor(core, "*.h", "");
+		char *tmp = rz_core_editor(core, NULL, "");
 		if (tmp) {
 			char *error_msg = NULL;
 			int result = rz_type_parse_string_stateless(typedb->parser, tmp, &error_msg);
@@ -1024,24 +1024,4 @@ RZ_IPI bool rz_types_open_file(RzCore *core, const char *path) {
 	}
 	free(homefile);
 	return true;
-}
-
-RZ_IPI void rz_types_open_editor(RzCore *core, const char *typename) {
-	RzTypeDB *typedb = core->analysis->typedb;
-	char *str = rz_core_cmd_strf(core, "tc %s", typename ? typename : "");
-	char *tmp = rz_core_editor(core, "*.h", str);
-	if (tmp) {
-		char *error_msg = NULL;
-		int result = rz_type_parse_string_stateless(typedb->parser, tmp, &error_msg);
-		if (result) {
-			// TODO: remove previous types and save new edited types
-			//rz_type_db_purge(typedb);
-		}
-		if (error_msg) {
-			eprintf("%s\n", error_msg);
-			free(error_msg);
-		}
-		free(tmp);
-	}
-	free(str);
 }
--- a/librz/core/cmd_descs/cmd_descs.c
+++ b/librz/core/cmd_descs/cmd_descs.c
@@ -185,7 +185,6 @@ static const RzCmdDescArg type_link_del_args[2];
 static const RzCmdDescArg type_list_noreturn_args[2];
 static const RzCmdDescArg type_noreturn_del_args[2];
 static const RzCmdDescArg type_open_file_args[2];
-static const RzCmdDescArg type_open_editor_args[2];
 static const RzCmdDescArg type_open_sdb_args[2];
 static const RzCmdDescArg type_print_args[3];
 static const RzCmdDescArg type_print_value_args[3];
@@ -4023,21 +4022,6 @@ static const RzCmdDescHelp type_open_file_help = {
 	.args = type_open_file_args,
 };
 
-static const RzCmdDescArg type_open_editor_args[] = {
-	{
-		.name = "type",
-		.type = RZ_CMD_ARG_TYPE_STRING,
-		.flags = RZ_CMD_ARG_FLAG_LAST,
-		.optional = true,
-
-	},
-	{ 0 },
-};
-static const RzCmdDescHelp type_open_editor_help = {
-	.summary = "Open cfg.editor to edit types",
-	.args = type_open_editor_args,
-};
-
 static const RzCmdDescArg type_open_sdb_args[] = {
 	{
 		.name = "file",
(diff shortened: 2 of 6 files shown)


XVilka:
@ret2libc the most problematic case is when you try to change the multiple existing types - due to how C parser is implemented - it adds the types to the types hashtable on the fly during the parsing, there is no easy way to know what exact type changes occured. There are functions that use separate hashtables for that (e.g. `rz_type_parse_string_stateless()` allows to specify new parser state with custom hashtables), but we don't have `ht_pp_merge()` function yet.

ret2libc:
> @ret2libc the most problematic case is when you try to change the multiple existing types

What do you mean exactly? Does that work but it was hard to implement or it doesn't work?

librz/core/ctypes.c
```diff
@@ -1026,22 +1026,78 @@ RZ_IPI bool rz_types_open_file(RzCore *core, const char *path) {
...
+	}
+	// Get the all types
+	RzList *types = rz_type_db_get_base_types(typedb);
+	if (!types || rz_list_empty(types)) {
+		return false;
+	}
+	RzListIter *iter;
+	RzBaseType *t;
+	rz_list_foreach (types, iter, t) {
+		char *str = rz_type_db_base_type_as_string(typedb, t);
```
ret2libc: Is this really the right string to print? I'd expect more something like `tc` output, which is a bit better formatted.

librz/core/ctypes.c
```diff
@@ -1026,22 +1026,78 @@ RZ_IPI bool rz_types_open_file(RzCore *core, const char *path) {
...
-		}
-		if (error_msg) {
-			eprintf("%s\n", error_msg);
-			free(error_msg);
+	RzStrBuf *contents = rz_strbuf_new("");
+	if (!contents) {
+		return false;
+	}
+	// Get the all types
+	RzList *types = rz_type_db_get_base_types(typedb);
```
ret2libc: Does it make sense to list the atomic types like char/int/etc?

librz/core/ctypes.c
```diff
@@ -1026,22 +1026,78 @@ RZ_IPI bool rz_types_open_file(RzCore *core, const char *path) {
...
+RZ_IPI bool rz_types_open_editor_type(RzCore *core, RZ_NONNULL const char *name) {
+	rz_return_val_if_fail(name, false);
+
+	RzTypeDB *typedb = core->analysis->typedb;
+	RzBaseType *btype = rz_type_db_get_base_type(typedb, name);
+	if (!btype) {
+		RZ_LOG_ERROR("Cannot find type \"%s\"\n", name);
+		return false;
+	}
+	char *str = rz_type_db_base_type_as_string(typedb, btype);
```
ret2libc: same here... I'd expect things with multiline and better printed in the vim buffer. With this function we get everything in one line.

librz/core/ctypes.c
```diff
@@ -1026,22 +1026,78 @@ RZ_IPI bool rz_types_open_file(RzCore *core, const char *path) {
...
+	free(tmp);
+	free(str);
+	return success;
+}
+
+RZ_IPI bool rz_types_open_editor_type(RzCore *core, RZ_NONNULL const char *name) {
+	rz_return_val_if_fail(name, false);
+
+	RzTypeDB *typedb = core->analysis->typedb;
+	RzBaseType *btype = rz_type_db_get_base_type(typedb, name);
```
ret2libc: Probably it does not make sense to accept atomic type here either, right? You can't modify them anyway AFAIK.

librz/core/ctypes.c
XVilka: Do we need typedefs? Or only enums, structures and unions?

ret2libc: I think also typedefs make sense

ret2libc (changes requested):
We should really provide an API for editing an existing type. This is useful for Cutter as well, where a user wants to edit the type X. I think ideally we'd have something like:
```C
RZ_API int rz_type_edit_type(RzTypeDB *typedb, RzType/RzBaseType *type, const char *code, char **error_msg);
```

XVilka:
@ret2libc what if editing an old type *adds* a new one?
For example, you have structure like this:
```c
struct bla {
   uint64_t a;
   void *b;
};
```
Then you call this function and type instead this:
```c
typedef struct uint128_t {
   uint64_t high;
   uint64_t low;
};
struct bla {
    uint128_t  a;
    void *b;
};
```

ret2libc:
> @ret2libc what if editing an old type _adds_ a new one?

I think `rz_type_parse_string` (and the new `rz_type_edit`) returns the number of types that were parsed. If that number != 1, we fail with an error.

For not freeing the elem in the ht, I see we already do a hack in function.c. I suggest you do it here as well and then later we can add the new API in sdb with more calm.
```
		// Do not free the elements of the Ht, because they were moved to ctx.inst_vars_new
		fcn->inst_vars->opt.freefn = NULL;
		ht_up_free(fcn->inst_vars);
		fcn->inst_vars = ctx.inst_vars_new;
```

thestr4ng3r:
I think `rz_type_parse_string()` should be a 2-pass thing.
First you call `rz_type_parse_string()`, it will purely parse everything and return some datastructure with all the new stuff inside of it.
Then you can decide what to do with it:
* Fail if there are already existing types
* Apply them to the database, overwriting any existing types
* For editing a single type:
  * If there is a single type of the same name in the result, apply it
  * If there is a single type of another name in the result, delete the old type and apply the new (renaming is also supported then)
  * If there are multiple types in the result, fail hard.

XVilka:
@thestr4ng3r the internals of the parser allow to import the hashtable and put the parsed types into a different one. We will need things like `ht_pp_merge()` first, before implementing this two-stage parsing:
```c
/**
 * \brief Creates a new instance of the C type parser
 *
 * Creates the new instance of the C types parser preloaded
 * hashtables for RzBaseTypes and RzCallable types. It will
 * use provided hashtables for storing the parsed types as well.
 *
 * \param type RzBaseTypes hashtable to preload into the parser state
 * \param type RzCallable hashtable to preload into the parser state
 */
RZ_API RZ_OWN RzTypeParser *rz_type_parser_init(HtPP *types, HtPP *callables) 
```
```c
/**
 * \brief Parses the C type string reusing the existing parser state
 *
 * \param parser RzTypeParser instance
 * \param code The C type itself
 * \param error_msg A pointer where all error messages will be stored
 */
RZ_API int rz_type_parse_string_stateless(RzTypeParser *parser, const char *code, char **error_msg) {
	return type_parse_string(parser->state, code, error_msg);
}
```

Also, we will need to have two sets of the hashtables then - one is for preloading, read-only, and one is for the storage - write-only. Once parsing is complete, we can merge the storage inside the output hashtable.

The main problem arises if we have a type called "struct qwe" which is defined in the read-only hashtable but not defined in the storage. Do we remove "struct qwe" first? Or what?

thestr4ng3r:
Yes, that sounds like a good solution.
I am not sure though what you mean by the `struct qwe`, could you maybe give a more verbose example?

XVilka:
Superseded by https://github.com/rizinorg/rizin/issues/1642

Issue #1442, opened 2021-08-13
Convert analysis `a` (cmd_analysis.c) commands to rzshell


XVilka:
Some of the commands are already converted, we should finish the complete migration. Moreover, some analysis logic is inside the `librz/core/cmd_analysis.c`. We should move as much as possible inside either `librz/analysis/*` or `librz/core/canalysis.c` from `librz/core/cmd_analysis.c`.

The task implies also possible API changes and some refactorings, see https://github.com/rizinorg/rizin/issues/1223, https://github.com/rizinorg/rizin/issues/1117, https://github.com/rizinorg/rizin/issues/992, and https://github.com/rizinorg/rizin/issues/489

- [x] `ar` https://github.com/rizinorg/rizin/pull/2006
- [x] `af` https://github.com/rizinorg/rizin/pull/2040, https://github.com/rizinorg/rizin/pull/2586
- [x] `aa` https://github.com/rizinorg/rizin/pull/2594
- [x] `a8` https://github.com/rizinorg/rizin/pull/2396
- [x] `ab` https://github.com/rizinorg/rizin/pull/2395
- [x] `ac` https://github.com/rizinorg/rizin/pull/2160
- [x] `aC` https://github.com/rizinorg/rizin/pull/2793
- [x] `ad`
- [ ] `ae`  https://github.com/rizinorg/rizin/issues/1533
- [x] `aF`
- [x] `ag` https://github.com/rizinorg/rizin/pull/2916
- [x] `ah` https://github.com/rizinorg/rizin/pull/2130
- [x] `ai`
- [x] `aL`
- [x] `an` https://github.com/rizinorg/rizin/pull/2447
- [x] `ao` https://github.com/rizinorg/rizin/pull/2396
- [x] `aO` https://github.com/rizinorg/rizin/pull/2396
- [x] `ap` (removed)
- [x] `as` https://github.com/rizinorg/rizin/pull/2323
- [x] `av`
- [x] `ax`

See these files:
- `librz/core/cmd_analysis.c`
- `librz/core/cmd_descs/cmd_analysis.yaml`

The task is important for further porting of different architectures to the global variables, new IL.

p371jah:
`ax` commands have already been ported

PR #1655, opened 2021-09-10
Port ax commands to newshell

p371jah:

**Detailed description**

- Optional `address from` argument was removed for all commands (`axc addr_to addr_from` --> `axc addr_to @ addr_from`).
- `ax` was renamed to `axl`
- `axf.` was removed as undocmunted
- `axtg*` was removed as undocmunted
- `axv` wasn't ported as dup of `afxv`
- `axff` wasn't ported as dup of `afx`
- `rz_analysis_xrefs_list()` now just returns RzList of xrefs
- with `ax-`you can delete xrefs from `UT64_MAX`
- output of `axlj`, `axtj`, `axfj`, `afxj` was unified
- output of `axl*`, `axt*`, `axf*` was unified
- `ax.` was removed as useless 

Related issues: #1442
...

**Test plan**

...

**Closing issues**

...


librz/core/cmd_analysis.c
```diff
@@ -6041,6 +6035,19 @@ static void analysis_axg(RzCore *core, const char *input, int level, Sdb *db, in
...
+		PJ *pj = pj_new();
+		analysis_axg(core, addr, 0, ht, mode, pj);
+		rz_cons_printf("%s\n", pj_string(pj));
+		pj_free(pj);
+	} else {
+		analysis_axg(core, addr, 0, ht, mode, NULL);
+	}
+	ht_uu_free(ht);
+}
+
```
wargio: do not forget to run clang-format

ret2libc:
Why the different order of arguments for ax- and axc?

p371jah:
> 
> 
> Why the different order of arguments for ax- and axc?

For `ax-` command `addr_from` argument is optional: if it is undefined `ax-` removes all xrefs to `addr_to`. In case of `ax- to from` we can't use `@`. So I don't know what is the best way to make `ax-` work with `@` and keep `addr_from` optional.

XVilka:
What about moving the listing feature of `ax` to the `axl` and keep `ax` command instead of the `ax0` for adding the reference? I think it makes more sense this way.

ret2libc:
> > Why the different order of arguments for ax- and axc?
> 
> For `ax-` command `addr_from` argument is optional: if it is undefined `ax-` removes all xrefs to `addr_to`. In case of `ax- to from` we can't use `@`. So I don't know what is the best way to make `ax-` work with `@` and keep `addr_from` optional.

I see. I'll think a bit about this... I think it *might* be confusing having to specify `to` after `@` in some cases and `from` in other cases.

XVilka:
@ret2libc we can use range `@(0x1000,0xFFFF)` where `0x1000` is the "from" value and `0xFFFF` is the "to" value.
We could allow syntax like `@(,0xFFFF)` for example, and `@(0x1000,)`.

ret2libc:
> @ret2libc we can use range `@(0x1000,0xFFFF)` where `0x1000` is the "from" value and `0xFFFF` is the "to" value.
> We could allow syntax like `@(,0xFFFF)` for example, and `@(0x1000,)`.

Why are you talking about ranges? This is just a command that needs to operate on two addresses. Also, I don't think this command has anything special to deserve an adjustment at the grammar.

p371jah:
Well we can keep `ax- to from`

XVilka:
@ret2libc ignore me, I thought `from-to` as an interval, not the reference.

XVilka:
@ret2libc @pelijah what do you think about my suggestion to keep `ax` for adding and creating `axl` for listing? It will also be consistent with `af` and `afl`, for example.

ret2libc:
> Well we can keep `ax- to from`

I think it makes sense, yeah



> @ret2libc @pelijah what do you think about my suggestion to keep `ax` for adding and creating `axl` for listing? It will also be consistent with `af` and `afl`, for example.

I'm ok with that as well, if @pelijah wants to implement it.

p371jah:
Should I move out printing stuff from `rz_analysis_xrefs_list()` to a separate function?

XVilka:
@pelijah yes, please do. We try to follow the names:
- `*_list()` returns either `RzList` or `RzVector`
- `*_print()` actually prints stuff.

Moreover, we usually try to avoid printing stuff outside of the `librz/core/*`, to remove the `RzCons` dependency in the future.

librz/core/cmd_analysis.c
```diff
@@ -9721,3 +9157,401 @@ RZ_IPI RzCmdStatus rz_analysis_function_vars_sp_setref_handler(RzCore *core, int
...
+		rz_cons_printf("%s 0x%" PFMT64x " @ 0x%" PFMT64x "\n", xrefs_type2cmd(xref->type), xref->to, xref->from);
+	}
+}
+
+RZ_IPI RzCmdStatus rz_analysis_xrefs_list_handler(RzCore *core, int argc, const char **argv, RzCmdStateOutput *state) {
+	RzListIter *iter;
+	RzAnalysisXRef *xref;
+	if (state->mode == RZ_OUTPUT_MODE_JSON && !state->d.pj) {
+		return RZ_CMD_STATUS_ERROR;
+	}
```
ret2libc: you can probably remove this. Or convert it into a `rz_return_val_if_fail`, since it is expected this function is always called with the proper things initialized. In particular, if mode == JSON || LONG_JSON, d.pj should be initialized. Same thing for TABLE.

librz/core/cmd_analysis.c
```diff
@@ -9721,3 +9157,401 @@ RZ_IPI RzCmdStatus rz_analysis_function_vars_sp_setref_handler(RzCore *core, int
...
+	rz_list_free(list);
+	return status;
+}
+
+RZ_IPI RzCmdStatus rz_analysis_xrefs_to_list_handler(RzCore *core, int argc, const char **argv, RzCmdStateOutput *state) {
+	RzAnalysisXRef *xref;
+	RzListIter *iter;
+	if (state->mode == RZ_OUTPUT_MODE_JSON && !state->d.pj) {
+		return RZ_CMD_STATUS_ERROR;
+	}
```
ret2libc: same

librz/core/cmd_analysis.c
```diff
@@ -9721,3 +9157,401 @@ RZ_IPI RzCmdStatus rz_analysis_function_vars_sp_setref_handler(RzCore *core, int
...
+	rz_list_free(list);
+	return status;
+}
+
+RZ_IPI RzCmdStatus rz_analysis_xrefs_from_list_handler(RzCore *core, int argc, const char **argv, RzCmdStateOutput *state) {
+	RzAnalysisXRef *xref;
+	RzListIter *iter;
+	if (state->mode == RZ_OUTPUT_MODE_JSON && !state->d.pj) {
+		return RZ_CMD_STATUS_ERROR;
+	}
```
ret2libc: same for this and others, if there are.

ret2libc (commented):
LGTM

ret2libc (approved):

p371jah:
Should we keep `axff` command?

ret2libc:
> Should we keep `axff` command?

I'd say kill it. We can always re-introduce, but I think it's better to keep less commands at this point.

p371jah:
I have removed `axff`

ret2libc:
LGTM. There are issues with pylint, which I think are discovered only because you touched one py file and the linter got updated and probably it discovered new stuff...

wargio (approved):

XVilka (approved):

PR #1656, opened 2021-09-11
Add demangling as plugins and libdemangle as external dependency

### LINKED ISSUE #1600: Extract demangling code into the separate library
- [ ] Move the code from `librz/bin/mangling/*` to the separate library `rz-demangle`
- [ ] Make it a Meson subproject in Rizin
- [ ] Move issues and pure demangling tests to this library instead
- [ ] Move `rz-libswift` code either completely or as a subproject of this library as well
- [ ] Move [issues with label "magling"](https://github.com/rizinorg/rizin/labels/mangling) to this new library repository

wargio:

**Detailed description**

Now the demanglers are all handled as external libs (https://github.com/rizinorg/rz-libdemangle).

Also https://github.com/rizinorg/rz-libswift is now a plugin

Closes https://github.com/rizinorg/rizin/issues/1600


librz/bin/bin.c
```diff
@@ -1293,3 +1293,149 @@ RZ_API const RzBinLdrPlugin *rz_bin_ldrplugin_get(RzBin *bin, const char *name)
 	}
 	return NULL;
 }
+
+static char *bin_demangle_cxx(RzBinFile *bf, const char *symbol, ut64 vaddr) {
```
XVilka, line 1310: Please add `rz_return_val_if_fail(symbol)`

XVilka: Also why `bin_demangle_cxx()` is static while `bin_demangle_rust()` is `RZ_API`?

librz/bin/bin.c
```diff
@@ -1293,3 +1293,149 @@ RZ_API const RzBinLdrPlugin *rz_bin_ldrplugin_get(RzBin *bin, const char *name)
...
+RZ_API char *bin_demangle_rust(RzBinFile *binfile, const char *symbol, ut64 vaddr) {
+	char *str = NULL;
+	if (!(str = bin_demangle_cxx(binfile, symbol, vaddr))) {
+		return str;
+	}
+	free(str);
+	return rz_demangler_rust(symbol);
+}
+
+RZ_API char *rz_bin_demangle(RzBinFile *bf, const char *language, const char *symbol, ut64 vaddr, bool libs) {
```
XVilka: Please add doxygen, also check `language` for NULL

librz/bin/bin.c
```diff
@@ -1293,3 +1293,149 @@ RZ_API const RzBinLdrPlugin *rz_bin_ldrplugin_get(RzBin *bin, const char *name)
...
+		language = rz_bin_language_to_string(type);
+	}
+	if (!language) {
+		return NULL;
+	}
+	char *demangled = NULL;
+	switch (type) {
+	case RZ_BIN_LANGUAGE_UNKNOWN: return NULL;
+	case RZ_BIN_LANGUAGE_KOTLIN:
+		/* fall-thru */
```
XVilka, line 1446: It's `fall-through`

XVilka (changes requested):
Looks much better and cleaner this way 👍 🥇

librz/bin/bin.c
wargio: mistake. it should be static.

librz/bin/bin.c
wargio: there is a check on language below. it is nullable at the end.

librz/bin/bin.c
wargio: it is not a mistake https://developers.redhat.com/blog/2017/03/10/wimplicit-fallthrough-in-gcc-7

librz/bin/bin.c
wargio: > Please add `rz_return_val_if_fail(symbol)`

This can't happen because the symbol value is checked by `rz_bin_demangle`

librz/bin/bin.c
XVilka: It's a bad language. What about `/* Falls through. */`?

XVilka (approved):

librz/bin/bin.c
wargio: for me is ok. i can make the change in another PR because there are multiple `/* fall-thru */` in the codebase

wargio: https://github.com/rizinorg/rizin/issues/1674

XVilka:
@wargio please rebase again to fix the Windows builds.

librz/bin/bin.c
```diff
@@ -1295,3 +1295,159 @@ RZ_API const RzBinLdrPlugin *rz_bin_ldrplugin_get(RzBin *bin, const char *name)
...
+		return out;
+	}
+	char *sign = (char *)strchr(out, '(');
+	if (!sign) {
+		return out;
+	}
+
+	char *str = out;
+	char *ptr = NULL;
+	char *method_name = NULL;
```
ret2libc, line 1322: is it reasonable to have this method extraction thing moved into libdemangler? I expect it's a common operation across languages to extract the method name of a demangled string, right?

librz/bin/bin_language.c
```diff
@@ -0,0 +1,244 @@
...
+/**
+ * \brief Tries to detect which language is used in the binary based on symbols and libraries
+ *
+ * Currently this method can detect the language only from bins that are either ELF, PE,
+ * Mach-O, Java Class and Dex.
+ * 
+ * The current supported languages are: c, cxx, dart, dlang, go, groovy, java, kotlin, msvc,
+ * objc, rust, swift.
+ */
+RZ_API RzBinLanguage rz_bin_language_detect(RzBinFile *binfile) {
```
ret2libc, line 67: do we have at least one test for each case? c++, rust, dart, etc. ?

librz/core/cmd_descs/cmd_info.yaml
```diff
@@ -137,21 +137,34 @@ commands:
...
+          - name: lang
+            type: RZ_CMD_ARG_TYPE_CHOICES
+            choices:
+              - c++
+              - java
+              - objc
+              - swift
+              - dlang
+              - msvc
+              - rust
```
ret2libc, line 155: Maybe later it's worth to introduce a new ARG_TYPE which is autocompleted at runtime instead of having this manual list

librz/core/cmd_info.c
```diff
@@ -594,8 +570,51 @@ RZ_IPI RzCmdStatus rz_cmd_info_pdb_download_handler(RzCore *core, int argc, cons
 	return RZ_CMD_STATUS_OK;
 }
 
+static bool print_demangler_info(const RzDemanglerPlugin *plugin, void *user) {
+	if (!user) {
+		rz_cons_printf("%-6s %-8s %s\n", plugin->language, plugin->license, plugin->author);
+		return true;
+	}
```
ret2libc, line 563: split this in a separate function.

meson_options.txt
```diff
@@ -33,6 +33,8 @@ option('use_sys_openssl', type: 'feature', value: 'disabled')
 option('use_sys_tree_sitter', type: 'feature', value: 'disabled')
 option('use_sys_libuv', type: 'feature', value: 'auto', description: 'Whether to force, suggest or not use at all the system version of libuv. If system version is not found, one is built statically, unless use_libuv is false.')
 option('use_libuv', type: 'boolean', value: true, description: 'If true, libuv is used to handle remote features')
+option('use_swift_demangler', type: 'boolean', value: true, description: 'If false, disables the swift demangler')
+option('use_gpl', type: 'boolean', value: true, description: 'Set to false when you want to disable gpl code')
```
ret2libc, line 37: I'd add "not completely supported yet" because I think there's other code that is GPL which is not disabled if you set this to false, right?

librz/core/cmd_descs/cmd_info.yaml
wargio: that is a great idea

librz/bin/bin.c
wargio: this is more a hack of the old code that adds classes and methods to the bin based on the demangled output. i would keep this here and then fix this by having this logic within bin by looking for this stuff.

librz/bin/bin_language.c
wargio: in tests, yes with the exclusion of dlang and dart.

meson_options.txt
wargio: if you mean in rizin codebase, yes, but if it is referred to libdemangler, no. it is supported and not built when this flag is enabled.

ret2libc: i mean in rizin codebase. This is rizin meson file, so one may think that it applies to the whole rizin.

wargio: then no. but i could find all GPL code and add the `#if USE_GPL` in another commit.

ret2libc: i think it's a separate thing. Maybe let's create an issue and do it separately after this if you want.

librz/bin/bin.c
ret2libc: ok yeah if it's all moved/refactored code that's fine.

meson_options.txt
wargio: ok, i have added the comment. i will make another PR with all GPL guards once this is merged.

XVilka: There are already issues for that:
- https://github.com/rizinorg/rizin/issues/70
- https://github.com/rizinorg/rizin/issues/209

XVilka (approved):

librz/main/rizin.c
```diff
@@ -389,6 +391,7 @@ RZ_API int rz_main_rizin(int argc, const char **argv) {
 	rz_signal_sigmask(SIG_BLOCK, &sigBlockMask, NULL);
 #endif
 
+	rz_demangler_plugin_init();
```
ret2libc: Why is this needed? I think it should not be here... maybe somewhere inside RzCore? Please look at how all the other plugins are initialized and do the same, this feels wrong.

ret2libc (changes requested):
I think most things look good, apart from the following:
Most `librz/main` binaries should not have to care about rz_demanglaer_plugin_init/fini IMHO. Maybe only rz-bin needs it.

- [ ] TODO: create issue to create new RZ_ARG_TYPE_LANG with autocompletion for `iD`

librz/main/rizin.c
wargio: no, i disagree.
it should be initialized by the tool only once and i want to avoid to implement ref counting for this.
i do not want to place this into RzCore because rz-bin needs it since it provides demangling and you can create n-th instances of the RzCore.
Also it is required to have dynamically allocated lists where plugins of this type are registered when loading shared libs.

ret2libc: I'm not convinced. Look at how other plugins do their initialization. Look at config.h.in and maybe let's add a `RZ_DEMANGLER_STATIC_PLUGINS` there with the list of plugins you need to build. You can easily define them in meson.build instead of at runtime.

wargio: it does not change the fact that they cannot be initialized by RzCore

ret2libc: You do not need to initialize/finalize anything if you just move the list in RZ_DEMANGLER_STATIC_PLUGINS, like what's done for all the other plugins types. It is no different than a RzBin plugin, needed by rz-bin.

wargio: no. all the plugins are still initialized when you call the `rz_<something>_new()`

librz/demangler/demangler.c
```diff
@@ -0,0 +1,204 @@
+// SPDX-FileCopyrightText: 2021 RizinOrg <info@rizin.re>
+// SPDX-FileCopyrightText: 2021 deroad <wargio@libero.it>
+// SPDX-License-Identifier: LGPL-3.0-only
+
+#include <rz_demangler.h>
+#include <rz_util.h>
+#include <rz_list.h>
+#include <rz_libdemangle.h>
+
+static RzList *demangler_plugins = NULL;
```
ret2libc: Just to wrap up what was discussed on mattermost:
1) no global
2) use RzDemangler object to keep track of demangler_plugins (and whatever can come later)
3) use `RZ_DEMANGLER_STATIC_PLUGINS` in config.h.in and initialize those plugins in `rz_demangler_new`

ret2libc (changes requested):

XVilka (approved):

wargio:
requires a version bump once https://github.com/rizinorg/rz-libdemangle/pull/10 is merged
**edit**: merged.

librz/demangler/demangler.c
```diff
@@ -0,0 +1,184 @@
+// SPDX-FileCopyrightText: 2021 RizinOrg <info@rizin.re>
+// SPDX-FileCopyrightText: 2021 deroad <wargio@libero.it>
+// SPDX-License-Identifier: LGPL-3.0-only
+
+#include <rz_demangler.h>
+#include <rz_util.h>
+#include <rz_libdemangle.h>
+
+#include "demangler_plugins.c"
```
ret2libc: let's avoid this inclusion of c files in new files if not strictly necessary. Just put the whole content here, I think it's fine.

librz/include/rz_demangler.h
```diff
@@ -0,0 +1,48 @@
...
+
+#define rz_demangler_plugin_demangle(x, y) ((x) && RZ_STR_ISNOTEMPTY(y) ? (x)->demangle(y) : NULL)
+
+RZ_API RZ_OWN char *rz_demangler_java(RZ_NULLABLE const char *symbol);
+RZ_API RZ_OWN char *rz_demangler_cxx(RZ_NONNULL const char *symbol);
+RZ_API RZ_OWN char *rz_demangler_objc(RZ_NONNULL const char *symbol);
+RZ_API RZ_OWN char *rz_demangler_rust(RZ_NONNULL const char *symbol);
+RZ_API RZ_OWN char *rz_demangler_msvc(RZ_NONNULL const char *symbol);
+
+RZ_API RZ_OWN RzDemangler *rz_demangler_new();
```
ret2libc:
```suggestion
RZ_API RZ_OWN RzDemangler *rz_demangler_new(void);
```
?? i don't remember if it's the same or not..

librz/main/rz-gg.c
```diff
@@ -6,6 +6,7 @@
 #include <rz_main.h>
 #include <rz_util/rz_print.h>
 #include <rz_util.h>
+#include <rz_demangler.h>
```
ret2libc: not necessary?

librz/main/rz-find.c
```diff
@@ -7,6 +7,7 @@
 #include <rz_main.h>
 #include <rz_types.h>
 #include <rz_search.h>
+#include <rz_demangler.h>
```
ret2libc: not necessary?

librz/main/rz-diff.c
```diff
@@ -6,6 +6,7 @@
 #include <rz_bin.h>
 #include <rz_diff.h>
 #include <rz_util.h>
+#include <rz_demangler.h>
```
ret2libc: not necessary?

librz/main/rz-sign.c
```diff
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: LGPL-3.0-only
 #include <rz_main.h>
 #include <rz_core.h>
+#include <rz_demangler.h>
```
ret2libc: not necessary?

ret2libc (commented):
LGTM

librz/include/rz_demangler.h
wargio: yes, it is the same. i can add that.

ret2libc (approved):

PR #1659, opened 2021-09-12
Fix `toe` command

### LINKED ISSUE #1642: toe does not load existing type in editor and it does not change it afterward

XVilka:

**Detailed description**

Changed the `toe` implementation to accept only one type name per call, fixed bugs.

**Test plan**

Try `toe tm` for example

**Closing issues**

Closes https://github.com/rizinorg/rizin/issues/1642

### DIFF
--- a/librz/core/ctypes.c
+++ b/librz/core/ctypes.c
@@ -123,33 +123,40 @@ RZ_IPI void rz_core_types_enum_print_all(RzCore *core, RzOutputMode mode) {
 	}
 }
 
-RZ_IPI void rz_core_types_enum_print_c(RzTypeDB *typedb, const RzBaseType *btype, bool multiline) {
-	rz_return_if_fail(btype);
-	rz_return_if_fail(btype->kind == RZ_BASE_TYPE_KIND_ENUM);
+RZ_IPI RZ_OWN char *rz_core_types_enum_as_c(RzTypeDB *typedb, const RzBaseType *btype, bool multiline) {
+	rz_return_val_if_fail(btype, NULL);
+	rz_return_val_if_fail(btype->kind == RZ_BASE_TYPE_KIND_ENUM, NULL);
 
 	char *separator;
+	RzStrBuf *buf = rz_strbuf_new("");
 	if (!rz_vector_empty(&btype->enum_data.cases)) {
-		rz_cons_printf("enum %s {%s", btype->name, multiline ? "\n" : "");
+		rz_strbuf_appendf(buf, "enum %s {%s", btype->name, multiline ? "\n" : "");
 		separator = multiline ? "\t" : "";
 		RzTypeEnumCase *cas;
 		rz_vector_foreach(&btype->enum_data.cases, cas) {
-			rz_cons_printf("%s%s = %" PFMT64u, separator, cas->name, cas->val);
+			rz_strbuf_appendf(buf, "%s%s = %" PFMT64u, separator, cas->name, cas->val);
 			separator = multiline ? ",\n\t" : ", ";
 		}
-		rz_cons_println(multiline ? "\n};" : "};");
+		rz_strbuf_append(buf, multiline ? "\n};\n" : "};\n");
 	} else {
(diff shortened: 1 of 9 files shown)


librz/core/ctypes.c
```diff
@@ -1004,44 +1066,68 @@ RZ_IPI bool rz_types_open_file(RzCore *core, const char *path) {
...
+	if (t->kind == RZ_BASE_TYPE_KIND_ATOMIC) {
+		RZ_LOG_ERROR("Cannot edit atomic type \"%s\"\n", typename);
+		return false;
+	}
+	char *str = rz_core_base_type_as_c(core, t, true);
+	if (!str) {
+		RZ_LOG_ERROR("Cannot genereate C representation of type \"%s\"\n", typename);
+		return false;
+	}
+	char *tmp = rz_core_editor(core, NULL, str);
```
thestr4ng3r, line 1104: Please add the code below as an API, so we can use it in Cutter.

wargio (approved):

ret2libc (changes requested):
Doesn't this suffer from the problem we discussed the other day? That is that if the new type is invalid, you will end up deleting the existing type?

Few solutions:
1) save the original C string representation and re-parse it in case the new one fails
2) remove the original basetype from the hashtable, but don't free it. If the new one fails you re-insert the basetype in the hashtable

XVilka:
@ret2libc it already does that: https://github.com/rizinorg/rizin/pull/1659/files#diff-9ad7c99ed0a1b70e3fd759554220a7ff325910afc935d78fb8c342387de91e81R1119

librz/core/ctypes.c
```diff
@@ -1004,44 +1066,99 @@ RZ_IPI bool rz_types_open_file(RzCore *core, const char *path) {
...
+	char *error_msg = NULL;
+	int result = rz_type_parse_string_stateless(typedb->parser, new, &error_msg);
+	if (result) {
 		if (error_msg) {
-			eprintf("%s\n", error_msg);
+			RZ_LOG_ERROR("%s\n", error_msg);
+		}
+		free(error_msg);
+		// There is an error during the parsing thus we restore the old type
+		int result = rz_type_parse_string_stateless(typedb->parser, old, &error_msg);
```
thestr4ng3r: What about not using `rz_type_db_delete_base_type()` to delete the base type entirely, but just removing it from the hashtable without freeing to re-insert it later?
Using the `old` string means that it must be guaranteed that printing a type to C and then parsing it again will always result in exactly the original type, which I would not trust if not strictly necessary.

XVilka:
@thestr4ng3r @ret2libc added new API function:
```c
/**
 * \brief Edits the existing base type given the new C code
 *
 * Searches the base type in the types database given the \p name.
 * If it exists - parses the \p typestr as the new C type. If there is
 * any error during the parsing it restores the original type in the
 * database.
 *
 * \param typedb Type Database instance
 * \param name Name of the base type
 * \param typestr C string of the new definition of the type
 */
RZ_API bool rz_type_db_edit_base_type(RzTypeDB *typedb, RZ_NONNULL const char *name, RZ_NONNULL const char *typestr)
```

thestr4ng3r (approved):
Perfect!

PR #1678, opened 2021-09-14
Type pretty printing API

### LINKED ISSUE #1624: Provide `rz_type_as_string_pretty` API for Cutter and others to pretty-print a type

DMaroo:

**Detailed description**

This provides a generalized function to pretty print types. The pretty printing style can be modified using `RzTypePrintOpts`. As of the current implementation, this is not easily scalable/modular in terms of options, since new options will require changing the logic of the existing function. But the current set of options provide a decent set of combinations, and we would probably not need to add anything more.

**Test plan**

The unit tests are diverse enough to see the functionality of the new pretty printing function.

Also, as mentioned above, the current implementation is somewhat complicated (in my opinion). It would be nicer if this could be made more modular/simpler, so that maintaining it becomes easier.

**Relevant PRs and issues**

#1637, #1645, #1624, rizinorg/cutter#2744

**Closing issues**

Closes #1624.


DMaroo:
I will be adding the tests. That will make it easy to review and find mistakes, if any.

librz/util/strbuf.c
```diff
@@ -307,6 +307,46 @@ RZ_API bool rz_strbuf_vappendf(RzStrBuf *sb, const char *fmt, va_list ap) {
...
 }
 
+/**
+ * \brief Strips the last n characters from the buffer
+ * 
+ * \param buf Buffer to strip
+ * \param n Number of bytes to strip
+ * \return bool true if successfull, false otherwise (buffer won't be modified in this case)
+ */
+RZ_API bool rz_strbuf_strip(RzStrBuf *buf, int n) {
```
ret2libc: Add few unit tests for this.

XVilka:
@DMaroo could you please extract `strbuf` changes and tests in a separate PR?

DMaroo:
> I will be adding the tests. That will make it easy to review and find mistakes, if any.

I have added 3 tests. Please let me know if this is the way I should go forward with. Also, exhaustively testing the pretty printing functions will require many varied combinations, so it will also be good if any of the reviewers could suggest the type and extent of test cases to cover.

librz/type/type.c
```diff
@@ -1055,6 +1055,260 @@ RZ_API RZ_OWN char *rz_type_identifier_declaration_as_string(const RzTypeDB *typ
...
+ * \param unfold_level level of unfolding to do in case of nested structures/unions (any negative number means maximum unfolding, i.e. INT32_MAX. 0 means no unfolding, just the typename and identifier, if any)
+ * \return char* string in pretty printed form, ending with a ";"
+ */
+RZ_API RZ_OWN char *rz_type_as_pretty_string(const RzTypeDB *typedb, RZ_NONNULL const RzType *type, RZ_NULLABLE const char *identifier, unsigned int opts, int unfold_level) {
+	rz_return_val_if_fail(typedb && type, NULL);
+
+	if (unfold_level < 0) { // any negative number means maximum unfolding
+		unfold_level = INT32_MAX;
+	}
+	HtPP *used_types = ht_pp_new0();
```
XVilka: Check if NULL

librz/type/type.c
```diff
@@ -1055,6 +1055,260 @@ RZ_API RZ_OWN char *rz_type_identifier_declaration_as_string(const RzTypeDB *typ
...
+		ht_pp_insert(used_types, self_ref_typename, NULL); // add the type to the ht
+	}
+	RzBaseType *btype = NULL;
+	bool is_anon = false;
+	if (type->kind == RZ_TYPE_KIND_IDENTIFIER) {
+		is_anon = !strncmp(type->identifier.name, "anonymous ", 10);
+		btype = rz_type_db_get_base_type(typedb, type->identifier.name);
+	}
+	if ((type->kind == RZ_TYPE_KIND_POINTER && rz_type_is_callable_ptr_nested(type)) || type->kind == RZ_TYPE_KIND_CALLABLE) {
+		identifier = NULL; // no need to separately print identifier for funtion pointers or functions
```
XVilka: `funtion` typo.

XVilka (changes requested):
Looks good! 👍

DMaroo:
Need confirmations on the *style* of pretty printing:

 1. Do we need the pretty printing function to print the semicolon (`;`) at the end or not?
 2. For structs/unions/enums being printed on a single line, I am adding two spaces before and after the braces, i.e.
 ```c
struct person {  char *firstname; char *lastname;  };
 ```
 Does this look fine, or should there be only one space?

DMaroo:
Also, shall I open another PR to reuse this new function for the rest of the type printing API (once this gets merged)? That will reduce redundant code and allow for a consistent *style* across the pretty printing API.

XVilka:
1. I think, at the top-level there should be semicolon. Note, that in case of the identifier used like `struct { int a; char *b; } foo;` semicolon should be put only *after* identifier.
2. Better to keep it just one space

DMaroo:
> 1. I think, at the top-level there should be semicolon. Note, that in case of the identifier used like `struct { int a; char *b; } foo;` semicolon should be put only _after_ identifier.

So is there anything to be modified with respect to the current API, because the example which you have given is what the current function seems to do?

test/unit/test_type.c
```diff
@@ -512,6 +512,185 @@ static bool test_type_as_string(void) {
...
+	mu_assert_notnull(ttype, "struct array ptr func ptr type parse unsuccessfull");
+	mu_assert_null(error_msg, "parsing errors");
+
+	error_msg = NULL;
+	ttype = rz_type_parse_string_single(typedb->parser, pretty_complex_const_pointer, &error_msg);
+	mu_assert_notnull(ttype, "struct array ptr func ptr type parse unsuccessfull");
+	mu_assert_null(error_msg, "parsing errors");
+
+	mu_end;
+}
```
ret2libc: is this here on purpose?

test/unit/test_type.c
```diff
@@ -512,6 +512,185 @@ static bool test_type_as_string(void) {
...
+	char *pretty_str = rz_type_as_pretty_string(typedb, ttype, "c", RZ_TYPE_PRINT_NO_OPTS, 1);
+	mu_assert_streq(pretty_str, pretty_complex_const_pointer, "complex const pointer type is ugly");
+	free(pretty_str);
+	free(ttype);
+
+	error_msg = NULL;
+	ttype = rz_type_parse_string_single(typedb->parser, pretty_struct_array_ptr_func_ptr, &error_msg);
+	mu_assert_notnull(ttype, "struct array ptr func ptr type parse unsuccessfull");
+	mu_assert_null(error_msg, "parsing errors");
+	pretty_str = rz_type_as_pretty_string(typedb, ttype, NULL, RZ_TYPE_PRINT_NO_OPTS, 2);
```
ret2libc: Why 2?

test/unit/test_type.c
```diff
@@ -512,6 +512,185 @@ static bool test_type_as_string(void) {
...
+						    "} maxmult;";
+static char *pretty_enum = "enum MCU { IRON = 0x1001, HAWK = 0x337, DOCS = 0x1337, CAPM = 0x2077 };";
+static char *pretty_enum_multiline = "enum MCU {\n"
+				     "\tIRON = 0x1001,\n"
+				     "\tHAWK = 0x337,\n"
+				     "\tDOCS = 0x1337,\n"
+				     "\tCAPM = 0x2077\n"
+				     "} enumult;";
+
+static bool test_type_as_pretty_string(void) {
```
ret2libc, line 586: I love all these test <3 Thanks!

ret2libc (commented):
Good work! I think what is missing now is that you use the new function rz_type_as_pretty_string to implement rz_type_as_string, rz_type_declaration_as_string, rz_type_identifier_declaration_as_string and also you use it for `tc` commands.

Right @XVilka  ? I think that would be the proof that everything works well ;)

test/unit/test_type.c
DMaroo: Oh, sorry. It was just a temporary test :)

test/unit/test_type.c
DMaroo: Using 1 would also work, but I just wanted to test that the function doesn't go crazy when given a larger number than expected.

test/unit/test_type.c
DMaroo: > I think what is missing now is that you use the new function rz_type_as_pretty_string to implement rz_type_as_string, rz_type_declaration_as_string, rz_type_identifier_declaration_as_string and also you use it for tc commands.

Agreed! But should that be done in this PR, or should I be creating another PR for that?

test/unit/test_type.c
ret2libc: Ok, makes sense. Maybe add a comment on top, just to make it clear. Or put it in the message string of the mu_assert after this line.

test/unit/test_type.c
ret2libc: I think this PR should be fine because if all works well it should be just a matter of removing a lot of the existing code and replace it with a call to `rz_type_as_pretty_string` with the right parameters. If it does not work that way, it means you probably have to adjust the API anyway, i think.

DMaroo: I have refactored the other type print functions to use this API. I had to add a few more pretty printing options (though it was very easy to add them) so that the expected output of the original functions stays the same. I will do the same for `tc` once you review the refactored functions. Thank you.

DMaroo:
What should be done about the integration tests? Should I change the API, or should I change the tests?

XVilka:
```

[XX] db/cmd/cmd_avg avgt # retype global variable
RZ_NOPLUGINS=1 rizin -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -Qc 'avga foo 0x100 int
avg
avgt foo char
avg
avgt foo "struct a { int a; char b; };"
avg
avgt foo "struct { int a; char b; void (*c)(int a, char b); };"
avg
avgt foo "union { int a; char b; };"
avg
avgt foo "enum { A=0,B,C,D };"
avg
avgt foo "enum foo { A=0,B,C,D };"
avg
avgt foo "void (*a)(int a, char b);"
avg
avgt foo "char *"
avg
avgt foo "char [50]"
avg
' =
-- stdout
--- expected
+++ actual
@@ -1,10 +1,10 @@
 global int foo @ 0x100
 global char foo @ 0x100
 global struct a foo @ 0x100
-global struct anonymous struct 0 foo @ 0x100
-global union anonymous union 0 foo @ 0x100
-global anonymous enum 0 foo @ 0x100
-global foo foo @ 0x100
+global struct foo @ 0x100
+global union foo @ 0x100
+global enum foo @ 0x100
+global enum foo foo @ 0x100
 global void foo @ 0x100
 global char * foo @ 0x100
 global char [50] foo @ 0x100
```
You lost `anonymous`

DMaroo:
```diff
-global anonymous enum 0 foo @ 0x100
-global foo foo @ 0x100
+global enum anonymous enum 0 foo @ 0x100
+global enum foo foo @ 0x100
```

How do I deal with this in `db/cmd/cmd_avg`?

ret2libc:
@XVilka what does `global struct anonymous struct 0 foo @ 0x100` really mean? I'm not sure I understand the type.

XVilka:
`global struct anonymous struct 0 foo @ 0x100` - `global` means `global` variable, not a part of the type. `struct anonymous struct 0` is the type, where first `struct` well, just structure keyword, while `anonymous struct 0` is the type name (it's made like on purpose, since it should not be reparsed, thus invalid name), and `foo` is the variable name.

ret2libc:
> `global struct anonymous struct 0 foo @ 0x100` - `global` means `global` variable, not a part of the type. `struct anonymous struct 0` is the type, where first `struct` well, just structure keyword, while `anonymous struct 0` is the type name (it's made like on purpose, since it should not be reparsed, thus invalid name), and `foo` is the variable name.

Mmm it's a bit hard to understand, I think. Maybe the name should be quoted `"..."`.

DMaroo:
@XVilka What should be done in the case of an enum? Modifying the pretty print API just for this specific enum printing task does not seem like a good idea for me. Should I modify the test cases instead?

XVilka:
@DMaroo I think the new enum output is fine as is, just change the test for this.

DMaroo:
@XVilka I suppose that it is also fine to remove the space between `*` and `const`, like I did in `test/db/cmd/types`:

```diff
-boo_t = const char *** const *** const * const * const
+boo_t = const char ***const ***const *const *const
```

XVilka:
@DMaroo to be honest, I think with spaces it looks better. Is it possible to keep them still?

DMaroo:
> @DMaroo to be honest, I think with spaces it looks better. Is it possible to keep them still?

Yeah, it is possible to keep them. But I don't think we need to keep a separate option for that. I'll keep those spaces everywhere.

XVilka:
@DMaroo right, no need for the option, just keep these spaces everywhere.

XVilka (approved):
Once you fix the spaces - I think it's good to merge.

librz/type/type.c
```diff
@@ -791,268 +791,305 @@ RZ_API ut64 rz_type_db_get_bitsize(const RzTypeDB *typedb, RZ_NONNULL RzType *ty
...
+	rz_return_val_if_fail(typedb && type && used_types && self_ref, false);
+
+	bool is_anon = false;
 	switch (type->kind) {
 	case RZ_TYPE_KIND_IDENTIFIER: {
-		// Here it can be any of the RzBaseType
+		if (!type->identifier.name) {
+			return false;
+		}
+		is_anon = !strncmp(type->identifier.name, "anonymous ", 10);
```
ret2libc: What is 10?
If it's the length of the string `anonymous `, just use `strlen("....")`, the compiler will optimize it and it's easier to understand.

ret2libc (approved):
Just minor things.

Apart from this, PR looks good to me. It remains another big chunk of code that I think can be replaced with your new API though, and that is the code in `ctypes.c` related to `rz_core_types_as_c`. Probably all those functions `types_as_c` in cmd_type.c and ctypes.c can be removed as well, but a separate PR is ok if you still want to work on this.

librz/type/type.c
DMaroo: Thanks for pointing this out!

Also, I think I will make a different PR for refactoring all the `*types_as_c` functions. It will be easier to later bisect in case any of the refactoring breaks anything (also, since those functions don't seem to have tests, I do not want to introduce untested code (and probably bugs) in this PR).

PR #1694, opened 2021-09-15
io/io_ar: Check for NULL when doing strdup

ret2libc:
# SQUASH ME

"Inspired" by radare2 commit
https://github.com/radareorg/radare2/commit/1d747fa37190298394b10f7ad55b8fa853324c05

### DIFF
--- a/librz/io/p/io_ar.c
+++ b/librz/io/p/io_ar.c
@@ -12,8 +12,15 @@ static bool rz_io_ar_plugin_open(RzIO *io, const char *file, bool many) {
 
 static RzIODesc *rz_io_ar_open(RzIO *io, const char *file, int perm, int mode) {
 	RzIODesc *res = NULL;
-	char *url = strdup(file);
-	char *arname = strstr(url, "://") + 3;
+	char *uri = strdup(file);
+	if (!uri) {
+		return NULL;
+	}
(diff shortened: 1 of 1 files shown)


librz/io/p/io_ar.c
```diff
@@ -12,19 +12,32 @@ static bool rz_io_ar_plugin_open(RzIO *io, const char *file, bool many) {
 
 static RzIODesc *rz_io_ar_open(RzIO *io, const char *file, int perm, int mode) {
```
XVilka, line 13: I suggest to add also `RZ_NONNULL` for `file` and `rz_return_val_if_fail(file, NULL)` since `strdup()` cannot accept NULL pointers

XVilka (changes requested):

librz/io/p/io_ar.c
```diff
@@ -12,19 +12,32 @@ static bool rz_io_ar_plugin_open(RzIO *io, const char *file, bool many) {
...
 	RzIODesc *res = NULL;
```
wargio, line 15:
```suggestion
	rz_return_val_if_fail(io && file, NULL);
	RzIODesc *res = NULL;
```

XVilka (approved):

PR #1700, opened 2021-09-16
Add `not-windows-x64` category to `archos/`

kazarmy:

**Detailed description**

This pr adds the `nonwindows-x64` category to `archos/`. This is to:
1. Prevent duplicate tests in `linux-x64` and `darwin-x64`.
2. Allow the BSD OSes to join in on the Linux/Darwin testing party as well.

Imho it's quite obvious that mainstream OSes can generally be classified into 2 categories: Windows and Unix-like. Imo if anything more complicated than this is needed (e.g. not Windows and not OpenBSD), it should be done as part of a unit test.

**Test plan**

All builds are green. Tests in `archos/nonwindows-x64` are run as appropriate.

**Closing issues**

...


wargio:
how about just defining `not-` prefix ?

kazarmy:
the negation only applies to the `windows` part, not the `x64` part

wargio:
yes, and this is my point, so we can negate any arch

kazarmy:
`not-windows-x64` does also include `windows-x86`, which is not the meaning here

wargio:
hmmmm i understand what you are referring to, but we can just limit that not to arch and not bits

kazarmy:
ok as you wish 9e9e1a4711800d8fb4263961cd50d33b13b94623 ... btw, `windows` is the os, and `x64` is arch and bitness combined

wargio:
actually what i meant is that if the filename starts with `not-` then we negate the OS check.
example:
- `not-linux-x86` means run on any OS that is **not linux** which has arch `x86`
- `not-darwin-arm` means run on an OS that is **not darwin** which has arch `arm`

kazarmy:
that's what i understood too, and `not-windows-x64` is exactly according to what you meant

binrz/rz-test/load.c
```diff
@@ -612,7 +612,10 @@ static bool database_load(RzTestDatabase *db, const char *path, int depth) {
 				eprintf("Skipping %s" RZ_SYS_DIR "%s because it requires additional dependencies.\n", path, subname);
 				continue;
 			}
-			if ((!strcmp(path, "archos") || rz_str_endswith(path, RZ_SYS_DIR "archos")) && strcmp(subname, RZ_TEST_ARCH_OS)) {
+			if ((!strcmp(path, "archos") || rz_str_endswith(path, RZ_SYS_DIR "archos")) && strcmp(subname, RZ_TEST_ARCH_OS) &&
+				!(!strcmp(subname, "not-windows-x64") &&
+					!rz_str_startswith(RZ_TEST_ARCH_OS, "windows-") &&
+					rz_str_endswith(RZ_TEST_ARCH_OS, "-x64"))) {
```
ret2libc: do we really need the specific reference to windows here? Can't we rewrite it with something like `startswith(not-) && strstr(RZ_TEST_ARCH_OS, second-part-of-the-path)`?

kazarmy: no, yes we can 3794dbd2d457c240297aff94468b3e85fa7d5f89

wargio (approved):

PR #1708, opened 2021-09-17
Add `rz_strbuf_strip` to allow stripping end of `RzStrBuf`

DMaroo:

**Detailed description**

This utility function will be useful in general, but presently it is being used in #1678.

**Test plan**

Added unit tests. The unit tests are passing. CI should also pass, since no existing API has been changed.

**Closing issues**

None.

### DIFF
--- a/librz/include/rz_util/rz_strbuf.h
+++ b/librz/include/rz_util/rz_strbuf.h
@@ -26,6 +26,7 @@ RZ_API bool rz_strbuf_append_n(RzStrBuf *sb, const char *s, size_t l);
 RZ_API bool rz_strbuf_prepend(RzStrBuf *sb, const char *s);
 RZ_API bool rz_strbuf_appendf(RzStrBuf *sb, const char *fmt, ...) RZ_PRINTF_CHECK(2, 3);
 RZ_API bool rz_strbuf_vappendf(RzStrBuf *sb, const char *fmt, va_list ap);
+RZ_API bool rz_strbuf_strip(RzStrBuf *buf, int n);
 RZ_API char *rz_strbuf_get(RzStrBuf *sb);
 RZ_API RZ_OWN char *rz_strbuf_drain(RzStrBuf *sb);
 RZ_API RZ_OWN char *rz_strbuf_drain_nofree(RzStrBuf *sb);
--- a/librz/util/strbuf.c
+++ b/librz/util/strbuf.c
@@ -327,7 +327,7 @@ RZ_API bool rz_strbuf_strip(RzStrBuf *buf, int n) {
 		buf->len -= n;
 		buf->ptrlen -= n;
 	} else if (buf->len - n < sizeof(buf->buf)) {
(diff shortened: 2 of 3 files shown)


XVilka (approved):

librz/include/rz_util/rz_strbuf.h
```diff
@@ -26,6 +26,7 @@ RZ_API bool rz_strbuf_append_n(RzStrBuf *sb, const char *s, size_t l);
 RZ_API bool rz_strbuf_prepend(RzStrBuf *sb, const char *s);
 RZ_API bool rz_strbuf_appendf(RzStrBuf *sb, const char *fmt, ...) RZ_PRINTF_CHECK(2, 3);
 RZ_API bool rz_strbuf_vappendf(RzStrBuf *sb, const char *fmt, va_list ap);
+RZ_API bool rz_strbuf_strip(RzStrBuf *buf, int n);
```
wargio, line 29: i do not like the name, because i expect `rz_strbuf_strip` and `rz_strbuf_trim` to do the same thing (i.e. remove whitespaces from header and tail). i would suggest to rename it as `rz_strbuf_slice` and add start and end?

```c
RZ_API bool rz_strbuf_slice(RzStrBuf *buf, size_t start, size_t end);
```

Also @DMaroo avoid using `int` for sizes. use `size_t`

wargio (changes requested):

librz/include/rz_util/rz_strbuf.h
wargio: you can also add in the header
```c
#define rz_strbuf_slice_tail(buf, end) rz_strbuf_slice(buf, 0, end)
#define rz_strbuf_slice_head(buf, start) rz_strbuf_slice(buf, start, 0)
```

DMaroo: A `rz_strbuf_slice` function already exists.

https://github.com/rizinorg/rizin/blob/dev/librz/util/strbuf.c#L111-L130

It slices the `RzStrBuf` from `from` till `from + len`. The only reason why I created the strip function was that I needed to remove the trailing commas in the type representation of enums (#1678). I could also use this function, but for that I would have to manually access the length member of the `RzStrBuf`. To avoid that, I just created a tail strip function. What shall I do? 

TLDR: `rz_strbuf_slice` will also work in my case, but it will require me to access `RzStrBuf.len` explicitly.

ret2libc: you can use rz_strbuf_length function. I did not notice there is already `slice` API. If so, I think we don't need this new API. Is using `rz_strbuf_length` ok for you @DMaroo ?

DMaroo: Yup, makes sense. No need to have such a specific function. I'll close this PR then, and remove the `rz_strbuf_strip` commits from the other PR.

Issue #1342, opened 2021-07-21
Tracking porting the commands parser to the tree-sitter-based one


XVilka:
# Finished porting

- [x] `!` (external commands) - https://github.com/rizinorg/rizin/pull/1235
- [x] `i` (information) https://github.com/rizinorg/rizin/pull/1238
- [x] `b` (block size) https://github.com/rizinorg/rizin/pull/1745
- [x] `r` (resize) https://github.com/rizinorg/rizin/pull/1737
- [x] `y` (yank) https://github.com/rizinorg/rizin/pull/1734
- [x]  `g` (shellcodes) https://github.com/rizinorg/rizin/pull/1744
- [x] `c` (compare) https://github.com/rizinorg/rizin/pull/1854
- [x] `C` (meta information) https://github.com/rizinorg/rizin/pull/1752
- [x] `f` (flags) #1587 https://github.com/rizinorg/rizin/pull/2211
- [x] `m` (magic) #1588
- [x] `-` (open editor) #1585
- [x] `o` (open) #1589
- [x] #2363
- [x] #1584
- [x] #2834
- [x] #1583
- [x] `ae` (analysis) - https://github.com/rizinorg/rizin/issues/1533
- [x] `k` (query) - not sure if necessary as is since many parts migrated from SDB to hashtables. On the other hand, a hashtable-based query feature would also be useful.
- [x]  `d` (debug)
- [x] `p` (print) - mostly done, only `px/` left, blocked on search (`/`) commands porting first: #1590
- [x] `R` (remote) - partially ported, missing `Rg`, `Rh`, `RH`, `R!`
- [x] `x` (hexdump) - alias for `px`

# Unfinished

- [x] #1586

ret2libc:
>  0 (alias for s 0x...) - actually not sure if it's better to remove it completely

I propose to remove this. @XVilka @ITAYC0HEN @thestr4ng3r @wargio @kazarmy

wargio:
yes, is better to uniform

thestr4ng3r:
remove

PR #1734, opened 2021-09-23
Port `y` (yank) commands to the rzshell

XVilka:

**Detailed description**

- Migrate all `y` (yank) commands to the rzshell, with some slight API change
- Removed `yl` command that doesn't make sense, undocumented, and untested

**Test plan**

CI is green

Partially addresses https://github.com/rizinorg/rizin/issues/1342


ret2libc (commented):
I suggest to:
1) remove `addr` wherever it makes sense and just use `@` instead
2) try to use `@!` for blocksize instead of `len` everywhere. WDYT?

XVilka:
@ret2libc I removed the address. As for length, I think it's better to keep this parameter

wargio (approved):

ret2libc (changes requested):
Going to review

librz/core/cmd_descs/cmd_yank.yaml
```diff
@@ -0,0 +1,84 @@
...
+    type: RZ_CMD_DESC_TYPE_ARGV_STATE
+    modes:
+      - RZ_OUTPUT_MODE_STANDARD
+      - RZ_OUTPUT_MODE_RIZIN
+      - RZ_OUTPUT_MODE_JSON
+    args:
+      - name: len
+        type: RZ_CMD_ARG_TYPE_RZNUM
+        optional: true
+  - name: y!
```
ret2libc: `ye`? I think in other parts we used the `e`.

librz/core/cmd_descs/cmd_yank.yaml
```diff
@@ -0,0 +1,84 @@
...
+    cname: yank_editor
+    summary: Open cfg.editor to edit the clipboard
+    args: []
+  - name: yf
+    cname: yank_file
+    summary: Yank [blocksize] bytes from file
```
ret2libc: is this accurate? What is `len` for then? What is its relationship with `blocksize`. That's why I was suggesting to just use `@!`(at least in some cases), because that way you know there is just one size and that is the blocksize.

librz/core/cmd_descs/cmd_yank.yaml
```diff
@@ -0,0 +1,84 @@
...
+        type: RZ_CMD_ARG_TYPE_FILE
+  - name: yfa
+    cname: yank_whole_file
+    summary: Yank whole file into clipboard
+    args:
+      - name: file
+        type: RZ_CMD_ARG_TYPE_FILE
+  - name: yp
+    cname: yank_hexpairs_print
+    summary: Print contents of clipboards as hexpairs
```
ret2libc: Does it print everything if len is not specified?

librz/core/cmd_descs/cmd_yank.yaml
```diff
@@ -0,0 +1,84 @@
...
+        type: RZ_CMD_ARG_TYPE_RZNUM
+      - name: file
+        type: RZ_CMD_ARG_TYPE_FILE
+  - name: yfa
+    cname: yank_whole_file
+    summary: Yank whole file into clipboard
+    args:
+      - name: file
+        type: RZ_CMD_ARG_TYPE_FILE
+  - name: yp
```
ret2libc, line 37: Did you change it on purpose from `yq` to `yp`? `yp` was useless, right?

ret2libc: Also... Isn't this the same as `yx`?

librz/core/cmd_descs/cmd_yank.yaml
```diff
@@ -0,0 +1,84 @@
...
+  - name: yx
+    cname: yank_hex_print
+    summary: Print contents of clipboard in hexadecimal
+    args:
+      - name: len
+        type: RZ_CMD_ARG_TYPE_RZNUM
+        optional: true
+  - name: yy
+    cname: yank_paste
+    summary: Paste N bytes from yank clipboard
```
ret2libc: s/N/len ?

Also, what is used as length if len is not specified? The whole clipboard?

librz/core/yank.c
```diff
@@ -193,267 +163,188 @@ RZ_API int rz_core_yank_paste(RzCore *core, ut64 addr, int len) {
...
+	char *buf = rz_cons_hud_path(input, dir);
+	ut64 len = buf ? strlen((const char *)buf) + 1 : 0;
+	bool res = rz_core_yank_set_str(core, RZ_CORE_FOREIGN_ADDR, buf, len);
 	free(buf);
 	return res;
 }
 
-RZ_API bool rz_core_yank_hexpair(RzCore *core, const char *input) {
-	if (!input || !*input) {
+RZ_API bool rz_core_yank_hexpair(RzCore *core, const char *str) {
```
ret2libc, line 362: Maybe it's just me, but I have an hard time understanding the yank functions where they mean from clipboard to file or from file to clipboard... Could you add only the `brief` doxygen doc for some of these API? even a very short one is fine.

ret2libc:
Also, do you know what is `RZ_CORE_FOREIGN_ADDR`? Can you document it if you know?

librz/core/yank.c
```diff
@@ -94,30 +68,29 @@ static int perform_mapped_file_yank(RzCore *core, ut64 offset, ut64 len, const c
...
+	bool res = rz_core_yank_set(core, addr, (ut8 *)str, len);
 	if (res == true) {
 		ut8 zero = 0;
 		rz_buf_write_at(core->yank_buf, len - 1, &zero, sizeof(zero));
 	}
 	return res;
 }
 
-RZ_API int rz_core_yank(struct rz_core_t *core, ut64 addr, int len) {
+RZ_API bool rz_core_yank(struct rz_core_t *core, ut64 addr, int len) {
```
ret2libc: i'm wondering why this function needs to use `rz_core_seek`... Do you know? I'm trying to add a bunch of simple unit tests for yank API.

ret2libc:
```C
// SPDX-FileCopyrightText: 2021 ret2libc <sirmy15@gmail.com>

#include <rz_core.h>
#include "minunit.h"

static RzCore *fake_core_new(void) {
	RzCore *core = rz_core_new();
	RzCoreFile *file = rz_core_file_open(core, "malloc://1024", RZ_PERM_RW, 0);
	mu_assert_notnull(file, "open file");
    rz_config_set_b(core->config, "cfg.bigendian", false);
	return core;
}

bool test_yank(void) {
    RzCore *core = fake_core_new();
    bool r = rz_core_write_value_at(core, 0, 0x11223344, 4);
    mu_assert_true(r, "value should be written at 0");
    r = rz_core_yank(core, 0, 4);
    mu_assert_true(r, "4 bytes should be yanked from 0");
    r = rz_core_seek(core, 4, true);
    mu_assert_true(r, "seek should be moved to 4");
    r = rz_core_yank_paste(core, 4, 4);
    mu_assert_true(r, "clipboard content should be pasted at address 4");
    r = rz_core_block_read(core) > 0;
    mu_assert_true(r, "more than 0 bytes should be read at address 4");
    mu_assert_memeq(core->block, (const ut8 *)"\x44\x33\x22\x11", 4, "yanked bytes should be pasted at address 4");
    rz_core_free(core);
    mu_end;
}

int all_tests() {
	mu_run_test(test_yank);
	return tests_passed != tests_run;
}

mu_main(all_tests)
```

I'm trying to add a simple unit tests, but it does not work... Maybe i don't understand well the API. Do you know?

librz/core/cmd_descs/cmd_yank.yaml
XVilka: Yes

librz/core/cmd_descs/cmd_yank.yaml
XVilka: Initially I thought the same, but they produce different output. I am not sure what `yp` is, but I kept the previous output

librz/core/cmd_descs/cmd_yank.yaml
XVilka: Yes

librz/core/cmd_descs/cmd_yank.yaml
ret2libc: I'd try to review a bit the existing commands before merging this, since `y` commands are quite limited in scope it should be feasible. I'll try to help on this.

librz/core/cmd_descs/cmd_yank.yaml
XVilka: Fixed.

librz/core/cmd_descs/cmd_yank.yaml
XVilka: It wasn't accurate, now it's `len`

librz/core/cmd_descs/cmd_yank.yaml
XVilka: Ok, I returned back `yq` as `y` with `RZ_OUTPUT_MODE_QUIET`, and fixed `yp` description as printing the raw data.

XVilka:
Added missing Doxygen to the yank functions.

librz/core/yank.c
```diff
@@ -94,40 +68,61 @@ static int perform_mapped_file_yank(RzCore *core, ut64 offset, ut64 len, const c
...
+ * \p addr Address "where" the information was yanked from
+ * \p buf Contents of the buffer to be set as the clipboard
+ * \p len Length of the buffer
+ */
+RZ_API bool rz_core_yank_set(RzCore *core, ut64 addr, const ut8 *buf, ut32 len) {
+	if (!buf || !len) {
+		return false;
+	}
+	// FIXME: direct access to base should be avoided (use _sparse
+	// when you need buffer that starts at given addr)
```
ret2libc: I think this is old stuff and can be removed.

librz/core/yank.c
```diff
@@ -94,40 +68,61 @@ static int perform_mapped_file_yank(RzCore *core, ut64 offset, ut64 len, const c
...
+/* \brief Sets the contents of the yank clipboard
+ *
+ * Function sets the contents of the yank clibpoards as the NULL-terminated string
+ *
+ * \p core RzCore instance
+ * \p addr Address "where" the information was yanked from
+ * \p str Zero-terminated string to be set as the clipboard
+ * \p len Length of the buffer
+ */
+RZ_API bool rz_core_yank_set_str(RzCore *core, ut64 addr, RZ_NONNULL const char *str, ut32 len) {
```
ret2libc:
```suggestion
RZ_API bool rz_core_yank_set_str(RzCore *core, ut64 addr, RZ_NONNULL const char *str) {
```
Then call `rz_core_yank_set` with `strlen(str)+1`

librz/core/yank.c
```diff
@@ -94,40 +68,61 @@ static int perform_mapped_file_yank(RzCore *core, ut64 offset, ut64 len, const c
...
 		ut8 zero = 0;
 		rz_buf_write_at(core->yank_buf, len - 1, &zero, sizeof(zero));
 	}
 	return res;
 }
 
-RZ_API int rz_core_yank(struct rz_core_t *core, ut64 addr, int len) {
-	ut64 curseek = core->offset;
-	ut8 *buf = NULL;
+/* \brief Yank the data into the clipboard
```
ret2libc:
```suggestion
/* \brief Yank the data at address \p addr into the clipboard
```

librz/core/yank.c
```diff
@@ -193,267 +194,215 @@ RZ_API int rz_core_yank_paste(RzCore *core, ut64 addr, int len) {
...
+		break;
+	}
+	free(str);
+	return true;
 }
 
-RZ_API int rz_core_yank_cat(RzCore *core, ut64 pos) {
+/* \brief Shows yank clipboard contents as columns of hexadecimal output with headers
+ */
+RZ_API bool rz_core_yank_hexdump(RzCore *core, ut64 pos) {
```
ret2libc:
```suggestion
RZ_API bool rz_core_yank_print_hexdump(RzCore *core, ut64 pos) {
```
? WDYT?

librz/core/yank.c
```diff
@@ -193,267 +194,215 @@ RZ_API int rz_core_yank_paste(RzCore *core, ut64 addr, int len) {
...
+	rz_buf_read_at(core->yank_buf, pos, buf, ybl - pos);
+	rz_print_hexdump(core->print, pos,
+		buf, ybl - pos, 16, 1, 1);
+	return true;
 }
 
-RZ_API int rz_core_yank_cat_string(RzCore *core, ut64 pos) {
+/* \brief Shows yank clipboard contents as raw string starting from the \p pos position
+ */
+RZ_API bool rz_core_yank_cat(RzCore *core, ut64 pos) {
```
ret2libc:
```suggestion
RZ_API bool rz_core_yank_print(RzCore *core, ut64 pos) {
```

librz/core/yank.c
```diff
@@ -193,267 +194,215 @@ RZ_API int rz_core_yank_paste(RzCore *core, ut64 addr, int len) {
...
 }
 
-RZ_API int rz_core_yank_hud_file(RzCore *core, const char *input) {
-	char *buf = NULL;
-	bool res = false;
-	ut32 len = 0;
-	if (!input || !*input) {
+/* \brief Shows yank clipboard contents as zero-terminated string starting from the \p pos position
+ */
+RZ_API bool rz_core_yank_cat_string(RzCore *core, ut64 pos) {
```
ret2libc:
```suggestion
RZ_API bool rz_core_yank_print_string(RzCore *core, ut64 pos) {
```

ret2libc (commented):
LGTM. Just very minor stuff. I've also pushed a commit with 2 basics unit tests, that can be extended later by other people.

librz/core/yank.c
```diff
@@ -94,40 +68,60 @@ static int perform_mapped_file_yank(RzCore *core, ut64 offset, ut64 len, const c
...
+ *
+ * \p core RzCore instance
+ * \p addr Address "where" the information was yanked from
+ * \p buf Contents of the buffer to be set as the clipboard
+ * \p len Length of the buffer
+ */
+RZ_API bool rz_core_yank_set(RzCore *core, ut64 addr, const ut8 *buf, ut32 len) {
+	if (!buf || !len) {
+		return false;
+	}
```
ret2libc, line 83:
```suggestion
	rz_return_val_if_fail(buf, false);
	if (!len) {
		return false;
	}
```

ret2libc (approved):

librz/core/cmd_yank.c
```diff
@@ -0,0 +1,107 @@
...
+	ut64 addr = rz_num_math(core->num, argv[2]);
+	return bool2status(rz_core_yank_to(core, len, addr));
+}
+
+RZ_IPI RzCmdStatus rz_yank_hexpairs_handler(RzCore *core, int argc, const char **argv) {
+	return bool2status(rz_core_yank_hexpair(core, argv[1]));
+}
+
+RZ_IPI RzCmdStatus rz_yank_paste_handler(RzCore *core, int argc, const char **argv) {
+	ut64 pos = argc > 1 ? rz_num_math(core->num, argv[1]) : core->offset;
```
wargio: tbh, we should only have core->offset and then use `@ something` for temp seek

wargio (approved):
i'm ok with these changes, but i would like to avoid to have commands that behaves in a different way each time.

ret2libc (changes requested):
Good point from Giovanni. Also, yank paste command seems to accept a Len but you are using core offset as default instead .

XVilka:
Fixed the issue for `yy` (yank paste), added the test. Merging.

PR #1749, opened 2021-09-25
[WIP] Add RzOption implementation ##util

ghost:

**Detailed description**

The RzOption implementation is a list of macro to defined and implement some custom type.

The best practice is to define the struct in a public header and implement the struct in a private header.

**Test plan**

I have added some simple check.

**Closing issues**

...


ret2libc:
What are exactly the problem we are trying to solve with this?
Do you have specific examples of such problems?
Did you think about other solutions (e.g. returning a bool as error value and passing the return value as `*` argument)?

ghost:
IMHO, this object could be useful to represent addresses that are not defined (better alternative than UT64_MAX). The use of boolean isn't a bad idea if localized to a specific problem. The use of a common struct could harmonized a lot of APIs (the value undefined would always be None).

ret2libc:
@thestr4ng3r cc

XVilka:
Closing since abandoned

PR #1760, opened 2021-09-27
`ij`: Represent non-utf8 binary filename as number array

kazarmy:

**Detailed description**

This pr represents a non-utf8 binary filename in `ij` as a number array and unbreaks a test.

**Test plan**

All builds are green.

**Closing issues**

...

### DIFF
--- a/librz/core/cbin.c
+++ b/librz/core/cbin.c
@@ -2843,7 +2843,12 @@ RZ_API bool rz_core_file_info_print(RzCore *core, RzBinFile *binfile, RzCmdState
 		break;
 	case RZ_OUTPUT_MODE_JSON:
 		pj_o(state->d.pj);
-		pj_ks(state->d.pj, "file", filename);
+		const char *file_tag = "file";
+		if (rz_str_is_utf8(filename)) {
+			pj_ks(state->d.pj, file_tag, filename);
+		} else {
+			pj_kr(state->d.pj, file_tag, (unsigned char *)filename, strlen(filename));
+		}
 		if (desc) {
 			ut64 fsz = rz_io_desc_size(desc);
 			pj_ki(state->d.pj, "fd", desc->fd);
--- a/librz/util/str.c
+++ b/librz/util/str.c
@@ -1782,7 +1782,17 @@ RZ_API bool rz_str_is_ascii(const char *str) {
 	return true;
 }
 
-RZ_API bool rz_str_is_utf8(const char *str) {
+/**
+ * @brief Returns true if the input string is correctly UTF-8-encoded.
+ *
+ * Goes through a null-terminated string and returns false if there is a byte
+ * sequence that does not encode a valid UTF-8 code point (as determined by
+ * rz_utf8_decode()). If there are no such sequences, it returns true.
+ *
+ * @param str Input string to check for UTF-8 validity.
+ */
+RZ_API bool rz_str_is_utf8(RZ_NONNULL const char *str) {
+	rz_return_val_if_fail(str, false);
 	const ut8 *ptr = (const ut8 *)str;
 	size_t len = strlen(str);
 	while (len) {
--- a/librz/util/str.c
+++ b/librz/util/str.c
@@ -1782,6 +1782,20 @@ RZ_API bool rz_str_is_ascii(const char *str) {
 	return true;
 }
 
+RZ_API bool rz_str_is_utf8(const char *str) {
+	const ut8 *ptr = (const ut8 *)str;
+	size_t len = strlen(str);
+	while (len) {
+		int bytes = rz_utf8_decode(ptr, len, NULL);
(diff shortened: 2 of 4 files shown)


ret2libc (approved):

librz/util/str.c
```diff
@@ -1782,6 +1782,20 @@ RZ_API bool rz_str_is_ascii(const char *str) {
 	return true;
 }
 
+RZ_API bool rz_str_is_utf8(const char *str) {
```
XVilka: Doxygen please. Also `RZ_NONNULL` and `rz_return_val_if_fail(str, false)`.

XVilka (changes requested):

librz/util/str.c
kazarmy: Ok ff35d9121dd42c747e55cafef2be9ebae7ce02cc

librz/util/str.c
```diff
@@ -1782,6 +1782,30 @@ RZ_API bool rz_str_is_ascii(const char *str) {
 	return true;
 }
 
+/**
+ * @brief Returns true if the input string is correctly UTF-8-encoded.
```
XVilka: `\brief` syntax please.

XVilka (approved):

librz/util/str.c
kazarmy: Ok f2e25e8da7bfecdbf8608e22b342e8631e9f8cee

XVilka (approved):

librz/core/cbin.c
```diff
@@ -2843,7 +2843,12 @@ RZ_API bool rz_core_file_info_print(RzCore *core, RzBinFile *binfile, RzCmdState
 		break;
 	case RZ_OUTPUT_MODE_JSON:
 		pj_o(state->d.pj);
-		pj_ks(state->d.pj, "file", filename);
+		const char *file_tag = "file";
+		if (rz_str_is_utf8(filename)) {
+			pj_ks(state->d.pj, file_tag, filename);
+		} else {
+			pj_kr(state->d.pj, file_tag, (unsigned char *)filename, strlen(filename));
```
wargio, line 2850: i disagree on this. this should be encoded properly by pj_ks using `\uxxxx` codes.
https://tools.ietf.org/id/draft-ietf-json-rfc4627bis-09.html#rfc.section.7

wargio (commented):
.

wargio:
The output should be `"B\u001b\u00c2\u00a2\u00c2\u0080\u0081\u00e2\u0082\u00ac\u00f0\u009d\u0084\u009e\u00f3\u00a0\u0081\u0081\\.bin"`

```python
import json
x = "".join(map(chr, [66,27,194,162,194,128,129,226,130,172,240,157,132,158,243,160,129,129,92,46,98,105,110]))
print(json.dumps(x))
```

kazarmy:
The problem is that `\u00c2` is the byte sequence `0xc3 0x82` (UTF-8 encoding) and not just the single byte `0xc2` .

wargio:
it is true, but you can represent the whole chain also in this way, because the parser will have to convert it to a byte. the string will be preserved as it should have been.

kazarmy:
```py
>>> import json
>>> x = "".join(map(chr, [66,27,194,162,194,128,129,226,130,172,240,157,132,158,243,160,129,129,92,46,98,105,110]))
>>> print(json.dumps(x))
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\Python\lib\json\__init__.py", line 244, in dumps
    return _default_encoder.encode(obj)
  File "C:\Python\lib\json\encoder.py", line 201, in encode
    return encode_basestring_ascii(o)
UnicodeDecodeError: 'utf8' codec can't decode byte 0x81 in position 6: invalid start byte
```

wargio:
mine works..

wargio:
image

kazarmy:
> mine works..

I know. It works because it uses [`chr()`](https://docs.python.org/3.6/library/functions.html#chr), which converts numbers to the corresponding Unicode characters (code points). E.g. `chr(194) == chr(0xc2)` which is [U+00C2](https://unicode-table.com/en/00C2/). So the code isn't concatenating bytes but concatenating Unicode chars.

kazarmy:
... in Python 3.6. The code doesn't work in Python 2.7 because [`chr()`](https://docs.python.org/2.7/library/functions.html#chr) has a different definition there.

kazarmy:
Under Python 3.7:

```py
>>> bytes([66,27,194,162,194,128,129,226,130,172,240,157,132,158,243,160,129,129,92,46,98,105,110]).decode()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
UnicodeDecodeError: 'utf-8' codec can't decode byte 0x81 in position 6: invalid start byte
```

Issue #1413, opened 2021-08-09
Support separate help entries when number of arguments is different


ret2libc:
**Is your feature request related to a problem? Please describe.**
Sometimes you want to have one command name to deal with multiple operations. A classical example is the "print" command, which can often be executed without arguments to print a list of items and with an argument to print the details of one specific item. Right now we have to put together the helps of both functionalities and everything has to be the same (e.g. the modes).

```
[0x00006b60]> t?
Usage: t[?]   # Types, noreturn, signatures, C parser and more
| t[j*] [<type>] # List all types / Show type information
```

**Describe the solution you'd like**
It would be awesome to be able to specify two separate entries in the YAML files and use the same name for both of them but provide separate handlers. RzShell should be able to know which handler to call based on the number of arguments provided by the user.

```
[0x00006b60]> t?
Usage: t[?]   # Types, noreturn, signatures, C parser and more
| t[jtq*] # List all types
| t[j*] <type> # Show type information
```

Notice both commands have the same name (`t`, but they can have different modes available, different summary/description, etc.). One possible problem with that is how to get the description/summary. Because if you do `t??` what do you want to get? The first? The second? Both? This shall be thought well.

**Describe alternatives you've considered**
None.

**Additional context**
None

ret2libc:
Probably low prio at this moment. It can be improved later so I'm not going to focus on this just right now.

PR #1762, opened 2021-09-28
Port `db` commands to newshell [WIP]

DMaroo:

**Detailed description**

Porting `db` and related commands to newshell.

**Test plan**

CI should be green.

**Closing issues**

None.


DMaroo:
Currently, the build fails because the `cmd_debug.yaml` configuration is invalid. Any way to make the configuration valid, and also preserve the current state of commands.

XVilka:
```
If a command has the same name as its parent, it can only be the first child. See parent of Command cmd_debug_db
```

DMaroo:
> ```
> If a command has the same name as its parent, it can only be the first child. See parent of Command cmd_debug_db
> ```

Yes. But I want to have two commands named `db`, one which lists the breakpoints, and other which adds a breakpoint. I do not want to merge them into a single command with optional parameter. Neither do I want to make any one of them a subcommand of another, since that will mean that the subcommand won't show up when using `db?`.

ret2libc:
> > ```
> > If a command has the same name as its parent, it can only be the first child. See parent of Command cmd_debug_db
> > ```
> 
> Yes. But I want to have two commands named `db`, one which lists the breakpoints, and other which adds a breakpoint. I do not want to merge them into a single command with optional parameter. Neither do I want to make any one of them a subcommand of another, since that will mean that the subcommand won't show up when using `db?`.

This is not possible for now. I would first need to resolve https://github.com/rizinorg/rizin/issues/1413 before this is possible. Sorry. The solution we have for now is to just merge the two things with optional parameters.

DMaroo:
I have ported the `db` commands. Now I'll have to fix some integration tests (I hope they fail, rather than just pass silently).

I have also made some subtle changes to the actions done by some of the commands. This might take a bit long to review, so please take your time and review the PR carefully. Thank you.

DMaroo:
I'll probably add a few more integration tests to test the new command handlers. Other than that, there is nothing to do according to me. Please let me know if anything's missing.

XVilka:
@DMaroo could you please close this PR and send a new one from under the rizinorg account called `dist-` to trigger AppVeyor and BSD builds to check if nothing was missed in the `db` code, especially for Windows-specific commands?

PR #1766, opened 2021-09-28
Fix function return type for `rz_debug_.*_wait` functions

DMaroo:

**Detailed description**

Gets rid of warnings on Windows build. Also, makes dev green,

**Test plan**

Green dev.

**Closing issues**

None.


librz/debug/p/debug_native.c
```diff
@@ -418,7 +418,7 @@ static RzDebugReasonType rz_debug_native_wait(RzDebug *dbg, int pid) {
 }
 // FIXME: Should WAIT_ON_ALL_CHILDREN be a compilation flag instead of runtime debug config?
 #elif __linux__ && !defined(WAIT_ON_ALL_CHILDREN) // __WINDOWS__
-static RzDebugReasonType rz_debug_native_wait(RzDebug *dbg, int pid) {
+static int rz_debug_native_wait(RzDebug *dbg, int pid) {
```
yossizap: Linux native debug returns `RzDebugReasonType` as well and this is the signature in rz_debug:

```RZ_API RzDebugReasonType rz_debug_wait(RzDebug *dbg, RzBreakpointItem **bp);```

I am not sure if this should just be changed to a less indicative `int` fix a windows specific warning. I would prefer to separate this warning from the build issue.

File: librz/include/rz_debug.h [hunk 2]
```diff
@@ -387,7 +387,11 @@ typedef struct rz_debug_plugin_t {
 	int (*step)(RzDebug *dbg);
 	int (*step_over)(RzDebug *dbg);
 	int (*cont)(RzDebug *dbg, int pid, int tid, int sig);
+#if __WINDOWS__
+	RzDebugReasonType (*wait)(RzDebug *dbg, int pid);
```
XVilka, line 390: Why just not use it for everyone? This looks wrong, to be honest

XVilka (changes requested):

librz/include/rz_debug.h
DMaroo: Hmm... Agreed. So shall I modify the function signatures for the rest of the functions as well?

XVilka: yes

DMaroo: Many of the other functions, like `rz_debug_qnx_wait`, `rz_debug_bochs_wait` return integers as their return value. Directly changing the signature will probably not work.

XVilka: Why? It will solve the problem of the type mismatch.

DMaroo: It will require changing the function internals, since right now many of those return true or false or negative integers. We need to map those return values correctly to `RzDebugReasonType`. Not a big task, but could maybe introduce regressions. Anyways, I'll push a commit doing the forementioned. You can review it, then.

DMaroo: It will require changing the function internals, since right now many of those return true or false or negative integers. We need to map those return values correctly to `RzDebugReasonType`. Not a big task, but could maybe introduce regressions. Anyways, I'll push a commit doing the forementioned. You can review it, then.

XVilka (approved):

librz/debug/p/debug_bochs.c
```diff
@@ -268,9 +268,9 @@ static void bochs_debug_break(void *u) {
 	bBreak = true;
 }
 
-static int rz_debug_bochs_wait(RzDebug *dbg, int pid) {
+static RzDebugReasonType rz_debug_bochs_wait(RzDebug *dbg, int pid) {
 	if (!isBochs(dbg)) {
-		return false;
+		return RZ_DEBUG_REASON_NONE;
```
yossizap: I think that this should be RZ_DEBUG_REASON_ERROR

librz/debug/p/debug_bochs.c
```diff
@@ -322,7 +322,7 @@ static int rz_debug_bochs_wait(RzDebug *dbg, int pid) {
 	}
 	desc->data[0] = 0;
 
-	return true;
+	return RZ_DEBUG_REASON_SIGNAL;
```
yossizap: It would be better to return RZ_DEBUG_REASON_NONE in this case instead of reporting that a signal was received

librz/debug/p/debug_bf.c
```diff
@@ -120,9 +120,9 @@ static int rz_debug_bf_continue_syscall(RzDebug *dbg, int pid, int num) {
 	return true;
 }
 
-static int rz_debug_bf_wait(RzDebug *dbg, int pid) {
+static RzDebugReasonType rz_debug_bf_wait(RzDebug *dbg, int pid) {
 	/* do nothing */
-	return true;
+	return RZ_DEBUG_REASON_SIGNAL;
```
yossizap: This should be REASON_NONE to avoid the "child received signal" log. You can even remove the entire function since it doesn't do anything and `rz_debug_wait` will return REASON_UNKNOWN.

yossizap: Same issue in debug_esil/io/rap

yossizap (changes requested):

DMaroo:
@yossizap I have tried to keep the numerical value of the return values the same (`RZ_DEBUG_REASON_SIGNAL` for `true` and `RZ_DEBUG_REASON_NONE` for `false`). This was done so that any sort of conditional checks requiring the return value are evaluated as they were evaluated before the change.

I agree with your suggestion, but that might break some conditionals. However, I'll go ahead merge your changes and fix everywhere the return value is used. Please let me know in case I should be doing something different.

yossizap (approved):

PR #1794, opened 2021-10-04
Fix changing `rflags` register using `dr`

### LINKED ISSUE #1710: unable to change rflags register via dr
### Work environment

| OS/arch/bits (mandatory)                             | Ubuntu x86 32
| File format of the file you reverse (mandatory)      | elf64
| Architecture/bits of the file (mandatory)            | x86
| `rizin -v` full output, **not truncated** (mandatory)  | rizin 0.3.0-git @ linux-x86-64 commit: f211163b70eaf9748404da88b29f764581a1b757, build: 2021-09-17__17:50:56

### Expected behavior
```
[0x5595cf20077b]> dr rflags
0x00000206
[0x5595cf20077b]> dr rflags = 0x246
[0x5595cf20077b]> dr rflags
0x00000246 # this should be the value of rflags
```

### Actual behavior

```
[0x5595cf20077b]> dr rflags
0x00000206
[0x5595cf20077b]> dr rflags = 0x246
[0x5595cf20077b]> dr rflags
0x00000202 # this is the actual value
```

- Zip with the binary:
[crackme1.zip](https://github.com/rizinorg/rizin/files/7186627/crackme1.zip)
- [asciinema video](https://asciinema.org/a/436360)
- The video above in a single screenshot:
image

### Additional notes
- As seen in the video/screenshot - with `ar` it does work
- I've opened an issue for this on the cutter repo too ( [#2763](https://github.com/rizinorg/cutter/issues/2763) ), because it affects cutter gui too

DMaroo:

**Detailed description**

Pass the numerical value string as parameter to `rz_core_debug_reg_set` instead of the register name string. This now allows to change `rflags` using `dr` in the same way it can be done using `ar`.

**Test plan**

Maybe, I should add integration tests which check whether the `rflags` value changes when using `dr`? I don't know.

**Closing issues**

Closes #1710.


XVilka (approved):
Good catch! Yes, having some test for `dr rflags` and `ar rflags` would certainly be helplful.

XVilka (approved):

wargio (approved):

kazarmy (changes requested):
Sorry for being a killjoy on this but this fix doesn't work if rizin is debugging therefore it doesn't fix #1710. I think one needs to do some code injection to fix this ala `0xcc` breakpoint injection.

DMaroo:
> Sorry for being a killjoy on this but this fix doesn't work if rizin is debugging therefore it doesn't fix #1710. I think one needs to do some code injection to fix this ala `0xcc` breakpoint injection.

It does work for me when I open an ELF, and set a breakpoint and run

```
ood
dc
dr rflags
dr rflags=0x101
dr rflags
```

Am I doing something wrong?

kazarmy:
> It does work for me when I open an ELF, and set a breakpoint and run
> 
> ```
> ood
> dc
> dr rflags
> dr rflags=0x101
> dr rflags
> ```
> 
> Am I doing something wrong?

Try removing the `dc` and see whether the `dr rflags=0x101` still work.

DMaroo:
It still seems to work. Could you send a screenshot where this fails?

kazarmy:
Sure

dr_rflags

DMaroo:
Ah, I know that the `rflags` value doesn't always change, since (I suppose) some bits are reserved, and thus some values are just illegal (and I suppose 0x200 might be illegal). However, if you change it to some different value, it gets changed, as shown below

rizin_drflags

The `0x2` bit is reserved, so maybe we cannot change the `rflags` to a value like `0x200` or `0x340`. I am just guessing. I'll step through the execution to make sure it is intended behavior and not a regression.

kazarmy:
> The `0x2` bit is reserved, so maybe we cannot change the rflags to a value like `0x200` or `0x340`.

Hmm you're right!

dr_0x200

Well `rflags` starts with 0x200 ... I wonder why

kazarmy (approved):

Issue #276, opened 2020-12-27
`librz/cons/*` (TUI library/) - remove global state


ghost:
# Remove global variables

## `librz/cons/input.c`
```c
#define I rz_cons_singleton()
```

**Additional context**

Also related: https://github.com/rizinorg/rizin/issues/985

XVilka:
I don't think we need to use any external library. So far, Rizin's own is battle tested across different systems and environments, and much smaller than alternatives. We don't require all features of the "bigger" libraries, thus I think it's better to keep it.

ret2libc:
That said, it would definitely benefit from more unit tests and better API, as all the other librz modules :(

XVilka:
A good starting point would be removing global variables, e.g. in `librz/cons/editor.c`:
```c
/* TODO: remove global vars */
static char *lines = NULL;
static char *path = NULL;
static char prompt[32];
static int bytes, nlines, _n = 0;
```

PR #1826, opened 2021-10-12
Remove global state from RzCons dietline

XVilka:

**Detailed description**

Remove global state from dietline.

**Test plan**

Try to use `yank` actions in various modes in the Rizin command line.

See commands in https://book.rizin.re/basic_commands/dietline.html
In particular, commands like `Ctrl-k`, `Ctrl-w`, `Ctrl-y` in Emacs mode and `dw`, `d$`, `p` in the Vi mode.

**Closing issues**

Partially addresses https://github.com/rizinorg/rizin/issues/276

### DIFF
--- a/librz/cons/dietline.c
+++ b/librz/cons/dietline.c
@@ -27,8 +27,6 @@ typedef enum {
 	MAJOR_BREAK
 } BreakMode;
 
-bool enable_yank_pop = false;
-
 static inline bool is_word_break_char(char ch, bool mode) {
 	int i;
 	if (mode == MAJOR_BREAK) {
@@ -88,7 +86,7 @@ static void kill_word(BreakMode mode) {
 	I.buffer.length -= len;
 }
 
-static void paste(void) {
+static void paste(bool *enable_yank_pop) {
 	if (I.clipboard) {
 		char *cursor = I.buffer.data + I.buffer.index;
 		int dist = (I.buffer.data + I.buffer.length) - cursor;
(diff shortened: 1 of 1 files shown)


wargio (approved):

librz/cons/dietline.c
```diff
@@ -88,7 +86,7 @@ static void kill_word(BreakMode mode) {
 	I.buffer.length -= len;
 }
 
-static void paste(void) {
+static void paste(bool enable_yank_pop) {
```
ret2libc: in this function `enable_yank_pop` is set to true, but by moving it as an argument, the change to the variable is lost at the end of the function.

ret2libc (changes requested):

librz/cons/dietline.c
XVilka: Good catch, now I pass pointers everywhere instead!

ret2libc (approved):

ret2libc (commented):
Only one thing... maybe the property shall go in RzCons?

XVilka:
@ret2libc I thought about that, it's unclear where it should exactly go - RzCons structure already has too many parameters, there are two contexts - global and input ones. Maybe we should introduce a new context?

ret2libc:
> @ret2libc I thought about that, it's unclear where it should exactly go - RzCons structure already has too many parameters, there are two contexts - global and input ones. Maybe we should introduce a new context?

I have no idea eheh i'm not that familiar with the rzcons module to know that. I'm ok with this approach for now until someone has the time/will to really review the whole module.

ret2libc (approved):
OK. CTRL-w/CTRL-y seem a bit broken anyway, but it's unrelated to your changes.

PR #1839, opened 2021-10-15
Add rzshell_which script to look for the handler of a rzshell cmd

ret2libc:

**Detailed description**

Provide a way to quickly find the C handler of a rizin command, useful to developers to know where to modify code when dealing with existing commands.

Example:
```
Rizin Command: dbl
C handler: rz_cmd_debug_list_bp_handler
Git command to get it: git grep -nWG "^[^[:blank:]].*rz_cmd_debug_list_bp_handler(" *.c
```
(actually the values are printed in red, in the console).

**Test plan**

Try `./librz/core/cmd_descs/rzshell_find_handler.py avgt` or others.

**Closing issues**

Closes #https://github.com/rizinorg/rizin/issues/691


doc/rzshell.md
```diff
@@ -138,13 +138,16 @@ files: `cmd_descs.c` and `cmd_descs.h`.
...
 ## Where is the handler of command `x`?
 
-If you are looking for the handler of command `x`, you just have to look at
-the file
+You can use the script `librz/core/cmd_descs/rzshell_find_handler.py` to get the
+name of the function handling the specified command.
+
+If that doesn't work, please report the problem to us! However, you can still
+find the handler yourself by looking at the file
 [`librz/core/cmd_descs.yaml`](https://github.com/rizinorg/rizin/blob/9ce5003ca647cdfc181ac3c0f6206762ebb9e3e9/librz/core/cmd_descs.yaml).
```
XVilka, line 146: By the way, while you are on it, why not to refer to the `cmd_descs.yaml` from the `dev`, not a particular commit?

XVilka (approved):

doc/rzshell.md
ret2libc: I did that initially because I wanted to have a fixed reference (at that time things were changing a bit..)

thestr4ng3r:
Can we have a more convenient shortcut to this? Something like `sys/which.py avgt`?

ret2libc:
> Can we have a more convenient shortcut to this? Something like `sys/which.py avgt`?

I initially did it in sys actually, but then I realized I wanted things to be shared between cmd_descs_generate.py and this new script and having the shared functions in librz/core/cmd_descs and the script in sys was a bit of a mess... Do you know of a way to make this work?

thestr4ng3r:
There are a few possibilities:
* Just symlink from sys (doesn't work on windows probably)
* Make `sys/which` a shell script that calls the actual python one in the subdirectory
* Make `sys/which.py` call the script as a subprocess
* Make `sys/which.py` do some `sys.path.insert` magic and use the shared code or entrypoint like that

ret2libc:
> Make sys/which a shell script that calls the actual python one in the subdirectory

This

ret2libc:
Actually, I did 

> Make sys/which.py call the script as a subprocess

Please review again.

thestr4ng3r:
Can we call it just `which.py` though? But I would also like to hear all of your opinions on this.

thestr4ng3r:
Also, what about running `git grep` automatically?

ret2libc:
> Can we call it just `which.py` though? But I would also like to hear all of your opinions on this.

I prefer rzshell prefix actually... which seems a bit too generic.

ret2libc:
> Also, what about running `git grep` automatically?

We could, but... is it really useful? For example, I just want to get the function name and look it up on my editor, rather than getting it printed somehow on the terminal through git... WDYT?

XVilka:
Yes, let's keep things simple.

thestr4ng3r:
> Yes, let's keep things simple.

Simple in what sense?


> We could, but... is it really useful? For example, I just want to get the function name and look it up on my editor, rather than getting it printed somehow on the terminal through git... WDYT?

In my case, I would probably almost always follow the search by a `git grep` anyway and then open the appropriate file in vim. Not having it printed by default is fine for me, also since the function could get pretty long. I would still like to have a cmd line argument for this though, so I don't always have to copy+paste the command. That can be added later too though.

thestr4ng3r (approved):

PR #1854, opened 2021-10-18
Port `c` commands to newshell

DMaroo:
`cv`, `cV` and `cg` haven't been ported yet

 

**Detailed description**

To port:
 - [x] `cv` (implemented in `cb`)
 - [x] `cV` (implemented in `ca`)
 - [x] `cg` (to be removed)

**Test plan**

Add tests for new commands. Test coverage for `c` commands is not very good.

**Closing issues**

Tracking issue: #1342


librz/core/cmd/cmd_cmp.c
```diff
@@ -88,7 +88,7 @@ RZ_API int rz_core_cmpwatch_del(RzCore *core, ut64 addr) {
 	return ret;
 }
 
-RZ_API int rz_core_cmpwatch_show(RzCore *core, ut64 addr, int mode) {
+RZ_API void rz_core_cmpwatch_show(RzCore *core, ut64 addr, int mode) {
```
XVilka: Can you please also introduce a new enum for the mode? `RzCompareOutputMode` or something like that.

librz/core/cmd/cmd_cmp.c
```diff
@@ -955,3 +954,374 @@ RZ_IPI int rz_cmd_cmp(void *data, const char *input) {
...
+	if (val != UT64_MAX) {
+		core->num->value = val;
+		return RZ_CMD_STATUS_OK;
+	}
+	return RZ_CMD_STATUS_ERROR;
+}
+
+// c1
+RZ_IPI RzCmdStatus rz_cmd_cmp_num1_handler(RzCore *core, int argc, const char **argv) {
+	__core_cmp_bits(core, rz_num_math(core->num, argv[1]));
```
XVilka: Could it return different values?

librz/core/cmd/cmd_cmp.c
```diff
@@ -345,7 +344,7 @@ static void cmd_cmp_watcher(RzCore *core, const char *input) {
 	}
 }
 
-static int cmd_cmp_disasm(RzCore *core, const char *input, int mode) {
+static bool cmd_cmp_disasm(RzCore *core, const char *input, int mode) {
```
XVilka: Can you split this function to two: one for comparison, and another for the output?

librz/core/cmd/cmd_cmp.c
```diff
@@ -955,3 +954,374 @@ RZ_IPI int rz_cmd_cmp(void *data, const char *input) {
...
+// c1
+RZ_IPI RzCmdStatus rz_cmd_cmp_num1_handler(RzCore *core, int argc, const char **argv) {
+	__core_cmp_bits(core, rz_num_math(core->num, argv[1]));
+	return RZ_CMD_STATUS_OK;
+}
+
+// c2
+RZ_IPI RzCmdStatus rz_cmd_cmp_num2_handler(RzCore *core, int argc, const char **argv) {
+	ut16 v16 = (ut16)rz_num_math(core->num, argv[1]);
+	ut64 val = rizin_compare(core, core->block, (ut8 *)&v16, sizeof(v16), 0);
```
XVilka: I think that `rizin_compare()` would make a good API function. But I would change the arguments sligtly - e.g. drop the `core->block` argument, and use only the address.

librz/core/cmd/cmd_cmp.c
```diff
@@ -955,3 +954,374 @@ RZ_IPI int rz_cmd_cmp(void *data, const char *input) {
...
+
+// cc
+RZ_IPI RzCmdStatus rz_cmd_cmp_hex_block_handler(RzCore *core, int argc, const char **argv) {
+	ut64 addr = rz_num_math(core->num, argv[1]);
+	bool col = core->cons->columns > 123;
+	ut8 *b = malloc(core->blocksize);
+	if (b) {
+		memset(b, 0xff, core->blocksize);
+		rz_io_read_at(core->io, addr, b, core->blocksize);
+		rz_print_hexdiff(core->print, core->offset, core->block, addr, b, core->blocksize, col);
```
XVilka, line 194: @wargio please take a look at these two functions, since you wrote that.

librz/core/cmd/cmd_cmp.c
```diff
@@ -955,3 +954,374 @@ RZ_IPI int rz_cmd_cmp(void *data, const char *input) {
...
+// ccd
+RZ_IPI RzCmdStatus rz_cmd_cmp_disasm_handler(RzCore *core, int argc, const char **argv) {
+	return cmd_cmp_disasm(core, argv[1], 'c') ? RZ_CMD_STATUS_OK : RZ_CMD_STATUS_ERROR;
+}
+
+// cd
+RZ_IPI RzCmdStatus rz_cmd_chdir_handler(RzCore *core, int argc, const char **argv) {
+	static char *olddir = NULL;
+	if (argc == 2) {
+		if (!strcmp(argv[1], "-")) {
```
XVilka: I think we can move some of this logic to the `RzUtil` `rz_sys_*()`, it can be useful in many other places.

librz/core/cmd/cmd_cmp.c
```diff
@@ -955,3 +954,374 @@ RZ_IPI int rz_cmd_cmp(void *data, const char *input) {
...
+
+// clear | cls
+RZ_IPI RzCmdStatus rz_cmd_cmp_clear_screen_handler(RzCore *core, int argc, const char **argv) {
+	rz_cons_clear00();
+	return RZ_CMD_STATUS_OK;
+}
+
+// cu
+RZ_IPI RzCmdStatus rz_cmd_cmp_unified_handler(RzCore *core, int argc, const char **argv) {
+	rizin_compare_unified(core, core->offset, rz_num_math(core->num, argv[1]), core->blocksize);
```
XVilka: For these functions, `rizin_compare_unified()`, `rizin_compare_words()`, etc I recommend the same split between comparison functions that can be made an API and the printing ones that can be kept `RZ_IPI` or `static`.

XVilka (changes requested):

librz/core/cmd_descs/cmd_cmp.yaml
```diff
@@ -0,0 +1,177 @@
...
+      - RZ_OUTPUT_MODE_STANDARD
+      - RZ_OUTPUT_MODE_RIZIN
+    args:
+      - name: string
+        type: RZ_CMD_ARG_TYPE_STRING
+    details:
+      - name: Rizin output mode
+        entries:
+          - text: "Print commands to write the <string> at current offset"
+            comment: "c* <string>"
```
ret2libc: I'm tempted to say that having:
```
[0x00006b60]> c* hello
wx 68 @ 0x00006b60
wx 65 @ 0x00006b61
wx 6c @ 0x00006b62
wx 6c @ 0x00006b63
wx 6f @ 0x00006b64
wx 00 @ 0x00006b65
```

Is quite useless... Shall we remove this?

librz/core/cmd_descs/cmd_cmp.yaml
```diff
@@ -0,0 +1,177 @@
...
+  - name: c1
+    summary: Compare 8-bit data at current offset with a number
```
ret2libc:
```suggestion
    summary: Compare 8-bit data at current offset with a number <num>
```

ret2libc: same for the others? now arguments are highlighted for better visibility btw.

librz/core/cmd/cmd_cmp.c
```diff
@@ -955,3 +954,374 @@ RZ_IPI int rz_cmd_cmp(void *data, const char *input) {
...
+	return RZ_CMD_STATUS_ERROR;
+}
+
+// cat
+RZ_IPI RzCmdStatus rz_cmd_cat_handler(RzCore *core, int argc, const char **argv) {
+	if (argv[1][0] == '$') { // an alias
+		const char *old_text = rz_cmd_alias_get(core->rcmd, argv[1], 1);
+		if (old_text) {
+			rz_cons_printf("%s\n", old_text + 1);
+		}
```
ret2libc: @XVilka WDYT about this? shall we kill it? aliases should be handled by the `$` command, no?

librz/core/cmd/cmd_cmp.c
ret2libc: we already have `rz_sys_chdir`, maybe this additional logic should be placed there?

ret2libc: Probably not the `olddir` part, but the `~` handling can be put in `rz_sys_chdir`.

ret2libc (commented):
As you noticed, it seems this whole file is not very tested... You can see the coverage on https://app.codecov.io/gh/rizinorg/rizin/blob/6da244eca72902c3939f1ecc5a1b566c93e4b415/librz/core/cmd/cmd_cmp.c . I think adding a basic test for each command would be great.

librz/core/cmd/cmd_cmp.c
XVilka: Yes, lets remove it.

librz/core/cmd_descs/cmd_cmp.yaml
ret2libc: @XVilka

librz/core/cmd/cmd_cmp.c
DMaroo: Should I remove `'o'` as a mode? It seems to be unimplemented and unused.

XVilka: Yes, feel free.

librz/core/cmd/cmd_cmp.c
DMaroo: Also, identifiers containing two consecutive underscores (`__`) are reserved in C. I know that `__core_cmp_bits` would not lead to any name collision, but should we just rename it to something without the `__`?

librz/core/cmd_descs/cmd_cmp.yaml
XVilka: Safe to remove

XVilka: Safe to remove

librz/core/cmd/cmd_cmp.c
XVilka: Yes, of course.

librz/core/cmd/cmd_cmp.c
DMaroo: What shall the comparison moiety return?

XVilka: Same as in `rz_diff_` API - a list of strings with the differences. @wargio might provide a more precise example.

XVilka:
```
[XX] db/cmd/bitmask bitmask cmp data
RZ_NOPLUGINS=1 rizin -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -Qc 'b 16
wx 414243
e scr.null=1
cx 41..43 2>/dev/null
e scr.null=0
?! ?e thats fine
e scr.null=1
cx 41..11 2>/dev/null
e scr.null=0
?! ?e thats wrong
' =
-- stdout
--- expected
+++ actual
@@ -1,1 +1,0 @@
-thats fine

-- exit status: -1
[XX] db/cmd/data cmp data
RZ_NOPLUGINS=1 rizin -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -Qc 'b 16
wx 41
e scr.null=1
cx 41 2>/dev/null
e scr.null=0
?! ?e thats fine
e scr.null=1
cx 11 2>/dev/null
e scr.null=0
?! ?e thats wrong
' =
-- stdout
--- expected
+++ actual
@@ -1,1 +1,0 @@
-thats fine

-- exit status: -1
[XX] db/cmd/cmd_c cmd_c*
RZ_NOPLUGINS=1 rizin -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -Qc 'c* 1234567890' bins/elf/ls
-- stdout
--- expected
+++ actual
@@ -1,11 +1,0 @@
-wx 31 @ 0x00005ae0
-wx 32 @ 0x00005ae1
-wx 33 @ 0x00005ae2
-wx 34 @ 0x00005ae3
-wx 35 @ 0x00005ae4
-wx 36 @ 0x00005ae5
-wx 37 @ 0x00005ae6
-wx 38 @ 0x00005ae7
-wx 39 @ 0x00005ae8
-wx 30 @ 0x00005ae9
-wx 00 @ 0x00005aea

-- stderr
Command 'c*' does not exist.
```
Also forgotten `e` part in:
```
 [XX] db/cmd/regexp "/e /t\wst\d\d\d\s\w\w/i" - oldshell
RZ_NOPLUGINS=1 rizin -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -Qc 'cfg.oldshell=true
w "test123 ab"
w "Test123 ab" @444
?e
e search.in=block
b 777
"/e /t\wst\d\d\d\s\w\w/i"
' malloc://1024
-- stdout
--- expected
+++ actual
@@ -1,3 +1,0 @@
-
-0x00000000 hit0_0 "test123 ab"
-0x000001bc hit0_1 "Test123 ab"

-- stderr
Command 'cfg.oldshell=true' does not exist.
```

DMaroo:
Thank you for pointing out. However, I will fix the errors once the porting is complete.

librz/core/cmd/cmd_cmp.c
DMaroo: I have split the function into `RZ_API` functions. Please have a look and let me know if they are fine.

librz/core/cmd_descs/cmd_cmp.yaml
DMaroo: Should I also remove `cx*`, which also does the same, but with hexpairs instead?

XVilka: Yes

librz/core/cmp.c
```diff
@@ -0,0 +1,361 @@
+// SPDX-FileCopyrightText: 2021 Dhruv Maroo <dhruvmaru007@gmail.com>
+// SPDX-License-Identifier: LGPL-3.0-only
+
+#include <rz_cmp.h>
+#include <rz_asm.h>
+#include <rz_list.h>
+
+RZ_API RZ_OWN RzCompareData *rz_cmp_data_data(RZ_NONNULL RzCore *core, ut64 addr1, ut64 addr2, ut32 len) {
```
XVilka: Please add Doxygen to these functions, and a new unit test.

librz/core/cmp.c
```diff
@@ -0,0 +1,361 @@
...
+	rz_return_val_if_fail(core, NULL);
+
```
XVilka, line 19: Maybe we should check first if `addr1`, `addr2`, `addr1+len`, `addr2+len` are reachable before allocating reading?

librz/core/cmp.c
```diff
@@ -0,0 +1,361 @@
...
+error_goto:
+	free(buf1);
+	free(buf2);
+	return NULL;
+}
+
+RZ_API RZ_OWN RzCompareData *rz_cmp_data_str(RZ_NONNULL RzCore *core, ut64 addr1, RZ_NONNULL const ut8 *str, ut32 len) {
+	rz_return_val_if_fail(core && str, NULL);
+
```
XVilka, line 58: Same here

librz/core/cmp.c
```diff
@@ -0,0 +1,361 @@
...
+	cmp->addr2 = UT32_MAX;
+	cmp->same = rz_mem_eq(cmp->data1, cmp->data2, len);
+	return cmp;
+
+error_goto:
+	free(buf1);
+	return NULL;
+}
+
+RZ_API int rz_cmp_print(RZ_NONNULL RzCore *core, RZ_NONNULL const RzCompareData *cmp, RzCompareOutputMode mode) {
```
XVilka: Please add Doxygen and explain what is the return value

librz/core/cmp.c
```diff
@@ -0,0 +1,361 @@
...
+			if (hascolor) {
+				rz_cons_print(Color_RESET);
+			}
+		}
+	}
+
+	return true;
+}
+
+/* cmpwatch API */
```
XVilka: You can group functions together using Doxygen section/paragraph/group notation..

XVilka (changes requested):
Good job!

librz/core/cmp.c
DMaroo: How do I check if they are reachable? Also, won't the reading fail if they aren't reachable, so anyways, we won't be returning any garbage.

XVilka: There is a quick check in `librz/io/ioutils.c` called `rz_io_addr_is_mapped()`. This check should be faster than reading, I think.

DMaroo: Also, `rz_io_read_at` is said to be deprecated. Should I keep it?

librz/core/cmp.c
DMaroo: Need guidance on how to get a memory mapped (of a binary, or maybe manually setting it) for testing this in the unit tests.

librz/core/cmp.c
ret2libc: I'm not sure if the checks below really help. Things could be anyway unmapped in between addr1 and addr1+len, so not sure what's the added value of this. Also, don't we want to have the ability to compare two addresses even if one has only some bytes mapped? That's a difference anyway.

librz/core/cmp.c
```diff
@@ -0,0 +1,411 @@
...
+	if (!rz_io_addr_is_mapped(core->io, addr2) || !rz_io_addr_is_mapped(core->io, addr2 + len)) {
+		RZ_LOG_ERROR("Cannot reach addresses: 0x%" PFMT64x " 0x%" PFMT64x "\n", addr2, addr2 + len);
+		return NULL;
+	}
+	ut8 *buf1 = malloc(len * sizeof(ut8));
+	ut8 *buf2 = malloc(len * sizeof(ut8));
+	if (!buf1 || !buf2) {
+		goto error_goto;
+	}
+	if (!rz_io_read_at(core->io, addr1, buf1, len) || !rz_io_read_at(core->io, addr2, buf2, len)) {
```
ret2libc: Use one of the non deprecated API. We should really review those RzIO module at some point...

librz/core/cmp.c
```diff
@@ -0,0 +1,411 @@
...
+	return cmp;
+
+error_goto:
+	free(buf1);
+	free(buf2);
+	return NULL;
+}
+
+/**
+ * \brief Compare data at \p addr with the string \p str
```
ret2libc: `str` is not really a string, so I wouldn't call it so.

librz/include/rz_cmp.h
```diff
@@ -0,0 +1,29 @@
+// SPDX-FileCopyrightText: 2021 Dhruv Maroo <dhruvmaru007@gmail.com>
```
ret2libc, line 1: I think this should just go into `rz_core.h`. Also, all functions should have the prefix `rz_core_cmp`

librz/include/rz_cmp.h
```diff
@@ -0,0 +1,29 @@
...
+	ut32 len; // max 255 bytes diff in one struct
+	bool same;
+	ut8 *data1;
+	ut8 *data2;
+	ut64 addr1;
+	ut64 addr2;
+} RzCompareData;
+
+RZ_API RZ_OWN RzCompareData *rz_cmp_data_data(RZ_NONNULL RzCore *core, ut64 addr1, ut64 addr2, ut32 len);
+RZ_API RZ_OWN RzCompareData *rz_cmp_data_str(RZ_NONNULL RzCore *core, ut64 addr, RZ_NONNULL const ut8 *str, ut32 len);
```
ret2libc:
```suggestion
RZ_API RZ_OWN RzCompareData *rz_cmp_data_mem(RZ_NONNULL RzCore *core, ut64 addr, RZ_NONNULL const ut8 *str, ut32 len);
```

librz/include/rz_cmp.h
```diff
@@ -0,0 +1,29 @@
...
+RZ_API int rz_cmp_print(RZ_NONNULL RzCore *core, RZ_NONNULL const RzCompareData *cmp, RzComparePrintMode mode);
```
ret2libc:
```suggestion
RZ_API int rz_cmp_print(RZ_NONNULL RzCore *core, RZ_NONNULL const RzCompareData *cmp, RzCmdStateOutput *state);
```
like all the other `_print` functions. I don't think you need a special enum in this case. Also, by doing so you can directly pass the `state` from the command handler. The parser will take care of allocating the PJ and printing + freeing it as necessary.

librz/core/cmp.c
```diff
@@ -0,0 +1,411 @@
...
+	RzListIter *iter;
+	RzCoreCmpWatcher *w;
+	rz_list_foreach (core->watchers, iter, w) {
+		int is_diff = w->odata ? memcmp(w->odata, w->ndata, w->size) : 0;
+		switch (mode) {
+		case RZ_COMPARE_MODE_RIZIN:
+			rz_cons_printf("cw 0x%08" PFMT64x " %d %s%s\n",
+				w->addr, w->size, w->cmd, is_diff ? " # differs" : "");
+			break;
+		case RZ_COMPARE_MODE_DIFF: // diff
```
ret2libc: when is this mode used?

librz/include/rz_cmp.h
DMaroo: Since there are a handful of `cmp` functions, I thought I would keep them as a separate API component. @XVilka had approved. However, I will convert these to `rz_core_cmp_*` functions if that is better.

librz/core/cmp.c
DMaroo: Agreed, the function names need to be revised appropriately.

librz/core/cmp.c
DMaroo: `rz_core_cmpwatch_show`, which is used by `cwl`.

librz/core/cmp.c
XVilka: Well, I think we can drop those. But it will fail quicker if the boundaries are unmapped. Both approaches are fine, I think. If you prefer we could drop these checks completely.

librz/core/cmp.c
XVilka: I think if we use `rz_io_nread_at()` and compare the number of the bytes read first.

librz/include/rz_cmp.h
XVilka: @ret2libc Well, I think separating some logicaly different components is good, we already have a bunch of `RzCore` APIs that are located in other headers, for example `librz/include/rz_cmd.h`, `librz/include/rz_agraph.h`, etc.

ret2libc: As you prefer for the separate header file, though I still think `rz_core_cmp_` prefix is slightly better.

librz/core/cmp.c
ret2libc: Yes I think we should just rely on the read operation and check failure there.

librz/core/cmd/cmd_cmp.c
```diff
@@ -234,237 +76,25 @@ static int rizin_compare_unified(RzCore *core, ut64 of, ut64 od, int len) {
...
-			pj_kn(pj, "offset", core->offset + i);
-			pj_ki(pj, "rel_offset", i);
-			pj_ki(pj, "value", (int)f[i]);
-			pj_ki(pj, "cmp_value", (int)d[i]);
-			pj_end(pj);
-			break;
-		}
-	}
-	if (mode == 0) {
-		eprintf("Compare %d/%d equal bytes (%d%%)\n", eq, len, (eq / len) * 100);
```
DMaroo, line 280: Changing this `eprintf` to `rz_cons_printf` in `rz_cmp_print` results in this line being printed the last. I would want it to print before the rest of the lines. Any way to do it without storing the diffs in a buffer and printing them in the end?

DMaroo:
Now, I am adding the unit tests for the new compare API. I will be adding integration tests also. Please try to suggest any API changes before I finish writing the tests.

librz/include/rz_cmp.h
XVilka: @DMaroo could you please also change the function prefix to `rz_core_cmp`? Since it requires `RzCore`.

librz/include/rz_core.h
```diff
@@ -989,13 +989,13 @@ RZ_API bool rz_core_diff_show_function(RzCore *core, RzCore *core2, ut64 addr, b
...
 
 /* watchers */
-RZ_API void rz_core_cmpwatch_free(RzCoreCmpWatcher *w);
-RZ_API RzCoreCmpWatcher *rz_core_cmpwatch_get(RzCore *core, ut64 addr);
-RZ_API int rz_core_cmpwatch_add(RzCore *core, ut64 addr, int size, const char *cmd);
-RZ_API int rz_core_cmpwatch_del(RzCore *core, ut64 addr);
-RZ_API int rz_core_cmpwatch_update(RzCore *core, ut64 addr);
-RZ_API int rz_core_cmpwatch_show(RzCore *core, ut64 addr, int mode);
-RZ_API int rz_core_cmpwatch_revert(RzCore *core, ut64 addr);
+RZ_API void rz_core_cmpwatch_free(RZ_NONNULL RzCoreCmpWatcher *w);
```
XVilka: I think it's unnecessary for the `RzCore` to be specified as `RZ_NONNULL` as it would be implied by the `rz_core_cmp_*` prefix.

librz/core/cmp.c
```diff
@@ -0,0 +1,412 @@
...
+/**
+ * \brief Compare memory at \p addr1 with the memory at \p addr2
+ *
+ * \param core Current RzCore instance
+ * \param addr1 address to read data from
+ * \param addr2 address to read data from
+ * \param len Number of bytes to compare
+ * \return RzCompareData* A pointer to RzCompareData comparison
+ */
+RZ_API RZ_OWN RzCompareData *rz_cmp_mem_mem(RZ_NONNULL RzCore *core, ut64 addr1, ut64 addr2, ut32 len) {
```
XVilka: Same here for `RzCore`

librz/core/cmp.c
```diff
@@ -0,0 +1,412 @@
...
+	RzCoreCmpWatcher *w;
+	rz_list_foreach (core->watchers, iter, w) {
+		if (addr == w->addr) {
+			return w;
+		}
+	}
+	return NULL;
+}
+
+RZ_API bool rz_core_cmpwatch_add(RZ_NONNULL RzCore *core, ut64 addr, int size, const char *cmd) {
```
XVilka: Please add Doxygen for those

XVilka (changes requested):

librz/include/rz_cmp.h
DMaroo: Alright, sure. Also, should I keep the headers in `rz_core.h` or let them stay in `rz_cmp.h`?

XVilka: Lets keep them in `rz_cmp.h`. In the future, we should split big headers like `rz_core.h` or `rz_analysis.h` into smaller headers, like we have for `librz/include/rz_util/*.h`, for example.

librz/core/cmd_descs/cmd_cmp.yaml
```diff
@@ -0,0 +1,145 @@
...
+  - name: ca
+    summary: Compare <n> bytes of data at <addr> with the data at current offset (also return in $?)
+    cname: cmd_cmp_addr
+    args:
+      - name: addr
+        type: RZ_CMD_ARG_TYPE_RZNUM
+      - name: n
+        type: RZ_CMD_ARG_TYPE_NUM
+  - name: cb
+    summary: Compare <n> (upto 8) bytes at current offset with a number <num> (also return in $?)
```
ret2libc:
```suggestion
    summary: Compare <n> (up to 8) bytes at current offset with a number <num> (also return in $?)
```

ret2libc: Is this a comparison of bytes or of "values"?

librz/core/cmd_descs/cmd_cmp.yaml
```diff
@@ -0,0 +1,145 @@
...
+        cname: cmd_cmp_unified4
+        args:
+          - name: offset
+            type: RZ_CMD_ARG_TYPE_RZNUM
+      - name: cu8
+        summary: Compare qwords (64-bit) at <offset> with the current offset
+        cname: cmd_cmp_unified8
+        args:
+          - name: offset
+            type: RZ_CMD_ARG_TYPE_RZNUM
```
ret2libc, line 112: I think we should use the same structure as we do for `c`. If we remove `c1/c2/c4/c8`, then we should probably remove also `cu1/cu2/cu4/cu8` similarly.

librz/include/rz_core.h
```diff
@@ -989,13 +989,13 @@ RZ_API bool rz_core_diff_show_function(RzCore *core, RzCore *core2, ut64 addr, b
...
+RZ_API RzCoreCmpWatcher *rz_core_cmpwatch_get(RZ_NONNULL RzCore *core, ut64 addr);
+RZ_API bool rz_core_cmpwatch_add(RZ_NONNULL RzCore *core, ut64 addr, int size, const char *cmd);
+RZ_API bool rz_core_cmpwatch_del(RZ_NONNULL RzCore *core, ut64 addr);
+RZ_API bool rz_core_cmpwatch_update(RZ_NONNULL RzCore *core, ut64 addr);
+RZ_API void rz_core_cmpwatch_show(RZ_NONNULL RzCore *core, ut64 addr, RzOutputMode mode);
+RZ_API bool rz_core_cmpwatch_revert(RZ_NONNULL RzCore *core, ut64 addr);
```
ret2libc: do we want to move also these to rz_cmp.h?

librz/core/cmp.c
```diff
@@ -0,0 +1,465 @@
...
+ */
+RZ_API RZ_OWN RzCompareData *rz_core_cmp_mem_mem(RzCore *core, ut64 addr1, ut64 addr2, ut32 len) {
+	rz_return_val_if_fail(core, NULL);
+
+	ut8 *buf1 = malloc(len * sizeof(ut8));
+	ut8 *buf2 = malloc(len * sizeof(ut8));
+	if (!buf1 || !buf2) {
+		goto error_goto;
+	}
+	if (rz_io_nread_at(core->io, addr1, buf1, len) == -1 || rz_io_nread_at(core->io, addr2, buf2, len) == -1) {
```
ret2libc: I think you should get the results of both `rz_io_nread_at` calls and compare them to also determine if the comparison is true or false. It could happen that you just partially reads from `addr1` (or `addr2`) but you don't notice.

librz/core/cmp.c
```diff
@@ -0,0 +1,465 @@
...
+ */
+RZ_API RZ_OWN RzCompareData *rz_core_cmp_mem_data(RzCore *core, ut64 addr, RZ_NONNULL const ut8 *data, ut32 len) {
+	rz_return_val_if_fail(core && data, NULL);
+
+	ut8 *buf1 = malloc(len * sizeof(ut8));
+	if (!buf1) {
+		RZ_LOG_ERROR("Cannot read at address: 0x%" PFMT64x "\n", addr);
+		goto error_goto;
+	}
+	if (rz_io_nread_at(core->io, addr, buf1, len) == -1) {
```
ret2libc: also here, the result should be used to make sure you really read all the data you want.

librz/core/cmp.c
```diff
@@ -0,0 +1,465 @@
...
+
+/**
+ * \brief Print a comparison \p cmp according to the print mode \p mode
+ *
+ * \param core Current RzCore instance
+ * \param cmp RzCompareData instance to be printed
+ * \param mode Mode to be used (options: default, diff, json)
+ * \return int Number of lines/diffs printed (-1 if failed)
+ */
+RZ_API int rz_core_cmp_print(RzCore *core, RZ_NONNULL const RzCompareData *cmp, RzOutputMode mode) {
```
ret2libc: Maybe you already answered somewhere, but did you consider using `RzCmdStateOutput` (and RZ_CMD_DESC_TYPE_ARGV_STATE in the command instead of ARGV_MODES) ?

librz/core/cmp.c
```diff
@@ -0,0 +1,465 @@
...
+		}
+		int is_diff = w->odata ? memcmp(w->odata, w->ndata, w->size) : 0;
+		switch (mode) {
+		case RZ_OUTPUT_MODE_RIZIN:
+			rz_cons_printf("cw 0x%08" PFMT64x " %d %s%s\n",
+				w->addr, w->size, w->cmd, is_diff ? " # differs" : "");
+			break;
+		case RZ_OUTPUT_MODE_STANDARD:
+			rz_cons_printf("0x%08" PFMT64x "%s\n", w->addr, is_diff ? " modified" : "");
+			snprintf(cmd, sizeof(cmd), "%s@%" PFMT64d "!%d",
```
ret2libc:
```suggestion
			snprintf(cmd, sizeof(cmd), "%s @ %" PFMT64d " @!%d",
```

ret2libc (commented):
I think things are alright. You already did a great job, I think it's almost ready to be merged ;)

librz/core/cmd_descs/cmd_cmp.yaml
DMaroo: For doing that, I will have to rewrite/refactor `rizin_compare_words`, doing which is not as easy as it was for the `c` commands. Also, the main reason why I replaced `c1`, `c2`, `c4` and `c8` was because they were taking up unnecessary space in the help section of `c`. I don't feel a need to do the same for `cu` commands since there is no such cluttering in the help section.

librz/core/cmp.c
DMaroo: The reason why I am not using `RzCmdStateOutput` is because to me it feels easier to just pass around an enum instead of creating a `RzCmdStateOutput` object and then pass it. Also, it is possible that in future this might be called by other functions completely unrelated to command handling, therefore, I wanted this to be general and not just focused on command handling.

librz/core/cmp.c
DMaroo: Yes. I'll add the checks. That'll also make it easy for me to write the tests where the data read is less than the data expected (earlier, such test used to cause a segfault).

librz/include/rz_core.h
DMaroo: Moving them to `rz_cmp.h` is problematic since `core.c` uses the function `rz_core_cmpwatch_free`. Is it fine to `#include <rz_cmp.h>` in `core.c`?

librz/core/cmd_descs/cmd_cmp.yaml
DMaroo: It will compare the bytes at a memory location with the number `<num>`. So, if a memory location has the following bytes: 0xff32, and we input `<num>` as 0x10 and `<n>` as 2, it will show the comparison as follows (assuming little-endian):
 
 ```
0x32 -- 0x10
0xff -- 0x00
 ```

librz/core/cmd_descs/cmd_cmp.yaml
ret2libc: for `c1`/`c2`/`c4`/`c8`, it could have been possible to use the same as done for `wv` in `w?`. (or `w<1248><+->`).

librz/include/rz_core.h
ret2libc: yes, absolutely. rz_cmp is just a part of rzcore AFAIK.

librz/core/cmp.c
ret2libc: > because to me it feels easier to just pass around an enum instead of creating a RzCmdStateOutput object and then pass it

You don't need to create a `RzCmdStateOutput` in most cases because it comes directly from the command handler if you use `RZ_CMD_DESC_TYPE_ARGV_STATE`. It actually isn't strictly related to "command handling" but more to "output handling" as it makes sure that you have all the right data structures correctly initialized for the right mode you want to output (e.g. if output mode is RZ_OUTPUT_MODE_TABLE you'd have the RzTable in ->d.t, etc.).

BTW, there are functions like `rz_cmd_state_output_init` and others to quickly create a `RzCmdStateOutput` object, though in your cases you probably don't even need it.

> Also, it is possible that in future this might be called by other functions completely unrelated to command handling

I doubt that's ever going to be the case considering that the `_print` function directly prints to RzCons, thus something like Cutter can hardly calls this function to get any useful results (apart from just printing stuff).

An added bonus is that e.g. `cmd_cmp_bytes`, `cmd_cmp_addr` and other functions where you are now using `rz_core_cmp_print(...., RZ_CMD_STATE_OUTPUT_STANDARD)` could become `rz_core_cmp_print(...., state)` and you would get for free JSON support also for those commands (`ca`, `cb` and others) just by adding the modes to the cmd_cmp.yaml file.

librz/core/cmd_descs/cmd_cmp.yaml
DMaroo: Yes, but it just felt easier and neater to use a single command for `c`.

DMaroo:
Why are the integration tests `cmd_cc` and `cmd_ccc` failing? Any ideas? The text is the same. I suspect it is because of the ANSI color codes.

librz/core/cmd/cmd_cmp.c
```diff
@@ -473,384 +103,324 @@ static void __core_cmp_bits(RzCore *core, ut64 addr) {
...
+
+// ccd
+RZ_IPI RzCmdStatus rz_cmd_cmp_disasm_handler(RzCore *core, int argc, const char **argv) {
+	RzList *cmp = rz_core_cmp_disasm(core, core->offset, rz_num_math(core->num, argv[1]), core->blocksize);
+	bool ret = rz_core_cmp_disasm_print(core, cmp, false);
+	rz_list_free(cmp);
+	return ret ? RZ_CMD_STATUS_OK : RZ_CMD_STATUS_ERROR;
+}
+
+// cf
```
ret2libc, line 225: i'm not convinced we should really add these comments... These can become outdated without anyone noticing and grepping for "// cf" is meh. Easier and more accurate to use `./sys/rizin_which.py cf` IMHO.

librz/core/cmd/cmd_debug.c
```diff
@@ -4410,23 +4410,67 @@ RZ_IPI RzCmdStatus rz_cmd_debug_add_bp_handler(RzCore *core, int argc, const cha
...
+	PJ *pj = NULL;
+	if (state) {
+		mode = state->mode;
+		pj = state->d.pj;
+		rz_cmd_state_output_array_start(state);
+	}
+	rz_list_foreach (core->dbg->bp->bps, iter, b) {
+		switch (mode) {
+		case RZ_OUTPUT_MODE_STANDARD: {
+			core->dbg->bp->cb_printf("0x%08" PFMT64x " - 0x%08" PFMT64x
```
ret2libc:
```suggestion
			rz_cons_printf("0x%08" PFMT64x " - 0x%08" PFMT64x
```

librz/core/cmd/cmd_debug.c
```diff
@@ -4410,23 +4410,67 @@ RZ_IPI RzCmdStatus rz_cmd_debug_add_bp_handler(RzCore *core, int argc, const cha
...
-	case RZ_OUTPUT_MODE_JSON:
-		rz_bp_list(core->dbg->bp, 'j');
-		break;
-	case RZ_OUTPUT_MODE_QUIET:
-		rz_bp_list(core->dbg->bp, -1);
-		break;
-	default:
-		rz_warn_if_reached();
-		break;
+	RzBreakpointItem *b;
```
ret2libc: Avoid this change if not strictly necessary. I think it has not anything to do with `c` porting.

ret2libc: (saying because otherwise there are few things to fix here as well and they are not `c` related).

librz/core/cmd/cmd_cmp.c
```diff
@@ -473,384 +103,324 @@ static void __core_cmp_bits(RzCore *core, ut64 addr) {
...
+	}
+
+	int val = rz_core_cmp_print(core, cmp, state);
+	if (val != -1) {
+		ret = RZ_CMD_STATUS_OK;
+	}
+
+end:
+	rz_core_cmp_free(cmp);
+	return ret;
```
ret2libc: I suggest to create a small function that accepts `cmp` and sets core->num->value based on `cmp`, prints it and, free it and then returns the right RZ_CMD_STATUS_ value. You can use this in most handlers to avoid duplicate code and ensure more consistency (e.g. in some handlers you are setting core->num->value, in others you aren't).

librz/core/cmd/cmd_cmp.c
DMaroo: Makes sense. Had totally forgotten about the `rizin_which.py` script.

librz/core/cmd/cmd_debug.c
DMaroo: So shall I move this commits to another branch, and then open a separate PR for this?

ret2libc: yes please. there, i suggest to check if rz_bp_list is still needed, if not to delete it, then to use rz_cons_Printf instead of cb_printf.

librz/core/cmd/cmd_cmp.c
DMaroo: On a second thought, I think I should let them be there. It is makes it much easier to figure out which handler corresponds to what. Having a 3 letter comment won't hurt in my opinion. So if it isn't causing any problems, I think it should stay.

librz/core/cmd/cmd_debug.c
DMaroo: I have moved these commits in a new PR (#2012).

librz/core/cmd/cmd_cmp.c
ret2libc: > Having a 3 letter comment won't hurt in my opinion. So if it isn't causing any problems, I think it should stay.

I agree it doesn't hurt, right now, but I fear for the long-term maintainability of such comments and the fact that relying on them is not something one should do, because by relying on them you might find `cf` command but not `wz` (which did not have that comment on top) or many other commands that we converted but did not have any comment on top like this.

Either a solution is trustworthy for all cases, or it isn't IMO.

cc @thestr4ng3r @XVilka @wargio for your input as well.

ret2libc:
@DMaroo I think you messed up with the branch because you ended up including some changes done by @thestr4ng3r and if you look at the diff right now there are also changes from him related to DYLD.

librz/core/cmd/cmd_cmp.c
ret2libc: Were you able to solve this?

DMaroo:
Yes. My bad. Moving commits from this branch to the `dbl` PR might have caused it. I'll fix them.

librz/core/cmd/cmd_cmp.c
DMaroo: As per the discussion on Mattermost, I am removing setting the `core->num->value` where it is not necessary.

librz/core/cmd/cmd_cmp.c
XVilka: I don't have any strong opinion on that. Keeping them is fine, removing as well.

XVilka (approved):
Please fix the remaining tests and we are good to go, IMHO:
```
[XX] db/cmd/data cmp data
RZ_NOPLUGINS=1 rizin -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -Qc 'b 16
wx 41
e scr.null=1
cx 41 2>/dev/null
e scr.null=0
?! ?e thats fine
e scr.null=1
cx 11 2>/dev/null
e scr.null=0
?! ?e thats wrong
' =
-- stdout
--- expected
+++ actual
@@ --1,3 +-1,4 @@
 thats fine
+thats wrong

[XX] db/cmd/bitmask bitmask cmp data
RZ_NOPLUGINS=1 rizin -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -Qc 'b 16
wx 414243
e scr.null=1
cx 41..43 2>/dev/null
e scr.null=0
?! ?e thats fine
e scr.null=1
cx 41..11 2>/dev/null
e scr.null=0
?! ?e thats wrong
' =
-- stdout
--- expected
+++ actual
@@ --1,3 +-1,4 @@
 thats fine
+thats wrong
```

librz/core/cmd/cmd_cmp.c
```diff
@@ -473,384 +103,301 @@ static void __core_cmp_bits(RzCore *core, ut64 addr) {
...
+	int val = rz_core_cmp_print(core, cmp, state);
+	rz_core_cmp_free(cmp);
+	if (val == -1) {
+		goto return_goto;
+	}
+	ret = true;
+
+return_goto:
+	free(buf);
+	return ret ? RZ_CMD_STATUS_OK : RZ_CMD_STATUS_ERROR;
```
ret2libc:
```suggestion
	ret = val != -1;

return_goto:
	free(buf);
	return ret ? RZ_CMD_STATUS_OK : RZ_CMD_STATUS_ERROR;
```

librz/core/cmd/cmd_cmp.c
```diff
@@ -473,384 +103,301 @@ static void __core_cmp_bits(RzCore *core, ut64 addr) {
...
+		goto return_goto;
+	}
+	core->num->value = cmp->same ? 0 : 1;
+	int val = rz_core_cmp_print(core, cmp, state);
+	rz_core_cmp_free(cmp);
+	if (val == -1) {
+		ret = false;
+		goto return_goto;
+	}
+	ret = true;
```
ret2libc:
```suggestion
	ret = val != -1;
```

librz/core/cmd/cmd_cmp.c
```diff
@@ -473,384 +103,301 @@ static void __core_cmp_bits(RzCore *core, ut64 addr) {
...
+	RzCompareData *cmp = rz_core_cmp_mem_data(core, core->offset, buf, core->blocksize);
+	if (!cmp) {
+		goto return_goto;
+	}
+	int val = rz_core_cmp_print(core, cmp, state);
+	rz_core_cmp_free(cmp);
+	if (val == -1) {
+		goto return_goto;
+	}
+	stat = RZ_CMD_STATUS_OK;
```
ret2libc:
```suggestion
	stat = val != -1 ? RZ_CMD_STATUS_OK : RZ_CMD_STATUS_ERROR;
```

librz/core/cmd/cmd_cmp.c
```diff
@@ -473,384 +103,301 @@ static void __core_cmp_bits(RzCore *core, ut64 addr) {
...
+	mask >>= 8 - sz;
+	ut64 valid_num = num & mask;
+	RzCompareData *cmp = rz_core_cmp_mem_data(core, core->offset, (ut8 *)&valid_num, sz);
+	if (!cmp) {
+		goto end;
+	}
+	int val = rz_core_cmp_print(core, cmp, state);
+	if (val != -1) {
+		ret = RZ_CMD_STATUS_OK;
+	}
```
ret2libc:
```suggestion
	ret = val != -1 ? RZ_CMD_STATUS_OK : RZ_CMD_STATUS_ERROR;
```

librz/core/cmd/cmd_cmp.c
```diff
@@ -473,384 +103,301 @@ static void __core_cmp_bits(RzCore *core, ut64 addr) {
...
+	RzCompareData *cmp = rz_core_cmp_mem_mem(core, core->offset, rz_num_math(core->num, argv[1]), rz_num_math(core->num, argv[2]));
+	if (!cmp) {
+		return ret;
+	}
+	int val = rz_core_cmp_print(core, cmp, state);
+	rz_core_cmp_free(cmp);
+	if (val != -1) {
+		ret = RZ_CMD_STATUS_OK;
+	}
+	return ret;
```
ret2libc:
```suggestion
	return val != -1 ? RZ_CMD_STATUS_OK : RZ_CMD_STATUS_ERROR;
```

librz/core/cmd/cmd_cmp.c
```diff
@@ -473,384 +103,301 @@ static void __core_cmp_bits(RzCore *core, ut64 addr) {
...
-		if (rz_file_size(file2) <= 0) {
-			eprintf("Cannot compare with file %s\n", file2);
-			return false;
-		}
+// ca
+RZ_IPI RzCmdStatus rz_cmd_cmp_addr_handler(RzCore *core, int argc, const char **argv, RzCmdStateOutput *state) {
+	RzCmdStatus ret = RZ_CMD_STATUS_ERROR;
+	RzCompareData *cmp = rz_core_cmp_mem_mem(core, core->offset, rz_num_math(core->num, argv[1]), rz_num_math(core->num, argv[2]));
+	if (!cmp) {
+		return ret;
```
ret2libc:
```suggestion
		return RZ_CMD_STATUS_ERROR;
```

librz/core/cmd/cmd_cmp.c
```diff
@@ -473,384 +103,301 @@ static void __core_cmp_bits(RzCore *core, ut64 addr) {
...
+	return ret;
+}
 
-		if (rz_file_size(file2) <= 0) {
-			eprintf("Cannot compare with file %s\n", file2);
-			return false;
-		}
+// ca
+RZ_IPI RzCmdStatus rz_cmd_cmp_addr_handler(RzCore *core, int argc, const char **argv, RzCmdStateOutput *state) {
+	RzCmdStatus ret = RZ_CMD_STATUS_ERROR;
```
ret2libc:
```suggestion
```

librz/core/cmd/cmd_cmp.c
```diff
@@ -473,384 +103,301 @@ static void __core_cmp_bits(RzCore *core, ut64 addr) {
...
+	RzCompareData *cmp = rz_core_cmp_mem_mem(core, core->offset, rz_num_math(core->num, argv[1]), 1);
+	if (!cmp) {
+		return ret;
+	}
+	bool val = core_cmp_bits(core, cmp);
+	if (val) {
+		ret = RZ_CMD_STATUS_OK;
+	}
+	rz_core_cmp_free(cmp);
+	return ret;
```
ret2libc:
```suggestion
	rz_core_cmp_free(cmp);
	return val ? RZ_CMD_STATUS_OK : RZ_CMD_STATUS_ERROR;
```

librz/core/cmd/cmd_cmp.c
```diff
@@ -473,384 +103,301 @@ static void __core_cmp_bits(RzCore *core, ut64 addr) {
...
-			rz_core_cmd_help(core, help_message);
-			return false;
-		}
-		}
+// c1
+RZ_IPI RzCmdStatus rz_cmd_cmp_bits_handler(RzCore *core, int argc, const char **argv) {
+	RzCmdStatus ret = RZ_CMD_STATUS_ERROR;
+	RzCompareData *cmp = rz_core_cmp_mem_mem(core, core->offset, rz_num_math(core->num, argv[1]), 1);
+	if (!cmp) {
+		return ret;
```
ret2libc:
```suggestion
		return RZ_CMD_STATUS_ERROR;
```

librz/core/cmd/cmd_cmp.c
```diff
@@ -473,384 +103,301 @@ static void __core_cmp_bits(RzCore *core, ut64 addr) {
...
-				"cgo", "", "Opcode-bytes code graph diff",
-				NULL
-			};
-			rz_core_cmd_help(core, help_message);
-			return false;
-		}
-		}
+// c1
+RZ_IPI RzCmdStatus rz_cmd_cmp_bits_handler(RzCore *core, int argc, const char **argv) {
+	RzCmdStatus ret = RZ_CMD_STATUS_ERROR;
```
ret2libc:
```suggestion
```

librz/core/cmd/cmd_cmp.c
```diff
@@ -473,384 +103,301 @@ static void __core_cmp_bits(RzCore *core, ut64 addr) {
...
+	char *unescaped = strdup(argv[1]);
+	int len = rz_str_unescape(unescaped);
+	RzCompareData *cmp = rz_core_cmp_mem_data(core, core->offset, (ut8 *)unescaped, len);
+	if (!cmp) {
+		goto end;
+	}
+	int val = rz_core_cmp_print(core, cmp, state);
+	if (val != -1) {
+		ret = RZ_CMD_STATUS_OK;
+	}
```
ret2libc:
```suggestion
	ret = val != -1 ? RZ_CMD_STATUS_OK : RZ_CMD_STATUS_ERROR;
```

ret2libc (commented):
LGTM. Just few minor/last nitpicks.

ret2libc (approved):
Good job!

PR #1857, opened 2021-10-20
Test Windows CI

ret2libc:


ret2libc:
@XVilka windows CI is working... let's see with your new merge if it still works now

ret2libc:
@XVilka seems to still work... maybe a temporary failure? not sure why though.

Issue #1676, opened 2021-09-14
Convert `ec` commands to the newshell and proper API


XVilka:
In `librz/core/cmd_eval.c` there is a function called `rz_eval_color()` that still represents the "oldshell-style parsing". It should be:
- Migrated to the corresponding separate rzshell commands: `ec`, `eco`, `ecr`, etc
- Use `RzOutputMode` enum and output state for JSON
- Refactor the APIs like `rz_cons_pal_list()` to return the `RzList` instead, and introduce a separate function to print this `RzList` depending on the separate modes using `RzOutputMode` enum as well.
- Refactor the `rz_core_theme_nextpal()` to use `rz_core_list_themes()` and proper output modes. Moreover, 'n' and 'p' modes probably should be extracted in separate function[s]. In function `rz_eval_color()` for the JSON mode it's better to use the better name like `rz_core_themes_print()` and use also `rz_core_list_themes()` API and `RzOutputMode` enum

See files:
- `librz/core/cmd_eval.c`
- `librz/core/cmd_descs/cmd_eval.yaml`
- `librz/cons/pal.c`

ret2libc:
This is not perfect, yet, but I think we can close this for now. The API for `ec` could be better though...

PR #1809, opened 2021-10-08
Convert `ec` to newshell

officialcjunior:
**Your checklist for this pull request**
- [x] I've read the [guidelines for contributing](https://github.com/rizinorg/rizin/blob/master/DEVELOPERS.md) to this repository
- [x] I made sure to follow the project's [coding style](https://github.com/rizinorg/rizin/blob/master/DEVELOPERS.md#code-style)
- [ ] I've added tests that prove my fix is effective or that my feature works (if possible)
- [ ] I've updated the documentation and the [rizin book](https://github.com/rizinorg/book) with the relevant information (if needed)

**Detailed description**

Converts `ec` and its tree of commands to RzShell.

**Test plan**

- [ ] Check the implementation and fix the broken tests, if any.

**Closing issues**

Partially addresses #1676

PR #1879, opened 2021-10-23
Refactor APIs related to themes and remove global variables from `cmd_eval`

officialcjunior:

**Detailed description**

Current status:
* Refactored `rz_core_theme_nextpal()` to use `rz_core_list_themes()`
* Used PJ for `ecoj`
*  Added tests for `ecp`, `ecn` and `ecoj`
* Remove all global values from `cmd_eval.c`.
      - Added `curtheme` into `RzCore` which points to the current name of the theme.
* Sorts the themes according to the name so that it shows up the same way on all platforms.
* This part from the linked issue: `In function rz_eval_color() for the JSON mode it's better to use the better name like rz_core_themes_print() and use also rz_core_list_themes() API and RzOutputMode enum`
* Renames the APIs related to the themes to follow a standard nomenclature.
```
RZ_API RzList *rz_core_theme_list(RzCore *core);
RZ_API char *rz_core_theme_get(void);
RZ_API bool rz_core_theme_load(RzCore *core, const char *name);
RZ_API void rz_core_theme_nextpal(RzCore *core, RzConsPalSeekMode mode);
```

**Test plan**

- [x] CI passes

**Closing issues**

Partially address #1676

### DIFF
--- a/librz/core/cmd/cmd_eval.c
+++ b/librz/core/cmd/cmd_eval.c
@@ -131,88 +131,37 @@ RZ_API RZ_OWN RzList *rz_core_list_themes(RzCore *core) {
 }
 
 RZ_IPI void rz_core_theme_nextpal(RzCore *core, int mode) {
-	// TODO: use rz_core_list_themes() here instead of rewalking all the time
 	RzList *files = NULL;
 	RzListIter *iter;
 	const char *fn;
-	char *path = NULL;
 	int ctr = 0;
-	char *home = rz_str_home(RZ_HOME_THEMES RZ_SYS_DIR);
+	files = rz_core_list_themes(core);
 
-	if (mode == 'j') {
-		rz_cons_printf("[");
-	}
-	if (home) {
-		files = rz_sys_dir(home);
-		rz_list_foreach (files, iter, fn) {
-			if (*fn && *fn != '.') {
-				if (mode == 'p') {
-					const char *nfn = iter->n ? iter->n->data : NULL;
-					if (!curtheme) {
-						free(home);
-						rz_list_free(files);
-						return;
-					}
-					eprintf("%s %s %s\n", rz_str_get_null(nfn), curtheme, fn);
-					if (nfn && !strcmp(nfn, curtheme)) {
-						free(curtheme);
-						curtheme = strdup(fn);
-						RZ_FREE(home);
-						goto done;
-					}
-				} else {
-					if (!nextpal_item(core, mode, fn, ctr++)) {
-						RZ_FREE(home);
-						goto done;
-					}
+	rz_list_foreach (files, iter, fn) {
+		if (*fn && *fn != '.') {
+			if (mode == 'p') {
+				const char *nfn = iter->n ? iter->n->data : NULL;
+				if (!curtheme) {
+					rz_list_free(files);
+					return;
+				}
+				if (nfn && !strcmp(nfn, curtheme)) {
+					free(curtheme);
+					curtheme = strdup(fn);
+					goto done;
+				}
+			} else {
+				if (!nextpal_item(core, mode, fn, ctr++)) {
+					goto done;
 				}
 			}
-		}
-		RZ_FREE(home);
-		rz_list_free(files);
-		files = NULL;
-	}
-
-	path = rz_str_rz_prefix(RZ_THEMES RZ_SYS_DIR);
-	if (path) {
-		files = rz_sys_dir(path);
-		rz_list_foreach (files, iter, fn) {
-			if (*fn && *fn != '.') {
-				if (mode == 'p') {
-					const char *nfn = iter->n ? iter->n->data : NULL;
-					if (!curtheme) {
-						free(home);
-						rz_list_free(files);
-						return;
-					}
-					eprintf("%s %s %s\n", rz_str_get_null(nfn), curtheme, fn);
-					if (nfn && !strcmp(nfn, curtheme)) {
-						free(curtheme);
-						curtheme = strdup(fn);
-						goto done;
-					}
-				} else {
-					if (!nextpal_item(core, mode, fn, ctr++)) {
-						goto done;
-					}
-				}
-			}
-		}
-	}
-
-done:
-	free(path);
-	if (mode == 'l' && !curtheme && !rz_list_empty(files)) {
-		//rz_core_theme_nextpal (core, mode);
-	} else if (mode == 'n' || mode == 'p') {
-		if (curtheme) {
-			rz_core_load_theme(core, curtheme);
 		}
 	}
 	rz_list_free(files);
-	if (mode == 'j') {
-		rz_cons_printf("]\n");
-	}
+	files = NULL;
+done:
+	rz_core_load_theme(core, curtheme);
+	rz_list_free(files);
 }
 
 RZ_IPI RzCmdStatus rz_cmd_eval_color_list_handler(RzCore *core, int argc, const char **argv, RzOutputMode mode) {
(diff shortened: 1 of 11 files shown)


librz/core/cmd/cmd_eval.c
```diff
@@ -157,88 +157,37 @@ RZ_API RzList *rz_core_list_themes(RzCore *core) {
...
 
 RZ_IPI void rz_core_theme_nextpal(RzCore *core, int mode) {
-	// TODO: use rz_core_list_themes() here instead of rewalking all the time
 	RzList *files = NULL;
 	RzListIter *iter;
 	const char *fn;
-	char *path = NULL;
 	int ctr = 0;
-	char *home = rz_str_home(RZ_HOME_THEMES RZ_SYS_DIR);
+	files = rz_core_list_themes(core);
```
XVilka: Just move definition here: `RzList *files = ...`

librz/core/cmd/cmd_eval.c
```diff
@@ -318,9 +267,21 @@ RZ_IPI RzCmdStatus rz_cmd_eval_color_load_theme_handler(RzCore *core, int argc,
 		return bool2status(rz_core_load_theme(core, argv[1]));
 	}
 	switch (mode) {
-	case RZ_OUTPUT_MODE_JSON:
-		rz_core_theme_nextpal(core, 'j');
+	case RZ_OUTPUT_MODE_JSON: {
```
XVilka: Maybe makes sense to port this command to use `RzCmdStateOutput` then and just use `state.pj`?

librz/core/cmd/cmd_eval.c
```diff
@@ -318,9 +267,21 @@ RZ_IPI RzCmdStatus rz_cmd_eval_color_load_theme_handler(RzCore *core, int argc,
...
+		PJ *pj = pj_new();
+		if (!pj) {
+			break;
+		}
+		pj_a(pj);
+		themes_list = rz_core_list_themes(core);
```
XVilka: Check for NULL here please

XVilka (changes requested):

librz/core/cmd/cmd_eval.c
officialcjunior: Sure. Should I `break;` or return `RZ_CMD_STATUS_ERROR`?

librz/core/cmd/cmd_eval.c
officialcjunior: Yeah, I'll do that.

librz/core/cmd/cmd_eval.c
XVilka: ERROR, I think

officialcjunior:
https://github.com/rizinorg/rizin/pull/1809#discussion_r732423844
> In function rz_eval_color() for the JSON mode it's better to use the better name like rz_core_themes_print() and use also rz_core_list_themes() API and RzOutputMode enum.

This part is now covered, right? I'm a bit confused with which function exactly did you mean by `rz_eval_color()`

XVilka:
@officialcjunior yes, it's covered now. It was refering to the `rz_eval_color()` before your previous refactoring.

XVilka:
@officialcjunior I was suggesting to split ` nextpal_item()` function to the one that returns next/previous palette, and printing logic:
```
switch (mode) {
	case 'j': // json
		rz_cons_printf("%s\"%s\"", ctr ? "," : "", fn);
		break;
	case 'l': // list
		rz_cons_println(fn);
		break;
```
Remove these from the `nextpal_item()`. I also suggest to rename it to `pal_seek()` and a special enum - `RzConsPalSeekMode` with `RZ_CONS_PAL_SEEK_PREVIOUS` and `RZ_CONS_PAL_SEEK_NEXT` members.
And switch `mode` type to this enum. For printing another function should be used.

ret2libc:
I think it should be `rz_core_themes_list` instead of `rz_core_list_themes`.

ret2libc (commented):
Is it possible to remove the global `curtheme` as part of this PR?

officialcjunior:
> I think it should be `rz_core_themes_list` instead of `rz_core_list_themes`.

So, I was looking into all the API related to the themes:
```bash
RZ_API RzList *rz_core_list_themes(RzCore *core);
RZ_API char *rz_core_get_theme(void);
RZ_API bool rz_core_load_theme(RzCore *core, const char *name);
RZ_API void rz_core_theme_nextpal(RzCore *core, RzConsPalSeekMode mode);
```

Well, the word 'theme' seem to be at different places in the names of the APIs related to it.
Which way should we stick to?

officialcjunior:
> Is it possible to remove the global `curtheme` as part of this PR?

Yeah, there's also `getNext`, which is also global. I am assuming that we can eliminate that with by changing the logic.
But, I'm not sure about `curtheme`. Can we put in inside some thing and pass it around? How do you think we should do this?

XVilka:
I suggest to rename them all:
```c
RZ_API RzList *rz_core_theme_list(RzCore *core);
RZ_API char *rz_core_theme_get(void);
RZ_API bool rz_core_theme_load(RzCore *core, const char *name);
RZ_API void rz_core_theme_nextpal(RzCore *core, RzConsPalSeekMode mode);
```
Regarding the `curtheme`, it's either for `RzCore` or for `RzCons`. @ret2libc what do you think?

ret2libc:
> I suggest to rename them all:
> 
> ```c
> RZ_API RzList *rz_core_theme_list(RzCore *core);
> RZ_API char *rz_core_theme_get(void);
> RZ_API bool rz_core_theme_load(RzCore *core, const char *name);
> RZ_API void rz_core_theme_nextpal(RzCore *core, RzConsPalSeekMode mode);
> ```
> 
> Regarding the `curtheme`, it's either for `RzCore` or for `RzCons`. @ret2libc what do you think?

RzCore for now.

`rz_core_theme_get` will probably need to have `RzCore` as arg.

XVilka:
Please rebase, since I did some changes in the same code recently, note the changes as well: https://github.com/rizinorg/rizin/commit/91ef6da7adc2559d9a96f01f2145b437b04b9d4a

XVilka (approved):
Looks good now. Please fix the tests and this could be merged, IMHO.

officialcjunior:
On Linux and macOS, the order of the themes are different for some reason I didn't get time to look into. That makes `ecn` and `ecp` load a different theme on the different OSes even if we have currently selected the same theme. And the tests succeed in Linux because I had used Linux for developing this and added its output.

https://github.com/rizinorg/rizin/runs/4031176426?check_suite_focus=true#step:16:584

librz/core/cmd/cmd_eval.c
```diff
@@ -284,26 +222,42 @@ RZ_IPI RzCmdStatus rz_cmd_eval_color_highlight_list_handler(RzCore *core, int ar
...
 	RzListIter *th_iter;
 	const char *th;
 	if (argc == 2) {
-		return bool2status(rz_core_load_theme(core, argv[1]));
+		return bool2status(rz_core_theme_load(core, argv[1]));
 	}
-	switch (mode) {
-	case RZ_OUTPUT_MODE_JSON:
-		rz_core_theme_nextpal(core, 'j');
+	switch (state->mode) {
```
ret2libc: Get the list of themes outside the `switch` and put the switch inside the loop instead of doing the same things 3 times.

That is:
```C
themes_list = rz_core_theme_list(core);
if (!themes_list) {
...
}
rz_list_foreach (themes_list, th_iter, th) {
   switch (mode) {
   case RZ_OUTPUT_MODE_STANDARD:
      ....
      ....
  }
}
```

ret2libc (commented):
I guess the removal of global `curtheme` will be done another time, right?

librz/core/cmd/cmd_eval.c
officialcjunior: Great catch. I hope that the fix is alright now.

XVilka (approved):

XVilka:
@officialcjunior I think it might be related that these lists are unsorted. I recommend to sort list on loading themes, this will ensure the same order on all systems.

officialcjunior:
> @officialcjunior I think it might be related that these lists are unsorted. I recommend to sort list on loading themes, this will ensure the same order on all systems.

Yes. I've had the same idea. I am looking into it. :+1

librz/core/cmd/cmd_eval.c
```diff
@@ -126,93 +128,27 @@ RZ_API RZ_OWN RzList *rz_core_list_themes(RzCore *core) {
 		list_themes_in_path(list, path);
 		RZ_FREE(path);
 	}
-
+	rz_list_sort(list, cmpthemes);
```
ret2libc: just use `strcmp` directly with a cast, if necessary.

ret2libc (approved):

PR #1897, opened 2021-10-28
Add simple POST-command and changed upload path (##remote)

ogre2007:
- [x] I've added tests that prove my fix is effective or that my feature works (test plan)
- [ ] I've updated the documentation and the [rizin book](https://github.com/rizinorg/book) with the relevant information (will do after rz-pipe pr)

**Detailed description**

remote files can be uploaded now via /upload/<filename>, because root of the POST request path (/<filename>) shoud not be used for this
commands now can be transmitted via POST body, because its not good to transmit special characters via URI path (allows pipelinig, etc.)
should be enough to update rz-pipe in order to enable remote analysis

**Test plan**

start rizin as an HTTP-service:
`rizin -q -e http.port=8080 -e http.root=$(pwd) -e http.upload=1 -e http.uproot=$(pwd) -cRh`
upload file as "temp":
`curl -si http://localhost:8080/upload/temp -F upload=@/bin/true`
analyze uploaded file:
`curl -si http://localhost:8080/cmd/ -d "o temp;pd  @ main;o-*"`
and see the result.
**Closing issues**

idk


XVilka:
Please add a test

librz/core/rtr_http.c
```diff
@@ -432,31 +438,51 @@ static int rz_core_rtr_http_run(RzCore *core, int launch, int browse, const char
...
-						char *filename = rz_file_root(
-							rz_config_get(core->config, "http.uproot"),
-							rs->path + 4);
-						http_logf(core, "UPLOADED '%s'\n", filename);
-						rz_file_dump(filename, ret, retlen, 0);
-						free(filename);
-						snprintf(buf, sizeof(buf),
-							"<html><body><h2>uploaded %d byte(s). Thanks</h2>\n", retlen);
-						rz_socket_http_response(rs, 200, buf, 0, headers);
+			if (!strncmp(rs->path, "/upload/", 8)) {
```
XVilka: While you are on it, I wonder if you are willing to also split this giant function into multiple. Something like router + handlers style.

XVilka (changes requested):

XVilka:
You need also to run it as part of the rz-test, see e.g. `test/db/cmd/cmd_pipe`. Please also run `black` on this Python file.

ret2libc (changes requested):
Blocking this to discuss further:

why do we need this? Isn't it the same as moving the file with `nc` or with some other ways?

XVilka:
@ret2libc I think the idea is to allow the upload file to the remotely opened Rizin session.

ogre2007:
> Blocking this to discuss further:
> 
> why do we need this? Isn't it the same as moving the file with `nc` or with some other ways?

The idea behind this is to allow uploading data (via rz-pipe with further changes, for example), and analyze it. However, this functions already were in r2/rizin, I've just refactored them a little now and added a feature that allows transmitting command not over GET-request path, but in POST-request body, which is more handy.

ret2libc:
> I've just refactored them a little now and added a feature that allows transmitting command not over GET-request path, but in POST-request body, which is more handy.

This seems better indeed, thanks for that!

> I think the idea is to allow the upload file to the remotely opened Rizin session.

@XVilka what's the difference with doing a separate `nc`(or something) to upload the binary and then analyzing it? You are assuming anyway that on the test machine there is a way to send a POST request to a remote server.

XVilka:
You need to install them first:
```
[XX] db/cmd/cmd_http_post http-post-cmd
RZ_NOPLUGINS=1 rizin -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -Qc '!python3 scripts/http_post_cmd_upload.py
' =
-- stdout
--- expected
+++ actual
@@ -1,2 +1,0 @@
-New and old cmd results equal
-Test succeeded

-- stderr
Traceback (most recent call last):
  File "/Users/runner/work/rizin/rizin/test/scripts/http_post_cmd_upload.py", line 14, in <module>
    import requests
ModuleNotFoundError: No module named 'requests'
```

```
python -m pip install requests
````

File: librz/core/rtr_http.c [hunk 2]
```diff
@@ -3,6 +3,311 @@
...
+		if (*index == '/') {
+			path = strdup(index);
+		} else {
+		}
+	}
+	// FD IS OK HERE
+	if (rs->path[strlen(rs->path) - 1] == '/') {
+		path = (*index == '/') ? strdup(index) : rz_str_append(path, index);
+	} else {
+		//snprintf (path, sizeof (path), "%s/%s", root, rs->path);
```
XVilka: Remove all unused code also.

test/db/cmd/cmd_http_post
```diff
@@ -0,0 +1,11 @@
+NAME=http-post-cmd
+FILE==
+CMDS=<<EOF
+!python3 -m pip install requests
```
XVilka: No, this is not the way to do it. Please update these files instead:
- `.builds/*`
- `.github/workflows/*`
- `.appveyor.yml`
- `.travis.yml`

XVilka (changes requested):

librz/core/rtr_http.c
```diff
@@ -14,7 +311,7 @@ static int rz_core_rtr_http_run(RzCore *core, int launch, int browse, const char
 	char *dir;
 	int iport;
 	const char *host = rz_config_get(core->config, "http.bind");
-	const char *index = rz_config_get(core->config, "http.index");
+	//const char *index = rz_config_get(core->config, "http.index");
```
ret2libc: remove instead of commenting

XVilka:
Use files from `test/bins/` instead:
```

[XX] db/cmd/cmd_http_post http-post-cmd
RZ_NOPLUGINS=1 rizin -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -Qc '!python3 scripts/http_post_cmd_upload.py
' =
-- stdout
--- expected
+++ actual
@@ -1,1 +1,0 @@
-Test succeeded

-- stderr
Traceback (most recent call last):
  File "/Users/runner/work/rizin/rizin/test/scripts/http_post_cmd_upload.py", line 85, in <module>
    main()
  File "/Users/runner/work/rizin/rizin/test/scripts/http_post_cmd_upload.py", line 52, in main
    with open(TARGET, "rb") as file:
FileNotFoundError: [Errno 2] No such file or directory: '/bin/true'

[**]                               db/cmd/cmd_http_post    14372 OK       961 BR        1 XX       22 FX
```

XVilka:
@ogre2007 any plans to finish this?

PR #1901, opened 2021-10-28
[RzIL] implemented ao statement output

wargio:

**Detailed description**

Adds tests for RzIL linked to brainfuck and implements ao output of the stmt grammar (also in json format)


librz/il/rzil_print.c
```diff
@@ -0,0 +1,451 @@
...
+	if (sb) {
+		rz_strbuf_append(sb, "invalid");
+	} else {
+		pj_o(pj);
+		pj_ks(pj, "opcode", "invalid");
+		pj_end(pj);
+	}
+}
+
+RZ_API void rz_il_dump_list(RZ_NONNULL RzPVector *op_list, RZ_NONNULL RzStrBuf *sb) {
```
XVilka: Add Doxygen.

librz/il/rzil_print.c
```diff
@@ -0,0 +1,451 @@
+// SPDX-FileCopyrightText: 2021 RizinOrg <info@rizin.re>
```
XVilka: I suggest putting a small description of the output format with an example within a `\file` Doxygen section.

librz/il/rzil_print.c
```diff
@@ -0,0 +1,451 @@
...
+		if (i > 0) {
+			rz_strbuf_append(sb, ", ");
+		}
+		il_op_resolve(ilop, sb, NULL);
+		i++;
+	}
+	rz_strbuf_append(sb, "]");
+}
+
+RZ_API void rz_il_json_list(RZ_NONNULL RzPVector *op_list, RZ_NONNULL PJ *pj) {
```
XVilka: Doxygen.

librz/include/rz_il/rzil_vm.h
```diff
@@ -79,6 +79,9 @@ RZ_API void rz_il_vm_add_reg(RZ_NONNULL RzILVM *vm, RZ_NONNULL const char *name,
 RZ_API void rz_il_vm_store_opcodes_to_addr(RzILVM *vm, RzILBitVector *addr, RzPVector *oplist);
 RZ_API RzPVector *rz_il_make_oplist(int num, ...);
 
+RZ_API void rz_il_dump_list(RZ_NONNULL RzPVector *oplist, RZ_NONNULL RzStrBuf *sb);
```
XVilka: In general, please avoid calling printing functions with `_list()` suffix. In this case that would be `rz_il_oplist_print()` and maybe `rz_il_oplist_json()`.

XVilka (changes requested):
Overall is good, have a few nitpicks.

librz/il/rzil_print.c
ret2libc: Few comments on this whole file:
- name maybe could be just `print.c`? Maybe there are already other files with the `rzil_` prefix, but i think it's redundant.
- do we really print? maybe it's an export (as string, object, json or other methods)?
- i'm not a fan of having json and strings here in this way. we easily fall in the same thing we had before, with printing being in the modules instead of being in core. I think the `IL` module should just return the objects and the way those are printed (in a string, in a GUI toolbox, in a table, in json, etc.) should be chosen by RzCore or by Cutter. By having instead the formatting here, you are forced to use the format we are providing here. What about having this in `cil.c` or in RzCore anyway?

librz/include/rz_il/rzil_vm.h
```diff
@@ -79,6 +79,9 @@ RZ_API void rz_il_vm_add_reg(RZ_NONNULL RzILVM *vm, RZ_NONNULL const char *name,
...
+RZ_API void rz_il_json_list(RZ_NONNULL RzPVector *oplist, RZ_NONNULL PJ *pj);
```
ret2libc: I don't like too much either how these two APIs behave.

librz/il/rzil_print.c
XVilka: I think it should be split in two:
- Return `char*` in here
- Move printing to `librz/core/cil.c`

wargio: i can rename the file to rzil_dump.c/rzil_export.c but i believe this is the right place to implement this behavior.
It is true what you say, but the data is just a representation of the actual code and we are not going to use it for anything else besides ao or rz-test

ret2libc: I think the printing is already done in `cil.c`, but anyway returning an opaque string representing the RzIl ops does not allow you to do anything with it and it doesn't even allow you to easily reuse the output in other tools/libraries (e.g. Cutter) because for example Cutter may want to colorize the output or to do other things on that string. Or maybe cutter wants to create a tree instead of just displaying a string. I think having this in `IL` module limits us.

ret2libc: What about using a "visitor" pattern in this case? I think it fits pretty well with what we are trying to do here and it allows much more flexibility because the visit functions can be chosen by the caller.

librz/include/rz_il/rzil_vm.h
wargio: what is wrong with these?

librz/il/rzil_print.c
XVilka: @ret2libc we need text representation anyway. It's like assembly language of RZIL. If Cutter wants to do a syntax highlight, it should parse these expressions or just use JSON input.

ret2libc: Suppose you want to do anything else on this RzIl, is there a way right now to perform custom operations on the IL without implementing the full code in `librz/il/`? I think a visitor here allows you to do everything (text representation included).

XVilka: Visitor pattern sounds good. But we need **standard** text representation of RZIL anyway. We might want even to be able to read it from the text files in the future.

thestr4ng3r: You don't really need a specified visitor pattern here since the data structures are exposed anyway, so you can just recurse over them.
I agree that the text-format should be implemented in `il` and printing in `core`, exactly because of what @XVilka mentioned of the il needing a text-based representation in any case and this representation is part of the il specification, not the core.
Whether this specific way of formatting will stay the final representation or not is still open, but for the time being I think it serves its purpose quite well.

ret2libc: For example, I want also a RzTable output in 1 week from now. I have to change again the RzIL module. I want to look for all vars in a list of opcodes: i have to change again the RzIL module. I want to print the opcodes in a slightly different way (e.g. a tree form): i have to change again the RzIL module. I want to visit all memory operations to know what's the maximum&minimum offset of an operation involving RBP register: i have to either do the parsing of everything manually or change again the RzIL module.

By using a visitor i think you solve a lot of problems and you can implement most operations in those terms:
```C
struct rz_il_visitor_t {
   void (*visit_ite)(RzILOpIte *, struct rz_il_visitor_t *);
   void (*visit_var)(RzILOpVar *, struct rz_il_visitor_t *);
   void (*visit_and)(RzILOpAnd *, struct rz_il_visitor_t *);
   void (*visit_perform)(RzILOpPerform *, struct rz_il_visitor_t *);
   void (*visit_load)(RzILOpLoad *, struct rz_il_visitor_t *);
   void (*visit_store)(RzILOpStore *, struct rz_il_visitor_t *);
   [....]
}

void visit_load_check_offsets(RzILOpLoad *op, RzILVisitor *vis) {
   struct my_data_struct *d = vis.data;
   d.b = RZ_MAX(op.mem, d.b);
}

RzILVisitor visitor = { 0 };
struct my_data_struct some_data = {a,b,c};
visitor.visit_load = visit_load_check_offsets;
visitor.data = &some_data;
rz_il_visit(opcodes, &visitor);
int result_max_offset = visitor.data.b;
[...]
```

Just to give you an idea. In the case of the printing, you'd have to create a visitor that implements all various types of RzILOp and creates the string for each one, but this is much more flexible IMHO.


cc @thestr4ng3r @kazarmy btw

ret2libc: The reason why I think something like this makes sense is because RzILOps are essentially a tree and visiting a tree is a basic operation. the same shall be true for our RzIL. Visiting the ops should be a basic operation that one can customize as he wants.

ret2libc: > You don't really need a specified visitor pattern here since the data structures are exposed anyway, so you can just recurse over them.

This is a lot of boilerplate code that does not add value and I think there's no reason to force user of the API to reimplement all of that themselves.

> But we need standard text representation of RZIL anyway. We might want even to be able to read it from the text files in the future.

This is true, yeah. If you want the standard representation of a RzILOp there should be a `rz_il_op_string`(or `rz_il_op_repr` or whatever). If you want the representation of a RzILOpStore you should have `rz_il_op_store_string` and so on for all types. Those Op->string methods could also be easily put in a visitor structure like what mentioned above (wrapped, maybe).

What I'm saying is that I feel this is a good moment to start working on a single visiting algorithm/code and just make it flexible instead of making it inflexible.

wargio: i do not see how this can be useful, since the RzILOp is already in a tree form.

thestr4ng3r: How would you implement printing with this? Let's say we want to make s-expressions like `(+ (& 42 1337) 13)`. Then we define:
```c
void my_visit_add(RzILOpAdd *op, ...) {
    printf("(+ ");
    // how do I tell the visitor to descend here into x?
    printf(" ");
    // how do I tell the visitor to descend here into y?
    printf(")");
}
```
You could specify two functions per op, one for pre and one for post and something for inner, but that means you have to define a huge amount of functions. Also if for example you want some extra logic for indentation, you will have to repeat this in every single of these functions (even if you have the core indentation logic in a helper function you still have to call this function everywhere).
So while I don't disagree on a visitor pattern to have its usefulness for the il in general, I don't see it being beneficial here over just something like:
```c
void print_op(RzILOp *op) {
    // do indent
    printf("(");
    switch (op->code) {
    case RZ_IL_OP_ADD:
        printf("+ ");
        print_op(op->op.add->x);
        printf(" ");
        print_op(op->op.add->x);
    }
    printf(")");
} 
```

ret2libc: It is probably not super flexible to have iteration of children by default, right. Maybe something like this instead.

```
void my_visit_int(RzILOpInt *op, RzILVisitor *vis) {
   vis.a = rz_str_newf("%d", op->value);
}

void my_visit_and(RzILOpAnd *op, RzILVisitor *vis) {
   rz_il_visit(op->x, vis);
   char *a = strdup(vis.a);
   rz_il_visit(op->y, vis);
   char *b = strdup(vis.b);
   vis.a = rz_str_newf("(& %s %s)", a, b));
}

void my_visit_add(RzILOpAdd *op, RzILVisitor *vis) {
   rz_il_visit(op->x, vis);
   char *a = strdup(vis.a);
   rz_il_visit(op->y, vis);
   char *b = strdup(vis.b);
   vis.a = rz_str_newf("(+ %s %s)", a, b));
}

rz_il_visit(ops, &vis);
printf("%s\n", vis.a);
```

Anyway, maybe it was not a good idea after all :) If the text representation is a standard of the module, then it's ok to have it in `librz/il`

librz/il/rzil_print.c
```diff
@@ -0,0 +1,451 @@
...
+		pj_end(pj);
+	}
+}
+
+RZ_API void rz_il_dump_list(RZ_NONNULL RzPVector *op_list, RZ_NONNULL RzStrBuf *sb) {
+	rz_return_if_fail(sb);
+	if (!op_list) {
+		rz_strbuf_append(sb, "[]");
+		return;
+	}
```
ret2libc: I think this should be handled differently than an empty op_;list. Also, if op_list can't be null according to the API it should be in the `rz_return_if_fail` condition.

librz/il/rzil_print.c
```diff
@@ -0,0 +1,451 @@
...
+static void il_opdmp_set(RzILOp *op, RzStrBuf *sb, PJ *pj);
+static void il_opdmp_jmp(RzILOp *op, RzStrBuf *sb, PJ *pj);
+static void il_opdmp_goto(RzILOp *op, RzStrBuf *sb, PJ *pj);
+static void il_opdmp_seq(RzILOp *op, RzStrBuf *sb, PJ *pj);
+static void il_opdmp_blk(RzILOp *op, RzStrBuf *sb, PJ *pj);
+static void il_opdmp_repeat(RzILOp *op, RzStrBuf *sb, PJ *pj);
+static void il_opdmp_branch(RzILOp *op, RzStrBuf *sb, PJ *pj);
+static void il_opdmp_invalid(RzILOp *op, RzStrBuf *sb, PJ *pj);
+
+static dump_op_t il_opdmp[RZIL_OP_MAX] = {
```
ret2libc: this ordering seems a bit hard to debug. I feel it would be easy to forget to modify this whenever there's a change in the ordering or number of RzILOPCode

XVilka (approved):

librz/il/rzil_export.c
```diff
@@ -0,0 +1,545 @@
...
+		return;
+	}
+}
+
+/**
+ * Generates the string representation of the IL statements
+ * \param op_list RzPVector*, array of IL statements
+ * \param sb RzStrBuf*, a pointer to the string buffer
+ */
+RZ_API void rz_il_oplist_stringify(RZ_NULLABLE RzPVector *op_list, RZ_NONNULL RzStrBuf *sb) {
```
ret2libc, line 535: for the sake of testing and API point of view I think we should have also a `rz_il_op_stringify` and this function should just iterate over the vector and call `rz_il_op_stringify`

wargio: probably a good idea. before mergin, let me fix the memleaks

ret2libc (approved):

XVilka:
```
--- stderr ---
WARNING: (../librz/il/rzil_opcodes.c:258):rz_il_free_op: code should not be reached
ERROR: RzIl: unknown opcode 3553874899
WARNING: (../librz/il/rzil_opcodes.c:258):rz_il_free_op: code should not be reached
ERROR: RzIl: unknown opcode 3553874899
WARNING: (../librz/il/rzil_opcodes.c:258):rz_il_free_op: code should not be reached
ERROR: RzIl: unknown opcode 3553874899
WARNING: (../librz/il/rzil_opcodes.c:258):rz_il_free_op: code should not be reached
ERROR: RzIl: unknown opcode 3553874899
WARNING: (../librz/il/rzil_opcodes.c:258):rz_il_free_op: code should not be reached
ERROR: RzIl: unknown opcode 3553874899
WARNING: (../librz/il/rzil_opcodes.c:258):rz_il_free_op: code should not be reached
ERROR: RzIl: unknown opcode 3553874899
WARNING: (../librz/il/rzil_opcodes.c:258):rz_il_free_op: code should not be reached
ERROR: RzIl: unknown opcode 3553874899
malloc_consolidate(): invalid chunk size
-------
```

wargio:
i do not understand why the web UI is missing a commit but the CI is running on it

XVilka (approved):

PR #1911, opened 2021-10-31
Add local=absolute option to meson for absolute rpaths

thestr4ng3r:

**Detailed description**

OpenBSD has a different behavior for `$ORIGIN` in rpaths than for example Linux: https://www.mail-archive.com/misc@openbsd.org/msg143490.html
In particular, when an executable is run through `PATH` instead of using its direct path, it uses the cwd, which breaks our local installations. So this adds the option to have absolute rpaths in the installation:
rpath
This has the disadvantage that the installation is not relocatable anymore, so it's only used with `auto` on OpenBSD.


meson.build
```diff
@@ -545,6 +557,9 @@ rpath_lib = ''
 if use_rpath
   rpath_exe = '$ORIGIN/../' + get_option('libdir')
   rpath_lib = '$ORIGIN'
+elif use_rpath_absolute
+  rpath_exe = get_option('prefix') / get_option('libdir')
```
wargio, line 563: maybe use path join?

thestr4ng3r: The `/` operator is syntax for exactly that: https://mesonbuild.com/Reference-manual_functions.html#join_paths

wargio (approved):

meson.build
```diff
@@ -753,7 +768,8 @@ summary({
   'System zlib library': zlib_dep.found() and zlib_dep.type_name() != 'internal',
   'System zip library': libzip_dep.found() and libzip_dep.type_name() != 'internal',
   'Use ptrace-wrap': use_ptrace_wrap,
-  'Use RPATH': use_rpath,
+  'Use relative RPATH': use_rpath,
+  'Use absolute RPATH': use_rpath_absolute,
```
ret2libc: Can we have a summary line `RPATH: <relative|absolute|disabled>` instead?

ret2libc (commented):
What would be the disadvantage of always using absolute rpath? I *think* those rpaths are just used at install time, while at build time things should keep working... Could you double check this? If so, maybe we can just use absolute rpaths.

thestr4ng3r:
> What would be the disadvantage of always using absolute rpath?

You can't move the entire prefix to other locations anymore.

> I think those rpaths are just used at install time, while at build time things should keep working... Could you double check this?

Yes, they are independent of build-time rpaths. But as noted above, relative rpaths are usually better for every other platform.

ret2libc:
> What would be the disadvantage of always using absolute rpath?
You can't move the entire prefix to other locations anymore.

You should not do this anyway unless you also specify `portable=true`, which usually goes together with `static`, because the .sdb files won't be found otherwise. In which case do you want to install and then move the prefix around on your own system?

thestr4ng3r:
> You should not do this anyway unless you also specify `portable=true`, which usually goes together with `static`, because the .sdb files won't be found otherwise. In which case do you want to install and then move the prefix around on your own system?

Hm, that's right indeed. But wouldn't it also make sense to have a portable build that is not statically linked? That is probably how I would prefer to use it if I would need one, to avoid repeating the same code over multiple executables.

ret2libc:
> > You should not do this anyway unless you also specify `portable=true`, which usually goes together with `static`, because the .sdb files won't be found otherwise. In which case do you want to install and then move the prefix around on your own system?
> 
> Hm, that's right indeed. But wouldn't it also make sense to have a portable build that is not statically linked? That is probably how I would prefer to use it if I would need one, to avoid repeating the same code over multiple executables.

In theory one might want that, but in practice I'm not sure how much useful it would be. Portable builds make sense when you have to move things from one system to another, when you don't know where a user is going to place the "rizin" directory. But if you are on your own system, you can just install wherever you want (be it user-wide or system-wide).

That said, I'm not against merging this PR as-is, I was just trying to figure out if we could somehow reduce the complexity of the meson.build file by using just one way.

ret2libc (approved):

PR #1890, opened 2021-10-27
[RzIL] AVR lifter

wargio:
# squash me
 

**Your checklist for this pull request**
- [x] I've read the [guidelines for contributing](https://github.com/rizinorg/rizin/blob/master/DEVELOPERS.md) to this repository
- [x] I made sure to follow the project's [coding style](https://github.com/rizinorg/rizin/blob/master/DEVELOPERS.md#code-style)
- [x] I've added tests that prove my fix is effective or that my feature works (if possible)
- [ ] I've updated the documentation and the [rizin book](https://github.com/rizinorg/book) with the relevant information (if needed)

**Detailed description**

Adds AVR support
- `des` and `spm` ops are not implemented due the complexity behind them.

Most 80% of the ops were tested comparing the results with a debugger, remaining ones using simulators.

**Test plan**

Added IL statements and tests under `db/rzil/avr`

PR #1920, opened 2021-11-01
Unscramble Register Printing and Flags Logic

thestr4ng3r:

**Detailed description**

The final problem I am trying to solve by this is a lot of registers not being shown when remote-debugging 6502:
Bildschirmfoto 2021-11-01 um 13 01 30
While it should actually look something like this, perhaps without the individual CZIDVN flags, but a, x and y are the main registers of 6502:
Bildschirmfoto 2021-11-01 um 13 30 09

The reason they are not shown is that only registers of the size of `analysis->bits` are shown. For `ar`, there was a 6502-specific (`!strcmp(core->analysis->cur->arch, "6502") && bits == 8`) hack in core for this, but that doesn't apply for debug and also it's dirty obv.
Another place where similar filtering applied was when setting flags. Here I think it actually makes sense to do it because flags that can't possibly contain an address probably also make no sense to have a flag.

So the changes done here are:
* Removed all the strcmp-hacks from core and replace by `rz_analysis_get_address_bits()` for the filtering logic for flags. Plugins may supply address sizes different to their bitness (8-bit 6502 with 16-bit addrs, arm thumb having 32-bit addrs, ...). This function can also be used for other places where the address size is needed.
* `dr` and `ar` just show all gpr registers

The downside from the second point above is now that e.g. for x86, you always see `rax`, `eax`, and `ax` while you probably only want to see `rax` by default (see the tests diff). Here are two ideas hot to solve this better:
* Add the info that a register is "less important" to the register profile and hide these in the plain `ar`/`dr` commands
* Check whether a register is contained entirely within another one and only show the bigger one then

I'm slightly leaning towards the second solution right now. Please comment with thoughts or other suggestions.


XVilka:
I think, second solution is the better one.

wargio:
yeah, second one is much better.

librz/core/core_private.h
```diff
@@ -119,7 +119,7 @@ RZ_IPI RzCmdStatus rz_core_binldr_plugin_print(const RzBinLdrPlugin *ld, RzCmdSt
 /* cdebug.c */
 RZ_IPI bool rz_core_debug_reg_set(RzCore *core, const char *regname, ut64 val, const char *strval);
 RZ_IPI bool rz_core_debug_reg_list(RzCore *core, int type, int size, PJ *pj, int rad, const char *use_color);
-RZ_IPI void rz_core_debug_regs2flags(RzCore *core, int bits);
```
officialcjunior, line 122: Since you touched `rz_core_debug_regs2flags`, I hope you can add a oneliner Doxygen doc for the API.

thestr4ng3r:
Ok, second solution is implemented now. There are two relevant changes in test results now:

In ppc,  instead of e.g. `ibat0l`/`ibat0h` registers, we get full `ibat0` ones: https://github.com/rizinorg/rizin/pull/1920/files#diff-e1956f476af1cd42f8405481dfb245f5add863d88a535fa1c1da7a68ec7dc6e4L301-R302
I'm not sure which is better here.

In avr, instead of e.g. `r0` and `r1`, we only get the `r1_r0` which covers both: https://github.com/rizinorg/rizin/pull/1920/files#diff-97c2da55cfebebd53e703431d82be100350be49badac1a55bfbd03f2308f922dL757-L795
These registers are apparently only hacks for esil, e.g. from c1c75f76d71535cd139dc016bc8d26e2f3a169a0, so they should just be removed and replaced by esil and/shift ops if possible.

wargio:
> Ok, second solution is implemented now. There are two relevant changes in test results now:
> 
> In ppc, instead of e.g. `ibat0l`/`ibat0h` registers, we get full `ibat0` ones: https://github.com/rizinorg/rizin/pull/1920/files#diff-e1956f476af1cd42f8405481dfb245f5add863d88a535fa1c1da7a68ec7dc6e4L301-R302 I'm not sure which is better here.
> 
> In avr, instead of e.g. `r0` and `r1`, we only get the `r1_r0` which covers both: https://github.com/rizinorg/rizin/pull/1920/files#diff-97c2da55cfebebd53e703431d82be100350be49badac1a55bfbd03f2308f922dL757-L795 These registers are apparently only hacks for esil, e.g. from [c1c75f7](https://github.com/rizinorg/rizin/commit/c1c75f76d71535cd139dc016bc8d26e2f3a169a0), so they should just be removed and replaced by esil and/shift ops if possible.

definitely having just the full reg without having it the low/high value is better (this applied to MIPS/AVR/PPC, etc..).

thestr4ng3r:
Good, then I will check if it's easy to remove the avr esil hacks, but in light of #1890 I will try not to spend too much time on it and maybe implement a workaround.

librz/analysis/p/analysis_arm_cs.c
```diff
@@ -4507,6 +4507,10 @@ static RzList *analysis_preludes(RzAnalysis *analysis) {
 	return l;
 }
 
+static int address_bits(RzAnalysis *analysis, int bits) {
+	return bits == 16 ? 32 : -1;
```
XVilka, line 4512: Why `-1`?

thestr4ng3r:
Default `ar` on AVR now looks like this:
```
r0 = 0x00000000
r1 = 0x00000000
r2 = 0x00000000
r3 = 0x00000000
r4 = 0x00000000
r5 = 0x00000000
r6 = 0x00000000
r7 = 0x00000000
r8 = 0x00000000
r9 = 0x00000000
r10 = 0x00000000
r11 = 0x00000000
r12 = 0x00000000
r13 = 0x00000000
r14 = 0x00000000
r15 = 0x00000000
r16 = 0x00000000
r17 = 0x00000000
r18 = 0x00000000
r19 = 0x00000000
r20 = 0x00000000
r21 = 0x00000000
r22 = 0x00000000
r23 = 0x00000000
r24 = 0x00000000
r25 = 0x00000000
x = 0x00000000
y = 0x00000000
z = 0x00000000
pc = 0x00000000
sp = 0x00000000
sreg = 0x00000000
rampx = 0x00000000
rampy = 0x00000000
rampz = 0x00000000
rampd = 0x00000000
eind = 0x00000000
_prog = 0x00000000
_page = 0x00000000
_eeprom = 0x00000000
_ram = 0x00000000
_io = 0x00000000
_sram = 0x00000000
spmcsr = 0x00000000
```

Much better imo. r26, ..., r31 are omitted because they are overlapped by x, y and z, but I think that is fine because these registers are defined by the ISA and not just random hacks. All r-regs can be shown with `ar 8`.

librz/analysis/p/analysis_arm_gnu.c
```diff
@@ -473,13 +473,18 @@ static int archinfo(RzAnalysis *analysis, int q) {
 	return 4; // XXX
 }
 
+static int address_bits(RzAnalysis *analysis, int bits) {
+	return bits == 16 ? 32 : -1;
```
XVilka, line 478: Why `-1`?

librz/core/cdebug.c
```diff
@@ -251,19 +238,71 @@ RZ_IPI bool rz_core_debug_reg_set(RzCore *core, const char *regname, ut64 val, c
...
+			// nope
+			return true;
+		}
+	}
+	// from_list ends before from_tree, so it is covered and smaller
+	return false;
+}
+
+/// Filter out all registers that are smaller than but covered entirely by some other register
+static RZ_OWN RzList *regs_filter_covered(RZ_BORROW const RzList /* <RzRegItem> */ *regs) {
```
XVilka, line 270: We should convert `regs` to vector in the future btw.

librz/core/cdebug.c
```diff
@@ -251,19 +238,71 @@ RZ_IPI bool rz_core_debug_reg_set(RzCore *core, const char *regname, ut64 val, c
...
+			// foreach_reg_cb break-ed so it found a cover
+			continue;
+		}
+		rz_list_push(ret, item);
+	}
+	rz_interval_tree_fini(&t);
+	return ret;
+}
+
+RZ_IPI bool rz_core_debug_reg_list(RzCore *core, int type, int size, bool skip_covered, PJ *pj, int rad, const char *use_color) {
```
XVilka, line 300: What about killing `rad` and use `RzOutputMode`, or even better `RzCmdStateOutput`, to avoid passing `pj`, and adding table mode in the future. Moreover, some of the options like `skip_covered` and `use_color` could be combined in some kind of structure.

librz/core/cmd/cmd_analysis.c
```diff
@@ -3408,8 +3408,10 @@ RZ_IPI int rz_cmd_analysis_fcn(void *data, const char *input) {
 	return true;
 }
 
-// size: 0: bits; -1: any; >0: exact size
-static void __analysis_reg_list(RzCore *core, int type, int bits, char mode) {
+/**
+ * \param bits if > 0, show only regs with this bit size, otherwise all
+ */
+static void print_reg_list(RzCore *core, int type, int bits, bool skip_covered, char mode) {
```
XVilka, line 3414: Same for `mode` and options.

XVilka (changes requested):
Overall looks good.

librz/analysis/p/analysis_arm_cs.c
thestr4ng3r: Added a comment.

librz/analysis/p/analysis_arm_gnu.c
thestr4ng3r: same

librz/core/cdebug.c
thestr4ng3r: Maybe. `RzList` isn't too bad here though since the number of regs should generally be well-controlled.

librz/core/cdebug.c
thestr4ng3r: Agreed, there is definitely more to improve here. Also converting `ar`/`dr` to newshell and merging their code into one.

XVilka (approved):

Issue #1918, opened 2021-11-01
Implement missing RZIL operation handlers


XVilka:
## Bitvectors
- [x] SLE https://github.com/rizinorg/rizin/pull/2094
- [x] ULE https://github.com/rizinorg/rizin/pull/2094
- [ ] CONCAT (postponed until there is a real need)
- [x] APPEND  https://github.com/rizinorg/rizin/pull/1936
## Effects
- [x] JMP https://github.com/rizinorg/rizin/pull/1924
- [x] BLK https://github.com/rizinorg/rizin/pull/2119
- [x] REPEAT https://github.com/rizinorg/rizin/pull/2119

PR #1924, opened 2021-11-01
[RzIL] fix jmp & added bitvector type due limitation of int type

### LINKED ISSUE #1859: Use RzILBitVector directly in rzil_op_int_t
The int opcode currently looks like this:
```c
/**
 *  \struct rzil_op_int_t
 *  \brief op structure for `int` (val int : 's Bitv.t Value.sort -> word -> 's bitv)
 *
 *  int s x is a bitvector constant x of sort s.
 */
struct rzil_op_int_t {
	ut32 length; ///< s -- sort(type), length of bitvector
	int value; ///< x -- value of bitvector
};
```

It simply represents a constant bitvector value and in practice will be converted to RzILBitVector for further usage:
```c
void *rz_il_handler_int(RzILVM *vm, RzILOp *op, RzILOpArgType *type) {
	RzILOpInt *op_int = op->op.int_;

	ut32 length = op_int->length;
	int value = op_int->value;
	RzILBitVector *bv = rz_il_bv_new_from_ut32(length, value);

	*type = RZIL_OP_ARG_BITV;
	return bv;
}
```

So the op could actually be just:
```c
struct rzil_op_int_t {
	RzILBitVector bv; // no pointer
};
```

It might make sense to do #1858 first, for less memory usage.

wargio:

**Detailed description**

- Adds RZIL_OP_BITV and removes RZIL_OP_INT because it is very limited to what ìnt` is on the host arch and platform
- Fixed type RZIL_OP_JMP to work and apply effect.
- Implemented  `rz_il_bv_new_from_st32` `rz_il_bv_new_from_st64`

Closes https://github.com/rizinorg/rizin/issues/1859


wargio:
Linked to https://github.com/rizinorg/rizin/issues/1918

wargio:
fail on windows is a false positive.

XVilka (changes requested):
Looks good, could you please add a unit test for those?
- `test/unit/test_il_*`

thestr4ng3r (changes requested):
This should replace the int op entirely, otherwise it's redundant: https://github.com/rizinorg/rizin/issues/1859

wargio:
if we do this, i think i should add a method to convert signed numbers to bitv

XVilka:
Yes, please do. In fact, it was the original meaning of this operation,
just to show a representation, not to use it in the VM.

XVilka (approved):

PR #1936, opened 2021-11-03
[RzIL] Add concat, append handlers etc.

ogre2007:

**Detailed description**

I've just added concat IL opcode handler, and fixed mistake in the definition of the concat bv operation. If this is ok, I'll add another opcodes later

**Test plan**

...

**Closing issues**

linked to #1918


wargio (approved):

wargio:
looks like you forgot to change the unit test

XVilka:
Please also rebase.

librz/include/rz_il/rzil_opcodes.h
```diff
@@ -118,6 +118,16 @@ struct rzil_op_cast_t {
 	RzILOp *val; ///< value to cast
 };
 
+
+/**
+ *  \struct rzil_op_concat_t
+ *  \brief op structure for concating bitv
```
thestr4ng3r: This does not add any value over the name the op already has, please describe exactly what it does, for example from BAP documentation:
> Appends x and y so that in the resulting vector the vector x occupies the most significant part and y the least significant part.

thestr4ng3r (changes requested):
It should be called "append" to correspond to BAP: http://binaryanalysisplatform.github.io/bap/api/master/bap-core-theory/Bap_core_theory/Theory/module-type-Minimal/index.html#val-append

librz/include/rz_il/rzil_opcodes.h
XVilka: Yes, append operates on two bitvectors, while concat operates on the list of bitvectors.

wargio (changes requested):
unit test needs to be changed (https://github.com/rizinorg/rizin/blob/dev/test/unit/test_il_definitions.c) and also you need to add `void *rz_il_handler_concat(RzILVM *vm, RzILOp *op, RzILOpArgType *type);` definition in vm_layer.c

ogre2007:
I've renamed concat into append, and added new concat operation, but there is not RzList ARG type in opcodes, so its not correct currently. Should I add new ARG_TYPE?

librz/include/rz_il/rzil_vm.h
```diff
@@ -100,6 +100,54 @@ RZ_API RzILEffect *rz_il_evaluate_effect(RzILVM *vm, RzILOp *op, RzILOpArgType *
...
+void *rz_il_handler_jmp(RzILVM *vm, RzILOp *op, RzILOpArgType *type);
+void *rz_il_handler_goto(RzILVM *vm, RzILOp *op, RzILOpArgType *type);
+void *rz_il_handler_seq(RzILVM *vm, RzILOp *op, RzILOpArgType *type);
+void *rz_il_handler_branch(RzILVM *vm, RzILOp *op, RzILOpArgType *type);
+
+void *rz_il_handler_load(RzILVM *vm, RzILOp *op, RzILOpArgType *type);
+void *rz_il_handler_store(RzILVM *vm, RzILOp *op, RzILOpArgType *type);
+
+//TODO: remove me when all the handlers are implemented
+void *rz_il_handler_unimplemented(RzILVM *vm, RzILOp *op, RzILOpArgType *type);
```
wargio: i have removed them for a reason. just add the concat/append at the beginning of vm_layer.c

librz/il/definitions/bitvector.c
```diff
@@ -251,24 +251,48 @@ RZ_API RZ_OWN RzILBitVector *rz_il_bv_cut_tail(RZ_NONNULL RzILBitVector *bv, ut3
...
+			tmp = ret;
+			continue;
+		}
+		ret = rz_il_bv_append(ret, bv);
+		rz_il_bv_free(tmp);
+		tmp = ret;
+	}
+	tmp = NULL;
+	return ret;
+}
```
wargio: i'm not sure about this. @XVilka

wargio (changes requested):

librz/include/rz_il/rzil_vm.h
ogre2007: Sure! But the question is how to provide RzList as argument for concat inside the operation object, bcs there is no argtype for it

librz/il/definitions/bitvector.c
ogre2007: It actually works though. I can do it w/o append function, but it takes two foreach loops: 1st for collecting actual lengths, then we can allocate resulting bv, then we copy bvs content in the second foreach loop.

librz/include/rz_il/rzil_vm.h
wargio: i think you need to edit the RzILOpConcat structure to have an RzList, but i would wait for @XVilka for his input. i'm not sure if that is the right way to implement this.

librz/il/definitions/bitvector.c
wargio: looks correct, but the problem is if this is actually needed, or not. so this is why i asked to wait for him

XVilka:
On the one hand, such an operation could be emulated by calling multiple `append` operations. On the other hand, it might become handy for uplifting more complex instructions. I guess we can add a new `list` type for the arguments. @thestr4ng3r your thoughts?

thestr4ng3r:
From what I can tell, `RzILOpArgType` is only needed when an op **returns** a list, but not if it takes one. Or are there ops in BAP that can return lists of bitvectors? That would seem overcomplicated to me.
Regardless, I would just omit the concat completely for now. `RzList` is a linked list anyway and chaining multiple append ops would actually be one too.

librz/il/definitions/bitvector.c
```diff
@@ -251,12 +251,12 @@ RZ_API RZ_OWN RzILBitVector *rz_il_bv_cut_tail(RZ_NONNULL RzILBitVector *bv, ut3
 }
 
 /**
- * Concat bv1 and bv2 to get new bitvector
```
XVilka, line 264: Append then

test/unit/test_il_definitions.c
```diff
@@ -179,6 +179,7 @@ bool test_rzil_bv_algorithm(void) {
 	RzILBitVector *x, *y;
 	RzILBitVector *result;
 	RzILBitVector *add, *sub, *mul, *div, *mod;
+
```
XVilka: Unnecessary change

XVilka (changes requested):

XVilka:
@ogre2007 please rebase and address my comments, we can merge it then.

ogre2007:
ok, Ill do it today

XVilka (approved):

librz/il/vm_layer.c
```diff
@@ -177,8 +177,8 @@ RZ_API bool rz_il_vm_init(RzILVM *vm, ut64 start_addr, ut32 addr_size, ut32 data
 	vm->op_handler_table[RZIL_OP_SLE] = &rz_il_handler_unimplemented; // &rz_il_handler_sle;
 	vm->op_handler_table[RZIL_OP_ULE] = &rz_il_handler_unimplemented; // &rz_il_handler_ule;
 	vm->op_handler_table[RZIL_OP_CAST] = &rz_il_handler_cast;
-	vm->op_handler_table[RZIL_OP_CONCAT] = &rz_il_handler_unimplemented; // &rz_il_handler_concat;
-	vm->op_handler_table[RZIL_OP_APPEND] = &rz_il_handler_unimplemented; // &rz_il_handler_append;
+	vm->op_handler_table[RZIL_OP_CONCAT] = &rz_il_handler_unimplemented; // &rz_il_handler_concat
+	vm->op_handler_table[RZIL_OP_APPEND] = &rz_il_handler_append; // &rz_il_handler_append;
```
wargio:
```suggestion
	vm->op_handler_table[RZIL_OP_APPEND] = &rz_il_handler_append;
```

librz/include/rz_il/rzil_vm.h
```diff
@@ -100,6 +100,52 @@ RZ_API RzILEffect *rz_il_evaluate_effect(RzILVM *vm, RzILOp *op, RzILOpArgType *
...
+
```
wargio:
```suggestion
```

wargio: please add the definition (`void *rz_il_handler_append(RzILVM *vm, RzILOp *op, RzILOpArgType *type);`) in vm_layer.c

This header mush **NOT** have any of these definitions there.

wargio (changes requested):

wargio (approved):

PR #1910, opened 2021-10-31
Add asan build for clang

kazarmy:
**Your checklist for this pull request**
- [ ] I've read the [guidelines for contributing](https://github.com/rizinorg/rizin/blob/master/DEVELOPERS.md) to this repository
- [ ] I made sure to follow the project's [coding style](https://github.com/rizinorg/rizin/blob/master/DEVELOPERS.md#code-style)
- [ ] I've added tests that prove my fix is effective or that my feature works (if possible)
- [ ] I've updated the documentation and the [rizin book](https://github.com/rizinorg/book) with the relevant information (if needed)

**Detailed description**

The reasoning for this pr goes like this:

1. The non-optimized OpenBSD builds seem to be far more stable than the optimized builds.
2. OpenBSD uses Clang.
3. Clang might treat some UB code differently from gcc.
4. Ergo, we should have an optimized asan build for Clang as well.

Fortunately, the default Clang version for GitHubCI's `ubuntu-20.04` is the same as for OpenBSD i.e. 11.

**Test plan**

All builds are green. The `linux-clang-tests-asan` build works as expected.

**Closing issues**

...

---

- The removal of the `pkgcfg_sanitize_libs` lines was done using

```sh
REGEX='pkgcfg_sanitize_libs,$' bash -c 'git grep -l "$REGEX" | xargs -n 1 sed -i "/$REGEX/d"'
```

- The addition of the `library_linkflags` lines was done using

```sh
REGEX="^\([[:blank:]]*\)\(soversion: rizin_libversion\)" bash -c 'git grep -l "$REGEX" | xargs -n 1 sed -i "s/$REGEX/\1link_args: library_linkflags,\n\1\2/"'
```

with some manual minor editing done afterwards.

- The addition of the `binrz` `executable_cflags` lines was done using

```sh
REGEX="^\([[:blank:]]*\)\(include_directories: \[platform_inc\],\)" bash -c 'git grep -l "$REGEX" | grep "binrz/\(rizin/\)\|\(rz-.*/\)" | xargs -n 1 sed -i "s/$REGEX/\1c_args: executable_cflags,\n\1\2/"'
```

- The addition of the `binrz` `executable_linkflags` lines was done using

```sh
[...]

PR #1941, opened 2021-11-05
rz_pvector_foreach: Check for zero vector length to prevent UB

kazarmy:

**Detailed description**

This pr cherry-picks the fix for the following error from #1910:

cbin-UB
(https://github.com/rizinorg/rizin/runs/4101766412?check_suite_focus=true#step:16:314)

**Test plan**

All builds are green.

**Closing issues**

...


ret2libc (commented):
Why is `it` NULL ? Is it because the pvector is empty?

kazarmy:
Might be an `RzPVector` straight from an `rz_pvector_new()` but didn't check. Does it matter?

kazarmy:
btw, the pr checks whether `it` is null, not whether `*it` is null

kazarmy:
> Might be an RzPVector straight from an rz_pvector_new() but didn't check. Does it matter?

The point I was trying to make here is that

```c
RzPVector *a_vector = rz_pvector_new(NULL);
void **itr;
rz_pvector_foreach (a_vector, itr) {
        // ...
}
```

i.e. iterating over an empty `RzPVector` ... that shouldn't invoke UB, so it shouldn't be an issue as to whether there is a problem with the vector that triggered the UBSan runtime error -- the patch has to happen anyway.

kazarmy:
Ok I'm going to follow https://patchwork.kernel.org/project/git/patch/20200125053834.GB744673@coredump.intra.peff.net/ and check whether the vector length is 0. Need to confirm whether this works first though.

pelijah:
I don't understand. In case of `v.len = 0` a loop body won't be executed since you get `for (it = (vec)->v.a; it != (vec)->v.a; it++)`. Why this extra check is needed?

ret2libc (approved):

PR #1948, opened 2021-11-08
RzShell: convert ou command

ret2libc:


librz/core/cmd/cmd_open.c
```diff
@@ -1588,3 +1567,25 @@ RZ_IPI RzCmdStatus rz_open_close_all_handler(RzCore *core, int argc, const char
...
+			if (bf) {
+				rz_core_bin_raise(core, bf->id);
+				rz_core_block_read(core);
+				RZ_LOG_INFO("Switched to fd %d (%s)\n", fdnum, bf->file);
+			}
+			return RZ_CMD_STATUS_OK;
+		}
+	}
+	RZ_LOG_ERROR("Could not find any opened file with fd %d\n", fdnum);
+	return RZ_CMD_STATUS_ERROR;
```
wargio: this is just a comment, but this code does not really work. we need to fix this in another commit.

wargio (approved):

librz/core/cmd/cmd_open.c
ret2libc: what do you mean? What does not work?

wargio: try it. just does not really work. print a function, change fd and then try again with another function

ret2libc: ah if you mean analysis I didn't test it. It should work though at a more low level.

XVilka (approved):

PR #1975, opened 2021-11-13
Rewrite and fix AVR DES ESIL

thestr4ng3r:

**Detailed description**

Pulled out of #1920.
The primary motivation here is to get rid of the `text` and `deskey` register hacks. But the op was implemented incorrectly anyway (actually it was a nop), so I rewrote it from scratch and against the "official" AVR Simulator's behavior because the instruction set manual is useless.

The test `EXPECTED` parts are unfortunately a bit ugly because `ar=8` is broken here. In the mentioned pr this is fixed.

**Test plan**

see the added tests


wargio (approved):

librz/analysis/p/analysis_avr.c
```diff
@@ -599,11 +597,10 @@ INST_HANDLER(dec) { // DEC Rd
 }
 
 INST_HANDLER(des) { // DES k
-	if (desctx.round < 16) { //DES
-		op->type = RZ_ANALYSIS_OP_TYPE_CRYPTO;
-		op->cycles = 1; //redo this
-		rz_strbuf_setf(&op->esil, "%d,des", desctx.round);
-	}
+	op->type = RZ_ANALYSIS_OP_TYPE_CRYPTO;
+	op->cycles = 1; //redo this
```
XVilka: Add `TODO` or `FIXME` please, it will make easier to see/find.

XVilka (approved):

librz/analysis/p/analysis_avr.c
thestr4ng3r: Actually I don't know why this comment was there in the first place. `des` takes 1 cycle, or 2 cycles if following a non-des op, and since we can only insert a single value here, 1 would be ok.
I'll just remove the comment once travis is through.

PR #1996, opened 2021-11-19
Add support for DYLD_CHAINED_PTR_64_OFFSET

thestr4ng3r:

**Detailed description**

These kinds of chained fixups have started to show up in built binaries around Monterey.

**Test plan**

see the added test


librz/bin/format/mach0/mach0_defines.h
```diff
@@ -1519,6 +1520,22 @@ struct dyld_chained_ptr_arm64e_auth_bind {
...
+		next : 12,
+		bind : 1; // == 0
+};
+
+struct dyld_chained_ptr_64_bind {
+	uint64_t ordinal : 24,
+		addend : 8,
+		reserved : 19,
+		next : 12,
+		bind : 1; // == 1
```
wargio, line 1536: probably these needs to be parsed with endianness in mind.

thestr4ng3r: Judging from my recent experiments with endianness and bitfields on tree-sitter there is actually a chance this might work as-is, but I am not sure yet. I would still leave them as-is right now and convert later, along with all the other structs, because these are the official definitions from dyld code.

wargio: ok

thestr4ng3r: Ok, just checked and it does not work on big endian. But I would still like to fix it later.

ret2libc (approved):

PR #2012, opened 2021-11-23
Cleanup `dbl` command handler; Remove `rz_bp_list`

DMaroo:

**Detailed description**

More cleaner implementation of `dbl`, and it uses `RzCmdStateOutput` correctly. `rz_bp_list` was marked to be deprecated and has now been removed.

**Test plan**

None. The integration tests already test `dbl`.

**Closing issues**

None.


librz/core/cmd/cmd_debug.c
```diff
@@ -4410,23 +4410,67 @@ RZ_IPI RzCmdStatus rz_cmd_debug_add_bp_handler(RzCore *core, int argc, const cha
...
+				b->hw ? "hw" : "sw",
+				b->trace ? "trace" : "break",
+				b->enabled ? "enabled" : "disabled",
+				rz_bp_is_valid(core->dbg->bp, b) ? "valid" : "invalid",
+				rz_str_get(b->data),
+				rz_str_get(b->cond),
+				rz_str_get(b->name),
+				rz_str_get(b->module_name));
+			break;
+		}
```
wargio: @ret2libc @XVilka maybe using the table api is better?

librz/core/cmd/cmd_debug.c
```diff
@@ -4410,23 +4410,67 @@ RZ_IPI RzCmdStatus rz_cmd_debug_add_bp_handler(RzCore *core, int argc, const cha
...
+	rz_list_foreach (core->dbg->bp->bps, iter, b) {
+		switch (mode) {
+		case RZ_OUTPUT_MODE_STANDARD: {
+			rz_cons_printf("0x%08" PFMT64x " - 0x%08" PFMT64x
+						 " %d %c%c%c %s %s %s %s cmd=\"%s\" cond=\"%s\" "
+						 "name=\"%s\" module=\"%s\"\n",
+				b->addr, b->addr + b->size, b->size,
+				((b->perm & RZ_BP_PROT_READ) | (b->perm & RZ_BP_PROT_ACCESS)) ? 'r' : '-',
+				((b->perm & RZ_BP_PROT_WRITE) | (b->perm & RZ_BP_PROT_ACCESS)) ? 'w' : '-',
+				(b->perm & RZ_BP_PROT_EXEC) ? 'x' : '-',
```
wargio: i would suggest to create a static function where you give the flags and a char buffer and you set the flags accordingly.
the char buffer can be declared next to the PJ declaration as `char perms[10];`

librz/core/cmd/cmd_debug.c
```diff
@@ -4410,23 +4410,67 @@ RZ_IPI RzCmdStatus rz_cmd_debug_add_bp_handler(RzCore *core, int argc, const cha
...
+			} else {
+				core->dbg->bp->cb_printf("db @ 0x%08" PFMT64x "\n", b->addr);
+			}
+			break;
+		}
+		case RZ_OUTPUT_MODE_JSON: {
+			pj_o(pj);
+			pj_kN(pj, "addr", b->addr);
+			pj_ki(pj, "size", b->size);
+			pj_ks(pj, "perm", rz_str_rwx_i(b->perm & 7)); /* filter out R_BP_PROT_ACCESS */
```
wargio: why here you use the rz_str_rwx_i but not above? are we also sure that the enums are mapped in the same way? also why filtering out the access flag? this looks wrong

wargio (changes requested):

librz/core/cmd/cmd_debug.c
DMaroo: This was the way the old command was implemented. I will change it.

wargio: yes, please.

DMaroo: I have changed it. Removed the deprecated `RZ_BP_PROT` enum.

librz/core/cmd/cmd_debug.c
```diff
@@ -4410,23 +4406,65 @@ RZ_IPI RzCmdStatus rz_cmd_debug_add_bp_handler(RzCore *core, int argc, const cha
...
+				rz_bp_is_valid(core->dbg->bp, b) ? "valid" : "invalid",
+				rz_str_get(b->data),
+				rz_str_get(b->cond),
+				rz_str_get(b->name),
+				rz_str_get(b->module_name));
+			break;
+		}
+		case RZ_OUTPUT_MODE_RIZIN: {
+			if (b->module_name) {
+				core->dbg->bp->cb_printf("dbm %s %" PFMT64d "\n", b->module_name, b->module_delta);
```
ret2libc: rz_cons_printf here and in other places in this function

librz/core/cmd/cmd_debug.c
```diff
@@ -4410,23 +4406,65 @@ RZ_IPI RzCmdStatus rz_cmd_debug_add_bp_handler(RzCore *core, int argc, const cha
...
-		rz_bp_list(core->dbg->bp, -1);
-		break;
-	default:
-		rz_warn_if_reached();
-		break;
+	RzBreakpointItem *b;
+	RzListIter *iter;
+	RzOutputMode mode = RZ_OUTPUT_MODE_STANDARD;
+	PJ *pj = NULL;
+	if (state) {
```
ret2libc: remove this check and just use a `rz_return_val_if_fail(state, RZ_CMD_STATUS_ERROR)` at the beginning of the function.

librz/core/cmd/cmd_debug.c
```diff
@@ -4410,23 +4406,65 @@ RZ_IPI RzCmdStatus rz_cmd_debug_add_bp_handler(RzCore *core, int argc, const cha
...
+		mode = state->mode;
+		pj = state->d.pj;
+		rz_cmd_state_output_array_start(state);
+	}
+	rz_list_foreach (core->dbg->bp->bps, iter, b) {
+		switch (mode) {
```
ret2libc:
```suggestion
		switch (state->mode) {
```

librz/core/cmd/cmd_debug.c
```diff
@@ -4410,23 +4406,65 @@ RZ_IPI RzCmdStatus rz_cmd_debug_add_bp_handler(RzCore *core, int argc, const cha
...
+			core->dbg->bp->cb_printf("0x%08" PFMT64x "\n", b->addr);
+			break;
+		}
+		default:
+			rz_warn_if_reached();
+		}
+	}
+	if (state) {
+		rz_cmd_state_output_array_end(state);
 	}
```
ret2libc:
```suggestion
	rz_cmd_state_output_array_end(state);
```

librz/core/cmd/cmd_debug.c
```diff
@@ -4410,23 +4406,65 @@ RZ_IPI RzCmdStatus rz_cmd_debug_add_bp_handler(RzCore *core, int argc, const cha
...
+			pj_ks(pj, "cond", rz_str_get(b->cond));
+			pj_end(pj);
+			break;
+		}
+		case RZ_OUTPUT_MODE_QUIET: {
+			core->dbg->bp->cb_printf("0x%08" PFMT64x "\n", b->addr);
+			break;
+		}
+		default:
+			rz_warn_if_reached();
```
ret2libc, line 4466:
```suggestion
			rz_warn_if_reached();
			break;
```

librz/core/cmd/cmd_debug.c
```diff
@@ -4410,23 +4406,65 @@ RZ_IPI RzCmdStatus rz_cmd_debug_add_bp_handler(RzCore *core, int argc, const cha
...
+			pj_kb(pj, "valid", rz_bp_is_valid(core->dbg->bp, b));
+			pj_ks(pj, "data", rz_str_get(b->data));
+			pj_ks(pj, "cond", rz_str_get(b->cond));
+			pj_end(pj);
+			break;
+		}
+		case RZ_OUTPUT_MODE_QUIET: {
+			core->dbg->bp->cb_printf("0x%08" PFMT64x "\n", b->addr);
+			break;
+		}
```
ret2libc:
```suggestion
		case RZ_OUTPUT_MODE_QUIET:
			rz_cons_printf("0x%08" PFMT64x "\n", b->addr);
			break;
```

librz/core/cmd/cmd_debug.c
```diff
@@ -4410,23 +4406,65 @@ RZ_IPI RzCmdStatus rz_cmd_debug_add_bp_handler(RzCore *core, int argc, const cha
...
+				b->trace ? "trace" : "break",
+				b->enabled ? "enabled" : "disabled",
+				rz_bp_is_valid(core->dbg->bp, b) ? "valid" : "invalid",
+				rz_str_get(b->data),
+				rz_str_get(b->cond),
+				rz_str_get(b->name),
+				rz_str_get(b->module_name));
+			break;
+		}
+		case RZ_OUTPUT_MODE_RIZIN: {
```
ret2libc:
```suggestion
		case RZ_OUTPUT_MODE_RIZIN:
```

librz/core/cmd/cmd_debug.c
```diff
@@ -4410,23 +4406,65 @@ RZ_IPI RzCmdStatus rz_cmd_debug_add_bp_handler(RzCore *core, int argc, const cha
...
+			if (b->module_name) {
+				core->dbg->bp->cb_printf("dbm %s %" PFMT64d "\n", b->module_name, b->module_delta);
+			} else {
+				core->dbg->bp->cb_printf("db @ 0x%08" PFMT64x "\n", b->addr);
+			}
+			break;
+		}
```
ret2libc:
```suggestion
```

librz/core/cmd/cmd_debug.c
DMaroo: I have used `rz_str_rwx` instead.

librz/core/cmd/cmd_debug.c
ret2libc: I would love to have it, yes, but for the sake of this single PR i'm ok also as is.

DMaroo: I will add tables. Looks nice to me.

librz/core/cmd/cmd_debug.c
```diff
@@ -4410,24 +4406,61 @@ RZ_IPI RzCmdStatus rz_cmd_debug_add_bp_handler(RzCore *core, int argc, const cha
...
-		break;
+	rz_return_val_if_fail(state, RZ_CMD_STATUS_ERROR);
+	RzBreakpointItem *b;
+	RzListIter *iter;
+	PJ *pj = NULL;
+	pj = state->d.pj;
+	rz_cmd_state_output_array_start(state);
+	rz_list_foreach (core->dbg->bp->bps, iter, b) {
+		switch (state->mode) {
+		case RZ_OUTPUT_MODE_STANDARD: {
```
ret2libc:
```suggestion
		case RZ_OUTPUT_MODE_STANDARD:
```

librz/core/cmd/cmd_debug.c
```diff
@@ -4410,24 +4406,61 @@ RZ_IPI RzCmdStatus rz_cmd_debug_add_bp_handler(RzCore *core, int argc, const cha
...
+			break;
+		}
+		case RZ_OUTPUT_MODE_RIZIN:
+			if (b->module_name) {
+				rz_cons_printf("dbm %s %" PFMT64d "\n", b->module_name, b->module_delta);
+			} else {
+				rz_cons_printf("db @ 0x%08" PFMT64x "\n", b->addr);
+			}
+			break;
+		case RZ_OUTPUT_MODE_JSON: {
```
ret2libc:
```suggestion
		case RZ_OUTPUT_MODE_JSON:
```

librz/core/cmd/cmd_debug.c
```diff
@@ -4410,24 +4406,61 @@ RZ_IPI RzCmdStatus rz_cmd_debug_add_bp_handler(RzCore *core, int argc, const cha
...
+			pj_kb(pj, "hw", b->hw);
+			pj_kb(pj, "trace", b->trace);
+			pj_kb(pj, "enabled", b->enabled);
+			pj_kb(pj, "valid", rz_bp_is_valid(core->dbg->bp, b));
+			pj_ks(pj, "data", rz_str_get(b->data));
+			pj_ks(pj, "cond", rz_str_get(b->cond));
+			pj_end(pj);
+			break;
+		}
+		case RZ_OUTPUT_MODE_QUIET: {
```
ret2libc:
```suggestion
		case RZ_OUTPUT_MODE_QUIET:
```

librz/core/cmd/cmd_debug.c
```diff
@@ -4410,24 +4406,58 @@ RZ_IPI RzCmdStatus rz_cmd_debug_add_bp_handler(RzCore *core, int argc, const cha
...
+				rz_cons_printf("dbm %s %" PFMT64d "\n", b->module_name, b->module_delta);
+			} else {
+				rz_cons_printf("db @ 0x%08" PFMT64x "\n", b->addr);
+			}
+			break;
+		case RZ_OUTPUT_MODE_JSON:
+			pj_o(pj);
+			pj_kN(pj, "addr", b->addr);
+			pj_ki(pj, "size", b->size);
+			pj_ks(pj, "perm", rz_str_rwx_i(b->perm & 7)); // filter out only rwx bits
```
wargio:
```suggestion
			pj_ks(pj, "perm", rz_str_rwx_i(b->perm));
```

wargio (changes requested):

librz/core/cmd/cmd_debug.c
DMaroo: I don't really see how the higher bits (`RZ_PERM_(SHAR|PRIV|ACCESS|CREAT)`) make sense in the context of breakpoints. Shouldn't we just discard them?

wargio: i do not think you ever set them, so they cannot be set. if that is an issue, just `b->perm & 7` every time you call `rz_str_rwx_i`

DMaroo: Yeah, I filter once while setting the perms. No need to filter while printing again. Agreed.

librz/core/cmd/cmd_debug.c
```diff
@@ -2884,15 +2884,11 @@ static void add_breakpoint(RzCore *core, ut64 addr, const char *arg_perm, bool h
...
 	if (watch) {
-		if (!strcmp(arg_perm, "r")) {
-			rw = RZ_BP_PROT_READ;
-		} else if (!strcmp(arg_perm, "w")) {
-			rw = RZ_BP_PROT_WRITE;
-		} else if (!strcmp(arg_perm, "rw")) {
-			rw = RZ_BP_PROT_ACCESS;
-		} else {
+		rw = rz_str_rwx(arg_perm);
+		rw &= 7; // filter out the rwx bits only
```
XVilka: I think there was named mask for that. If not - please add.

librz/core/cmd/cmd_debug.c
```diff
@@ -4410,24 +4406,58 @@ RZ_IPI RzCmdStatus rz_cmd_debug_add_bp_handler(RzCore *core, int argc, const cha
...
-	case RZ_OUTPUT_MODE_JSON:
-		rz_bp_list(core->dbg->bp, 'j');
-		break;
-	case RZ_OUTPUT_MODE_QUIET:
-		rz_bp_list(core->dbg->bp, -1);
-		break;
-	default:
-		rz_warn_if_reached();
-		break;
+	rz_return_val_if_fail(state, RZ_CMD_STATUS_ERROR);
```
XVilka: I recommend to check here also the `core->dbg` and `core->dbg->bp`.

XVilka (changes requested):
Overall is good, just a few nitpicks.

XVilka (approved):
From the code point of view looks OK. Need some manual testing though, just to be sure.

librz/core/cmd/cmd_debug.c
ret2libc: This was not resolved yet, was it? @DMaroo are you going to do it in this PR?

DMaroo: Yes, I am.

DMaroo: I tried implementing it using `RzTable`, but I now do not feel it is worth it. Anyways, we won't be able to use `RzTable` for `RZ_OUTPUT_MODE_JSON` since that would require us the JSON object (which the current `RzTable` does not provide). Therefore, the only use case for the table is `RZ_OUTPUT_MODE_STANDARD`, which already uses a single `rz_cons_printf` line. Using tables would require me to create a `RzList` of all the attributes and then create a row from it and then insert it to the table. I don't think the extra complexity is necessary here. Sorry for delaying, I don't think we need tables. This is resolved from my end.

ret2libc:
@DMaroo conflicts need to be solved.

yossizap (approved):
LGTM

librz/core/cmd/cmd_debug.c
ret2libc: I'm ok to add it later, but the table does not require you to have rzlist.
You can just use rz_table_add_rowf, e.g. `rz_table_add_rowf(state->d.t, "sssss", str, plugin->license, plugin->name, plugin->uris, plugin->desc);`. In general, it allows you more control on the output and it provides the headers, which are always nice to have as a user immediately understands what the output means.

wargio (approved):
@DMaroo i have not seen the table use. do you plan to have it in this PR or to another one? regardless, good job.

librz/core/cmd/cmd_debug.c
DMaroo: That would be great. I'll just add it.

DMaroo:
Yes. Please don't merge. I will add the `RzTable`.

DMaroo:
I have added tables. Please review and merge once green.

librz/core/cmd/cmd_debug.c
```diff
@@ -4411,24 +4407,64 @@ RZ_IPI RzCmdStatus rz_cmd_debug_add_bp_handler(RzCore *core, int argc, const cha
...
+				b->trace ? "trace" : "break",
+				b->enabled ? "enabled" : "disabled",
+				rz_bp_is_valid(core->dbg->bp, b) ? "valid" : "invalid",
+				rz_str_get(b->data),
+				rz_str_get(b->cond),
+				rz_str_get(b->name),
+				rz_str_get(b->module_name));
+			break;
+		case RZ_OUTPUT_MODE_TABLE:
+			rz_table_set_columnsf(state->d.t, "XXdsssssssss", "start addr", "end addr", "size", "perm", "hw/sw", "type", "state", "validity", "cmd", "cond", "name", "module");
```
ret2libc: move this outside the `rz_list_foreach`.

librz/core/cmd/cmd_debug.c
```diff
@@ -4411,24 +4407,64 @@ RZ_IPI RzCmdStatus rz_cmd_debug_add_bp_handler(RzCore *core, int argc, const cha
...
+			t->showFancy = true;
```
ret2libc: I don't think this is needed either. We should probably have some sort of global setting or something.

DMaroo: Do we have such a global setting?

DMaroo: TBF, the fancy output looks much more prettier than the default output.

ret2libc: I do agree :D

librz/core/cmd/cmd_debug.c
```diff
@@ -4411,24 +4407,63 @@ RZ_IPI RzCmdStatus rz_cmd_debug_add_bp_handler(RzCore *core, int argc, const cha
...
+				b->trace ? "trace" : "break",
+				b->enabled ? "enabled" : "disabled",
+				rz_bp_is_valid(core->dbg->bp, b) ? "valid" : "invalid",
+				rz_str_get(b->data),
+				rz_str_get(b->cond),
+				rz_str_get(b->name),
+				rz_str_get(b->module_name));
+			break;
+		case RZ_OUTPUT_MODE_TABLE:
+			rz_table_add_rowf(t, "XXdsssssssss", b->addr, b->addr + b->size, b->size, rz_str_rwx_i(b->perm), b->hw ? "hw" : "sw", b->trace ? "trace" : "break", b->enabled ? "enabled" : "disabled", rz_bp_is_valid(core->dbg->bp, b) ? "valid" : "invalid", rz_str_get(b->data), rz_str_get(b->cond), rz_str_get(b->name), rz_str_get(b->module_name));
```
ret2libc: oops :D please split this. it's a huge line

librz/core/cmd_descs/cmd_debug.yaml
```diff
@@ -16,6 +16,7 @@ commands:
         type: RZ_CMD_DESC_TYPE_ARGV_STATE
         modes:
            - RZ_OUTPUT_MODE_STANDARD
+           - RZ_OUTPUT_MODE_TABLE
```
```suggestion
        default_mode: RZ_OUTPUT_MODE_TABLE
        modes:
           - RZ_OUTPUT_MODE_TABLE
```

FYI, this is also an option if we want. It makes the default mode (the one you get when just doing `dbl`) the table mode.

ret2libc (approved):
Just the long line comment, but LGTM overall.

wargio (approved):
great job

librz/core/cmd_descs/cmd_debug.yaml
DMaroo: Oh, didn't know that. Thanks. But I don't think we should remove the standard mode output. What if the user wants to use standard mode output?

ret2libc: When you use default_mode there is no STANDARD mode anymore. The standard mode becomes the TABLE. One could argue what is the additional benefit of a "STANDARD" mode vs the table mode, which has the same info, with headers, with a syntax to filter columns/rows/etc.

DMaroo: Oh, yes. I was talking about the plain standard mode only. I thought that is more concise and has less fluff, but I also agree that it is not especially useful compared to the other modes. I'll do this.

librz/core/cmd/cmd_debug.c
DMaroo: I have opened a PR for this #2027.

DMaroo:
Please review the integration tests.

PR #2027, opened 2021-11-25
Allow users to decide how their tables look

DMaroo:

**Detailed description**

Currently, there is no way for the user to configure how their table will be printed/shown. It would be nice to allow the user to decide the printing style and whether to print headers or not.

Need help in figuring out how to use the config within the `rz_table_new` function. Maybe I could use a `RzCoreBind`. Would like to hear some suggestions.

**Test plan**

Add integration tests.

**Closing issues**

None.

### DIFF
--- a/librz/core/cconfig.c
+++ b/librz/core/cconfig.c
@@ -3257,6 +3257,10 @@ RZ_API int rz_core_config_init(RzCore *core) {
 	SETPREF("prj.file", "", "Path of the currently opened project");
 	SETBPREF("prj.compress", "false", "Compress the project file while saving");
 
+	/* table */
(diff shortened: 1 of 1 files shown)


DMaroo:
Rizin already has such a feature using the `:` option (`aflt:fancy`). Sorry for the noise.

ret2libc:
Yes, it has that. When I was talking about a way to set it for all tables I meant a var that would select a table mode by default without having to specify `:fancy`, for example.

DMaroo:
Oh, I see. Anyways, I don't think it is a priority. I'll finish the memleaks PR first.

ret2libc:
> Oh, I see. Anyways, I don't think it is a priority. I'll finish the memleaks PR first.

Absolutely not a priority. Just throwing ideas but there are many more important/fun things ;)

PR #2038, opened 2021-11-27
Refactor `rz_cons_pal_list()` to return an `RzList`

### LINKED ISSUE #1676: Convert `ec` commands to the newshell and proper API

officialcjunior:

**Detailed description**

The linked issue suggests to
```
Refactor the APIs like rz_cons_pal_list() to return the RzList instead, 
and introduce a separate function to print this RzList depending on the 
separate modes using RzOutputMode enum as well.
```
This is an early, draft PR and so far, I've just appended the `char *` (which was getting printed) to the `RzList` in each iteration. Feel free to suggest
other ideas.

Things to do:
* Currently, the PJ is printed inside `rz_cons_pal_list()` itself. Need to see if it can be moved with introducing `RzCmdStateOutput` or something else.
* Add the Doxygen doc for the API
* Looks like the inline grep `~` is broken on this PR

**Test plan**
`ec*` has a test, tests for `ecj` can also maybe be added.

**Closing issues**

Closes #1676

### DIFF
--- a/librz/cons/pal.c
+++ b/librz/cons/pal.c
@@ -521,7 +521,7 @@ RZ_API void rz_cons_pal_show(void) {
 typedef struct {
 	int val;
 	const char *str;
-} RAttrStr;
+} RzAttrStr;
 
 RZ_API RzList *rz_cons_pal_list(int rad, const char *arg) {
 	char *name = NULL, *current, **color;
@@ -571,7 +571,7 @@ RZ_API RzList *rz_cons_pal_list(int rad, const char *arg) {
 				rz_str_append(name, rz_str_newf(" rgb:%02x%02x%02x", rcolor->r2, rcolor->g2, rcolor->b2));
 			}
 			if (rcolor->attr) {
-				const RAttrStr attrs[] = {
+				const RzAttrStr attrs[] = {
 					{ RZ_CONS_ATTR_BOLD, "bold" },
 					{ RZ_CONS_ATTR_DIM, "dim" },
 					{ RZ_CONS_ATTR_ITALIC, "italic" },
--- a/librz/cons/pal.c
+++ b/librz/cons/pal.c
@@ -523,20 +523,32 @@ typedef struct {
 	const char *str;
 } RzAttrStr;
 
-RZ_API RzList *rz_cons_pal_list(int rad, const char *arg) {
-	char *name = NULL, *current, **color;
+/**
+ * \brief Returns a list of console eval colors and keys
+ *
+ * \param mode reference to RzOutputMode
+ * \param prefix reference to the optional argument for printing palettles in CSS
+ */
+RZ_API RzList RZ_OWN *rz_cons_pal_list(RzOutputMode mode, const char *prefix) {
+	char *name = NULL, *item, **color;
 	const char *hasnext;
-	PJ *pj = pj_new();
-	RzList *ret = rz_list_new();
 	int i;
-	if (rad == 'j') {
+	PJ *pj = pj_new();
+	if (!pj) {
+		return NULL;
+	}
+	RzList *ret = rz_list_new();
+	if (!ret) {
+		return NULL;
+	}
+	if (mode == RZ_OUTPUT_MODE_JSON) {
 		pj_o(pj);
 	}
 	for (i = 0; keys[i].name; i++) {
 		RzColor *rcolor = RZCOLOR_AT(i);
 		color = COLOR_AT(i);
-		switch (rad) {
-		case 'j':
+		switch (mode) {
+		case RZ_OUTPUT_MODE_JSON:
 			pj_k(pj, keys[i].name);
 			pj_a(pj);
 			pj_n(pj, rcolor->r);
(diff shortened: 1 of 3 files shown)


librz/cons/pal.c
```diff
@@ -521,28 +521,32 @@ RZ_API void rz_cons_pal_show(void) {
 typedef struct {
 	int val;
 	const char *str;
-} RAttrStr;
+} RzAttrStr;
 
-RZ_API void rz_cons_pal_list(int rad, const char *arg) {
-	char *name, **color;
+RZ_API RzList *rz_cons_pal_list(int rad, const char *arg) {
```
XVilka: Add ownership mark please. Also swithc `rad` to the `RzOutputMode`

librz/cons/pal.c
```diff
@@ -521,28 +521,32 @@ RZ_API void rz_cons_pal_show(void) {
...
+	char *name = NULL, *current, **color;
 	const char *hasnext;
+	PJ *pj = pj_new();
+	RzList *ret = rz_list_new();
```
XVilka, line 540: Check `pj` and `ret` for NULL.

librz/include/rz_cons.h
```diff
@@ -950,7 +950,7 @@ RZ_API const char *rz_cons_pal_get_name(int index);
 RZ_API int rz_cons_pal_len(void);
 RZ_API int rz_cons_rgb_parse(const char *p, ut8 *r, ut8 *g, ut8 *b, ut8 *a);
 RZ_API char *rz_cons_rgb_tostring(ut8 r, ut8 g, ut8 b);
-RZ_API void rz_cons_pal_list(int rad, const char *arg);
+RZ_API RzList *rz_cons_pal_list(int rad, const char *arg);
```
XVilka: Ownership.

XVilka (changes requested):

librz/cons/pal.c
officialcjunior: > Also swithc rad to the RzOutputMode

Yeah, about that. 
There was this weird edge case inside that, a `case 'c'` which handled `ecc`. `ecc` is a command which prints the shows the palette in CSS. I just put that under `RZ_OUTPUT_MODE_LONG`. See if that sounds alright.

XVilka: I think it's better to make this a separate command. @ret2libc what do you think?

XVilka (changes requested):
I am not sure if this "list" is a right solution though. This output mode-dependent logic should be only in the printing function. The `rz_cons_pal_list()` just should return a `RzList` of some C structure, for example `RzConsPalette`, that's it.

ret2libc (changes requested):
The idea is that the `pal_list` function should return a list of objects that are independent of the output mode. This means that still having `RzOutputMode mode` as argument of `rz_cons_pal_list` is wrong.

You should have:
```C
// rz_cons.h
RzList *rz_cons_pal_list(const char *prefix);

// rz_core.h
void rz_core_cons_pal_print(RzList *colors, RzCmdStateOutput *state);
```

Please think about what would be the type of the element inside the `RzList`.

b1llow:
Is this PR still going on?

officialcjunior:
> Is this PR still going on?

I've had long COVID and is still recovering, so there's a bit of a drop in the motivation.

Feel free to steal it if want to :)

officialcjunior:
> I am not sure if this "list" is a right solution though. This output mode-dependent logic should be only in the printing function. The `rz_cons_pal_list()` just should return a `RzList` of some C structure, for example `RzConsPalette`, that's it.

Thanks for the review.

Yeah, you're right. and I was looking into what sort of C structure the `RzList` should contain. On a quick look, it looked
like an `RzList` of `RzColor`s would do, but there's two more things that I think we'll need to carry with us and take care of: the name of the variable (`keys.name`, global and static) and a scary variable `color` (`char **`) which holds the actual color of the key when `ec` is printed (as you can see [here](https://github.com/rizinorg/rizin/blob/dev/librz/cons/pal.c#L597)). I was wondering about how to handle that.

XVilka:
@officialcjunior any chance to finish this?

officialcjunior:
Yes, I was looking at this again and was still stuck at figuring out what sort of C structure to make an RzList out of.

> The rz_cons_pal_list() just should return a RzList of some C structure, for example RzConsPalette, that's it.

> Yeah, you're right. and I was looking into what sort of C structure the RzList should contain. On a quick look, it looked
> like an RzList of RzColors would do, but there's two more things that I think we'll need to carry with us and take > care of: the name of the variable (keys.name, global and static) and a scary variable color (char **) which holds > the actual color of the key when ec is printed (as you can see [here](https://github.com/rizinorg/rizin/blob/dev > /librz/cons/pal.c#L597)). I was wondering about how to handle that.


I think we can make this happen with just `keys.name` and an `RzColor`. So we need to have a structure which has a `char *` and an `RzColor` together.

I don't know how colors generally work, and especially how it works here. Let me try to learn more and break this into first principles and see if I can find another solution.

XVilka:
Closing this due to the lack of activity. If you plan to continue working on this - please open a new one.

PR #2045, opened 2021-11-29
Remove binrz/preload

thestr4ng3r:
This is not used at all.


ret2libc:
Could it be used by rz-run preload ?

ret2libc:
```
librz/io/p/io_self.c:   // !!! may die if not running from r2preload !!! //
```

wargio (approved):

XVilka (approved):
We could remove `self://` as well, @ret2libc  ;)

thestr4ng3r:
> Could it be used by rz-run preload ?

No, rz-run preload is for setting things like `LD_PRELOAD` on a child.

> We could remove `self://` as well, @ret2libc  ;)

Actually I would move that to an external plugin. It's completely useless but fun.

ret2libc (approved):

PR #2068, opened 2021-12-04
Add rz_bv_set_from_bytes_le()

thestr4ng3r:

**Detailed description**

This can be used to set a bit vector from raw bytes interpreted as little endian.
I played around with big endian too, but there are multiple ways to implement it and the one I would need to replicate the register arena behavior would be very weird. So this is left for later.

Also make rz_bv_as_hex_string() always pad with the same number of 0s if len % 4 != 0. This means e.g. that a zero bv of size 7 shows as 0x00 instead of 0x0.

**Test plan**

see the tests

### DIFF
--- a/librz/util/bitvector.c
+++ b/librz/util/bitvector.c
@@ -5,6 +5,9 @@
 #include <stdlib.h>
 #include <stdio.h>
 
+#define NELEM(N, ELEMPER) ((N + (ELEMPER)-1) / (ELEMPER))
+#define BV_ELEM_SIZE      8U
+
 /**
  * \brief Initialize a RzBitVector structure
  * \param bv Pointer to a uninitialized RzBitVector instance
@@ -98,7 +101,7 @@ RZ_API RZ_OWN char *rz_bv_as_hex_string(RZ_NONNULL RzBitVector *bv) {
 
 	if (bv->len <= 64) {
 		char format[32] = { 0 };
-		rz_strf(format, "0x%%0%d" PFMT64x, bv->len / 4);
+		rz_strf(format, "0x%%0%d" PFMT64x, (bv->len + 3) / 4);
 		return rz_str_newf(format, bv->bits.small_u);
 	}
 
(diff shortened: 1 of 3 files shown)


librz/util/bitvector.c
```diff
@@ -1174,6 +1189,34 @@ RZ_API bool rz_bv_set_from_st64(RZ_NONNULL RzBitVector *bv, st64 value) {
...
 }
 
+/**
+ * Set the bitvector's contents from the given bits. The bitvector's size is unchanged.
+ * If bv->len < size, additional bits are cut off, if bv->len > size, the rest is filled up with 0.
+ * \param buf little endian buffer of at least (bit_offset + size + 7) / 8 bytes
+ * \param bit_offset offset inside buf to start reading from, in bits
+ * \param size number of bits to read from buf
+ */
+RZ_API void rz_bv_set_from_le(RZ_NONNULL RzBitVector *bv, RZ_IN const ut8 *buf, ut32 bit_offset, ut32 size) {
```
XVilka: Check for `buf` to be not NULL

librz/util/bitvector.c
```diff
@@ -1133,6 +1136,18 @@ RZ_API RZ_OWN RzBitVector *rz_bv_new_from_st64(ut32 length, st64 value) {
 	return bv;
 }
 
+/**
+ * Create a new bitvector of size bits and apply rz_bv_set_from_le() to it
+ */
+RZ_API RZ_OWN RzBitVector *rz_bv_new_from_le(RZ_IN const ut8 *buf, ut32 bit_offset, ut32 size) {
```
XVilka: Check for `buf` to be not NULL

XVilka (approved):
Looks ok, just please add these NULL checks

librz/util/bitvector.c
```diff
@@ -1174,6 +1189,34 @@ RZ_API bool rz_bv_set_from_st64(RZ_NONNULL RzBitVector *bv, st64 value) {
...
+			val |= (ut64)buf[i] << (i * 8);
+		}
+		val &= (UT64_MAX >> (64 - size));
+		rz_bv_set_from_ut64(bv, val);
+		return;
+	}
+	for (ut32 i = 0; i < bv->len; i++) {
+		bool bit = false;
+		if (i < size) {
+			bit = !!(buf[(bit_offset + i) / 8] & (1 << (bit_offset + i) % 8));
```
wargio:
```suggestion
			bit = (buf[(bit_offset + i) >>> 3] & (1 << (bit_offset + i) & 7));
```

wargio:
tbh, i'm not sure if the name makes much sense. i would call this more `rz_bv_set_from_big_num`

thestr4ng3r:
> tbh, i'm not sure if the name makes much sense. i would call this more `rz_bv_set_from_big_num`

It is meant to be consistent with `rz_read_le*` and similar functions, only that this one reads into a bitvector. It also leaves open a big endian version.

wargio:
> > tbh, i'm not sure if the name makes much sense. i would call this more `rz_bv_set_from_big_num`
> 
> It is meant to be consistent with `rz_read_le*` and similar functions, only that this one reads into a bitvector. It also leaves open a big endian version.

what about `rz_bv_set_from_bytes_le` ?

thestr4ng3r:
That sounds good to me.

wargio (approved):

wargio:
i have restarted travis

PR #2073, opened 2021-12-06
Adapt basic block analysis for the Hexagon architecture.

### LINKED ISSUE #36: Allow multiple `endloop` packets per loop instruction
Multiple `endloop` packets can belong to a single `loop` instruction.

Example:
loop

Currently the jump address of each endloop packet after the first one are set to `0x0`. Therefore the analysis has errors.

The fix should be fairly simple. Just not set the static variables of loop starts to `0x0` once an endloop instruction is disasssembled.
See cases in `hexagon_disas.c::hexagon_disasm_instruction()`.

Prerequisite: https://github.com/rizinorg/rizin/pull/2073 should be merged.

### LINKED ISSUE #1796: `af` interprets conditional jumps as end of function
This is the same issue as https://github.com/rizinorg/rz-hexagon/issues/30.

The function analysis of `af` simply interprets conditional jump instructions as the end of a function.

Example:

rz-example

The instruction `if (!P1) jumpr:nt R31` has the `RzAnalysisOp.type = RZ_ANALYSIS_OP_TYPE_RCJMP` set.

If instructions of type `RZ_ANALYSIS_OP_TYPE_RCJMP` are analyzed the block analysis function will simply `goto beach`:

```c
static int run_basic_block_analysis(RzAnalysisTaskItem *item, RzVector *tasks) {
    [...]

    case RZ_ANALYSIS_OP_TYPE_CJMP:
    case RZ_ANALYSIS_OP_TYPE_MCJMP:
    case RZ_ANALYSIS_OP_TYPE_RCJMP:
    case RZ_ANALYSIS_OP_TYPE_UCJMP:
	    if (analysis->opt.cjmpref) {
		    (void)rz_analysis_xrefs_set(analysis, op.addr, op.jump, RZ_ANALYSIS_REF_TYPE_CODE);
	    }
    [...]
    // XXX breaks mips analysis too !op.delay
    // this will be all x86, arm (at least)
    // without which the analysis is really slow,
    // presumably because each opcode would get revisited
    // (and already covered by a bb) many times
    goto beach;
    // For some reason, branch delayed code (MIPS) needs to continue
    break;
```

If the `goto beach` statement is removed all kinds of tests fail. What do you think is the right way to address this issue?

[Referenced code](https://github.com/rizinorg/rizin/blob/d6ac6b93ef999d906b20dd9d6116d220bf8df5c8/librz/analysis/fcn.c#L1107)

### Work environment

| OS/arch/bits (mandatory)                             | Ubuntu x86 64
| Architecture/bits of the file (mandatory)            | Hexagon (from [this PR](https://github.com/rizinorg/rizin/pull/1614))
| `rizin -v` full output, **not truncated** (mandatory)  |rizin 0.3.0-git @ linux-x86-64 commit: d878f0babdd0425d0896ab806a84839b94406586, build: 2021-10-01

**(Although this is not the most recent state of `dev` the code in question has not changed!)**
### Expected behavior

The function should continue after the conditional jump (if `analysis.jump.after = true`).

### Actual behavior

After a conditional jump the function is not recognized as such.

```bash

$ > rizin -a hexagon -e asm.pcalign=4 =

[0x00000000] > wx 11400275114001f20e4021f1e0c282752943028c00417f53e34b82750ec22ef102478e8500c001940242636b027f427446ea005c3260
[0x00000000] > af
[0x00000000] > pif
/ P1 = !cmp.eq(R2,##0x0)
| P1 = !cmp.eq(R1,R0)
| R14 = or(R1,R0)
\ P0 = cmp.gtu(R2,##0x17)
/ R9 = lsr(R2,#0x3)
| if (!P1) jumpr:nt R31

# Should be longer than that.
```

### LINKED ISSUE #2047: `run_basic_block_analysis` ignores `analysis.jmp.after` flag.
`run_basic_block_analysis` ([src](https://github.com/rizinorg/rizin/blob/d6ac6b93ef999d906b20dd9d6116d220bf8df5c8/librz/analysis/fcn.c#L536)) almost always ignores the `analysis.jmp.after` option and treats `jumps` as return instructions. This behavior results in a low analysis quality for binaries with little to no symbol ELF information (think of firmware binaries).

A specific problem caused by this behavior is described in issue https://github.com/rizinorg/rizin/issues/1796.

### Work environment

| OS/arch/bits (mandatory)                             | Ubuntu x86 64
| Architecture/bits of the file (mandatory)            | x86/64.
| `rizin -v` full output, **not truncated** (mandatory)  | rizin 0.4.0-git @ linux-x86-64 commit: 6942f6c9083f5f2347f697b95d93ae09f10b0a88, build: 2021-11-29__14:18:00

### Expected behavior

If `analysis.jmp.after = true` (which is the default case) this is the expected behavior:
If a `jump` instruction is encountered by the basic block analysis, it should continue the analysis with the next instruction (after setting xrefs of cause).

### Actual behavior

It treats the `jump` instruction as return. This in turn results in a really flawed function detection.

Same as in https://github.com/rizinorg/rizin/issues/1796

```bash

$ > rizin -a hexagon -e asm.pcalign=4 =

[0x00000000] > wx 11400275114001f20e4021f1e0c282752943028c00417f53e34b82750ec22ef102478e8500c001940242636b027f427446ea005c3260
[0x00000000] > af
[0x00000000] > pif
/ P1 = !cmp.eq(R2,##0x0)
| P1 = !cmp.eq(R1,R0)
| R14 = or(R1,R0)
\ P0 = cmp.gtu(R2,##0x17)
/ R9 = lsr(R2,#0x3)
| if (!P1) jumpr:nt R31

# Should be longer than that.
```

### Additional context

In [this branch](https://github.com/Rot127/rizin/tree/af-jmp-is-ret) I added multiple checks which check for `analysis.jmp.after`. Those changes let to a firmware binary of mine to be analyzed way better than before.

On the other hand **a lot** of tests broke afterwards.
One of the tests were the `afx` test for the Hexagon architecture. But it could be simply fixed by executing `aaa` instead of `af` since this adds the jumps to the xref list for some reason.
Maybe all the other analysis test could be fixed like this as well?

Rot127:

**Detailed description**

The `run_basic_block_analysis` function does not continue analysis if it encounters an register jump and `analysis.jmp.after=true`. This is not correct, since `analysis.jmp.after=true` explicitly states that the analysis should continue after a jump instruction.
 
Now to the big problem.

All additions I made to the code are only executed if the analysis runs on the Hexagon architecture (see usage of `is_hexagon` flag). Because if it changes its analysis behavior for all architectures **a lot** (>60) tests fail.

**Test plan**

Tests were added.
~~Only the `analysis.ijmp missing branch bug` failed. Could someone, who is more familiar than me with the x86_64 arch, please take a look at this one? I can't really tell whether it is good that it failed or not.~~

**Closing issues**

kinda closes https://github.com/rizinorg/rizin/issues/2047 (Since for all other architectures than Hexagon the problem persists.)
closes https://github.com/rizinorg/rizin/issues/1796
closes https://github.com/rizinorg/rz-hexagon/issues/36

**Tiny additional feature added**
Added `analysis.trap.after=false` config. Set to `true`, it will continue the analysis after a trap instruction.


Rot127:
Just a random idea which came to my mind.
It would be possible to allow the user to fine grade when the analysis should be continue after jumps.
Something like that:
```c
analysis.jmp.after.outofmap     // Continue analysis if jump leaves mapped memory region.
analysis.jmp.after.reg          // Continue analysis after a register jump.
analysis.jmp.after.cond         // Continue analysis after a conditional jump.
analysis.jmp.after.outfcn       // Continue analysis if jump leads out of function.
```
It would allow the user to configure the analysis style best fitting to the platform being analysed.
Do you like the idea?

librz/analysis/fcn.c
```diff
@@ -1067,6 +1101,9 @@ static int run_basic_block_analysis(RzAnalysisTaskItem *item, RzVector *tasks) {
 				bb->fail = UT64_MAX;
 			}
 			rz_analysis_task_item_new(analysis, tasks, fcn, NULL, op.jump);
+			if (continue_after_jump && is_hexagon) {
```
XVilka, line 1107: Now you have `continue_after_jump`, is `is_hexagon` really necessary? I think it should not break other architectures then.

XVilka:
Regarding:
```
analysis.jmp.after.outofmap     // Continue analysis if jump leaves mapped memory region.
analysis.jmp.after.reg          // Continue analysis after a register jump.
analysis.jmp.after.cond         // Continue analysis after a conditional jump.
analysis.jmp.after.outfcn       // Continue analysis if jump leads out of function.
```
The idea is good, but this should be thoroughly tested with various combinations for main architectures.

librz/analysis/fcn.c
Rot127: If the `is_hexagon` flag is removed 15 tests fail. Some fail because the jump instruction should have been treated (correctly) as return statement and now it isn't. But for others I am not sure.

<details>
<summary> For example </summary>

```diff
[XX] /home/user/rizin/test/db/analysis/avr avr ATTiny RzArchProfile display
RZ_NOPLUGINS=1 rizin -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -e asm.cpu=ATTiny48 -Qc 'aa
pd 5 @ 0x0
fs~reg
' bins/firmware/arduino_avr.bin
-- stdout
--- expected
+++ actual
@@ -1,11 +1,15 @@
-        ,=< 0x00000000  ~   jmp   entry0
-        |   ;-- PINB:
-        |   0x00000003      unaligned
-        |   ;-- DDRB:
-       ,==< 0x00000004  ~   jmp   0x1b6
-       ||   ;-- PORTB:
-       ||   0x00000005      unaligned
-       ||   ;-- PINC:
-       ||   0x00000006  ~   invalid
+|           ;-- (0x00000003) PINB:
+|       ,=< 0x00000000  ~   jmp   entry0
+|       |   ;-- DDRB:
+|       |   ;-- (0x00000005) PORTB:
+|      ,==< 0x00000004  ~   jmp   0x1b6
+|      ||   ;-- PORTC:
+|      ||   ;-- (0x00000009) PIND:
+|     ,===< 0x00000008  ~   jmp   0x1b6
+|     |||   ;-- PINA:
+|     |||   ;-- (0x0000000d) DDRA:
+|    ,====< 0x0000000c  ~   jmp   0x1b6
+|    ||||   ;-- (0x00000012) PORTCR:
+|   ,=====< 0x00000010  ~   jmp   0x1b6
    37 * registers.extended
    41 * registers.mmio

[x] Analyze all flags starting with sym. and entry0 (aa)

[XX] /home/user/rizin/test/db/analysis/avr test adding fake CPUs and loading new profiles
RZ_NOPLUGINS=1 rizin -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -Qc 'e asm.cpu=ATTiny4_fake_cpu
aa
pd 5 @ 0x03c
' bins/firmware/arduino_avr.bin
-- stdout
--- expected
+++ actual
@@ -1,8 +1,6 @@
-        ,=< 0x0000003c  ~   jmp   0x1b6
-        |   ;-- SPL:
-        |   0x0000003d      unaligned
-        |   ;-- SPH:
-        |   0x0000003e  ~   invalid
-        |   ;-- SREG:
-        |   0x0000003f      unaligned
-       ,==< 0x00000040      jmp   0x3f8
+|           ;-- (0x0000003d) SPL:
+|       ,=< 0x0000003c  ~   jmp   0x1b6
+|      ,==< 0x00000040      jmp   0x3f8
+|     ,===< 0x00000044      jmp   0x1b6
+|    ,====< 0x00000048      jmp   0x143a
+|   ,=====< 0x0000004c      jmp   0x154e

[XX] /home/user/rizin/test/db/analysis/x86_32 hang-on-aa/-aac fix (#15004)
RZ_NOPLUGINS=1 rizin -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -Qc 'e asm.bytes=true
e analysis.vars.stackname=true
aa
e asm.var=false
pdb
' bins/pe/15004.file
-- stdout
--- expected
+++ actual
@@ -1,4 +1,2 @@
-/ (loc) entry0 (int32_t arg_40h, int32_t arg_44h);
+/ (loc) entry0 (int32_t arg_44h);
 |           0x006453a1      e8d8170000     call  fcn.00646b7e
-|           0x006453a6      8d642444       lea   esp, [arg_44h]
-|       ,=< 0x006453aa      0f85a6300000   jne   0x648456

[XX] /home/user/rizin/test/db/analysis/x86_64 noreturn of reloc-functions
RZ_NOPLUGINS=1 rizin -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -Qc 'aaa
afbr 0x80b0
' bins/elf/ls
-- stdout
--- expected
+++ actual
@@ --1,4 +-1,3 @@
 0x000081cb
-0x00008237

[XX] /home/user/rizin/test/db/analysis/mach0 aav find xrefs
RZ_NOPLUGINS=1 rizin -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -Qc 'e analysis.loads=true
e analysis.in=range
e analysis.from=0x8000
e analysis.to=0x8000+4K
e asm.bytes=true
aac
afna
aar
aan
aav
# aa;afna;aac;aar;aan;aav
pd 17 @ 0x8058
' bins/mach0/macho-bash
-- stdout
--- expected
+++ actual
@@ --1,34 +-1,34 @@
 |       `=< 0x00008058      03e4           b     0x7862
-            0x0000805a      0000           movs  r0, r0
-            ; DATA XREF from sub.free_73a0 @ 0x7d30
-            0x0000805c      .dword 0x000693da
-            ; DATA XREF from sub.free_73a0 @ 0x7d3e
-            0x00008060      .dword 0x0006c1f4
-            ; DATA XREF from sub.free_73a0 @ 0x7d48
-            0x00008064      .dword 0x0006c1a6
-            ; DATA XREF from sub.free_73a0 @ 0x7e30
-            0x00008068      .dword 0x0006c102
-            ; DATA XREF from sub.free_73a0 @ 0x7e3a
-            0x0000806c      .dword 0x0006c0b4
-            ; DATA XREF from sub.free_73a0 @ 0x7e58
-            0x00008070      .dword 0x0006c0da
-            ; DATA XREF from sub.free_73a0 @ 0x7e62
-            0x00008074      .dword 0x0006c08c
-            ; DATA XREF from sub.strcpy_7890 @ 0x7e80
-            0x00008078      .dword 0x0006c0b2
-            ; DATA XREF from sub.strcpy_7890 @ 0x7e8a
-            0x0000807c      .dword 0x0006c064
-            ; DATA XREF from sub.strcpy_7890 @ 0x7ed8
-            0x00008080      .dword 0x00069232
-            ; DATA XREF from sub.strcpy_7890 @ 0x7ee8
-            0x00008084      .dword 0x0006c04a
-            ; DATA XREF from sub.strcpy_7890 @ 0x7ef2
-            0x00008088      .dword 0x0006bffc
-            ; DATA XREF from sub.free_73a0 @ 0x7f64
-            0x0000808c      .dword 0x0006bfce
-            ; DATA XREF from sub.free_73a0 @ 0x7f72
-            0x00008090      .dword 0x0006bf7c
-            ; DATA XREF from sub.free_73a0 @ 0x7fcc
-            0x00008094      .dword 0x0006bf66
+|           0x0000805a      0000           movs  r0, r0
+|           ; DATA XREF from sub.free_73a0 @ 0x7d30
+|           0x0000805c      .dword 0x000693da
+|           ; DATA XREF from sub.free_73a0 @ 0x7d3e
+|           0x00008060      .dword 0x0006c1f4
+|           ; DATA XREF from sub.free_73a0 @ 0x7d48
+|           0x00008064      .dword 0x0006c1a6
+|           ; DATA XREF from sub.free_73a0 @ 0x7e30
+|           0x00008068      .dword 0x0006c102
+|           ; DATA XREF from sub.free_73a0 @ 0x7e3a
+|           0x0000806c      .dword 0x0006c0b4
+|           ; DATA XREF from sub.free_73a0 @ 0x7e58
+|           0x00008070      .dword 0x0006c0da
+|           ; DATA XREF from sub.free_73a0 @ 0x7e62
+|           0x00008074      .dword 0x0006c08c
+|           ; DATA XREF from sub.free_73a0 @ 0x7e80   <--------- free instead of strcpy
+|           0x00008078      .dword 0x0006c0b2
+|           ; DATA XREF from sub.free_73a0 @ 0x7e8a   <--------- free instead of strcpy
+|           0x0000807c      .dword 0x0006c064
+|           ; DATA XREF from sub.free_73a0 @ 0x7ed8   <--------- free instead of strcpy
+|           0x00008080      .dword 0x00069232
+|           ; DATA XREF from sub.free_73a0 @ 0x7ee8   <--------- free instead of strcpy
+|           0x00008084      .dword 0x0006c04a
+|           ; DATA XREF from sub.free_73a0 @ 0x7ef2   <--------- free instead of strcpy
+|           0x00008088      .dword 0x0006bffc
+|           ; DATA XREF from sub.free_73a0 @ 0x7f64
+|           0x0000808c      .dword 0x0006bfce
+|           ; DATA XREF from sub.free_73a0 @ 0x7f72
+|           0x00008090      .dword 0x0006bf7c
+|           ; DATA XREF from sub.free_73a0 @ 0x7fcc
+|           0x00008094      .dword 0x0006bf66

```

</details>

Rot127:
> The idea is good, but this should be thoroughly tested with various combinations for main architectures.

After spending more time to adapt `run_basic_block_analysis()` for Hexagon, I think it is better to write a function for each architecture.
`run_basic_block_analysis()` is already really complex and adding exceptions everywhere will make it only worse.
Besides, shouldn't there only one way to correctly analyze assembly of an architecture? Or are there exceptions?

librz/analysis/fcn.c
```diff
@@ -533,7 +533,19 @@ static int analyze_function_locally(RzAnalysis *analysis, RzAnalysisFunction *fc
 	return ret;
 }
 
-static int run_basic_block_analysis(RzAnalysisTaskItem *item, RzVector *tasks) {
+/**
+ * \brief Analyses the given task item \p item for branches.
+ *
+ * Analsysis starts for all instructions from \p item->start_address. If a branch is
```
XVilka: A typo

librz/analysis/fcn.c
XVilka: OK.

librz/include/rz_analysis.h
```diff
@@ -471,13 +472,13 @@ enum {
...
 
-enum {
+typedef enum {
 	RZ_ANALYSIS_RET_NOP = 0,
-	RZ_ANALYSIS_RET_ERROR = -1,
-	RZ_ANALYSIS_RET_DUP = -2,
-	RZ_ANALYSIS_RET_NEW = -3,
-	RZ_ANALYSIS_RET_END = -4
-};
+	RZ_ANALYSIS_RET_ERROR = -1, // Basic block ended because of analysis error.
```
XVilka, line 477: @ret2libc @thestr4ng3r @wargio opinions on this change? I personally have nothing against it, for the record.

librz/include/rz_analysis.h
```diff
@@ -523,6 +524,7 @@ typedef struct rz_analysis_options_t {
 	bool trycatch;
 	bool norevisit;
 	int afterjmp; // continue analysis after jmp eax or forward jmp // option
+	int aftertrap; // continue analysis after trap instructions
```
XVilka, line 527: This might be useful for many architectures, indeed.

XVilka (changes requested):

librz/include/rz_analysis.h
Rot127: Does also work for all architectures. But the config is turned of by default.

XVilka (approved):

Rot127:
Please wait with merging. It needs more testing.
Recognized a few mistakes and meanwhile am confused at which point it is right to treat an instruction as branch and when to continue analysis etc. Need to draw it out on paper again.

librz/include/rz_analysis.h
wargio: i think is ok

Rot127:
After doing some reversing with the current state of this PR I think it should be ready to merge. The result is fairly decent.
Sorry for not marking it as draft in the first place.

XVilka (approved):

thestr4ng3r (approved):

librz/analysis/fcn.c
```diff
@@ -569,6 +586,7 @@ static int run_basic_block_analysis(RzAnalysisTaskItem *item, RzVector *tasks) {
 	bool is_x86 = is_arm ? false : analysis->cur->arch && !strncmp(analysis->cur->arch, "x86", 3);
 	bool is_amd64 = is_x86 ? fcn->cc && !strcmp(fcn->cc, "amd64") : false;
 	bool is_dalvik = is_x86 ? false : analysis->cur->arch && !strncmp(analysis->cur->arch, "dalvik", 6);
+	bool is_hexagon = is_x86 ? false : analysis->cur->arch && !strncmp(analysis->cur->arch, "hexagon", 7);
```
wargio, line 589: these checks are kinda weird, i'm not sure about this, but maybe i would change this to a if-else-if-else statement.

wargio (approved):

PR #2092, opened 2021-12-10
[RzIL] refactor rz_il_op to provide rz_il_op_new_* per op

wargio:

**Detailed description**

The original code was a bit messy to use, so i have refactored the rz_il_new_op to split it for each single operation to simplify the creation of the operation.

**Test plan**

Tests have been converted, including bf code.


librz/include/rz_il/rzil_opcodes.h
```diff
@@ -275,7 +275,6 @@ struct rzil_op_bool_operation_t {
  */
 struct rzil_op_bool_inv_t {
 	RzILOp *x; ///< index of the BOOL operand
-	RzILOp *ret; ///< index of store the BOOL result
```
XVilka, line 278: Why removing this? RZIL is pure and should not change operands in place

XVilka (approved):
Apart from my question, the rest is good.

librz/include/rz_il/rzil_opcodes.h
thestr4ng3r: I think that is exactly the point of this change, the `ret` looks like an old leftover to me.

thestr4ng3r (approved):
Much nicer!

librz/include/rz_il/rzil_opcodes.h
wargio: it's a leftover. it is not used and there is no reason for being there because inverting a number requires only 1 argument.

PR #2004, opened 2021-11-20
Fix aae overwriting registers of debuggee

thestr4ng3r:
**Your checklist for this pull request**
- [x] I've read the [guidelines for contributing](https://github.com/rizinorg/rizin/blob/master/DEVELOPERS.md) to this repository
- [x] I made sure to follow the project's [coding style](https://github.com/rizinorg/rizin/blob/master/DEVELOPERS.md#code-style)
- [ ] I've documented or updated the documentation of every function and struct this PR changes. If not so I've explained why.
- [ ] I've added tests that prove my fix is effective or that my feature works (if possible)
- [ ] I've updated the [rizin book](https://github.com/rizinorg/book) with the relevant information (if needed)

**Detailed description**

Running `aaa` (in particular its `aae`) part would mess up the register contents of the debuggee.
The changes in tests are because some of the internal effects of analysis are not persisted anymore.

**Test plan**

Since this depends on debug specifically, it's not that easy to automate testing:

* Install https://git.sr.ht/~thestr4ng3r/rz-commodore
* Start `x64sc` with binary remote monitor enabled
* Open `rz -a 6502 -d vice://`
* Note the output of `dr`
* Run `aaa`
* Run `dr` again. The contents should not have changed compared to before the analysis.
* Run `dc` and observe that the debuggee does not get stuck or do any other weird stuff.

PR #2097, opened 2021-12-12
Split RzAnalysis.reg and RzDebug.reg

thestr4ng3r:
# DO NOT SQUASH THIS!!!

**Detailed description**

The core of these changes is removing this single piece of code:
https://github.com/rizinorg/rizin/blob/96a85e573e766e3870a01482f36c92df403cc4cd/librz/debug/dreg.c#L167-L170

This would make `analysis->reg` and `dbg->reg` point to the same `RzReg` which causes multiple issues, for example:
* Even though both debug and analysis plugins supply register profiles, only one could be used at a time. If the debug one was loaded (which it must be as long as debug is active) and at some point e.g. an `analysis.bits` change was triggered, the analysis one was loaded and debug registers were completely messed up because the plugin supplies data only for its own profile.
* Emulation-based analysis can leak into the debuggee. This was partially addressed by #2004 but would still be dangerous.

Unfortunately, pretty much all code was written with this behavior in place. The entire rest of this pr is for fixing these things. I went through every single access of RzDebug.reg and checked it. Hopefully nothing was missed.

**Test plan**

Try out anything that accesses registers in both debug and non-debug/emulation mode. Help on testing here is welcome, especially for debug.

# DO NOT SQUASH THIS!!!


librz/core/creg.c
```diff
@@ -0,0 +1,96 @@
...
+ * from debug or analysis (emulation).
+ * Before using this function, think twice whether it would not make more sense
+ * to use core->dbg->reg or analysis->dbg->reg directly, depending on what you want
+ * to do with it.
+ */
+RZ_API RzReg *rz_core_reg_default(RzCore *core) {
+	return rz_core_is_debug(core) ? core->dbg->reg : core->analysis->reg;
+}
+
+/// rz_reg_getv_by_role_or_name() on rz_core_reg_default()
```
ret2libc: is `///` alright?

shouldn't it be:
```C
/**
  * xxxxxx
  */
```
?

librz/core/creg.c
```diff
@@ -0,0 +1,96 @@
...
+	// this logic has to be in sync with rz_core_reg_default().
+	if (rz_core_is_debug(core)) {
+		// call this instead of rz_reg_getv_... directly because it also syncs
+		return rz_debug_reg_set(core->dbg, name, num);
+	}
+	RzRegItem *ri = rz_reg_get_by_role_or_name(core->analysis->reg, name);
+	if (!ri) {
+		return false;
+	}
+	return rz_reg_set_value(core->analysis->reg, ri, num);
```
ret2libc: shouldn't this become:
```C
reg = rz_core_reg_default(core);
RzRegItem *ri = rz_reg_get_by_role_or_name(reg, name);
return rz_reg_set_value(reg, ri, num);
```
?

ret2libc (commented):
remember to add doxygen doc to new RZ_API functions before marking this PR ready.

librz/core/creg.c
thestr4ng3r: It is ok like this because `rz_core_is_debug(core)` has already been checked. This is what the comment `this logic has to be in sync with rz_core_reg_default()` means.

ret2libc: Yes but why do you want to sync regs in one case and not in the other?

thestr4ng3r: It's only for syncing with the debuggee. In the case of analysis->regs, there is no such thing.

ret2libc: Don't you need to sync the registers with the flags anyway?

thestr4ng3r: Theoretically yes, but flags are completely separate, it's done by the caller and not here. In general I am not a fan at all of all the sync calls for both debug and flags. At least flags could be done much better with events, but that is another issue.

ret2libc: I'm just not sure about this because callers of this function would have different visible results based on debug vs non-debug mode. In debug mode they can expect flags to be synced, in non-debug mode they can't. That means that the caller needs to be aware of what this function does, which counters a bit the purpose of having a unified function IMO.

thestr4ng3r: No, flags are not synced at all in this function, regardless of the mode.

ret2libc: Oh, I thought `because it also syncs` was referring to the flags, sorry

thestr4ng3r: After looking at more code that will use this, I have changed my mind about syncing flags. It now also syncs the flags and that happens in both debug and emulation mode.

librz/core/creg.c
thestr4ng3r: Just re-checked and you are right. Seems like I have been doing this wrong for quite a while...

thestr4ng3r: To be exact: just using `///` works if there are at least two lines. With one it does not. Documented here: https://www.doxygen.nl/manual/docblocks.html#cppblock

librz/core/carg.c
```diff
@@ -5,6 +5,70 @@
...
+				return (ut64)n32;
+			}
+		}
+		const char *rn = rz_analysis_cc_arg(core->analysis, cc, num);
+		if (rn) {
+			return rz_core_reg_getv_by_role_or_name(core, rn);
+		}
+	}
+	char reg[32];
+	snprintf(reg, sizeof(reg) - 1, "A%d", num);
```
wargio: rz_strf

librz/core/carg.c
```diff
@@ -5,6 +5,70 @@
...
+	if (!RZ_STR_ISEMPTY(cc)) {
+		cc = rz_analysis_syscc_default(core->analysis);
+	}
+	const char *rn = rz_analysis_cc_arg(core->analysis, cc, num);
+	if (rn) {
+		rz_core_reg_set_by_role_or_name(core, rn, val);
+		return true;
+	}
+	char reg[32];
+	snprintf(reg, sizeof(reg) - 1, "A%d", num);
```
wargio: same here.

wargio (changes requested):

wargio (approved):

PR #2098, opened 2021-12-13
Replace non-GPL compatible code

### LINKED ISSUE #2084: Incompatible (L)GPL licenses
As of https://github.com/rizinorg/rizin/commit/1d3f029d0775ab39cfbc14f322a328e820694ea5 , this is the list of licenses we have in our codebase:

Apache-2.0, BSD-2-Clause, BSD-3-Clause, CC-BY-SA-4.0, CC0-1.0, GPL-1.0-or-later, GPL-2.0-only, GPL-2.0-or-later, GPL-3.0-or-later, LGPL-2.0-or-later, LGPL-2.1-only, LGPL-2.1-or-later, LGPL-3.0-only, MIT, MS-PL, NCSA, RSA-MD

Apache, BSD, CC, GPL, LGPL, MIT, NCSA should be compatible with GPL AFAIK.

However, MS-PL and RSA-MD seem to not be compatible.
* RSA-MD is used in `librz/hash/algorithms/md5/`
* MS-PL is used in `librz/crypto/p/crypto_aes_algo.c`

ret2libc:

**Detailed description**

MD5 code taken from freeradius project, LGPL.
AES taken from Nettle project, LGPLv3. We could also use other things from there.

In any case, these are just plugins, so we could easily switch them off, if we see the need.

**Test plan**

...

**Closing issues**

Closes #2084


wargio (approved):

librz/hash/algorithms/md5/md5.c
```diff
@@ -1,242 +1,270 @@
...
+ * To compute the message digest of a chunk of bytes, declare an
+ * MD5Context structure, pass it to MD5Init, call MD5Update as
+ * needed on buffers full of bytes, and then call MD5Final, which
+ * will fill a supplied 16-byte array with the digest.
+ */
 
-	/* Zeroize sensitive information. check if compiler optimizes out this */
-	rz_mem_memzero((void *)x, sizeof(x));
-}
+/*#include <sys/param.h>*/
```
XVilka: Just remove.

librz/hash/algorithms/md5/md5.c
```diff
@@ -1,242 +1,270 @@
...
+ * Final wrapup - pad to 64-byte boundary with the bit pattern
+ * 1 0* (64-bit count of bits processed, MSB-first)
+ */
+void MD5Final(unsigned char digest[MD5_DIGEST_LENGTH], MD5_CTX *ctx) {
+	ut8 count[8];
+	size_t padlen;
+	int i;
+
+	/* Convert count to 8 bytes in little endian order. */
+	PUT_64BIT_LE(count, ctx->count);
```
XVilka: Use function from `rz_util_*()` maybe?

librz/hash/algorithms/md5/md5.c
```diff
@@ -1,242 +1,270 @@
...
+	if (padlen < 1 + 8)
+		padlen += MD5_BLOCK_LENGTH;
+	MD5Update(ctx, PADDING, padlen - 8); /* padlen - 8 <= 64 */
+	MD5Update(ctx, count, 8);
 
-	/* Save number of bits */
-	Encode(bits, context->count, 8);
+	if (digest != NULL) {
+		for (i = 0; i < 4; i++)
+			PUT_32BIT_LE(digest + i * 4, ctx->state[i]);
```
XVilka: Use function from `rz_util_*`

librz/hash/algorithms/md5/md5.c
```diff
@@ -1,242 +1,270 @@
...
+/*
+ * The core of the MD5 algorithm, this alters an existing MD5 hash to
+ * reflect the addition of 16 longwords of new data.  MD5Update blocks
+ * the data and converts bytes into longwords for this routine.
+ */
+void MD5Transform(ut32 state[4], const ut8 block[MD5_BLOCK_LENGTH]) {
+	ut32 a, b, c, d, in[MD5_BLOCK_LENGTH / 4];
+
+	for (a = 0; a < MD5_BLOCK_LENGTH / 4; a++) {
+		in[a] = (ut32)((ut32)(block[a * 4 + 0]) |
```
XVilka: Use function from `rz_util`?

librz/hash/algorithms/md5/md5.h
```diff
@@ -9,14 +9,76 @@
 #define RZ_HASH_MD5_DIGEST_SIZE  0x10
 #define RZ_HASH_MD5_BLOCK_LENGTH 0x40
 
-typedef struct {
-	ut32 state[4];
-	ut32 count[2];
-	ut8 buffer[RZ_HASH_MD5_BLOCK_LENGTH];
+/*
+ * md5.h        Structures and prototypes for md5.
```
XVilka: This comment could be removed, unnecessary with SPDX

librz/hash/algorithms/md5/md5.h
```diff
@@ -9,14 +9,76 @@
...
+ *
+ * Version:     $Id$
+ * License:		LGPL, but largely derived from a public domain source.
+ *
+ */
+
+#ifdef HAVE_INTTYPES_H
```
XVilka: Are those really unnecessary? Just remove those.

librz/hash/algorithms/md5/md5.h
```diff
@@ -9,14 +9,76 @@
...
 
-void MD5_Init(MD5_CTX *context);
-void MD5_Update(MD5_CTX *context, const ut8 *data, unsigned int length);
-void MD5_Final(ut8 *digest, MD5_CTX *context);
+/* include <sys/cdefs.h> */
+
+/* __BEGIN_DECLS */
+void MD5Init(MD5_CTX *);
+void MD5Update(MD5_CTX *, const ut8 *, size_t)
+	/*		__attribute__((__bounded__(__string__,2,3)))*/;
```
XVilka: Remove unused attributes

XVilka (changes requested):
Is the nettle subproject was added by mistake? Why not use the external library with a custom meson file as we have for other dependencies?

ret2libc:
@XVilka considering that the md5 code was taken from somewhere else, I didn't modified it too much and I prefer to keep it as simple as possible, without rz_util-ifying it too much or doing anything fancy. I have removed the unused #include and some header for licenses, which is already covered by SPDX, but for the rest of the code I kept it as is for now.

With regards to nettle, it was added on purpose to provide the AES code. I didn't use a wrap referencing the nettle project because it includes many other files that we don't need (nettle supports a lot of encryption stuff). Since we just need the few files for AES i just copied-pasted them, but this is still ok and it is done also for e.g. xxhash. The reason I'm ok with it even if it's copy-pasted code is that 1) it is very clear where the code comes from and from which version 2) it is in a subproject, with its separate meson file 3) AES is just needed for a crypto plugin, so in theory at some point it could even be easily compiled out.

I used nettle because it is a nice project and it provides simple implementations for common crypto algorithms that are LGPL as well.

librz/crypto/p/crypto_aes.c
```diff
@@ -104,8 +160,8 @@ static bool aes_ecb_fini(RzCrypto *cry) {
 
 RzCryptoPlugin rz_crypto_plugin_aes = {
 	.name = "aes-ecb",
-	.author = "Karl Malbrain",
-	.license = "MS-PL",
+	.author = "Nettle project (algorithm implementation), pancake (plugin)",
```
wargio, line 163:
```suggestion
	.author = "Nettle project",
```

ret2libc: I'd prefer to keep it this way for now... unless I rewrite even all the other functions.

XVilka:
To be honest, I think, it's totally OK to modify the MD5 code as we see fit. Nobody updates the algorithm or implementations, since it is an effectively deprecated algorithm and is useful in Rizin due to the RE specifics. Thus, porting the code to use Rizin-specific API, e.g. RzUtil, is the way to go, in my opinion.

wargio:
i agree. it's LGPL at the end.

ret2libc:
I have slightly changed the md5 code to use some rz-util function, but *please review that file very carefully now*, especially with regards to the rz_util functions I used.

librz/hash/algorithms/md5/md5.c
```diff
@@ -1,242 +1,252 @@
...
-static void Encode(ut8 *output, ut32 *input, ut32 len) {
-	ut32 i, j;
-	for (i = 0, j = 0; j < len; i++, j += 4) {
-		output[j] = (ut8)(input[i] & 0xff);
-		output[j + 1] = (ut8)((input[i] >> 8) & 0xff);
-		output[j + 2] = (ut8)((input[i] >> 16) & 0xff);
-		output[j + 3] = (ut8)((input[i] >> 24) & 0xff);
-	}
-}
+void librad_md5_calc(unsigned char *output, unsigned char *input,
```
XVilka: Rename the `librad` prefix as well, please.

librz/hash/algorithms/md5/md5.c
```diff
@@ -1,242 +1,240 @@
...
+	rz_MD5Update(ctx, count, 8);
 
-	/* Zeroize sensitive information.  */
-	rz_mem_memzero((void *)context, sizeof(*context));
+	if (digest != NULL) {
+		for (i = 0; i < 4; i++) {
+			rz_write_le32(digest + i * 4, ctx->state[i]);
+		}
+	}
+	bzero(ctx, sizeof(*ctx)); /* in case it's sensitive */
```
wargio: i'm not 100% sure that bzero is portable. (https://stackoverflow.com/questions/17096990/why-use-bzero-over-memset/17097088) so i would suggest to change this to memset

librz/hash/p/algo_md5.c
```diff
@@ -77,8 +77,8 @@ static bool plugin_md5_small_block(const ut8 *data, ut64 size, ut8 **digest, RzM
 
 RzMsgDigestPlugin rz_msg_digest_plugin_md5 = {
 	.name = "md5",
-	.license = "RSA-MD",
-	.author = "RSA Data Security, Inc.",
+	.license = "LGPL2",
+	.author = "Alan DeKok (md5 algorithm implementation), deroad (plugin)",
```
wargio, line 81: damn, i should have added `.author_lib` and `.author_plugin` to simplify this text in the first place. maybe i will refactor this once we merge it :D

test/db/tools/rz_hash
```diff
@@ -241,7 +241,7 @@ ____h_ xxhash32       LGPL3      deroad
 ____h_ parity         LGPL3      deroad
 ____h_ entropy        LGPL3      deroad
 ____h_ entropy_fract  LGPL3      deroad
-ED____ aes-ecb        MS-PL      Karl Malbrain
+ED____ aes-ecb        LGPL3      Nettle project (algorithm implementation), pancake (plugin)
 ED____ aes-cbc        LGPL-3     rakholiyajenish.07
 __ed__ base64         LGPL-3     rakholiyajenish.07
 __ed__ base91         LGPL-3     rakholiyajenish.07
```
wargio, line 247: maybe this can be changed from `LGPL-3` to `LGPL3` but can be done later on.

wargio (changes requested):

librz/hash/algorithms/md5/md5.c
ret2libc: Good catch, let's use `rz_mem_memzero`

wargio (approved):

XVilka (approved):

PR #2099, opened 2021-12-13
Check kind of DWARF attribute before demangling

### LINKED ISSUE #2089: Access to dwarf attribute value without kind checking leads to segfault
### Work environment

| OS/arch/bits (mandatory)                             | Ubuntu 20.04
| File format of the file you reverse (mandatory)      | ELF64
| Architecture/bits of the file (mandatory)            | amd64
| `rizin -v` full output, **not truncated** (mandatory)  | rizin 0.4.0-git @ linux-x86-64 commit: 5b6aaed68d84ac13de49b9400bd3bc7fe44eacbc, build: 2021-12-10__11:04:46

### Expected behavior

Rizin shouldn't try to access invalid memory locations when trying to demangle a function name.

### Actual behavior

On some elf64 binaries with debug information, rizin will read a pointer straight from the binary and store it as a char pointer, allowing the binary to control that pointer and most likely lead to a segfault (or some other kind of undesirable behavior).

Analyze the binary attached below with `aaa` or `-A`.
[binary.zip](https://github.com/rizinorg/rizin/files/7693386/binary.zip)

[asciicast](https://asciinema.org/a/oV2p7hr9UJx8fhlRuo5mFMGky)

The cause seems to be that when parsing the binary, an attribute with format `DW_FORM_ref_sup8` gets read, and it's value gets stored at the field named `reference` (at `dwarf.c:1693`) of the attribute struct. Later, when parsing a function with that attribute as it's name, that same attribute value gets read as char pointer representing the name of the function (because the `parse_function` function attempts to read the `string.content` field of the attribute struct, and both variables share the same address since they are stored in a union type). This happens at `dwarf_process.c:1643`, and since there is no kind check the pointer that was read from the binary gets stored 'as is' within the function structure, leading to a crash when trying to demangle it.

This seems to be part of a larger issue, though, since in many places within the codebase the same pattern emerges (someone tries to access an attribute without checking the kind first, which could lead to similar issues). However, after reviewing the code a bit, we didn't find any places where this could have more severe implications.

DMaroo:

**Detailed description**

This prevents the crash encountered in #2089.

**Test plan**

All tests should pass. I think that we should add the binary ([binary.zip](https://github.com/rizinorg/rizin/files/7693386/binary.zip)) in our test suite. Please let me know if I should do that.

**Closing issues**

Closes #2089.

### DIFF
--- a/librz/analysis/dwarf_process.c
+++ b/librz/analysis/dwarf_process.c
@@ -1413,17 +1413,17 @@ static void parse_function(Context *ctx, ut64 idx) {
 		switch (die->attr_values[i].attr_name) {
 		case DW_AT_name:
 			if (!get_linkage_name || !has_linkage_name) {
-				fcn.name = val->string.content;
+				fcn.name = val->kind == DW_AT_KIND_STRING ? val->string.content : fcn.name;
 			}
 			break;
 		case DW_AT_linkage_name:
 		case DW_AT_MIPS_linkage_name:
-			fcn.name = val->string.content;
+			fcn.name = val->kind == DW_AT_KIND_STRING ? val->string.content : fcn.name;
 			has_linkage_name = true;
 			break;
 		case DW_AT_low_pc:
 		case DW_AT_entry_pc:
-			fcn.addr = val->address;
+			fcn.addr = val->kind == DW_AT_KIND_ADDRESS ? val->address : fcn.addr;
 			break;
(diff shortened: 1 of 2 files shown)


XVilka (approved):

wargio (approved):

DMaroo:
Should I add this binary in the rizinorg/rizin-testbins?

XVilka:
@DMaroo please do

PR #2105, opened 2021-12-14
[RzIL] Add loadw/storew and bind memories against RzBuffer

wargio:

**Detailed description**

1. Implements memory reads/writes via RzBuffer/RzIO
2. Adds RzBitVector features like `rz_bv_set_from_bytes_be`, `rz_bv_set_to_bytes_le` and `rz_bv_set_to_bytes_be`

**Test plan**

Added unit tests where possible.


librz/core/cmd/cmd_analysis.c
```diff
@@ -9640,7 +9640,35 @@ RZ_IPI RzCmdStatus rz_rzil_vm_step_with_events_handler(RzCore *core, int argc, c
 	return RZ_CMD_STATUS_OK;
 }
 
+RZ_IPI RzCmdStatus rz_rzil_vm_step_till_addr_handler(RzCore *core, int argc, const char **argv) {
```
thestr4ng3r: I find "until" better than "till" but that is subjective and both would be correct.

librz/include/rz_il/rzil_opcodes.h
```diff
@@ -307,9 +307,8 @@ struct rzil_op_load_t {
  *  store m k x a memory m in which the key k is associated with the word x.
  */
 struct rzil_op_store_t {
-	int mem; ///< index of memory in VM
```
thestr4ng3r, line 299: Don't remove the mem index, it's an important feature of the IL. Multiple memories are absolutely possible, even if our RzIO currently has no good way to represent it.

librz/include/rz_il/rzil_opcodes.h
```diff
@@ -296,8 +296,8 @@ struct rzil_op_bool_inv_t {
  *  load m k is the value associated with the key k in the memory m.
  */
 struct rzil_op_load_t {
-	int mem; ///< index of the memory in VM (different from the temp_val_list)
-	RzILOp *key; ///< index of the RzBitVector key (address)
+	RzILOp *key; ///< memory index of the RzBitVector key (address)
+	ut32 len; ///< n of bits to read
 };
```
thestr4ng3r: What you have implemented here is not `load` and `store` (which load single "bytes"), but `loadw` and `storew` which load entire words of a given length.
See
http://binaryanalysisplatform.github.io/bap/api/master/bap-core-theory/Bap_core_theory/Theory/module-type-Basic/index.html#val-load
http://binaryanalysisplatform.github.io/bap/api/master/bap-core-theory/Bap_core_theory/Theory/module-type-Basic/index.html#val-loadw

We should do this in the same way. Essentially that would mean leaving the original implementation of load/store and adding what you have written as loadw/storew.

librz/il/vm_layer.c
```diff
@@ -128,16 +128,21 @@ static void free_bind_var_val(HtPPKv *kv) {
...
-	vm->addr_size = addr_size;
-	vm->data_size = data_size;
+RZ_API bool rz_il_vm_init(RzILVM *vm, ut64 start_addr, ut32 addr_size, RzBuffer *ro_memory, bool big_endian) {
+	rz_return_val_if_fail(vm && ro_memory, false);
 
 	rz_pvector_init(&vm->vm_global_variable_list, (RzPVectorFree)rz_il_variable_free);
 	rz_pvector_init(&vm->vm_local_variable_list, (RzPVectorFree)rz_il_variable_free);
-	rz_pvector_init(&vm->vm_memory, (RzPVectorFree)rz_il_mem_free);
+
+	vm->vm_memory = rz_buf_new_sparse_overlay(ro_memory, RZ_BUF_SPARSE_WRITE_MODE_SPARSE);
```
thestr4ng3r: In the long run, using a sparse overlay buffer should not be hardcoded here since there are multiple use-cases of for example also just letting the vm persist everything in io directly. What is desired can be controlled by just supplying a buffer in the right form into this function.
But this is easy to refactor, so no blocker imo.

thestr4ng3r (dismissed):

wargio:
Ok, i'll revert back and keep the two approches and then allow also the user to set two different type of memory based on the op type (i.e. `load` `store` or `loadw` and `storew`)

thestr4ng3r:
What exactly do you mean by different types of memory? Both load/store and loadw/storew can access the same memories. The only difference is that load/store access exactly one byte and loadw/storew can access multiple ones depending on the requested size and endianness.

librz/core/cil.c
```diff
@@ -422,6 +422,62 @@ RZ_IPI void rz_core_analysis_rzil_reinit(RzCore *core) {
 	}
 }
 
+RZ_IPI void rz_core_analysis_rzil_vm_set(RzCore *core, const char *var_name, ut64 value) {
```
XVilka: Maybe a brief description what this function for

librz/core/cmd/cmd_analysis.c
XVilka: Agree with `until`

librz/il/vm_layer.c
XVilka: Agree on this point. We can open an issue to track that.

librz/include/rz_il/rzil_opcodes.h
XVilka: Yes, this precisely.

librz/util/bitvector.c
```diff
@@ -1150,6 +1169,19 @@ RZ_API RZ_OWN RzBitVector *rz_bv_new_from_bytes_le(RZ_IN RZ_NONNULL const ut8 *b
 	return bv;
 }
 
+/**
+ * Create a new bitvector of size bits and apply rz_bv_set_from_bytes_be() to it
```
XVilka: You can use `\p size` syntax to reference the parameter inline.

XVilka (changes requested):
Overall good job! Please address the comments, especially the `load`/`loadw` one and would be good to merge 💯

librz/il/rzil_export.c
```diff
@@ -329,15 +329,15 @@ static void il_opdmp_append(RzILOp *op, RzStrBuf *sb, PJ *pj) {
 static void il_opdmp_load(RzILOp *op, RzStrBuf *sb, PJ *pj) {
 	RzILOpLoad *opx = op->op.load;
 	if (sb) {
-		rz_strbuf_append(sb, "load(key:");
+		rz_strbuf_appendf(sb, "load(mem:%u, key:", (unsigned int)opx->mem);
```
wargio, line 332: maybe let's avoid `unsigned int` and instead use `ut32`

librz/include/rz_il/definitions/mem.h
```diff
@@ -10,16 +10,31 @@
 extern "C" {
 #endif
 
-struct rzil_mem_t {
-	HtPP *kv_map;
-	ut32 min_unit_size; // minimal unit size in bit (len of value bv)
-};
-typedef struct rzil_mem_t RzILMem;
+typedef unsigned int RzILMemIndex;
```
wargio:
```suggestion
typedef ut32 RzILMemIndex;
```

wargio (commented):
TBH, this works, but the issue is RzIOBind that uses the phy memory and not virt/phy :/
anyway, that is fine, since the only one impacted is me with ELF bins used in AVR.
Please `sed 's/unsigned int/ut32/g'`

librz/il/vm_layer.c
```diff
@@ -321,40 +337,60 @@ RZ_API RzBitVector *rz_il_vm_mem_load(RzILVM *vm, ut32 mem_index, RzBitVector *k
...
-	RzBitVector *zero = rz_bv_new(m->min_unit_size);
-	RzBitVector *old_value = rz_il_mem_load(m, key);
-	rz_bv_free(old_value);
-	if (value) {
-		*value = zero;
+RZ_API RzBitVector *rz_il_vm_mem_loadw(RzILVM *vm, RzILMemIndex index, RzBitVector *key, ut32 n_bits) {
+	rz_return_val_if_fail(vm && key, NULL);
+	RzILMem *mem = rz_il_vm_get_mem(vm, index);
+	if (!mem) {
+		RZ_LOG_ERROR("Non-existent mem %u referenced\n", (unsigned int)index);
```
XVilka, line 367: Same, `ut32`?

librz/il/vm_layer.c
```diff
@@ -321,40 +337,60 @@ RZ_API RzBitVector *rz_il_vm_mem_load(RzILVM *vm, ut32 mem_index, RzBitVector *k
...
+ * an RZIL_EVENT_MEM_WRITE event
+ * \param  vm    RzILVM* pointer to VM
+ * \param  key   RzBitVector, aka address, a key to store data from memory
+ * \param  value RzBitVector, aka value to store in memory
+ */
+RZ_API void rz_il_vm_mem_storew(RzILVM *vm, RzILMemIndex index, RzBitVector *key, RzBitVector *value) {
+	rz_return_if_fail(vm && key && value);
+	RzILMem *mem = rz_il_vm_get_mem(vm, index);
+	if (!mem) {
+		RZ_LOG_ERROR("Non-existent mem %u referenced\n", (unsigned int)index);
```
XVilka, line 387: And here.

librz/include/rz_il/definitions/mem.h
```diff
@@ -10,16 +10,31 @@
...
 
-RZ_API RzILMem *rz_il_mem_new(ut32 min_unit_size);
+/**
+ * \brief A single memory as part of the RzIL VM.
+ *
+ * This can be seen as an of bitvectors, indexed by bitvector keys, covering a
```
XVilka: Missing "array"

XVilka (approved):
Good job! Only a few nitpicks and let's merge this!

librz/il/rzil_export.c
thestr4ng3r: `%u` is `unsigned int` so I **must** cast it to that when it's `ut32`.

librz/include/rz_il/definitions/mem.h
thestr4ng3r: Sounds good, pretty much any type is fine here.

librz/il/vm_layer.c
thestr4ng3r: see https://github.com/rizinorg/rizin/pull/2105#discussion_r774716561

thestr4ng3r:
> TBH, this works, but the issue is RzIOBind that uses the phy memory and not virt/phy :/

You are right, the impl in `buf_io.c` indeed always reads only from a specific fd, not the mapping. But that just means we have to extend it or add another RzBuffer implementation, whatever turns out less awkward.

librz/il/vm_layer.c
```diff
@@ -290,29 +293,42 @@ RZ_API void rz_il_vm_free(RzILVM *vm) {
...
+	}
+	rz_pvector_reserve(&vm->vm_memory, index + 1);
+	// Fill up with NULLs until the given index
+	while (rz_pvector_len(&vm->vm_memory) < index + 1) {
+		rz_pvector_push(&vm->vm_memory, NULL);
+	}
+	rz_pvector_set(&vm->vm_memory, index, mem);
+}
+
+RZ_API RzILMem *rz_il_vm_get_mem(RzILVM *vm, RzILMemIndex index) {
```
XVilka, line 311: Missing Doxygen

librz/il/rzil_export.c
wargio: > `%u` is `unsigned int` so I **must** cast it to that when it's `ut32`.

not really. ut32 is unsigned int

XVilka (approved):

PR #2109, opened 2021-12-16
[RzIL] Fix concept of Effects and Effect ops

thestr4ng3r:

**Detailed description**

This brings our concept of effects in line with BAP and fixes issues like values being evaluated prematurely in seq. The main point is that ops themselves **are** the effects, rather than returning them to be evaluated later.

**Test plan**

added unit tests and bf.


DMaroo:
For the `rz_il_*_handler`s, why return a `void*` if `NULL` is the only thing they'll be returning? Can't we just make them `void` return functions, and refactor accordingly wherever they get called?

thestr4ng3r:
It's because they are part of the dispatch table here: https://github.com/rizinorg/rizin/blob/d7de2bd0195bb8b2e5eeb95e8abe431b05b4d35f/librz/il/vm_layer.c#L55
You are right that we could make them `void` functions in the long run. This is the rough plan:
* Fix the perform/effect concept without refactoring too much outside this scope (this pr)
* Make ops more statically typed, i.e. `RzILOp` is split into `RzILOpBitvector`, `RzILOpEffect`, etc. like in BAP. This also has the effect that the handlers may have different signatures (grouped by type), so this is where we can turn them into `void`.
* Get rid of `RzPVector` oplists because `seq` can be used instead.

XVilka (approved):

PR #2122, opened 2021-12-20
Remove the wrong PACKAGE_INIT defined by meson and completely replace it

ret2libc:
Just remove the PACKAGE_INIT because broken and use our own definitions
of PACKAGE_PREFIX_DIR and set_and_check.


thestr4ng3r (approved):

XVilka (approved):
LGTM if green

PR #2126, opened 2021-12-21
Move remaining things from shlr/ to meson subprojects

ret2libc:

**Detailed description**

1) clean the meson.build files so that the top dirs does not include random things from shlr/x/d/ and librz/x/y but it just includes librz, which takes care of the other things under it
2) avoid the dependency problems with some shlr stuff that depends on rz_* modules while being dependencies of other rz_* modules
3) preparation work for building some plugins as shared libs instead of statically and potentially extract some of those into their own repo/rizin-extras. or also, if we want to make keystone the default assembler we could probably build it as shared plugin within a subproject so the whole project does not need to be converted to c++

**Test plan**

...

**Closing issues**

...


XVilka (changes requested):
Is it possible not to move jemalloc to the RzCore? Maybe also a subproject? See https://github.com/rizinorg/rizin/issues/209

ret2libc:
> Is it possible not to move jemalloc to the RzCore? Maybe also a subproject? See #209

Done

XVilka (approved):
No objections. As long as it's green.

thestr4ng3r (approved):

Issue #752, opened 2021-03-02
There seems to be no way to determine the base of a number when there are no analysis hints defined


ret2libc:
Copied from https://github.com/radareorg/radare2/issues/17727

# Description
I'm trying to implement a hotkey to toggle the immediate base in Cutter. For the hotkey to always work correctly, I need to obtain the current immediate base for the selected instruction. However, there seems to be no way to get that information, unless an analysis hint was defined for that particular address earlier.
The only alternative I have considered is parsing the opcode, but I'm not sure if that's a good idea (radare supports many different instruction sets, syntaxes, and disassemblers). `aoj`/`ao `does not seem to be useful in this case either.

See rizinorg/cutter#2429 (comment) :

@ret2libc

    I think there is no way currently to determine what is the base used for a number when no hint is defined yet. I see https://github.com/radareorg/radare2/blob/master/libr/parse/filter.c#L402 , which seems to mean that whatever is returned by the disassembler is shown to the user. Capstone seems to show the number as "decimal" if < 10 (by "decimal" i mean that it doesn't show 0x, but of course numbers < 10 are the same in hex and dec), but you probably can't assume that, as radare2 supports multiple disassemblers.

# Expected behavior
```
[0x00003362]> pd1
│           0x00003362      mov     edx, 5 
[0x00003362]> ahj 0x00003362
[{"addr":13154,"immbase":10}] 
[0x00003362]> pd1 @ 0x00003a94
│           0x00003a94      lea     rax, [0x00007766] 
[0x00003a94]> ahj 0x00003a94
[{"addr":14996,"immbase":16}] 
```
(Or maybe a different command to get the default immediate base or other information that could be useful in this case)
# Actual behavior
```
[0x00003362]> pd1
│           0x00003362      mov     edx, 5 
[0x00003362]> ahj 0x00003362
[] 
[0x00003362]> ahi 10 @ 0x00003362

[0x00003362]> ahj 0x00003362
[{"addr":13154,"immbase":10}] 
(...)
```
# Related to

rizinorg/cutter#2429

ret2libc:
@plaets I have copied the issue in Rizin repository as well, just FYI.

PR #2130, opened 2021-12-21
Convert ah commands to rzshell

p371jah:

**Detailed description**

- `rz_core_analysis_hint_list()` was renamed to `rz_core_analysis_hint_list_print()` 
- `rz_core_analysis_hint_list_print()` and `rz_core_analysis_hint_print()` use `RzCmdStateOutput`
- all commands work only with `core->offset`
- `ah` was replaced with `ahl`
- `ah.` and `ah <addr>` were replaced with `ahl.`
- missing `aho-` and `aht-` were added

**Test plan**

...

**Closing issues**

Related: #1442
...


XVilka:
Maybe it worth to address also the https://github.com/rizinorg/rizin/issues/752 while you are changing this code.

p371jah:
I guess #752 requires unifying of all disasm plugins.

ret2libc:
> Maybe it worth to address also the #752 while you are changing this code.

Completely separate problem I think. out of scope for this rzshell conversion IMO.

librz/core/canalysis.c
```diff
@@ -1258,7 +1258,7 @@ static void print_hint_h_format(HintNode *node) {
 // if mode == 'j', pj must be an existing PJ!
 static void hint_node_print(HintNode *node, int mode, PJ *pj) {
```
XVilka: Maybe switch the function argument to `RzOutputMode` enum?

XVilka (approved):
LGTM apart that minor thing.

p371jah:
I have ported the code but I don't know what to use as a summary for commands. Previously there was no description in the help,  but only command examples. Well I think I should move these examples to `yaml`.

ret2libc:
> I have ported the code but I don't know what to use as a summary for commands. Previously there was no description in the help, but only command examples. Well I think I should move these examples to `yaml`.

Yes move the examples there please. For the actual text, let's not worry too much at this time. We will improve the descriptions as we see them.

XVilka (approved):

XVilka:
@ret2libc take one more look, please, apart from that it's good to merge, I think.

wargio (approved):
LGTM !

librz/core/cmd/cmd_analysis.c
```diff
@@ -8905,3 +8542,268 @@ RZ_IPI RzCmdStatus rz_rzil_vm_status_handler(RzCore *core, int argc, const char
...
+RZ_IPI RzCmdStatus rz_analysis_hint_del_offset_handler(RzCore *core, int argc, const char **argv) {
+	rz_analysis_hint_unset_offset(core->analysis, core->offset);
+	return RZ_CMD_STATUS_OK;
+}
+
+RZ_IPI RzCmdStatus rz_analysis_list_struct_offsets_handler(RzCore *core, int argc, const char **argv) {
+	ut64 toff = rz_num_math(core->num, argv[1]);
+	if (!toff) {
+		return RZ_CMD_STATUS_WRONG_ARGS;
+	}
```
ret2libc, line 8799: I think this should be replaced with `rz_num_is_valid_input`

ret2libc (commented):
LGTM apart from that small thing with rznum.

librz/core/cmd/cmd_analysis.c
p371jah: 1) About check on zero: zero offset is not allowed.
2) About `rz_num_is_valid_input`: I expect that rzshell validates arguments of type `RZ_CMD_ARG_TYPE_NUM`/`RZ_CMD_ARG_TYPE_RZNUM`

ret2libc: > About check on zero: zero offset is not allowed

Ah ok

> About rz_num_is_valid_input: I expect that rzshell validates arguments of type RZ_CMD_ARG_TYPE_NUM/RZ_CMD_ARG_TYPE_RZNUM

This is a good point actually. It is not done at the moment, but maybe it should be done indeed.

ret2libc (dismissed):

librz/core/cmd/cmd_analysis.c
ret2libc: Mmmm why zero offset is not allowed by the way?

p371jah: Because all structs have field at zero offset.

PR #2137, opened 2021-12-24
Plugins reorganization.

ret2libc:
I will clean up this PR and split commits later, but first I'd like some feedback on what you think about this approach to deal with plugins.

This PR allows to build selected plugins and choose whether to build them statically inside rizin or as shared libraries that are installed in the <prefix>/<plugindir> (like others 3rd party plugins). For now this is just a PoC so please just look at `lang` and `bp` plugins, but the others would look the same.

One can specify the plugins to enable in this way:
```
$ meson -Dlang_plugins=shared.* build # all lang plugins are shared libs
$ meson -Dlang_plugins=static.* build # all lang plugins are statically linked in librz_lang
$ meson -Dlang_plugins=[] build # no lang plugins are built
$ meson -Dlang_plugins=[shared.pipe,static.lib] build # lang pipe plugin is built in a shared lib, lang lib plugin statically
```

Pros:
- provide a fine grained way to select the plugins and how to build them
- provide a way to compile a very basic rizin with only the few things that one use instead of all of them
- make separation of plugins and core clearer. each plugin now has its own (very simple in most cases) meson.build file. This means that dependencies for a particular plugin are explicitly dependencies of the plugin and not of the whole librz_* module. ofc in case of static plugins, the dependency would become a dependency of the librz_* module, but if the plugin is dynamically linked we can keep the plugin dependencies separated from the core dependencies.
- ease the transition of some plugins to an external repo or a "contrib_plugins" repo with some less maintained stuff, because each plugin should be already quite confined and independent of others things.
- ease introduction of heavier plugins like rz_keystone, which we could have by default as a shared plugin instead of a static one, so that the whole project can remain C while the plugin become C++.

**Detailed description**

- [ ] provide example of shared plugin
- [ ] update doc with shared/static settings
- [ ] add CI jobs for all shared and all static plugins
- [ ] add CI job for no-plugins configuration

**Test plan**

...

**Closing issues**

...


thestr4ng3r (commented):
The general approach looks good to me. I especially like the generic `librz/shared_plugin.c.in` to be reused for all plugins.
The only complaint I have at the moment is that the directory nesting for plugins that are just one c file, e.g. `librz/bp/p/arm/bp_arm.c` instead of `librz/bp/p/bp_arm.c` is too much. For multi-file plugins this can make sense, but single-file ones in my opinion should be in the root `/p/` and their meson files flattened out in `p/meson.build`.

thestr4ng3r:
Some inspiration from qemu: They have a lot of different targets to compile from different sources but in a similar way, where each is in its own subdirectory.
But instead of doing the `executable(...` etc. calls in the subdirectories, they collect them in some global dictionaries (e.g. `target_arch` in https://gitlab.com/qemu-project/qemu/-/blob/master/target/avr/meson.build) and then they loop over the requested targets in a parent dir and pick whatever is needed from the dicts: https://gitlab.com/qemu-project/qemu/-/blob/master/meson.build#L2977-2996

In our case this would mean that instead of calling `subdir(p)` in a loop, some plugins may be in subdirs, some not, but they just fill a global dictionary and the rest is done outside.

XVilka:
@ret2libc should this be closed?

ret2libc:
Ok to close but don't delete the branch please!

PR #2147, opened 2021-12-28
Add IO/Map-based RzBuffer Implementation

thestr4ng3r:

**Detailed description**

This adds an RzBuffer implementation that binds against RzIO's `rz_io_read_at`/`rz_io_write_at` in order to access the mapped memory. This is in contrast to the previous io RzBuffer, which has been renamed to `..._io_fd` and which reads directly from a single file descriptor without mapping.
To keep bf working, `bin_bf` now maps an area of zeroes because otherwise the initial memory contents in RzIL would read as 0xff.

**Test plan**

run the unit tests


librz/util/buf_io.c
```diff
@@ -1,82 +1,58 @@
...
-	return priv->iob->fd_seek(priv->iob->io, priv->fd, addr, io_whence);
+	BufIOPriv *priv = b->priv;
+	priv->offset = rz_seek_offset(priv->offset, 0, addr, whence);
+	return priv->offset; // TODO: this might be negative and be detected as error
 }
 
 static ut64 buf_io_get_size(RzBuffer *b) {
-	struct buf_io_priv *priv = get_priv_io(b);
-	return priv->iob->fd_size(priv->iob->io, priv->fd);
+	return 0; // TODO: overflow :-(
```
ret2libc: ?

librz/util/buf.c
```diff
@@ -484,17 +489,30 @@ RZ_API RZ_OWN RzBuffer *rz_buf_new_with_bytes(RZ_NULLABLE RZ_OWN const ut8 *byte
...
+RZ_API RZ_OWN RzBuffer *rz_buf_new_with_io_fd(RZ_NONNULL void *iob, int fd) {
 	rz_return_val_if_fail(iob && fd >= 0, NULL);
 
-	struct buf_io_user u = { 0 };
+	struct buf_io_fd_user u = { 0 };
 
 	u.iob = (RzIOBind *)iob;
 	u.fd = fd;
 
 	return new_buffer(RZ_BUFFER_IO, &u);
```
ret2libc:
```suggestion
	return new_buffer(RZ_BUFFER_IO_FD, &u);
```

ret2libc (changes requested):
Where do you need this?

thestr4ng3r:
> Where do you need this?

It will be used to bind RzIL memory against io when desired.

librz/util/buf_io.c
thestr4ng3r: io has a size of 2^64, which we can't express. I've just removed the size callback entirely since not implementing it is supported by RzBuffer too.

XVilka (approved):

PR #2155, opened 2021-12-30
[RzIL] RzReg/Variable Binding

thestr4ng3r:

**Detailed description**

Plugins do not create register-based variables themselves anymore, but they are derived from the register profile.
However not all registers may be bound and not all variables may be actual registers.
The concrete relations between an `RzILVM` and registers is described by an `RzILRegBinding` object. It is simply a list of register names and variables are created of the same names.
Such a binding can be created in two ways currently:
* `rz_il_reg_binding_derive()` automatically selects a set of non-overlapping registers, favoring the "larger" ones (e.g. for rax and eax, you will get rax). It's currently used for all plugins, always.
* `rz_il_reg_binding_exactly()` simply takes the given list as-is. Right now it's only used for testing, but it is plausible that the automatic approach above might not fit some plugin the future well, so they might also (optionally) specify the registers themselves at some point.

When stepping with `aezs`, the registers are synchronized automatically. From now on, `aezv` is primarily useful for debugging, but interacting with emulation from the user side should be done with `ar`.

**Test plan**

* Added unit tests
* Execute instructions with `aezs` and observe the register contents from `ar` being used and updated.


librz/include/rz_il/rz_il_reg.h
```diff
@@ -0,0 +1,14 @@
...
+
+#ifndef RZIL_REG_H
+#define RZIL_REG_H
+
+#include <rz_reg.h>
+
+typedef struct rz_il_reg_binding_t {
+	size_t regs_count;
+	char **regs; //< regs_count names of registers that are bound to variables of the same name
+} RzILRegBinding;
```
wargio: maybe use RzPVector ?

thestr4ng3r: It should be immutable, so no reason to use RzPVector after the initial creation anymore imo.

wargio: i would still use RzPVector, since the implementation of these bindings is a representation of RzVector

thestr4ng3r: It's different now anyway. They are not pointers anymore, but structs and the array gives us much more type-safety than RzVector (which is still used in building, but only there).

XVilka (approved):
Looks good. Can it handle the case of EFLAGS registers? Where there is a register that consists of many particular bits. If yes - please add a test for this.

See, for example `SREG` case from AVR:
```
		// status bit register (SREG)
		"gpr	sreg	.8	38	0\n"
		"gpr	cf	.1	38.0	0\n" // Carry. This is a borrow flag on subtracts.
		"gpr	zf	.1	38.1	0\n" // Zero. Set to 1 when an arithmetic result is zero.
		"gpr	nf	.1	38.2	0\n" // Negative. Set to a copy of the most significant bit of an arithmetic result.
		"gpr	vf	.1	38.3	0\n" // Overflow flag. Set in case of two's complement overflow.
		"gpr	sf	.1	38.4	0\n" // Sign flag. Unique to AVR, this is always (N ^ V) (xor), and shows the true sign of a comparison.
		"gpr	hf	.1	38.5	0\n" // Half carry. This is an internal carry from additions and is used to support BCD arithmetic.
		"gpr	tf	.1	38.6	0\n" // Bit copy. Special bit load and bit store instructions use this bit.
		"gpr	if	.1	38.7	0\n" // Interrupt flag. Set when interrupts are enabled.
```
https://github.com/rizinorg/rizin/blob/dev/librz/analysis/p/analysis_avr.c#L1987

thestr4ng3r:
In that case, you should only get the `sreg`, which would work. It's probably nicer to have the individual flags as variables though and I think BAP does it like this too.
One way would be to add an additional heuristic for single-bit registers and favor them over the "larger" ones.

thestr4ng3r:
@XVilka Added support for this. Now for the following profile:
```
		"gpr	rax	.64	0	0\n"
		"gpr	eax	.32	0	0\n"
		"gpr	ax	.16	0	0\n"
		"gpr	al	.8	0	0\n"
		"gpr	ah	.8	1	0\n"
		"gpr	rbx	.64	8	0\n"
		"gpr	ebx	.32	8	0\n"
		"gpr	bx	.16	8	0\n"
		"gpr	bl	.8	8	0\n"
		"gpr	bh	.8	9	0\n"
		"gpr	sreg	.8	40	0\n"
		"gpr	cf	.1	40.0	0\n"
		"gpr	zf	.1	40.1	0\n"
		"gpr	nf	.1	40.2	0\n"
		"gpr	vf	.1	40.3	0\n"
		"gpr	hfsf	.2	40.4	0\n"
		"gpr	tf	.1	40.6	0\n"
		"gpr	if	.1	40.7	0\n";
```
you get rax, rbx, cf, zf, nf, vf, hfsf, tf and if.

XVilka (approved):
Looks complete now.

PR #2156, opened 2021-12-31
RzDiff optimizations and fix #1770

### LINKED ISSUE #1770: rz-diff -H view is wrong when one file is much shorter than the other
Version:
```
$ rz-diff -v
rz-diff 0.3.0 @ linux-x86-64, package: 0.3.0-1.fc33 (Fedora)
```

```
$ python3 -c 'print("A"*0x1000)' > /tmp/a
$ python3 -c 'print("B"*0x100)' > /tmp/b
$ rz-diff -H /tmp/a /tmp/b
```

scroll down until the end of the /tmp/b file. things at some point the /tmp/b file will stop scrolling but the /tmp/a one keeps going. Then if you scroll up the /tmp/b file starts scrolling up immediately.

image
image

notice the offsets: in the first image, it is 0x3b0 on /tmp/a and still 0x100 on /tmp/b; in the second immage i have scrolled up a bit and now it is 0x0 on /tmp/b, while being just 0x5b0 on /tmp/a.

I would expect the two offsets to always be in sync.

wargio:

**Detailed description**

- Implements infinite scroll which fixes #1770
image

- Optimization on diffing functions by avoiding analyzing the whole binary before hand and instead diffing the functions themselves directly.
- Added `-B` to analyze the whole binary via `aaa` optionally


librz/core/gdiff.c
```diff
@@ -447,6 +447,13 @@ static int graph_construct_nodes(RzCore *core, RzCore *core2, RzAnalysisFunction
 				core2->config = core->config;
 
 				char *modified = rz_core_cmd_strf(core2, "pdb @ 0x%08" PFMT64x, bbi->diff->addr);
```
XVilka, line 449: 1. Check if the `modified` is NULL before passing to the `strcmp()`
2. Is it possible to use the API instead of calling the `pdb` command while you are on it?

librz/core/gdiff.c
```diff
@@ -463,6 +470,7 @@ static int graph_construct_nodes(RzCore *core, RzCore *core2, RzAnalysisFunction
 				free(modified);
 				core2->config = oc;
 			} else {
+			graph_construct_nodes_is_same:
```
XVilka, line 473: Maybe add a  new line before the label - otherwise it looks weird a bit, IMHO.

librz/main/rz-diff.c
```diff
@@ -1465,100 +1480,97 @@ typedef enum diff_hex_len_t {
...
-			lp += printline("%s%02x%s ", same ? match : unmatch, buffer_b[pos + i], reset);
-		} else {
-			lp += printline("%s%02x%s ", unmatch, buffer_b[pos + i], reset);
-		}
-	}
-	if (i < hexlen) {
-		memset(line + lp, ' ', (hexlen - i) * 3);
-		lp += (hexlen - i) * 3;
-	}
+static inline void diff_hexdump_line(DiffHexView *hview, DiffHexLen hlen, ut64 pos, ssize_t read_a, ssize_t read_b, ssize_t skip_a, ssize_t skip_b) {
```
XVilka, line 1560: Is it possible to avoid so many arguments? Maybe combine them into some kind of structure?

librz/main/rz-diff.c
```diff
@@ -1465,100 +1480,97 @@ typedef enum diff_hex_len_t {
...
+	case DIFF_HEX_32:
+		hexlen = 32;
+		break;
+	default:
+		hexlen = 8;
+		break;
 	}
 
+#define printline(fmt, ...) snprintf(line + lp, RZ_MAX(lsize - lp, 0), fmt, ##__VA_ARGS__)
+	lp = diff_hexdump_partial(hview, hexlen, 0, lsize, buffer_a, buffer_b, address_a, address_b, hview->size_a, hview->size_b, pos, read_a, read_b, skip_a, skip_b);
```
XVilka, line 1585: Same applies here

XVilka (changes requested):

librz/main/rz-diff.c
wargio: it's an inline function, there is no real difference on using a struct or not in this specific case.

librz/core/gdiff.c
wargio: i just checked the pdb api, and i'm definetely not going to use the native one for this. there is no way to get a text unless i use rz_cons which is a mess.

wargio:
@XVilka

librz/core/gdiff.c
XVilka: I meant before the label itself, not before the code.

XVilka (approved):

PR #2157, opened 2021-12-31
Use analysis_get_function_in() wherever possible

p371jah:

**Detailed description**

- Fixes memory leaks related to `rz_analysis_get_functions_in()` usage.
- Get rid of some calls to deperecated `rz_analysis_get_fcn_in()`

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/librz/core/cmd/cmd_analysis.c
+++ b/librz/core/cmd/cmd_analysis.c
@@ -433,7 +433,8 @@ static const char *help_msg_as[] = {
 /**
  * \brief Helper to get function in \p offset
  *
- * Case of overlapped functions is treated as an error.
+ * Case of overlapped functions is treated as an error
+ * if \p offset is not an entry point.
  */
 static RzAnalysisFunction *analysis_get_function_in(RzAnalysis *analysis, ut64 offset) {
 	RzAnalysisFunction *fcn = rz_analysis_get_function_at(analysis, offset);
--- a/librz/core/cmd/cmd_analysis.c
+++ b/librz/core/cmd/cmd_analysis.c
@@ -430,6 +430,31 @@ static const char *help_msg_as[] = {
 	NULL
 };
 
+/**
+ * \brief Helper to get function in \p offset
+ *
+ * Case of overlapped functions is treated as an error.
+ */
+static RzAnalysisFunction *analysis_get_function_in(RzAnalysis *analysis, ut64 offset) {
+	RzAnalysisFunction *fcn = NULL;
+	RzList *list = rz_analysis_get_functions_in(analysis, offset);
+	if (rz_list_empty(list)) {
+		RZ_LOG_ERROR("No function found\n");
+		goto exit;
+	}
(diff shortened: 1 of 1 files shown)


librz/core/cmd/cmd_analysis.c
```diff
@@ -430,6 +430,31 @@ static const char *help_msg_as[] = {
 	NULL
 };
 
+/**
+ * \brief Helper to get function in \p offset
+ *
+ * Case of overlapped functions is treated as an error.
+ */
+static RzAnalysisFunction *analysis_get_function_in(RzAnalysis *analysis, ut64 offset) {
```
XVilka, line 439: Please name to `analysis_get_first_function_in()`

librz/core/cmd/cmd_analysis.c
```diff
@@ -430,6 +430,31 @@ static const char *help_msg_as[] = {
...
+	RzAnalysisFunction *fcn = NULL;
+	RzList *list = rz_analysis_get_functions_in(analysis, offset);
```
XVilka, line 444: Irrelevant to this PR but maybe we want to return a vector from this function in the future, to reduce the memory consumption.

XVilka (changes requested):
Overall I approve the change.

librz/core/cmd/cmd_analysis.c
p371jah: Right above you can see `Case of overlapped functions is treated as an error`. So this function returns `NULL` if there are more than one function in offset.

librz/core/cmd/cmd_analysis.c
XVilka: Ah, ok, makes sense then.

XVilka (approved):

librz/core/cmd/cmd_analysis.c
thestr4ng3r, line 444: Better to first try `rz_analysis_get_function_at()` before falling back to `_in` here (or if the list contains multiple entries and there is one function whose entrypoint is exactly `offset`, pick this instead of erroring, which would be equivalent).
This way, even for overlapping functions, these commands will still work if you execute them at the exact entrypoints like `afb @ fcb.ulumulu`, since entrypoints are unique.
Then in the `RZ_LOG_ERROR("Overlapped functions found\n");` below, you could also print the hint that running the code at the entrypoint of one of the functions will disambiguate.

thestr4ng3r (changes requested):

librz/core/cmd/cmd_analysis.c
p371jah: Done except for the hint. I just don't know how to formulate it briefly.

thestr4ng3r: "Multiple overlapping functions found at 0x%08x. Re-run this command at the entrypoint of one of them to disambiguate."

thestr4ng3r (approved):

PR #2178, opened 2022-01-05
Add Documentation for RzIL

thestr4ng3r:
as discussed on mattermost with @XVilka

### DIFF
--- a/doc/rzil.md
+++ b/doc/rzil.md
@@ -1,9 +1,9 @@
 RzIL
 ====
 
-RzIL is Rizin's new intermediate language, primarily for representing the
-semantics of machine code. It is designed as a clone of BAP's [Core
-Theory](http://binaryanalysisplatform.github.io/bap/api/master/bap-core-theory/Bap_core_theory/),
+RzIL is the new intermediate language in Rizin, primarily intended for
+representing the semantics of machine code. It is designed as a clone of BAP's
+[Core Theory](http://binaryanalysisplatform.github.io/bap/api/master/bap-core-theory/Bap_core_theory/),
 with minor deviations where necessary.
 
 Background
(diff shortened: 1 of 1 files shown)


doc/rzil.md
```diff
@@ -0,0 +1,132 @@
+RzIL
+====
+
+RzIL is Rizin's new intermediate language, primarily for representing the
```
XVilka: `is the`

doc/rzil.md
```diff
@@ -0,0 +1,132 @@
...
+Background
+----------
+
+RzIL was introduced as a consequence of dissatisfaction with certain properties
+of ESIL, in particular its highly ambiguous and weak typing, requiring [major
+hacks](https://github.com/rizinorg/rizin/blob/2e065789a70edd20909aadcdf7f9c45b9af699fb/librz/analysis/esil/esil.c#L1025-L1032)
+to determine the bit-width of a value required for further calculations, for
+example.
+
+At this point, the two options to proceed were to either enhance ESIL and
```
XVilka: "point, two options"

doc/rzil.md
```diff
@@ -0,0 +1,132 @@
...
+but goes no further than that. The typedefs `RzILOpBitVector` and `RzILOpBool`
+are mere aliases of `RzILOpPure` to indicate which type is meant whenever it is
+known in the code. Further splitting is not possible while keeping ops like
+`ite` polymorphic.
+
+So far, this is only an implementation detail, but does not actually affect the
+language itself. However, as there are also certain limitations of OCaml's type
+system, our dynamic typing opens up other possibilities.
+
+For instance, the `append` op is used for appending two bitvectors of arbitrary
```
XVilka, line 88: Probably also worth mentioning the lack of `concat` op.

doc/rzil.md
```diff
@@ -0,0 +1,132 @@
...
+limited to the body of the `let` expression. As part of the RzIL vm, these
+variables are called "local pure variables".
+
+`Var` vars in Core Theory are so-called virtual variables, used mostly for
+temporary scratch locations. Their scope is unlimited, but when an instruction
+is lifted to Core Theory as part of the BAP Knowledge Base (this is
+approximately a very innovative kind of database),
+[fresh](https://github.com/BinaryAnalysisPlatform/bap/blob/92d67c83fe0988b8a25bf563bdf33a9594db3e54/lib/bap_core_theory/bap_core_theory_var.ml#L99-L101)
+temporary variables are always chosen with a globally unique identifier, thus
+automatically distinguishing temporary variables from multiple lifted
```
XVilka, line 131: In the future, we might actually consider implementing something similar to KB as well. It will be certainly helpful for some analyses.

XVilka (approved):
Looks good!

doc/rzil.md
```diff
@@ -0,0 +1,132 @@
...
+OCaml is automatically well-typed on the level of the IL.
+
+As a specific example, BAP uses the OCaml type `'s bitv` to express pure
+bitvector ops, where the size of the bitvector is statically given by the type
+variable `'s`. This ensures for example that the op `add : 's bitv -> 's bitv ->
+'s bitv` will only be able to take bitvectors of identical size as operands and
+return a bitvector of the same size as well. At the same time, it is possible to
+have polymorphic ops such as the if-then-else `ite : bool -> 'a pure -> 'a pure
+-> 'a pure` which can operate on both `bool` and `'s bitv` values (`bool` is an
+alias for `Bool.t pure` and `'s bitv` is an alias for `'s Bitv.t pure`).
```
wargio, line 74: i'm not surehow can these be clear even for those who are not familiar with bap/ocaml stuff. i would simplify the stmts to something like `ite(bool condition, pure true_val, pure false_val) -> pure`

doc/rzil.md
thestr4ng3r: How exactly?

doc/rzil.md
thestr4ng3r: Yes, I think there is a lot to learn from the KB approach. Still, I don't think this would apply as-is to variables like in BAP as there is a lot in KB that is pretty much impossible or impractical to have in C like the monadic code.

doc/rzil.md
wargio: `RzIL is the new Rizin's intermediate language, primarily for representing the`

thestr4ng3r: What is the `new Rizin`?
I think it would only make sense as `RzIL is the new intermediate language in Rizin`

doc/rzil.md
thestr4ng3r: Dropping the type variables makes it lose its meaning here. In fact, the entire chapter's core is explaining the differences between the exact details of BAP's OCaml code and RzIL. I've just added a sentence above to indicate that some ML knowledge is required here.

XVilka (approved):

PR #1829, opened 2021-10-12
Fix a crash during variables type propagation

XVilka:
**Your checklist for this pull request**
- [x] I've read the [guidelines for contributing](https://github.com/rizinorg/rizin/blob/master/DEVELOPERS.md) to this repository
- [x] I made sure to follow the project's [coding style](https://github.com/rizinorg/rizin/blob/master/DEVELOPERS.md#code-style)
- [ ] I've added tests that prove my fix is effective or that my feature works (if possible)
- [ ] I've updated the documentation and the [rizin book](https://github.com/rizinorg/book) with the relevant information (if needed)

**Detailed description**

Clone the variables vector before retyping it, since retyping might trigger overlapping var deletion in case of the new type is big enough to overshadow the next variables.

**Test plan**

CI is green

PR #2222, opened 2022-01-19
Fix OOB in overlapping variables resolution

b1llow:
# SQUASH ME

**Detailed description**

Fixes the problem of removing the element from vector during the iteration of the vector elements.

**Test plan**

CI is green


ret2libc (commented):
Is the body of this loop changing `fcn->vars`? If so, I think this is a problem and this kind of fix does not look ideal (e.g. what if the array is re-allocated but with an offset that makes everything wrong, while some parts of it are still within the older bounds of the array?).

librz/core/analysis_tp.c
```diff
@@ -920,6 +920,10 @@ RZ_API void rz_core_analysis_type_match(RzCore *core, RzAnalysisFunction *fcn, H
 	// Type propagation for register based args
 	void **vit;
 	rz_pvector_foreach (&fcn->vars, vit) {
+		// maybe fcn->vars has changed
+		if (vit >= (void **)fcn->vars.v.a + fcn->vars.v.len) {
+			return;
```
p371jah: You can't just use `return` here.

b1llow:
> Is the body of this loop changing `fcn->vars`? If so, I think this is a problem and this kind of fix does not look ideal (e.g. what if the array is re-allocated but with an offset that makes everything wrong, while some parts of it are still within the older bounds of the array?).

Maybe I didn't explain it clearly, It's just "remove some variables to the overlaps resolution".
This leads to a shorter length of `fcn->vars`, which leads to Address boundary error.

https://github.com/rizinorg/rizin/blob/fc6f146f98ece802fd48c19a3c7a65509e10fd45/librz/core/analysis_tp.c#L934-L941

https://github.com/rizinorg/rizin/blob/21584e416cdcef2fa7d855c5aabf592a965f0e8d/librz/analysis/var.c#L96-L108

librz/core/analysis_tp.c
b1llow: maybe use `break` is enough

wargio:
i guess this is actually a problem. i think instead of removing those in that loop, the elements to remove should be added to a list which can then later be iterated and used to remove them from the vector

b1llow:
> i guess this is actually a problem. i think instead of removing those in that loop, the elements to remove should be added to a list which can then later be iterated and used to remove them from the vector

But that's bad, even need to rewrite `var_type_set` and `resolve_var_overlaps`

ret2libc:
> > i guess this is actually a problem. i think instead of removing those in that loop, the elements to remove should be added to a list which can then later be iterated and used to remove them from the vector
> 
> But that's bad, even need to rewrite `var_type_set` and `resolve_var_overlaps`

Why is that bad? I agree with @wargio that is the right way to do the thing. I think just checking the boundaries like what you are doing now would lead to other kind of (more subtle) errors.

Moreover:
```C
#define rz_pvector_foreach(vec, it) \
	for (it = (void **)(vec)->v.a; (vec)->v.len && it != (void **)(vec)->v.a + (vec)->v.len; it++)
```

1) if only elements *after* the current one can be deleted, fixing `rz_pvector_foreach` to check for `<` instead of `!=` might be ok, however you'd have to address for the case where you don't have to increase `it` because something has been deleted. In particular, if you have `[1,2,3,4(cur),5]` and while iterating over `4` you delete `4`, now `5` would be shifted in the position of `4`, but due to `it++` in the foreach loop you'd skip the element `5`. I hope it's clear what I mean.

2) if random elements in the vector can be deleted, there is no good way to fix it just within that loop. Collecting the list of elements to remove first and then remove them in a following loop is the onlly solution i can think of.

@thestr4ng3r cc.

wargio:
because iterating over a vector that is also modified during the iteration is generally bad.

wargio:
also if you iterate over an array `[1,2,3,4(cur),5]` and `4` gets deleted, then `5` gets skipped. so it is probably better to clone the array and iterate over the copy, instead of actually using the original one.

b1llow:
> also if you iterate over an array `[1,2,3,4(cur),5]` and `4` gets deleted, then `5` gets skipped. so it is probably better to clone the array and iterate over the copy, instead of actually using the original one.

OK, this approach is acceptable, for me.

@ret2libc Because I'm a bit lazy, and there are quite a few places where `var_type_set` is referenced, I think if it's not necessary, try not to refactor `var_type_set`.

thestr4ng3r:
I am out of the loop. What is the problem that is trying to be solved here?

wargio:
> I am out of the loop. What is the problem that is trying to be solved here?

the loop is iterated while the array used in the loop is also modified.

thestr4ng3r:
Is there a test case for this?

b1llow:
> Is there a test case for this?

no test case.

wargio (approved):

librz/core/analysis_tp.c
```diff
@@ -917,9 +917,11 @@ RZ_API void rz_core_analysis_type_match(RzCore *core, RzAnalysisFunction *fcn, H
 			rz_list_free(fcns);
 		}
 	}
+
 	// Type propagation for register based args
+	RzPVector *cloned_vars = (RzPVector *)rz_vector_clone((RzVector *)&fcn->vars);
 	void **vit;
-	rz_pvector_foreach (&fcn->vars, vit) {
+	rz_pvector_foreach (cloned_vars, vit) {
```
thestr4ng3r: Wouldn't this uaf if a var is removed inside the loop? The variable would not only be removed from `fcn->vars`, but also freed.

thestr4ng3r (changes requested):
Please add a test for this, or at least provide a binary/example how to reproduce the original bug.

librz/core/analysis_tp.c
b1llow: In theory it will.

And `resolve_var_overlaps` using the same method seems to be no problem, my fault.

https://github.com/rizinorg/rizin/blob/21584e416cdcef2fa7d855c5aabf592a965f0e8d/librz/analysis/var.c#L96-L108

XVilka:
Here is the test https://github.com/rizinorg/rizin/pull/1829/files#diff-1e6df3f8c133fea40c760e319419668a5a4cea9a75f10b0c3cc462fa1ab60e11

thestr4ng3r:
I see. I think another good way to test these specific edge cases might be to write unit tests where you create an `RzCore` and manually craft some minimal functions with variables that will overlap in exactly the way to currently cause such uafs, if that is possible. But it might also be tricky because the type propagation depends on many factors.

Regarding solving this problem, perhaps one could first remove all non-reg vars from the cloned vector, because these are the only ones we iterate over and overlap resolving will not delete them.

If that does not work, the more general solution would be to not do any overlap resolving inside the loop at all (basically change `var_type_set()` to not do it), but in a second pass do the overlap resolution for all vars. This may be possible by first sorting all vars by offset and then doing the overlap resolution from start to end (iterating with an index instead of pointer to avoid reallocation issues), since overlaps should only remove variables of higher offset iirc.

XVilka:
Please add a test from https://github.com/rizinorg/rizin/pull/1829/files

b1llow:
Is this correct behavior ?

```
 [XX] db/analysis/x86_64 x86_64 int64_t variable overlapping removal
RZ_NOPLUGINS=1 rizin -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -Qc 's 0x10270
af
afv
aaft
afv
' bins/elf/arch-x86_64-ls
-- stdout
--- expected
+++ actual
@@ -1,4 +1,3 @@
-void fcn.00010270(int64_t arg1, int64_t arg2, int64_t arg3, int64_t arg4, int64_t arg5);
 var int64_t var_10h_2 @ rsp+0x10
 var int64_t var_8h @ rsp+0x18
 var int64_t var_10h @ rsp+0x20
@@ -18,7 +17,6 @@
 arg int64_t arg3 @ rdx
 arg int64_t arg2 @ rsi
 arg int64_t arg1 @ rdi
-void fcn.00010270(int64_t arg1, int64_t arg2, const char **s, int64_t arg4, int64_t arg5);
 var const char * var_10h_2 @ rsp+0x10
 var void ** s1 @ rsp+0x18
 var uint64_t var_10h @ rsp+0x20
```

XVilka (approved):

thestr4ng3r (approved):

wargio (approved):

PR #2225, opened 2022-01-20
Change OpenBSD CI MALLOC_OPTIONS to be faster

thestr4ng3r:

**Detailed description**

From malloc(3):
```
     C       "Canaries".  Add canaries at the end of allocations in order to
             detect heap overflows.  The canary's content is checked when free
             is called.  If it has been corrupted, the process is aborted.

     D       "Dump".  malloc() will dump statistics to the file ./malloc.out,
             if it already exists, at exit.  This option requires the library
             to have been compiled with -DMALLOC_STATS in order to have any
             effect.

     F       "Freecheck".  Enable more extensive double free and use after
             free detection.  All chunks in the delayed free list will be
             checked for double frees.  Unused pages on the freelist are read
             and write protected to cause a segmentation fault upon access.

     G       "Guard".  Enable guard pages.  Each page size or larger
             allocation is followed by a guard page that will cause a
             segmentation fault upon any access.

     J       "More junking".  Increase the junk level by one if it is smaller
             than 2.

     j       "Less junking".  Decrease the junk level by one if it is larger
             than 0.  Junking writes some junk bytes into the area allocated.
             Junk is bytes of 0xdb when allocating; freed chunks are filled
             with 0xdf.  By default the junk level is 1: after free, small
             chunks are completely junked; for pages the first part is junked.
             After a delay, the filling pattern is validated and the process
             is aborted if the pattern was modified.  For junk level 2,
             junking is done on allocation as well and without size
             restrictions.  If the junk level is zero, no junking is
             performed.

     R       "realloc".  Always reallocate when realloc() is called, even if
             the initial allocation was big enough.

     S       Enable all options suitable for security auditing.

     U       "Free unmap".  Enable use after free protection for larger
             allocations.  Unused pages on the freelist are read and write
             protected to cause a segmentation fault upon access.

     X       "xmalloc".  Rather than return failure, abort(3) the program with
             a diagnostic message on stderr.  It is the intention that this
             option be set at compile time by including in the source:

                   extern char *malloc_options;
                   malloc_options = "X";

             Note that this will cause code that is supposed to handle out-of-
             memory conditions gracefully to abort instead.

     <       "Halve the cache size".  Decrease the size of the free page cache
             by a factor of two.

     >       "Double the cache size".  Increase the size of the free page
             cache by a factor of two.
```

Trying out some options, `J` primarily seems to introduce slowdown, so let's see how it works with these.


thestr4ng3r:
Results:

|  | |
|-|-|
| `MALLOC_OPTIONS=S` | 83 minutes |
| `MALLOC_OPTIONS=CFGU` | 27 minutes |
| `MALLOC_OPTIONS=` | 21 minutes |

thestr4ng3r:
@XVilka @kazarmy What do you think about removing the `openbsd_fast` job and only using `MALLOC_OPTIONS=CFGU`?

XVilka (approved):
No objections. Also good it tests the shared memory feature 👍

kazarmy:
I have a feeling it's the junking that makes OpenBSD's `malloc()` to be significantly different from ASAN's `malloc()` but I also have a feeling that any bugs caught by it will be seriously difficult to debug.

PR #2177, opened 2022-01-05
Support debugging DMP files

GustavoLCR:
# DO NOT SQUASH

**Your checklist for this pull request**
- [x] I've read the [guidelines for contributing](https://github.com/rizinorg/rizin/blob/master/DEVELOPERS.md) to this repository
- [x] I made sure to follow the project's [coding style](https://github.com/rizinorg/rizin/blob/master/DEVELOPERS.md#code-style)
- [ ] I've documented or updated the documentation of every function and struct this PR changes. If not so I've explained why.
- [x] I've added tests that prove my fix is effective or that my feature works (if possible)
- [ ] I've updated the [rizin book](https://github.com/rizinorg/book) with the relevant information (if needed)

**Detailed description**

* Add Triage Dump (Minidump) parsing support for dmp64 bin plugin
* Support debugging DMP files (Processes, Threads)
* Adds dependency of `rz_type` to `rz_debug` so I can use the type API to get structure offsets from PDB
* Fix some bugs that I found on the way (6cd8f03a8ddd9635a4c27f1fe261f4f881cbf918, fed4eee2bab06cfa9d676b75b0df62b45345dde8, 4ed8f64e90f98008c09efd234d219796a428b1a3, ff56f18fac0d407806f20e56889aec42183f3a47, d8174e84ca05751236cc09d6eaa4d7764cf09404)

Moved from https://github.com/rizinorg/rizin/pull/2171

**Test plan**

* PDB unit tests pass
* DMP tests pass
* WinKD plugin still works

**Closing issues**

Addresses most of #189

Issue #1241, opened 2021-06-24
Support for alignment and packing in composite types


XVilka:
Support:
- `#pragma` syntax for packing
- `__attribute__()` syntax for packing
- `#pragma` syntax for alignment
- `__attribute__()` syntax for alignment

Depends on https://github.com/tree-sitter/tree-sitter-c/issues/74, https://github.com/tree-sitter/tree-sitter-c/issues/47

PR #2230, opened 2022-01-21
`rz_type_db_struct_member_offset()` changes

GustavoLCR:
# PLEASE DO NOT SQUASH 

**Detailed description**

* Rename `rz_type_db_struct_member_offset()`->`rz_type_db_struct_member_packed_offset()`
* Implement `rz_type_db_struct_member_offset()`

**Test plan**

Tests pass

**Closing issues**

None

### DIFF
--- a/librz/include/rz_type.h
+++ b/librz/include/rz_type.h
@@ -301,7 +301,7 @@ RZ_API void rz_type_path_free(RZ_NULLABLE RzTypePath *tpath);
 RZ_API st64 rz_type_offset_by_path(const RzTypeDB *typedb, RZ_NONNULL const char *path);
 RZ_API RZ_OWN RzList /* RzTypePath */ *rz_type_path_by_offset(const RzTypeDB *typedb, RzBaseType *btype, ut64 offset);
 RZ_API RZ_OWN RzList /* RzTypePath */ *rz_type_db_get_by_offset(const RzTypeDB *typedb, ut64 offset);
-RZ_API ut64 rz_type_db_struct_member_offset(const RzTypeDB *typedb, RZ_NONNULL const char *name, RZ_NONNULL const char *member);
+RZ_API ut64 rz_type_db_struct_member_packed_offset(RZ_NONNULL const RzTypeDB *typedb, RZ_NONNULL const char *name, RZ_NONNULL const char *member);
 
 // Type parser low-level API
 
--- a/librz/include/rz_type.h
+++ b/librz/include/rz_type.h
@@ -302,6 +302,7 @@ RZ_API st64 rz_type_offset_by_path(const RzTypeDB *typedb, RZ_NONNULL const char
 RZ_API RZ_OWN RzList /* RzTypePath */ *rz_type_path_by_offset(const RzTypeDB *typedb, RzBaseType *btype, ut64 offset);
 RZ_API RZ_OWN RzList /* RzTypePath */ *rz_type_db_get_by_offset(const RzTypeDB *typedb, ut64 offset);
 RZ_API ut64 rz_type_db_struct_member_packed_offset(RZ_NONNULL const RzTypeDB *typedb, RZ_NONNULL const char *name, RZ_NONNULL const char *member);
+RZ_API ut64 rz_type_db_struct_member_offset(RZ_NONNULL const RzTypeDB *typedb, RZ_NONNULL const char *name, RZ_NONNULL const char *member);
 
 // Type parser low-level API
 
--- a/librz/type/path.c
+++ b/librz/type/path.c
@@ -258,3 +258,25 @@ RZ_API ut64 rz_type_db_struct_member_packed_offset(RZ_NONNULL const RzTypeDB *ty
 	}
 	return result;
 }
+
+/**
+ * \brief Returns the offset in bytes of the structure member if there is a match
+ *
+ * \param typedb Types Database instance
+ * \param name The structure type name
+ * \param name The structure member name
+ */
+RZ_API ut64 rz_type_db_struct_member_offset(RZ_NONNULL const RzTypeDB *typedb, RZ_NONNULL const char *name, RZ_NONNULL const char *member) {
+	rz_return_val_if_fail(typedb && name && member, 0);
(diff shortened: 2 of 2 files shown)


librz/type/path.c
```diff
@@ -96,7 +96,7 @@ static st64 path_walker(const RzTypeDB *typedb, const char *path) {
 					return -1;
 				}
 			}
-			offset += rz_type_db_struct_member_offset(typedb, parent->identifier.name, member);
+			offset += rz_type_db_struct_member_packed_offset(typedb, parent->identifier.name, member);
```
ret2libc, line 99: Whether to use packed or not packed shouldn't depend on the struct itself?

XVilka:
This is related to https://github.com/rizinorg/rizin/issues/1241
I personally think this change in the meantime is good enough. We can expand further non-packed structures support at a later stage. @thestr4ng3r your thoughts on this?

thestr4ng3r:
I agree with @ret2libc that the choice of packed/not packed depends on the struct itself, so every struct will have exactly one layout and can never be interpreted in multiple ways.
Thus rizin should either:
* not have an explicit `offset` member at all and consider structs always packed (padding can be done by just inserting stub arrays in between)
* not have an explicit `offset` member at all and have rules attached to a struct that can specify the alignment
* use the explicit `offset` for everything, in which case `rz_type_db_struct_member_packed_offset()` should not exist at all. This has the downside that when a type's size changes, this change might have to be propagated to all structs that contain a member of this type.

If you have a concrete use-case for the current changes in the pr, I am ok-ish with merging it as-is, but this should be clarified asap, otherwise it will be a complete mess again.

XVilka:
@thestr4ng3r it's required for https://github.com/rizinorg/rizin/pull/2177

thestr4ng3r (approved):

ret2libc:
> I agree with @ret2libc that the choice of packed/not packed depends on the struct itself, so every struct will have exactly one layout and can never be interpreted in multiple ways. Thus rizin should either:
> 
> * not have an explicit `offset` member at all and consider structs always packed (padding can be done by just inserting stub arrays in between)
> * not have an explicit `offset` member at all and have rules attached to a struct that can specify the alignment
> * use the explicit `offset` for everything, in which case `rz_type_db_struct_member_packed_offset()` should not exist at all. This has the downside that when a type's size changes, this change might have to be propagated to all structs that contain a member of this type.
> 
> If you have a concrete use-case for the current changes in the pr, I am ok-ish with merging it as-is, but this should be clarified asap, otherwise it will be a complete mess again.

Do we have the `packed` attribute stored somewhere for a struct BTW?

thestr4ng3r:
No, and that is kind of the point why this needs to be designed right. Just packed/not packed is probably thought too simple for all use-cases.

librz/type/path.c
```diff
@@ -258,3 +258,25 @@ RZ_API ut64 rz_type_db_struct_member_offset(const RzTypeDB *typedb, RZ_NONNULL c
...
 }
+
+/**
+ * \brief Returns the offset in bytes of the structure member if there is a match
+ *
+ * \param typedb Types Database instance
+ * \param name The structure type name
+ * \param name The structure member name
+ */
+RZ_API ut64 rz_type_db_struct_member_offset(const RzTypeDB *typedb, RZ_NONNULL const char *name, RZ_NONNULL const char *member) {
```
wargio:
```suggestion
RZ_API ut64 rz_type_db_struct_member_offset(RZ_NONNULL const RzTypeDB *typedb, RZ_NONNULL const char *name, RZ_NONNULL const char *member) {
```

librz/type/path.c
```diff
@@ -235,13 +235,13 @@ RZ_API RZ_OWN RzList /* RzTypePath */ *rz_type_db_get_by_offset(const RzTypeDB *
...
 /**
- * \brief Returns the offset of the structure member if there is a match
+ * \brief Returns the packed offset in bits of the structure member if there is a match
  *
  * \param typedb Types Database instance
  * \param name The structure type name
  * \param name The structure member name
  */
-RZ_API ut64 rz_type_db_struct_member_offset(const RzTypeDB *typedb, RZ_NONNULL const char *name, RZ_NONNULL const char *member) {
+RZ_API ut64 rz_type_db_struct_member_packed_offset(const RzTypeDB *typedb, RZ_NONNULL const char *name, RZ_NONNULL const char *member) {
```
wargio:
```suggestion
RZ_API ut64 rz_type_db_struct_member_packed_offset(RZ_NONNULL const RzTypeDB *typedb, RZ_NONNULL const char *name, RZ_NONNULL const char *member) {
```

librz/include/rz_type.h
```diff
@@ -301,6 +301,7 @@ RZ_API void rz_type_path_free(RZ_NULLABLE RzTypePath *tpath);
 RZ_API st64 rz_type_offset_by_path(const RzTypeDB *typedb, RZ_NONNULL const char *path);
 RZ_API RZ_OWN RzList /* RzTypePath */ *rz_type_path_by_offset(const RzTypeDB *typedb, RzBaseType *btype, ut64 offset);
 RZ_API RZ_OWN RzList /* RzTypePath */ *rz_type_db_get_by_offset(const RzTypeDB *typedb, ut64 offset);
+RZ_API ut64 rz_type_db_struct_member_packed_offset(const RzTypeDB *typedb, RZ_NONNULL const char *name, RZ_NONNULL const char *member);
 RZ_API ut64 rz_type_db_struct_member_offset(const RzTypeDB *typedb, RZ_NONNULL const char *name, RZ_NONNULL const char *member);
```
wargio:
```suggestion
RZ_API ut64 rz_type_db_struct_member_packed_offset(RZ_NONNULL const RzTypeDB *typedb, RZ_NONNULL const char *name, RZ_NONNULL const char *member);
RZ_API ut64 rz_type_db_struct_member_offset(RZ_NONNULL const RzTypeDB *typedb, RZ_NONNULL const char *name, RZ_NONNULL const char *member);
```

XVilka:
@GustavoLCR please address @wargio's feedback (and rebase on top of the latest `dev`) then we will merge.

XVilka (approved):

PR #2258, opened 2022-01-27
bf_syscall_write: Use `fflush` and don't print NUL

kazarmy:

**Detailed description**

This pr modifies `bf_syscall_write()` to use ~~`rz_cons`~~ ~~`cb_printf`~~ `fflush` so that its output is not out of order with regard to other command output, and to not print the NUL char to prevent the char from affecting the stdout sink (according to [Aivosto](https://www.aivosto.com/articles/control-characters.html#NUL), nothing should be printed for the NUL char).

**Test plan**

All builds are green.

**Closing issues**

...


librz/analysis/p/analysis_bf.c
```diff
@@ -44,7 +45,9 @@ static void bf_syscall_write(RzILVM *vm, RzILOpEffect *op) {
 	}
 	RzBitVector *bv = rz_il_vm_mem_load(vm, 0, ptr_val->data.bv);
 	ut32 c = rz_bv_to_ut32(bv);
-	putchar(c);
+	if (c) {
+		rz_cons_printf("%c", c);
+	}
```
ret2libc: I think analysis plugins (nor even analysis module itself) should have a dependency on rz_cons.

ret2libc (changes requested):

librz/analysis/p/analysis_bf.c
XVilka: Should not?

ret2libc: Should not, ofc :D

kazarmy: Ok I'll see how I can pass a callback to the plugin

ret2libc: Why the analysis plugin has to write stuff on the console?

thestr4ng3r: It's part of brainfuck emulation, which has a print instruction.

ret2libc: :( I see. What would be a solution then? Just depend on rz_cons?

kazarmy: 52b042706cead4b5029f4c3f31c1e7bbaf17fa49 is a solution

ret2libc: Kinda ok for now, though ideally I'd love to have RzAnalysis, RzAsm, RzFlag, etc. without that "cb_printf" callback...  Not sure if it will be fully possible.

librz/analysis/p/analysis_bf.c
```diff
@@ -36,6 +36,8 @@ static void bf_syscall_read(RzILVM *vm, RzILOpEffect *op) {
 	rz_bv_free(bv);
 }
 
+static PrintfCallback cb_printf = NULL;
```
wargio: tbh, i really do not like this...

kazarmy: okay dded0c2f7faa420e246f1a5a16af8c004e969099 works too and is simpler (and has flushing)

ret2libc:
I honestly don't know... @wargio @thestr4ng3r @kazarmy feel free to choose what's best without me, I'm not sure what would be a good option here.

thestr4ng3r (approved):
I would just use the simplest possible solution, which I think this is now.

XVilka (approved):
Same, I think `putchar()` is the lesser evil than the rest of alternatives. It's specific only to the Brainfuck anyway.

kazarmy:
I think the `putchar()` option is the best because the flushing is done immediately. For the `rz_cons` and `cb_printf()` options, it's unclear when the flushing will happen, other than it happens in time for the test to work.

PR #2259, opened 2022-01-27
Use `rz_file_path_join` in rz-find to avoid duplicated dir separator

PeiweiHu:
This commit contains two parts. One is adding functions
  used to trim specific char in str. The other one is a
  fix to remove redundant slash in tail.

 

**Detailed description**

While using `rz-find -s abc /bin/`, the output will contain duplicate slash, like `/bin//ls`. 

...

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/librz/main/rz-find.c
+++ b/librz/main/rz-find.c
@@ -322,7 +322,7 @@ static int rzfind_open_dir(RzfindOptions *ro, const char *dir) {
 			if (*fname == '.') {
 				continue;
 			}
-			fullpath = rz_str_newf("%s" RZ_SYS_DIR "%s", dir, fname);
+			fullpath = rz_file_path_join(dir, fname);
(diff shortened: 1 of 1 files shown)


ret2libc (changes requested):
I think we don't need to do all this, because the problem is that the directory is just concatenated to the filename in `rzfind_open_dir`. I suggest to replace `rz_str_newf` there with `rz_file_path_join`.

PeiweiHu:
> I think we don't need to do all this, because the problem is that the directory is just concatenated to the filename in `rzfind_open_dir`. I suggest to replace `rz_str_newf` there with `rz_file_path_join`.

Totally agree. I just saw that API......

XVilka (approved):

ret2libc (approved):

PR #2241, opened 2022-01-23
[RzIL] Add lifter for 32-bit Arm, including thumb

thestr4ng3r:
# SQUASH THIS

PR #2261, opened 2022-01-28
Enable specifying exact reg bindings in RzAnalysisILConfig

thestr4ng3r:

**Detailed description**

Cherry-pick from #2241 
ARM has register aliases of the same size, so the current `rz_il_reg_binding_derive()` just picks one of them (which one is unspecified). Trying to make automatic derivation account for this would become too awkward, so it's better to just specify the binding explicitly in such cases, which plugins can do now.

### DIFF
--- a/librz/il/il_reg.c
+++ b/librz/il/il_reg.c
@@ -149,7 +149,7 @@ err:
 
 /**
  * Create a new binding that binds exactly the given register names, querying \p reg for any additionally needed info
- * \param regs array of \p regs_count names of registers. Each of these must be part of \p reg.
+ * \param regs array of \p regs_count names of registers. Each of these must be part of \p reg and they must not overlap.
  */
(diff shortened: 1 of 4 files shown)


librz/il/il_reg.c
```diff
@@ -158,27 +158,47 @@ RZ_API RzILRegBinding *rz_il_reg_binding_exactly(RZ_NONNULL RzReg *reg, size_t r
...
 		rb->regs[i].size = ri->size;
+		items[i] = ri;
 	}
+	free(items);
 	return rb;
 err_regs:
 	for (size_t i = 0; i < regs_count; i++) {
 		reg_binding_item_fini(&rb->regs[i], NULL);
-	}
+	};
```
XVilka: Is this really necessary?

XVilka (approved):

librz/il/il_reg.c
thestr4ng3r: no

PR #2266, opened 2022-01-29
Fix #2217 - Use subprocess API in `rz_sys_cmd_str_full()` & extras

### LINKED ISSUE #2217: Setting `scr.pager=less` on Windows causes rizin to hang
### Work environment

| OS/arch/bits (mandatory)                             | Windows 10.0.19044.0 x64
| File format of the file you reverse (mandatory)      | PE
| Architecture/bits of the file (mandatory)            | x64
| `rizin -v` full output, **not truncated** (mandatory)  | rizin 0.3.4 @ windows-x86-64 commit: 2373c53880e017be86afed5454bca5b5017b920e, build: Fri 01/14/2022__18:38:44.79

### Expected behavior
Setting `e scr.pager=less` or `e scr.pager='less -R'` should work.

### Actual behavior
It causes rizin to hang on the next command that tries to use the pager.

1. Install `less` on Windows via `scoop` or this binary (Which is what scoop installs): https://github.com/jftuga/less-Windows/releases/download/less-v590/less.exe (Adding it to a folder on the `PATH`)
2. Start rizin on any binary.
3. Set `e scr.pager=less` or `e scr.pager='less -R'`.
4. Run any command that will use the pager, e.g. `pd`.
5. rizin hangs.

rizin-bug

GustavoLCR:

**Detailed description**

* Use Subprocess API in `rz_sys_cmd_str_full()`
* Save/restore console mode in subprocess API on Windows 
* Fix starting a process without an ".exe" at the end on Windows
* Support long paths on Windows with subprocess API
* Remove unused rz_sys_cmd_str_full_w32()

**Test plan**

Tests pass

**Closing issues**

Fixes #2217


XVilka:
1. Could you please resend the PR from the branch inside this repo and start from `dist-`? This will ensure subprocess changes work everywhere.
2. Please fix the following warning:
```
../librz/util/sys.c: In function ‘rz_sys_cmd_str_full’:
../librz/util/sys.c:594:41: error: pointer targets in passing argument 2 of ‘rz_subprocess_stdin_write’ differ in signedness [-Werror=pointer-sign]
  594 |   rz_subprocess_stdin_write(subprocess, input, strlen(input) + 1);
      |                                         ^~~~~
      |                                         |
      |                                         const char *
In file included from ../librz/include/rz_util.h:79,
                 from ../librz/util/sys.c:35:
../librz/include/rz_util/rz_subprocess.h:89:73: note: expected ‘const unsigned char *’ but argument is of type ‘const char *’
   89 | RZ_API ssize_t rz_subprocess_stdin_write(RzSubprocess *proc, const ut8 *buf, size_t buf_size);
      |     
```

ret2libc:
Thanks for this! <3

ret2libc:
@segevfiner would you care to test this as well?

segevfiner:
Seemed to work using just a basic `e scr.pager=less` (I have `LESS=-FR --mouse`, set globally).

There is a cosmetic issue of the ANSI color of `call` instructions somehow leaking to the next line:
screenshot
Which seems to only happen with `e scr.pager=less`.

librz/util/subprocess.c
```diff
@@ -19,6 +19,9 @@ struct rz_subprocess_t {
 };
 
 static volatile long pipe_id = 0;
+static DWORD mode_stdin;
+static DWORD mode_stdout;
+static DWORD mode_stderr;
```
wargio, line 24: i'm not a big fan of global variables. do you think is possible to avoid this and have local ones? after all these are only used when subprocess is executed.

ret2libc: There are in general some global vars in this file that can't be removed because they are functional to the way rzsubprocess works. However, it remains to see whether these vars could be part of the `rz_subprocess_t` instead of globals. That I don't know.

thestr4ng3r: I would say if the variable contains state directly related to the global state of the OS running the process, then it is fine to have global vars since only one instance of them can be valid anyway. But depends on the case.

GustavoLCR: It wouldn't make sense to put this in `rz_subprocess_t` as it should be saved/restore with `rz_subprocess_init/fini()`, as you can spawn n number of processes in between these calls and any subprocess could mess with the console state.
We could make `rz_subprocess_init()` return a struct with the initial process state and have `rz_subprocess_fini()` take that as an argument and restore it, but is it really a better design if it ends up modifying global state anyways?

ret2libc: It's ok as globals then ;)

XVilka (approved):

XVilka:
@GustavoLCR please resend this PR from `dist-`, we will merge if green.

XVilka:
I checked with https://github.com/rizinorg/rizin/pull/2311 and it's green, except unrelated bug of https://github.com/rizinorg/rizin/issues/2313
Merging.

PR #2274, opened 2022-02-01
Remove `-b32` from bf rzil tests

kazarmy:

**Detailed description**

This pr just removes `-b32` from the `bf` rzil tests. I don't think there's any good reason to have a 32-bit `bf` machine.

**Test plan**

All builds are green.

**Closing issues**

...

### DIFF
--- a/test/db/rzil/bf
+++ b/test/db/rzil/bf
@@ -1,6 +1,6 @@
 NAME=hello world
 FILE=bins/bf/hello-ok.bf
-ARGS=-b32 -eio.cache=1
+ARGS=-eio.cache=1
 CMDS=<<EOF
 e asm.arch=bf
 e asm.bytes=true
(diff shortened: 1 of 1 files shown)


test/db/rzil/bf
```diff
@@ -1,6 +1,6 @@
 NAME=hello world
 FILE=bins/bf/hello-ok.bf
-ARGS=-b32 -eio.cache=1
+ARGS=-eio.cache=1
```
wargio, line 3: is even io.cache needed? i think not.

kazarmy: currently it's needed ... it shouldn't be needed though

thestr4ng3r: It is needed for the emulation. The alternative is mapping a new malloc.

kazarmy: Well if `oml` says the memory area is writable then i think it should be writable no matter the value of `io.cache`

wargio (approved):

PR #2275, opened 2022-02-02
Add rzil bf "repeat-emu-run" test

kazarmy:

**Detailed description**

This pr adds rzil `bf` tests that:

1. Run emulation on a `bf` file twice, the 2nd time after resetting both registers and memory.
2. [Broken] Run emulation with a `pc` that is not at initial location.

**Test plan**

All builds are green.

**Closing issues**

...

### DIFF
--- a/test/db/rzil/bf
+++ b/test/db/rzil/bf
@@ -1727,3 +1727,44 @@ EXPECT=<<EOF
 7
 EOF
 RUN
+
+NAME=repeat emu run after reset everything
+FILE=bins/bf/2+5.bf
+CMDS=<<EOF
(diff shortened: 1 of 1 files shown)


thestr4ng3r (approved):
As a hint, `aezv` in most cases is only useful for debugging since before stepping it is overwritten by the analysis register state and after stepping it is copied back to registers, so usually `ar` is the state to use.

kazarmy:
Ok got it (i think). Probably some more consistency is still needed to prevent confusion and perhaps `aezi` (and `aei`) shouldn't clobber the `pc` register. Thanks for the hint!

PR #2306, opened 2022-02-12
Fix pcalign after changing analysis bits

thestr4ng3r:

**Detailed description**
This fixes a bug that showed up after 2bdec10f98b3dc9dbb7f5855010b091dad1f4f13
because the bits value was set after cpu, but pcalign was already 4 then
and not updated by bits for example in:
```
rz-asm -a arm -b 16 -o 0x1059e -A dff81000
```

**Test plan**

#2241 contains test cases that cover this. Adding them separately does not really make much sense since this is very implementation-specific. The rz-test test that is added in this pr succeeds both before and after the code change because core does a lot more stuff when switching, so it's only for testing the thumb/arm pcalign.

### DIFF
--- a/librz/analysis/analysis.c
+++ b/librz/analysis/analysis.c
@@ -279,6 +279,8 @@ RZ_API bool rz_analysis_set_bits(RzAnalysis *analysis, int bits) {
 		if (analysis->bits != bits) {
 			bool is_hack = is_arm_thumb_hack(analysis, bits);
 			analysis->bits = bits;
+			int v = rz_analysis_archinfo(analysis, RZ_ANALYSIS_ARCHINFO_ALIGN);
+			analysis->pcalign = RZ_MAX(0, v);
 			rz_type_db_set_bits(analysis->typedb, bits);
 			if (!is_hack) {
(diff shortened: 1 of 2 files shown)


wargio (approved):

librz/analysis/analysis.c
```diff
@@ -279,6 +279,8 @@ RZ_API bool rz_analysis_set_bits(RzAnalysis *analysis, int bits) {
 		if (analysis->bits != bits) {
 			bool is_hack = is_arm_thumb_hack(analysis, bits);
 			analysis->bits = bits;
+			int v = rz_analysis_archinfo(analysis, RZ_ANALYSIS_ARCHINFO_ALIGN);
+			analysis->pcalign = RZ_MAX(0, v);
```
wargio, line 283: this should be at least 1, not zero.

XVilka (approved):

librz/analysis/analysis.c
thestr4ng3r: 0 means no restrictions at all (which in practice is indeed behaves the same as 1, but the meaning is slightly different). It is checked at all places where it's used so 0 is ok.

PR #2315, opened 2022-02-13
Add an option to provide a custom prefix for portable binaries

yossizap:

**Detailed description**

Provide a way to set a custom prefix in [rz_path_prefix](https://github.com/rizinorg/rizin/blob/dev/librz/util/path.c#L25) for binaries that bundle with rizin.

This change aims to solve [#2897](https://github.com/rizinorg/cutter/issues/2897) - the Cutter build artifact fails to find the themes directory due to changes in [rz_path_prefix](https://github.com/rizinorg/rizin/blob/dev/librz/util/path.c#L25), which only returns the relative prefix for portable binaries that return true for `rz_str_endswith(<pid path>, RZ_SYS_DIR RZ_BINDIR)`. Since Windows Cutter sets `RZ_BINDIR=.` and Linux Cutter runs from the root directory while `RZ_BINDIR=bin`, the original build directory prefix is returned in all cases.

**Test plan**

Verify that [#2897](https://github.com/rizinorg/cutter/issues/2897) is solved by these changes.

**Closing issues**

...

### DIFF
--- a/librz/include/rz_userconf.h.in
+++ b/librz/include/rz_userconf.h.in
@@ -54,6 +54,8 @@
 #define RZ_DATDIR  "@DATADIR@"
 #define RZ_WWWROOT "@WWWROOT@"
 
+#define RZ_BINDIR_DEPTH  @BINDIR_DEPTH@
+
 #define RZ_PLUGINS  "@PLUGINS@"
 #define RZ_DATADIR  "@DATADIR_RZ@"
 #define RZ_SDB      "@SDB@"
--- a/librz/util/path.c
+++ b/librz/util/path.c
@@ -16,21 +16,24 @@ RZ_API RZ_OWN char *rz_path_prefix(RZ_NULLABLE const char *path) {
 #if RZ_IS_PORTABLE
 	char *pid_to_path = rz_sys_pid_to_path(rz_sys_getpid());
 	if (pid_to_path) {
-		char *t = rz_file_dirname(pid_to_path);
+		char *it = rz_file_dirname(pid_to_path);
 		free(pid_to_path);
-		// When rz_path_prefix is called from a unit test or from a
-		// not-yet-instazled rizin binary this would return the wrong path.
-		// In those cases, just return RZ_PREFIX.
-		char *result = NULL;
-		if (rz_str_endswith(t, RZ_SYS_DIR RZ_BINDIR)) {
-			char *r = rz_file_dirname(t);
-			result = rz_file_path_join(r, path);
-			free(r);
+
+		for (int i = 0; i < RZ_BINDIR_DEPTH; i++) {
+			char *tmp = it;
+			it = rz_file_dirname(tmp);
+			free(tmp);
 		}
-		free(t);
-		if (result) {
+
+		// When rz_path_prefix is called from a unit test or from a
+		// not-yet-installed rizin binary this would return the wrong path.
+		// In those cases, just return RZ_PREFIX.
+		if (rz_file_is_directory(it)) {
+			char *result = rz_file_path_join(it, path);
+			free(it);
 			return result;
 		}
+		free(it);
 	}
 #endif
 	return rz_file_path_join(RZ_PREFIX, path);
--- a/meson.build
+++ b/meson.build
@@ -177,11 +177,8 @@ if (b_sanitize_opt.contains('address') or b_sanitize_opt.contains('undefined')) 
   library_linkflags += '-shared-libasan'
 endif
 
-if get_option('portable_prefix') != '' and get_option('portable') == true
-  rizin_prefix = get_option('portable_prefix')
-else
-  rizin_prefix = get_option('prefix')
-endif
+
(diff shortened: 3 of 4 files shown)


ret2libc (commented):
Why can't you just set a different `prefix` value while building cutter?

yossizap:
You can't set prefix to a relative path since it's a meson variable and Rizin has to be installed to the prefix ${CMAKE_CURRENT_BINARY_DIR} as a part of the build process

ret2libc:
> You can't set prefix to a relative path since it's a meson variable 

Relative path to what?

> Rizin has to be installed to the prefix ${CMAKE_CURRENT_BINARY_DIR} as a part of the build process

Where it needs to be installed can be different from the prefix used (DESTDIR can be used, for example).

yossizap:
>Relative path to what?

To the Cutter binary, the folder is portable and it's released as a ZIP. It works for portable Rizin since the executed binary is in BINDIR.

Capture123123112

```
~/workspace/cutter/build (dev*) $ ls
CMakeCache.txt       CPackConfig.cmake        dist                  Rizin-prefix
CMakeFiles           CPackSourceConfig.cmake  Makefile              src
cmake_install.cmake  cutter                   Rizin-Bundled-prefix
```

>Where it needs to be installed can be different from the prefix used (DESTDIR can be used, for example).

But you can't set prefix to `./` even if DESTDIR is different.

ret2libc:
I think the right way to fix this would be to improve the check in `rz_path_prefix`. That check probably assumes that BINDIR is always one level below the top directory of rizin, but that may not be always the case. We should compute the "difference" between the current exe dir and the top dir and calculate the top dir correctly.

I'd prefer to avoid adding yet another build-time option, if possible.

yossizap:
>We should compute the "difference" between the current exe dir and the top dir and calculate the top dir correctly.

How would you suggest to do this? The top dir is portable so you can't store it's path and you can't determine that your binary is in the top dir(Cutter's case) since BINDIR(which is also set to directly install into topdir in Cutter Windows), LIBDIR, INCLUDEDIR, DATADIR can be configured.

My previous solution was this patch but it only works if Rizin dirs are installed in the same folder as the binary. Cutter Linux installs to Rizin-prefix/, this could be changed but it's still limiting for other apps that might bundle rizin compared to the build config.

```diff
        if (pid_to_path) {
                char *t = rz_file_dirname(pid_to_path);
                free(pid_to_path);
-               // When rz_path_prefix is called from a unit test or from a
-               // not-yet-instazled rizin binary this would return the wrong path.
-               // In those cases, just return RZ_PREFIX.
+
                char *result = NULL;
                if (rz_str_endswith(t, RZ_SYS_DIR RZ_BINDIR)) {
                        char *r = rz_file_dirname(t);
                        result = rz_file_path_join(r, path);
                        free(r);
+               } else {
+                       result = rz_file_path_join(t, path);
                }
                free(t);
-               if (result) {
+
+               // When rz_path_prefix is called from a unit test or from a
+               // not-yet-instazled rizin binary rz_file_path_join would return
+               // the wrong path. In those cases, just return RZ_PREFIX.
+               if (rz_file_is_directory(result)) {
                        return result;
                }
+               free(result);
        }
```

Maybe use getenv/setenv or a config variable that Cutter will change?

ret2libc:
> How would you suggest to do this? The top dir is portable so you can't store it's path and you can't determine that your binary is in the top dir(Cutter's case) since BINDIR(which is also set to directly install into topdir in Cutter Windows), LIBDIR, INCLUDEDIR, DATADIR can be configured.

Even if the build is portable you have `RZ_PREFIX` in rz_userconf.h and also `RZ_BINDIR`. I think we can get the result of `rz_sys_pid_to_path` and retrieve the top dir based on the diff between RZ_BINDIR and RZ_PREFIX. The current code in `dev` branch assumes that this diff is 1, but in the cutter case it would be 0.

Two notes:
1. the diff between RZ_BINDIR and RZ_PREFIX could probably be computed statically one time and then re-used for all calls to rz_path_prefix
2. the `if` in rz_path_prefix was really meant for unit tests, which run from ./<build-dir>/test/unit directory, thus they have a very different result for rz_sys_pid_to_path .

ret2libc (commented):
The meson part looks good (please resubmit this PR on a dist-* branch, so we can test if the python line works also on windows).

For the `rz_path_prefix` part I would change to something like:
```
if (pid_to_path endswith lastDirOf(realpath/normalize(rz_file_path_join(prefix, RZ_BINDIR))) {
     for (i.......) {
         .... it = rz_file_dirname(tmp);
     }
     if (rz_file_is_directory(it)) { 
           char *result = rz_file_path_join(it, path);
           ....
     }
}
```

The idea is that when this function is called by programs in <build>/test/unit directory the rest of the code is not executed.

yossizap:
>please resubmit this PR on a dist-* branch, so we can test if the python line works also on windows

Reopened @ #2332

PR #2323, opened 2022-02-14
Convert `as` commands to the rzshell

XVilka:

**Detailed description**

- `asj` now became `aslj`
- `ask` was removed
- `as <name>` became `asn <name>`
- `as`, `asca`, `asc` take now only the number as the argument

I have a question though - should I ignore the name argument in `asca` and `asc` completely, or return it back and allow the argument to be **either** syscall number or syscall name?

**Test plan**

CI is green

**Closing issues**

Partially addresses https://github.com/rizinorg/rizin/issues/1442

### DIFF
--- a/librz/core/cmd/cmd_analysis.c
+++ b/librz/core/cmd/cmd_analysis.c
@@ -296,19 +296,6 @@ static const char *help_msg_agn[] = {
 	NULL
 };
 
-static const char *help_msg_as[] = {
-	"Usage: as[ljk?]", "", "syscall name <-> number utility",
-	"as", "", "show current syscall and arguments",
-	"as", " 4", "show syscall 4 based on asm.os and current regs/mem",
-	"asc[a]", " 4", "dump syscall info in .asm or .h",
-	"asj", "", "list of syscalls in JSON",
-	"asl", "", "list of syscalls by asm.os and asm.arch",
-	"asl", " close", "returns the syscall number for close",
-	"asl", " 4", "returns the name of the syscall number 4",
-	"ask", " [query]", "perform syscall/ queries",
-	NULL
-};
-
 /**
  * \brief Helper to get function in \p offset
  *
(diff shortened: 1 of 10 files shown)


ret2libc:
> I have a question though - should I ignore the name argument in asca and asc completely, or return it back and allow the argument to be either syscall number or syscall name?

I think it's oki to accept just a string and recognize the type of argument, whether number or string.

librz/core/cmd/cmd_analysis.c
```diff
@@ -3942,131 +3841,118 @@ static void cmd_sdbk(Sdb *db, const char *input) {
...
+	if (!sysc) {
+		RZ_LOG_ERROR("Cannot resolve syscall: %" PFMT64d "\n", num);
+		return RZ_CMD_STATUS_ERROR;
+	}
+	rz_cons_println(sysc);
+	free(sysc);
+	return RZ_CMD_STATUS_OK;
+}
+
+static RzCmdStatus syscall_dump(RzCore *core, int argc, const char **argv, bool is_c) {
```
ret2libc: please simplify this function, i think it should be quite straightforward

librz/core/csyscall.c
```diff
@@ -0,0 +1,96 @@
...
+ *
+ * Given the syscall number and address it resolves the syscall
+ * for the selected `asm.arch` and `asm.os` values and print
+ * its arguments.
+ *
+ * \param core RzCore instance
+ * \param n number of the syscall
+ * \param addr address of the syscall
+ */
+RZ_API RZ_OWN char *rz_core_syscall_as_string(RzCore *core, st64 n, ut64 addr) {
```
ret2libc, line 26: this function is such a mess :( i understand you just copy-pasted it, so don't worry

librz/core/cmd/cmd_analysis.c
```diff
@@ -3942,131 +3841,118 @@ static void cmd_sdbk(Sdb *db, const char *input) {
...
+	return sdb_fmt(n > 1000 ? "0x%x" : "%d", n);
+}
+
+RZ_IPI RzCmdStatus rz_analysis_syscall_print_handler(RzCore *core, int argc, const char **argv, RzCmdStateOutput *state) {
 	RzListIter *iter;
-	RzList *list;
-	RzNum *num = NULL;
-	int n;
+	RzSyscallItem *si;
+	RzList *list = rz_syscall_list(core->analysis->syscall);
```
ret2libc, line 3016: move this to a separate function in csyscall.c imo. you avoid having syscallNumber function both here and there. Also, move the other function that uses syscallNumber there.

librz/core/csyscall.c
XVilka: And it seems not working, `rz_core_syscall_as_string(core, 4, 0x0)` returns `read(0)`, not `stat()`.

librz/core/cmd_descs/cmd_analysis.yaml
```diff
@@ -1453,3 +1453,49 @@ commands:
...
+        cname: analysis_syscall_name
+        args:
+          - name: name
+            type: RZ_CMD_ARG_TYPE_STRING
+      - name: asr
+        summary: Returns the syscall name by the number
+        cname: analysis_syscall_number
+        args:
+          - name: number
+            type: RZ_CMD_ARG_TYPE_STRING
```
ret2libc: this is string here but i think it is just converted to a number in the handler.

librz/core/cmd_descs/cmd_analysis.yaml
```diff
@@ -1453,3 +1453,49 @@ commands:
...
+            optional: true
+      - name: asl
+        summary: List syscalls
+        cname: analysis_syscall_print
+        type: RZ_CMD_DESC_TYPE_ARGV_STATE
+        default_mode: RZ_OUTPUT_MODE_STANDARD
+        modes:
+          - RZ_OUTPUT_MODE_STANDARD
+          - RZ_OUTPUT_MODE_JSON
+        args: []
```
ret2libc, line 1859: In the past it was possible to list just one syscall. I think we can keep this feature or actually we could even accept an array of strings and return all syscalls matching the array.

librz/core/csyscall.c
wargio: maybe let's open an issue so we track that this does not work.

librz/core/cmd_descs/cmd_analysis.yaml
XVilka: I moved this feature into the `asn` and `asr` commands respectively.

librz/core/cmd/cmd_analysis.c
XVilka: `syscallNumber()` is used in multiple places and is short enough to be duplicated. Further refactoring is out of the scope of this PR, IMHO. A lot of things has to be changed in how syscall are handled.

librz/core/csyscall.c
```diff
@@ -0,0 +1,96 @@
...
+
+/**
+ * \brief Returns the syscall representation as a string
+ *
+ * Given the syscall number and address it resolves the syscall
+ * for the selected `asm.arch` and `asm.os` values and print
+ * its arguments.
+ *
+ * \param core RzCore instance
+ * \param n number of the syscall
```
ret2libc: there is no mention here about what this function does when `n` is `-1` or whether it is even valid (according to the code, it is valid). Please add doc for that so we don't have to read the function to know it's an expected value.

librz/core/cmd/cmd_analysis.c
```diff
@@ -3095,143 +2994,122 @@ RZ_API void rz_cmd_analysis_calls(RzCore *core, const char *input, bool printCom
...
+static RzCmdStatus syscalls_dump(RzCore *core, int argc, const char **argv, bool is_c) {
+	if (argc > 1) {
+		st64 n = rz_num_math(core->num, argv[1]);
+		if (n < 1) {
+			n = rz_syscall_get_num(core->analysis->syscall, argv[1]);
+			if (n == -1) {
+				RZ_LOG_ERROR("Cannot resolve syscall: %s\n", argv[1]);
+				return RZ_CMD_STATUS_ERROR;
 			}
+			rz_cons_printf(".equ SYS_%s %s\n", argv[1], syscallNumber(n));
```
ret2libc, line 3083: use `syscall_dump`

librz/core/cmd/cmd_analysis.c
```diff
@@ -3095,143 +2994,122 @@ RZ_API void rz_cmd_analysis_calls(RzCore *core, const char *input, bool printCom
...
+	if (num < 1) {
+		RZ_LOG_ERROR("Cannot resolve syscall: %s\n", argv[1]);
+		return RZ_CMD_STATUS_ERROR;
+	}
+	rz_cons_println(syscallNumber(num));
+	return RZ_CMD_STATUS_OK;
+}
+
+RZ_IPI RzCmdStatus rz_analysis_syscall_number_handler(RzCore *core, int argc, const char **argv) {
+	st64 num = rz_num_math(core->num, argv[1]);
```
ret2libc: AFAIR we should use `rz_num_math(NULL, ...)` when we don't want resolution of flags, function names, etc. like in this case (it is a _NUM argument, not a _RZNUM).

librz/core/cmd/cmd_analysis.c
XVilka: I can't since it doesn't search the `RzSyscallItem *si` here, just the `char *` string.

ret2libc: actually... can't we just remove these two lines of code and let the code fall through the `rz_syscall_get`, etc. ?

XVilka: @ret2libc at practice the logic between `rz_syscall_get_num()` and `rz_syscall_get()` is a bit different. I would prefer not to touch too much of this code until we decide to refactor the `librz/syscall/*` code which is a mess.

librz/core/csyscall.c
XVilka: Will create an issue once this merged.

ret2libc (approved):

librz/core/csyscall.c
XVilka: https://github.com/rizinorg/rizin/issues/2583

PR #2320, opened 2022-02-14
Fix #2072: Convert main vaddr to paddr first

kazarmy:
**Your checklist for this pull request**
- [X] I've read the [guidelines for contributing](https://github.com/rizinorg/rizin/blob/master/DEVELOPERS.md) to this repository
- [X] I made sure to follow the project's [coding style](https://github.com/rizinorg/rizin/blob/master/DEVELOPERS.md#code-style)
- [ ] I've documented or updated the documentation of every function and struct this PR changes. If not so I've explained why.
- [X] I've added tests that prove my fix is effective or that my feature works (if possible)
- [ ] I've updated the [rizin book](https://github.com/rizinorg/book) with the relevant information (if needed)

**Detailed description**

This pr fixes #2072 by converting the found main virtual address to a physical address first before returning it, as required by the `binsym()` in `librz/bin/p/bin_elf.inc`.

**Test plan**

All builds are green.

**Closing issues**

Closes #2072.

PR #2324, opened 2022-02-15
get_main_offset_linux_64_pie, lea: Apply `maindelta` to vaddr of `entry`

kazarmy:

**Detailed description**

~~This pr follows up on #2320 by making `get_main_offset_linux_64_pie()` return a physical address on `lea`.~~

This pr reworks the `lea` algo in `get_main_offset_linux_64_pie()` so that `maindelta` is applied to the virtual address of `entry`. This is because `maindelta` should only be applied in the virtual address space.

**Test plan**

All builds are green.

**Closing issues**

...

### DIFF
--- a/librz/bin/format/elf/elf_info.c
+++ b/librz/bin/format/elf/elf_info.c
@@ -321,7 +321,7 @@ static ut64 get_main_offset_linux_64_pie(ELFOBJ *bin, ut64 entry, ut8 *buf) {
 			ut64 vmain = (ut64)(entry + bo + maindelta) + 7;
 			ut64 ventry = Elf_(rz_bin_elf_p2v_new)(bin, entry);
 			if (vmain >> 16 == ventry >> 16) {
-				return (ut64)vmain;
+				return Elf_(rz_bin_elf_v2p_new)(bin, vmain);
 			}
 		} else if (ch == 0xc7) { // mov rdi, 0xADDR
 			ut8 *p = buf + bo + 3;
--- a/librz/bin/format/elf/elf_info.c
+++ b/librz/bin/format/elf/elf_info.c
@@ -315,14 +315,15 @@ static ut64 get_main_offset_linux_64_pie(ELFOBJ *bin, ut64 entry, ut8 *buf) {
 	}
 	if (buf[bo] == 0x48) {
 		ut8 ch = buf[bo + 1];
-		if (ch == 0x8d) { // lea rdi, qword [rip-0x21c4]
+		if (ch == 0x8d) { // lea rdi, qword [rip + MAINDELTA]
 			ut8 *p = buf + bo + 3;
 			st32 maindelta = (st32)rz_read_le32(p);
-			ut64 vmain = (ut64)(entry + bo + maindelta) + 7;
 			ut64 ventry = Elf_(rz_bin_elf_p2v_new)(bin, entry);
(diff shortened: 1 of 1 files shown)


ret2libc (commented):
Is `entry` a virtual address?

kazarmy:
No it's physical because it's from `rz_bin_elf_get_entry_offset()` and that function converts `e_entry` from the ELF header into a physical address.

ret2libc:
> No it's physical because it's from `rz_bin_elf_get_entry_offset()` and that function converts `e_entry` from the ELF header into a physical address.

If `entry` is physical, then:

```C
ut64 vmain = (ut64)(entry + bo + maindelta) + 7;
ut64 ventry = Elf_(rz_bin_elf_p2v_new)(bin, entry);
if (vmain >> 16 == ventry >> 16) {
	return Elf_(rz_bin_elf_v2p_new)(bin, vmain);
}
```

how is `vmain` virtual? `vmain = entry + bo + maindelta + 7`, where `entry` is physical. `bo` is a small offset. `maindelta` is a st32, so potentially a small value i guess.

kazarmy:
> ... how is `vmain` virtual?

Hmm, good question.

I've rewritten the `lea` algo (7ade88aaec3bcf00e326863242372319cd168fed) so that `maindelta` is added to the virtual address of `entry`. The reasoning for this change is that `maindelta` is certainly valid in the virtual address space, but not necessarily so in the physical address space.

kazarmy:
Updated https://github.com/rizinorg/rizin/pull/2324#issue-853078315.

ret2libc (commented):
Why are you doing `ventry = p2v(entry); <op>; return v2p(vmain)`? Is it different than just doing the `<op>` directly on `entry`?

kazarmy:
Based on the following example in [`test/bins/elf/analysis/fedora_35_x86_64bit_ls`](https://github.com/rizinorg/rizin-testbins/blob/424e9325cc7e9f71dd77f25610399f0093f228f2/elf/analysis/fedora_35_x86_64bit_ls):

```
$ rizin bins/elf/analysis/fedora_35_x86_64bit_ls
 -- Toggle between disasm and graph with the space key
[0x00006b10]> e asm.bytes=1
[0x00006b10]> e asm.sub.rel=0
[0x00006b10]> pd 1 @ entry0+0x18
            0x00006b28      488d3d51e2ff.  lea   rdi, [rip - 0x1daf]   ; main
                                                                       ; 0x4d80
```

> Why are you doing `ventry = p2v(entry); <op>; return v2p(vmain)`? Is it different than just doing the `<op>` directly on `entry`?

I can think of 2 reasons why:

1. The `lea` instruction is executed in the virtual address space and the target `main` address (`0x4d80`) is a virtual address, so it makes more sense for the Rizin code to reflect this as close as possible.
2. The `lea` instruction assumes that `main` is basically `0x1daf` bytes before it. This must be true in the virtual address space (otherwise execution is not going to the start of `main`) but can be false in the physical address space e.g. it is possible that `entry0` and `main` could be in different `LOAD` segments that are physically very far from each other but very close virtually (this appears unlikely but could happen).

And thus yes it is different theoretically (though maybe not practically).

ret2libc:
You definitely convinced me @kazarmy ;) that's right!

ret2libc (approved):

PR #2328, opened 2022-02-16
Add option to install sigdb systemwise and custom user location for additional signatures

wargio:

**Detailed description**

1. Allows to download the latest sigdb repository and install it in the prefix folder (`meson -Dinstall_sigdb=true build && ninja -C build install`); by default this option is **disabled**.

image

### DIFF
--- a/librz/core/cconfig.c
+++ b/librz/core/cconfig.c
@@ -3797,9 +3797,7 @@ RZ_API int rz_core_config_init(RzCore *core) {
 		"pe", "pilot", "srec", "w32run", "zip", "all", "none", NULL);
 	SETB("flirt.sig.deflate", false, "enables/disables FLIRT zlib compression when creating a signature file (available only for .sig files)");
 	SETI("flirt.node.optimize", RZ_FLIRT_NODE_OPTIMIZE_MAX, "FLIRT optimization option when creating a signature file (none: 0, normal: 1, smallest: 2)");
-	char *home_sigdb = rz_path_system(RZ_SIGDB);
-	SETPREF("flirt.sigdb.path", home_sigdb, "Rizin sigdb location on the filesystem.");
-	free(home_sigdb);
+	SETPREF("flirt.sigdb.path", "", "Additional user defined rizin sigdb location to load on the filesystem.");
 
 	rz_config_lock(cfg, true);
 	return true;
--- a/librz/core/cconfig.c
+++ b/librz/core/cconfig.c
@@ -3797,7 +3797,9 @@ RZ_API int rz_core_config_init(RzCore *core) {
 		"pe", "pilot", "srec", "w32run", "zip", "all", "none", NULL);
 	SETB("flirt.sig.deflate", false, "enables/disables FLIRT zlib compression when creating a signature file (available only for .sig files)");
 	SETI("flirt.node.optimize", RZ_FLIRT_NODE_OPTIMIZE_MAX, "FLIRT optimization option when creating a signature file (none: 0, normal: 1, smallest: 2)");
-	SETPREF("flirt.sigdb.path", "", "Rizin sigdb location on the filesystem.");
+	char *home_sigdb = rz_path_system(RZ_SIGDB);
+	SETPREF("flirt.sigdb.path", home_sigdb, "Rizin sigdb location on the filesystem.");
+	free(home_sigdb);
 
 	rz_config_lock(cfg, true);
 	return true;
--- a/librz/signature/meson.build
+++ b/librz/signature/meson.build
@@ -52,7 +52,10 @@ pkgconfig_mod.generate(rz_sign,
     'rz_type',
     'rz_flag',
   ],
-  description: 'rizin foundation libraries'
+  description: 'rizin foundation libraries',
+  variables: [
+    'datdir=@0@'.format(rizin_datdir_rz),
+  ],
 )
 
 if not is_static_libs_only
--- a/meson.build
+++ b/meson.build
@@ -195,6 +195,7 @@ if host_machine.system() == 'windows'
   rizin_wwwroot = rizin_datdir / 'www'
   rizin_sdb = rizin_datdir / ''
   rizin_zigns = rizin_datdir / 'zigns'
+  rizin_sigdb = rizin_datdir / 'sigdb'
   rizin_themes = rizin_datdir / 'cons'
   rizin_fortunes = rizin_datdir / 'fortunes'
   rizin_flags = rizin_datdir / 'flag'
@@ -206,6 +207,7 @@ else
   rizin_wwwroot = rizin_datdir_rz / 'www'
   rizin_sdb = rizin_datdir_rz
   rizin_zigns = rizin_datdir_rz / 'zigns'
+  rizin_sigdb = rizin_datdir_rz / 'sigdb'
   rizin_themes = rizin_datdir_rz / 'cons'
   rizin_fortunes = rizin_datdir_rz / 'fortunes'
   rizin_flags = rizin_datdir_rz / 'flag'
@@ -224,6 +226,7 @@ opts_overwrite = [
   'rizin_wwwroot',
   'rizin_sdb',
   'rizin_zigns',
+  'rizin_sigdb',
   'rizin_themes',
   'rizin_fortunes',
   'rizin_flags',
@@ -364,6 +367,14 @@ use_ptrace_wrap = use_ptrace_wrap and has_debugger
 message('HAVE_PTRACE: @0@'.format(have_ptrace))
 message('USE_PTRACE_WRAP: @0@'.format(use_ptrace_wrap))
 
+if get_option('install_sigdb')
+  r = run_command(py3_exe, check_meson_subproject_py, 'sigdb')
+  if r.returncode() == 1 and get_option('subprojects_check')
+    error('Subprojects are not updated. Please run `git clean -dxff subprojects/` to delete all local subprojects directories. If you want to compile against current subprojects then set option `subprojects_check=false`.')
+  endif
+  subproject('sigdb', default_options: ['install_dir=@0@'.format(get_option('prefix') / rizin_sigdb)])
+endif
+
 checks_level = get_option('checks_level')
 if checks_level == 9999
   if get_option('buildtype') == 'release'
@@ -395,6 +406,7 @@ userconf.set('DATADIR', rizin_datdir)
 userconf.set('WWWROOT', rizin_wwwroot)
 userconf.set('SDB', rizin_sdb)
 userconf.set('ZIGNS', rizin_zigns)
+userconf.set('SIGDB', rizin_sigdb)
 userconf.set('THEMES', rizin_themes)
 userconf.set('FORTUNES', rizin_fortunes)
 userconf.set('FLAGS', rizin_flags)
@@ -704,6 +716,7 @@ summary({
   'wwwroot': rizin_wwwroot,
   'sdb': rizin_sdb,
   'zigns': rizin_zigns,
+  'sigdb': rizin_sigdb,
   'themes': rizin_themes,
   'fortunes': rizin_fortunes,
   'flags': rizin_flags,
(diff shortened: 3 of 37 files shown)


wargio:
question worth asking is if we should include the sigdb in the rizin release bin.

Maijin:
For me - it's really important to have this part of the default release both for rizin and cutter.

ITAYC0HEN:
I want it! Joining Maijin and think it should be default in Rizin and cutter releases 🤝

XVilka:
I think we could do something like:
- rizin-with-signatures.zip
- rizin-without-zignatures.zip
This way users and distributions will have a choice.

XVilka (approved):
Code changes look OK to me. Not sure about the Meson/installation process, better to wait for @ret2libc opinion.

thestr4ng3r:
We might want to consider either locking this to specific versions on stable, or making concrete releases of the sigdb (probably better) in order to help distributions create packages for this as well.

meson.build
```diff
@@ -358,6 +361,14 @@ use_ptrace_wrap = use_ptrace_wrap and has_debugger
 message('HAVE_PTRACE: @0@'.format(have_ptrace))
 message('USE_PTRACE_WRAP: @0@'.format(use_ptrace_wrap))
 
+if get_option('install_sigdb')
+  r = run_command(py3_exe, check_meson_subproject_py, 'sigdb')
+  if r.returncode() == 1 and get_option('subprojects_check')
+    error('Subprojects are not updated. Please run `git clean -dxff subprojects/` to delete all local subprojects directories. If you want to compile against current subprojects then set option `subprojects_check=false`.')
+  endif
+  subproject('sigdb', default_options: ['install_dir=@0@'.format(get_option('prefix') / rizin_sigdb)])
```
thestr4ng3r: Can the prefix be handled automatically by the subproject?
Something like: `['prefix=@0@'.format(get_option('prefix'), 'install_dir=@0@'.format(rizin_sigdb)]`
Or even better: `['prefix=@0@'.format(get_option('prefix')]` if the sigdb meson knows the right subdirectory itself.

subprojects/packagefiles/sigdb/meson.build
```diff
@@ -0,0 +1,3 @@
+project('sigdb', version: 'main')
+
+install_subdir('.', install_dir : get_option('install_dir'))
```
thestr4ng3r: What about adding this entire meson build system to the sigdb repo?

subprojects/packagefiles/sigdb/meson_options.txt
```diff
@@ -0,0 +1 @@
+option('install_dir', type: 'string', value: '', description: 'Install path for sigdb files')
```
thestr4ng3r: By setting the `value` here to the right directory, installing it with default settings into the prefix would work.

subprojects/packagefiles/sigdb/meson.build
wargio: yes, this can be done.

wargio:
i was thinking that maybe it could be much better to have releases for distros and this can allow us to have easy upgrades for the users (i.e. deb, rpm, etc.. with versioning).

meson.build
wargio: this does make sense to just use the prefix and just let the repo to decide exactly in which subfolder.

wargio:
i have splitted the PR to merge the changes related to flirt separately

ret2libc:
I think the source code used by distros should not include signatures, however the static binaries of rizin should have them. Distros should probably package sigdb as a separate package (a suggested package, for example)

ret2libc (changes requested):
Will review soon, sorry for missing this

Maijin:
While it makes sense on the dev side and on the long run to separate it - I'm really afraid that separating the signature at the package level now, that most users will just don't know about it and think we don't have signature support. The types are included in the source-code which is inconsistent too.

I do think that splitting in small module will make sense when rizin is ready to be fully modular (like can install specific modules etc.) but until then, the distribution install should be complete because that's a really important feature. Maybe the package manager could mark that package as essential dependancy? (which later could be marked as optional somehow when things become more modular)

Maybe a third way / idea would be to provide rizin/cutter users a continuous notification to let them know that when they do `aaa` that FLIRT signature are not present but can be downloaded from their package manager or via the official repository.

ITAYC0HEN:
I agree with @Maijin here (it became a trend or what?). The signatures are too useful to take the chance that some users won't have them (unless users specified they don't want it, or downloaded a 'slim' version)

subprojects/packagefiles/sigdb/meson.build
ret2libc: I agree with moving this to sigdb itself. sigdb should be installable also after rizin was already installed, so i expect to download sigdb and do a `meson build; ninja -C build install` there and work the same way.

meson.build
ret2libc: I think we should export the datadir path in .pc and .cmake files so that "plugins" (not really the usual kind of plugins, but i'd consider sigdb kinda a plugin anyway) can just read the datadir from there and install additional subdirectories/files there directly.

I think sigdb meson file should look something like:
```
rz_sign_dep = dependency('rz_sign')
sigdb_path = get_option('prefix') / rz_sign_dep.get_variable(pkgconfig: 'datdir', cmake: 'rz_sign_DATDIR')
install('.', install_path: sigdb_path)
```

WDYT?

librz/core/cconfig.c
```diff
@@ -3797,7 +3797,9 @@ RZ_API int rz_core_config_init(RzCore *core) {
 		"pe", "pilot", "srec", "w32run", "zip", "all", "none", NULL);
 	SETB("flirt.sig.deflate", false, "enables/disables FLIRT zlib compression when creating a signature file (available only for .sig files)");
 	SETI("flirt.node.optimize", RZ_FLIRT_NODE_OPTIMIZE_MAX, "FLIRT optimization option when creating a signature file (none: 0, normal: 1, smallest: 2)");
-	SETPREF("flirt.sigdb.path", "", "Rizin sigdb location on the filesystem.");
+	char *home_sigdb = rz_path_system(RZ_SIGDB);
+	SETPREF("flirt.sigdb.path", home_sigdb, "Rizin sigdb location on the filesystem.");
+	free(home_sigdb);
```
ret2libc: This way one cannot install sigdb both in user-wide location and system-wide locations (e.g. if we provide a set of signatures, but someone else creates a new signature database that a user wants to install, he may want to install them just for his user).

Can we assume that when flirt.sigdb.path == NULL we use both `rz_path_system(RZ_SIGDB)` and `rz_path_home(RZ_SIGDB)`?

ret2libc (commented):
I think including the binary blobs in the source code is not ideal. Types are in "source form", AFAIK, signatures are binary blobs and they are quite big.

For the static/executable versions of rizin shipped in releases: I agree, we can provide both `-signature` version and `-nosignature` ones. Users can choose which one to take.

For the src tarball, this is mainly used by developers and/or distributions: I think we should work *with* them to make sure they know everything about rizin (what are the best compile options to specify, what's best to exclude, etc.) and in this case we should made it super easy for them to package `sigdb` as a separate package.

As a mid-ground we could provide a `-signature` src tarball for some time, but I'd personally keep the "default" src tarball without any additional binary blob. That said, let's "put it to votes" :)

librz/core/cconfig.c
wargio: maybe i can move the `rz_path_system(RZ_SIGDB)` within the function that parses the sigdb values and use this as user-defined location?

wargio:
> I'd personally keep the "default" src tarball without any additional binary blob. That said, let's "put it to votes" :)

the default tarball will still be without the bin blobs. i think @Maijin and @ITAYC0HEN point is for those users that will download the deb, rpm, static builds of rizin

ret2libc:
> > I'd personally keep the "default" src tarball without any additional binary blob. That said, let's "put it to votes" :)
> 
> the default tarball will still be without the bin blobs. i think @Maijin and @ITAYC0HEN point is for those users that will download the deb, rpm, static builds of rizin

As said, I agree that for deb, RPMs, static builds, etc. I agree 100% to have sigdb included. For the source code, I have some small concerns.

Maijin:
For the source code exposed on Github Release yep I don't see also the point of having the signatures - not lot of users download these hopefully.

librz/core/cconfig.c
ret2libc: Is it possible to load a sigdb from multiple directories and "merge" the results?

wargio: well, is not implemented, but we could have that evar as the user db and the system db hardcoded.

meson.build
wargio: yes, but is `rz_sign_DATDIR` already available?

librz/core/cconfig.c
ret2libc: I was thinking about the possibility of a user to "overwrite" an existing signature (e.g. he wants  his own signature to take precedence over the system one). Is that possible right now?

wargio: nope. but the system will apply all the signatures based on bin type, arch and arch bits

meson.build
ret2libc: no, it isn't right now

subprojects/packagefiles/sigdb/meson.build
wargio: moved to sigdb

meson.build
wargio: i have added rz_sign_DATDIR and i tried manually after installing rizin and does work: https://github.com/rizinorg/sigdb/blob/main/meson.build

wargio: image
Before and after install

librz/core/cconfig.c
wargio: i have added user + system support. essentially the system is loaded where RZ_SIGDB is pointing and adds also the user one when defined.

XVilka (approved):

librz/signature/meson.build
```diff
@@ -52,7 +52,10 @@ pkgconfig_mod.generate(rz_sign,
     'rz_type',
     'rz_flag',
   ],
-  description: 'rizin foundation libraries'
+  description: 'rizin foundation libraries',
```
ret2libc, line 55: I would add this to all .pc files, not only this one. Also, the .cmake files as well. It shouldn't be too big of a change in cmake, hopefully.

ret2libc (approved):

meson_options.txt
```diff
@@ -34,6 +35,7 @@ option('use_sys_libuv', type: 'feature', value: 'auto', description: 'Whether to
 option('use_libuv', type: 'boolean', value: true, description: 'If true, libuv is used to handle remote features')
 option('use_swift_demangler', type: 'boolean', value: true, description: 'If false, disables the swift demangler')
 option('use_gpl', type: 'boolean', value: true, description: 'Set to false when you want to disable gpl code')
+option('install_sigdb', type: 'boolean', value: false, description: 'Downloads and installs rizin sigdb')
```
ret2libc, line 38: shall we have this as `feature`? We can have `auto`, `true`,`false`. Maybe `auto` can default to `true` in case it is a release buildtype, false otherwise? cc @wargio @thestr4ng3r @Maijin @XVilka

ret2libc (changes requested):
`subprojects/sigdb/meson.build:11:2: ERROR: Could not get cmake variable and no default provided for <CMakeDependency rz_sign: True None>`

ret2libc:
> `subprojects/sigdb/meson.build:11:2: ERROR: Could not get cmake variable and no default provided for <CMakeDependency rz_sign: True None>`

The sigdb project should be configured in meson.build *after* defining the rz_sign module, which is under `librz`, otherwise sigdb will look for the system-installed rz_sign.

meson_options.txt
wargio: i think false is enough, no need to have that, because i believe this has to be explicitly defined.

ret2libc (approved):

meson_options.txt
Maijin: depends on how many users are using git version or release these days tbh

PR #2332, opened 2022-02-17
Fix rz_path_prefix not working when BINDIR is at the root directory or several folders deep

yossizap:

**Detailed description**

Fix `rz_path_prefix` not working when BINDIR is at the root directory or several folders deep.

This change aims to solve [#2897](https://github.com/rizinorg/cutter/issues/2897) - the Cutter build artifact fails to find the themes directory due to changes in [rz_path_prefix](https://github.com/rizinorg/rizin/blob/dev/librz/util/path.c#L25), which only returns the relative prefix for portable binaries that return true for `rz_str_endswith(<pid path>, RZ_SYS_DIR RZ_BINDIR)`. Since Windows Cutter sets `RZ_BINDIR=.` and Linux Cutter runs from the root directory while `RZ_BINDIR=bin`, the original build directory prefix is returned in all cases.

**Test plan**

- Check that Rizin is able to load themes with `-Dbindir=multiple/prefixes`
- Check that Rizin is able to load themes with `-Dbindir=.`
- Verify that [#2897](https://github.com/rizinorg/cutter/issues/2897) is solved by these changes.

**Closing issues**

...


XVilka:
```
 "cl" "-Ilibrz\util\rz_util.lib.p" "-I." "-ID:\a\rizin\rizin" "-Ilibrz" "-ID:\a\rizin\rizin\librz" "-Ilibrz\include" "-ID:\a\rizin\rizin\librz\include" "-Isubprojects\sdb" "-ID:\a\rizin\rizin\subprojects\sdb" "-Isubprojects\sdb\src" "-ID:\a\rizin\rizin\subprojects\sdb\src" "-Isubprojects\zlib-1.2.11" "-ID:\a\rizin\rizin\subprojects\zlib-1.2.11" "/MT" "/nologo" "/showIncludes" "/utf-8" "/W2" "/O2" "/Gw" "-DRZ_PLUGIN_INCORE=1" "/Fdlibrz\util\rz_util.lib.p\D__a_rizin_rizin_librz_util_path.c.pdb" /Folibrz/util/rz_util.lib.p/D__a_rizin_rizin_librz_util_path.c.obj "/c" D:/a/rizin/rizin/librz/util/path.c
D:/a/rizin/rizin/librz/util/path.c(225): error C2059: syntax error: 'return'
D:/a/rizin/rizin/librz/util/path.c(226): error C2059: syntax error: '}'
```

librz/util/path.c
```diff
@@ -180,3 +198,29 @@ RZ_API RZ_OWN char *rz_path_home_expand(RZ_NULLABLE const char *path) {
...
+	const char *rp = realpath(path, buf);
+	if (rp) {
+		return strdup(rp);
+	}
+#elif __WINDOWS__
+	char buf[MAX_PATH] = { 0 };
+	char *basename;
+
+	DWORD len = GetFullPathName(path, MAX_PATH, buf, &basename);
+	if (len != 0 && len < MAX_PATH - 1)
```
XVilka: Missing opening `{`, also broken indentation.

XVilka (changes requested):

librz/util/path.c
```diff
@@ -180,3 +198,29 @@ RZ_API RZ_OWN char *rz_path_home_expand(RZ_NULLABLE const char *path) {
...
+	char buf[PATH_MAX] = { 0 };
+	const char *rp = realpath(path, buf);
+	if (rp) {
+		return strdup(rp);
+	}
+#elif __WINDOWS__
+	char buf[MAX_PATH] = { 0 };
+	char *basename;
+
+	DWORD len = GetFullPathName(path, MAX_PATH, buf, &basename);
```
GustavoLCR: Use `GetFullPathNameW()` and use `rz_utf8_to_utf16()` and `rz_utf16_to_utf8()` to convert from `char *` -> `wchar_t *`->`char *`, currently it will break if `UNICODE` is defined, and its better to always support Unicode paths, whether `UNICODE` is defined or not.

GustavoLCR (changes requested):

librz/util/path.c
```diff
@@ -180,3 +199,33 @@ RZ_API RZ_OWN char *rz_path_home_expand(RZ_NULLABLE const char *path) {
...
+	if (rp) {
+		return strdup(rp);
+	}
+#elif __WINDOWS__
+	wchar_t buf[MAX_PATH] = { 0 };
+
+	wchar_t *wpath = rz_utf8_to_utf16(path);
+	DWORD len = GetFullPathNameW(wpath, MAX_PATH, buf, NULL);
+	free(wpath);
+	if (len != 0 && len < MAX_PATH - 1) {
```
wargio:
```suggestion
	if (len > 0 && len < MAX_PATH - 1) {
```

librz/util/path.c
```diff
@@ -180,3 +199,33 @@ RZ_API RZ_OWN char *rz_path_home_expand(RZ_NULLABLE const char *path) {
...
+ * references to /./, /../ and extra '/' characters.
+ *
+ * \param path Original file path.
+ * \return New canonicalized absolute path.
+ */
+RZ_API RZ_OWN char *rz_path_realpath(RZ_NULLABLE const char *path) {
+	if (!path) {
+		return NULL;
+	}
+#if __UNIX__
```
ret2libc: Instead of this i would use HAVE_REALPATH computed from meson

yossizap: Also went ahead and added it in file.c

XVilka (approved):

wargio (approved):
wait for ret2libc approval

librz/util/path.c
```diff
@@ -5,6 +5,11 @@
 #include <rz_util/rz_file.h>
 #include <rz_util/rz_sys.h>
 #include <rz_util/rz_str.h>
+#include <rz_util/rz_utf8.h>
+
+// NOTE: This path is for internal use and it's only valid in non-portable binaries,
+// rz_path_bindir should be used to find the prefix during runtime.
+#define RZ_BINDIR_PREFIXED RZ_JOIN_2_PATHS(RZ_PREFIX, RZ_BINDIR)
```
ret2libc: since this seems to be used only in one place, isn't it better to just use `RZ_JOIN_2_PATHS(RZ_PREFIX, RZ_BINDIR)` there instead of this define with 2 lines comment?

meson.build
```diff
@@ -214,6 +214,12 @@ else
   rizin_bindings = rizin_libdir / 'rizin-bindings'
 endif
 
+# Calcualte BINDIR's depth to be able to find the root directory during runtime
+# in portable builds
+py_cmd = 'import os; print(len(os.path.normpath(r"@0@").split(os.sep)) - len(os.path.normpath(r"@1@").split(os.sep)))'.format(rizin_prefix / rizin_bindir, rizin_prefix)
+py_cmd = run_command(py3_exe, '-c', py_cmd)
+bindir_depth = py_cmd.stdout().strip()
+
```
ret2libc, line 222: bindir_depth should be somehow used as well in :
```
if use_rpath
  rpath_exe = '$ORIGIN/../' + get_option('libdir')
  rpath_lib = '$ORIGIN'
  rpath_summary = 'relative'
elif use_rpath_absolute
```

As you can see it assumes that the binaries are in `<top-prefix>/bin`, but this is not true. That should be fixed as well to ensure proper rpath.

ret2libc (commented):
Just these two comments. Anyway, rz_path_prefix logic looks good.

meson.build
yossizap: I had to set the minimum meson version to 0.58.0 for range() support

ret2libc (approved):

XVilka (approved):

PR #2336, opened 2022-02-19
Refactor Arm IT handling and fix for 4-byte ops

thestr4ng3r:

**Detailed description**

The code was duplicated in both asm and anylsis, so now it's in
arm_it.c.
Previously, conditions were applied to N offsets after an IT instruction
using a fixed stride of 2 bytes. This obviously breaks for 4-byte thumb2
instructions, but whenever we find such an instruction, we can check if
there is a condition inside of it and readjust if necessary.

**Test plan**

see the added test

### DIFF
--- a/librz/asm/arch/arm/arm_it.c
+++ b/librz/asm/arch/arm/arm_it.c
@@ -1,0 +1,131 @@
+// SPDX-FileCopyrightText: 2022 Florian Märkl <info@florianmaerkl.de>
+// SPDX-FileCopyrightText: 2013-2018 pancake <pancake@nopcode.org>
+// SPDX-License-Identifier: LGPL-3.0-only
+
+#include "arm_it.h"
+
+typedef union arm_cs_itblock_t {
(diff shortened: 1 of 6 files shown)


librz/asm/arch/arm/arm_it.c
```diff
@@ -0,0 +1,130 @@
...
+}
+
+/**
+ * Signal that a non-IT instruction was disassembled and clear any block at the same address.
+ */
+RZ_API void rz_arm_it_update_nonblock(RzArmITContext *ctx, cs_insn *insn) {
+	rz_return_if_fail(ctx && insn);
+	bool found;
+	ArmCSITBlock block = { .packed = ht_uu_find(ctx->ht_itblock, insn->address, &found) };
+	if (found) {
```
XVilka: Maybe invert this condition?

XVilka (approved):
LGTM.

librz/asm/arch/arm/arm_it.c
thestr4ng3r: done

PR #2351, opened 2022-02-22
Meson 0.59.0

ret2libc:
We are already using `import(..., required: ..)` where `required`
argument was introduced in 0.59.0, so let's update the required version
of meson necessary to build rizin.


XVilka (approved):

wargio:
shouldn't this PR have a branch name like dist- ?

PR #2367, opened 2022-02-27
aezsu: Fix outdated `pc` when checking for completion

kazarmy:

**Detailed description**

This pr fixes a problem with `aezsu` in which the initial `pc` value was outdated when checking whether the `aezsu` address has been reached, thus causing `aezsu` to fail when run twice.

**Test plan**

All builds are green.

**Closing issues**

...

### DIFF
--- a/librz/core/cmd/cmd_analysis.c
+++ b/librz/core/cmd/cmd_analysis.c
@@ -8035,7 +8035,7 @@ RZ_IPI RzCmdStatus rz_il_vm_step_until_addr_handler(RzCore *core, int argc, cons
 	}
 	RzILVM *vm = core->analysis->il_vm->vm;
 
-	ut64 pc = rz_bv_to_ut64(vm->pc);
+	ut64 pc = rz_reg_get_value_by_role(core->analysis->reg, RZ_REG_NAME_PC);
 	while (pc != address) {
 		if (rz_cons_is_breaked()) {
 			rz_cons_printf("CTRL+C was pressed.\n");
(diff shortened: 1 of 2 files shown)


librz/core/cmd/cmd_analysis.c
```diff
@@ -8035,7 +8035,7 @@ RZ_IPI RzCmdStatus rz_il_vm_step_until_addr_handler(RzCore *core, int argc, cons
 	}
 	RzILVM *vm = core->analysis->il_vm->vm;
 
-	ut64 pc = rz_bv_to_ut64(vm->pc);
+	ut64 pc = rz_reg_get_value_by_role(core->analysis->reg, RZ_REG_NAME_PC);
```
thestr4ng3r: What about using `rz_reg_get_value_by_role()` for checks after stepping too? Maybe even making it a `while (1)` with explicit break to avoid writing it twice.

kazarmy: okay 5653db1d2eb221a6c9219600965db0bb225c1176 ... not sure which is more expensive, `rz_reg_get_value_by_role()` or  `rz_bv_to_ut64()`

thestr4ng3r: That won't matter since stepping at least syncs back and forth, which is way more expensive anyway. But as `rz_core_il_step()` seen from the outside operates on the `RzReg` and `vm->pc` is more just an internal side-product, this is actually "correcter".

thestr4ng3r (approved):

PR #2376, opened 2022-03-03
Refactoring for rz_print_2bpp_* and ps<cmd>

### LINKED ISSUE #2338: pswj does not return 16-bit wide string as expected
### Work environment

| OS/arch/bits (mandatory)                             | Arch Linux x86_64
| File format of the file you reverse (mandatory)      | Any
| Architecture/bits of the file (mandatory)            | Any
| `rizin -v` full output, **not truncated** (mandatory)  | rizin 0.4.0-git @ linux-x86-64 commit: 102549102d508b24c226a67d3a1152fe360f64a3, build: 2022-02-19__22:54:58

(xpost from radareorg/radare2#19729)

### Expected behavior

The command should return properly NUL terminated, properly decoded string.

### Actual behavior

The command did not decode the string and return them while ignoring the NUL termination.

```
$ bin/rizin malloc://512
 -- Enable the PAGER with 'e scr.pager=less -R'
[0x00000000]> wx 680065006c006c006f000000
[0x00000000]> psw
hello
[0x00000000]> pswj
{"string":"h\u0000e\u0000l\u0000l\u0000o\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000","offset":0,"section":"unknown","length":256,"type":"wide16"}
[0x00000000]> 
```

wargio:

**Detailed description**

- Moved rz_print_2bpp_ under core since it is used only there
- Moved rz_print_string out of rz_print into rz_str since it just stringify a raw buffer
- rz_str_stringify_raw_buffer now implements correctly urlencode and encoding/decoding of utf8,16,32 (le and be)
- Added `pu0` to print url encoded strings and stop at first nil value.
- Removed RzCoreStringKind and RzCoreString to use RzStrEnc
- Implemented `psm` and `psM` for utf16be and utf32be (newshell only)
- Implemented `psw` and `psW` in newshell
- Uniformed output of `ps` commands with the output of `iz` regarding string types.
- Fix #2338

**Tests**

Added test for `pu`, `puw`, `pu0`, `pswj`, `psm`, `psM`


librz/util/str.c
```diff
@@ -3819,3 +3819,71 @@ RZ_API char *rz_str_version(const char *program) {
...
 
 #undef RZ_STR_PKG_VERSION_STRING
+
+/**
+ * \brief Converts a raw buffer to a printable string based on the selected options
+ *
+ * \param  buf          The buffer to read
+ * \param  len          The length of the buffer
+ * \param  max_wrap_len The max line length when RZ_STR_STRINGIFY_WRAP is set
+ * \param  options      The options flags to follow to generate the output.
```
ret2libc: Could you put `RZ_STR_STRINGIFY_WIDE16_LE` and the others in an enum? I know that the option will still be a int because you want to do `... | ... | ...` but at least you can specify what options is referring to.

librz/util/str.c
```diff
@@ -3819,3 +3819,71 @@ RZ_API char *rz_str_version(const char *program) {
...
+ * \return The stringified raw buffer
+ */
+RZ_API RZ_OWN char *rz_str_stringify_raw_buffer(RZ_NONNULL const ut8 *buf, ut32 len, ut32 options, ut32 max_wrap_len) {
+	rz_return_val_if_fail(buf && len > 0, NULL);
```
ret2libc:
```suggestion
	rz_return_val_if_fail(buf && len >= 0, NULL);
```
I think having len=0 is a valid case when the API should not do anything.

librz/util/str.c
```diff
@@ -3819,3 +3819,71 @@ RZ_API char *rz_str_version(const char *program) {
 }
 
 #undef RZ_STR_PKG_VERSION_STRING
+
+/**
+ * \brief Converts a raw buffer to a printable string based on the selected options
+ *
+ * \param  buf          The buffer to read
+ * \param  len          The length of the buffer
+ * \param  max_wrap_len The max line length when RZ_STR_STRINGIFY_WRAP is set
```
ret2libc: i think this is not very clear. what does it do? (just to understand without me having to look at the implementation :D)

librz/include/rz_util/rz_str.h
```diff
@@ -256,6 +256,15 @@ RZ_API const char *rz_str_sep(const char *base, const char *sep);
 RZ_API const char *rz_str_rsep(const char *base, const char *p, const char *sep);
 RZ_API char *rz_str_version(const char *program);
 
+#define RZ_STR_STRINGIFY_DEFAULT     (0)
```
ret2libc: as said, i think it makes sense to have an enum for this (with the same `1<<x` values as now)

librz/core/cmd/cmd_print.c
```diff
@@ -6070,14 +6128,13 @@ RZ_IPI int rz_cmd_print(void *data, const char *input) {
 			if (l > 0) {
 				int h, w = rz_cons_get_size(&h);
 				int colwidth = rz_config_get_i(core->config, "hex.cols") * 2;
-				core->print->width = (colwidth == 32) ? w : colwidth; // w;
```
ret2libc, line 5840: I see this is used only in very few places... could you double check if `print->width` can be removed completely from RzPrint please?

librz/core/cmd/cmd_print.c
```diff
@@ -4953,6 +4953,67 @@ static void cmd_pxr(RzCore *core, int len, int mode, int wordsize, const char *a
...
+		}
+		if (useColor) {
+			rz_cons_printf(Color_RESET "\n");
+		} else {
+			rz_cons_printf("\n");
+		}
+	}
+}
+
+static inline void core_print_raw_buffer(const ut8 *buffer, ut32 length, ut32 option, ut32 max_wrap_len) {
```
```suggestion
static void core_print_raw_buffer(const ut8 *buffer, ut32 length, ut32 option, ut32 max_wrap_len) {
```
no need for inline, it doesn't do anything here.

ret2libc (commented):
Nice!

librz/core/cmd/cmd_print.c
wargio: i have set it as inline on purpose to avoid duplication since this code is so small.

librz/core/cmd/cmd_print.c
wargio: i will remove it in another PR.

librz/core/cmd/cmd_print.c
```diff
@@ -500,8 +500,8 @@ static const char *help_msg_ps[] = {
 	"psp", "[j]", "print pascal string",
 	"pss", "", "print string in screen (wrap width)",
 	"psu", "[zj]", "print utf16 unicode (json)",
-	"psw", "[j]", "print 16bit wide string",
-	"psW", "[j]", "print 32bit wide string",
+	"psw", "[j]", "print 16bit wide little endian string",
```
XVilka, line 503: I suggest adding also to add BE commands for the sake of symmetry

XVilka: Maybe we could do like this:
- `psw` - 16-bit LE string
- `psm` - 16-bit BE string
- `psW` - 32-bit LE string
- `psM` - 32-bit BE string

XVilka: Maybe we could do like this:
- `psw` - 16-bit LE string
- `psm` - 16-bit BE string
- `psW` - 32-bit LE string
- `psM` - 32-bit BE string

librz/include/rz_util/rz_str.h
```diff
@@ -256,6 +256,18 @@ RZ_API const char *rz_str_sep(const char *base, const char *sep);
...
 RZ_API char *rz_str_version(const char *program);
 
+typedef enum {
+	RZ_STR_STRINGIFY_DEFAULT = 0,
+	RZ_STR_STRINGIFY_ESCAPE_NL = 1 << 0,
+	RZ_STR_STRINGIFY_WRAP = 1 << 1,
+	RZ_STR_STRINGIFY_STOP_AT_NIL = 1 << 2,
+	RZ_STR_STRINGIFY_URLENCODE = 1 << 3,
+	RZ_STR_STRINGIFY_WIDE16_LE = 1 << 4,
+	RZ_STR_STRINGIFY_WIDE32_LE = 1 << 5,
```
XVilka: Add BE please

XVilka (changes requested):

XVilka:
Probably also worth to fix this in the same PR https://github.com/rizinorg/rizin/issues/2338

librz/core/cmd/cmd_print.c
wargio: yeah, but the output is not correct anyway and needs to be implemented. this should be done in another PR

librz/include/rz_util/rz_str.h
wargio: i did not write the LE code, i kept the old one but is not complete and supports only ascii values which is wrong.

librz/core/cmd/cmd_print.c
wargio: implemented `psm` and `psM`

librz/core/cmd/cmd_print.c
ret2libc: It doesn't work that way AFAIK. `static` is enough to make compiler do its magic. `inline` is useful only in `.h` files (again, AFAIK). @thestr4ng3r cc

wargio: ok, i will remove it then.

librz/core/cmd_descs/cmd_print.yaml
```diff
@@ -101,3 +101,43 @@ commands:
...
+    cname: print_utf32le
+    modes:
+      - RZ_OUTPUT_MODE_STANDARD
+      - RZ_OUTPUT_MODE_JSON
+    args:
+      - name: type
+        type: RZ_CMD_ARG_TYPE_NUM
+        optional: true
+  - name: psm
+    summary: Print buffer as a utf16le string
```
XVilka: `utf16be`

librz/core/cmd_descs/cmd_print.yaml
```diff
@@ -101,3 +101,43 @@ commands:
...
+    cname: print_utf16be
+    modes:
+      - RZ_OUTPUT_MODE_STANDARD
+      - RZ_OUTPUT_MODE_JSON
+    args:
+      - name: type
+        type: RZ_CMD_ARG_TYPE_NUM
+        optional: true
+  - name: psM
+    summary: Print buffer as a utf32le string
```
XVilka: `utf32be`

librz/util/utf8.c
```diff
@@ -598,7 +598,7 @@ RZ_API int rz_utf8_strlen(const ut8 *str) {
 	return len;
 }
 
-RZ_API int rz_isprint(const RzRune c) {
+RZ_API bool rz_rune_is_printable(const RzRune c) {
```
XVilka, line 607: Add Doxygen please

librz/core/cmd/cmd_print.c
```diff
@@ -4483,64 +4483,60 @@ static inline int cmd_pxb_k(const ut8 *buffer, int x) {
...
-		kindstr = "utf32";
-		break;
-	default:
-		kindstr = "unknown";
-		break;
+	return section_name;
+}
+
+static void print_json_string(RzCore *core, const ut8 *block, ut32 len, RzStrEnc encoding, bool stop_at_nil) {
+	const char *kindstr = rz_str_enc_as_string(encoding);
```
XVilka: could be moved later in the code

XVilka (changes requested):
LGTM apart from these minor things

librz/core/cmd/cmd_print.c
wargio: where you want me to move this?

librz/core/cmd_descs/cmd_print.yaml
wargio: good catches!

librz/core/cmd/cmd_print.c
XVilka: After the next check, so the unnecessary execution of that line could be avoided. Miniscule but free optimization ;)

XVilka (approved):

PR #2385, opened 2022-03-06
cons: Show last line of output even if it has no `\n`

kazarmy:

**Detailed description**

This pr makes the rz_cons prompt show the last line of command output even if it has no '`\n`' (newline). The following before/after images illustrate this.

From `rizin bins/bf/2+5.bf`:

__Before:__

before-nonl_last_line

__After:__

after-nonl_last_line

For this to work with `analysis_bf.c`, it needs to call into rz_cons (currently indirectly using `cb_printf`) instead of calling `putchar()` directly.

**Test plan**

All builds are green.

**Closing issues**

...

### DIFF
--- a/librz/analysis/p/analysis_bf.c
+++ b/librz/analysis/p/analysis_bf.c
@@ -36,6 +36,8 @@ static void bf_syscall_read(RzILVM *vm, RzILOpEffect *op) {
 	rz_bv_free(bv);
 }
 
+static PrintfCallback cb_printf = NULL;
+
 static void bf_syscall_write(RzILVM *vm, RzILOpEffect *op) {
 	RzILVal *ptr_val = rz_il_vm_get_var_value(vm, RZ_IL_VAR_KIND_GLOBAL, "ptr");
 	if (ptr_val->type != RZ_IL_TYPE_PURE_BITVECTOR) {
@@ -44,9 +46,8 @@ static void bf_syscall_write(RzILVM *vm, RzILOpEffect *op) {
 	}
 	RzBitVector *bv = rz_il_vm_mem_load(vm, 0, ptr_val->data.bv);
 	ut32 c = rz_bv_to_ut32(bv);
-	if (c) {
-		putchar(c);
-		fflush(stdout);
+	if (c && cb_printf) {
+		cb_printf("%c", c);
 	}
 	rz_bv_free(bv);
 }
@@ -154,6 +155,7 @@ RzILOpEffect *bf_rlimit(RzAnalysis *analysis, ut64 addr, ut64 target) {
 }
 
 static RzAnalysisILConfig *il_config(RzAnalysis *analysis) {
+	cb_printf = analysis->cb_printf;
 	RzAnalysisILConfig *cfg = rz_analysis_il_config_new(64, false, 64);
 	cfg->init_state = rz_analysis_il_init_state_new();
 	if (!cfg->init_state) {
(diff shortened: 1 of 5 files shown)


librz/cons/cons.c
```diff
@@ -1023,6 +1025,18 @@ RZ_API void rz_cons_flush(void) {
...
+		const char *p = &I.context->buffer[I.context->buffer_len - 1];
+		for (; *p != '\n' && p > I.context->buffer; p--)
+			;
+		if (*p == '\n') {
+			RZ_FREE(I.context->last_nonl_line);
+			p++;
+		}
+		I.context->last_nonl_line = rz_str_append(I.context->last_nonl_line, p);
+	}
+
```
ret2libc, line 1039: Is this really necessary? Can't we force the output by adding a `rz_cons_printf("\n");` in the `aezsu` command? I think it is better than having rz_cons automagically perform this.

librz/analysis/p/analysis_bf.c
```diff
@@ -36,6 +36,8 @@ static void bf_syscall_read(RzILVM *vm, RzILOpEffect *op) {
 	rz_bv_free(bv);
 }
 
+static PrintfCallback cb_printf = NULL;
```
wargio, line 39: i really do not like this.

wargio:
i have a counter proposal, what if we implement stdout via RzStrBuf ? or even better, how about just read/write from some memory spaces? like add a fake register and have these to hold a circular buffer that is used as stdin/out ?

Thse fake registers can be simpply called stdin and stdout.

librz/cons/cons.c
kazarmy: > Is this really necessary?

Strictly speaking, for `aeszu` yes an extra newline is just an inaccuracy, but suppose a buggy command produces output that is not terminated with a newline. The current code will just make the final line disappear and you'd be lucky if you realize what was happening.

> Can't we force the output by adding a `rz_cons_printf("\n");` in the `aezsu` command?

Yes but if the emulated code does actually print a newline then there will be a double newline. The only way `aezsu` can determine whether a newline is needed or not is to check what the final character of the output is, and for that you need code in rz_cons that is probably similar to the above.

ret2libc: > Yes but if the emulated code does actually print a newline then there will be a double newline.

So be it. If you are emulating a new line, than it must print the newline and that should not have anything to do with the additional newline that we print by defaul imo.

librz/analysis/p/analysis_bf.c
kazarmy: it is decreed that all rizin output must go through rz_cons but i see your comment below

librz/cons/cons.c
kazarmy: Besides, the bash shell works like this and I'm just following bash in this regard.

kazarmy:
> what if we implement stdout via RzStrBuf ?

rz_cons isn't a string buffer?

> or even better, how about just read/write from some memory spaces? like add a fake register and have these to hold a circular buffer that is used as stdin/out ?

this should be done as well, but if you wanted to see output on-the-fly ... but I suppose output like that should be printed on a different terminal other than the rizin one

librz/cons/cons.c
ret2libc: > Besides, the bash shell works like this

What do you mean?

```
ret2libc@fedora:~$ echo -n "Hello"
Helloret2libc@fedora:~$ echo -ne "Hello\n\n"
Hello

ret2libc@fedora:~$ 
```

kazarmy: Exactly!

> So be it.

Ok.

kazarmy:
Closed to pursue alternative similar to https://github.com/rizinorg/rizin/pull/2385#discussion_r820618089.

kazarmy:
Or maybe the same as https://github.com/rizinorg/rizin/pull/2385#discussion_r820618089.

PR #2388, opened 2022-03-07
cmd_open.c: clear previous items when reopen another file

PeiweiHu:

**Detailed description**

Unexpected behavior happens while reopening another file.

reproduce step:
1. `rizin`
2. `o arm1.bin`
3. `o hello-v850e`
4. `s sym.main`

There is no `sym.main` in `hello-v850e`, the address of `sym.main` of `arm1.bin` will be selected. 

BTW, plz check is there any other info needs to be removed while reopening another file. I am not familiar with RzCore structure. 

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/librz/core/cmd/cmd_open.c
+++ b/librz/core/cmd/cmd_open.c
@@ -786,6 +786,9 @@ static RzCmdStatus open_file(RzCore *core, const char *filepath, ut64 addr, int 
 		return RZ_CMD_STATUS_ERROR;
 	}
 
+	// clear the previous flag items
+	rz_flag_unset_all(core->flags);
+
 	core->num->value = cfile->fd;
 	if (addr == 0) { // if no baddr defined, use the one provided by the file
 		addr = UT64_MAX;


ret2libc:
`o` does not "reopen another file", but it just opens another file. So you will have two files loaded in your rizin instance (see `ol` to see the list of opened files, or just `o` if you are not using the dev branch).

If you want to close the current file and open a new one you should do `oc`.
See `o?` for more info. Please note that in any case, reopening of files is still very messy and does  not work perfectly, so fixes are welcome!

(As a side note, fixing things in the command handler is usually not the right approach. If you have an API, like `rz_core_bin_load` that does not behave correctly you should usually fix it directly and not add additional code in the command handler, which should just calls existing APIs).



That said, maybe we should consider if it makes more sense to have `o` reopening a new file instead of opening an additional file. I think it is more common wanting to open a separate file (like when you do "Open file..." in a regular GUI tool) rather than opening another file in the same address space of rizin. WDYT @XVilka @wargio @thestr4ng3r @kazarmy

thestr4ng3r:
I think `o` should stay as-is in the sense of opening an additional file. That is because when you want to "reopen another file", the usual thing to do is to just quit rizin and restart with another file. And if you really do want to do the same thing in a single instance, you can just use `o--; o <file>`.
It might be worth enhancing the command documentation of `o` slightly to express that it opens **another** file.

wargio:
i actually agree with @thestr4ng3r

PeiweiHu:
Thanks sincerely for the detailed explanation.

PR #2392, opened 2022-03-08
aezsu: Always print `\n` in case output has none

kazarmy:

**Detailed description**

Per https://github.com/rizinorg/rizin/pull/2385#discussion_r820618089, this pr makes `aezsu` print an `\n` always. This is so the prompt will not clobber the last line of `aeszu` output if the last line is not terminated by an `\n`.

**Test plan**

All builds are green.

**Closing issues**

...

### DIFF
--- a/test/db/rzil/bf
+++ b/test/db/rzil/bf
@@ -1707,7 +1707,7 @@ FILE=bins/bf/2+5.bf
 CMDS=<<EOF
 e io.cache=1
 aezi
-aezsu $s; ?e
+aezsu $s
 EOF
 EXPECT=<<EOF
 7
@@ -1721,13 +1721,13 @@ e io.cache=1
 w + @ 2
 aezi
 ara+
-aezsu $s; ?e
+aezsu $s
 ar
 ?e --- Reset everything ---
 ara-
 wc-*
 ar
-aezsu $s; ?e
+aezsu $s
 EOF
 EXPECT=<<EOF
 8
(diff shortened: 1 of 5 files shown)


thestr4ng3r (changes requested):
I don't think this is a good idea. Brainfuck printing things is really just a special case. In the general case, emulation has no notion of "printing" at all so `aezsu` should not have to care about this at all. I think it's fine to have broken output for weird code in bf.

test/db/rzil/bf
```diff
@@ -1721,13 +1721,13 @@ e io.cache=1
...
 ara+
-aezsu $s; ?e
+aezsu $s
 ar
 ?e --- Reset everything ---
 ara-
 wc-*
 ar
-aezsu $s; ?e
+aezsu $s
```
ret2libc, line 1730: btw, this was also a possibility no? instead of having rz_cons_newline just add a ?e whenever necessary.
WDYT?

thestr4ng3r: I think putting `?e` is a good solution indeed.

wargio: agreed

XVilka: Angree.

kazarmy: I prefer #2385 over this solution (#2392) since this solution is kind of ugly in terms of user experience, but I'm fine with having to put `?e`. It's just that I realized rather quickly back then that I had to attach `; ?e` to a bare `aezsu`, largely because rz-test doesn't support non-nl-terminated last lines with `<<EOF`. I wonder whether a new Rizin user would have the same realization.

PR #2401, opened 2022-03-10
`!!`: Fix extra `\r` on Windows

kazarmy:

**Detailed description**

This pr prevents the `\r\r\n` sequence from occurring in `!!` output on Windows, by converting `\r\n` to `\n` in the output of the client process.

**Test plan**

All builds are green.

**Closing issues**

...


librz/core/cmd/cmd_system.c
```diff
@@ -191,6 +191,21 @@ static RzCmdStatus system_common_handler(RzCore *core, bool force_rzcons, int ar
 	bool succ = system_exec(core, argc - 1, &argv[1], need_rzcons ? &out : NULL, &length, &ret);
 	rz_cons_sleep_end(bed);
 	if (need_rzcons) {
+#if __WINDOWS__
+		char *src = out;
+		char *dest = src;
+		char *end = out + length;
+		while (src != end) {
+			*dest = *src;
+			if (src[0] == '\r' && src[1] == '\n') {
```
ret2libc: is `src[1]` always within the bounds?

ret2libc (commented):
Why is the \r\r\n happening? what's the problem with that?

librz/core/cmd/cmd_system.c
kazarmy: hmm good catch ceaca2e217cdb516d10ac52e6aa2ca5d75c766d7

kazarmy:
> Why is the \r\r\n happening? what's the problem with that?

On Windows, when `!!` captures the output of a process, the output lines can be terminated with `\r\n` as a result of normal stdio printing ... when rz_cons then tries to print the captured output using `write()` (alias of [`_write()`](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/write?view=msvc-160)), it helpfully converts `\n` into `\r\n`, thus `\r\n` becomes `\r\r\n`[^1]. 




The problem here is that the additional `\r` isn't added by the process but by Windows, and scripts shouldn't really need to treat Windows `!!` output differently from POSIX `!!` output.

[^1]: The relevant line in the `_write()` documentation is:
  "If the file is opened in text mode, each line feed character is replaced with a carriage return-line feed pair in the output."
  It says nothing about converting carriage return characters.

kazarmy:
> ... as a result of normal stdio printing

This is because stdout [defaults to text mode](https://docs.microsoft.com/en-us/cpp/c-runtime-library/text-and-binary-mode-file-i-o?view=msvc-160). What text mode does is stated [here](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/setmode?view=msvc-160#remarks).

ret2libc (approved):

Issue #2125, opened 2021-12-21
Wrong calculation of the string length in Visual Graphs


XVilka:
### Work environment

| Questions                                            | Answers
|------------------------------------------------------|--------------------
| OS/arch/bits (mandatory)                             | Fedora 35 x86_64
| File format of the file you reverse (mandatory)      | -
| Architecture/bits of the file (mandatory)            | -
| `rizin -v` full output, **not truncated** (mandatory)  | rizin 0.2.0-git @ linux-x86-64, package: 0.2.0.1 (ret2libc) commit: c875be9afde5a6eed037249854b8a8759517263c, build: 2021-01-26__18:17:13

Instead of the code points we should calculate grapheme clusters of the Unicode characters to understand how many actual characters are printed on the screen.

### Expected behavior

Calculate the length of the string correctly and draw the graph node border line correctly

### Actual behavior

![image](https://user-images.githubusercontent.com/203261/146886019-867ac65f-2501-43ad-b27f-7bc20a0023a7.png)

### Steps to reproduce the behavior
- `rizin /bin/ls`
- `aaa`
- `s main`
- `afn 🔥sfddsf🔥🔥🔥🔥🔥🔥🔥`
- `VV`

### Additional Logs, screenshots, source code,  configuration dump, ...

I also suggest to use the correct name as well, as mentioned in the below comments:
 - `rz_str_unicode_codepoints()`
 - `rz_str_unicode_graphemes()`

- https://github.com/ziglang/zig/issues/234#issuecomment-905129195
- https://unicode.org/reports/tr29/
- https://github.com/jecolon/ziglyph/blob/main/src/segmenter/Grapheme.zig
- https://git.suckless.org/libgrapheme/files.html (maybe we could use it as a subproject, actually)

XVilka:
@imbillow as part of your debuginfo work, and loading Unicode information from binaries, especially https://github.com/rizinorg/rizin/issues/3196, it will be necessary also to make sure the loaded information can be displayed correctly.

wargio:
already fixed.

PR #2417, opened 2022-03-16
Tokenize and colorize asm strings

### LINKED ISSUE #9: Syntax highlighting is broken
The syntax highlighting for jump instructions is a bit off at the moment.

In case of jump instruction the first few characters get colored green.
This is not what we want in our case because:
- the first character always indicates package position
- the `jump` keyword comes not always first in the instruction and the wrong characters get colored
- instructions like loop get not colored green at all.

Example:
ex1
ex2
ex3
ex4

I haven't checked it, but this needs probably be fixed in the rizin core.

Rot127:
# SQUASH ME
 

**Detailed description**

This is the first implementation for tokenization of asm strings. It adds:
- A general way to split the colorless asm string into tokens.
- Assign special types to each token (e.g. mnemonic, register, separator etc.)
- Color these tokens according its type.
- In case of number tokens: Store the numbers value in the token.

It also cleans up the coloring of the asm strings quite a bit. Additionally it possibly comes in handy in the future to have the asm string nicely split into parts. 

**Still missing:**

- [x] Move token logic into RzAsm
- [x] Replace all usage of current coloring functions with the token based one.
  - [x] vmenus.c::colorize_asm_string()
  - [x] print.c::rz_print_colorize_opcode()
  - [x] print.c::colorize_asm_string()
- [x] Do not colorize if it is turned off.
- [x] Replace `token_val` with `strtoull()`.
  - [x] `0` and `0x0` are not recognized as number.
- [x] Tests.
  - [x] Generic tokenizer.
  - [x] Hexagon custom tokenizer
  - [x] Coloring generic
    - [x] x86
    - [x] arm
  - [x] Coloring Hexagon custom
- [x] Allow to color only specific terms instead of all mnemonic tokens.
- [x] Some documentation is missing.
- [x] Allow custom coloring schemes.

**Test plan**

Added

**Closing issues**

closes https://github.com/rizinorg/rz-hexagon/issues/9

**Examples**

Here are some (out of date) pictures to compare the current coloring vs. the token based coloring.
On the top is the token base coloring. Below that the current way.

<details>
  <summary>Examples</summary>
  
 ### x86
c-x86

### arm
c-arm

### mips
c-mips

### dex
c-dex

### hexagon
c-hexagon

  
</details>


librz/core/disasm.c
```diff
@@ -1172,7 +1172,16 @@ static void ds_build_op_str(RzDisasmState *ds, bool print_color) {
 				core->parser->subrel_addr = killme;
 			}
 		}
-		char *asm_str = colorize_asm_string(core, ds, print_color);
+		bool partial_reset = line_highlighted(ds) ? true : false;
+		char *source = ds->opstr ? ds->opstr : rz_asm_op_get_asm(&ds->asmop);
+		RzAsmColorizeOptions opts = (RzAsmColorizeOptions){ .reset_bg = partial_reset, .hl_addr = 0, .analysis_op_type = ds->analysis_op.type };
```
wargio:
```suggestion
		RzAsmColorizeOptions opts = { 0 };
		opts.reset_bg = partial_reset;
		opts.analysis_op_type = ds->analysis_op.type;
```

librz/core/disasm.c
```diff
@@ -1200,7 +1209,14 @@ static void ds_build_op_str(RzDisasmState *ds, bool print_color) {
 		free(ds->opstr);
 		ds->opstr = strdup(ds->str);
 	} else {
-		char *asm_str = colorize_asm_string(core, ds, print_color);
+		bool partial_reset = line_highlighted(ds) ? true : false;
```
wargio:
```suggestion
		bool partial_reset = line_highlighted(ds);
```

librz/core/disasm.c
```diff
@@ -1200,7 +1209,14 @@ static void ds_build_op_str(RzDisasmState *ds, bool print_color) {
...
+		char *source = ds->opstr ? ds->opstr : rz_asm_op_get_asm(&ds->asmop);
+		RzAsmColorizeOptions opts = (RzAsmColorizeOptions){ .reset_bg = partial_reset, .hl_addr = 0, .analysis_op_type = ds->analysis_op.type };
```
wargio: same as above.

librz/core/disasm.c
```diff
@@ -5794,7 +5810,14 @@ RZ_API int rz_core_print_disasm_instructions_with_buf(RzCore *core, ut64 address
 				ds_build_op_str(ds, true);
 				free(ds->opstr);
 				ds->opstr = strdup(ds->str);
-				asm_str = colorize_asm_string(core, ds, true);
+				bool partial_reset = line_highlighted(ds) ? true : false;
```
wargio:
```suggestion
				bool partial_reset = line_highlighted(ds);
```

librz/include/rz_util/rz_print.h
```diff
@@ -41,6 +42,49 @@ typedef const char *(*RzPrintSectionGet)(void *user, ut64 addr);
...
+/**
+ *  \brief A token of an asm string holding meta data.
+ */
+typedef struct {
+	size_t start; // byte-offset into `str` where this token starts. Must be exactly at a utf-8 codepoint boundary.
+	size_t len; // `str` length of token in bytes.
+	RzAsmTokenType type;
+	union {
+		ut64 number; // Number of RZ_ASM_TOKEN_NUMBER
+		ut16 reg_num; // Register number
```
wargio: what is a `register number` ?

librz/core/disasm.c
```diff
@@ -1172,7 +1172,16 @@ static void ds_build_op_str(RzDisasmState *ds, bool print_color) {
...
-		char *asm_str = colorize_asm_string(core, ds, print_color);
+		bool partial_reset = line_highlighted(ds) ? true : false;
+		char *source = ds->opstr ? ds->opstr : rz_asm_op_get_asm(&ds->asmop);
+		RzAsmColorizeOptions opts = (RzAsmColorizeOptions){ .reset_bg = partial_reset, .hl_addr = 0, .analysis_op_type = ds->analysis_op.type };
+		char *asm_str = rz_print_tokenize_colorize_asm_str(core->print,
+			source,
+			NULL,
+			opts);
+
+		// char *asm_str = colorize_asm_string(core, ds, print_color);
```
XVilka: Remove this one?

librz/core/disasm.c
```diff
@@ -1172,7 +1172,16 @@ static void ds_build_op_str(RzDisasmState *ds, bool print_color) {
 				core->parser->subrel_addr = killme;
 			}
 		}
-		char *asm_str = colorize_asm_string(core, ds, print_color);
+		bool partial_reset = line_highlighted(ds) ? true : false;
```
XVilka: Same here, no need for `?:`.

librz/core/disasm.c
```diff
@@ -1200,7 +1209,14 @@ static void ds_build_op_str(RzDisasmState *ds, bool print_color) {
...
 	} else {
-		char *asm_str = colorize_asm_string(core, ds, print_color);
+		bool partial_reset = line_highlighted(ds) ? true : false;
+		char *source = ds->opstr ? ds->opstr : rz_asm_op_get_asm(&ds->asmop);
+		RzAsmColorizeOptions opts = (RzAsmColorizeOptions){ .reset_bg = partial_reset, .hl_addr = 0, .analysis_op_type = ds->analysis_op.type };
+		char *asm_str = rz_print_tokenize_colorize_asm_str(core->print,
+			source,
+			NULL,
+			opts);
+		// char *asm_str = colorize_asm_string(core, ds, print_color);
```
XVilka: Remove this please

librz/include/rz_util/rz_print.h
```diff
@@ -41,6 +42,49 @@ typedef const char *(*RzPrintSectionGet)(void *user, ut64 addr);
 typedef const char *(*RzPrintColorFor)(void *user, ut64 addr, bool verbose);
 typedef char *(*RzPrintHasRefs)(void *user, ut64 addr, int mode);
 
+typedef enum {
+	RZ_ASM_TOKEN_NONE, // Type which does not fit any else.
```
XVilka: Please use Doxygen `//<` style documentation for the members, here and below.

librz/include/rz_util/rz_print.h
```diff
@@ -41,6 +42,49 @@ typedef const char *(*RzPrintSectionGet)(void *user, ut64 addr);
...
+	RzAsmTokenType type;
+	union {
+		ut64 number; // Number of RZ_ASM_TOKEN_NUMBER
+		ut16 reg_num; // Register number
+	};
+} RzAsmToken;
+
+/**
+ * \brief An tokenized asm string.
+ *
```
XVilka: Could remove this unnecessary empty line here and below.

librz/include/rz_util/rz_print.h
```diff
@@ -193,6 +238,9 @@ RZ_API char *rz_print_json_indent(const char *s, bool color, const char *tab, co
 RZ_API char *rz_print_json_human(const char *s);
 RZ_API char *rz_print_json_path(const char *s, int pos);
 
+// String tokenizing
+RZ_API void rz_print_tokenize_asm_string(RzStrBuf *str, RZ_BORROW RzAsmTokenString *toks, RZ_NULLABLE void *mapping);
+RZ_API char *rz_print_tokenize_colorize_asm_str(RzPrint *p, char *str, RZ_NULLABLE void *mapping, const RzAsmColorizeOptions opt);
```
XVilka: Please add ownership of the output.

librz/util/print.c
```diff
@@ -1518,10 +1526,46 @@ static bool check_arg_name(RzPrint *print, char *p, ut64 func_addr) {
 	return false;
 }
 
-static bool ishexprefix(char *p) {
+static inline bool is_hex_prefix(const char *p) {
 	return (p[0] == '0' && p[1] == 'x');
 }
 
+static inline bool is_hex_postfix(const char c) {
```
XVilka: Please be careful to also properly handle UTF-8 Unicode output, at least maybe a couple unit tests could be helpful here.

librz/util/print.c
```diff
@@ -1872,3 +1916,257 @@ RZ_API void rz_print_rowlog_done(RzPrint *print, const char *str) {
...
+		t->reg_num = (ut16)val;
+	}
+	return t;
+}
+
+static void add_token(RZ_OUT RzAsmTokenString *token_str, const size_t i, const size_t j, const RzAsmTokenType type, const char *token_val, const bool val_is_hex) {
+	RzAsmToken *t = asm_token_create(i, j - i, type, strtoll(token_val, NULL, val_is_hex ? 16 : 10));
+	if (t) {
+		rz_vector_push(token_str->tokens, t);
+	} else {
```
XVilka: You could reverse the condition and handle error first, then one nesting level becomes unnecessary.

librz/util/print.c
```diff
@@ -1872,3 +1916,257 @@ RZ_API void rz_print_rowlog_done(RzPrint *print, const char *str) {
...
+
+/**
+ * \brief Parses an asm string genericly. So only numbers, separators, operations and basic x86 registers
+ * will be recognized as tokens.
+ * Mnemonics will be guessed. Anything else will be marked as an unknown token.
+ *
+ * \param str The plain asm string.
+ * \param toks The parsed string as tokens.
+ */
+static void tokenize_asm_generic(const char *str, RZ_BORROW RzAsmTokenString *toks) {
```
XVilka: Check for NULLs here?

librz/util/print.c
```diff
@@ -1872,3 +1916,257 @@ RZ_API void rz_print_rowlog_done(RzPrint *print, const char *str) {
...
+	bool val_is_hex = false;
+	while (str[i]) {
+		if (!str[j]) {
+			// j reached end of str. Finish last token and return.
+			add_token(toks, i, j, type, token_val, val_is_hex);
+			return;
+		}
+
+		// Alphanumeric tokens
+		if (is_alpha_num(str + j)) {
```
XVilka: Same as I said before - please be careful to handle also UTF-8 Unicode characters.

librz/util/print.c
```diff
@@ -1872,3 +1916,257 @@ RZ_API void rz_print_rowlog_done(RzPrint *print, const char *str) {
...
+	ut8 reset_len = (opt.reset_bg ? sizeof(Color_RESET_NOBG) : sizeof(Color_RESET)) - 1;
+	// mnemonic color
+	const char *mnem_col = rz_print_color_op_type(p, opt.analysis_op_type);
+
+	// Length of colorized string is: length of the black and white string
+	// + color and reset code for each token.
+	size_t num_toks = rz_vector_len(toks->tokens);
+	// Each element of a color code needs 4 characters.
+	size_t size_color = sizeof(RzColor) * 4;
+	size_t out_len = strlen(bw_str) + num_toks * (size_color + sizeof(Color_RESET_NOBG)) + 1;
```
XVilka: Be careful about `strlen()` and Unicode strings.

librz/util/print.c
```diff
@@ -1872,3 +1916,257 @@ RZ_API void rz_print_rowlog_done(RzPrint *print, const char *str) {
...
+ * \brief Converts a not colored asm string into a colored one. First it splits it into tokens, then colorizes it.
+ *
+ * \param p The RzPrint struct.
+ * \param str The plain asm string.
+ * \param mapping A mapping to parse the asm string. If NULL it is parsed by a generic method.
+ * \param opt Options for colorizing.
+ *
+ * \return char* The colorized asm string.
+ */
+RZ_API char *rz_print_tokenize_colorize_asm_str(RzPrint *p, char *str, RZ_NULLABLE void *mapping, const RzAsmColorizeOptions opt) {
```
XVilka: Please add ownership for the output and if nullable for the `str`.

XVilka (changes requested):

librz/core/disasm.c
Rot127: lol. Never program if too tired.

librz/include/rz_util/rz_print.h
Rot127: ~~In case the analysis plugin knows the number of the general register, it can enter it here. E.g. R0 = 0, R31 = 31 etc.
Not sure whether it is useful.~~ Going to remove it until it is needed.

librz/util/print.c
Rot127: Works fine here since strlen() counts the bytes. Which is the behavior wanted.

librz/util/print.c
```diff
@@ -1518,10 +1526,46 @@ static bool check_arg_name(RzPrint *print, char *p, ut64 func_addr) {
 	return false;
 }
 
-static bool ishexprefix(char *p) {
+static inline bool is_hex_prefix(const char *p) {
```
wargio:
```suggestion
static bool is_hex_prefix(const char *p) {
```

librz/util/print.c
```diff
@@ -1518,10 +1526,46 @@ static bool check_arg_name(RzPrint *print, char *p, ut64 func_addr) {
...
+	return (c == 'r' || c == 'e' || c == 'a' || c == 'R' || c == 'E' || c == 'A');
+}
+
+static inline bool is_x86_reg_posfix(const char *c) {
+	return ((c[0] == 'a' || c[0] == 'b' || c[0] == 'c' || c[0] == 'd') && c[1] == 'x') || (c[0] == 's' && c[1] == 'i') || (c[0] == 'd' && c[1] == 'i') || (c[0] == 's' && c[1] == 'p') || (c[0] == 'b' && c[1] == 'p') || (c[0] == 'x');
+}
+
+static bool is_x86_register(const char *c) {
+	return (is_x86_reg_prefix(c[0])) && (is_num(c + 1) || is_x86_reg_posfix(c + 1));
+}
```
wargio: no need for inline

librz/include/rz_util/rz_print.h
wargio: i suggest to replace it with a string pointer and make it more generic?

Rot127: The idea was to allow easy recognition of flag values in the future. So in case those tokens are used for more than coloring (e.g. replacing numbers with the flags) the comparison would not be against strings.

librz/include/rz_util/rz_print.h
```diff
@@ -41,6 +42,47 @@ typedef const char *(*RzPrintSectionGet)(void *user, ut64 addr);
...
 
+typedef enum {
+	RZ_ASM_TOKEN_NONE, //< Type which does not fit any else.
+	RZ_ASM_TOKEN_MNEMONIC, //< Asm mnemonics like: mov, push, lea...
+	RZ_ASM_TOKEN_OPERATOR, //< Arithmetic operators: +,-,<< etc.
+	RZ_ASM_TOKEN_NUMBER, //< Numbers
+	RZ_ASM_TOKEN_REGISTER, //< Registers
+	RZ_ASM_TOKEN_SEPARATOR, //< Brackets, comma etc.
+	RZ_ASM_TOKEN_META, //< Meta information (e.g Hexagon packet prefix, ARM & Hexagon number prefix).
+	RZ_ASM_TOKEN_UNKNOWN, //< Does not fit to any token above.
```
ret2libc: difference with `NONE`?

ret2libc (commented):
could you provide some unit tests for this new API before moving too ahead so that we can get an idea of how it is going to be used in practice?

librz/util/print.c
```diff
@@ -1872,3 +1926,256 @@ RZ_API void rz_print_rowlog_done(RzPrint *print, const char *str) {
...
+
+/**
+ * \brief Parses an asm string genericly. So only numbers, separators, operations and basic x86 registers
+ * will be recognized as tokens.
+ * Mnemonics will be guessed. Anything else will be marked as an unknown token.
+ *
+ * \param str The plain asm string.
+ * \param toks The parsed string as tokens.
+ */
+static void tokenize_asm_generic(const char *str, RZ_BORROW RzAsmTokenString *toks) {
```
ret2libc: i'm not sure it is a good idea to have a generic tokenizer, given that each disassembler/architecture has different rules. sometimes `#` is a comment, other times it may be part of an address IIRC. Maybe the tokenization should be done by the disasm plugins?

cc @XVilka @wargio @thestr4ng3r

XVilka: Yes, I think so. We could keep generic tokenizer meanwhile but mark it as DEPRECATED. Once the we finish the transition of every plugin - it can be removed. What do you think?

Rot127: @thestr4ng3r and I discussed this already a bit on Mattermost. The generic tokenizer works with similar rules as the opcode colorizing function does (see: `print.c::rz_print_colorize_opcode()`).

You are of cause correct. The disasm plugin should be able to customize the token build process.
How it does this needs more consideration. Maybe a list with regex patterns and the associated token type?

wargio: i agree.

librz/include/rz_util/rz_print.h
Rot127: Added an explanation in https://github.com/rizinorg/rizin/pull/2417/commits/d3582abfa1829c9de161c2649c203ea568ff3e77

librz/include/rz_util/rz_print.h
```diff
@@ -41,6 +42,58 @@ typedef const char *(*RzPrintSectionGet)(void *user, ut64 addr);
...
+} RzAsmTokenType;
+
+#define RZ_ASM_NUM_TOKEN_PATTERNS 6 //< Number of token patterns which need its own regex (no None and unknown tokens).
```
ret2libc: Understanding where this comes from and keeping this updated is gonna be hard...

@thestr4ng3r can we do something like `RZ_ASM_TOKEN_UNKNOWN - RZ_ASM_TOKEN_NONE + 1`? Plus, I would put a big comment on top of RzAsmTokenType to say that orders of UNKNOWN and NONE tokens matter.. or something like that anyway. I think it's better than having a random number here.

librz/include/rz_asm.h
```diff
@@ -141,6 +141,7 @@ typedef struct rz_asm_plugin_t {
 	char *(*mnemonics)(RzAsm *a, int id, bool json);
 	const char *features;
 	const char *platforms;
+	RzAsmTokenPattern token_patterns[RZ_ASM_NUM_TOKEN_PATTERNS]; //< All asm token patterns, ordered by priority. Highest priority at index 0.
```
ret2libc: I think this might be a problem because every time we add an enum value we need rebuild of asm plugins (I know we already do in practice, but...). I think it's better to have a function that returns a `const RzAsmTokenPattern*` instead of the data directly here.

@thestr4ng3r @wargio WDYT?

librz/util/print.c
```diff
@@ -1,8 +1,17 @@
 // SPDX-FileCopyrightText: 2007-2020 pancake <pancake@nopcode.org>
 // SPDX-License-Identifier: LGPL-3.0-only
 
+#include "rz_regex.h"
```
ret2libc: use `<...>` instead of `"...."` for this and the following lines.

librz/util/print.c
```diff
@@ -1872,3 +1947,351 @@ RZ_API void rz_print_rowlog_done(RzPrint *print, const char *str) {
...
+ *
+ * \param p The RzPrint struct.
+ * \param str The plain asm string.
+ * \param patterns A patterns to parse the asm string. If NULL it is parsed by a generic method.
+ * \param opt Options for colorizing.
+ *
+ * \return char* The colorized asm string.
+ */
+RZ_API RZ_OWN char *rz_print_tokenize_colorize_asm_str(RzPrint *p, char *str, RZ_NULLABLE RzAsmTokenPattern *patterns, const RzAsmColorizeOptions opt) {
+	RzAsmTokenString *toks = RZ_NEW0(RzAsmTokenString);
```
ret2libc:
```suggestion
	RzAsmTokenString *toks = RZ_NEW0(RzAsmTokenString);
	if (!toks) {
	   return NULL;
	}
```

librz/util/print.c
```diff
@@ -1872,3 +1947,351 @@ RZ_API void rz_print_rowlog_done(RzPrint *print, const char *str) {
...
+	toks->str = rz_strbuf_new(str);
```
ret2libc: check NULL here as well.

librz/util/print.c
```diff
@@ -1872,3 +1947,351 @@ RZ_API void rz_print_rowlog_done(RzPrint *print, const char *str) {
...
+	toks->tokens = rz_vector_new(sizeof(RzAsmToken), NULL, NULL);
```
ret2libc: also here.

librz/util/print.c
```diff
@@ -1872,3 +1947,351 @@ RZ_API void rz_print_rowlog_done(RzPrint *print, const char *str) {
...
+	rz_print_tokenize_asm_string(toks->str, toks, patterns);
+	char *clr_asm = rz_print_colorize_asm_str(p, toks, opt);
+	rz_strbuf_free(toks->str);
```
ret2libc: i think you also need to free toks->tokens and toks itself.

librz/util/print.c
```diff
@@ -1872,3 +1947,351 @@ RZ_API void rz_print_rowlog_done(RzPrint *print, const char *str) {
...
+		tokenize_asm_generic(sbuf, toks);
+	}
+}
+
+/**
+ * \brief Colorizes a tokenized asm string.
+ *
+ * \param p The RzPrint struct. Used to retrieve the color palette.
+ * \param toks The tokenized asm string.
+ * \param opt Options for colorizing. E.g. reset background color, an address to highlight etc.
```
ret2libc, line 1572: I'm not sure what is the purpose of the colorize options structure... Isn't all the color information supposed to be in RzPrint?

librz/util/print.c
```diff
@@ -1872,3 +1947,351 @@ RZ_API void rz_print_rowlog_done(RzPrint *print, const char *str) {
...
+	// mnemonic color
+	const char *mnem_col = rz_print_color_op_type(p, opt.analysis_op_type);
+
+	// Length of colorized string is: length of the black and white string
+	// + color and reset code for each token.
+	size_t num_toks = rz_vector_len(toks->tokens);
+	// Each element of a color code needs 4 characters.
+	size_t size_color = sizeof(RzColor) * 4;
+	size_t out_len = strlen(bw_str) + num_toks * (size_color + sizeof(Color_RESET_NOBG)) + 1;
+	char *out = calloc(1, out_len);
```
ret2libc:
```suggestion
	char *out = RZ_NEWS0(char, out_len);
```

ret2libc: Maybe it's easier to start by using a RzStrBuf instead of computing the length beforehand... I'm worried we might miss something. For example, is the reset_len that is added for each token computed here?

librz/include/rz_asm.h
wargio: RzPVector? but yes, a pointer is better

librz/util/print.c
Rot127: Unfortunately not. Coloring depends on the analysis operation type (jump instructions are bright green, returns are red etc.) and the background color should not reset if a line is highlighted.
This is not known in RzPrint but only by the caller (RzAsm mostly).

librz/util/print.c
Rot127: Oh yes. Indeed it is. I literally missed the `rz_strbuf_append()` for like a year now.

librz/include/rz_asm.h
XVilka: Pointer is fine for me as well :+1:

librz/util/print.c
Rot127: Though what we could do is to move it into RzPrint. Is this better architectural vise?

librz/include/rz_asm.h
thestr4ng3r: The important point is api and abi compatibility. I think a callback like `char *(*token_pattern)(RzAsmTokenType type)` (returning just the pattern because all other members of `RzAsmTokenPattern` are then implicit or generated) can work best.

Rot127: > The important point is api and abi compatibility.

Sorry if I missed the docs about this. But what do you mean with API compatibility?

> `char *(*token_pattern)(RzAsmTokenType type)`

What about `RzList /* RzAsmTokenPattern */ *get_token_patterns()` in https://github.com/rizinorg/rizin/pull/2417/commits/daf7e147f08daac123cbecebec79b1cd367ada8b ?
Sometimes there are more patterns assigned to the same type (see code). So a `type <-> pattern` one to one match is not always possible.

ret2libc: RzList is fine. If we don't want to rely on RzList we can just use a simple array terminated by a NULL element.

librz/util/print.c
ret2libc: If it is something that can be moved into RzPrint and does not required analysis/asm knowledge, then I think it makes sense to have that there directly.

I *think* that it should be the tokenizer that creates different kind of tokens based on whether the instruction is a jump, a return, or whatever. So the printer just needs to know the color for each token type, without really caring about what the token means. I expect the tokenizer will be in the asm/analysis plugins and they know that a `ret` is not just a mnemonic, but it's a RET_MNEMONIC. a `jmp` is not just a mnemonic but it's a JMP_MNEMONIC and so on. WDYT?

Rot127: > If it is something that can be moved into RzPrint and does not required analysis/asm knowledge, then I think it makes sense to have that there directly.

The line highlight info is stored in `RzDisasmState`. But we could refactor this (quick global search gives ~20-30 usages of the relevant struct members). So probably in another PR and rebase this one on top.

> I think that it should be the tokenizer that creates different kind of tokens based on whether the instruction is a jump.

I agree on that and can add this to the custom tokenizer.
But we still have to support asm plugins without token patterns. In this case the coloring function still needs to know the analysis type. Because the generic tokenizer should not scan for hundreds of specific keywords of 15+ archs, just to find out whether it is a `jump`, `ret` or `muleq`.

Or we do add token patterns for each arch (though I would need help then, beacuse I don't have time for all of this currently).
What do you think?

librz/util/print.c
wargio: > @thestr4ng3r and I discussed this already a bit on Mattermost. The generic tokenizer works with similar rules as the opcode colorizing function does (see: `print.c::rz_print_colorize_opcode()`).
> 
> You are of cause correct. The disasm plugin should be able to customize the token build process. How it does this needs more consideration. Maybe a list with regex patterns and the associated token type?

one thing i learned from r2dec/jsdec is that regexes for parsing asm are absolutely a mess.

Rot127: You can checkout Hexagons `librz/asm/p/asm_hexagon.c::get_token_patterns()`. This is the amount of regex patterns currently needed to parse Hexagons asm strings.
The runtime increase is barely existent (+-0.005s) compared to the generic method (going from left to right over the string).

What problems have you run into?

librz/include/rz_asm.h
thestr4ng3r: > Sometimes there are more patterns assigned to the same type (see code). So a `type <-> pattern` one to one match is not always possible.

Would `(<pattern0>)|(<pattern1>)|...` as part of the regex work too for you?

Rot127: No. The patterns need a priority because the same symbols can occur in different types.
E.g. mnemonics can contain digits (`trap0`) and numbers can contain letters (`0xff, 0b10` etc.) and so on.

Since there can be so many different mnemonics, mnemonic regex should match any kind of alphanumeric string (`[[:alnum:]]+`) with exceptions (see below).

Lets assume we have the asm string `trap0(0xff) ; R13 = memw(0x5000 + R13in)`. We have to:
1. Match hex numbers  - `type = number`
2. Match registers - `type = register`
3. Match mnemonics with number (`trap0`). Necessary, otherwise the `0` would be matched by the decimal numbers - `type = mnemonic`
4. Match decimal numbers - `type = number`
5. Match other mnemonics (`memw`) - `type = mnemonic`
6. Match separators - `type = separator`

Simply ORing of regex patterns would unfortunately not work here.
Also think of more complex asm strings like dex. Maybe an actual parser with a grammar for each arch is better?

thestr4ng3r: I see. For such an approach, returning a list from a callback sounds good. A context free grammar is probably strong enough for most archs too. mpc comes to mind as a solution to implement this, but we want to get rid of that dependency.

However I wonder if it wouldn't be better to make the generic parser parameterizeable, but still quite strict, so it only parses things that follow a fixed `<mnemonic> <op>, <op>, ...` structure for example and nothing else. For that, supplying a a more restricted parameter set, like one (optional) pattern per op might actually work. So kind of a mixture of `tokenize_asm_custom()` and `tokenize_asm_generic()` as it is now.
And the alternative if a plugin needs more freedom would be allowing them to do the entire parsing in C by themselves, without forcing them to use regex or anything else. The tokenized string could be returned as part of the regular `op` callback already.

> Sorry if I missed the docs about this. But what do you mean with API compatibility?

In short, for plugins, API compatibility is preserving the ability to compile the same plugin source against updated rizin. ABI compatibility is preserving the ability to use the same already compiled binary plugin inside updated rizin. For example extending the size of an array in the plugin struct may preserve API compatibility (mostly), but break ABI compatibility because the layout is moved around.

Rot127: How about this:

The generic method uses the register profile as parameter.
- If the current asm plugin provides no register profile, it parses the string like this: `<mnemmonic> <op>, <op>` where each non number `<op>` is treated as register.
- With register profile it still parses like `<mnemmonic> <op>, <op>`, but with less false positive register tokens. Every alphanumeric `<op>` which is no register is an unknown.

This way the generic method can be used right away without adding some kind of reduced parameter set for each arch.

Custom method:
- Do the parsing completely in the plugin if `op()` or `disassembly()` is called and save the `RzList*` in `RzAsmOp`/`RzAnalysisOp` (in the future maybe `RzArchOp`?).
- Coloring simply uses the generic method if the `RzList *` is NULL.
- This way @ret2libc remarks regarding keyword highlighting are addressed as well, because the plugin knows the analysis type.

librz/util/print.c
wargio: This is how x86 is parsed: https://github.com/wargio/r2dec-js/blob/master/libdec/arch/x86.js#L2187

librz/include/rz_asm.h
thestr4ng3r: This sounds perfect to me. Parameters could still be added to the generic method, but only if they turn out to really be needed. If it works well enough from just the reg profile, then that is even better of course.

test/unit/test_tokens.c
```diff
@@ -0,0 +1,233 @@
...
+static bool test_rz_tokenize_generic_2(void) {
+	RzStrBuf *asm_str = rz_strbuf_new("mov pc, 🍍");
+	RzAsmToken tokens[5] = {
+		{ .start = 0, .len = 3, .type = RZ_ASM_TOKEN_MNEMONIC, .val.number = 0 },
+		{ .start = 3, .len = 1, .type = RZ_ASM_TOKEN_SEPARATOR, .val.number = 0 },
+		{ .start = 4, .len = 2, .type = RZ_ASM_TOKEN_MNEMONIC, .val.number = 0 },
+		{ .start = 6, .len = 2, .type = RZ_ASM_TOKEN_SEPARATOR, .val.number = 0 },
+		{ .start = 8, .len = 4, .type = RZ_ASM_TOKEN_UNKNOWN, .val.number = 0 }
+	};
+	RzAsmTokenString *toks = rz_asm_token_string_new();
```
ret2libc: Is there any reason to have an "empty" asmtokenString? I think there should be just one constructor that immediately takes the string (essentially, just return the RzAsmTokenString from `rz_print_tokenize_asm_string`, instead of having an empty constructor).

test/unit/test_tokens.c
```diff
@@ -0,0 +1,233 @@
...
+	int i = 0;
+	RzAsmToken *it;
+	rz_vector_foreach(toks->tokens, it) {
+		mu_assert_eq(it->start, tokens[i].start, "Token start");
+		mu_assert_eq(it->len, tokens[i].len, "Token length");
+		mu_assert_eq(it->type, tokens[i].type, "Token type");
+		mu_assert_eq(it->val.number, tokens[i].val.number, "Token value");
+		++i;
+	}
+
```
ret2libc, line 216: what are you really testing here? I think you are testing `rz_print_tokenize_asm_string` and if so, IMHO you should test the bytes written in the rz_cons buffer, not the toks->tokens.

Rot127: In this specific case I checked for UTF-8 characters in the asm string.
These tests are not about coloring yet. So no `rz_cons` involved (please correct me if I'm wrong. Haven't had used `rz_cons` yet.).
I just want to make sure all the indices, size and types of the tokens are correct.

librz/include/rz_asm.h
```diff
@@ -198,6 +201,14 @@ RZ_API int rz_asm_op_set_hexbuf(RzAsmOp *op, const ut8 *buf, int len);
 RZ_API void rz_asm_op_set_buf(RzAsmOp *op, const ut8 *str, int len);
 RZ_API ut8 *rz_asm_op_get_buf(RzAsmOp *op);
 
+// String tokenizing
+RZ_API RZ_OWN RzAsmTokenString *rz_asm_token_string_new(const char *asm_str);
+RZ_API void rz_asm_token_string_free(RZ_OWN RzAsmTokenString *toks);
+RZ_API RZ_OWN RzAsmTokenString *rz_asm_token_string_clone(RZ_OWN RzAsmTokenString *toks);
+RZ_API void rz_asm_token_pattern_free(void *p);
+RZ_API RZ_OWN RzAsmTokenString *rz_asm_tokenize_asm_regex(RZ_BORROW RzStrBuf *asm_str, RzList /* RzAsmTokenPattern */ *patterns);
+RZ_DEPRECATE RZ_API RZ_OWN RzAsmTokenString *rz_asm_tokenize_asm_string(RZ_BORROW RzStrBuf *asm_str, RZ_NULLABLE RzAsmParseParam *param);
```
ret2libc: ?? why deprecated?

librz/include/rz_util/rz_print.h
```diff
@@ -41,6 +42,61 @@ typedef const char *(*RzPrintSectionGet)(void *user, ut64 addr);
...
 typedef char *(*RzPrintHasRefs)(void *user, ut64 addr, int mode);
 
+typedef enum {
+	RZ_ASM_TOKEN_MNEMONIC, //< Asm mnemonics like: mov, push, lea...
+	RZ_ASM_TOKEN_OPERATOR, //< Arithmetic operators: +,-,<< etc.
+	RZ_ASM_TOKEN_NUMBER, //< Numbers
+	RZ_ASM_TOKEN_REGISTER, //< Registers
+	RZ_ASM_TOKEN_SEPARATOR, //< Brackets, comma etc.
+	RZ_ASM_TOKEN_META, //< Meta information (e.g Hexagon packet prefix, ARM & Hexagon number prefix).
+	RZ_ASM_TOKEN_UNKNOWN, //< Does not fit to any token above.
```
ret2libc: Is it better to place this at the beginning of the enum definition and make it ` = 0`? So by default zero-initialized variables will be UNKNOWN.

librz/util/unum.c
```diff
@@ -8,6 +8,14 @@
 
 static ut64 rz_num_tailff(RzNum *num, const char *hex);
 
+RZ_API bool rz_num_is_hex_prefix(const char *p) {
```
XVilka, line 17: Please add Doxygen.

librz/asm/arch/hexagon/hexagon.h
```diff
@@ -106,8 +107,10 @@ typedef struct {
 	char mnem_infix[128]; // The mnemonic without the pre- and postfix.
```
XVilka: Could you please convert other members' comments of this struct to Doxygen format too?

librz/asm/arch/hexagon/hexagon.h
```diff
@@ -135,6 +138,7 @@ typedef struct {
 	RzList *const_ext_l; // Constant extender values.
 	RzAsm rz_asm; // Copy of RzAsm struct. Holds certain flags of interesed for disassembly formatting.
```
XVilka, line 137: Same - please convert other comments to this struct members' to Doxygen.

librz/asm/arch/hexagon/hexagon.h
```diff
@@ -135,6 +138,7 @@ typedef struct {
...
 	RzConfig *cfg;
+	RzList /* RzAsmTokenPattern */ *token_patterns; ///< NULL terminated arry with token patterns.
```
XVilka: A typo. Also, I wonder if using `RzPVector` better for this - likely you wont need to have random insertion/deletion for those, right?

librz/asm/arch/hexagon/hexagon_arch.h
```diff
@@ -42,7 +42,7 @@ typedef struct {
 
 #define HEX_PKT_UNK           "?   "
 #define HEX_PKT_SINGLE        "[   "
-#define HEX_PKT_SINGLE_UTF8   "[     "
+#define HEX_PKT_SINGLE_UTF8   "[   "
```
XVilka, line 45: Could you extract this change and Doxygen comments to the Hexagon structures into a separate PR? We could merge it before the 0.4.0 release even.

librz/asm/asm.c
```diff
@@ -14,6 +18,69 @@
...
+		return false; // UTF-8
+	}
+	return is_num(c) || isalpha(c[0]);
+}
+
+static bool is_separator(const char c) {
+	if (!isascii(c)) {
+		return false; // UTF-8
+	}
+	return (c == '(' || c == ')' || c == '[' || c == ']' || c == '{' || c == '}' || c == ',' || c == '.' || c == '#' || c == ':' || c == ' ');
```
XVilka: In TMS320 series of DSP, `||` is oftenly used to separate two parallel instructions in the block.

``` nasm
mov *(ar1+t0b) << t3, ac2 || mov ac1.l , *ar4
```

currently determined "mov ac1.l, *ar4" instruction as part of "ac2" operand

``` nasm
rptblocal #0x000011 || sfts ac0,  #-1
```

currently determined "sfts" as part of 0x000011 number

``` nasm
cmp t2 == t3, tc1
```

currently determined "t2 ==" as part of instruction keyword "cmp"

``` nasm
xccpart ac0 <=  #0 || mov #0x0, t0
```

currently determined "ac0" as part of "xccpart" instruction keyword, "|| mov #0x0, t0" as part of number #0

librz/asm/asm.c
```diff
@@ -1210,3 +1286,395 @@ RZ_API RzAsmCode *rz_asm_rasm_assemble(RzAsm *a, const char *buf, bool use_spp)
...
+RZ_API void rz_asm_token_string_free(RZ_OWN RzAsmTokenString *toks) {
+	if (!toks) {
+		return;
+	}
+	rz_strbuf_free(toks->str);
+	rz_vector_free(toks->tokens);
+	free(toks);
+}
+
+RZ_API RZ_OWN RzAsmTokenString *rz_asm_token_string_clone(RZ_OWN RzAsmTokenString *toks) {
```
XVilka: Add `RZ_NONNULL` also then

XVilka (changes requested):
Please add also test for strings containing also more complex Unicode strings, see, for example, https://github.com/rizinorg/rizin/issues/2125

librz/asm/arch/hexagon/hexagon_arch.h
Rot127: Unfortunately no. The current coloring function removes some spaces after the `[` or adds some after the UTF-8 brackets (or something similar, can't remember at the moment). If we merge this change now the assembly would be misaligned in visual mode:

```
     ┌     immext(##sym.__cxa_finalize) 
     └     R0 = ##sym.__cxa_finalize 
     [   P0 = cmp.eq(R0,##0x0)
     [   if (P0) jumpr:nt LR   
```

The new coloring method leaves the asm string as is, which makes the deletion of spaces necessary for alignment.

librz/asm/arch/hexagon/hexagon.h
Rot127: I do this in another PR. This way all comments can be updated at once.

librz/asm/asm.c
Rot127: I added `||`, `==` and `<=` to the separator list. But operators and separators are currently printed in the same color anyway.

Rot127:
Please let me know of any asm syntax which should be tested as well. Preferably complex and weird syntax constructs.

@XVilka I excluded TMS320 in the tests because it doesn't have a register profile. Due to that the color of the asm string is mostly white, which needs no testing.

XVilka:
@Rot127 no idea, I think adding TMS320 register profile[s] would be the best solution for this problem, it's the most complex assembly in Rizin and certainly would be the best test case for this new code.

Here is the instruction/architecture description for TMS320C64x ([`librz/analysis/p/analysis_tms320c64x.c`](https://github.com/rizinorg/rizin/blob/dev/librz/analysis/p/analysis_tms320c64x.c)):
- https://www.ti.com/lit/ug/spru732j/spru732j.pdf (ISA description)
- https://www.ti.com/lit/ug/spru395b/spru395b.pdf?ts=1657122578474 (see 2.1.1 C6000 CPU)
- https://github.com/capstone-engine/capstone/blob/master/arch/TMS320C64x/TMS320C64xGenRegisterInfo.inc#L13

Rot127:
This is the how a TM320 asm string looks now:
rz-asm

rz-

If you are fine with it I would wrap this PR up tomorrow.

XVilka:
@Rot127 it could be improved but for the purpose of this PR it's enough, I think.

Rot127:
@XVilka I agree. But TMS320 would also be a prime example where custom parsing would be very useful :)

XVilka:
```
Summary of Failures:
  8/109 rizin:unit / analysis_op                           FAIL            0.25s   exit status 1
 87/109 rizin:unit / tokens                                FAIL            0.38s   killed by signal 11 SIGSEGV
Ok:                 107 
Expected Fail:      0   
Fail:               2   
Unexpected Pass:    0   
Skipped:            0   
Timeout:            0   
```
Please also update the branch.

Rot127:
~~I can not reproduce the sigsegv of the token test.
` 87/109 rizin:unit / tokens                                FAIL            0.38s   killed by signal 11 SIGSEGV`~~

XVilka:
```

[XX] db/analysis/hexagon hexagon continue analysis after unmapped jumps
RZ_NOPLUGINS=1 /usr/bin/rizin -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -Qc 'wx 9852730118c00058824700e010406070
e asm.arch=hexagon
e analysis.arch=hexagon
af
pd 4
' =
-- stdout
--- expected
+++ actual
@@ -0,6 +0,6 @@
 / fcn.00000000 ();
 |           0x00000000      ?       immext(##0x1734a600)
-|       ,=< 0x00000004      ?       jump 0x1734a60c
+|       ,=< 0x00000004      ?       jump 0x1734a610
 |       |   0x00000008      /       R2 = +mpyi(R0,##0x3c)
 |       |   0x0000000c      |       R16 = R0

[XX] db/cmd/midbb asm.bb.middle and asm.flags.middle
RZ_NOPLUGINS=1 /usr/bin/rizin -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -Qc 'e asm.bytes=true
e asm.xrefs=false
s 0x560e67
af
e asm.bb.middle=true
e asm.flags.middle=2
f sym.dummy @ 0x560e7d
pd 2 @ 0x560e7a
?e
fm 0x560e7c @ sym.dummy
pd 3 @ 0x560e7a
?e
fm 0x560e96 @ sym.dummy
pd 2 @ 0x560e95
?e
fm 0x560e97 @ sym.dummy
pd 3 @ 0x560e95
?e
e asm.bb.middle=false
f sym.real @ 0x00560e95
pd 3 @ 0x560e95
?e
pd 3 @ 0x560e95 @e:scr.color=1
' bins/pe/cmd_adf_sample0.exe
-- stdout
--- expected
+++ actual
@@ -25,5 +25,5 @@
;-- sym.real:
|;-- (0x00560e97) sym.dummy:
0x00560e95  ~   ba0f31e914movedx, 0x14e9310f
-|0x00560e9a0000add[48;5;52m3mbyte [eax], al
-|0x00560e9c00a836dc5264add[48;5;52m3mbyte [eax+0x6452dc36], ch
+|0x00560e9a0000add[48;5;22m7mbyte [eax], al
+|0x00560e9c00a836dc5264add[48;5;22m7mbyte [eax+0x6452dc36], ch

[XX] db/cmd/cmd_pd2 newlined-comment-refline fix (#14627), wayward rip-relative flag fixes
RZ_NOPLUGINS=1 /usr/bin/rizin -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -Qc 'e asm.bytes=false
e asm.sub.names=true
(pd_tests; e asm.sub.rel=true; pd 15; ?e; e asm.sub.rel=false; pd 1 @ 0x000041ee; pd 1 @ 0x00004225)
(insn_tests; e scr.color=0; .(pd_tests); ?e; e scr.color=1; .(pd_tests))
s 0x000041e0
.(insn_tests)
?e
e io.cache=true
wa "mov rdx, [rip + 0x17792]; mov rsi, [rip + 0x1d84b]" @ 0x000041e7
wa "mov rdi, [rip + 0x147f6]" @ 0x00004225
.(insn_tests)
?e
wx 483b1592770100483b354bd80100 @ 0x000041e7  # cmp rdx, [rip + 0x17792]; cmp rsi, [rip + 0x1d84b]
wx 483b3df6470100 @ 0x00004225  # cmp rdi, [rip + 0x147f6]
.(insn_tests)
' bins/elf/ls
-- stdout
--- expected
+++ actual
@@ -27,7 +27,7 @@
|0x000041f8call0xc660
|0x000041fetesteax, eax
,==<0x00004200js0x4ae5
-       ||0x00004206cdqe
+       ||0x00004206cdqe
||0x00004208leardx, [0x0001b980]
||0x0000420fxoredi, edi
||0x00004211movesi, dword [rdx+rax*4]
@@ -69,7 +69,7 @@
|0x000041f8call0xc660
|0x000041fetesteax, eax
,==<0x00004200js0x4ae5
-       ||0x00004206cdqe
+       ||0x00004206cdqe
||0x00004208leardx, [0x0001b980]
||0x0000420fxoredi, edi
||0x00004211movesi, dword [rdx+rax*4]
@@ -111,7 +111,7 @@
|0x000041f8call0xc660
|0x000041fetesteax, eax
,==<0x00004200js0x4ae5
-       ||0x00004206cdqe
+       ||0x00004206cdqe
||0x00004208leardx, [0x0001b980]
||0x0000420fxoredi, edi
||0x00004211movesi, dword [rdx+rax*4]
```

XVilka (changes requested):
```
test_rz_colorize_custom_hexagon_0ERR
[XX] Fail at line 511: Colors of "?trap0(#0x3)" are incorrect. Should be "?if(cmp.eq(<err>.new,#0x0))jump:nt0x40"
.: expected true, got false
test_rz_colorize_custom_hexagon_1ERR
[XX] Fail at line 537: Colors of "?trap0(#0x3)" are incorrect. Should be "?LR:FP=dealloc_return(FP):raw"
.: expected true, got false
```

Rot127:
My internet connection is slow so setting up the Windows environment will take a while...

XVilka:
@Rot127 error seems there:
```
\bin\ConnectionManagerExe C:\projects\rizin\build\test\unit\test_tokens.exe
------------------------------------- 8< -------------------------------------
Listing only the last 100 lines from a long log.
WARNING: Failed to create token. Asm strings will be flawed.
WARNING: (../librz/asm/asm.c:1376):add_token: code should not be reached
WARNING: Failed to create token. Asm strings will be flawed.
WARNING: (../librz/asm/asm.c:1376):add_token: code should not be reached
WARNING: Failed to create token. Asm strings will be flawed.
WARNING: (../librz/asm/asm.c:1376):add_token: code should not be reached
WARNING: Failed to create token. Asm strings will be flawed.
WARNING: (../librz/asm/asm.c:1376):add_token: code should not be reached
WARNING: Failed to create token. Asm strings will be flawed.
WARNING: (../librz/asm/asm.c:1376):add_token: code should not be reached
WARNING: Failed to create token. Asm strings will be flawed.
WARNING: (../librz/asm/asm.c:1376):add_token: code should not be reached
WARNING: Failed to create token. Asm strings will be flawed.
WARNING: (../librz/asm/asm.c:1376):add_token: code should not be reached
WARNING: Failed to create token. Asm strings will be flawed.
WARNING: (../librz/asm/asm.c:1376):add_token: code should not be reached
WARNING: Failed to create token. Asm strings will be flawed.
WARNING: (../librz/asm/asm.c:1376):add_token: code should not be reached
WARNING: Failed to create token. Asm strings will be flawed.
WARNING: (../librz/asm/asm.c:1376):add_token: code should not be reached
WARNING: Failed to create token. Asm strings will be flawed.
WARNING: (../librz/asm/asm.c:1376):add_token: code should not be reached
WARNING: Failed to create token. Asm strings will be flawed.
WARNING: (../librz/asm/asm.c:1376):add_token: code should not be reached
WARNING: Failed to create token. Asm strings will be flawed.
WARNING: (../librz/asm/asm.c:1376):add_token: code should not be reached
WARNING: Failed to create token. Asm strings will be flawed.
WARNING: (../librz/asm/asm.c:1376):add_token: code should not be reached
WARNING: Failed to create token. Asm strings will be flawed.
WARNING: (../librz/asm/asm.c:1376):add_token: code should not be reached
WARNING: Failed to create token. Asm strings will be flawed.
WARNING: (../librz/asm/asm.c:1376):add_token: code should not be reached
WARNING: Failed to create token. Asm strings will be flawed.
WARNING: (../librz/asm/asm.c:1376):add_token: code should not be reached
WARNING: Failed to create token. Asm strings will be flawed.
WARNING: (../librz/asm/asm.c:1376):add_token: code should not be reached
WARNING: Failed to create token. Asm strings will be flawed.
WARNING: (../librz/asm/asm.c:1376):add_token: code should not be reached
WARNING: Failed to create token. Asm strings will be flawed.
WARNING: (../librz/asm/asm.c:1376):add_token: code should not be reached
WARNING: Failed to create token. Asm strings will be flawed.
WARNING: (../librz/asm/asm.c:1376):add_token: code should not be reached
WARNING: Failed to create token. Asm strings will be flawed.
WARNING: (../librz/asm/asm.c:1376):add_token: code should not be reached
WARNING: Failed to create token. Asm strings will be flawed.
WARNING: (../librz/asm/asm.c:1376):add_token: code should not be reached
WARNING: Failed to create token. Asm strings will be flawed.
WARNING: (../librz/asm/asm.c:1376):add_token: code should not be reached
WARNING: Failed to create token. Asm strings will be flawed.
WARNING: (../librz/asm/asm.c:1376):add_token: code should not be reached
WARNING: Failed to create token. Asm strings will be flawed.
WARNING: (../librz/asm/asm.c:1376):add_token: code should not be reached
WARNING: Failed to create token. Asm strings will be flawed.
WARNING: (../librz/asm/asm.c:1376):add_token: code should not be reached
WARNING: Failed to create token. Asm strings will be flawed.
WARNING: (../librz/asm/asm.c:1376):add_token: code should not be reached
WARNING: Failed to create token. Asm strings will be flawed.
WARNING: (../librz/asm/asm.c:1376):add_token: code should not be reached
WARNING: Failed to create token. Asm strings will be flawed.
WARNING: (../librz/asm/asm.c:1376):add_token: code should not be reached
WARNING: Failed to create token. Asm strings will be flawed.
WARNING: (../librz/asm/asm.c:1376):add_token: code should not be reached
WARNING: Failed to create token. Asm strings will be flawed.
WARNING: (../librz/asm/asm.c:1376):add_token: code should not be reached
WARNING: Failed to create token. Asm strings will be flawed.
WARNING: (../librz/asm/asm.c:1376):add_token: code should not be reached
WARNING: Failed to create token. Asm strings will be flawed.
WARNING: (../librz/asm/asm.c:1376):add_token: code should not be reached
WARNING: Failed to create token. Asm strings will be flawed.
WARNING: (../librz/asm/asm.c:1376):add_token: code should not be reached
WARNING: Failed to create token. Asm strings will be flawed.
WARNING: (../librz/asm/asm.c:1376):add_token: code should not be reached
WARNING: Failed to create token. Asm strings will be flawed.
WARNING: (../librz/asm/asm.c:1376):add_token: code should not be reached
WARNING: Failed to create token. Asm strings will be flawed.
WARNING: (../librz/asm/asm.c:1376):add_token: code should not be reached
WARNING: Failed to create token. Asm strings will be flawed.
WARNING: (../librz/asm/asm.c:1376):add_token: code should not be reached
WARNING: Failed to create token. Asm strings will be flawed.
WARNING: (../librz/asm/asm.c:1376):add_token: code should not be reached
WARNING: Failed to create token. Asm strings will be flawed.
WARNING: (../librz/asm/asm.c:1376):add_token: code should not be reached
WARNING: Failed to create token. Asm strings will be flawed.
WARNING: (../librz/asm/asm.c:1376):add_token: code should not be reached
WARNING: Failed to create token. Asm strings will be flawed.
WARNING: (../librz/asm/asm.c:1376):add_token: code should not be reached
WARNING: Failed to create token. Asm strings will be flawed.
WARNING: (../librz/asm/asm.c:1376):add_token: code should not be reached
WARNING: Failed to create token. Asm strings will be flawed.
WARNING: (../librz/asm/asm.c:1376):add_token: code should not be reached
WARNING: Failed to create token. Asm strings will be flawed.
WARNING: (../librz/asm/asm.c:1376):add_token: code should not be reached
WARNING: Failed to create token. Asm strings will be flawed.
WARNING: (../librz/asm/asm.c:1376):add_token: code should not be reached
WARNING: Failed to create token. Asm strings will be flawed.
WARNING: (../librz/asm/asm.c:1376):add_token: code should not be reached
WARNING: Failed to create token. Asm strings will be flawed.
WARNING: (../librz/asm/asm.c:1376):add_token: code should not be reached
WARNING: Failed to create token. Asm strings will be flawed.
WARNING: (../librz/asm/asm.c:1376):add_token: code should not be reached
------------------------------------------------------------------------------
```

Rot127:
Will fix the rest of the color errors tomorrow.

librz/asm/arch/hexagon/hexagon_arch.h
```diff
@@ -40,27 +40,40 @@ typedef struct {
...
-#define HEX_PKT_SDK_PADDING   "   "
-#define HEX_PKT_LAST_SDK      " }"
-#define HEX_PKT_FIRST         "/   "
-#define HEX_PKT_MID           "|   "
-#define HEX_PKT_LAST          "\\   "
+#define HEX_PKT_UNK         "?   "
+#define HEX_PKT_SINGLE      "[   "
+#define HEX_PKT_SINGLE_UTF8 "[   "
+#if __WINDOWS__
+// regex matching doesn't work on Windows with UTF-8 characters
```
XVilka: Not a big fan of this change. @GustavoLCR, could you please take a look? Is this an inherent problem or something that could be fixed?

XVilka (changes requested):

librz/asm/arch/hexagon/hexagon_arch.h
Rot127: @GustavoLCR I would be deeply grateful. I haven't understood a thing in the regex code.

XVilka (approved):

XVilka:
Looks good but lets wait for the regex fix to land in "dev" first.

XVilka (approved):

test/db/analysis/x86_64
```diff
@@ -2309,7 +2309,7 @@ aae
 pd 1 @ 0x08048432
 EOF
 EXPECT=<<EOF
-[36m|[0m           [32m0x08048432[0m      [37mlea  [36m eax[0m,[36m [0m[33mstr.Hello_PIC[0m[36m[0m[0m
+[36m|[0m           [32m0x08048432[0m      [37mlea[0m[37m   [0m[36meax[0m[37m, [33mstr.Hello_PIC[0m[0m[0m
```
ret2libc, line 2312: Could you put in a comment here the differences in terms of colors?

test/db/cmd/cmd_ahi
```diff
@@ -58,8 +58,8 @@ ahi 10
 pd 1
 EOF
 EXPECT=<<EOF
-            [32m[7m0x00000000[0m      [33m48[37m81[37mec[33m48[37m06[32m00[37m.[0m  [33msub  [36m rsp[0m,[36m[36m [33m0x648[0m[0m[0m
-            [32m[7m0x00000000[0m      [33m48[37m81[37mec[33m48[37m06[32m00[37m.[0m  [33msub  [36m rsp[0m,[36m[36m [33m1608[0m[0m[0m
+            [32m[7m0x00000000[0m      [33m48[37m81[37mec[33m48[37m06[32m00[37m.[0m  [33msub[0m[37m   [0m[36mrsp[0m[37m, [0m[33m0x648[0m[0m[0m
+            [32m[7m0x00000000[0m      [33m48[37m81[37mec[33m48[37m06[32m00[37m.[0m  [33msub[0m[37m   [0m[36mrsp[0m[37m, [0m[33m1608[0m[0m[0m
```
ret2libc, line 62: also here,please, just so i/we understand how it changes.

wargio (approved):
the lea hack is just horrible, please add a comment that says that it needs to be refactored and simplified

librz/include/rz_asm.h
```diff
@@ -198,6 +201,16 @@ RZ_API int rz_asm_op_set_hexbuf(RzAsmOp *op, const ut8 *buf, int len);
...
 
+// String tokenizing
+RZ_API RZ_OWN RzAsmTokenString *rz_asm_token_string_new(const char *asm_str);
+RZ_API void rz_asm_token_string_free(RZ_OWN RzAsmTokenString *toks);
+RZ_API RZ_OWN RzAsmTokenString *rz_asm_token_string_clone(RZ_OWN RZ_NONNULL RzAsmTokenString *toks);
+RZ_API void rz_asm_token_pattern_free(void *p);
+RZ_API RZ_OWN RzAsmTokenString *rz_asm_tokenize_asm_regex(RZ_BORROW RzStrBuf *asm_str, RzPVector /* RzAsmTokenPattern* */ *patterns);
+RZ_API RZ_OWN RzAsmParseParam *rz_asm_get_parse_param(RZ_NULLABLE const RzReg *reg, ut32 ana_op_type);
+RZ_DEPRECATE RZ_API RZ_OWN RzAsmTokenString *rz_asm_tokenize_asm_string(RZ_BORROW RzStrBuf *asm_str, RZ_NULLABLE const RzAsmParseParam *param);
+RZ_DEPRECATE RZ_API RZ_OWN RzStrBuf *rz_asm_colorize_asm_str(RZ_BORROW RzStrBuf *asm_str, RZ_BORROW RzPrint *p, RZ_NULLABLE const RzAsmParseParam *param, RZ_NULLABLE const RzAsmTokenString *toks);
```
ret2libc, line 212: why deprecated?

ret2libc, line 212: I wonder if these should really stay here... For example, RzReg does not seem to be a dependency of RzAsm (judging by meson.build)

librz/asm/asm.c
```diff
@@ -1210,3 +1291,472 @@ RZ_API RzAsmCode *rz_asm_rasm_assemble(RzAsm *a, const char *buf, bool use_spp)
 	free(lbuf);
 	return acode;
 }
+
+RZ_API RZ_OWN RzAsmTokenString *rz_asm_token_string_new(const char *asm_str) {
+	RzAsmTokenString *s = RZ_NEW0(RzAsmTokenString);
```
ret2libc, line 1296: if (!s) return NULL;

librz/asm/asm.c
```diff
@@ -1210,3 +1291,472 @@ RZ_API RzAsmCode *rz_asm_rasm_assemble(RzAsm *a, const char *buf, bool use_spp)
...
+	}
+	rz_strbuf_free(toks->str);
+	rz_vector_free(toks->tokens);
+	free(toks);
+}
+
+RZ_API RZ_OWN RzAsmTokenString *rz_asm_token_string_clone(RZ_OWN RZ_NONNULL RzAsmTokenString *toks) {
+	rz_return_val_if_fail(toks, NULL);
+
+	RzAsmTokenString *new = RZ_NEW0(RzAsmTokenString);
```
ret2libc, line 1318: if (!new) return NULL;

librz/asm/asm.c
```diff
@@ -1210,3 +1291,472 @@ RZ_API RzAsmCode *rz_asm_rasm_assemble(RzAsm *a, const char *buf, bool use_spp)
...
+ * \param start Index in the asm string of the token.
+ * \param len The length in bytes of the token.
+ * \param type The token type.
+ * \param val The value of the token (should be 0 if token has no value).
+ * \return RzAsmToken* Pointer to the newly created token or NULL in case of failure.
+ */
+static RZ_OWN RzAsmToken *asm_token_create(const size_t start, const size_t len, const RzAsmTokenType type, const ut64 val) {
+	rz_return_val_if_fail(len > 0, NULL);
+	RzAsmToken *t = RZ_NEW0(RzAsmToken);
+	rz_return_val_if_fail(t, NULL);
```
ret2libc:
```suggestion
	if (!t) {
		return NULL;
	}
```

librz/asm/asm.c
```diff
@@ -1210,3 +1291,472 @@ RZ_API RzAsmCode *rz_asm_rasm_assemble(RzAsm *a, const char *buf, bool use_spp)
...
+				i, pi, pj, ci, cj);
+			error = true;
+		}
+		i = cur->start + cur->len;
+		prev = cur;
+	}
+	if (error) {
+		RZ_LOG_WARN("Parsing errors in asm str: %s\n", rz_strbuf_get(toks->str));
+	}
+}
```
ret2libc, line 1460:
```suggestion
}
#else
static void check_token_coverage(RzAsmTokenString *toks) {
}
```

librz/asm/asm.c
```diff
@@ -1210,3 +1291,472 @@ RZ_API RzAsmCode *rz_asm_rasm_assemble(RzAsm *a, const char *buf, bool use_spp)
...
+			}
+			add_token(toks, i, l, pat->type, strtoull(str + i, NULL, 0));
+			j = i + l;
+		}
+	}
+
+	rz_vector_sort(toks->tokens, (RzVectorComparator)cmp_tokens, false);
+#if RZ_CHECKS_LEVEL == 2
+	check_token_coverage(toks);
+#endif
```
ret2libc:
```suggestion
	check_token_coverage(toks);
```

librz/include/rz_asm.h
Rot127: The idea was to mark it as deprecated so people are encouraged to implement a custom token parser for a plugin. This way there wouldn't be two ways for parsing around in the future. @XVilka  brought it in if I recall correctly. But since the generic parsing is rewritten completely we can change it back.

Rot127: Another advantage of nudging devs to the custom parsing method is to get rid of those dependencies between `RzAsm` <-> `RzReg` (like below). Because the plugin has all these information anyways.

Rot127: Will add it as such. Or do you have a suggestion where to move it?

test/db/analysis/x86_64
Rot127: The space characters are now classified as separator tokens. Since each token gets is own color assigned they now get "colored" in white (`37m`) or any other color of cause. Before the refactor, spaces where simply "colored" like the string before them and the color reset escape string came after the spaces.

Rot127:
@wargio 
> the lea hack is just horrible, please add a comment that says that it needs to be refactored and simplified

This is already addressed here: https://github.com/rizinorg/rizin/issues/2766

librz/include/rz_asm.h
XVilka: Yes, the idea is to discourage using generic parsing and move everything to plugins.

XVilka: I think until we merge RzAsm and RzAnalysis into RzArch, RzAsm is the best place to have it. @wargio, what do you think?

ret2libc:
image

Any chance those functions in the `call X` instructions can be colored? I find white to be a bit anonymous. My theme is "default".

ret2libc (approved):
Apart from comment above, looks good to me, good work :)

librz/include/rz_asm.h
wargio: yeah, i want to move this code to RzArch and have it there, since it will be the perfect place.

wargio (approved):

Rot127:
Please wait with merging. Just found a bug with the `pi` command.

Rot127:
> Any chance those functions in the `call X` instructions can be colored? I find white to be a bit anonymous. My theme is "default".

@ret2libc Couldn't find a quick fix for that. But the call addresses were replaced before the tokenization somewhere. Otherwise would be yellow, just like the other flag name at 0x4440. I could look into it next week and add it in another PR.

ret2libc:
> > Any chance those functions in the `call X` instructions can be colored? I find white to be a bit anonymous. My theme is "default".
> 
> @ret2libc Couldn't find a quick fix for that. But the call addresses were replaced before the tokenization somewhere. Otherwise would be yellow, just like the other flag name at 0x4440. I could look into it next week and add it in another PR.
> 

I'm ok with it, thanks. I consider that a small regression but i think it's fine if fixed before next release.

librz/asm/p/asm_hexagon.c
```diff
@@ -13,30 +13,118 @@
...
+	);
+	rz_pvector_push(pvec, pat);
+
+	pat = RZ_NEW0(RzAsmTokenPattern);
+	pat->type = RZ_ASM_TOKEN_META;
+	pat->pattern = strdup(
+		"(#{1,2})|(\\}$)|" // Immediate prefix, Closing packet bracket
+		"\\.new|:n?t|:raw|<err>" // .new and jump hints
+	);
+	rz_pvector_push(pvec, pat);
```
ret2libc, line 44: For next PR or a following one... I feel these operations are not really specific to hexagon. Maybe a sort of "builder API" would help here to avoid duplication.

Something like:
```c
RzPVector *pvec = rz_asm_token_pattern_builder();
rz_asm_token_pattern_builder_add(RZ_ASM_TOKEN_META, ".....");
```

librz/asm/p/asm_hexagon.c
```diff
@@ -13,30 +13,118 @@
...
+	pat->pattern = strdup(
+		"([[:alnum:]]+)|" // Alphanumeric mnemonics
+		"([[:alnum:]]+_[[:alnum:]]+)" // Menmonics with "_" e.g dealloc_return
+	);
+	rz_pvector_push(pvec, pat);
+
+	return pvec;
+}
+
+static void compile_token_patterns(RZ_INOUT RzPVector /* RzAsmTokenPattern* */ *patterns) {
```
ret2libc, line 114: also this function likely will be common to all?

```c
rz_asm_token_pattern_builder_compile(pvec);
```

librz/include/rz_util/rz_print.h
```diff
@@ -41,6 +42,62 @@ typedef const char *(*RzPrintSectionGet)(void *user, ut64 addr);
...
+} RzAsmParseParam;
+
+/**
+ * \brief Pattern for a asm string token.
+ */
+typedef struct {
+	RzAsmTokenType type; //< Asm token type.
+	char *pattern; //< The regex pattern describing the tokens.
+	RzRegex *regex; //< Compiled regex pattern.
+} RzAsmTokenPattern;
```
ret2libc, line 90: should these structs really be in RzPrint? They are also called **RzAsm**something so maybe they belong to RzAsm?

WDYT?

ret2libc (commented):
The rz_asm_token_pattern_builder pattern maybe is not really necessary and it can be introduced only once we see the code being repeated in multiple places. Right now there is only hexagon.

librz/asm/p/asm_hexagon.c
Rot127: Sounds like a great idea to me!

librz/include/rz_util/rz_print.h
Rot127: I agree that they should go to RzAsm modukle. But `rz_print_colorize_asm_str` needs the RzAsmTokenStr as type.
So I see
- Give a `const void *` to  `rz_print_colorize_asm_str` instead of `RzAsmTokenString*` (bit ugly)
- or those structs blong to `RzPrint` (which doesn't fit that well).
Or is there a third way?

ret2libc: move `rz_print_colorize_asm_str` to RzCore?

XVilka:
I suggest opening an issue with suggestions/feedback here so it's not forgotten, we also can add more during the process.

PR #2436, opened 2022-03-23
Optimize `REBASE_PADDR` macro

GustavoLCR:

**Detailed description**

Skips loop if `loadaddr` is 0 as it doesn't have an effect.

**Test plan**

Tests pass

**Closing issues**

None

### DIFF
--- a/librz/include/rz_bin.h
+++ b/librz/include/rz_bin.h
@@ -686,10 +686,12 @@ typedef struct rz_bin_class_t {
 
 #define REBASE_PADDR(o, l, type_t) \
 	do { \
-		RzListIter *_it; \
-		type_t *_el; \
-		rz_list_foreach ((l), _it, _el) { \
-			_el->paddr += (o)->opts.loadaddr; \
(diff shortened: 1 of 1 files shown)


XVilka (approved):

ret2libc (changes requested):
I think the change is ok, but we should leave the `do { ... } while(0)`. That is done so that a `;` is required at the end of the REBASE_PADDR macro. By having just the `if`, instead, the `;` is not required.

librz/include/rz_bin.h
```diff
@@ -687,13 +687,13 @@ typedef struct rz_bin_class_t {
...
 #define REBASE_PADDR(o, l, type_t) \
-	do { \
+	if ((o)->opts.loadaddr) { \
 		RzListIter *_it; \
 		type_t *_el; \
 		rz_list_foreach ((l), _it, _el) { \
 			_el->paddr += (o)->opts.loadaddr; \
 		} \
-	} while (0)
+	}
```
wargio: add the if inside the do .. while statement.

Issue #2357, opened 2022-02-23
libzip subproject - cannot find size of the `off_t` type with Muon


XVilka:
### Work environment

| Questions                                            | Answers
|------------------------------------------------------|--------------------
| OS/arch/bits (mandatory)                             | Fedora 35 x86_64
| File format of the file you reverse (mandatory)      | -
| Architecture/bits of the file (mandatory)            | -
| `rizin -v` full output, **not truncated** (mandatory)  | 7c5e4f1fad87be2fbb7eff94a9f2dd2cba759410

### Expected behavior

No error

### Actual behavior

```
[1063/1649] compiling c subprojects/libzip-1.7.3/liblibzip.a.p/lib/zip_add.c.o
FAILED: subprojects/libzip-1.7.3/liblibzip.a.p/lib/zip_add.c.o 
cc -I subprojects/libzip-1.7.3 -I ../subprojects/libzip-1.7.3 -I ../subprojects/libzip-1.7.3/lib -I subprojects/libzip-1.7.3 -I ../subprojects/libzip-1.7.3 -std=c99 -O3 -Wpedantic -Wextra -Wall -Werror=sizeof-pointer-memaccess -MD -MQ subprojects/libzip-1.7.3/liblibzip.a.p/lib/zip_add.c.o -MF subprojects/libzip-1.7.3/liblibzip.a.p/lib/zip_add.c.o.d -o subprojects/libzip-1.7.3/liblibzip.a.p/lib/zip_add.c.o -c ../subprojects/libzip-1.7.3/lib/zip_add.c
In file included from ../subprojects/libzip-1.7.3/lib/zipint.h:38,
                 from ../subprojects/libzip-1.7.3/lib/zip_add.c:36:
../subprojects/libzip-1.7.3/lib/compat.h:146:2: error: #error unsupported size of off_t
  146 | #error unsupported size of off_t
      |  ^~~~~
[1064/1649] /home/xvilka/rizin/muon/build/muon internal exe -- /usr/bin/python3 /home/xvilka/rizin/rizin_copy/subprojects/libzip-1.7.3/create_zip_err_str.py /home/xvilka/rizin/rizin_copy/build/subprojects/libzip-1.7.3/zip_err_str.c /home/xvilka/rizin/rizin_copy/subprojects/libzip-1.7.3/lib/zip.h
[1065/1649] compiling c subprojects/libzip-1.7.3/liblibzip.a.p/lib/zip_add_entry.c.o
[...]

PR #2449, opened 2022-03-25
Keep targeting meson 0.55

ret2libc:

**Detailed description**

0.58.0 was used for `range` function, which can be easily replaced with python calls.
0.59.0 was used because `import('cmake', required: false)` was used, where `required` was introduced in 0.59.0.

However, it is probably better to keep compatibility with a slightly older meson as that is the one used still in Ubuntu, Debian and many other distros.

I have also added the `check:` argument because there is now a warning in meson 0.62 due to a change in behaviour.

**Test plan**

Dist- tests should pass.

**Closing issues**

...

### DIFF
--- a/librz/core/cmd_descs/meson.build
+++ b/librz/core/cmd_descs/meson.build
@@ -31,7 +31,7 @@ cmd_descs_yaml = files(
 )
 cmd_descs_src_c = files('cmd_descs.c', 'cmd_descs.h')
 
-r = run_command(py3_exe, '-c', 'import yaml')
+r = run_command(py3_exe, '-c', 'import yaml', check: false)
 if r.returncode() == 0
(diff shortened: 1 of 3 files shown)


ret2libc:
@XVilka would it be a problem to break muon support? I have no idea how well it works or if it currently works at all, but I think it's more important to support meson on ubuntu/debian rather than muon, at this point.

wargio (approved):

XVilka (approved):
It will break Muon, yes: https://todo.sr.ht/~lattis/muon/24#event-112004

Nevertheless, you are right, also it's not the only problem for now: https://github.com/rizinorg/rizin/issues/2357

We can wait until Ubuntu updates their Meson and reintroduce this change. I will wait and poke the package maintainers.

PR #2453, opened 2022-03-26
Hexagon: Update register profile

### LINKED ISSUE #6: Assign specific register types to HVX registers and control registers
The details are explained here: https://github.com/rizinorg/rizin/issues/1287

In short:
HVX and Control Registers are marked as GPRs in the register profile. That is because `rizin` does not have vector or control register types.

Once they are implemented in `rizin` we should add them here as well.
An example is already implemented: https://github.com/rizinorg/rz-hexagon/blob/7df09f53826546b8e0a328021813882e13e21de1/HardwareRegister.py#L109-L120

### LINKED ISSUE #28: Incorrect size of predicate registers in register profile
The predicate registers (P0-P3) have the incorrect size in the register profile set. The size is 8bit per register. Not 32bit.

### LINKED ISSUE #1287: Add vector and control register type

Rot127:

**Detailed description**

Companion PR to https://github.com/rizinorg/rz-hexagon/pull/51.

> Although the reg profile was not really used yet it will be relevant for the RZIL uplifting.
> 
> Introduces:
> 
> - Double/quadruple registers now properly overlap with their counterparts.
> - Registers have correct types assigned (`vec`, `ctr`, `sys` etc. Before, all regs were of type `gbr`)
> - Consistent register names (asm name style: `c3:2`, `r31` etc, alias style: `lc1:sa1`, `lr` etc.)
> - Adds `_tmp` registers in the profile. Later used by uplifted instructions (think of `.new` registers).

**Test plan**

Done

**Closing issues**

closes https://github.com/rizinorg/rz-hexagon/issues/6
closes https://github.com/rizinorg/rz-hexagon/issues/28
closes https://github.com/rizinorg/rizin/issues/1287


librz/include/rz_asm.h
```diff
@@ -91,6 +91,17 @@ typedef struct {
...
+/**
+ * \brief Holds several settings for plugins.
+ */
+typedef struct {
+	bool immdisp; //< Display immediates with # symbol (for arm/hexagon architectures). false = show hashs
+	bool immsign; //< Print signed immediates as negative values, not their unsigned representation.
+	bool utf8; //< Flag for plugins: Use utf-8 characters.
+	bool hex_sdk; //< Hexagon arch only. Print packet syntax in hexagon-objdump style.
+	bool hex_reg_alias; //< Print the alias of registers (Alias from C0 = SA0).
+} RzAsmPluginSettings;
```
wargio, line 103: How about using a RzConfig instead of this?
This struct could be passed as argument when initializing the asm plugin, so you can add `plugins.hexagon.sdk` and `plugins.hexagon.reg.alias` dynamically (kinda like i do for jsdec).
I'm proposing this because i hate to have duplicates like we are having here.

@thestr4ng3r @ret2libc @XVilka

librz/analysis/p/analysis_hexagon.c
```diff
@@ -29,15 +29,13 @@ RZ_API int hexagon_v6_op(RzAnalysis *analysis, RzAnalysisOp *op, ut64 addr, cons
 
 	return op->size;
 }
-
 RZ_API char *get_reg_profile(RzAnalysis *analysis) {
```
XVilka, line 33: I think removing an empty line between functions is a mistake

librz/core/cconfig.c
```diff
@@ -2410,7 +2410,7 @@ static bool cb_tracetag(void *user, void *data) {
 static bool cb_utf8(void *user, void *data) {
 	RzCore *core = (RzCore *)user;
 	RzConfigNode *node = (RzConfigNode *)data;
-	core->rasm->utf8 = (bool)node->i_value;
+	core->rasm->settings.utf8 = (bool)node->i_value;
```
XVilka, line 2413: Could you please extract these `RzAsm->settings` change into a separate PR?

librz/include/rz_asm.h
XVilka: Good idea about this one.

XVilka (changes requested):

librz/include/rz_asm.h
thestr4ng3r: The plugins should not be given a `RzConfig` to actively mess with imo. What I would imagine instead is the plugin declaring in some data structure that can be queried that it wants a config var e.g. `reg.alias` bound to some callback and then core/asm will automatically sort that in as `plugins.hexagon.reg.alias` into its `RzConfig`.
The idea is also that you will use these plugins without a `core` in the future, but maybe such a refactoring is best done in combination with `RzArch`.

Rot127: This sounds really cool.

Maybe something like:

Add callbacks in plugin:
```c
RZ_BORROW RzConfig *get_plugin_config(); // Returns plugins RzConfig object with config nodes it would like to set.
void update_config_node(RzConfigNode *node); // Updates value in the plugins RzConfig
```

Before the plugin is put in use the core does:
- `config = RzAsmPlugin::get_plugin_config()`
- Iterates of each `node` in `config`
  - Does `SETCB("plugin.<plugin_name>" + node.conf, <default>, update_config_node, node.desc)`.

And if the plugin is disabled:
- `config = RzAsmPlugin::get_plugin_config()`
- Iterates of each `node` in `config`
  - Do `UNSETCB("plugin.<plugin_name>" + node.conf)`.

Rot127: This would also have the advantage that the plugin specific settings don't clutter the enormous amount of settings if not in use.

wargio: this could be even better. i like the proposal from @Rot127

Rot127: @thestr4ng3r Any comments on the proposal? Otherwise would implement it tomorrow the way it is described up there.

thestr4ng3r: Seems like an interesting idea, if it works. Currently asm does not depend on config though, but since config only uses util, might be ok to add.

PR #2488, opened 2022-04-02
Add `RzConfig` for asm plugins

Rot127:

**Detailed description**

Adds an `RzConfig` to asm plugins and some callbacks to communicate with them.

It also updates the hexagon asm plugin to use the new plugins config.

It works roughly like that:

- On plugin `init()`: Plugin initializes its `RzConfig` struct and adds settings (see `hexagon_init(void **user)`).
- `RzAsm.get_config` callback returns the `RzConfig` struct from the plugin.
- On plugin load the config nodes are copied to the `RzCore` config.
- If a config is set via the shell, it sets the `RzCores` copy of the config node.
- The callback of this setting is called and the `RzCores` copy of the node is given to it.
- The cb should be implemented by the plugin!
  In the Hexagon case it simply copies the newly set settings from the `RzCore` node the plugins node.
  But in general the plugin can do with the information as it pleases.
- If the asm plugin was put out of use the `RzCores` copies of the config nodes are deleted.

**Test plan**

- The hexagon analysis tests cover the working of plugin settings.
- (un)set of plugins settings on asm plugin init was tested manually. 

**Closing issues**

None
Prerequisite for https://github.com/rizinorg/rizin/pull/2490


Rot127:
Would also add an entry in Rizin Book. But currently missing the time for that.

Rot127:
Also, if plugin settings are set in `./config/rizin/rizinrc` they won;t be applied because the plugin is not loaded yet.
I would like to address this bug after the next release. Because time and, I assume, most people won't need it yet.

librz/config/config.c
```diff
@@ -352,6 +352,17 @@ RZ_API RzConfigNode *rz_config_set(RzConfig *cfg, RZ_NONNULL const char *name, c
...
 }
 
+RZ_API bool rz_config_add_node(RZ_BORROW RzConfig *cfg, RZ_OWN RzConfigNode *node) {
+	rz_return_val_if_fail(cfg && node, false);
+	if (!cfg->lock) {
+		ht_pp_insert(cfg->ht, node->name, node);
+		rz_list_append(cfg->nodes, node);
+		return true;
+	}
+	RZ_LOG_WARN("Config locked. Plugin config node not copied.\n");
```
thestr4ng3r: Free the node here

thestr4ng3r (changes requested):
Apart from that, lgtm. The config registration should maybe happen at plugin load time rather than `rz_asm_use()` but since the init/fini also happens there, I think this is ok for now.

librz/asm/asm.c
```diff
@@ -304,13 +305,54 @@ RZ_API bool rz_asm_use_assembler(RzAsm *a, const char *name) {
...
+static void set_plugin_configs(RZ_BORROW RzAsm *rz_asm, RZ_BORROW RzConfig *pcfg) {
+	rz_return_if_fail(pcfg && rz_asm);
+
+	RzConfig *conf = ((RzCore *)(rz_asm->core))->config;
+	RzConfigNode *n;
+	RzListIter *it;
+	rz_list_foreach_iter(pcfg->nodes, it) {
+		n = it->data;
+		if (!rz_config_add_node(conf, rz_config_node_clone(n))) {
+			RZ_LOG_WARN("Failed add %s to the global config.\n", n->name)
```
XVilka: I recommend putting the name in the quotes: `\"%s\"`
Also `failed add` sounds odd.

librz/asm/asm.c
```diff
@@ -304,13 +305,54 @@ RZ_API bool rz_asm_use_assembler(RzAsm *a, const char *name) {
...
+static void unset_plugins_config(RZ_BORROW RzAsm *rz_asm, RZ_BORROW RzConfig *pcfg) {
+	rz_return_if_fail(pcfg && rz_asm);
+
+	RzConfig *conf = ((RzCore *)(rz_asm->core))->config;
+	RzConfigNode *n;
+	RzListIter *it;
+	rz_list_foreach_iter(pcfg->nodes, it) {
+		n = it->data;
+		if (!rz_config_rm(conf, n->name)) {
+			RZ_LOG_WARN("Failed remove %s from the global config.", n->name)
```
XVilka: Same here

librz/asm/asm.c
```diff
@@ -304,13 +305,54 @@ RZ_API bool rz_asm_use_assembler(RzAsm *a, const char *name) {
...
+/**
+ * \brief Deletes all copies of \p pcfg nodes in the RzConfig from \p rz_asm.
+ *
+ * \param rz_asm Pointer to RzAsm struct.
+ * \param pcfg Pointer to the plugins RzConfig struct.
+ */
+static void unset_plugins_config(RZ_BORROW RzAsm *rz_asm, RZ_BORROW RzConfig *pcfg) {
+	rz_return_if_fail(pcfg && rz_asm);
+
+	RzConfig *conf = ((RzCore *)(rz_asm->core))->config;
```
XVilka, line 337: I don't like this kind of access. But I guess right now there is no better way.

librz/config/config.c
```diff
@@ -352,6 +352,17 @@ RZ_API RzConfigNode *rz_config_set(RzConfig *cfg, RZ_NONNULL const char *name, c
 	return node;
 }
 
+RZ_API bool rz_config_add_node(RZ_BORROW RzConfig *cfg, RZ_OWN RzConfigNode *node) {
```
XVilka, line 362: Add Doxygen please.

librz/config/config.c
```diff
@@ -352,6 +352,17 @@ RZ_API RzConfigNode *rz_config_set(RzConfig *cfg, RZ_NONNULL const char *name, c
...
+	rz_return_val_if_fail(cfg && node, false);
+	if (!cfg->lock) {
```
XVilka: Maybe reverse the condition?

XVilka (changes requested):

XVilka (approved):

thestr4ng3r (approved):

PR #1699, opened 2021-09-16
Add support for more AArch64 ELF relocs

XVilka:
**Your checklist for this pull request**
- [x] I've read the [guidelines for contributing](https://github.com/rizinorg/rizin/blob/master/DEVELOPERS.md) to this repository
- [x] I made sure to follow the project's [coding style](https://github.com/rizinorg/rizin/blob/master/DEVELOPERS.md#code-style)
- [ ] I've added tests that prove my fix is effective or that my feature works (if possible)
- [ ] I've updated the documentation and the [rizin book](https://github.com/rizinorg/book) with the relevant information (if needed)

**Detailed description**

Added support for a few more missing AArch64 ELF relocs

**Test plan**

CI is green, compare the results with `readelf` and other tools

**Closing issues**

Closes https://github.com/rizinorg/rizin/issues/1698

PR #2502, opened 2022-04-05
Add big endian PPC relocs

Rot127:

**Detailed description**

Extracts the PPC changes made here: https://github.com/rizinorg/rizin/pull/1699

**Test plan**

Two tests for `R_PPC_JMP_SLOT` and `R_PPC_RELATIVE` were added.
The expected values are from IDA.

Unfortunately both tests fail and are therefore marked as broken.

Based on a little debugging the `P` values are incorrect. Possibly a general bug in the elf relocation code?

**Closing issues**

...

### DIFF
--- a/librz/bin/p/bin_elf.inc
+++ b/librz/bin/p/bin_elf.inc
@@ -934,35 +934,33 @@ static void patch_reloc(struct Elf_(rz_bin_elf_obj_t) * obj, RzBinElfReloc *rel,
 		default:
 			break;
 		}
-		RzBuffer *rbuf = rz_buf_new_with_bytes(buf, sizeof(buf));
 		if (low) {
 			switch (low) {
 			case 14:
 				val &= (1 << 14) - 1;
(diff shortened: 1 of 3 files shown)


wargio (approved):

ret2libc (changes requested):
I think it's better to keep the `word` as 1/2/4/8 because it seems to be so in the rest of the file.

ret2libc (changes requested):
Are you sure about the `rz_buf_write_ble32_at(rbuf, patch_addr, ...`? Before it was using `rz_write_le32` and I think it makes more sense because rbuf is probably 8 bytes at most, or somethning like that.

Rot127:
TBH I did not really think about it. @08A  noted it in https://github.com/rizinorg/rizin/pull/1699 and I just assumed it is better.
But I agree that  `rz_write_ble32_at` is not necessary.
Anyways. Found a typo this way.

XVilka (approved):

PR #2503, opened 2022-04-05
Fix `rz_str_ndup()`-related OOB read by using `rz_str_ndup_buflen()`

kazarmy:

**Detailed description**

This pr fixes this [`rz_str_ndup()`-related OOB read](https://github.com/rizinorg/rizin/runs/5828574225?check_suite_focus=true#step:19:26) by using `rz_str_ndup_buflen()`, since unlike [POSIX `strndup()`](https://pubs.opengroup.org/onlinepubs/9699919799/functions/strdup.html), it guarantees that the pointer returned is to a buffer of size `len + 1`.

**Test plan**

All builds are green.

**Closing issues**

...

### DIFF
--- a/librz/include/rz_util/rz_str.h
+++ b/librz/include/rz_util/rz_str.h
@@ -179,6 +179,7 @@ static inline const char *rz_str_get_null(const char *str) {
 	return str ? str : "(null)";
 }
 RZ_API char *rz_str_ndup(RZ_NULLABLE const char *ptr, int len);
+RZ_API char *rz_str_ndup_buflen(RZ_NULLABLE const char *ptr, int len);
 RZ_API char *rz_str_dup(char *ptr, const char *string);
 RZ_API int rz_str_inject(char *begin, char *end, char *str, int maxlen);
 RZ_API int rz_str_delta(char *p, char a, char b);
--- a/librz/util/astr.c
+++ b/librz/util/astr.c
@@ -56,11 +56,11 @@ RZ_API RASN1String *rz_asn1_stringify_string(const ut8 *buffer, ut32 length) {
 	if (!buffer || !length) {
 		return NULL;
 	}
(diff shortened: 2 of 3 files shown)


wargio (approved):

wargio:
can you reopen this PR with a branch containing the word `fuzz` ?

XVilka (approved):

ret2libc (changes requested):
I think this is the wrong place to fix. IMHO the problem is in `rz_str_filter`, which should always ensure that `str[i]` is not `\x00`. The `len` in `rz_str_filter` is a `max_length` imho. (if you change that function please add doxygen doc).

kazarmy:
> can you reopen this PR with a branch containing the word `fuzz` ?

sure, see #2504

kazarmy:
> the problem is in `rz_str_filter`, which should always ensure that `str[i]` is not `\x00`.

Not all strings are null-terminated, e.g. [Pascal strings](https://en.wikipedia.org/wiki/String_(computer_science)#Length-prefixed).

ret2libc:
> > the problem is in `rz_str_filter`, which should always ensure that `str[i]` is not `\x00`.
> 
> Not all strings are null-terminated, e.g. [Pascal strings](https://en.wikipedia.org/wiki/String_(computer_science)#Length-prefixed).

I think `rz_str_*` functions should deal with the C strings, aren't they? They are used to manipulate C strings (as used by rizin itself), rather than the "string concept" used in the binaries. Am I wrong?

kazarmy:
A look at `rz_str_filter()` indicates that it's being used to convert non-printable characters to dots, presumably for printing. Thus, it appears likely that it's being used on strings-of-the-binary rather than internal rizin strings. Perhaps it shouldn't be in `rz_str_*` but where should it be?

PR #2504, opened 2022-04-05
Fix `rz_str_ndup()`-related OOB read by using `rz_str_ndup_buflen()`

kazarmy:
Dup of #2503 with `fuzz` in branch name.

### DIFF
--- a/librz/include/rz_util/rz_str.h
+++ b/librz/include/rz_util/rz_str.h
@@ -179,6 +179,7 @@ static inline const char *rz_str_get_null(const char *str) {
 	return str ? str : "(null)";
 }
 RZ_API char *rz_str_ndup(RZ_NULLABLE const char *ptr, int len);
+RZ_API char *rz_str_ndup_buflen(RZ_NULLABLE const char *ptr, int len);
 RZ_API char *rz_str_dup(char *ptr, const char *string);
 RZ_API int rz_str_inject(char *begin, char *end, char *str, int maxlen);
 RZ_API int rz_str_delta(char *p, char a, char b);
--- a/librz/util/astr.c
+++ b/librz/util/astr.c
@@ -56,11 +56,11 @@ RZ_API RASN1String *rz_asn1_stringify_string(const ut8 *buffer, ut32 length) {
 	if (!buffer || !length) {
 		return NULL;
 	}
-	char *str = rz_str_ndup((const char *)buffer, length);
+	char *str = rz_str_ndup_buflen((const char *)buffer, length);
 	if (!str) {
 		return NULL;
 	}
-	rz_str_filter(str, length - 1);
+	rz_str_filter(str, length);
 	return rz_asn1_create_string(str, true, length);
 }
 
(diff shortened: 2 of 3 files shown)


XVilka (approved):

ret2libc (changes requested):
For discussion.

XVilka:
I am merging this to fix the `dev`. If you have a better fix - please send a separate PR.

ret2libc (changes requested):
Changes to any "public API" should be merged with care IMO.

ret2libc:
Moving conversation here:
> > the problem is in rz_str_filter, which should always ensure that str[i] is not \x00.
> 
> Not all strings are null-terminated, e.g. [Pascal strings](https://en.wikipedia.org/wiki/String_(computer_science)#Length-prefixed).

> A look at rz_str_filter() indicates that it's being used to convert non-printable characters to dots, presumably for printing. Thus, it appears likely that it's being used on strings-of-the-binary rather than internal rizin strings. Perhaps it shouldn't be in rz_str_* but where should it be?

rz_str_filter uses strlen if len < 1, so in some cases it seems it has the condition that `str` should be NULL terminated. Actually, looking at the use cases all of them use len=-1 or len=0 except the use case in astr.c, which created the string from `rz_str_ndup`(thus it is null terminated as well). For this reason I think checking just for s[i] in rz_str_filter is enough.

kazarmy:
> Actually, looking at the use cases all of them use len=-1 or len=0 except the use case in astr.c, which created the string from `rz_str_ndup`(thus it is null terminated as well).

Yes but after the `rz_str_filter()` line is this line:
https://github.com/rizinorg/rizin/blob/962ffe2755eb7c7d0f9a12ef9ac5ed773cf595c4/librz/util/astr.c#L64

What is the relationship between `length` and `str` at this point? Changing just `rz_str_filter()` doesn't affect this relationship.

I see that there are 2 options here:

1. Either `length` is updated using an (imho unnecessary) `strlen()` call. This can be done before the call to `rz_str_filter()`.
2. Or  `rz_str_ndup_buflen()` is used and `str[length]` will always remain valid i.e. not OOB.

I do not have strong opinions either way, but I would like to note that option 1 will actually result in 2 additional calls to `strlen()` or equivalent compared to option 2, since  `rz_str_ndup()` also finds the length of its input string.

kazarmy:
> 2. Or `rz_str_ndup_buflen()` is used and `str[length]` will always remain valid i.e. not OOB.

Here I mean the original value of `length` as passed to `rz_asn1_stringify_string()`.

kazarmy:
I don't think it's good to have a `length` that is larger than the memory block allocated to `str`.

ret2libc:
> I don't think it's good to have a `length` that is larger than the memory block allocated to `str`.

I agree, and it is also not good IMO to have a length field that describe the length of a "C string". Why would you want to create a `RASN1String` object that says the string is "hello" but with a length of 80? After rz_str_ndup `length` is not useful anymore, I think.

kazarmy:
> After rz_str_ndup `length` is not useful anymore, I think.

Perhaps but it might still be used in lieu of running `strlen()` every time the length of the string is needed.

Issue #2080, opened 2021-12-09
ESIL to RZIL conversion


XVilka:
# First stage (DONE)

<details>
## High-priority

- [x] `librz/analysis/p/analysis_avr.c` https://github.com/rizinorg/rizin/pull/1890
- [x] `librz/analysis/p/analysis_arm_cs.c` ARMv7 https://github.com/rizinorg/rizin/pull/2241
- [x] `librz/analysis/p/analysis_arm_cs.c` ARMv8 https://github.com/rizinorg/rizin/pull/2354
- [x] `librz/analysis/p/analysis_ppc_cs.c` https://github.com/rizinorg/rizin/pull/2823
- [x] `librz/analysis/p/analysis_x86_cs.c` https://github.com/rizinorg/rizin/pull/2747 https://github.com/rizinorg/rizin/pull/3865
- [x] `librz/analysis/p/analysis_mips_cs.c` by @wargio https://github.com/rizinorg/rizin/pull/4882
- [x] *`librz/analysis/p/analysis_mips_gnu.c`* (should be merged into `mips_cs` and removed) by @wargio (blocked on https://github.com/rizinorg/rizin/issues/4007)

## The rest

- [x] `librz/analysis/p/analysis_bf.c`
- [x] `librz/analysis/p/analysis_6502.c`
- [x] `librz/analysis/p/analysis_sh.c` https://github.com/rizinorg/rizin/pull/2518
- [x]  `librz/analysis/p/analysis_8051.c` https://github.com/rizinorg/rizin/pull/2999
- [x] `librz/analysis/p/analysis_gb.c` https://github.com/rizinorg/rizin/pull/3750
- [x] `librz/analysis/p/analysis_wasm.c` **REMOVED** https://github.com/rizinorg/rizin/pull/3497
- [x] `librz/analysis/p/analysis_rsp.c` **REMOVED** https://github.com/rizinorg/rizin/pull/3531
- [x] `librz/analysis/p/analysis_v810.c` https://github.com/rizinorg/rizin/pull/4096
- [x] `librz/analysis/p/analysis_v850.c` https://github.com/rizinorg/rizin/pull/4103
- [x] `librz/analysis/p/analysis_pic.c` https://github.com/rizinorg/rizin/pull/4322
- [x] `librz/analysis/p/analysis_xtensa.c` by @imbillow  https://github.com/rizinorg/rizin/pull/4712
[...]

PR #2518, opened 2022-04-07
[RzIL] IL lifting of SuperH ISA

DMaroo:

**Detailed description**

Lifting SuperH ISA to RzIL.
Architecture reference: https://www.st.com/resource/en/user_manual/cd00147165-sh-4-32-bit-cpu-core-architecture-stmicroelectronics.pdf

I have also implemented a disassembler and assembler for SuperH.

**Test plan**

> Add one IL test for every lifted instruction.

I have added one IL test for every instruction which was implemented. It is yet to be tested against `rz-tracetest`, but the asm tests are passing.

**Closing issues**

Tracking issue: #2080

**Unimplemented instructions**

 - [ ] `LDTLB`
 - [ ] `OCBI`
 - [ ] `OCBP`
 - [ ] `OCBWB`
 - [ ] `PREF`
 - [ ] `TRAPA`
 - [ ] All floating point instructions are unimplemented, in the disassembler and assembler, as well as in the IL
 - [ ] Delayed branch jumps are also implemented as just normal jumps
 - [x] ~For `LDC` and `STC`, the instruction involving banked registers as the operands is unimplemented (more like incorrectly implemented) in the IL~


librz/asm/arch/sh/disassembler.h
```diff
@@ -0,0 +1,56 @@
...
+
+typedef struct sh_opcode_t {
+	SHOpMnem mnemonic;
+	SHParam param[2];
+	SHScaling scaling;
+} SHOp;
+
+ut32 sh_disassembler(const ut8 *buffer, const ut32 size, ut64 pc, bool be, SHOp *aop, RzStrBuf *sb);
+
+#endif /* RZ_ASM_AVR_DISASSEMBLER_H */
```
wargio: you want to change this line

librz/analysis/arch/sh/sh_il.c
```diff
@@ -0,0 +1,653 @@
...
+	return BRANCH(eq, SETG(SH_SR_T, SH_BIT(1)), SETG(SH_SR_T, SH_BIT(0)));
+}
+
+/**
+ * DIV1  Rm, Rn
+ * 1-step division (Rn ÷ Rm) ; Calculation result -> T
+ * 0011nnnnmmmm0100
+ */
+static RzILOpEffect *sh_il_div1(SHOp *op, ut64 pc, RzAnalysis *analysis) {
+	// TODO
```
DMaroo: Implement `DIV1`.
Resources: https://www.renesas.com/eu/en/document/mah/sh-1sh-2sh-dsp-software-manual?language=en

librz/include/rz_il/rz_il_opbuilder_begin.h
```diff
@@ -85,7 +85,7 @@
 #define ULT(x, y)   rz_il_op_new_ult(x, y)
 #define ULE(x, y)   rz_il_op_new_ule(x, y)
 #define UGT(x, y)   rz_il_op_new_ugt(x, y)
-#define UGR(x, y)   rz_il_op_new_ugr(x, y)
+#define UGE(x, y)   rz_il_op_new_uge(x, y)
```
XVilka: I suggest to send this change in a separate PR.

DMaroo: Implemented in #2566.

librz/analysis/arch/arm/arm_il64.c
```diff
@@ -606,7 +606,7 @@ static RzILOpEffect *branch(cs_insn *insn) {
 	}
 	RzILOpBool *c = cond(insn->detail->arm64.cc);
 	if (c) {
-		return BRANCH(c, JMP(a), NOP);
+		return BRANCH(c, JMP(a), NOP());
```
XVilka: Could you please extract these `NOP()` changes into a separate PR?

librz/analysis/arch/sh/sh_il.c
```diff
@@ -0,0 +1,1386 @@
...
+ * PRIVILEGED (Only GBR is not privileged)
+ *
+ * LDC.L  @Rm+, REG
+ * REG := SR/GBR/VBR/SSR/SPC/DBR/Rn_BANK
+ * (Rm) -> REG ; Rm + 4 -> Rm
+ * PRIVILEGED (Only GBR is not privileged)
+ */
+static RzILOpEffect *sh_il_ldc(SHOp *op, ut64 pc, RzAnalysis *analysis) {
+	RzBitVector *priv_bit = rz_il_evaluate_bitv(analysis->il_vm->vm, VARG(SH_SR_D));
+	ut8 state = priv_bit->bits.small_u == 0 ? 0b1 : 0b0;
```
XVilka: Standard C does not support binary literals, which is so wrong, but yeah, better to avoid custom extensions.

https://stackoverflow.com/questions/18244726/why-doesnt-c-have-binary-literals

librz/analysis/arch/sh/sh_il.c
```diff
@@ -0,0 +1,1386 @@
...
+	state += op->param[1].param[0] != SH_REG_IND_GBR ? 0b10 : 0b00;
+	if ((state & 0x11) == 0x11) {
+		rz_il_vm_event_add(analysis->il_vm->vm, rz_il_event_exception_new("SuperH: RESINST"));
+	}
+	if (op->scaling == SH_SCALING_INVALID) {
+		if (state & 0b10) {
```
XVilka: Same

librz/analysis/arch/sh/sh_il.c
```diff
@@ -0,0 +1,1386 @@
...
+			return BRANCH(VARG(SH_SR_D), sh_il_set_pure_param(1, sh_il_get_pure_param(0)), NOP());
+		} else {
+			return sh_il_set_pure_param(1, sh_il_get_pure_param(0));
+		}
+	} else if (op->scaling == SH_SCALING_L) {
+		SHParamHelper rm = sh_il_get_param(op->param[0], op->scaling);
+		if (state & 0b10) {
```
XVilka: Same

librz/analysis/arch/sh/sh_il.c
```diff
@@ -0,0 +1,1386 @@
...
+ * PRIVILEGED (Only GBR is not privileged)
+ *
+ * STC.L  REG, @-Rn
+ * REG := SR/GBR/VBR/SSR/SPC/DBR/Rn_BANK
+ * Rn - 4 -> Rn ; REG -> (Rn)
+ * PRIVILEGED (Only GBR is not privileged)
+ */
+static RzILOpEffect *sh_il_stc(SHOp *op, ut64 pc, RzAnalysis *analysis) {
+	RzBitVector *priv_bit = rz_il_evaluate_bitv(analysis->il_vm->vm, VARG(SH_SR_D));
+	ut8 state = priv_bit->bits.small_u == 0 ? 0b1 : 0b0;
```
XVilka: Same, etc

XVilka (changes requested):
The code is in a good shape.

librz/analysis/arch/arm/arm_il64.c
DMaroo: Done: #2705

librz/analysis/arch/sh/sh_il.c
```diff
@@ -0,0 +1,1335 @@
...
+	return sh_apply_effects(ret, pre, post);
+}
+
+static inline RzILOpBool *sh_il_is_add_carry(RZ_OWN RzILOpPure *res, RZ_OWN RzILOpPure *x, RZ_OWN RzILOpPure *y) {
+	// res = x + y
+	// x & y
+	RzILOpPure *xy = LOGAND(x, y);
+
+	RzILOpPure *nres = LOGNOT(res);
+	// !res & y
```
XVilka, line 511: Maybe move this up one line to be clear where `nres` comes from?

librz/analysis/arch/sh/sh_il.c
```diff
@@ -0,0 +1,1335 @@
...
+	if (op->scaling == SH_SCALING_INVALID) {
+		return sh_il_set_pure_param(1, sh_il_get_pure_param(0));
+	} else if (op->scaling == SH_SCALING_L) {
+		SHParamHelper rm = sh_il_get_param(op->param[0], op->scaling);
+		return SEQ2(sh_il_set_pure_param(1, rm.pure), rm.post);
+	}
+	return NOP();
+}
+
+// TODO: Implement LDTLB, MOVCA.L, OCBI, OCBP, OCBWB, PREF
```
XVilka: Please list them in the PR description

librz/asm/arch/sh/disassembler.c
```diff
@@ -0,0 +1,471 @@
...
+		    .mode = addrmode } }, \
+		.is_param = true }
+
+// opcode for "weird" movl
+#define MOVL 0x1fff
+
+#define NOPARAM ADDR(NIB0, SH_ADDR_INVALID)
+
+/**
+ * @brief Get params for mov.l instruction (0001NMD)
```
XVilka: `\brief`

librz/asm/arch/sh/disassembler.c
```diff
@@ -0,0 +1,471 @@
...
+	}
+
+	RZ_LOG_WARN("SuperH: Invalid opcode encountered by disassembler")
+	return NULL;
+}
+
+#define MNEM_PADLEN 10
+#define ARG_PADLEN  14
+
+RZ_API RZ_OWN char *sh_op_param_to_str(SHParam param) {
```
XVilka: Doxygen? Also, I am not really sure it should be the API function.

librz/asm/arch/sh/disassembler.c
```diff
@@ -0,0 +1,471 @@
...
+		rz_strbuf_appendf(buf, "%#04x", param.param[0]);
+		break;
+	default:
+		rz_warn_if_reached();
+	}
+
+	return rz_strbuf_drain(buf);
+}
+
+RZ_API RZ_OWN char *sh_op_to_str(const SHOp *op) {
```
XVilka: Same here

XVilka (changes requested):
Looks good from the code point of view.

librz/asm/arch/sh/disassembler.c
DMaroo: I think it will be good to have an API function for disassembly.

DMaroo:
The assembler is not complete yet. The basic idea has been implemented. Also, the assembler when completed won't be completely robust, i.e. it will assemble correctly for correct disassembly, but might not error out correctly for incorrect assembly. However, for the purposes of writing tests, I believe, that this level of robustness is enough. It can always be improved later.

Also, it might be a good idea to use Keystone (or use it after adding our own patches, in case it does not work correctly) for all the assembly, since there is significant code duplication, and none of the duplicated code can be reused, since it is architecture specific.

We could write our own assembly framework, but is it worth the effort?

XVilka:
No, spending too much time on SuperH assempler certainly doesn't worth the effort. We should focus on disassembly and RZIL, also accuracy instead.

DMaroo:
I have written an assembler and disassembler for SuperH. They only work for CPU instructions, however, extending to FPU instructions would not be a hassle (since it would only involve adding instruction info to an array, no extra code is (hopefully) needed). They should work nicely for "normal" cases. The assembler cannot handle invalid instructions very well, since it does not check the validity of the operands. But it should be correct as long as legitimate instructions are passed to it.

librz/asm/arch/sh/assembler.c
```diff
@@ -0,0 +1,384 @@
...
+				return SH_REG_INDIRECT_INDEXED;
+			} else if (rz_str_endswith(param, ",gbr)")) {
+				return SH_GBR_INDIRECT_DISP;
+			} else if (rz_str_endswith(param, ",pc)")) {
+				return SH_PC_RELATIVE_DISP;
+			} else {
+				return SH_REG_INDIRECT_DISP;
+			}
+		}
+		// unreachable
```
XVilka: `rz_warn_if_reached()`?

librz/asm/arch/sh/disassembler.c
```diff
@@ -0,0 +1,223 @@
...
+
+/**
+ * \brief Disassemble \p opcode and return a SHOp
+ *
+ * \param opcode 16 bit wide opcode
+ * \return SHOp object corresponding to the opcode
+ */
+RZ_API RZ_OWN SHOp *sh_disassembler(ut16 opcode) {
+	for (ut16 i = 0; i < OPCODE_NUM; i++) {
+		if ((opcode | sh_op_lookup[i].mask) == sh_op_lookup[i].opcode) {
```
XVilka: You could reverse the condition and call `continue` instead, one level of nestiness would be gone.

librz/asm/arch/sh/disassembler.h
```diff
@@ -0,0 +1,237 @@
...
+	ut16 opcode;
+	const char *str_mnem;
+	SHOpMnem mnemonic;
+	SHParam param[2];
+	SHScaling scaling;
+} SHOp;
+
+RZ_API RZ_OWN SHOp *sh_disassembler(ut16 opcode);
+
+RZ_API RZ_OWN char *sh_op_param_to_str(SHParam param, SHScaling scaling, ut64 pc);
```
XVilka: I think, it either should not be `RZ_API`, or has a corresponding prefix `rz_asm_`

XVilka (changes requested):
So far, looks very good. We need more tests for the RzIL on the instruction level

librz/asm/arch/sh/disassembler.h
DMaroo: Ah, in that case I'll add the prefix `rz_asm`. This functions are pretty nifty if one wants to manually assemble and disassemble stuff without loading the SuperH plugin and the associated instance of `RzAnalysis`. So exposing them is probably a good idea.

DMaroo: @thestr4ng3r @wargio @ret2libc Please suggest what should be done.

thestr4ng3r: `rz_` prefix and public api is fine

wargio: i do not see any reason fo which this should be RZ_API

DMaroo: In conclusion, I will convert all of these to `RZ_IPI`

wargio (commented):
Nice optimization

XVilka (changes requested):
.

librz/asm/p/asm_sh.c
```diff
@@ -1,78 +1,51 @@
 // SPDX-FileCopyrightText: 2010-2011 eloi <limited-entropy.com>
 // SPDX-License-Identifier: LGPL-3.0-only
+// SPDX-FileCopyrightText: 2022 Dhruv Maroo <dhruvmaru007@gmail.com>
```
XVilka, line 2: No need to double `SPDX-License-Identifier` and put `SDPX-FileCopyrightText` together

librz/asm/p/asm_sh.c
```diff
@@ -1,78 +1,51 @@
...
+	SHOp *dis_op = sh_disassembler(rz_read_ble16(buf, a->big_endian));
+	op->size = 2;
+	if (!dis_op) {
+		rz_strbuf_set(&op->buf_asm, "invalid");
+	} else {
+		char *disasm = sh_op_to_str(dis_op, a->pc);
+		rz_strbuf_set(&op->buf_asm, disasm);
+		free(disasm);
+	}
+	RZ_FREE(dis_op);
```
XVilka, line 21: `RZ_FREE()` also sets `dis_op` to `NULL`, since it's a local variable and isn't used anymore, just `free()` would be sufficient

librz/util/str.c
```diff
@@ -1903,6 +1903,22 @@ RZ_API bool rz_str_is_ascii(const char *str) {
 	return true;
 }
 
+/**
+ * \brief Checks if the whole string is composed of whitespace
+ *
+ * \param str input string
+ * \return bool true if whitespace (or empty or NULL), false otherwise
```
XVilka: This description isn't true since it fails on NULL

librz/analysis/arch/sh/sh_il.c
```diff
@@ -0,0 +1,1772 @@
...
+append:
+	if (post) {
+		target = SEQ2(target, post);
+	}
+
+	return target;
+}
+
+/**
+ * @brief Set the value of the \p param at \p scaling to \p val
```
XVilka: `\brief`

XVilka (changes requested):

librz/asm/p/asm_sh.c
DMaroo: Yes, I just wanted to avoid any sort of use-after-free stuff (by mistake), so just to be sure.

XVilka (changes requested):
.

librz/analysis/arch/sh/sh_il.c
```diff
@@ -0,0 +1,1780 @@
...
+
+static bool sh_banked_reg(ut16 reg) {
+	return reg < SH_BANKED_REG_COUNT;
+}
+
+/**
+ * \brief Set to true whenever the privilege bit is calculated
+ * It is used to add a `SETL` effect for the privilege bit, in case it is used
+ */
+static bool privilege_check = false;
```
XVilka: Is it possible to avoid the global/static variable?

librz/asm/arch/sh/assembler.c
```diff
@@ -0,0 +1,388 @@
+// SPDX-FileCopyrightText: 2022 Dhruv Maroo <dhruvmaru007@gmail.com>
+// SPDX-License-Identifier: LGPL-3.0-only
+
+#include "assembler.h"
+#include "regs.h"
+
+/**
+ * \brief Get the addressing mode for \p pb
+ *
+ * @param pb SHParamBuilder
```
XVilka: `\param`

XVilka (changes requested):
Looks very solid. But global or static variables are a big no-no.

librz/analysis/arch/sh/sh_il.c
DMaroo: Yes, kinda. I can pass a `bool*` to an IL lifter and store the information whether the privilege bit was calculated or not within that `bool`. But this will result in one extra `bool*` in all the instruction lifters and many of the utility functions as well, leading to an ugly interface.

The `privilege_check` is essentially used as a local variable only (since it is initialized to `false` on every invocation of `rz_sh_il_opcode`, but I am using a global variable so as to avoid using an extra argument everywhere.

DMaroo: On second thought, I think I should avoid non-const static/global variables since IL lifting has good chances of being multi-threaded in the future. So it's better if I keep that in mind right now as well. I will get rid of the static global.

wargio: please get rid of this and just check for the register bit

librz/asm/arch/sh/common.h
```diff
@@ -0,0 +1,277 @@
...
+		    .mode = addrmode } }, \
+		.is_param = true }
+
+// opcode for "weird" movl
+#define MOVL 0x1fff
+
+#define NOPARAM ADDR(NIB0, SH_ADDR_INVALID)
+
+// Opcode lookup list
+static const SHOpRaw sh_op_lookup[] = {
```
wargio: This is not exactly the best idea ever to have this huge list duplicated on multiple files.

wargio (changes requested):
removed the static var

librz/asm/arch/sh/common.h
DMaroo: Hmmm... Should I make it a global variable? And use `extern` wherever needed?

wargio: yes

DMaroo: Any reason for making it a global variable? It is a `static const` variable, so I don't think it makes any difference functionally. The current approach just feels cleaner to me.

wargio: ypu may have issues when compiling and also you will increasen the size of the binary because it will be duplicated.

XVilka (approved):
Looks great! Ready to shine, IMHO.

thestr4ng3r (changes requested):
Still missing asm test coverage on some instructions: https://app.codecov.io/gh/rizinorg/rizin/blob/sh-uplifting-cov/librz/analysis/arch/sh/sh_il.c
(Almost) all code should be green there, except for error cases that can't be reached by any byte sequences.

You can temporarily enable the codecov job like this: https://github.com/rizinorg/rizin/commit/806dadc9e0314f74c1b075887970da0501d31fb7

DMaroo:
@thestr4ng3r Yes. That is expected. This is because I haven't written the asm tests for _all_ the instructions yet, which is why much of the code isn't tested. I am planning to add the tests for the remaining instructions in a separate PR.

thestr4ng3r:
The tests should be done in this pr since otherwise when disassembling code in a real sh binary in practice, one might hit the untested code and there is nothing to ensure it won't e.g. segfault somewhere.

wargio:
don't forget to run the linter.

XVilka (approved):
LGTM

wargio (commented):
wow, this PR is awesome :D

thestr4ng3r (approved):
Excellent, good job!
https://app.codecov.io/gh/rizinorg/rizin/blob/sh-uplifting-cov/librz/analysis/arch/sh/sh_il.c

wargio (approved):

PR #2513, opened 2022-04-07
rz-ax: count set bits in number

DhruvaG2000:
Signed-off-by: Dhruva Gole <goledhruva@gmail.com>

 

**Your checklist for this pull request**
- [x] I've read the [guidelines for contributing](https://github.com/rizinorg/rizin/blob/master/DEVELOPERS.md) to this repository
- [x] I made sure to follow the project's [coding style](https://github.com/rizinorg/rizin/blob/master/DEVELOPERS.md#code-style)
- [ ] I've documented or updated the documentation of every function and struct this PR changes. If not so I've explained why.
- [x] I've added tests that prove my fix is effective or that my feature works (if possible)
- [ ] I've updated the [rizin book](https://github.com/rizinorg/book) with the relevant information (if needed)

**Detailed description**
The idea is to allow a new bit mode for rz-ax that will print the position of set bits and their count in the following format:
```
$ rz-ax =2 0xb3
10110011b
$ rz-ax -B 0xb3
[0-1]: 1
[4-5]: 1
[7]: 1
```

...

**Test plan**
TODO

...

**Closing issues**
closes https://github.com/rizinorg/rizin/issues/2458

...

PR #2533, opened 2022-04-11
rz-ax: move str to bin to rz_util

### LINKED ISSUE #2512: move bin -> str to rz_util
Refer [rz-ax.c#L400](https://github.com/rizinorg/rizin/blob/dev/librz/main/rz-ax.c#L400).
Moving this binary to string converter to ``rz_util/rz_str.h`` will help create an ``RZ_API`` which can b reused in other places.

Could probably name it ``rz_bin_to_str()``

DhruvaG2000:
Signed-off-by: Dhruva Gole <goledhruva@gmail.com>

 

**Detailed description**

help create an RZ_API ``rz_str_to_bin( )`` which can be reused in other places
...

**Test plan**

TODO
...

**Closing issues**

closes #2512
...

### DIFF
--- a/librz/include/rz_util/rz_str.h
+++ b/librz/include/rz_util/rz_str.h
@@ -195,6 +195,7 @@ typedef void (*str_operation)(char *c);
 RZ_API int rz_str_do_until_token(str_operation op, char *str, const char tok);
 
 RZ_API void rz_str_reverse(char *str);
+RZ_API void rz_str_to_bin(char *str);
 RZ_API int rz_str_re_match(const char *str, const char *reg);
 RZ_API int rz_str_re_replace(const char *str, const char *reg, const char *sub);
 RZ_API int rz_str_path_unescape(char *path);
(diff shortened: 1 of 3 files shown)


DhruvaG2000:
I think there is a more elegant solution possible for this, which would perhaps either be to convert the input string into a binary as it is passed by reference. However this would still essentially be a char string of 0's and 1's. Another idea is to have a second string passed as a parameter that will have the 0's and 1's copied to it.

Let me know if this PR as of [29bb1e2] is enough or I should implement what I have just discussed above and which approach of the 2.

wargio:
to be fair, i do not see any reason for which that function should go on rz_util (the comment was set in this commit https://github.com/rizinorg/rizin/commit/82416b021239286833c2336dddcb3ac6d1dab165).
I would just delete the comment and keep it inside rz-ax since it is used only there.
Moving it under rz_util means that that printf needs to be replaced with a RzStrBuf and the function will need to return a buffer which will need to be freed afterwards.

so, for me i vote to just remove the line https://github.com/rizinorg/rizin/pull/2533/files#diff-e3a3e294ca71005889c57c12639ffb403ed92e6da1a6eb753b4f3d6ee533d377L400 and keep it in rz-ax

@XVilka @thestr4ng3r ?

XVilka:
Agree with @wargio

DhruvaG2000:
Closing this PR and doing this small tiny change in https://github.com/rizinorg/rizin/pull/2513 itself: 
29ac83f31620833d9d59d7c33776294fa8df6447

PR #2545, opened 2022-04-15
Add asm and analysis plugin for kvx architecture ##arch

jmaselbas:
**checklist**

KVX is VLIW architecture, kind of like a DSP, it can execute more than
one "instruction" per cycle. Each cycle a bundle of instruction is
executed which is comprised of up to 6 instructions. The general rules
for bundle are: only one BCU instruction (which comes first); only one
LSU instruction; and up to 4 ALU instructions (depending on the type).

In assembly the end of a bundle is marked by a `;;` in an empty line,
here in r2 (to make things more compact and readable) the end of bundle
is appended after the last instruction, but beware that this is not
valid kvx assembly.


wargio (commented):
I have a small problem with this code: contains some GPL code. we are trying to avoid having GPL code in the repository.
I think this should be placed in rizin-extras and not in the main repo.
@XVilka @thestr4ng3r @ret2libc what do you think?

wargio:
just to be clear, i saw that you used the values from the GPL repo to implement the decoder. i'm not 100% sure if this can be considered not a GPL code, since you only have extracted the masks/consts value

jmaselbas:
> just to be clear, i saw that you used the values from the GPL repo to implement the decoder. i'm not 100% sure if this can be considered not a GPL code, since you only have extracted the masks/consts value

Yes... the binutils `kv3-opc.c` is GPL-2.0-or-later,  `kvx-dis.c` have some part derived from binutils `kv3-opc.c` (which is GPL-3.0-or-later), `kvx-reg.c` is written by me.
The license might need to be GPL instead of LGPL.

XVilka:
I also think we should put it in extras both due to the license and architecture being quite rare/exotic. Also, when you add a new architecture - please add tests for it.

XVilka:
Is it this processor? https://www.kalrayinc.com/products/processors-many-core

jmaselbas:
> Is it this processor? https://www.kalrayinc.com/products/processors-many-core

yes it is

jmaselbas:
> I also think we should put it in extras both due to the license and architecture being quite rare/exotic. Also, when you add a new architecture - please add tests for it.

Where the tests goes ? Do you any example ? I've added a simply hello world program into radare testbin do you need the same thing?

wargio:
yes, you can add a bin in the bin repo

XVilka:
Could you please send the PR to https://github.com/rizinorg/rizin-extras instead?

XVilka:
@jmaselbas could you please send a PR to https://github.com/rizinorg/rizin-extras instead? A good example how it would look like is https://github.com/rizinorg/rizin-extras/tree/master/jaguar-gpu

XVilka:
Closed in favor of putting it into [**rizin-extras**](https://github.com/rizinorg/rizin-extras/): [**kvx**](https://github.com/rizinorg/rizin-extras/tree/master/kvx).

PR #2547, opened 2022-04-17
Fix bug in rz_copy_nbits and add unit test

Heersin:

**Detailed description**
Suppose we intend to copy `n` bits from a bitvector of length `x` into a bitvector of length `y`, if `copy` operation is valid, `rz_bv_copy_nbits` would copy as many as possible bits instead of `n` bits. 

Example bug case : 
copy 3 bits(0,1,2) into a new bitvector which has the same length. 
- assume the `src bitvector` is `0110 0101`
- expected result should be : `0000 0101` 
- actual bitvector : `0110 0101`

**Affected Code**
None. 
Currently rizin only uses it in `il_reg.c`, the `src length` ,`dst length` and `copy size` have the same value

**Test plan**
see test/unit/test_bitvector.c

### DIFF
--- a/librz/util/bitvector.c
+++ b/librz/util/bitvector.c
@@ -182,7 +182,7 @@ RZ_API ut32 rz_bv_copy(RZ_NONNULL const RzBitVector *src, RZ_NONNULL RzBitVector
 }
 
 /**
- * Copy n bits from start position of source to start position of dest
+ * Copy n bits from start position of source to start position of dest, return num of copied bits
  * \param src RzBitVector, data source
  * \param src_start_pos ut32, start position in source bitvector of copy
  * \param dst RzBitVector, destination of copy
--- a/librz/util/bitvector.c
+++ b/librz/util/bitvector.c
@@ -202,7 +202,7 @@ RZ_API ut32 rz_bv_copy_nbits(RZ_NONNULL const RzBitVector *src, ut32 src_start_p
 	}
 
 	// normal case here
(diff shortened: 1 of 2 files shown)


wargio (approved):

wargio (changes requested):
~~i think the current behavior is correct~~ Hmm. not sure

XVilka (approved):

wargio:
can you add a non zero test? like a bitvector with a number in which is copied another number?

Heersin:
> ~i think the current behavior is correct~ Hmm. not sure

I found this confused behavior when trying to copy lowest `23` bits only from a 32-bitvector to another 32-bitvector, but it copied the whole `32` bits.

Heersin:
> can you add a non zero test? like a bitvector with a number in which is copied another number?

sorry, what's a non-zero test ? Could u give me an example ?

wargio:
in your tests you always copy over a zero-filled bitv, can you add an extra test with one with a value?
```
dst = 0x12345678901235566
src = 0x19384
copy_n_bits(dst, src, 33)

Heersin:
> in your tests you always copy over a zero-filled bitv, can you add an extra test with one with a value?
> 
> ```
> dst = 0x12345678901235566
> src = 0x19384
> copy_n_bits(dst, src, 33)
> ```

Thanks, I added it in the latest commit

thestr4ng3r (approved):

PR #2557, opened 2022-04-22
RzArch

wargio:

**Detailed description**

It is time to remove RzAsm & RzAnalysis and replace them with RzArch.

### DIFF
--- a/librz/include/rz_arch.h
+++ b/librz/include/rz_arch.h
@@ -4,7 +4,9 @@
 #ifndef RZ_ARCH_H
 #define RZ_ARCH_H
 
+#include <rz_config.h>
 #include <rz_flag.h>
+#include <rz_il.h>
 #include <sdb.h>
 
 #ifdef __cplusplus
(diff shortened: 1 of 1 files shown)


librz/include/rz_arch.h
```diff
@@ -59,6 +61,245 @@ RZ_API bool rz_arch_load_platform_sdb(RZ_NONNULL RzArchPlatformTarget *t, RZ_NON
...
 
+typedef enum {
+	RZ_ARCH_INFO_CODE_ALIGN = 0, ///< code/text segment alignment
+	RZ_ARCH_INFO_DATA_ALIGN, ///< data segment alignment
+	RZ_ARCH_INFO_INSTRUCTION_MIN_SIZE, ///< minimum instruction length
+	RZ_ARCH_INFO_INSTRUCTION_MAX_SIZE, ///< maximum instruction length
+	RZ_ARCH_INFO_ADDRESS_SPACE, ///< address space bit-size
+	/* ignore */
+	RZ_ARCH_INFO_END, ///< Used only to define the max value of RZ_ARCH_INFO_* enum
+} RzArchInfo;
```
XVilka, line 72: Probably also endianess. Some architectures have multiple endianness variants, like PPC or ARM, sometimes even can switch on the fly, we should account for that.

librz/include/rz_arch.h
```diff
@@ -59,6 +61,245 @@ RZ_API bool rz_arch_load_platform_sdb(RZ_NONNULL RzArchPlatformTarget *t, RZ_NON
...
+
+typedef enum {
+	RZ_ARCH_EXTENSION_TYPE_GENERIC = 0, ///< Opcode extension generic
+	RZ_ARCH_EXTENSION_TYPE_FLOATING_POINT, ///< Opcode extension floating point
+	RZ_ARCH_EXTENSION_TYPE_VECTORIAL, ///< Opcode extension vectorial
+	RZ_ARCH_EXTENSION_TYPE_VIRTUALIZATION, ///< Opcode extension virtualization
+	RZ_ARCH_EXTENSION_TYPE_PRIVILEGED, ///< Opcode extension for instructions with privilieged execution
+	RZ_ARCH_EXTENSION_TYPE_MMX, ///< Opcode extension for multimedia instructions
```
XVilka, line 80: Isn't better to group them in something like:
- SIMD
- Vector extensions
- AUTH
?
It will be more generic.

librz/include/rz_arch.h
```diff
@@ -59,6 +61,245 @@ RZ_API bool rz_arch_load_platform_sdb(RZ_NONNULL RzArchPlatformTarget *t, RZ_NON
...
+	RZ_ARCH_OPCODE_TYPE_SIGN_EXTEND /*           */ = (1ull << 14), ///< Opcode performs a extend sign operation (cast to signed)
+
+	/* Mathematical Instructions */
+	RZ_ARCH_OPCODE_TYPE_ADD /*                   */ = (1ull << 15), ///< Opcode performs a addition (a + b)
+	RZ_ARCH_OPCODE_TYPE_SUBTRACT /*              */ = (1ull << 16), ///< Opcode performs a subtraction (a − b)
+	RZ_ARCH_OPCODE_TYPE_MULTIPLY /*              */ = (1ull << 17), ///< Opcode performs a multiplication (a * b)
+	RZ_ARCH_OPCODE_TYPE_DIVIDE /*                */ = (1ull << 18), ///< Opcode performs a division (b / a)
+	RZ_ARCH_OPCODE_TYPE_MODULO /*                */ = (1ull << 19), ///< Opcode performs a modulus, i.e. remainder of a division (b % a)
+	RZ_ARCH_OPCODE_TYPE_NEGATE /*                */ = (1ull << 20), ///< Opcode performs a negate operation, i.e inverts the sign of a value (a = -b)
+	RZ_ARCH_OPCODE_TYPE_MATH /*                  */ = (1ull << 21), ///< Opcode performs a math operation not described above (i.e. sqrt, log2, abs, etc...)
```
XVilka, line 122: We probably also should distinguish between Integer math and Real math instructions, especially in the light of RZIL.

librz/include/rz_arch.h
```diff
@@ -59,6 +61,245 @@ RZ_API bool rz_arch_load_platform_sdb(RZ_NONNULL RzArchPlatformTarget *t, RZ_NON
...
+	RZ_ARCH_OPCODE_TYPE_REGISTER_CALL = (RZ_ARCH_OPCODE_TYPE_CALL_UNKNOWN | RZ_ARCH_OPCODE_TYPE_REGISTER),
+	RZ_ARCH_OPCODE_TYPE_INDIRECT_CALL = (RZ_ARCH_OPCODE_TYPE_CALL_UNKNOWN | RZ_ARCH_OPCODE_TYPE_INDIRECT),
+	RZ_ARCH_OPCODE_TYPE_INDIRECT_REGISTER_CALL = (RZ_ARCH_OPCODE_TYPE_INDIRECT_CALL | RZ_ARCH_OPCODE_TYPE_REGISTER),
+	RZ_ARCH_OPCODE_TYPE_CONDITIONAL_CALL = (RZ_ARCH_OPCODE_TYPE_CALL | RZ_ARCH_OPCODE_TYPE_CONDITIONAL),
+	RZ_ARCH_OPCODE_TYPE_CONDITIONAL_CALL_UNKNOWN = (RZ_ARCH_OPCODE_TYPE_CALL_UNKNOWN | RZ_ARCH_OPCODE_TYPE_CONDITIONAL),
+	RZ_ARCH_OPCODE_TYPE_CONDITIONAL_RETURN = (RZ_ARCH_OPCODE_TYPE_RETURN | RZ_ARCH_OPCODE_TYPE_CONDITIONAL),
+} RzArchOpcodeType;
+
+typedef struct rz_arch_opcode_asm_t {
+	ut32 size; ///< opcode size in bytes
```
XVilka, line 183: I think it's better to compactify it here - use `ut8`.

librz/include/rz_arch.h
```diff
@@ -59,6 +61,245 @@ RZ_API bool rz_arch_load_platform_sdb(RZ_NONNULL RzArchPlatformTarget *t, RZ_NON
...
+	RzArchPluginInit init; ///< initialize plugin.
+	RzArchPluginUpdate update; ///< updates the plugin configuration.
+	RzArchPluginFini fini; ///< deinitialize plugin.
+	RzArchPluginDisassemble disassemble; ///< disassembles an instruction from the provided bytes.
+	RzArchPluginAssemble assemble; ///< assembles an instruction from the provided assembly.
+	RzArchPluginMaskOpcode mask_opcode; ///< generates a byte mask of the decoded instruction.
+	RzArchPluginInfo info; ///< provides the architecture information regarding mix/max instruction size, alignment, etc..
+	RzArchPluginFunctionPreludes func_preludes; ///< provides a list of expected masked patterns bytes which indicates if a function is about to start
+
+	/* ESIL */
```
XVilka, line 289: Move this to the end of the structure.

XVilka (changes requested):

librz/include/rz_arch.h
```diff
@@ -59,6 +61,245 @@ RZ_API bool rz_arch_load_platform_sdb(RZ_NONNULL RzArchPlatformTarget *t, RZ_NON
 RZ_API bool rz_arch_platform_init(RzArchPlatformTarget *t, RZ_NONNULL const char *arch, RZ_NONNULL const char *cpu,
 	const char *platform, RZ_NONNULL const char *platforms_dir);
```
thestr4ng3r, line 62: Probably we should start by renaming all existing `rz_arch` stuff because it does not really have much to do with the "new" `RzArch`.

librz/include/rz_arch.h
```diff
@@ -59,6 +61,245 @@ RZ_API bool rz_arch_load_platform_sdb(RZ_NONNULL RzArchPlatformTarget *t, RZ_NON
...
+	RZ_ARCH_OPCODE_TYPE_INDIRECT_REGISTER_CALL = (RZ_ARCH_OPCODE_TYPE_INDIRECT_CALL | RZ_ARCH_OPCODE_TYPE_REGISTER),
+	RZ_ARCH_OPCODE_TYPE_CONDITIONAL_CALL = (RZ_ARCH_OPCODE_TYPE_CALL | RZ_ARCH_OPCODE_TYPE_CONDITIONAL),
+	RZ_ARCH_OPCODE_TYPE_CONDITIONAL_CALL_UNKNOWN = (RZ_ARCH_OPCODE_TYPE_CALL_UNKNOWN | RZ_ARCH_OPCODE_TYPE_CONDITIONAL),
+	RZ_ARCH_OPCODE_TYPE_CONDITIONAL_RETURN = (RZ_ARCH_OPCODE_TYPE_RETURN | RZ_ARCH_OPCODE_TYPE_CONDITIONAL),
+} RzArchOpcodeType;
+
+typedef struct rz_arch_opcode_asm_t {
+	ut32 size; ///< opcode size in bytes
+	RzStrBuf buffer; ///< inline assembly string
+} RzArchOpcodeAsm;
```
Rot127, line 185: In Hexagon one instruction word (instruction packet) can encode multiple opcodes. Although `instruction word != opcode` they are treated similar in practice (single address for instruction word, executed atomically, execution commits or fails, never partially commits).
It would be nice to account for that here. Maybe a flag (`multiple_opcodes = <bool>`) and `buffer` can contain new lines (each line an opcode assembly string)

Rot127: Just recognized that this will interfere with the opcode type flags. Maybe an `instruction` struct could be added which holds x opcodes?

Rot127 (changes requested):

DMaroo (approved):

Rot127:
@wargio Another thing for the todo list would be a more detailed implementation of `analsysi.cpu`. Currently it is only represented by a string if I am not mistaken. But processors have many more properties which we might need to know in the future (see https://github.com/rizinorg/rizin/issues/2899#issuecomment-1729872777).
Especially if we ever want to do system emulation.

librz/include/rz_arch.h
Rot127: There is an additional problem of missing context.
Some instruction words only disassemble correctly if the previous instruction is valid.
Or when they are at a symbol offset.

Currently bytes at an address get decoded multiple times before anything is done (by asm, analysis, for emulation stepping etc.).
Each time it is possible that, with the exception of the instruction bytes at the address, everything else changes (e.g. buffer size, content of the buffer after the instruction bytes and their content (could be 0x00..., 0xff... or the next instruction bytes, no one knows!1!!) etc.).

Because the context changes, not even buffering helps reliably.

I think it is important that the passed data to the plugin is at a minimum nicely defined AND stays the same over multiple decoding requests for a single address.
Simply so the plugin doesn't get a random set of stuff to work with, whenever it is called at the same address.

I struggle so enormously often with this currently, that it would be a huge relive of pain.

librz/include/rz_arch.h
```diff
@@ -59,6 +61,245 @@ RZ_API bool rz_arch_load_platform_sdb(RZ_NONNULL RzArchPlatformTarget *t, RZ_NON
...
+	const char *author; ///< author name.
+	const char *features; ///< architecture features list.
+	const char *platforms; ///< architecture sdb platform name.
+	ut32 bits; ///< architecture supported bits.
+	ut32 endianness; ///< architecture supported endianness.
+	RzArchPluginInit init; ///< initialize plugin.
+	RzArchPluginUpdate update; ///< updates the plugin configuration.
+	RzArchPluginFini fini; ///< deinitialize plugin.
+	RzArchPluginDisassemble disassemble; ///< disassembles an instruction from the provided bytes.
+	RzArchPluginAssemble assemble; ///< assembles an instruction from the provided assembly.
```
Rot127, line 284: This needs another callback for plugins which don't disassemble or assemble.
E.g. rz-ghidra or similar.

Because plugins need different parameters for their analysis, we would need a general way to pass arbitrary arguments.
Or we just pass `core`. But in any way, we would need way to pass what ever the plugin needs.

wargio:
closing since it will be rewritten due the recent merge of parse, asm and analysis

PR #2576, opened 2022-05-01
Reg profile parsing

Rot127:

**Detailed description**

Refactors register profile string parsing and introduces a `RzRegProfile` to `RzReg`.

**Register profile parsing**

- Types, sizes and offsets get now parsed separately in their own functions.
- More logging added.
- The string is first parsed into a list of register definitions and alias (`RzRegProfileDef`, `RzRegProfileAlias`) and afterwards applied to `RzReg.regset`.
- Documentation was added how the register profile needs to be formatted.
- Functions are more consistent with return types (most return `bool` now instead of `Error Msg` + `NULL`).

**New register profile structs**

The register profile of `RzReg` consists of two lists with alias and definitions.
I chose lists over NULL terminated arrays because they were a little easier to handle during parsing (no memory management).
Please let me know if you would strongly prefer the NULL terminated version.

**Test plan**

The current tests are more than enough (everything burns if something is not equivalent to the previous parsing code).

**Closing issues**

None


librz/reg/profile.c
```diff
@@ -1,245 +1,452 @@
...
+ * * packed: Packed size of the register in bytes.
+ * * comment: A comment about the register.
+ * * Flags this register holds.
+ *
+ * \param reg Register struct which holds all register items.
+ * \param profile Register profile string.
+ * \return false On failure.
+ * \return true On success.
+ */
+static bool parse_reg_profile_str(RZ_BORROW RzReg *reg, const char *profile) {
```
thestr4ng3r: Better make this parse only the profile without touching any `RzReg` and only later apply it, something like:
```
static bool parse_reg_profile_str(RZ_OUT RzRegProfile *profile, const char *str) {
```

thestr4ng3r (commented):
> Reg profile strings are now enforced to be separated with tab characters.

I don't think this is a good idea, I personally would just accept `[ \t]+` as separators. But even if we decide to switch to tab-only, it will be better to do this in a later iteration because it will break external plugins and we can track down issues better to either the refactor or the format change.

librz/include/rz_reg.h
```diff
@@ -92,16 +92,38 @@ typedef enum {
 #define RZ_REG_COND_LE   12
 #define RZ_REG_COND_LAST 13
 
+typedef struct {
+	RzRegisterId role; ///< Register role (PC, Argument etc.)
+	char *alias; ///< Alias of register.
+	char *reg_name; ///< Real register name of alias.
+} RzRegProfileAlias;
```
XVilka, line 99: Missing new line

librz/reg/profile.c
```diff
@@ -1,245 +1,452 @@
...
-		int role = rz_reg_get_name_idx(tok[0] + 1);
-		return rz_reg_set_name(reg, role, tok[1])
-			? NULL
-			: "Invalid alias";
+static void rz_reg_profile_def_free(RzRegProfileDef *def) {
+	if (!def) {
+		return;
 	}
-	return "Invalid syntax";
+	if (def->name) {
```
XVilka: Just `free(def->name)` should be enough, all implementations handle `free(NULL)` properly.

librz/reg/profile.c
```diff
@@ -1,245 +1,452 @@
...
+	if (!alias) {
+		return;
 	}
-	char *has_dot = strchr(s, '.');
-	if (has_dot) {
-		*has_dot++ = 0;
-		ut64 a = strtoul(s, end, 0) << 3;
-		ut64 b = strtoul(has_dot, end, 0);
-		return a + b;
+	if (alias->reg_name) {
```
XVilka: Same.

librz/reg/profile.c
```diff
@@ -1,245 +1,452 @@
...
+ *
+ * \param item Register item whichs types are set.
+ * \param type_str The type string.
+ * \return true On success.
+ * \return false On failure.
+ */
+static bool parse_type(RZ_OUT RzRegProfileDef *def, const char *type_str) {
+	rz_return_val_if_fail(def && type_str, false);
+	char *s = strdup(type_str);
+
```
XVilka: Unnecesary empty line,

librz/reg/profile.c
```diff
@@ -1,245 +1,452 @@
...
+ * \param reg Register struct with the register profile.
+ * \param tokens List of strings of a single register definition.
+ * \return false On failure.
+ * \return true On success.
+ */
+static bool parse_def(RZ_INOUT RzReg *reg, RZ_OWN RzList *tokens) {
+	rz_return_val_if_fail(reg && tokens, false);
+
+	const char *name = rz_list_get_n(tokens, 1);
+	rz_return_val_if_fail(name, false);
```
XVilka: I am not sure it should be an assert, @wargio @thestr4ng3r  @ret2libc thoughts?

librz/reg/profile.c
XVilka: Unnecessary empty line

librz/reg/profile.c
```diff
@@ -1,245 +1,452 @@
...
-		eprintf("Ignoring duplicated register definition '%s'\n", tok[1]);
-		return NULL;
-		// return "Duplicate register definition";
+	free(s);
+	return true;
+}
+
+/**
+ * \brief Parses the size of a register definition.
+ * Sizes with . in fornt are in bits. Otherwise in bytes.
```
XVilka: A typo.

librz/reg/profile.c
```diff
@@ -1,245 +1,452 @@
...
+	reg->regset[t].maskregstype |= ((int)1 << item->sub_type);
 }
 
-#define PARSER_MAX_TOKENS 8
-RZ_API bool rz_reg_set_profile_string(RzReg *reg, const char *str) {
-	char *tok[PARSER_MAX_TOKENS];
-	char tmp[128];
-	int i, j, l;
-	const char *p = str;
+RZ_API bool rz_reg_set_reg_profile(RZ_BORROW RzReg *reg) {
```
XVilka, line 336: Add Doxygen, please.

XVilka (changes requested):
Agree with @thestr4ng3r on tabs change.

librz/reg/profile.c
thestr4ng3r: It should be a non-assert error since profiles can be user-supplied any time.

Rot127: Can even be removed. List length is checked before. So this will always result in a string.

thestr4ng3r: Then assert is fine.

librz/include/rz_reg.h
```diff
@@ -148,9 +171,10 @@ typedef struct rz_reg_flags_t {
 RZ_API void rz_reg_free(RzReg *reg);
 RZ_API void rz_reg_free_internal(RzReg *reg, bool init);
 RZ_API RzReg *rz_reg_new(void);
-RZ_API bool rz_reg_set_name(RzReg *reg, int role, const char *name);
-RZ_API bool rz_reg_set_profile_string(RzReg *reg, const char *profile);
+RZ_API bool rz_reg_set_name(RZ_BORROW RzReg *reg, RzRegisterId role, RZ_BORROW const char *name);
+RZ_API bool rz_reg_set_profile_string(RZ_BORROW RzReg *reg, const char *profile);
```
wargio: i do not think you need to deifne the borrow here, since it should be set for the return value, i guess a RZ_NONNULL will be enough.

XVilka (approved):
Looks quite good!

librz/util/str.c
```diff
@@ -3268,6 +3335,29 @@ RZ_API RzList *rz_str_split_duplist_n(const char *_str, const char *c, int n, bo
...
+ * optionally trim (aka remove spaces) the tokens and/or consider at most \p n
+ * delimiters. The result is a \p RzList with newly allocated strings for each
+ * token.
+ *
+ * \param str Input string to split
+ * \param r Delimiter regex string used to split \p str
+ * \param n If > 0 at most this number of delimiters are considered.
+ * \param trim If true each token is considered without trailing/leading whitespaces.
+ */
+RZ_API RzList *rz_str_split_duplist_n_regex(const char *_str, const char *r, int n, bool trim) {
```
XVilka: Please add ownership for the returning list.

librz/util/str.c
```diff
@@ -3228,6 +3278,23 @@ RZ_API RzList *rz_str_split_list(char *str, const char *c, int n) {
...
+ *
+ * Split a string \p str according to the regex specified in \p r and it
+ * considers at most \p n delimiters. The result is a \p RzList with pointers
+ * to the input string \p str. Each token is trimmed as well.
+ *
+ * \param str Input string to split. It will be modified by this function.
+ * \param r Delimiter regex used to split \p str
+ * \param n If > 0 at most this number of delimiters are considered.
+ */
+RZ_API RzList *rz_str_split_list_regex(char *str, const char *r, int n) {
```
XVilka: Please add ownership for the returning list.

librz/util/str.c
wargio: also RZ_NONNULL, etc..

XVilka (approved):

librz/reg/profile.c
```diff
@@ -1,142 +1,397 @@
...
+	rz_return_val_if_fail(def_lines, false);
+
+	st32 l = 0; // Line number
+	const char *line;
+	bool is_alias = false;
+	RzListIter *it;
+	RzList *toks = NULL;
+	rz_list_foreach (def_lines, it, line) {
+		++l;
+		if (strcmp(line, "") == 0) {
```
wargio: RZ_STR_ISEMPTY(line) instead of strcmp

librz/reg/profile.c
```diff
@@ -1,142 +1,397 @@
...
+			continue;
+		}
+		toks = rz_str_split_duplist_n_regex(line, "[[:blank:]]+", 0, true);
+		if (!toks || rz_list_length(toks) == 0) {
```
wargio: no need to check the null here, since `rz_list_length` accepts it and returns 0, but i guess is better to use directly `rz_list_is_empty`

librz/reg/profile.c
```diff
@@ -1,142 +1,397 @@
...
+			continue;
+		}
+		ut32 toks_len = rz_list_length(toks);
```
wargio: you can for me move this before the if and use toks_len in the if itself, so you avoid calling twice the function

librz/reg/reg.c
```diff
@@ -1,6 +1,7 @@
 // SPDX-FileCopyrightText: 2009-2020 pancake <pancake@nopcode.org>
 // SPDX-License-Identifier: LGPL-3.0-only
 
+#include "rz_list.h"
```
wargio:
```suggestion
#include <rz_list.h>
```

test/unit/test_str.c
```diff
@@ -200,13 +200,31 @@ bool test_rz_str_split_list(void) {
 	mu_assert_streq(rz_list_get_n(l, 3), "  Everyone", "fourth item");
 	rz_list_free(l);
 
+	char s1[] = "Hello  World\tAnd \t Everyone";
+	RzList *l1 = rz_str_split_duplist_n_regex(s1, "[[:blank:]]+", 0, false);
+	mu_assert_eq(rz_list_length(l1), 4, "string has been split in 4 items");
```
wargio, line 205:
```suggestion
	mu_assert_eq(rz_list_length(l1), 4, "string             has been split in 4 items");
```

wargio (approved):

PR #2572, opened 2022-04-30
Fix capstone include directory

hamarituc:
**Your checklist for this pull request**
- [x] I've read the [guidelines for contributing](https://github.com/rizinorg/rizin/blob/master/DEVELOPERS.md) to this repository
- [x] I made sure to follow the project's [coding style](https://github.com/rizinorg/rizin/blob/master/DEVELOPERS.md#code-style)
- [ ] I've documented or updated the documentation of every function and struct this PR changes. If not so I've explained why.
- [ ] I've added tests that prove my fix is effective or that my feature works (if possible)
- [ ] I've updated the [rizin book](https://github.com/rizinorg/book) with the relevant information (if needed)

**Detailed description**

Capstone requires to explicitly state the include path like `#include <capstone/capstone.h>` (see https://www.capstone-engine.org/lang_c.html). Until version 4 `#include <capstone.h>` worked as well. But from version 5 on, the former method is required (see https://github.com/capstone-engine/capstone/issues/1807).

This PR fixes the include paths such that Rizin can be built against newer versions of capstone.

**Test plan**

 * Install Capstone 5.0-rc2
 * Compile Rizin

**Closing issues**

nothing

PR #2579, opened 2022-05-02
Fix capstone include directory

hamarituc:
Fix capstone include directory

**Detailed description**

This is a second attempt of fixing capstone include directories like in #2572. But changing the branch name made it necessary to file an new pull request.

Capstone requires to explicitly state the include path like `#include <capstone/capstone.h>` (see https://www.capstone-engine.org/lang_c.html). Until version 4 `#include <capstone.h>` worked as well. But from version 5 on, the former method is required (see https://github.com/capstone-engine/capstone/issues/1807).

This PR fixes the include paths such that Rizin can be built against newer versions of capstone.

The bundled version of capstone 3 had to be patched, as the include file locations were inconsistent between the in-tree build directories an the system install directory. This issue was fixed first in capstone version 4.

**Test plan**

 * Install Capstone 5.0-rc2 system wide
 * Compile Rizin

**Closing issues**

nothing


XVilka:
@ret2libc what do you think about such an approach? Looks like the hack of adding custom headers is necessary only for the capstone-3 which is long time obsolete.

ret2libc:
> @ret2libc what do you think about such an approach? Looks like the hack of adding custom headers is necessary only for the capstone-3 which is long time obsolete.

I'm actually ok with that, if we need to slightly modify capstone-v3 (very "stable" at this point) to support future capstone version, I think it's ok.

However, I'm having problem with this change when e.g. you have capstone v4 installed system wide and you try to use `-Duse_capstone_version=v3`. It complains about `cs_regs_access` in /librz/analysis/p/analysis_x86_cs.c:1918, I think it is because it is getting the system headers instead of the local ones.

ret2libc:
Sorry for the previous comment, I had a dirty environment. LGTM

ret2libc (approved):
Thanks @hamarituc !

wargio (approved):

ret2libc (approved):

Issue #1514, opened 2021-08-24
`izzz` reports wrong addresses


ret2libc:
### Work environment

| Questions                                            | Answers
|------------------------------------------------------|--------------------
| OS/arch/bits (mandatory)                             | Fedora 33
| File format of the file you reverse (mandatory)      | Macho
| Architecture/bits of the file (mandatory)            | 64
| `rizin -v` full output, **not truncated** (mandatory)  | 3b0d11eebb3af4634741c7a22f70b6c8d4e174f5,

### Expected behavior
```
093 0x00002f3c 0x100002f3c  47  48 () ascii ,http://www.apple.com/appleca/codesigning.crl0\r
```

### Actual behavior
```
093 0x00002f3c 0x00002f3c  47  48 () ascii ,http://www.apple.com/appleca/codesigning.crl0\r
```

### Steps to reproduce the behavior
```
$ ./build-dev/binrz/rizin/rizin ./test/bins/mach0/fatmach0-3true
> [0x100000ef8]> izzz~http~codesigning
```

### Additional info
The address currently reported is wrong because there's nothing at that address, see:
```
[0x100000ef8]> pd 1 @ 0x2f3c
            0x00002f3c      add   byte [rax], al
[0x100000ef8]> px 30 @ 0x2f3c
- offset -   0 1  2 3  4 5  6 7  8 9  A B  C D  E F  0123456789ABCDEF
0x00002f3c  0000 0000 0000 0000 0000 0000 0000 0000  ................
0x00002f4c  0000 0000 0000 0000 0000 0000 0000       ..............
[0x100000ef8]> 
```

ret2libc:
After looking at the code that searches for strings... I'm not sure we can fix it in short time. The whole "string search" code is a quite convoluted that would likely require a big refactoring + unit tests to properly fix the problem and be sure things would still work.

XVilka:
Agree. It will also benefit from this work: https://github.com/rizinorg/rizin/pull/1535

XVilka:
I know the reason - it's actually quite trivial. The string search is ran using the physical address of the buf, for example in https://github.com/rizinorg/rizin/pull/1752 I did this code to search in the current address space:
```c
ut64 paddr = rz_io_v2p(core->io, addr);
int count = rz_scan_strings(bf->buf, str_list, &scan_opt, paddr, paddr + limit, encoding);
if (count <= 0) {
	rz_list_free(str_list);
	free(*name);
	return false;
}
```
[...]

PR #2590, opened 2022-05-05
RzBin: improve conversion of paddr/vaddr strings

ret2libc:

**Detailed description**

Introduce APIs such as `rz_bin_object_p2v` and `rz_bin_object_v2p` and use them in the string search API.
Moreover, consider the `bobj->boffset` before returning the found strings, to consider the case of fatmach0 and similar.

**Test plan**

Use `iz`/`izz` on both regular binaries and "fat" binaries (e.g. test/bins/mach0/fatmach0-true).

Some strings should now be shown as below:
```
61  0x000223af ---------- 4   5    .gnu_debugdata ascii    ,Lb
62  0x000223bc ---------- 4   5    .gnu_debugdata ascii   +(Ys
63  0x000223c7 ---------- 5   6    .gnu_debugdata ascii   $r+u$
```
To show that those strings do not have a place in the virtual space.

**Closing issues**

#1514

### DIFF
--- a/librz/bin/bin.c
+++ b/librz/bin/bin.c
@@ -678,7 +678,11 @@ RZ_API RzBinSection *rz_bin_get_section_at(RzBinObject *o, ut64 off, int va) {
 }
 
 /**
- * \brief Find the first binary map at offset \p off .
+ * \brief Find the last binary map at offset \p off .
+ *
+ * This function returns the last binary map that contains offset \p off,
+ * because it assumes that maps are sorted by priority, thus the last one will
+ * be the most important one.
  *
  * \param o Reference to the \p RzBinObject instance
  * \param off Address to search
@@ -692,7 +696,7 @@ RZ_API RzBinMap *rz_bin_object_get_map_at(RzBinObject *o, ut64 off, bool va) {
 	RzListIter *iter;
 	ut64 from, to;
 
-	rz_list_foreach (o->maps, iter, map) {
+	rz_list_foreach_prev(o->maps, iter, map) {
 		from = va ? rz_bin_object_addr_with_base(o, map->vaddr) : map->paddr;
 		to = from + (va ? map->vsize : map->psize);
 		if (off >= from && off < to) {
--- a/librz/bin/bin.c
+++ b/librz/bin/bin.c
@@ -649,6 +649,14 @@ RZ_API RzList *rz_bin_get_sections(RzBin *bin) {
 	return o ? (RzList *)rz_bin_object_get_sections_all(o) : NULL;
 }
 
+/**
+ * \brief Find the binary section at offset \p off.
+ *
+ * \param o Reference to the \p RzBinObject instance
+ * \param off Address to search
+ * \param va When 0 the offset \p off is considered a physical address, otherwise a virtual address
+ * \return Pointer to a \p RzBinSection containing the address
+ */
 RZ_API RzBinSection *rz_bin_get_section_at(RzBinObject *o, ut64 off, int va) {
 	RzBinSection *section;
 	RzListIter *iter;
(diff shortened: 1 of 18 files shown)


librz/bin/bin.c
```diff
@@ -649,6 +649,14 @@ RZ_API RzList *rz_bin_get_sections(RzBin *bin) {
...
 
+/**
+ * \brief Find the binary section at offset \p off.
+ *
+ * \param o Reference to the \p RzBinObject instance
+ * \param off Address to search
+ * \param va When 0 the offset \p off is considered a physical address, otherwise a virtual address
+ * \return Pointer to a \p RzBinSection containing the address
+ */
 RZ_API RzBinSection *rz_bin_get_section_at(RzBinObject *o, ut64 off, int va) {
```
XVilka, line 660: Please add also an ownership mark on the return.

librz/bin/bin.c
```diff
@@ -669,6 +677,57 @@ RZ_API RzBinSection *rz_bin_get_section_at(RzBinObject *o, ut64 off, int va) {
...
 
+/**
+ * \brief Find the first binary map at offset \p off .
+ *
+ * \param o Reference to the \p RzBinObject instance
+ * \param off Address to search
+ * \param va When false the offset \p off is considered a physical address, otherwise a virtual address
+ * \return Pointer to a \p RzBinMap containing the address
+ */
+RZ_API RzBinMap *rz_bin_object_get_map_at(RzBinObject *o, ut64 off, bool va) {
```
XVilka, line 692: Ownership

XVilka (changes requested):
Code looks good, but it causes significant amount of regressions, especially on PE files - see the broken tests.

wargio:
modify also the golang code which contains a piece of code to convert virt to phy

librz/bin/bin.c
ret2libc: added to the .h file, like the others

librz/bin/bin.c
ret2libc: added to the .h file, like the others

XVilka: Idea is to have the same markers in both `.c` and `.h` files - just look at the other APIs.

librz/core/cbin.c
```diff
@@ -2508,6 +2508,14 @@ RZ_API bool rz_core_bin_cur_section_print(RzCore *core, RzBinFile *bf, RzCmdStat
 	return rz_core_bin_sections_print(core, bf, state, &filter, hashes);
 }
 
+RZ_API bool rz_core_bin_cur_segment_print(RzCore *core, RzBinFile *bf, RzCmdStateOutput *state, RzList *hashes) {
+	rz_return_val_if_fail(core && state, false);
```
XVilka: `bf` too?

librz/bin/bobj.c
```diff
@@ -961,3 +961,71 @@ RZ_API RZ_OWN RzVector *rz_bin_object_sections_mapping_list(RzBinObject *obj) {
...
+}
+
+/**
+ * \brief Convert virtual address to offset in the file according to binary mappings
+ *
+ * \param obj Reference to \p RzBinObject
+ * \param paddr Virtual address
+ * \return Converted virtual address to offset in the file or UT64_MAX if the conversion cannot be done
+ */
+RZ_API ut64 rz_bin_object_v2p(RzBinObject *obj, ut64 vaddr) {
```
XVilka, line 1024: I suggest adding here `rz_return_val_if_fail(obj, UT64_MAX)`.

librz/bin/bobj.c
```diff
@@ -961,3 +961,71 @@ RZ_API RZ_OWN RzVector *rz_bin_object_sections_mapping_list(RzBinObject *obj) {
...
+/**
+ * \brief Convert offset in the file to all possible virtual addresses according to binary mappings
+ *
+ * \param obj Reference to \p RzBinObject
+ * \param paddr Offset in the file
+ * \return Vector containing \p ut64 values of all possible virtual addresses
+ */
+RZ_API RzVector *rz_bin_object_p2v_all(RzBinObject *obj, ut64 paddr) {
+	RzPVector *maps = rz_bin_object_get_maps_at(obj, paddr, false);
+	rz_return_val_if_fail(maps, NULL);
```
XVilka: Is this intended as an assert? Can it be NULL at the runtime?

XVilka (approved):
LGTM apart from a few minor things. The new output is so much better!

librz/bin/bobj.c
ret2libc: oh well.... actually it could be NULL. I need to return NULL when the allocation fails, so yeah i will change this.

librz/bin/bin.c
ret2libc: I see it is inconsistent... and actually that's why i don't like the approach as well. RZ_BORROW, RZ_OWN, etc. are already things that may be wrong because they are just indication for developers without any real way to "check" them automatically and duplicating the same info in two places will just lead to problems IMO, because it's easy for one of them to be out of sync.

XVilka (approved):

XVilka:
```
In file included from ../librz/include/rz_vector.h:5,
                 from ../librz/include/rz_util/rz_event.h:12,
                 from ../librz/include/rz_util.h:28,
                 from ../librz/include/rz_bin.h:4,
                 from ../librz/bin/bobj.c:6:
../librz/bin/bobj.c: In function ‘rz_bin_object_p2v_all’:
../librz/include/rz_util/rz_assert.h:112:11: error: returning ‘long long unsigned int’ from a function with return type ‘RzVector *’ {aka ‘struct rz_vector_t *’} makes pointer from integer without a cast [-Werror=int-conversion]
  112 |    return (val); \
      |           ^
../librz/bin/bobj.c:998:2: note: in expansion of macro ‘rz_return_val_if_fail’
  998 |  rz_return_val_if_fail(obj, UT64_MAX);
      |  ^~~~~~~~~~~~~~~~~~~~~
```

test/db/formats/mdmp
```diff
@@ -379,7 +379,7 @@ EXPECT=<<EOF
 0x00128fa8 0x74fc93d6 11 24 C:_Windows_System32_wow64win.dll utf16le VarFileInfo
 0x000cdc38 0x74f66066 15 32 C:_Windows_System32_wow64cpu.dll utf16le VS_VERSION_INFO
 0x000cdf84 0x74f663b2 11 24 C:_Windows_System32_wow64cpu.dll utf16le VarFileInfo
-0x00299106 0x77087534 11 12 C:_Windows_SysWOW64_kernel32.dll ascii _vfprintf_l
+0x00299106 0x77378534 11 12 C:_Windows_SysWOW64_kernel32.dll ascii _vfprintf_l
```
wargio, line 382: why?

wargio (approved):

XVilka (approved):

test/db/formats/mdmp
ret2libc: Because before it was wrong and there was no string there, now there is in the fixed address. That's the real fix

PR #2620, opened 2022-05-16
radare2 backports

ret2libc:


XVilka:
```
Run echo -e "#include <rz_util.h>\nint main(int argc, char **argv) { return rz_str_newf (\"%s\", argv[0]) != NULL? 0: 1; }" > test.c
In file included from test.c:1:
/usr/include/librz/rz_util.h:28:10: fatal error: 'rz_util/rz_axml.h' file not found
#include "rz_util/rz_axml.h"
         ^~~~~~~~~~~~~~~~~~~
1 error generated.
Error: Process completed with exit code 1.
```

XVilka:
Also missing binaries

ret2libc:
> Also missing binaries

I dobn't have any and I didn't see any attached to the r2 commit I'm backporting.

librz/core/cmd/cmd_print.c
```diff
@@ -1275,6 +1276,16 @@ static void cmd_print_fromage(RzCore *core, const char *input, const ut8 *data,
...
 		}
 	} break;
+	case 'A': // "pFA"
+	{
+		char *s = rz_axml_decode(data, size);
+		if (s) {
+			rz_cons_printf("%s", s);
+			free(s);
+		} else {
+			eprintf("Malformed object: did you supply enough data?\ntry to change the block size (see b?)\n");
```
wargio, line 1286: RZ_LOG_ERROR

librz/util/axml.c
```diff
@@ -0,0 +1,458 @@
...
+
+			return name;
+		}
+
+		if ((ut64)start > (ut64)data + data_size - sizeof(ut32) - n - 1) {
+			free(name);
+			return NULL;
+		}
+
+		memcpy(name, start, n);
```
wargio, line 154: why not use `rz_str_dup`?

librz/util/axml.c
```diff
@@ -0,0 +1,458 @@
...
+RZ_PACKED(
+	typedef struct {
+		ut16 type;
+		ut16 header_size;
+		ut32 size;
+	})
+chunk_header_t;
+
+// String pool referenced throughout the Binary XML, there must only be ONE
+RZ_PACKED(
```
wargio, line 44: these should be removed and endianness resolved.

librz/core/cmd/cmd_print.c
ret2libc: I already said so in the previous PR, this is the same as all the other cases in this function. I don't plan to address this in this PR but rather just convert it when the `pF` commands will be converted to rzshell.

librz/util/axml.c
ret2libc: because it's imported code of a new and separate feature, so I'm not that interested at this point in making this part perfect as that would not be a backport anymore but a full rewrite of a functionality.

XVilka (approved):
OK to merge but this code needs further improvements afterward.

PR #2632, opened 2022-05-24
Fix sigdb storage path and add theme doxygen

wargio:
# DO NOT SQUASH

**Detailed description**

Fix themes for cutter.

### DIFF
--- a/librz/core/canalysis.c
+++ b/librz/core/canalysis.c
@@ -5982,7 +5982,7 @@ static int core_sigdb_sorter(const RzSigDBEntry *a, const RzSigDBEntry *b) {
  * \param[in]  with_details  The reads the signature details and sets them in RzSigDBEntry
  * \return     On success a RzList containing RzSigDBEntry entries, otherwise NULL.
  */
-RZ_API RzList *rz_core_analysis_sigdb_list(RZ_NONNULL RzCore *core, bool with_details) {
+RZ_API RZ_OWN RzList *rz_core_analysis_sigdb_list(RZ_NONNULL RzCore *core, bool with_details) {
 	rz_return_val_if_fail(core, NULL);
 
 	RzList *list = NULL, *sigs = NULL;
--- a/librz/core/canalysis.c
+++ b/librz/core/canalysis.c
@@ -5970,37 +5970,63 @@ static int core_sigdb_sorter(const RzSigDBEntry *a, const RzSigDBEntry *b) {
 	return strcmp(a->short_path, b->short_path);
 }
 
-static RzList *core_load_all_signatures_from_sigdb(RzCore *core, bool with_details) {
-	RzList *sysdb = NULL, *userdb = NULL;
+/**
+ * \brief Returns all the signatures found in the default path.
+ *
+ * Scans for signature in the following paths:
+ * - home path + RZ_SIGDB
+ * - system install prefix path + RZ_SIGDB
+ * - flirt.sigdb.path user custom sigdb path
+ *
+ * \param      core          The RzCore to use.
+ * \param[in]  with_details  The reads the signature details and sets them in RzSigDBEntry
+ * \return     On success a RzList containing RzSigDBEntry entries, otherwise NULL.
+ */
+RZ_API RzList *rz_core_analysis_sigdb_list(RZ_NONNULL RzCore *core, bool with_details) {
+	rz_return_val_if_fail(core, NULL);
+
+	RzList *list = NULL, *sigs = NULL;
+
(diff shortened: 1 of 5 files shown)


librz/core/cmd/cmd_eval.c
```diff
@@ -110,15 +110,28 @@ static bool dict2keylist(void *user, const void *key, const ut64 value) {
 	return true;
 }
 
-RZ_API RZ_OWN RzList *rz_core_theme_list(RzCore *core) {
+/**
+ * \brief      Returns the list of the rizin themes.
+ *
+ * \param      core  The RzCore struct to use
+ *
```
ret2libc:
```suggestion
```

librz/core/cmd/cmd_eval.c
```diff
@@ -110,15 +110,28 @@ static bool dict2keylist(void *user, const void *key, const ut64 value) {
...
+		return NULL;
+	}
+
 	char *path = rz_path_home_prefix(RZ_THEMES);
 	if (path) {
 		list_themes_in_path(themes, path);
 		RZ_FREE(path);
 	}
 
-	path = rz_path_system(RZ_THEMES);
```
ret2libc: I think RZ_THEMES should be set correctly at build time. The problem is with cutter, right? I think that's because we are setting one prefix and then copying stuff to other dirs.

IMHO it should be possible to define the system themes/sigdb dirs at build time and those should be the very first paths to look at. `dir.prefix` should only be considered afterwards IMHO.

ret2libc:
cc @thestr4ng3r

wargio:
ok, i agree that we should use the same paths + "the portable option".

librz/core/canalysis.c
```diff
@@ -5932,37 +5932,62 @@ static int core_sigdb_sorter(const RzSigDBEntry *a, const RzSigDBEntry *b) {
...
+	sigs = rz_list_newf((RzListFree)rz_sign_sigdb_signature_free);
+	if (!sigs) {
+		return NULL;
+	}
+
 	char *system_sigdb = rz_path_system(RZ_SIGDB);
 	if (RZ_STR_ISNOTEMPTY(system_sigdb) && rz_file_is_directory(system_sigdb)) {
 		sysdb = rz_sign_sigdb_load_database(system_sigdb, with_details);
 	}
 	free(system_sigdb);
```
ret2libc: could you do that also for rz_path_user ? So one can save sigdb data also in ~/.local

librz/core/canalysis.c
```diff
@@ -5932,37 +5932,62 @@ static int core_sigdb_sorter(const RzSigDBEntry *a, const RzSigDBEntry *b) {
...
 
+	const char *prefix = rz_config_get(core->config, "dir.prefix");
```
ret2libc: I still wonder if we really need this... I don't like too much having an option to change the "prefix" considered by the tool. No strong opinion though.

librz/core/canalysis.c
wargio: hmmm makes sense.

librz/core/canalysis.c
wargio: i can add some `#if` that enables this only when portable is set.

ret2libc: but why would it be needed in that case? If rizin is compiled as portable, rz_path_system calls `rz_path_prefix`, which takes the path relative to where the rizin exe is placed.

wargio: actually it does not work.

librz/core/canalysis.c
```diff
@@ -5970,37 +5970,63 @@ static int core_sigdb_sorter(const RzSigDBEntry *a, const RzSigDBEntry *b) {
...
+ * Scans for signature in the following paths:
+ * - home path + RZ_SIGDB
+ * - system install prefix path + RZ_SIGDB
+ * - flirt.sigdb.path user custom sigdb path
+ *
+ * \param      core          The RzCore to use.
+ * \param[in]  with_details  The reads the signature details and sets them in RzSigDBEntry
+ * \return     On success a RzList containing RzSigDBEntry entries, otherwise NULL.
+ */
+RZ_API RzList *rz_core_analysis_sigdb_list(RZ_NONNULL RzCore *core, bool with_details) {
```
XVilka: Add ownership mark

XVilka (approved):

librz/core/canalysis.c
wargio: oh right, i forgot that

PR #2641, opened 2022-05-26
Fix: format/pe: baddr = 0x0000 0000

### LINKED ISSUE #2640: Fix: format PE32+ baddr
I got a PE32+ binary with baddr = 0x0 which is not "messed up".

I see no reason to set this to 0x10000

```
rz-bin -H myfile
paddr      name                                    vaddr      comment    
-------------------------------------------------------------------------
...
0x00000e88 ImageBase                               0x00000e88 0x00000000
```

### Work environment

| File format of the file you reverse (mandatory)      | PE32+
| Architecture/bits of the file (mandatory)            | ARM64
| `rizin -v` full output, **not truncated** (mandatory)  | rizin 0.3.1 @ linux-x86-64 commit: 6ef55aa63a563864774f4777dcd1a69933b62a72, build: 2021-11-24__12:29:16

### Expected behavior
```
rz-bin -I myfile
[Info]
arch     arm
cpu      N/A
baddr    0x00000000
binsz    0x000a0000
bintype  pe
bits     64
retguard false
class    PE32+
...

o myfile

[0x00001000]> om
 1 fd: 3 +0x00001000 0x00001000 - 0x00083fff r-x fmap..text
```

### Actual behavior
```
rz-bin -I myfile
[Info]
arch     arm
cpu      N/A
baddr    0x00010000
binsz    0x000a0000
bintype  pe
bits     64
retguard false
class    PE32+
...

[0x00011000]> om
 1 fd: 3 +0x00001000 0x00011000 - 0x00092fff r-x fmap..text
```

- rz-bin -H is correct
- rz-bin -I is not
- rizin "o myfile" mapping gets messed

if (!imageBase) {
//this should only happens with messed up binaries

Well, myfile is not messed up at all.

chr-chr:
This messed up loading binaries having ImageBase 0x0

 

**Detailed description**

See issue #2640

**Closing issues**

fixes #2640

### DIFF
--- a/librz/bin/format/pe/pe.c
+++ b/librz/bin/format/pe/pe.c
@@ -403,12 +403,6 @@ ut64 PE_(rz_bin_pe_get_image_base)(struct PE_(rz_bin_pe_obj_t) * bin) {
 		return 0LL;
 	}
 	imageBase = bin->nt_headers->optional_header.ImageBase;
-	if (!imageBase) {
-		// this should only happens with messed up binaries
-		// XXX this value should be user defined by bin.baddr
-		// but from here we can not access config API
-		imageBase = 0x10000;
-	}
 	return imageBase;
 }
 


chr-chr:
from the failing check:
```
[XX] db/formats/pe/virtrelocXP PE: corkami virtrelocXP.exe - entrypoint
RZ_NOPLUGINS=1 rizin -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -Qc s bins/pe/virtrelocXP.exe
-- stdout
--- expected
+++ actual
@@ -1,1 +1,1 @@
-0x10138
+0x138
```
where are those test files?

Is there anything in the PE32 spec that makes baddr=0x0 to 0x10000 ?!

wargio:
> from the failing check:
> 
> ```
> [XX] db/formats/pe/virtrelocXP PE: corkami virtrelocXP.exe - entrypoint
> RZ_NOPLUGINS=1 rizin -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -Qc s bins/pe/virtrelocXP.exe
> -- stdout
> --- expected
> +++ actual
> @@ -1,1 +1,1 @@
> -0x10138
> +0x138
> ```
> 
> where are those test files?
> 
> Is there anything in the PE32 spec that makes baddr=0x0 to 0x10000 ?!

i think is fine to just change the test to match an entry point as `0x138` the test is here: https://github.com/chr-chr/rizin/blob/dev/test/db/formats/pe/virtrelocXP#L12

ret2libc:
I would double check though. IDA loads this binary at 0x10000 as well, so there may be a reason after all.

ret2libc:
Please have a look at https://code.google.com/archive/p/corkami/wikis/PE.wiki, in particular the ImageBase section.


> is a multiple of 10000h
> can be null, under XP. In this case, the binary will be relocated to 10000h
ibnullXP.exe has a null imagebase, and relocations

chr-chr:
Hi,

I did a deeper look

> Please have a look at https://code.google.com/archive/p/corkami/wikis/PE.wiki, in particular the ImageBase section.
> 
> > can be null, under XP. In this case, the binary will be relocated to 10000h

So firstly this is not an official documentation. I have no clue about Windows but I'm not surprised if it does funny things.
I am wondering about the comment in the code: I think "messed up binaries" doesn't refer to windows but the testfile:

https://github.com/indrora/corkami/blob/master/src/PE/virtrelocXP.asm
> call [__imp__printf + 10000h]

Intentional crafted ?!

And there is documentation:

https://docs.microsoft.com/en-us/windows/win32/debug/pe-format

```
PE32 contains this additional field, which is absent in PE32+, following BaseOfCode.
Offs | size | Field
24 | 4 | BaseOfData | The address that is relative to the image base of the beginning-of-data section when it is loaded into memory.
...
28/24 | 4/8 | ImageBase:
The preferred address of the first byte of image when loaded into memory;
must be a multiple of 64 K. The default for DLLs is 0x10000000.
The default for Windows CE EXEs is 0x00010000.
The default for Windows NT, Windows 2000, Windows XP, Windows 95, Windows 98, and Windows Me is 0x00400000.
```

It says nothing about substituting 0x0

The test file is PE32 (magic 0x010b)

My files are PE32+ (magic 0x020b). They are from an arm64 UEFI firmware.
Double checked: they don't like this offset 0x10000

So which file flags we can consider?!
```
- if (!imageBase)
+ if ( !imageBase && (windows....whatever) )
```

chr-chr:
rz-bin -I myfile shows up "BaseOfData" which is absent in PE32+
Instead "ImageBase" is 8 bytes

needs to review the PE parser ....

XVilka:
> So firstly this is not official documentation. I have no clue about Windows but I'm not surprised if it does funny things.
I am wondering about the comment in the code: I think "messed up binaries" doesn't refer to windows but the test file:

Don't blindly trust documentation, it's sometimes imprecise or false, any kind of documentation. All files constructed by corkami are valid PE files and considered so by the Windows itself - they are still runnable despite all deviations from the "specification".

XVilka:
> My files are PE32+ (magic 0x020b). They are from an arm64 UEFI firmware.

Check the PE "Subsystem", you can apply this logic only if it's UEFI: https://docs.microsoft.com/en-us/windows/win32/debug/pe-format#windows-subsystem

XVilka:
Maybe some example UEFI app can be used as a test?

chr-chr:
> > My files are PE32+ (magic 0x020b). They are from an arm64 UEFI firmware.
> 
> Check the PE "Subsystem", you can apply this logic only if it's UEFI: https://docs.microsoft.com/en-us/windows/win32/debug/pe-format#windows-subsystem

yes, this can do the trick,

testfile is 0x3 IMAGE_SUBSYSTEM_WINDOWS_CUI
myfile is 0xa IMAGE_SUBSYSTEM_EFI_APPLICATION

So apply 0x10000 ("Windows CE EXE") to 0x3 only *)
or exclude # 10 to # 13 (EFI crap)

*) all listed OS are 32 bits only, can they PE32+ aka PE64?!

XVilka:
@chr-chr I think, it's better to be conservative and exclude only EFI stuff, at least for now.

chr-chr:
Well,

I asked on #winehackers for some enlightenment and I got:

> addresses below 0x10000 are invalid on Windows

At least we can drop a warning while opening such a file.

>  at very least you have a corner case here that I don't think we've 
 encountered. Of course maliciously or hand crafted binaries can trigger 
 whatever kind of corner cases, but usually we don't have to run them :-)

Great idea, let's try:

>  wine virtrelocXP.exe
> Application could not be started, or no application associated with the specified file.
> ShellExecuteEx failed: File not found.

I'd like to say if the file is broken, use hexedit. (I did, still file fails to execute)
or use -B $((0x10000)) command line option would make the test happy!

In my case -B 0 works on loading, but this fails loading via a project.
That deserves another bug report.

XVilka:
Wine is not Windows. And this file targets Windows XP in particular. Rizin is not an OS - it's a reverse engineering tool. It should be able to open any kind of file, broken or not, trying to extract as much information as possible. While for running Rizin itself we support only Windows 7 and higher, it should be able to open files for Windows 3.1, 3.11, NT 4, 95, 98, MS-DOS, FreeDOS, etc.
Thus, I suggest just adding that EFI check.

chr-chr:
> Wine is not Windows.

My aim was to ask who may know more about this.
Windows perspective is clear: don't load. There is no automagic relocating happening.
So, the test file is a rouge corner case.

> And this file targets Windows XP in particular.

It pretends to be XP from file name and IMAGE_SUBSYSTEM_WINDOWS_CUI = 3

> Rizin is not an OS - it's a reverse engineering tool. It should be able to open any kind of file, broken or not, trying to extract as much information as possible.

Of course, sir!

And here the guess for a XP would be 0x00400000.
0x10000 would be CE "IMAGE_SUBSYSTEM_WINDOWS_CE_GUI " = 9 ?!

What I don't like is that this magic/guess relocation happens without any notice.
Reminds me to the comment:
```
//this should only happens with messed up binaries
//XXX this value should be user defined by bin.baddr
//but from here we can not access config API
```
I agree, this happens at the wrong location. The legit way for this corner case is setting the address like the -B option does.

> Thus, I suggest just adding that EFI check.

If a file is rouge we can't trust the subsystem field either.

I'd like to drop this into PE_(read_nt_headers)():
```
if (headers->optional_header.ImageBase < 0x10000) {
		eprintf("ImageBase below 0x10000 is invalid on Windows target. Use the -B option to relocate.\n");
		eprintf("Preferred addresses are:\n");
		eprintf("0x10000000 DLLs\n");
		eprintf("0x00010000 Windows CE EXEs\n");
		eprintf("0x00400000 Windows NT, 2000, XP, 95, 98, and Me\n");
}
```
This would handle plus 0xfffe corner cases and a good hint to the user.

wargio:
If you want to add that check, i would suggest to use it and throw an error.
```c
if (headers->optional_header.ImageBase < 0x10000) {
	RZ_LOG_ERROR("ImageBase value is below 0x10000, which is invalid on Windows target. Please set the base address to relocate.\n"
					"Preferred addresses are:\n"
					"- 0x10000000 DLLs\n"
					"- 0x00010000 Windows CE EXEs\n"
					"- 0x00400000 Windows NT, 2000, XP, 95, 98, and Me\n");
	return false;
}
```

ret2libc:
> If you want to add that check, i would suggest to use it and throw an error.
> 
> ```c
> if (headers->optional_header.ImageBase < 0x10000) {
> 	RZ_LOG_ERROR("ImageBase value is below 0x10000, which is invalid on Windows target. Please set the base address to relocate.\n"
> 					"Preferred addresses are:\n"
> 					"- 0x10000000 DLLs\n"
> 					"- 0x00010000 Windows CE EXEs\n"
> 					"- 0x00400000 Windows NT, 2000, XP, 95, 98, and Me\n");
> 	return false;
> }
> ```

I'm not sure I like this. @chr-chr is correct saying that PE are not just windows. So failing to load a PE because windows does not load it is not good IMHO.

Anyway, my question is: what is the problem with having the binary loaded by default at 0x100000 (or `-B` if specified?)? Even if your UEFI bin (PE) has ImageBase=0, it should be relocatable, right? I think the problem I saw in the PE binary is that some relocations are not correctly applied.

XVilka:
I think PE binaries for UEFI are made unrelocatable to simplify compilation
and loading process.

chr-chr:
> Anyway, my question is: what is the problem with having the binary loaded by default at 0x100000 (or -B if specified?)?

That forced me to use -B 0 . The testfile can be happily use -B because it's test ends.

But I want to use the Ps feature. And the Po feature.
So this and that makes it showstopper to me:

https://github.com/rizinorg/cutter/issues/2541

Fixing the -B option is actually a greater challenge.

To sum this up:

I was expecting the file gets read in as it is (rz-bin -H header) and I was piss^Wwondering about this offset (rz-bin -I )
Windows would not load that testfile but rizin can. Even at a wrong and illegal location. No reason to fail.
I'd like to add some hinting of course but at the right location. My printf's are just a proposal as they would clobber rz-bin output.
Can this just added to rz-bin -I/-H output as a note/hint?

Also best location for an implicit setting ImageBase would be in librz/bin/p/bin_pe.inc
in load_buffer() or maps() ....

chr-chr:
I put some hacking on it:

https://github.com/rizinorg/rizin/compare/dev...chr-chr:dev.fixing

and by the way: the 'o' command was bugged aswell

PR #2652, opened 2022-06-01
RzCore: add 'aez' help reference

ret2libc:


XVilka (approved):
I think it was done on purpose since RzIL is not yet well integrated into the analysis loop.
Anyway, OK for me.

wargio (approved):

wargio:
isn't better to just use nnewshell?

ret2libc:
We haven't converted the whole ae commands yet. Also, since we mention it in the release notes it should be accessible

PR #2658, opened 2022-06-04
Dev.fixing2

### LINKED ISSUE #2640: Fix: format PE32+ baddr

chr-chr:

**Detailed description**

See issue #2640

and discussion on first pull request:
https://github.com/rizinorg/rizin/pull/2641

There are valid PE files having ImageBase=0x0 (UEFI)
but the PE loader set it to 0x10000 which is not correct:
"this should only happens with messed up binaries".

Also manual overriding to 0x0 fails (o command)

**Closing issues**

fixes #2640
and some other issues that raised up while diving trough code


librz/bin/format/pe/pe.c
```diff
@@ -403,12 +403,7 @@ ut64 PE_(rz_bin_pe_get_image_base)(struct PE_(rz_bin_pe_obj_t) * bin) {
 		return 0LL;
 	}
 	imageBase = bin->nt_headers->optional_header.ImageBase;
-	if (!imageBase) {
```
ret2libc, line 406: IMHO rizin should done the best by default. We could check the system in `bin->nt_headers->optional_header.Subsystem` and if the value is a windows one we set the imagebase=0x10000, otherwise use just the binary value.

librz/bin/p/bin_pe.inc
```diff
@@ -26,6 +26,21 @@ static bool load_buffer(RzBinFile *bf, RzBinObject *obj, RzBuffer *buf, Sdb *sdb
...
+				RZ_LOG_INFO("%s",
+					"ImageBase below 0x10000 is invalid on Windows.\n"
+					"Preferred addresses are:\n"
+					"0x10000000 DLLs\n"
+					"0x00010000 Windows CE EXEs\n"
+					"0x00400000 Windows NT, 2000, XP, 95, 98, and Me\n"
+					"Use the -B option to set base address.\n"
+				);
+			}
+		}
```
ret2libc, line 43: I would avoid this TBH, it is very system dependent and one should know what to do with its own files to reverse. Moreover, this is just windows specific and if I'm analyzing a EFI app I don't want to see this.

test/db/formats/pe/virtrelocXP
```diff
@@ -7,6 +7,7 @@ RUN
 
 NAME=PE: corkami virtrelocXP.exe - entrypoint
 FILE=bins/pe/virtrelocXP.exe
+ARGS=-B 0x10000
```
ret2libc, line 10: If we make rizin work well by default, this is not needed anymore.

ret2libc (changes requested):

librz/bin/format/pe/pe.c
chr-chr: If a file is bogus u can not rely on the subsystem flag either.

The test file is bogus. Windows does not load it.

test/db/formats/pe/virtrelocXP
chr-chr: What is default?

For XP it's 0x00400000

librz/bin/p/bin_pe.inc
chr-chr: I don't insist in keeping that

librz/bin/format/pe/pe.c
chr-chr: // but from here we can not access config API

Also this is the wrong location to handle this

ret2libc: > If a file is bogus u can not rely on the subsystem flag either.

True but you can provide useful defaults for the good case. If you analyze a specially crafted binary you can still use `-B` when loading the binary.

> 
> The test file is bogus. Windows does not load it.

The test file is not bogus AFAIK because it's a binary from Corkami.

test/db/formats/pe/virtrelocXP
ret2libc: AFAIU that is the preferred value chosen by the linker, not the default value.

librz/bin/format/pe/pe.c
chr-chr: // this should only happens with messed up binaries

if ( strcmp(filename, "virtrelocXP.exe") == 0) { ... }

librz/bin/p/bin_pe.inc
chr-chr: At least this is the right location for some auto magic.

librz/bin/format/pe/pe.c
chr-chr: > 
> > The test file is bogus. Windows does not load it.
> 
> The test file is not bogus AFAIK because it's a binary from Corkami.

Look at it's disassembly.

 * It set ImageBase to 0x0 but "call [0x10208]" the offset was crafted.
 * It's first instruction "push 0x150" is wrong if we set ImageBase to 0x10000 (to my understanding should be 0x10150, 1st arg to printf?!)
 * windows raises middle finger (I tried it)
 * the file is "synthetic"
 
https://github.com/indrora/corkami/blob/master/src/PE/virtrelocXP.asm

ret2libc: > > > The test file is bogus. Windows does not load it.
> > 
> > 
> > The test file is not bogus AFAIK because it's a binary from Corkami.
> 
> Look at it's disassembly.
> 
> * It set ImageBase to 0x0 but "call [0x10208]" the offset was crafted.
> * It's first instruction "push 0x150" is wrong if we set ImageBase to 0x10000 (to my understanding should be 0x10150, 1st arg to printf?!)

IDA disassembles that correctly. There is likely a relocation that relocates 0x150 to 0x10150.

> * windows raises middle finger (I tried it)

Yes, maybe there is some protection or new check in newer windows version that prevents that from loading. I bet if you spawn a Windows XP VM though it will run.

> * the file is "synthetic"
> 
> https://github.com/indrora/corkami/blob/master/src/PE/virtrelocXP.asm

The fact that it is synthetic does not really matter.


Anyway, I agree probably this is not the best place. I suggest the get_image_base in bin_pe plugin returns the value in the PE header, but somewhere else we set anyway the 0x10000 for windows binaries in case the user has not set anything else. I think we can rely on the value in RzBinObjectLoadOptions.baseaddr to check the baseaddr that is set.

ret2libc:
Hey @chr-chr how is it going on this? Do you have a clear path forward?

XVilka:
@chr-chr hi, had there been any updates?

PR #2659, opened 2022-06-04
RzAnalysis: enable io.cache to make mem operations work in RzIL

ret2libc:

**Detailed description**

Without this patch users are forced to set io.cache=true and they may not know about it and not even realize it is needed (because all mem write operation actually seem to succeed even if the data is actually not stored anywhere).

Instead of just adding warnings here and there, I think the user experience should be the best out-of-the-box, so whenever you do `aezi` io.cache should be set to true, so that caching in IO is eanbled and mem write/read operations can work.

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/librz/analysis/il/analysis_il.c
+++ b/librz/analysis/il/analysis_il.c
@@ -105,6 +105,8 @@ RZ_API RZ_OWN RzAnalysisILVM *rz_analysis_il_vm_new(RzAnalysis *a, RZ_NULLABLE R
 	if (!r) {
 		goto ruby_pool;
 	}
+	a->iob.io->cached |= RZ_PERM_RW;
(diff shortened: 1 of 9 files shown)


XVilka (changes requested):
Too many changes for the release in the RzIL code, also it breaks the bit.
Since for now, RzIL is not really exposed to the user until we port the majority of architectures, I suggest postponing this change to 0.5.0 instead.

```
test_rz_io_map_delERR
[XX] Fail at line 479: read before map: expected true, got false
test_rz_io_map_del_for_fdOK
test_rz_io_map_del_on_closeOK
test_rz_io_map_del_on_close_allERR
[XX] Fail at line 622: read after unmap: expected true, got false
```

PR #2660, opened 2022-06-04
Write logs to INFO when loading a library instead of DEBUG

wargio:

**Detailed description**

Sometimes is complicated to understand why a library is not being loaded, before you could set RZ_DEBUG=1 and see the messages, but now they are all hidden under RZ_LOG_DEBUG which is available only in debug builds.
This PR changes most of the RZ_LOG_DEBUG to RZ_LOG_INFO.

### DIFF
--- a/librz/util/lib.c
+++ b/librz/util/lib.c
@@ -254,7 +254,7 @@ RZ_API int rz_lib_close(RzLib *lib, const char *file) {
 				ret = p->handler->destructor(p,
 					p->handler->user, p->data);
 			}
-			eprintf("Unloaded %s\n", p->file);
+			RZ_LOG_INFO("Unloaded %s\n", p->file);
 			free(p->file);
 			rz_list_delete(lib->plugins, iter);
 			return ret;
@@ -349,7 +349,7 @@ RZ_API int rz_lib_open_ptr(RzLib *lib, const char *file, void *handler, RzLibStr
 		free(mm0);
 		free(mm1);
 		if (mismatch) {
-			eprintf("Module version mismatch %s (%s) vs (%s)\n",
+			RZ_LOG_WARN("Module version mismatch %s (%s) vs (%s)\n",
 				file, stru->version, RZ_VERSION);
 			if (stru->pkgname) {
 				const char *dot = strchr(stru->version, '.');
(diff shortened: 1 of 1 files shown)


ret2libc (commented):
Maybe only some of these are worth of being INFO/ERR/WARN ? I don't want to print too much stuff to the user...

librz/util/lib.c
```diff
@@ -61,7 +61,7 @@ RZ_API void *rz_lib_dl_open(const char *libname) {
 		ret = dlopen(NULL, RTLD_NOW);
 	}
 	if (!ret) {
-		RZ_LOG_DEBUG("rz_lib_dl_open: error: %s (%s)\n", libname, dlerror());
+		RZ_LOG_INFO("rz_lib_dl_open: error: %s (%s)\n", libname, dlerror());
```
ret2libc: is it an ERROR ?

librz/util/lib.c
```diff
@@ -80,7 +80,7 @@ RZ_API void *rz_lib_dl_open(const char *libname) {
 	ret = LoadLibrary(libname_);
 	free(libname_);
 	if (!ret) {
-		RZ_LOG_DEBUG("rz_lib_dl_open: error: %s\n", libname);
+		RZ_LOG_INFO("rz_lib_dl_open: error: %s\n", libname);
```
ret2libc: error?

ret2libc (commented):
some of these may be ERROR/WARN rather than generic INFO (if not DEBUG)

wargio:
> Maybe only some of these are worth of being INFO/ERR/WARN ? I don't want to print too much stuff to the user...

i think is better to keep them hidden.

XVilka (approved):

librz/util/lib.c
```diff
@@ -209,7 +209,7 @@ RZ_API int rz_lib_run_handler(RzLib *lib, RzLibPlugin *plugin, RzLibStruct *symb
 		RZ_LOG_DEBUG("PLUGIN LOADED %p fcn %p\n", h, h->constructor);
 		return h->constructor(plugin, h->user, symbol->data);
 	}
-	RZ_LOG_DEBUG("Cannot find plugin constructor\n");
+	RZ_LOG_INFO("Cannot find plugin constructor\n");
```
ret2libc: RZ_LOG_WARN

ret2libc (approved):

PR #2669, opened 2022-06-06
Bump version to v0.5.0 for development

ret2libc:


wargio (changes requested):
Let's wait for the release before bumping

XVilka (approved):

ret2libc:
@wargio I would go ahead with this. At this point we have already merged dev in stable, so the rest of the fixes need to be cherry-picked and we can move forward with dev.

wargio (approved):

PR #2683, opened 2022-06-08
Refactor thread code and add RzThreadQueue

wargio:

**Detailed description**

This is a big cleanup of the RzThread C API, also implements RzThreadQueue.


librz/util/thread.h
```diff
@@ -43,15 +43,11 @@
...
-
+#define RZ_TH_RET_T  void *
 #else
 #error Threading library only supported for pthread and w32
 #endif
 
-#ifdef __cplusplus
-extern "C" {
-#endif
-
```
ret2libc: why?

librz/include/rz_th.h
```diff
@@ -55,6 +50,12 @@ RZ_API int rz_th_lock_tryenter(RzThreadLock *thl);
 RZ_API int rz_th_lock_enter(RzThreadLock *thl);
 RZ_API int rz_th_lock_leave(RzThreadLock *thl);
 RZ_API void *rz_th_lock_free(RzThreadLock *thl);
+#define rz_th_lock_guard(lock, x) \
+	do { \
+		rz_th_lock_enter(lock); \
+		x; \
+		rz_th_lock_leave(lock); \
+	} while (0)
```
ret2libc: where is this used?

librz/util/thread.h
wargio: because is a private header.

wargio: also it did not have an end-guard. i suspect it was copied from somewhere else.

librz/include/rz_th.h
thestr4ng3r: I don't think this is a good idea. If you e.g. `return` or `break` inside of it, it will keep the lock. In C++ it's easy to do with RAII, but in C I fear there is no nice way.

librz/include/rz_th.h
```diff
@@ -5,44 +5,39 @@
...
 typedef struct rz_th_t RzThread;
-
-#define RZ_THREAD_POOL_ALL_CORES (0)
-
-typedef struct rz_th_pool_t {
-	size_t size;
-	RzThread **threads;
-} RzThreadPool;
+typedef struct rz_th_pool_t RzThreadPool;
+typedef struct rz_th_queue_t RzThreadQueue;
```
thestr4ng3r, line 27: Please add some docs on what exactly this queue is. I.e. "a thread-safe queue that can be listened on from multiple threads" for example.
Also please document the functions describing exactly the interesting parts of the expected behavior, e.g. when you pop, will it block and wait or will it return immediately, what happens when multiple threads pop concurrently, when the queue is empty, etc.

librz/util/thread.c
```diff
@@ -24,45 +26,43 @@
...
-		th->running = false;
-		rz_th_lock_enter(th->lock);
-	} while (ret);
-	rz_th_lock_leave(th->lock);
+#endif
+
+	RzThread *th = (RzThread *)_th;
+	RzThreadStatus status = RZ_TH_STATUS_LOOP;
+
+	while (status == RZ_TH_STATUS_LOOP) {
```
thestr4ng3r: The looping-"feature" should be removed from `RzThread` too since it has absolutely no benefits over just putting a while loop in `th->function`. On the contrary actually: by removing it, we will de-occupy the return value of `th->function` to be used by the user of `RzThread` for any purpose.

thestr4ng3r (changes requested):
Love it, this is a long overdue step in the right direction!

librz/include/rz_th.h
wargio: yeah i thought about this... there is no easy way to do this in C unfortunately

librz/util/thread.c
wargio: i agree/disagree. thread functions for me should be returning `void`.
i did not remove it because it required to rewrite the code related to the rtr server which i do not want to handle now.

test/unit/test_threads.c
```diff
@@ -10,14 +10,16 @@ bool test_thread_pool_cores(void) {
```
XVilka, line 12: You added Queue API, please add a unit test for it as well

XVilka (changes requested):

librz/util/thread.h
ret2libc: oh right it's in librz/util, ok!

librz/util/thread.c
thestr4ng3r: Ok, if it's not an easy change then it's ok for me to leave in for now. But please maybe mark with RZ_DEPRECATED where appropriate.

librz/include/rz_th.h
wargio: i removed it.

librz/util/thread.c
wargio: i think i will probably just rewrite the code, to make everybody happy.
but just to be sure, do we agree to return `void` and make the API consistent? or i need to return a pointer and add the return value to the structure?

thestr4ng3r: Imo `void *`. But technically it does not have to be added to the structure but can instead be returned directly when doing the join, see for example https://git.sr.ht/~thestr4ng3r/chiaki/tree/420809b24ef7bb8510a13d9a4c02ac260a9841a7/item/lib/src/thread.c#L59-73

thestr4ng3r: Interestingly, libuv, which provides similar platform-agnostic wrappers, does not use the native return values in `uv_thread_join()`: http://docs.libuv.org/en/v1.x/guide/threads.html#core-thread-operations

wargio: i have added `rz_th_retv` because the join one is not compatible on how the pool code works (i.e. a user will never be able to retrieve the return value).
i also changed it to `void *` as return type.

thestr4ng3r: I see what you mean. However I think you should not store the return value **inside** the thread, but rather return it from the outer thread function and retrieve it inside the join and then store away in the struct there, to avoid any potential synchronization issues and enforce that the thread has been joined when the return value is accessed from the outside.
From what I understand, the `pthread_exit()` can also be removed completely since it is useless when there is no code after it (like calling `exit(0)` at the end of a main function).

wargio: - regarding pthread_exit, you are right, we can avoid calling it.
- regarding where to store the return value, i disagree. that value is meant to not be accessed in a thread-safe way. even in your code the issue is the same on windows side. i think we can keep it in this way, otherwise i need to add something like a RzList just to get all return values..

thestr4ng3r: Oh you are perfectly right, I read my own code wrong there. Then it's fine and the caller has to make sure to join before accessing it.

thestr4ng3r (approved):

librz/core/rtr.c
```diff
@@ -877,16 +883,15 @@ RZ_API void rz_core_rtr_cmd(RzCore *core, const char *input) {
 			eprintf("This is experimental and probably buggy. Use at your own risk\n");
 		} else {
 			// TODO: use tasks
-			RapThread *RT = RZ_NEW0(RapThread);
-			if (RT) {
-				RT->core = core;
-				RT->input = strdup(input + 1);
+			RapThread *rap_th = RZ_NEW0(RapThread);
+			if (rap_th) {
```
XVilka: Reverse the condition?

librz/core/rtr.c
```diff
@@ -877,16 +883,15 @@ RZ_API void rz_core_rtr_cmd(RzCore *core, const char *input) {
 			eprintf("This is experimental and probably buggy. Use at your own risk\n");
 		} else {
 			// TODO: use tasks
```
XVilka, line 885: Maybe unnecessary  TODO now?

librz/include/rz_th.h
```diff
@@ -66,11 +58,18 @@ RZ_API size_t rz_th_physical_core_number();
...
-RZ_API bool rz_th_pool_wait_async(RZ_NONNULL RzThreadPool *pool);
-RZ_API bool rz_th_pool_kill(RZ_NONNULL RzThreadPool *pool, bool force);
-RZ_API bool rz_th_pool_kill_free(RZ_NONNULL RzThreadPool *pool);
+RZ_API bool rz_th_pool_kill(RZ_NONNULL RzThreadPool *pool);
+RZ_API size_t rz_th_pool_size(RZ_NULLABLE RzThreadPool *pool);
+
+RZ_API RZ_OWN RzThreadQueue *rz_th_queue_new(size_t max_size, RZ_NULLABLE RzListFree qfree);
+RZ_API void rz_th_queue_free(RZ_NULLABLE RzThreadQueue *queue);
+RZ_API bool rz_th_queue_push(RZ_NONNULL RzThreadQueue *queue, RZ_NONNULL void *user, bool tail);
+RZ_API void *rz_th_queue_pop(RZ_NONNULL RzThreadQueue *queue, bool tail);
```
XVilka: For these that return `void *` please note the ownership

librz/util/thread.c
```diff
@@ -197,107 +234,151 @@ RZ_API bool rz_th_setaffinity(RzThread *th, int cpuid) {
...
+/**
+ * \brief Yield the processor
+ *
+ * \return On success returns true, otherwise false
+ */
+RZ_API bool rz_th_yield(void) {
+#if __WINDOWS__
+	return SwitchToThread() != 0;
+#else
+	// sched_yield is not available everywhere.
```
XVilka, line 323: Maybe we could check the presence in Meson? cc @ret2libc

librz/util/thread.c
```diff
@@ -411,27 +497,16 @@ RZ_API bool rz_th_pool_add_thread(RZ_NONNULL RzThreadPool *pool, RZ_NONNULL RzTh
...
-			RZ_LOG_DEBUG("thread: started thread %u\n", i);
-			rz_th_start(pool->threads[i], enable);
-			started = true;
-		}
-	}
-	if (!started) {
-		RZ_LOG_ERROR("thread: cannot start thread pool when there are no threads in it\n");
-	}
-	return started;
+RZ_API RZ_OWN RzThread *rz_th_pool_get_thread(RZ_NONNULL RzThreadPool *pool, size_t index) {
```
XVilka: Are you sure it's not `RZ_BORROW`?

librz/util/thread.c
```diff
@@ -447,80 +522,189 @@ RZ_API bool rz_th_pool_wait(RZ_NONNULL RzThreadPool *pool) {
...
- * \param  th The thread to get the user pointer from
+ * \param  queue The RzThreadQueue to push to
+ * \param  tail  When true, pops the element from the tail, otherwise from the head
  *
- * \return user pointer set by the rz_th_new user parameter
+ * \return On success returns a valid pointer, otherwise NULL
  */
-RZ_API void *rz_th_get_user(RzThread *th) {
-	return th->user;
+RZ_API void *rz_th_queue_pop(RZ_NONNULL RzThreadQueue *queue, bool tail) {
```
XVilka: Please add the ownership marking

librz/util/thread.c
```diff
@@ -447,80 +522,189 @@ RZ_API bool rz_th_pool_wait(RZ_NONNULL RzThreadPool *pool) {
...
+
+/**
+ * \brief  Removes an element from the queue, but yields the thread till not empty.
+ *
+ * \param  queue The RzThreadQueue to push to
+ * \param  tail  When true, pops the element from the tail, otherwise from the head
+ *
+ * \return On success returns a valid pointer, otherwise NULL
+ */
+RZ_API void *rz_th_queue_wait_pop(RZ_NONNULL RzThreadQueue *queue, bool tail) {
```
XVilka: Same

librz/util/thread.c
```diff
@@ -447,80 +522,189 @@ RZ_API bool rz_th_pool_wait(RZ_NONNULL RzThreadPool *pool) {
...
+
+/**
+ * \brief  Returns true if the queue is empty (thread-safe)
+ *
+ * \param  queue The RzThreadQueue to check
+ *
+ * \return When empty returns true, otherwise false
+ */
+RZ_API bool rz_th_queue_is_empty(RZ_NONNULL RzThreadQueue *queue) {
+	rz_return_val_if_fail(queue, NULL);
```
XVilka: `false`?

librz/util/thread.c
```diff
@@ -447,80 +522,189 @@ RZ_API bool rz_th_pool_wait(RZ_NONNULL RzThreadPool *pool) {
...
+
+/**
+ * \brief  Returns true if the queue is full and when the size is not RZ_THREAD_QUEUE_UNLIMITED (thread-safe)
+ *
+ * \param  queue The RzThreadQueue to check
+ *
+ * \return When full returns true, otherwise false
+ */
+RZ_API bool rz_th_queue_is_full(RZ_NONNULL RzThreadQueue *queue) {
+	rz_return_val_if_fail(queue, NULL);
```
XVilka: `false`?

XVilka (changes requested):

librz/include/rz_th.h
wargio: the ownership depends on what you put, i guess it should always be RZ_OWN ?

librz/util/thread.c
wargio: Yes, you are right.

librz/core/rtr.c
wargio: that code is old. i would not even change it since it probably needs to be tested and rewritten.

librz/core/rtr.c
wargio: i would not change the code now since it needs to be revisited and probably rewritten

librz/include/rz_th.h
```diff
@@ -1,76 +1,78 @@
+// SPDX-FileCopyrightText: 2009-2017 pancake <pancake@nopcode.org>
+// SPDX-FileCopyrightText: 2021-2022 deroad <wargio@libero.it>
+// SPDX-License-Identifier: LGPL-3.0-only
+
 #ifndef RZ_TH_H
 #define RZ_TH_H
 
 #ifdef _GNU_SOURCE
```
XVilka, line 8: These lines are a bit weird, lol.

librz/util/thread.c
XVilka: https://man7.org/linux/man-pages/man2/sched_yield.2.html

librz/util/thread_cond.c
```diff
@@ -19,31 +20,34 @@ RZ_API RzThreadCond *rz_th_cond_new(void) {
...
+	rz_return_if_fail(cond);
 #if HAVE_PTHREAD
 	pthread_cond_broadcast(&cond->cond);
 #elif __WINDOWS__
 	WakeAllConditionVariable(&cond->cond);
 #endif
 }
 
-RZ_API void rz_th_cond_wait(RzThreadCond *cond, RzThreadLock *lock) {
+RZ_API void rz_th_cond_wait(RZ_NONNULL RzThreadCond *cond, RZ_NONNULL RzThreadLock *lock) {
```
XVilka, line 63: Please add Doxygen.

librz/util/thread_cond.c
```diff
@@ -19,31 +20,34 @@ RZ_API RzThreadCond *rz_th_cond_new(void) {
...
+	rz_return_if_fail(cond);
 #if HAVE_PTHREAD
 	pthread_cond_signal(&cond->cond);
 #elif __WINDOWS__
 	WakeConditionVariable(&cond->cond);
 #endif
 }
 
-RZ_API void rz_th_cond_signal_all(RzThreadCond *cond) {
+RZ_API void rz_th_cond_signal_all(RZ_NONNULL RzThreadCond *cond) {
```
XVilka, line 48: Please add a function description

librz/util/thread_cond.c
```diff
@@ -19,31 +20,34 @@ RZ_API RzThreadCond *rz_th_cond_new(void) {
 	return cond;
 }
 
-RZ_API void rz_th_cond_signal(RzThreadCond *cond) {
+RZ_API void rz_th_cond_signal(RZ_NONNULL RzThreadCond *cond) {
```
XVilka, line 34: Please add a function description

librz/util/thread_lock.c
```diff
@@ -1,74 +1,70 @@
...
-	return pthread_mutex_lock(&thl->lock);
+	pthread_mutex_lock(&thl->lock);
 #elif __WINDOWS__
 	EnterCriticalSection(&thl->lock);
-	return 0;
 #endif
 }
 
-RZ_API int rz_th_lock_tryenter(RzThreadLock *thl) {
+RZ_API bool rz_th_lock_tryenter(RZ_NONNULL RzThreadLock *thl) {
```
XVilka, line 61: Please add a function description

librz/util/thread_sem.c
```diff
@@ -76,15 +77,17 @@ RZ_API void rz_th_sem_free(RzThreadSemaphore *sem) {
 	free(sem);
 }
 
-RZ_API void rz_th_sem_post(RzThreadSemaphore *sem) {
+RZ_API void rz_th_sem_post(RZ_NONNULL RzThreadSemaphore *sem) {
```
XVilka, line 97: Please add a function description

librz/util/thread_sem.c
```diff
@@ -76,15 +77,17 @@ RZ_API void rz_th_sem_free(RzThreadSemaphore *sem) {
...
+	rz_return_if_fail(sem);
 #if HAVE_PTHREAD
 	sem_post(sem->sem);
 #elif __WINDOWS__
 	ReleaseSemaphore(sem->sem, 1, NULL);
 #endif
 }
 
-RZ_API void rz_th_sem_wait(RzThreadSemaphore *sem) {
+RZ_API void rz_th_sem_wait(RZ_NONNULL RzThreadSemaphore *sem) {
```
XVilka, line 111: Same

XVilka (approved):
Looks good apart from a few missing Doxygen things.

@GustavoLCR: could you please take a look if it's OK on Windows ARM?

@devnexen could you please help us to check if it works on Haiku and DragonFlyBSD?

devnexen:
works fine.

GustavoLCR:
Works on windows on arm

librz/util/thread_pool.c
```diff
@@ -0,0 +1,201 @@
...
+ *
+ * \return true if managed to kill all threads, otherwise false
+ */
+RZ_API bool rz_th_pool_kill(RZ_NONNULL RzThreadPool *pool) {
+	rz_return_val_if_fail(pool, false);
+	bool has_exited = false;
+	for (ut32 i = 0; i < pool->size; ++i) {
+		if (pool->threads[i]) {
+			RZ_LOG_DEBUG("thread: killing thread %u\n", i);
+			rz_th_kill(pool->threads[i]);
```
GustavoLCR, line 184: Forcefully killing threads is not a good idea and makes the entire process enter an undefined state. It would be best to have a way to signal the thread pool thread to exit normally.

wargio: that is up to the person using the code. ofc, you should not normally do this.

librz/socket/run.c
```diff
@@ -719,9 +719,12 @@ static int redirect_socket_to_stdio(RzSocket *sock) {
 }
 
 #if __WINDOWS__
-static RzThreadFunctionRet exit_process(RzThread *th) {
+static void *exit_process(void *user) {
+	int timeout = (int)(void *)user;
+	rz_sys_sleep(timeout);
 	// eprintf ("\nrz_run: Interrupted by timeout\n");
```
XVilka: Maybe use `RZ_LOG_DEBUG()` here?

XVilka (approved):

XVilka:
@wargio I think can be merged after the rebase. Also, please convert that `eprintf` into `RZ_LOG_DEBUG()`. Android failures are unrelated, we can open an issue to track them.

XVilka (approved):

PR #2709, opened 2022-06-19
CMake: Fix `_rizin_cmake_path`

kazarmy:

**Detailed description**

When building rz-ghidra with a local non-standard Rizin installation, i.e.:

```bash
Rizin_DIR=~/bin/rizin-git/lib/x86_64-linux-gnu/cmake/Rizin/ cmake -DCMAKE_INSTALL_PREFIX=~/.local ..
```

the build failed with the following error:

```
-- Could NOT find rz_core (missing: rz_core_DIR)
CMake Error at CMakeLists.txt:8 (find_package):
  Found package configuration file:

    /home/kazarmy/bin/rizin-git/lib/x86_64-linux-gnu/cmake/Rizin/RizinConfig.cmake

  but it set Rizin_FOUND to FALSE so package "Rizin" is considered to be NOT
  FOUND.  Reason given by package:

  Failed to find Rizin component Core
```

This pr fixes the above problem by setting `_rizin_cmake_path` to its proper value (based on the current structure of having multiple folders under `cmake/`). It also replaces use of `PATHS` with `CMAKE_PREFIX_PATH` since `PATHS` is the lowest item in the CMake [`find_package`](https://cmake.org/cmake/help/v3.0/command/find_package.html) totem pole, which in practice means that it can be slow.

**Test plan**

Repeat the above steps perhaps. All builds are green.

**Closing issues**

...


XVilka:
A good thing to have in the 0.4.0, IMHO

wargio:
@ret2libc

ret2libc:
@kazarmy did you try just setting `-DCMAKE_PREFIX_PATH=/tmp/rizin-install-path`?

thestr4ng3r (changes requested):
> It also replaces use of PATHS with CMAKE_PREFIX_PATH since PATHS is the lowest item in the CMake [find_package](https://cmake.org/cmake/help/v3.0/command/find_package.html) totem pole, which in practice means that it can be slow.

Setting `CMAKE_PREFIX_PATH` is not the right approach here I think since that applies globally.
To avoid all the options and skip right to checking `PATHS`, options like `NO_DEFAULT_PATH` to `find_package` may be used. But I would have to check what the best practices for such recursive local config files are.

In any case, for now I would make this pr only do the `/..` change and nothing else to be clear about what the actual fix is.

kazarmy:
> In any case, for now I would make this pr only do the `/..` change and nothing else to be clear about what the actual fix is.

Right, okay 630be501740173b0931072c91ab6311d6b99e11a

kazarmy:
> @kazarmy did you try just setting `-DCMAKE_PREFIX_PATH=/tmp/rizin-install-path`?

That option aka

```bash
cmake -DCMAKE_PREFIX_PATH=~/bin/rizin-git/ -DCMAKE_INSTALL_PREFIX=~/.local ..
```

does work but after doing `cmake  -DCMAKE_INSTALL_PREFIX=~/.local ..` (from the rz-ghidra README) and getting the following error message:

```
CMake Error at CMakeLists.txt:8 (find_package):
  By not providing "FindRizin.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "Rizin", but
  CMake did not find one.

  Could not find a package configuration file provided by "Rizin" with any of
  the following names:

    RizinConfig.cmake
    rizin-config.cmake

  Add the installation prefix of "Rizin" to CMAKE_PREFIX_PATH or set
  "Rizin_DIR" to a directory containing one of the above files.  If "Rizin"
  provides a separate development package or SDK, be sure it has been
  installed.
```

the second option should work too, largely because it appears to be more straightforward.

ret2libc (approved):

thestr4ng3r (approved):

PR #2754, opened 2022-06-28
Add RzBinStrDb and api to interact with it via rz_bin_object_string_*

wargio:

**Detailed description**

Adds `RzBinStrDb` which provides simple methods to retrieve/add/remove strings.

### DIFF
--- a/librz/bin/bin.c
+++ b/librz/bin/bin.c
@@ -595,37 +595,37 @@ RZ_API void rz_bin_set_baddr(RzBin *bin, ut64 baddr) {
 }
 
 // XXX: those accessors are redundant
-RZ_API RzList *rz_bin_get_entries(RzBin *bin) {
+RZ_DEPRECATE RZ_API RzList *rz_bin_get_entries(RzBin *bin) {
 	rz_return_val_if_fail(bin, NULL);
 	RzBinObject *o = rz_bin_cur_object(bin);
 	return o ? (RzList *)rz_bin_object_get_entries(o) : NULL;
 }
 
-RZ_API RzList *rz_bin_get_fields(RzBin *bin) {
+RZ_DEPRECATE RZ_API RzList *rz_bin_get_fields(RzBin *bin) {
 	rz_return_val_if_fail(bin, NULL);
 	RzBinObject *o = rz_bin_cur_object(bin);
 	return o ? (RzList *)rz_bin_object_get_fields(o) : NULL;
 }
 
-RZ_API RzList *rz_bin_get_imports(RzBin *bin) {
+RZ_DEPRECATE RZ_API RzList *rz_bin_get_imports(RzBin *bin) {
 	rz_return_val_if_fail(bin, NULL);
 	RzBinObject *o = rz_bin_cur_object(bin);
 	return o ? (RzList *)rz_bin_object_get_imports(o) : NULL;
 }
 
-RZ_API RzBinInfo *rz_bin_get_info(RzBin *bin) {
+RZ_DEPRECATE RZ_API RzBinInfo *rz_bin_get_info(RzBin *bin) {
 	rz_return_val_if_fail(bin, NULL);
 	RzBinObject *o = rz_bin_cur_object(bin);
 	return o ? (RzBinInfo *)rz_bin_object_get_info(o) : NULL;
 }
 
-RZ_API RzList *rz_bin_get_libs(RzBin *bin) {
+RZ_DEPRECATE RZ_API RzList *rz_bin_get_libs(RzBin *bin) {
 	rz_return_val_if_fail(bin, NULL);
 	RzBinObject *o = rz_bin_cur_object(bin);
 	return o ? (RzList *)rz_bin_object_get_libs(o) : NULL;
 }
 
-RZ_API RzList *rz_bin_get_sections(RzBin *bin) {
+RZ_DEPRECATE RZ_API RzList *rz_bin_get_sections(RzBin *bin) {
 	rz_return_val_if_fail(bin, NULL);
 	RzBinObject *o = rz_bin_cur_object(bin);
 	return o ? (RzList *)rz_bin_object_get_sections_all(o) : NULL;
@@ -718,46 +718,31 @@ RZ_API RZ_OWN RzPVector *rz_bin_object_get_maps_at(RzBinObject *o, ut64 off, boo
 	return res;
 }
 
-RZ_API RzList *rz_bin_reset_strings(RzBin *bin) {
-	rz_return_val_if_fail(bin, NULL);
-	RzBinFile *bf = rz_bin_cur(bin);
-	if (!bf || !bf->o) {
-		return NULL;
-	}
-	return (RzList *)rz_bin_object_reset_strings(bin, bf, bf->o);
-}
-
-RZ_API RzList *rz_bin_get_strings(RzBin *bin) {
+RZ_DEPRECATE RZ_API RzList *rz_bin_get_strings(RzBin *bin) {
 	rz_return_val_if_fail(bin, NULL);
 	RzBinObject *o = rz_bin_cur_object(bin);
 	return o ? (RzList *)rz_bin_object_get_strings(o) : NULL;
 }
 
-RZ_API int rz_bin_is_string(RzBin *bin, ut64 va) {
-	rz_return_val_if_fail(bin, false);
-	RzBinObject *o = rz_bin_cur_object(bin);
-	return o ? rz_bin_object_is_string(o, va) : false;
-}
-
-RZ_API RzList *rz_bin_get_symbols(RzBin *bin) {
+RZ_DEPRECATE RZ_API RzList *rz_bin_get_symbols(RzBin *bin) {
 	rz_return_val_if_fail(bin, NULL);
 	RzBinObject *o = rz_bin_cur_object(bin);
 	return o ? (RzList *)rz_bin_object_get_symbols(o) : NULL;
 }
 
-RZ_API RzList *rz_bin_get_mem(RzBin *bin) {
+RZ_DEPRECATE RZ_API RzList *rz_bin_get_mem(RzBin *bin) {
 	rz_return_val_if_fail(bin, NULL);
 	RzBinObject *o = rz_bin_cur_object(bin);
 	return o ? (RzList *)rz_bin_object_get_mem(o) : NULL;
 }
 
-RZ_API int rz_bin_is_big_endian(RzBin *bin) {
+RZ_DEPRECATE RZ_API int rz_bin_is_big_endian(RzBin *bin) {
 	rz_return_val_if_fail(bin, false);
 	RzBinObject *o = rz_bin_cur_object(bin);
 	return o ? rz_bin_object_is_big_endian(o) : false;
 }
 
-RZ_API int rz_bin_is_static(RzBin *bin) {
+RZ_DEPRECATE RZ_API int rz_bin_is_static(RzBin *bin) {
 	rz_return_val_if_fail(bin, false);
 	RzBinObject *o = rz_bin_cur_object(bin);
 	return o ? rz_bin_object_is_static(o) : false;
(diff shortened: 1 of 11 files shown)


librz/bin/bin.c
```diff
@@ -718,46 +718,31 @@ RZ_API RZ_OWN RzPVector *rz_bin_object_get_maps_at(RzBinObject *o, ut64 off, boo
...
-	rz_return_val_if_fail(bin, NULL);
-	RzBinFile *bf = rz_bin_cur(bin);
-	if (!bf || !bf->o) {
-		return NULL;
-	}
-	return (RzList *)rz_bin_object_reset_strings(bin, bf, bf->o);
-}
-
-RZ_API RzList *rz_bin_get_strings(RzBin *bin) {
+RZ_DEPRECATE RZ_API RzList *rz_bin_get_strings(RzBin *bin) {
```
XVilka, line 721: Could you please add those `RZ_DEPRECATE` into the header file as well?

librz/bin/bobj.c
```diff
@@ -817,48 +783,60 @@ RZ_API const RzList *rz_bin_object_get_resources(RzBinObject *obj) {
...
+
+	// RzBinStrDb becomes the owner of the RzList strings
+	obj->strings = rz_bin_string_database_new(strings, obj->opts.loadaddr, bin->debase64);
+	return obj->strings != NULL;
+}
+
+/**
+ * \brief Return RzBinString if at \p address \p there is an entry in the RzBinObject string database
+ */
+RZ_API RzBinString *rz_bin_object_string_at(RZ_NONNULL RzBinObject *obj, ut64 address, bool is_va) {
```
XVilka: Add ownership for the return value

librz/core/canalysis.c
```diff
@@ -645,12 +645,14 @@ static ut64 *next_append(ut64 *next, int *nexti, ut64 v) {
 }
 
 static void rz_analysis_set_stringrefs(RzCore *core, RzAnalysisFunction *fcn) {
+	bool is_va = core->io->va;
+	RzBinObject *bobj = rz_bin_cur_object(core->bin);
```
XVilka, line 649: Maybe check for NULL here?

librz/include/rz_bin.h
```diff
@@ -936,8 +936,10 @@ RZ_API const RzList *rz_bin_object_get_strings(RZ_NONNULL RzBinObject *obj);
 RZ_API const RzList *rz_bin_object_get_mem(RZ_NONNULL RzBinObject *obj);
 RZ_API const RzList *rz_bin_object_get_resources(RZ_NONNULL RzBinObject *obj);
 RZ_API const RzList *rz_bin_object_get_symbols(RZ_NONNULL RzBinObject *obj);
-RZ_API const RzList *rz_bin_object_reset_strings(RZ_NONNULL RzBin *bin, RZ_NONNULL RzBinFile *bf, RZ_NONNULL RzBinObject *obj);
-RZ_API bool rz_bin_object_is_string(RZ_NONNULL RzBinObject *obj, ut64 va);
+RZ_API bool rz_bin_object_reset_strings(RZ_NONNULL RzBin *bin, RZ_NONNULL RzBinFile *bf, RZ_NONNULL RzBinObject *obj);
+RZ_API RzBinString *rz_bin_object_string_at(RZ_NONNULL RzBinObject *obj, ut64 address, bool is_va);
```
XVilka: Ownership mark for the return value

XVilka (changes requested):
New API idea looks OK.

librz/bin/bobj.c
```diff
@@ -1017,3 +995,119 @@ RZ_API ut64 rz_bin_object_v2p(RzBinObject *obj, ut64 vaddr) {
...
+				free(decoded);
+				decoded = tmp;
+			} while (1);
+			if (decoded != bstr->string) {
+				free(bstr->string);
+				bstr->string = decoded;
+				bstr->type = RZ_STRING_ENC_BASE64;
+			}
+		}
+
```
ret2libc: I do wonder whether this should be done here or somewhere else. I mean, there could be several types of encoding and shall they all be done here? Maybe it should be the responsibility of the caller to convert the strings in the right encoding.

librz/bin/bobj.c
```diff
@@ -1017,3 +995,119 @@ RZ_API ut64 rz_bin_object_v2p(RzBinObject *obj, ut64 vaddr) {
...
+			}
+			free(decoded);
+			decoded = tmp;
+		} while (1);
+		if (decoded != bstr->string) {
+			free(bstr->string);
+			bstr->string = decoded;
+			bstr->type = RZ_STRING_ENC_BASE64;
+		}
+	}
```
ret2libc: I think this is dupped code

librz/bin/bin.c
wargio: they are already there.

librz/bin/bobj.c
wargio: this was already part of the old code, i just integrated it in the loop since it was a small optimization

ret2libc: I know there was a separate function for this however as you are defining new apis I wonder whether it's really the right place for this.

librz/core/golang.c
```diff
@@ -515,12 +512,12 @@ static bool add_new_bin_string(RzCore *core, char *string, ut64 vaddr, ut32 size
...
 	bstr->paddr = paddr;
 	bstr->vaddr = vaddr;
-	bstr->ordinal = rz_list_length(bf->o->strings);
+	bstr->ordinal = rz_list_length(strings);
 	bstr->length = bstr->size = size;
 	bstr->string = string;
 	bstr->type = RZ_STRING_ENC_UTF8;
-	if (!rz_list_append(bf->o->strings, bstr)) {
-		RZ_LOG_ERROR("Failed append new go string to strings list\n");
+	if (!rz_bin_object_string_add(bin, bf->o, bstr)) {
```
thestr4ng3r: I think it is questionable whether this kind of adding strings to a binobject should be possible at all, since the info in it is expected to be statically derived from the bin, not through analysis in core. But I don't know a definitive solution yet either so I am ok with keeping the adding in this pr.
But since this is currently only done in this single place and to avoid future code like this, I suggest to remove `rz_bin_object_string_add()` and `rz_bin_object_string_remove()` and call `rz_bin_string_database_add()` etc. directly on the contained db.

librz/bin/bobj.c
```diff
@@ -1017,3 +995,119 @@ RZ_API ut64 rz_bin_object_v2p(RzBinObject *obj, ut64 vaddr) {
...
+
+	if (!rz_list_add_sorted(db->list, bstr, (RzListComparator)string_compare_sort)) {
```
thestr4ng3r: This is horribly slow due to linear search. Two possibilities for different data structures would be:
* Sorted Rz(P)Vector, super fast access, but slower insertion
* RBTree, a bit more indirection, but fast insertion

For reasons I am to lazy to write down right now, I am leaning towards the RBTree, but in this pr I would leave it as-is since this is no regression here. So just see this as a general comment.

thestr4ng3r (changes requested):

librz/bin/bobj.c
```diff
@@ -817,48 +783,60 @@ RZ_API const RzList *rz_bin_object_get_resources(RzBinObject *obj) {
...
+/**
+ * \brief Return RzBinString if at \p address \p there is an entry in the RzBinObject string database
+ */
+RZ_API RzBinString *rz_bin_object_string_at(RZ_NONNULL RzBinObject *obj, ut64 address, bool is_va) {
+	rz_return_val_if_fail(obj, false);
+	if (is_va) {
+		return ht_up_find(obj->strings->virt, address, NULL);
 	}
-	return obj->strings;
+	return ht_up_find(obj->strings->phys, address, NULL);
```
thestr4ng3r, line 869: `obj->strings` can be NULL from what I can tell. Same goes for all other places that access it and don't check.

thestr4ng3r (changes requested):

librz/include/rz_bin.h
thestr4ng3r: I would call this `rz_bin_object_get_string_at` to be consistent with the other functions around

librz/bin/bobj.c
XVilka: +1 for RBTree

librz/bin/bobj.c
```diff
@@ -787,7 +789,7 @@ RZ_API const RzList *rz_bin_object_get_classes(RzBinObject *obj) {
  */
 RZ_API const RzList *rz_bin_object_get_strings(RzBinObject *obj) {
 	rz_return_val_if_fail(obj, NULL);
-	return obj->strings;
+	return obj->strings->list;
```
thestr4ng3r, line 803:
```suggestion
	if (!obj->strings) {
		return NULL;
	}
	return obj->strings->list;
```

test/db/analysis/golang
```diff
@@ -17,7 +17,7 @@ EXPECT=<<EOF
 0x0063e49b str.version
 0x0063e4ac str.HTTP_version
 0x0063e566 str.
-12879
+12917
```
thestr4ng3r: Where do those diffs come from?

thestr4ng3r (changes requested):

test/db/analysis/golang
wargio: i have checked the output, essentially before i was deleting certain strings, which were duplicated, now i do not.

librz/bin/bobj.c
```diff
@@ -7,6 +7,37 @@
 #include <rz_util.h>
 #include "i/private.h"
 
+RZ_API void rz_bin_string_decode_base64(RzBinString *bstr) {
```
XVilka: Add Doxygen please, also check for NULL

test/db/analysis/golang
XVilka: Can you please attach the diff?

XVilka (approved):
Looks good!

librz/bin/bobj.c
```diff
@@ -1017,3 +1017,105 @@ RZ_API ut64 rz_bin_object_v2p(RzBinObject *obj, ut64 vaddr) {
...
 }
+
+/**
+ * \brief   Allocates and initializes the RzBinStrDb structure with the given list of strings
+ *
+ * \param   list  The list of strings to initialize the database with
+ *
+ * \return  On success returns a valid pointer, otherwise NULL
+ */
+RZ_API RZ_OWN RzBinStrDb *rz_bin_string_database_new(RZ_NULLABLE RzList /*<RzBinString *>*/ *list) {
```
ret2libc, line 1036: shall the list be RZ_BORROW?

wargio: borrow is only for returns, but it does borrow anything, it will own it.

thestr4ng3r: No, borrow can also mean an argument's ownership is kept at the caller.

ret2libc: sorry, i meant RZ_OWN :DDD

thestr4ng3r (approved):

test/db/analysis/golang
ret2libc: I think this is the last thing to see.

librz/bin/bobj.c
ret2libc: was RZ_OWN added?

ret2libc: these cases are very important to notify and they are actually one of the very reasons we introduced RZ_OWN and RZ_BORROW, because otherwise people have no way to know (without reading all the code) that the list they are providing as argument is not theirs anymore.

test/db/analysis/golang
ret2libc: @wargio

wargio: not much to see, but essentially before there were huge long strings and substrings being removed, now there are no more
so you may have in the string list more strings than before, this is because golang optimizes the string buffer but squashing everything together

wargio: image

librz/bin/bobj.c
wargio: yes it was

wargio: image

PR #2762, opened 2022-06-30
Add liblzma subproject and parse .gnu_debugdata section

### LINKED ISSUE #747: Backports from radare2
- [x] #1422
- [x] #1423
- [x] #1424
- [x] #1425
- [x] #1426
- [x] #1427
- [x] #1430
- [x] #1428
- [x] #1429
- [x] [Fix emulation for AARCH64 ldr,str,stp,ldp instructions ##esil](https://github.com/radareorg/radare2/commit/cf93f2aaf96bd3d23cb59e9b380c927741b9bd55)

## Enhancements
- [x] https://github.com/radareorg/radare2/pull/18545
- [x] https://github.com/radareorg/radare2/commit/fcf3ad1e1d95f493f0648b42f60a6901be53c5e5 : ELF improvement
- [ ] https://github.com/radareorg/radare2/commit/17a160cc7705e4af7bf92e6d6b11e9f706cb31e2 : ELF improvement
- [x] https://github.com/radareorg/radare2/pull/19042 (issue https://github.com/radareorg/radare2/issues/19043): MZ underflow

<details>

## Finished and rejected
- [ ] ~~https://github.com/radareorg/radare2/commit/e3b599177432456b43ec4123a29128657fe22708~~ : see https://github.com/rizinorg/rizin/pull/1419 right now i could not find any issue with those functions, so i'm not backporting
- [x] https://github.com/radareorg/radare2/commit/0ace091ff4b95c194ef975f156e31150656e7ee9 : see https://github.com/rizinorg/rizin/pull/1278/
- [ ] ~~https://github.com/radareorg/radare2/commit/acf469352ccff0fe5c01b7f9758521bf55a4ceac~~ You can see the changes to tests has been essentially reverted in https://github.com/radareorg/radare2/commit/3fedf8003687cccd25bc739427b95b2f40742c03#diff-4966e4f1fa478276b4c6424b2ae54c6216ed6311b8c3b7d43976b71070307236 . The output of r2 and rizin are the same now, so we can skip this.
- [ ] ~~https://github.com/radareorg/radare2/commit/0977e6119fe8371e7bc24bd8043b774a36709f68~~ let's fix this when we encounter the problem. With proper tests as well.
- [ ] ~~https://github.com/radareorg/radare2/pull/18483/files~~ some things have already been patched. others i cannot find. i think we can just skip this.
- [ ] ~~https://github.com/radareorg/radare2/commit/a0213f0347b5a3a302566c060d1a486dd29be967~~ we can fix our coverity issues ourselves. I think it's safer, easier than trying to backport things happening on a now very different codebase.
- [ ] ~~https://github.com/radareorg/radare2/pull/18521~~ we can fix our coverity issues ourselves. it's just easier.
- [ ] ~~https://github.com/radareorg/radare2/commit/c150dc489fe68c3e02750ef72cc365c5cc5f549a~~ this seems wrong because `disassembler_options` is a `char *` and options is a local var on the stack. Anyway, we are fixing this in our own PR.
- [ ] ~~https://github.com/radareorg/radare2/commit/57a9926b1fbc83c756adda62243dbe251253cfda~~ some small refactoring in cons grep.c. I think we can do that ourselves at some point.
- [ ] https://github.com/radareorg/radare2/commit/be3b640725cbaf6e165447f705aa471a3b923330 : let's ignore pf for now
- [ ] https://github.com/radareorg/radare2/commit/6181e1e17f4dbe05d88fbf9b582b8657387949e1 : let's ignore pf for now
- [ ] https://github.com/radareorg/radare2/commit/42d05b7bd13f66580f62f9d29e04bf200c3570a0 : let's skip, we rewrite that part if necessary
# Plan 9 binary format
- [ ] https://github.com/radareorg/radare2/commit/592e6407d6cf26ea3dcf977d17965277783a8d78
- [ ] https://github.com/radareorg/radare2/commit/fbfebef9e659969d648774458cd5ae0ab16fdc6d
at this point rzbin plugins are quite different as we have maps to map things, so not sure it's worth backporting. it would be a rewrite of the plugin in other words.
- [ ] https://github.com/radareorg/radare2/pull/18490 : I know nothing about this ssdeep algo, but I used https://pypi.org/project/ssdeep/ and I got different results than what expected in the unit tests there. Also, ssdeep library is provided on many distros, so we should probably add the chance to compile with libssdeep instead of our own implementation, if we really want this. Also, probably it requires some adjustments due to the many changes done by @wargio on the hash code. ----- it doesn't seem to be much requested feature. skip given the above.
- [ ] https://github.com/radareorg/radare2/pull/18424 (use a separate plugin): same, it doesn't look like it's much requested feature so far. skip.

## Enhancements signatures
- [ ] https://github.com/radareorg/radare2/commit/0a10c3f13a5fb5548d7035ef5e54d2f840424069
- [ ] https://github.com/radareorg/radare2/commit/98150ad1829df99e52a96f2ed91fe4399c1ead5c
- [ ] https://github.com/radareorg/radare2/commit/086038fecaada0a6db3c8e1c49927614204d675a
- [ ] https://github.com/radareorg/radare2/commit/74b8b3cc99f293f101a616099a728247256a8e65
- [ ] https://github.com/radareorg/radare2/commit/00e8fd12533a6a6753dcc62e95e0eae8a74dc0e4
all skip because signatures were heavily refactored already by wargio.

</details>

ret2libc:

**Detailed description**

This PR adds support  for https://sourceware.org/gdb/onlinedocs/gdb/MiniDebugInfo.html by adding the liblzma dependency useful to decompress the .gnu_debugdata section. Then it parses it by using all the existing methods.

Backports https://github.com/radareorg/radare2/commit/17a160cc7705e4af7bf92e6d6b11e9f706cb31e2

**Test plan**

...

**Closing issues**

Closes https://github.com/rizinorg/rizin/issues/747


XVilka:
@ret2libc where it is planned to be used?
I wonder if we just should switch to the [libarchive](https://www.libarchive.org/) ([GitHub](https://github.com/libarchive/libarchive)) once and for all, and be done with various different compression formats one by one.

ret2libc:
> @ret2libc where it is planned to be used? I wonder if we just should switch to the [libarchive](https://www.libarchive.org/) ([GitHub](https://github.com/libarchive/libarchive)) once and for all, and be done with various different compression formats one by one.

https://sourceware.org/gdb/onlinedocs/gdb/MiniDebugInfo.html which is the last missing point from https://github.com/rizinorg/rizin/issues/747 . r2 implements it by directly calling `xz` executable, I want to have LZMA algo inside. I will make the whole thing optional, so if one does not want the dependency he can skip it altogether (at the cost of not having this feature ofc).

XVilka (changes requested):
Could you please rename then `librz/util/zip.c`? To be more generic.

librz/bin/format/elf/elf_symbols.c
```diff
@@ -364,6 +364,77 @@ static bool get_section_elf_symbols(ELFOBJ *bin, RzVector *result, RzBinElfSymbo
 	return true;
 }
 
+static bool get_gnu_debugdata_elf_symbols(ELFOBJ *bin, RzVector *result, RzBinElfSymbolFilter filter, HtUU *set) {
```
XVilka, line 367: While it's not exported function, I think it would be still nice to have a brief description what it does, maybe the link to the information about `gnu_debugdata`

XVilka (changes requested):

XVilka (approved):

librz/bin/format/elf/elf_symbols.c
```diff
@@ -364,6 +364,78 @@ static bool get_section_elf_symbols(ELFOBJ *bin, RzVector *result, RzBinElfSymbo
...
+	}
+
+	HtPP *name_set = ht_pp_new0();
+	if (!name_set) {
+		goto debug_symbols_err;
+	}
+
+	RzBinElfSymbol *sym;
+	rz_vector_foreach(result, sym) {
+		ht_pp_insert(name_set, sym->name, sym);
```
wargio, line 411: this will fail if the name is already in the map.

ret2libc: it's alright, isn't it? i just don't want to provide dupes of symbols when parsing .gnu_debugdata

XVilka (approved):

PR #2764, opened 2022-07-01
In Vv mode allow filter flags in the left column

### LINKED ISSUE #2750: In `Vv` mode allow filter flags in the left column
When you open `Vv` mode on the left you could see the list of the flags. It is somewhat similar to the default Cutter interface.
Unlike Cutter, Rizin doesn't have key to enter filter of these flags by name/pattern. It would be highly useful to have that feature and key.

PeiweiHu:

**Detailed description**

Add filter in Vv mode. `f` adds the new keyword. `F` clears all keywords. This pr also adds hints of the current keywords. 

...

**Test plan**

CI is green. 

**Closing issues**

closes https://github.com/rizinorg/rizin/issues/2750
...


librz/include/rz_core.h
```diff
@@ -349,6 +349,7 @@ struct rz_core_t {
 	bool keep_asmqjmps;
 	RzCoreVisual visual;
 	// visual // TODO: move them into RzCoreVisual
+	RzList *visual_filter; // store the filter keywords in visual mode
```
XVilka: I don't see any deletion of the elements, I wonder if RzPVector would be a better fit here.

librz/core/tui/vmenus.c
```diff
@@ -2708,6 +2729,29 @@ RZ_API void rz_core_visual_analysis(RzCore *core, const char *input) {
...
 		switch (ch) {
+		case 'f':
+			// add new keyword
+			if (!core->visual_filter) {
+				core->visual_filter = rz_list_newf(free);
+				if (!core->visual_filter) {
+					break;
+				}
+			}
+			char *keyword = malloc(256);
```
XVilka: Please extract this number at least to `#define`, here and below.

XVilka (changes requested):

librz/include/rz_core.h
PeiweiHu: Ok. I will change it later.

XVilka (changes requested):
I tested this and it works, thank you!
One minor thing - I think it's better to show the filter keywords outside of the "legend" area, maybe below the function list? Though that's my opinion, let's see what others say.

Another question would be, if it makes sense to start filtering *during the typing*, thus the results will be clear immediately. For example, you often can mistype or get the lowercase/uppercase wrong. Filtering on the go has some *explorative* spirit.

PeiweiHu:
> Another question would be, if it makes sense to start filtering _during the typing_, thus the results will be clear immediately. For example, you often can mistype or get the lowercase/uppercase wrong. Filtering on the go has some _explorative_ spirit.

Good idea. I'd like to give it a shot.

PeiweiHu:
> Another question would be, if it makes sense to start filtering _during the typing_, thus the results will be clear immediately. For example, you often can mistype or get the lowercase/uppercase wrong. Filtering on the go has some _explorative_ spirit.

Now it supports filtering on the go, as shown in the gif.

wargio (approved):

XVilka (approved):

PR #2768, opened 2022-07-02
Add RzBaseFindOpt for custom thread status callback

wargio:

**Detailed description**

Add RzBaseFindOpt to allow to have custom thread status callbacks

### DIFF
--- a/librz/include/rz_basefind.h
+++ b/librz/include/rz_basefind.h
@@ -21,7 +21,29 @@ typedef struct rz_basefind_t {
 	ut32 score;
 } RzBaseFindScore;
 
-RZ_API RZ_OWN RzList *rz_basefind(RZ_NONNULL RzCore *core, ut32 pointer_size);
+typedef struct rz_basefind_info_t {
+	ut32 thread_idx;
+	ut32 n_threads;
+	ut64 begin_address;
+	ut64 current_address;
+	ut64 end_address;
+	ut32 percentage;
+} RzBaseFindThreadInfo;
+
+typedef bool (*RzBaseFindThreadInfoCb)(const RzBaseFindThreadInfo *th_info, void *user);
+
+typedef struct rz_basefind_options_t {
+	ut32 pointer_size;
+	ut32 min_score;
+	ut64 start_address;
+	ut64 end_address;
+	ut64 increase_by;
+	size_t max_threads;
+	RzBaseFindThreadInfoCb callback;
+	void *user;
+} RzBaseFindOpt;
+
+RZ_API RZ_OWN RzList *rz_basefind(RZ_NONNULL RzCore *core, RzBaseFindOpt *options);
 
 #ifdef __cplusplus
 }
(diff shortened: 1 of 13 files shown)


XVilka (approved):
Looks good, but it would also benefit from unit/integration test, I think.

librz/include/rz_basefind.h
```diff
@@ -17,11 +17,33 @@ extern "C" {
...
+typedef struct rz_basefind_info_t {
+	ut32 thread_idx; ///< Thread number
+	ut32 n_threads; ///< Total number of threads
+	ut64 begin_address; ///< Thread begin address
+	ut64 current_address; ///< Thread current address
+	ut64 end_address; ///< Thread end address
+	ut32 percentage; ///< Current percentage of the thread scan
+} RzBaseFindThreadInfo;
+
+typedef bool (*RzBaseFindThreadInfoCb)(const RzBaseFindThreadInfo *th_info, void *user);
```
ret2libc: What is this used for exactly? Also, the doc for the fields might be preciser i guess?
I agree unit tests would help.

wargio: i do not think is a good idea to implement unit tests for this, since this will require a big amount of time to run (compared to the other tests).

librz/include/rz_basefind.h
```diff
@@ -17,11 +17,33 @@ extern "C" {
...
+	ut32 score; ///< Score of the candidate address
 } RzBaseFindScore;
 
-RZ_API RZ_OWN RzList *rz_basefind(RZ_NONNULL RzCore *core, ut32 pointer_size);
+typedef struct rz_basefind_info_t {
+	ut32 thread_idx; ///< Thread number
+	ut32 n_threads; ///< Total number of threads
+	ut64 begin_address; ///< Thread begin address
+	ut64 current_address; ///< Thread current address
+	ut64 end_address; ///< Thread end address
```
ret2libc: What is the "thread begin address'?

wargio: each thread has a range to scan, that is the thread range

ret2libc: ok maybe be a bit more explicit about this in the doc for these fields.

Moreover, from a more high level point of view, why do we care about the details? I mean, do we really need to export the information about threads running the search? Are you planning to export this info in cutter?

wargio: yes.

test/unit/test_basefind.c
```diff
@@ -0,0 +1,134 @@
...
+	rz_io_open(core->io, "bins/firmware/stm32f103-dapboot-v1.20-bluepill.bin", RZ_PERM_R, 0);
+	basefind_options_set_valid(&options);
+	result = rz_basefind(core, &options);
+	mu_assert_null(result, "file not loaded via core");
+
+	rz_core_free(core);
+	mu_end;
+}
+
+int all_tests() {
```
ret2libc: I think these tests should go in `test/integration` because they require bins to be loaded. They also use RzCore so a lot of stuff are just supposed to be in the right place.

ret2libc (commented):
LGTM

ret2libc (approved):

PR #2777, opened 2022-07-04
Use gnu99 globally and update libzip to v1.9.2

ret2libc:
Let's see how it goes with a full upgrade.


ret2libc:
only first two commits should be backported, not the updated libzip one.

thestr4ng3r (approved):

XVilka (approved):

PR #2787, opened 2022-07-06
Remove rz_th_kill and replace with safe halt. 

wargio:

**Detailed description**

...

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/librz/core/basefind.c
+++ b/librz/core/basefind.c
@@ -331,6 +331,7 @@ RZ_API RZ_OWN RzList *rz_basefind(RZ_NONNULL RzCore *core, RZ_NONNULL RzBaseFind
 	ut64 base_start = options->start_address;
 	ut64 base_end = options->end_address;
 	ut64 alignment = options->alignment;
+	options->callback = NULL;
 
 	if (options->pointer_size != 32 && options->pointer_size != 64) {
 		RZ_LOG_ERROR("basefind: supported pointer sizes are 32 and 64 bits.\n");
--- a/librz/core/basefind.c
+++ b/librz/core/basefind.c
@@ -23,6 +23,11 @@ typedef struct basefind_data_t {
 	BaseFindArray *array;
 } BaseFindData;
 
+typedef struct thread_bool_t {
+	bool value;
+	RzThreadLock *lock;
+} ThreadBool;
+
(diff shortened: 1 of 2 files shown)


XVilka (approved):

ret2libc:
Description in the PR please. What was the problem about? How does the PR solve it?

librz/core/basefind.c
```diff
@@ -35,14 +40,50 @@ typedef struct basefind_thread_data_t {
...
+static void thread_bool_set(ThreadBool *tl, bool value) {
+	rz_th_lock_enter(tl->lock);
+	tl->value = value;
+	rz_th_lock_leave(tl->lock);
+}
+
+static void thread_bool_fini(ThreadBool *tl) {
+	rz_th_lock_free(tl->lock);
+}
+
```
ret2libc, line 74: I'm not sure about this TBH... Is all this thing really necessary for a bool? I think there are some functions like cmp&xchg or something like that that can be used to atomically set a bool value.

librz/core/basefind.c
```diff
@@ -276,12 +318,12 @@ static void *basefind_thread_ui(BaseFindUIInfo *ui_info) {
...
 			basefind_set_thread_info(bftd, &th_info, i);
 			if (!callback(&th_info, user)) {
-				rz_th_pool_kill(pool);
+				basefind_stop_all_search_threads(pool);
 				goto end;
 			}
 		}
 		rz_sys_usleep(100000);
-	} while (1);
+	} while (thread_bool_get(&ui_info->loop));
```
ret2libc, line 326: is there anything preventing this condition to be true even though, one microsecond after this has been checked, the value of ->loop is set to false? Is it alright to have such case?

librz/core/basefind.c
wargio: This is what is what happens when you do multithreaded code, becomes very weird.

librz/core/basefind.c
wargio: What may happen is that while exiting from the `thread_bool_get` which holds a `true`, the value gets set to `false` but this means that the execution is still happening, which is fine, since it will be terminated during the next loop.
The thread is anyway waited by the main thread, so nothing to worry.

PR #2799, opened 2022-07-10
Update `RZ_BIRTH` on every build

kazarmy:

**Detailed description**

This pr properly updates `RZ_BIRTH` (the timestamp shown by `rizin -v`) on every build i.e. every time `ninja` is invoked, instead of only every time a `meson.build` file is changed.

**Test plan**

Check manually whether the timestamp shown by `rizin -v` reflects the time of the last build after making a change to a `.c` or `.h` file. All CI builds should be green.

**Closing issues**

...

### DIFF
--- a/librz/include/rz_version.h.in
+++ b/librz/include/rz_version.h.in
@@ -7,7 +7,6 @@
 #define RZ_VERSION_NUMBER @RZ_VERSION_NUMBER@
 #define RZ_VERSION "@RZ_VERSION@"
 #define RZ_GITTIP "@RZ_GITTIP@"
-#define RZ_BIRTH "@RZ_BIRTH@"
 #mesondefine RZ_PACKAGER_VERSION
 #mesondefine RZ_PACKAGER
 #endif
--- a/librz/util/meson.build
+++ b/librz/util/meson.build
@@ -55,6 +55,7 @@ rz_util_sources = [
   'str_constpool.c',
   'str_search.c',
   'str_trim.c',
+  'str_version.c',
   'strbuf.c',
   'strpool.c',
   'subprocess.c',
@@ -94,7 +95,7 @@ else
   rz_util_sources += ['big.c']
 endif
 
-rz_util = library('rz_util', rz_util_sources,
+rz_util = library('rz_util', rz_util_sources + [rz_build_h],
   include_directories: platform_inc,
   c_args: library_cflags,
   dependencies: rz_util_deps,
--- a/librz/util/str.c
+++ b/librz/util/str.c
@@ -3924,47 +3924,6 @@ RZ_API RzList *rz_str_wrap(char *str, size_t width) {
 	return res;
 }
 
-// version.c
-#include <rz_userconf.h>
-#include <rz_util.h>
-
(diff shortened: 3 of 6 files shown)


kazarmy:
Updated the [head comment](https://github.com/rizinorg/rizin/pull/2799#issue-1299874188).

ret2libc:
How many files are rebuilt when you don't change anything at all and you re-run ninja?

kazarmy:
Three basically:

1. "rz_build.h"
2. "str_version.c" because it depends on "rz_build.h"
3. "librz_util.so.0.5.0" because it depends on "str_version.o"

thestr4ng3r (changes requested):
That is quite a lot actually. Especially on slower devices, relinking rz_util can take a significant amount of time, and this also depends on python. So I don't think this is a good idea. Do you have any concrete use-case where you really need the exact build time rather than the configure time?

XVilka:
Another solution would be to move the `rz_str_version` from RzUtil to somewhere else, it doesn't really belong here anyway.

kazarmy:
> Especially on slower devices, relinking rz_util can take a significant amount of time, and this also depends on python.

I can have `str_version()` read the build timestamp from a file, instead of the timestamp being compiled in as part of librz_util. Would that be okay?

> Do you have any concrete use-case where you really need the exact build time rather than the configure time?

Nothing significant really. It's just that the output of `rizin -v` that we ask as part of the GitHub bug report template may not be entirely accurate, particularly the git hash. But we are going to check with HEAD anyway so this issue is somewhat moot.

This pr is about the build timestamp rather than the git hash because that appears easier to fix, and fixing the git hash would require this kind of infra anyway.

kazarmy:
> Another solution would be to move the `rz_str_version` from RzUtil to somewhere else, it doesn't really belong here anyway.

It could be in RzMain but I don't want to be delayed unnecessarily by a linking step either.

ret2libc:
I'm not sure any of the other tools around do something like this. It is kinda expected that if you run two times "make" or ninja or whatever without changing the source code nothing is done. With this PR there is always something to compile. 

TBH I'm not sure the build time is even that useful. The git tag is probably more useful if it's accurate and it's also more stable than the time.

kazarmy:
> TBH I'm not sure the build time is even that useful.

I agree, let's remove it ... never seen a tool include it as part of their `--version` output either. It's only that R2/Rizin has always been somewhat ... different.

thestr4ng3r:
I do remember seeing the build time in many tools actually, for example the macos kernel:
```
florian-macbook:rizin florian$ uname -v
Darwin Kernel Version 21.5.0: Tue Apr 26 21:08:29 PDT 2022; root:xnu-8020.121.3~4/RELEASE_ARM64_T8101
```

Linux:
```
florian@florian-desktop ~ $ uname -v
#1 SMP Sun Jul 3 12:57:10 CEST 2022
```

But it really is usually just the configure time, so just as we do it.

kazarmy:
Oh ok

kazarmy:
Still I don't see why the configure time is useful for a tool like Rizin, even for debugging purposes.

PR #2830, opened 2022-07-21
Make "rb" command rebase all analysis

BenH11235:

**Detailed description**

Currently, once analysis is performed (e.g via the `aaa` command), some of it (such as function names) remains bound to its current address and can't be manually rebased. A rebase will be performed once when running `oodf`, but if the user can't or doesn't want to run that specific command, or needs to perform another rebase after running it, there appears to be no way for them to do so.

The changes in this PR implement the TODO in `rb_resize_handler`: 

https://github.com/rizinorg/rizin/blob/6c665c609879c790d2d2d835ec7a38e5776d8a5c/librz/core/cmd/cmd_resize.c#L12 

Since the required functionality seems to be already implemented for the `oodf` command, the implementation exposes the internal `__rebase_everything` function in `librz/core/cfile.c` (used in `oodf`) as part of the `rz_core` API, then calls that function from `cmd_resize.c`. The rewrite allows the user to either use the current recorded binary's loaded address, or supply one of their own.

**Test plan**

```
o my_binary (note original loading address -- value of "baddr" in output of "i" command)
aaa
oob 0x800000
(V, return, v, note that symbols are out of sync)
rb original_loading_address 0x800000
(V, return, v, note that symbols in sync again)

### CODE REVIEW

**Comment by ret2libc** [comment]
Thanks *a lot* for looking at this issue!
It is a big one and a very important one to fix indeed.
We will look at your code and provide feedback ;)

Thanks again!

File: librz/core/cfile.c [hunk 1]
```diff
@@ -53,19 +53,25 @@ static void loadGP(RzCore *core) {
...
 }
 
-static RZ_OWN RzList *__save_old_sections(RzCore *core) {
+
+/**
+ * \brief Create a backup of binary's sections.
+ * \param core Rizin core.
+ * \return RzList<RzBinSection>
+ */
+RZ_API RZ_OWN RzList *rz_core_create_sections_backup(RZ_BORROW RzCore *core) {
```
DMaroo: Please add

```c
rz_return_val_if_fail(core, NULL)
```

That will return `NULL` and log an error on the console if `core` is `NULL`. You should put it elsewhere as well, wherever you expect the argument to be non-null.

librz/core/cfile.c
```diff
@@ -120,25 +126,35 @@ static bool __rebase_xrefs(void *user, const ut64 k, const void *v) {
...
-static void __rebase_everything(RzCore *core, RzList *old_sections, ut64 old_base) {
+
+/**
+ * \brief rebase all flags, binary, information, breakpoints, and analysis.
+ * \param core Rizin core.
+ * \param sections_backup Backup of original file sections.
+ * \param infer_new_baddr_shift Should the object's current baddr_shift be used to compute a new base address.
+ * \return void
+ */
+RZ_API void rz_core_rebase_everything(RZ_BORROW RzCore *core, RZ_BORROW RzList *sections_backup, bool infer_new_baddr_shift, ut64 old_baddr_shift, ut64 new_baddr_shift) {
```
DMaroo, line 159: It's also a good idea to use the `RZ_NONNULL` annotation for the arguments which you don't expect to be `NULL`. Like `sections_backup` in this case.

Also add the `rz_return_val_if_fail` check here as well (you should check both, `core` and `sections_backup`).

Side note: Usually, we don't use the `RZ_NONNULL` annotation with `RzCore` because of brevity and to reduce unnecessary clutter. It is implicit that any `RzCore` pointer being passed should not be `NULL`

DMaroo (commented):
Thank you for your contribution! :)

librz/core/cfile.c
XVilka: I also suggest removing `RZ_BORROW` from the `RzCore` as well.

librz/core/cfile.c
```diff
@@ -298,8 +312,8 @@ RZ_API void rz_core_file_reopen_debug(RzCore *core, const char *args) {
 		return;
 	}
 	core->dbg->main_arena_resolved = false;
-	RzList *old_sections = __save_old_sections(core);
-	ut64 old_base = core->bin->cur->o->baddr_shift;
+	RzList *sections_backup = rz_core_create_sections_backup(core);
```
XVilka, line 483: If you will add a return of NULL in case of error, then you would need to check `sections_backup` for NULL as well.

librz/core/cfile.c
```diff
@@ -1614,8 +1628,8 @@ RZ_API void rz_core_io_file_open(RZ_NONNULL RzCore *core, int fd) {
 	char *file = rz_str_path_escape(bf->file);
 	// Backup the baddr and sections that were already rebased to
 	// revert the rebase after the debug session is closed
-	ut64 orig_baddr = core->bin->cur->o->baddr_shift;
-	RzList *orig_sections = __save_old_sections(core);
+	ut64 orig_baddr_shift = core->bin->cur->o->baddr_shift;
+	RzList *orig_sections = rz_core_create_sections_backup(core);
```
XVilka, line 1836: And here

XVilka (changes requested):
Please regenerate the C code from YAML again:
```
Run ./librz/core/cmd_descs/cmd_descs_generate.py --output-dir /tmp ./librz/core/cmd_descs/*.yaml
731c731
< RZ_IPI RzCmdStatus rz_open_binary_rebase_handler(RzCore *core, int argc, const char **argv);
---
> RZ_IPI RzCmdStatus rz_open_binary_rebase_handler(RZ_BORROW RzCore *core, int argc, RZ_BORROW const char **argv);
Error: Process completed with exit code 1.
```

librz/core/cmd_descs/cmd_resize.yaml
```diff
@@ -32,6 +32,9 @@ commands:
     args:
       - name: oldbase
         type: RZ_CMD_ARG_TYPE_RZNUM
+      - name: newbase
+        type: RZ_CMD_ARG_TYPE_RZNUM
+        optional: true
```
ret2libc, line 37: Probably this command should be renamed and put somewhere else.. It is not related at all to "resize" functionality and I would expect it to be somewhere under `o` (for example `or`?)

Also, I believe the user should not have to pass the `oldbase` argument because Rizin should already be aware of that itself, without the user having to tell it.

librz/core/cmd_descs/cmd_descs.h
```diff
@@ -728,7 +728,7 @@ RZ_IPI RzCmdStatus rz_open_binary_list_ascii_handler(RzCore *core, int argc, con
 RZ_IPI RzCmdStatus rz_open_binary_show_handler(RzCore *core, int argc, const char **argv);
 RZ_IPI RzCmdStatus rz_open_binary_add_handler(RzCore *core, int argc, const char **argv);
 RZ_IPI RzCmdStatus rz_open_binary_file_handler(RzCore *core, int argc, const char **argv);
-RZ_IPI RzCmdStatus rz_open_binary_rebase_handler(RzCore *core, int argc, const char **argv);
+RZ_IPI RzCmdStatus rz_open_binary_rebase_handler(RZ_BORROW RzCore *core, int argc, RZ_BORROW const char **argv);
```
ret2libc, line 731:
```suggestion
RZ_IPI RzCmdStatus rz_open_binary_rebase_handler(RzCore *core, int argc, RZ_BORROW const char **argv);
```

ret2libc: this file is automatically generated (see the header of the file), don't touch it manually.

librz/core/cmd/cmd_resize.c
```diff
@@ -4,21 +4,55 @@
...
+	ut64 old_base = rz_num_math(core->num, argv[1]);
+	ut64 static_old_delta = old_base - static_base;
+
+	// perform actual rebase
+	RzList *sections_backup = rz_core_create_sections_backup(core);
+	if (!sections_backup) {
+		RZ_LOG_ERROR("Cannot create sections backup.\n");
+		return RZ_CMD_STATUS_ERROR;
+	}
+
```
ret2libc: Can you move this into the `rz_core_rebase_everything` function? If you do so you can make `rz_core_create_sections_backup` an internal function in cfile.c instead of an export RZ_API.

librz/core/cmd/cmd_resize.c
```diff
@@ -4,21 +4,55 @@
...
-}
+/**
+ * \brief rb - rebase all flags, binary, information, breakpoints, and analysis.
+ * \param core Rizin core.
+ * \param argc Count of command arguments.
+ * \param argv Vector containing command arguments.
+ * \return RZ_CMD_STATUS_OK or RZ_CMD_STATUS_ERROR.
+ */
+RZ_IPI RzCmdStatus rz_rebase_handler(RZ_BORROW RzCore *core, int argc, RZ_BORROW const char **argv) {
+
```
ret2libc:
```suggestion
```

librz/core/cmd/cmd_resize.c
```diff
@@ -4,21 +4,55 @@
...
+	// get current file and current object
+	RzBinFile *bf = rz_bin_cur(core->bin);
+	if (!(bf && rz_file_exists(bf->file))) {
```
ret2libc:
```suggestion
	if (!bf) {
```

I think you don't need to check if the file exists. Was there any particular reason why you did that? AFAIK bf->file could also refer to a non physical file.

ret2libc (commented):
I left few comments, however I will need a bit more time to carefully look at this rebase topic overall. I'm not 100% sure that what done by `oodf` is the best, I actually think the approach taken by the old `rb` code is cleaner (apart from the fact analysis functions are not moved, but as written in the TODO maybe it is just a matter of creating an API in RzAnalysis).

Anyway, thanks a lot for looking at this topic, it's not an easy one as said. I will try to provide a more definitive reply ASAP on what, I think, would be the best approach IMHO.

BenH11235:
Thank you for all the detailed feeback. I will be able to take a proper look at it on Sunday.

BenH11235:
Regarding `cmd_descs.c` and `cmd_descs.h`, I was somewhat confused that autogenerated files appeared in the repository (I'd expect them to be part of the build process). Anyway I am sure there is good reason for this and rolled back the manual modifications to them.

Regarding `rz_return_val_if_fail` I was apprehensive to call this macro that returns a `NULL` value inside a function with return type `void`. Instead in such functions I checked for the error condition manually, and on fail logged the error and used a naked `return`.

Regarding folding `create_sections_backup` into `rz_core_rebase_everything` I indeed remember that this seemed obvious to do and even remember trying it, and running into an issue. This was a few weeks ago so apologies if my memory is not perfectly clear, but IIRC in some places there is a call to `create_sections_backup` followed by several manipulations and only then the call to `rebase_everything`. If the section backup is made later while already in `rebase_everything` the sections have already changed, and then the rebase fails. Maybe this is possible to do but it is not as straightforward as simply using the functions the exact same way as before.

Regarding the possibility that it's the approach in `oodf` that needs a rewrite based on the one in `rb` instead, I will gladly help with this discussion to the best of my ability but please understand that I only became familiar with the Rizin code base a few weeks ago for attempting to contribute this specific feature, and this is in fact my first ever proposed contribution to a large open source project. It was only possible at all for me because "hey invoking from elsewhere functionality that already exists, how difficult could it be". Of course all decisions should be made according to what is right for the project, and if "just do what oodf does" is wrong to do then we should not do it, just be advised that in that case my ability to contribute further to this specific PR will probably be fairly limited.

Thanks again

DMaroo:
@BenH11235

Regarding generating `cmd_descs.{c,h}`, these are auto-generated during the build process if you have `pyyaml` installed on your system (the YAML command tree is parsed and thus the files are autogenerated). So, I believe, installing `pyyaml` should solve the problem.

Regarding your concern about returning `NULL` inside `void` return type functions, you can just use another macro `rz_return_if_fail`. It basically logs a warning and returns from the function.

ret2libc:
> @BenH11235
> 
> Regarding generating `cmd_descs.{c,h}`, these are auto-generated during the build process if you have `pyyaml` installed on your system (the YAML command tree is parsed and thus the files are autogenerated). So, I believe, installing `pyyaml` should solve the problem.

Please also make sure to delete and re-create your build directory. The reason why the files are *also* in the repository is that to generate them you need pyyaml, so we want to enable users to compile Rizin without this additional dependency. Developers who want to change commands will need to install pyyaml.

ret2libc:
> Regarding folding `create_sections_backup` into `rz_core_rebase_everything` I indeed remember that this seemed obvious to do and even remember trying it, and running into an issue. This was a few weeks ago so apologies if my memory is not perfectly clear, but IIRC in some places there is a call to `create_sections_backup` followed by several manipulations and only then the call to `rebase_everything`. If the section backup is made later while already in `rebase_everything` the sections have already changed, and then the rebase fails. Maybe this is possible to do but it is not as straightforward as simply using the functions the exact same way as before.



> Regarding the possibility that it's the approach in `oodf` that needs a rewrite based on the one in `rb` instead, I will gladly help with this discussion to the best of my ability but please understand that I only became familiar with the Rizin code base a few weeks ago for attempting to contribute this specific feature, and this is in fact my first ever proposed contribution to a large open source project. It was only possible at all for me because "hey invoking from elsewhere functionality that already exists, how difficult could it be".


> Of course all decisions should be made according to what is right for the project, and if "just do what oodf does" is wrong to do then we should not do it, just be advised that in that case my ability to contribute further to this specific PR will probably be fairly limited.

Thank you, I really appreciate your being honest and realistic about it. I understand this is your first contributions so I don't want to transform this first contribution into a huge change.

We are discussing just right now on the Mattermost channels how we would envision a "perfect" rebase feature: what can be rebased, what it means to rebase, what's needed, etc.

As a first and intermediate step which I think could be a good-sized chunk for a PR i suggest to mix the behaviour of `rebase_helper` and of `__rebase_everything` into a single function that is exported(RZ_API). For example, `rebase_helper` sets the new baddr and and use `	rz_core_bin_apply_all_info(core, rz_bin_cur(core->bin));`, however `__rebase_everything` moves functions, xrefs and meta. Probably the two functions should be merged because everything needs to be rebased.

Another good thing you could do to help us is by writing simple but effective tests involving rebasing. That is the best way to communicate what you expect from a rebase operation.

BTW, I think you merged some commits in this PR that should not be here (e.g. there are changes that were already applied to dev but that are shown as separate in this PR).


Again, thanks for looking at this. If you have any doubt, please let me/us know and we'll try to help!

XVilka:
Apart from the already mentioned points, it will also need to be rebased to resolve conflicts.

PR #2813, opened 2022-07-17
Convert `px[ dDhHwWqQol]` to new shell 

b1llow:
**Your checklist for this pull request**
- [ ] I've read the [guidelines for contributing](https://github.com/rizinorg/rizin/blob/master/DEVELOPERS.md) to this repository
- [ ] I made sure to follow the project's [coding style](https://github.com/rizinorg/rizin/blob/master/DEVELOPERS.md#code-style)
- [ ] I've documented or updated the documentation of every function and struct this PR changes. If not so I've explained why.
- [ ] I've added tests that prove my fix is effective or that my feature works (if possible)
- [ ] I've updated the [rizin book](https://github.com/rizinorg/book) with the relevant information (if needed)

**Detailed description**

convert `px[ dDhHwWqQol]` to new shell
- `pxd` -> `pxdw`
- `pxd 1` -> `pxd`
- `pxd 2` -> `pxdh`
- `pxd 4` -> `pxdw`
- `pxd 8` -> `pxdq`

**Test plan**

...

**Closing issues**

...

PR #2841, opened 2022-07-26
print.c refactor

b1llow:
# SQUASH ME

**Detailed description**

1/n of https://github.com/rizinorg/rizin/pull/2813

**Test plan**

...

**Closing issues**

...


librz/include/rz_util/rz_print.h
```diff
@@ -140,7 +140,7 @@ RZ_API void rz_print_set_flags(RzPrint *p, int _flags);
 RZ_API void rz_print_addr(RzPrint *p, ut64 addr);
 RZ_API void rz_print_section(RzPrint *p, ut64 at);
 RZ_API void rz_print_hexii(RzPrint *p, ut64 addr, const ut8 *buf, int len, int step);
-RZ_API void rz_print_hexdump(RzPrint *p, ut64 addr, const ut8 *buf, int len, int base, int step, size_t zoomsz);
+RZ_API RZ_OWN char *rz_print_hexdump_str(RZ_NONNULL RzPrint *p, ut64 addr, RZ_NONNULL const ut8 *buf, int len, int base, int step, size_t zoomsz);
```
wargio, line 201: i like the changes, but i would prefer to move `rz_print_hexdump_str` under rz_core and just have something like
```c
RZ_API void rz_core_hexdump_block(RZ_NONNULL RzCore *core, int base, int step, size_t zoomsz);
```
the function should use directly `core->offset` and `core->block` otherwise put it under rz_cons as this
```c
RZ_API void rz_cons_hexdump(ut64 addr, RZ_NONNULL const ut8 *buf, int len, int base, int step, size_t zoomsz);
```
@XVilka @ret2libc what do you think?

XVilka: Yes, but printing should not be in the API - it's useless if you use it as a library. String is more useful

XVilka: Fine to move this into the Core, just please return string

librz/core/cmd/cmd_print.c
```diff
@@ -2523,6 +2523,13 @@ static ut8 *old_transform_op(RzCore *core, const char *val, char op, int *buflen
 	return result;
 }
 
+static inline void print_hexdump(RZ_NONNULL RzPrint *p, ut64 addr, RZ_NONNULL const ut8 *buf,
```
```suggestion
static void print_hexdump(RZ_NONNULL RzPrint *p, ut64 addr, RZ_NONNULL const ut8 *buf,
```

librz/io/p/io_gprobe.c
```diff
@@ -1004,7 +1004,9 @@ static int gprobe_getinformation(struct gport *port) {
 
 	ut64 tmpsz;
 	const ut8 *tmp = rz_buf_data(reply, &tmpsz);
-	rz_print_hexdump(NULL, 0, tmp, tmpsz, 16, 1, 1);
+	char *dump = rz_print_hexdump_str(NULL, 0, tmp, tmpsz, 16, 1, 1);
```
ret2libc: I think the first param can't be NULL?

librz/util/print.c
```diff
@@ -653,11 +780,13 @@ RZ_API void rz_print_hexdump(RzPrint *p, ut64 addr, const ut8 *buf, int len, int
 					delta--;
 				}
 				for (i = 0; i < delta; i++) {
-					print(space);
+					rz_strbuf_append(sb, space);
 				}
 			}
+			ut32 K = 0;
+			ut32 k = 0;
```
ret2libc, line 705: dupped

librz/util/print.c
```diff
@@ -512,56 +512,180 @@ RZ_API void rz_print_section(RzPrint *p, ut64 at) {
...
+					rz_strbuf_appendf(sb, "0x%016" PFMT64x "%c", addr, ch);
+				} else {
+					rz_strbuf_appendf(sb, "0x%08" PFMT64x "%c", addr, ch);
+				}
+			}
+		}
+	}
+}
+
+static inline void print_section(RzStrBuf *sb, RzPrint *p, ut64 at) {
```
ret2libc, line 499: I think you can reuse this here in this PR to rewrite `rz_print_section` in terms of this.

ret2libc: Something like
```c
RZ_API void rz_print_section(RzPrint *p, ut64 at) {
   RzStrBuf sb;
   rz_strbuf_init(&sb);
   print_section(&sb, p, at);
   char *s = rz_strbuf_drain(&sb);
   p->cb_printf("%s", s);
   free(s);
   return;
}
```

librz/util/print.c
```diff
@@ -512,56 +512,180 @@ RZ_API void rz_print_section(RzPrint *p, ut64 at) {
...
-	rz_return_if_fail(p && buf && len > 0);
-	PrintfCallback printfmt = (PrintfCallback)printf;
-#define print(x) printfmt("%s", x)
-	bool c = p ? (p->flags & RZ_PRINT_FLAGS_COLOR) : false;
-	const char *color_title = c ? (Pal(p, offset)
-					      : Color_MAGENTA)
-				    : "";
-	int inc = p ? p->cols : 16;
-	size_t i, j, k;
+static inline void print_addr(RzStrBuf *sb, RzPrint *p, ut64 addr) {
```
ret2libc, line 403: please do the same suggested before for rz_print_section but for rz_print_addr and for possibly other code you have duplicated.

librz/include/rz_util/rz_print.h
ret2libc: I prefer avoiding direct use of core->offset and core->block in APIs because they require users of the api to call seek_set and restore the seek after the call, which is painful. It is much easier to use an API that automatically reads the data it needs from `addr`, passed as argument. Also, core->block is just a block of data and maybe the API needs more, so it is in general bad to use IMHO. AFAIR we are trying to move away from core->block in favour of rz_io_read_at.

ret2libc: About Core, I'm ok with keeping *this* function in RzPrint and add `rz_core_hexdump_block` as suggested by @wargio , which simply calls rz_print_hexdump_str and print it.

ret2libc (commented):
Thanks for splitting the previous PR! It is very much appreciated and much easier to review now, thanks!

I think overall it is a good change, just added some minor things.
Thanks again!

librz/include/rz_util/rz_print.h
wargio: :S i actually want to remove completely RzPrint, so this either goes to rz_cons or rz_core for me.

wargio: Since the others are saying to do the opposite as me, then just follow what @ret2libc said, just add an extra method to dump the core->block data.

ret2libc: I'm not saying RzPrint is good actually :D I also think it is a bit useless to have such a thing and probably better to have everything in RzCore, but... let's do that at some point later I guess.

librz/util/print.c
b1llow: There are indeed two variables used here, an uppercase and a lowercase k

ret2libc: oh :/ i did not see they were different cases.... what a nice choice of variable names :D

ret2libc (approved):

ret2libc (approved):

XVilka (approved):

PR #2846, opened 2022-07-27
Use clang-format 14 instead of 13

DMaroo:

**Detailed description**

Bump `clang-format` to version 14 instead of 13

**Test plan**

Linter CI should be green

**Closing issues**

None


librz/asm/arch/include/mybfd.h
```diff
@@ -1506,7 +1506,9 @@ extern const struct bfd_symbol *const bfd_ind_symbol;
 			(struct bfd_symbol **)SYM_PTR, \
 \
 			/* map_head, map_tail                                            */ \
-			{ NULL }, { NULL } \
+			{ NULL }, { \
```
XVilka: This one looks weird.

XVilka (approved):

librz/asm/arch/include/mybfd.h
DMaroo: I agree. Should I change the `.clang-format`?

XVilka: Do you know what would change this behavior? It might be just a one-off case.

DMaroo: No, I do not know. I'll have a look at the documentation and the changelog.

DMaroo: I am unable to find the option which can make these blocks look as they did in clang-format 13.

Changelog: https://releases.llvm.org/14.0.0/tools/clang/docs/ReleaseNotes.html#clang-format
Documentation: https://clang.llvm.org/docs/ClangFormatStyleOptions.html

XVilka: You can just add off/on comments

XVilka: Maybe it also makes sense to open a bug in clang-format itself, it does look like it

DMaroo: I'll probably ask first to the folks at clang-format, and see if it is really not possible to do this. If it indeed is not possible, we can add off/on directives till clang-format fixes it.

ret2libc: i think we can just remove this whole `BFD_FAKE_SECTION` macro as it does not seem to be used anywhere.

ret2libc (commented):
LGTM apart from removing that macro.

librz/asm/arch/include/mybfd.h
DMaroo: @XVilka I did ask in the official LLVM Discord. There has been no response. I guess, we'll just let it be for now. Maybe we can change the `.clang-format` later, if we find the correct option.

@ret2libc Sure, I will remove that macro.

XVilka (approved):

ret2libc (approved):

Issue #635, opened 2021-02-16
Interrupting the http server breaks the command prompt


caribpa:
### Work environment

| Questions                                            | Answers
|------------------------------------------------------|--------------------
| OS/arch/bits (mandatory)                             | Arch Linux x64
| File format of the file you reverse (mandatory)      | n/a
| Architecture/bits of the file (mandatory)            | n/a
| `rizin -v` full output, **not truncated** (mandatory)  | rizin 0.2.0-git @ linux-x86-64, package: 0.2.0.r25761.cf8f661b7-git (AUR) commit: cf8f661b736a8dd9bae2a4ef06b1aeff2041fc6f, build: 2021-02-16__13:10:51

### Expected behavior
Interrupting the http server doesn't break Rizin's command prompt.

### Actual behavior
Interrupting the http server breaks the Rizin's command prompt:
- The prompt becomes colorless
- Writing a command and erasing part of it doesn't clear the removed characters (like some ptys and `vi` versions in insert-mode)

### Steps to reproduce the behavior
- `rz --`
- `=h 8080`
- `<Ctrl-C>`
- 💔 
- `*writes something and deletes part of it*`
- 💔 

### Additional Logs, screenshots, source code,  configuration dump, ...
Exiting Rizin solves the issue.

This issue is also present in radare2.

Issue #560, opened 2021-02-09
Stderr is not passed to remote session


caribpa:
### Work environment

| Questions                                            | Answers
|------------------------------------------------------|--------------------
| OS/arch/bits (mandatory)                             | Arch Linux x64
| File format of the file you reverse (mandatory)      | ELF
| Architecture/bits of the file (mandatory)            | x86/64
| `rizin -v` full output, **not truncated** (mandatory)  | rizin 0.2.0-git @ linux-x86-64 commit: 3e64a2e16a4a4d4bc7879299152e67c99084a685, build: 2021-02-09__17:04:35

### Expected behavior
Opening a remote session with `== <fd>` shows the stderr messages in the client instead of the server.

### Actual behavior
Opening a remote session with `== <fd>` shows the stderr messages only in the server.

#### Server
- `cd /bin`
- `rz rap://:8080`

#### Client
- `rz --`
- `=+ rap://localhost:8080/rizin`
- `== 0`
- `aa`
- 💔 

Using a TCP server (`rz -q -c '=t 8080' rizin`) don't show the stderr in any of the parts of the connection.

Using `-2` in the server only silences the stderr, it is not passed to the client. Not that I expected that flag to do a miracle but, I guess, it proves that the `aa` output goes to stderr (haven't checked the internals).

This issue is also present in radare2.

caribpa:
After #552 resolution, I came to say that this issue is also present when using `=!= <fd>` instead of `== <fd>`.

What's weird is that this issue doesn't happen when directly connecting to the server:
```sh
# Server
$ rz rap://:8080

# Client
$ rz rap://localhost:8080/rizin
> aa
# [x] Analyze all flags starting with sym. and entry0 (aa) ✅ 
```

This issue is also present in radare2.

Issue #554, opened 2021-02-09
Background server blocks Rizin


caribpa:
### Work environment

| Questions                                            | Answers
|------------------------------------------------------|--------------------
| OS/arch/bits (mandatory)                             | Arch Linux x64
| File format of the file you reverse (mandatory)      | n/a
| Architecture/bits of the file (mandatory)            | n/a
| `rizin -v` full output, **not truncated** (mandatory)  | rizin 0.2.0-git @ linux-x86-64 commit: 3e64a2e16a4a4d4bc7879299152e67c99084a685, build: 2021-02-09__17:04:35

### Expected behavior
Starting a background rap server (`=&:`) starts the server in the background, allowing to perform other Rizin commands and exit.

### Actual behavior
Starting a background rap server blocks Rizin (though the server works), making it impossible to kill the server or exit Rizin. Performing `Ctrl+z` and killing the background process is the only way to exit.

### Steps to reproduce the behavior
- `rz -q -c '=&: 8080 ; echo hi' --`
- Mmm, `hi` wasn't printed?
- Presses `Ctrl+c`
- 💔 

This issue is also present in radare2, but I am pretty sure this used to work...

ret2libc:
Trying to reproduce, I get this:
```
$ rizin -q -c '=&: 8080 ; echo hi' --
Failed to set cpu affinity
Background rap server started.
hi
rap: listening at port 8080 ssl off
RAP Server started (rap.loop=true)
^Caccept: Invalid argument
```
and ^C worked.

Could you do `ldd $(which rizin)` please?

caribpa:
Ha, interesting.

Did your server exit after pressing Ctrl+C?

`ldd $(which rizin)`:
```
	linux-vdso.so.1 (0x00007ffebfff0000)
	librz_util.so.0.2.0-git => /usr/bin/../lib/librz_util.so.0.2.0-git (0x00007fe545dd8000)
	librz_main.so.0.2.0-git => /usr/bin/../lib/librz_main.so.0.2.0-git (0x00007fe545d9e000)
	libpthread.so.0 => /usr/bin/../lib/libpthread.so.0 (0x00007fe545d7d000)
	libc.so.6 => /usr/bin/../lib/libc.so.6 (0x00007fe545bb0000)
	libdl.so.2 => /usr/bin/../lib/libdl.so.2 (0x00007fe545ba9000)
	libm.so.6 => /usr/bin/../lib/libm.so.6 (0x00007fe545a64000)
[...]

PR #2860, opened 2022-07-31
Rewrite non-libuv tcp server and drop libuv

thestr4ng3r:
# DO NOT SQUASH

**Detailed description**

Despite being a fantastic library, libuv was only ever used for the tcp server for almost 4 years. This fixes the non-libuv tcp server and drops libuv in favor of it. See commit descriptions for details.

**Test plan**

Warning: if `tcp.islocal=false`, the following will open complete shell access to `0.0.0.0`!
`rz -c "Rt 1337"`, then from another terminal do e.g. `echo "?E hello" | nc localhost 1337` and see the command output.
Ctrl+C should also break the `Rt` command properly.


wargio (approved):
LGTM, but i cannot verify this now on windows

thestr4ng3r:
I did test on windows in detail, but still doesn't hurt to have someone else look through it too.

thestr4ng3r:
NetBSD needs some handling for EINTR on connect()

librz/socket/socket.c
```diff
@@ -898,4 +912,160 @@ RZ_API ut8 *rz_socket_slurp(RzSocket *s, int *len) {
...
+#endif
+}
+
+/**
+ * Block until one of the following events has happened and return:
+ * * \p stop_pipe has been stopped with `rz_stop_pipe_stop()` => `RZ_STOP_PIPE_STOPPED`
+ * * \p sock is ready for read or write (depending on \p sock_write) => `RZ_STOP_PIPE_SOCKET_READY`
+ * * \p timeout_ms milliseconds have passed => `RZ_STOP_PIPE_TIMEOUT`
+ *
+ * This is only supported for non-ssl sockets.
```
ret2libc, line 1035: why?

ret2libc (commented):
Does this fixes by any chance either of these https://github.com/rizinorg/rizin/issues/635 https://github.com/rizinorg/rizin/issues/560 https://github.com/rizinorg/rizin/issues/554 ?

Anyway, code LGTM

GustavoLCR (commented):
Tested on Windows, it works

librz/socket/socket.c
wargio: rz_socket just supports TLS why not just add a eval for this?

thestr4ng3r: The select operates on the raw `sock->fd`. When ssl is used in the `RzSocket`, then it does not have that and instead uses OpenSSL functions on an `s->sfd`. At the moment, this limitation is not an issue since the users of this don't use ssl anyway. Also, I think it is a bit questionable whether ssl support should be built into the socket anyway, but I can see it being useful in some cases.

PR #2867, opened 2022-08-03
Update MacOS installation information

pcolladosoto:
**Detailed description**
We have included information on how to install Rizin both through Homebrew and MacPorts on top of using the package file provided on the releases.

**Closing issues**
This PR is intended to add information on the installation procedures offered after closing issue #160.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -32,6 +32,15 @@ whole directory anywhere.
 ## MacOS
 On MacOS systems you can install Rizin in one of three ways.
 
+Bear in mind the formula offered by Homebrew and the port offered by MacPorts
+are not provided by Rizin: they are offered and actively maintained and updated
+by their respective communities. Rizin *might* help with maintenance and fixes
+in a punctual fashion, but it cannot guarantee these two installation methods
+offer the latest available version.
+
+You can always find the latest version in the package offered on the GitHub
+release.
+
 ### Homebrew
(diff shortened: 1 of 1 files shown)


XVilka (approved):

INSTALL.md
```diff
@@ -30,15 +30,26 @@ anywhere because Rizin is compiled in a "portable" way, allowing moving the
 whole directory anywhere.
 
 ## MacOS
-On MacOS systems, Rizin can be installed through .pkg files attached to the
+On MacOS systems you can install Rizin in one of three ways.
+
+### Homebrew
+Rizin offers a Homebrew [formula](https://formulae.brew.sh/formula/rizin)
```
ret2libc, line 45: both here and in the next section i feel we should be more explicit on the fact that it is not us who provide these things, but they are maintained by the respective communities (homebrew and macports). We might help them keep them updated or fix things here and there, but we are not actively keeping them updated.

ret2libc (commented):
Thanks a lot!

INSTALL.md
pcolladosoto: I added some information tackling this topic. What do you think?

wargio: LGTM

wargio (approved):

ret2libc (approved):

ret2libc:
Thanks! ;)

Issue #2865, opened 2022-08-03
[build] meson_git_wrapper.py called against build dir instead of source dir


alxchk:
Meson 0.63.0

How to reproduce:
```
/tmp > mkdir rizin                                                                                                                                                                                                                             
                                                                                                                                                                                                                                               
rizin > git clone --recursive https://github.com/rizinorg/rizin                                                                                                                                                                                
Cloning into 'rizin'...                                                                                                                                                                                                                        
remote: Enumerating objects: 270927, done.                                                                             
remote: Counting objects: 100% (1819/1819), done.                                                                                                                                                                                              
remote: Compressing objects: 100% (737/737), done.                                                                                                                                                                                             
remote: Total 270927 (delta 1254), reused 1469 (delta 1079), pack-reused 269108                                        
Receiving objects: 100% (270927/270927), 133.28 MiB | 9.06 MiB/s, done.                                                
Resolving deltas: 100% (213637/213637), done.                                                                          
rizin > mkdir rizin-build                                                                                                                                                                                                                      
rizin > cd rizin-build                                                                                                 
rizin-build > meson ../rizin                                                                               
.......
  Subprojects                                                                                                                                                                                                                                  
    capstone-bundled          : YES                                                                                                                                                                                                            
    libdemangle               : YES                                                                                                                                                                                                            
    liblzma                   : YES                                                                                    
    libmspack                 : YES                                                                                    
    libzip                    : YES                                                                                    
    lz4                       : YES                                                                                    
    mpc                       : YES                                                                                    
    nettle                    : YES                                                                                    
    ptrace-wrap               : YES                                                                                    
    rizin-shell-parser        : YES                                                                                    
    rzar                      : YES                                                                                                                                                                                                            
    rzbochs                   : YES                                                                                    
    rzgdb                     : YES                                                                                                                                                                                                            
    rzheap                    : YES                                                                                    
    rzqnx                     : YES                                                                                    
    rzspp                     : YES                                                                                                                                                                                                            
    rzwinkd                   : YES                                                                                    
    tree-sitter               : YES                                                                                                                                                                                                            
    tree-sitter-c             : YES                                                                                                                                                                                                            
    xxhash                    : YES                                                                                    
    yxml                      : YES                                                                                    
    zlib                      : YES                                                                                    
                                                                                                                       
Found ninja-1.11.0 at /usr/bin/ninja                                                                                                                                                                                                           
rizin-build > ninja                                                                                                    
[...]

PR #2868, opened 2022-08-03
build: use meson.current_source_dir() instead of '.'

ret2libc:
Let's see why this was not done from the beginning.

Closes #https://github.com/rizinorg/rizin/issues/2865


p371jah:
I think we can remove this `if meson.is_subproject()` block. It was added by me to introduce support for meson in cutter.

wargio (approved):

ret2libc:
Android needs a different fix as it is a separate thing.

PR #2869, opened 2022-08-04
gittip: Use Meson's `capture: true` for Python 3.6 compatibility

kazarmy:

**Detailed description**

This pr fixes the following problem (https://github.com/rizinorg/rizin/runs/7665608906?check_suite_focus=true#step:7:814):

gittip-capture-output

and reverts `sys/meson_git_wrapper.py` so any lost functionality should also be restored.

**Test plan**

The fix makes sense. All builds are green.

**Closing issues**

...

### DIFF
--- a/sys/meson_git_wrapper.py
+++ b/sys/meson_git_wrapper.py
@@ -25,9 +25,7 @@ def isCArgSupported(executable, path):
 
 def simple_git_execution(args):
     try:
-        called = subprocess.run(args, check=True, capture_output=True)
-        with open("gittip", "w", encoding="utf8") as f:
-            f.write(called.stdout.decode("utf8").strip())
+        called = subprocess.run(args, check=True)
         sys.exit(called.returncode)
     except subprocess.CalledProcessError as e:
         sys.exit(e.returncode)
--- a/librz/util/str.c
+++ b/librz/util/str.c
@@ -4067,6 +4067,7 @@ RZ_API char *rz_str_version(const char *program) {
 	if (!gittip || !*rz_str_trim_head_ro(gittip)) {
 		goto done;
 	}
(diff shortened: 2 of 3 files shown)


XVilka (approved):

sys/meson_git_wrapper.py
```diff
@@ -25,9 +25,7 @@ def isCArgSupported(executable, path):
 
 def simple_git_execution(args):
     try:
-        called = subprocess.run(args, check=True, capture_output=True)
-        with open("gittip", "w", encoding="utf8") as f:
-            f.write(called.stdout.decode("utf8").strip())
+        called = subprocess.run(args, check=True)
```
eli-schwartz, line 28: This is wrong, and using `capture: true` in Meson is bad -- it causes this python script to be run by `meson --internal exe`, which imports lots of stuff and is very slow and obscures the command line.

You don't need to throw the baby out with the bathwater. capture_output is python 3.7, sure, but it's also literally an alias for `stdout=subprocess.PIPE, stderr=subprocess.PIPE` and it's extremely easy to make that work with python 3.6.

ret2libc: Thanks a lot for this hint!

kazarmy: Okay #2873 ... Using `stdout=subprocess.PIPE` changes the behavior of `sys/meson_git_wrapper.py` away from expected behavior, but hopefully this divergence is only temporary.

eli-schwartz: Great. Happy to help. :)

Capture exists solely for the benefit of tools that don't support specifying the output file and usually cannot be modified as they are external processing tools rather than internal helper scripts, so you don't really have a choice.

PR #2878, opened 2022-08-05
Resolve golang string structures in data section

wargio:

**Detailed description**

This resolves the go string structures in a very simplistic way by looking for offsets in sections which names contains the `data` substring. all the golang strings in tables are always set in the following way: `pointer` followed by the string `size`, stored in 32 or 64 bit words.

These structures are especially used in go 1.12 and 1.16

### DIFF
--- a/librz/core/golang.c
+++ b/librz/core/golang.c
@@ -1651,7 +1651,7 @@ static void core_recover_golang_strings_from_data_pointers(RzCore *core, GoStrRe
 				break;
 			}
 
-			if (0 > rz_io_nread_at(core->io, current, buffer, length)) {
+			if (rz_io_nread_at(core->io, current, buffer, length) < 0) {
 				RZ_LOG_ERROR("Failed to read map at address %" PFMT64x "\n", current);
 				break;
 			}
--- a/librz/core/golang.c
+++ b/librz/core/golang.c
@@ -1616,7 +1616,7 @@ static ut32 golang_recover_string_riscv64(GoStrRecover *ctx) {
 // the first offset is always the pointer to the squashed strings and the next word
 // is the size of the string
 static void core_recover_golang_strings_from_data_pointers(RzCore *core, GoStrRecover *ctx) {
-	rz_core_notify_begin(core, "Recovering go strings from the data sections");
+	rz_core_notify_begin(core, "Recovering go strings from bin maps");
 
 	RzAnalysis *analysis = core->analysis;
 	const ut32 word_size = analysis->bits / 8;
@@ -1677,7 +1677,7 @@ static void core_recover_golang_strings_from_data_pointers(RzCore *core, GoStrRe
 
 end:
 	free(buffer);
-	rz_core_notify_done(core, "Recovering go strings from the data sections");
+	rz_core_notify_done(core, "Recovering go strings from bin maps");
 }
 
 /**
--- a/librz/core/golang.c
+++ b/librz/core/golang.c
@@ -4,6 +4,7 @@
 #include <rz_core.h>
 
 #define GO_MAX_STRING_SIZE 0x4000
+#define GO_MAX_TABLE_SIZE  0x10000
 
 #define GO_1_2  (12)
 #define GO_1_16 (116)
@@ -527,6 +528,12 @@ static bool add_new_bin_string(RzCore *core, char *string, ut64 vaddr, ut32 size
 static bool recover_string_at(GoStrRecover *ctx, ut64 str_addr, ut64 str_size) {
 	// check that the values are acceptable.
 	if (str_size < 2 || str_size > GO_MAX_STRING_SIZE || str_addr < 1 || str_addr == UT64_MAX) {
+		return false;
+	}
+
+	// skip possible pointers that matches to symbols flags, because these are already handled.
+	RzFlagItem *fi = rz_flag_get_by_spaces(ctx->core->flags, str_addr, RZ_FLAGS_FS_SYMBOLS, NULL);
+	if (fi && !strncmp(fi->name, "sym.", 4)) {
 		return false;
 	}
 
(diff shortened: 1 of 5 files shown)


librz/core/golang.c
```diff
@@ -530,6 +531,12 @@ static bool recover_string_at(GoStrRecover *ctx, ut64 str_addr, ut64 str_size) {
 		return false;
 	}
 
+	// do not remove any flag that a symbol.
```
ret2libc: rephrase this please

librz/core/golang.c
```diff
@@ -1605,14 +1612,78 @@ static ut32 golang_recover_string_riscv64(GoStrRecover *ctx) {
...
+				break;
+			}
+
+			length -= word_size;
+			for (size_t i = 0; i < length; i += word_size) {
+				string_addr = rz_read_ble(buffer + i, analysis->big_endian, analysis->bits);
+				string_size = rz_read_ble(buffer + i + word_size, analysis->big_endian, analysis->bits);
+				if (!string_addr || !string_size) {
+					continue;
+				} else if (word_size == sizeof(ut32) && string_addr == UT32_MAX) {
```
ret2libc, line 1665: can you have the case word_size == sizeof(ut64)?

wargio: yes, but is already handled by recover_string_at

test/db/analysis/golang
```diff
@@ -249,7 +249,7 @@ EOF
 EXPECT=<<EOF
 1788
 0x004a699b str.hello__hacktivity
-8555
+9222
```
ret2libc, line 266: can we have a test to check specifically for the strings that are detected by this new function?

wargio: i sure can try to find the smallest which doesnt print 9k strings.

ret2libc: just `~XXXX` to check if one (or few) specific string(s) are present. no need to print them all.

wargio: This is the current dev with the new test
image

ret2libc: 👍

librz/core/golang.c
```diff
@@ -1605,14 +1612,80 @@ static ut32 golang_recover_string_riscv64(GoStrRecover *ctx) {
...
+static void core_recover_golang_strings_from_data_pointers(RzCore *core, GoStrRecover *ctx) {
+	rz_core_notify_begin(core, "Recovering go strings from the data sections");
+
+	RzAnalysis *analysis = core->analysis;
+	const ut32 word_size = analysis->bits / 8;
+	RzListIter *iter;
+	RzBinSection *section;
+	ut8 *buffer = NULL;
+	ut64 string_addr, string_size;
+	RzList *section_list = rz_bin_get_sections(core->bin);
```
ret2libc: Do we really need to look at sections? We should tend to avoid relying on sections (and  section names ) if possible.

wargio: actually yes, i could just check if it does not have exec flag, but some PE has data that is executable

ret2libc: really?? that is weird, especially for common windows binaries

XVilka (approved):

XVilka:
```

[XX] db/analysis/golang Resolve all symbols on a stripped linux go1.12 binary
RZ_NOPLUGINS=1 /usr/bin/rizin -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -Qc 'aalg
fl~sym.go.~?
pdfs @ main ~str.
izq~?
' bins/golang/example-server-1.12-stripped
-- stdout
--- expected
+++ actual
@@ -8,4 +8,4 @@
 0x0065de6d str.version
 0x0065de8a str.HTTP_version
 0x0065dfe6 str.
-14338
+14889

-- stderr

[x] Found go 1.12 pclntab data.

[x] Recovered 5414 symbols and saved them at sym.go.*
[ ] Analyze all flags starting with sym.go. (aF @@f:sym.go.*)
[
[x] Analyze all flags starting with sym.go. (aF @@f:sym.go.*)
[ ] Recovering go strings from the data sections
[
[x] Recovering go strings from the data sections
[ ] Analyze all instructions to recover all strings used in sym.go.*
[
[x] Analyze all instructions to recover all strings used in sym.go.*

[x] Recovered 10082 strings from the sym.go.* functions.

[**]                       db/archos/linux-x64/dbg_dmhd     1836 OK        58 BR        0 XX        3 FX
[**]                     db/archos/linux-x64/asm_x64_as     1842 OK        58 BR        0 XX        3 FX
[**]                     db/archos/linux-x64/dbg_trace1     1843 OK        58 BR        0 XX        3 FX


[XX] db/analysis/golang Resolve all symbols on a stripped linux go1.16 binary
RZ_NOPLUGINS=1 /usr/bin/rizin -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -Qc 'aalg
fl~sym.go.~?
pdfs @ main ~str.
izq~?
' bins/golang/example-server-1.16-stripped
-- stdout
--- expected
+++ actual
@@ -12,4 +12,4 @@
 ;-- str.flag:
 ;-- str.sort:
 ;-- str.sync:
-13492
+14028

-- stderr

[x] Found go 1.16 pclntab data.

[x] Recovered 4657 symbols and saved them at sym.go.*
[ ] Analyze all flags starting with sym.go. (aF @@f:sym.go.*)
[
[x] Analyze all flags starting with sym.go. (aF @@f:sym.go.*)
[ ] Recovering go strings from the data sections
[
[x] Recovering go strings from the data sections
[ ] Analyze all instructions to recover all strings used in sym.go.*
[
[x] Analyze all instructions to recover all strings used in sym.go.*

[x] Recovered 10098 strings from the sym.go.* functions.
```

ret2libc (changes requested):
Will review

wargio:
@ret2libc if we merge this then i can reuse the same function on the other PR.

librz/core/golang.c
ret2libc: I'm still not sure about this... By using `rz_bin_get_sections` we are just getting the sections of the binary which, as said elsewhere, do not really mean too much for ELF, for example. For ELF we should use segments, but for other binaries maybe sections are enough... I think in general we may just want to look at `RzBinMap`, which represent a memory mapping from a binary into the virtual memory.

Another reason for using RzBinMap is that actually we want to look at the mapped portions of a binary and sections may contain stuff like .dynamic, .comment, etc. which are not mapped anywhere (thus the read would fail).

wargio: i agree but in this case, we do not care, because these structures are pointing to RO memory normally on golang bins. since we do have different bins using golang we only check for "data" sections/segments/etc..

ret2libc: I think our users would like to reverse even binaries that are stripped and do not have non essential information (e.g. sections). Rizin should try to do its best with a regularly compiled binary but also with a manipulated one.

Do you see any drawback for the regularly compiled binaries in using the rzbinmap approach instead of the sections?

wargio: no.

librz/core/golang.c
```diff
@@ -1605,14 +1612,81 @@ static ut32 golang_recover_string_riscv64(GoStrRecover *ctx) {
...
+		ut64 current = map->vaddr;
+		ut64 end = map->vaddr + map->psize;
+
+		do {
+			size_t length = RZ_MIN((end - current), GO_MAX_TABLE_SIZE);
+			if (length < (word_size * 2)) {
+				break;
+			}
+
+			if (0 > rz_io_nread_at(core->io, current, buffer, length)) {
```
ret2libc: this looks a bit strange TBH. We usually write `if (xxxxx(....) < 0)` rather than `0 > xxxx`.

ret2libc (approved):

librz/core/golang.c
wargio: sure, let me patch that.

PR #2881, opened 2022-08-06
Add references and flags to constant pointers to data maps

wargio:
# DO NOT SQUASH

**Detailed description**

...

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/librz/analysis/analysis.c
+++ b/librz/analysis/analysis.c
@@ -277,7 +277,7 @@ RZ_API bool rz_analysis_set_bits(RzAnalysis *analysis, int bits) {
 		if (analysis->bits != bits) {
 			bool is_hack = is_arm_thumb_hack(analysis, bits);
 			analysis->bits = bits;
-			int v = rz_analysis_archinfo(analysis, RZ_ANALYSIS_ARCHINFO_ALIGN);
+			int v = rz_analysis_archinfo(analysis, RZ_ANALYSIS_ARCHINFO_TEXT_ALIGN);
 			analysis->pcalign = RZ_MAX(0, v);
 			rz_type_db_set_bits(analysis->typedb, bits);
 			rz_type_db_set_address_bits(analysis->typedb, rz_analysis_get_address_bits(analysis));
@@ -311,7 +311,7 @@ RZ_API int rz_analysis_get_address_bits(RzAnalysis *analysis) {
 RZ_API void rz_analysis_set_cpu(RzAnalysis *analysis, const char *cpu) {
 	free(analysis->cpu);
 	analysis->cpu = cpu ? strdup(cpu) : NULL;
-	int v = rz_analysis_archinfo(analysis, RZ_ANALYSIS_ARCHINFO_ALIGN);
+	int v = rz_analysis_archinfo(analysis, RZ_ANALYSIS_ARCHINFO_TEXT_ALIGN);
 	if (v != -1) {
 		analysis->pcalign = v;
 	}
@@ -446,18 +446,37 @@ RZ_API void rz_analysis_purge(RzAnalysis *analysis) {
 	rz_analysis_purge_imports(analysis);
 }
 
-RZ_API int rz_analysis_archinfo(RzAnalysis *analysis, int query) {
-	rz_return_val_if_fail(analysis, -1);
+/**
+ * \brief      Returns the queried information regarding the current architecture
+ *
+ * \param      analysis  The RzAnalysis object to use
+ * \param[in]  query     The architecture detail to query
+ *
+ * \return     Negative when fails.
+ */
+RZ_API int rz_analysis_archinfo(RzAnalysis *analysis, RzAnalysisInfoType query) {
+	rz_return_val_if_fail(analysis && query < RZ_ANALYSIS_ARCHINFO_ENUM_SIZE, -1);
+	if (!analysis->cur || !analysis->cur->archinfo) {
+		switch (query) {
+		case RZ_ANALYSIS_ARCHINFO_CAN_USE_POINTERS:
+			return true;
+		default:
+			return -1;
+		}
+	}
+
+	int value = analysis->cur->archinfo(analysis, query);
 	switch (query) {
 	case RZ_ANALYSIS_ARCHINFO_MIN_OP_SIZE:
-	case RZ_ANALYSIS_ARCHINFO_MAX_OP_SIZE:
-	case RZ_ANALYSIS_ARCHINFO_ALIGN:
-		if (analysis->cur && analysis->cur->archinfo) {
-			return analysis->cur->archinfo(analysis, query);
-		}
-		break;
+		// Always consume at least 1 byte
+		return value > 0 ? value : 1;
+	case RZ_ANALYSIS_ARCHINFO_CAN_USE_POINTERS:
+		// When negative (i.e. error) we assume the architecture does use them.
+		return value < 0 ? true : value;
+	default:
+		return value;
 	}
-	return -1;
+	return value;
 }
 
 #define K_NORET_ADDR(x) sdb_fmt("addr.%" PFMT64x ".noreturn", x)
--- a/librz/core/canalysis.c
+++ b/librz/core/canalysis.c
@@ -5755,6 +5755,121 @@ RZ_IPI bool rz_core_analysis_var_rename(RzCore *core, const char *name, const ch
 	return true;
 }
 
+static bool is_in_data_map(RzCore *core, const ut64 address) {
+	if (address < 256 || address == UT64_MAX) {
+		// prevents a lot of false positives when it comes
+		// to invalid addresses, especially in x86 arch with
+		// mov and jmp instructions.
+		return false;
+	}
+	RzBinObject *o = rz_bin_cur_object(core->bin);
+	if (!o) {
+		return false;
+	}
+
+	const RzBinMap *map = rz_bin_object_get_map_at(o, address, core->io->va);
+	return map && map->psize > 2 && rz_bin_map_is_data(map);
+}
+
+static ut32 add_data_pointer(RzCore *core, const ut8 *bytes, const ut32 size, ut64 pc, ut32 min_op_size) {
+	RzAnalysis *analysis = core->analysis;
+	RzAnalysisOp aop;
+	ut32 isize = 0;
+	ut64 pointer = 0;
+
+	rz_analysis_op_init(&aop);
+	if (rz_analysis_op(analysis, &aop, pc, bytes, size, RZ_ANALYSIS_OP_MASK_DISASM) < 1) {
+		rz_analysis_op_fini(&aop);
+		return min_op_size;
+	}
+
+	isize = aop.size;
+	pointer = aop.ptr;
+	rz_analysis_op_fini(&aop);
+
+	if (pointer == pc ||
+		!is_in_data_map(core, pointer) ||
+		rz_flag_get_list(core->flags, pointer) ||
+		rz_analysis_get_function_at(analysis, pointer)) {
+		return isize;
+	}
+
+	char *flagname = rz_str_newf("data.%08" PFMT64x, pointer);
+	if (!flagname) {
+		RZ_LOG_ERROR("Failed allocate flag name buffer for pointer to data\n");
+		return 0;
+	}
+
+	rz_flag_space_push(core->flags, RZ_FLAGS_FS_POINTERS);
+	rz_flag_set(core->flags, flagname, pointer, analysis->bits / 8);
+	rz_flag_space_pop(core->flags);
+	free(flagname);
+	rz_analysis_xrefs_set(analysis, pc, pointer, RZ_ANALYSIS_XREF_TYPE_DATA);
+	return isize;
+}
+
+/**
+ * \brief      Tries to resolve all the constant pointers and adds flags named data.XXXXXX
+ *
+ * \param      core  The RzCore to analyze
+ */
+RZ_IPI void rz_core_analysis_resolve_pointers_to_data(RzCore *core) {
+	RzAnalysis *analysis = core->analysis;
+
+	bool can_use_pointers = rz_analysis_archinfo(analysis, RZ_ANALYSIS_ARCHINFO_CAN_USE_POINTERS);
+	if (!can_use_pointers) {
+		// never run this for archs that does not use pointers
+		return;
+	}
+
+	RzListIter *it, *it2;
+	RzAnalysisFunction *func = NULL;
+	RzAnalysisBlock *block = NULL;
+	ut8 *bytes = NULL;
+	void *archbits = NULL;
+	ut32 isize = 0, bsize = 0;
+	ut64 pc = 0;
+	ut32 min_op_size = rz_analysis_archinfo(analysis, RZ_ANALYSIS_ARCHINFO_MIN_OP_SIZE);
+
+	// ignore any hint.
+	RZ_PTR_MOVE(archbits, analysis->coreb.archbits);
+
+	rz_list_foreach (analysis->fcns, it, func) {
+		if (rz_cons_is_breaked()) {
+			break;
+		}
+		rz_list_foreach (func->bbs, it2, block) {
+			if (block->size < 1) {
+				continue;
+			}
+
+			bytes = malloc(block->size);
+			if (!bytes) {
+				RZ_LOG_ERROR("Failed allocate basic block bytes buffer\n");
+				goto end;
+			} else if (rz_io_nread_at(core->io, block->addr, bytes, block->size) < 0) {
+				free(bytes);
+				RZ_LOG_ERROR("Failed to read function basic block at address %" PFMT64x "\n", block->addr);
+				goto end;
+			}
+
+			for (ut32 i = 0; i < block->size;) {
+				pc = block->addr + i;
+				bsize = block->size - i;
+				if (!(isize = add_data_pointer(core, bytes + i, bsize, pc, min_op_size))) {
+					free(bytes);
+					goto end;
+				}
+				i += isize;
+			}
+			free(bytes);
+		}
+	}
+
+end:
+	analysis->coreb.archbits = archbits;
+}
+
 static bool is_unknown_file(RzCore *core) {
 	if (core->bin->cur && core->bin->cur->o) {
 		return (rz_list_empty(core->bin->cur->o->sections));
@@ -5937,6 +6052,14 @@ RZ_API bool rz_core_analysis_everything(RzCore *core, bool experimental, char *d
 		rz_core_notify_begin(core, "%s", notify);
 		rz_analysis_dwarf_integrate_functions(core->analysis, core->flags, dwarf_sdb);
 		rz_core_notify_done(core, "%s", notify);
+	}
+
+	if (rz_config_get_b(core->config, "analysis.resolve.pointers")) {
+		notify = "Resolve pointers to data sections";
+		rz_core_notify_begin(core, "%s", notify);
+		rz_core_analysis_resolve_pointers_to_data(core);
+		rz_core_notify_done(core, "%s", notify);
+		rz_core_task_yield(&core->tasks);
 	}
 
 	if (experimental) {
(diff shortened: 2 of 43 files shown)


librz/core/canalysis.c
```diff
@@ -5755,6 +5755,98 @@ RZ_IPI bool rz_core_analysis_var_rename(RzCore *core, const char *name, const ch
...
+	if (address < 256 || address == UT64_MAX) {
+		return false;
+	}
+	RzBinObject *o = rz_bin_cur_object(core->bin);
+	const RzBinSection *section = o ? rz_bin_get_section_at(o, address, core->io->va) : NULL;
+	return section && section->vsize > 0 && strstr(section->name, "data");
+}
+
+/**
+ * \brief      Tries to resolve all the constant pointers and adds flags named ptr.XXXXXX
```
ret2libc:
```suggestion
 * \brief      Tries to resolve all the constant pointers and adds flags named data.XXXXXX
```

librz/core/canalysis.c
```diff
@@ -5755,6 +5755,98 @@ RZ_IPI bool rz_core_analysis_var_rename(RzCore *core, const char *name, const ch
...
+ *
+ * \param      core  The RzCore to analyze
+ */
+RZ_IPI void rz_core_analysis_resolve_pointers_to_data(RzCore *core) {
+	if (rz_str_startswith(rz_config_get(core->config, "asm.arch"), "dalvik")) {
+		// never run this for dalvik
+		return;
+	}
+
```
ret2libc: why?

librz/core/canalysis.c
```diff
@@ -5755,6 +5755,98 @@ RZ_IPI bool rz_core_analysis_var_rename(RzCore *core, const char *name, const ch
...
+			for (ut32 i = 0; i < block->size;) {
+				ut64 pc = block->addr + i;
+				RzAnalysisOp aop;
+				rz_analysis_op_init(&aop);
+				if (rz_analysis_op(core->analysis, &aop, pc, bytes + i, (block->size - i), RZ_ANALYSIS_OP_MASK_DISASM) < 1) {
+					rz_analysis_op_fini(&aop);
+					i += min_op_size;
+					continue;
+				}
+				i += aop.size;
```
ret2libc: we probably need an API to iterate over all instructions in a basicblock and that calls a callback passing each RzAnalysisOp, so that we can avoid  this thing every time.

librz/core/canalysis.c
```diff
@@ -5755,6 +5755,98 @@ RZ_IPI bool rz_core_analysis_var_rename(RzCore *core, const char *name, const ch
 	return true;
 }
 
+static bool is_in_data_section(RzCore *core, const ut64 address) {
+	if (address < 256 || address == UT64_MAX) {
+		return false;
+	}
+	RzBinObject *o = rz_bin_cur_object(core->bin);
+	const RzBinSection *section = o ? rz_bin_get_section_at(o, address, core->io->va) : NULL;
+	return section && section->vsize > 0 && strstr(section->name, "data");
```
ret2libc: what if we rely on segment instead and check for r-- or rw- permissions?

ret2libc (commented):
nice stuff

librz/core/canalysis.c
wargio: there is no need to do this with dalvik stuff since everything is resolved by default. there are no pointers in memory.

librz/core/canalysis.c
wargio: because is not reliable. there are bins with data that has the EXEC flag.

librz/core/canalysis.c
thestr4ng3r: Add this as a comment to the code please

ret2libc: The same might be true for other plugins as well? I don't like having special case in generic code like this, it feels hacky. Would anything happen if we just do it anyway? We should consider that someone not in rizinorg might write a plugin that we are not aware about and the non-plugin-specific code should work well nonetheless.

librz/core/canalysis.c
ret2libc: Btw, i was not suggesting you do this yourself in this PR:) i was just saying, feel free to ignore this comment

librz/core/canalysis.c
ret2libc: Sections are even less reliable considering that in binaries like ELF they can even be removed completely without affecting the program execution at all. 

I think we can rely on both sections and segments. If the address falls in a section with data name, consider it data. Otherwise look at segment and their permissions.

ret2libc: We can consider both Rx and rw and r as data. Rwx means self modifying code and who knows what happens in that case.

librz/core/canalysis.c
wargio: i think many of these things can be solved once we have RzArch.

librz/core/canalysis.c
wargio: i wanted to add this in the PR and then i was like, probably a bad idea :D KISS!!

librz/core/canalysis.c
ret2libc: How? I don't think It will be the magic solution for all of these special if we have around the code base

librz/core/canalysis.c
wargio: hmm, ok agreed.

librz/core/canalysis.c
wargio: just some options to say if the arch does use pointers etc.. like wasm does not either. it is a stack only based arch and the stack does not have a size. just a limit to how many values can be stored.

librz/core/canalysis.c
```diff
@@ -5755,6 +5755,110 @@ RZ_IPI bool rz_core_analysis_var_rename(RzCore *core, const char *name, const ch
...
+	}
+
+	if (strstr(section->name, "data")) {
+		// this ensures that the section will always be used if
+		// contains `data` substring regardless of their permissions.
+		return true;
+	}
+
+	return section->perm && !(section->perm & RZ_PERM_X);
+}
```
wargio: This will be replaced by the new function in https://github.com/rizinorg/rizin/pull/2878 PR

librz/core/canalysis.c
ret2libc: Can we do that already in reanalysis plugins please? It's easier to port to rzarch if it's already in the interface rather than having to chase all ifs around

wargio: what analysis plugin? i think this is the right place to add such check and filter out useless loops. tbh, this probably should be done when we analyze any instruction.

XVilka (approved):
Not a fan of such nestiness, but LGTM until the BB iterator is done.

wargio:
Before this fix, `rz_analysis_archinfo(RZ_ANALYSIS_ARCHINFO_DATA_ALIGN)` always returned `-1`, because the `RZ_ANALYSIS_ARCHINFO_DATA_ALIGN` value was not set in the switch; when the code was refactored, this broke `esil` because several checks that were always true, but now they are actually correctly executed.
To resolve this the dead code has been removed.

librz/analysis/analysis.c
```diff
@@ -446,18 +446,24 @@ RZ_API void rz_analysis_purge(RzAnalysis *analysis) {
 	rz_analysis_purge_imports(analysis);
 }
 
-RZ_API int rz_analysis_archinfo(RzAnalysis *analysis, int query) {
-	rz_return_val_if_fail(analysis, -1);
+RZ_API int rz_analysis_archinfo(RzAnalysis *analysis, RzAnalysisInfoType query) {
```
XVilka, line 457: Please add Doxygen what exactly this does

librz/core/canalysis.c
```diff
@@ -5755,6 +5755,105 @@ RZ_IPI bool rz_core_analysis_var_rename(RzCore *core, const char *name, const ch
 	return true;
 }
 
+static bool is_in_data_map(RzCore *core, const ut64 address) {
+	if (address < 256 || address == UT64_MAX) {
```
XVilka, line 5759: Why? It looks somewhat arbitrary. At least make it a define or something?

librz/core/canalysis.c
```diff
@@ -5755,6 +5755,105 @@ RZ_IPI bool rz_core_analysis_var_rename(RzCore *core, const char *name, const ch
...
+				RZ_LOG_ERROR("Failed allocate basic block bytes buffer\n");
+				goto end;
+			} else if (0 > rz_io_nread_at(core->io, block->addr, bytes, block->size)) {
+				free(bytes);
+				RZ_LOG_ERROR("Failed to read function basic block at address %" PFMT64x "\n", block->addr);
+				goto end;
+			}
+
+			for (ut32 i = 0; i < block->size;) {
+				ut64 pc = block->addr + i;
```
XVilka: I would extract the body of this function into a separate function.

librz/include/rz_analysis.h
```diff
@@ -95,10 +95,15 @@ enum {
 #define RZ_ANALYSIS_ADDR_TYPE_ASCII    1 << 10
 #define RZ_ANALYSIS_ADDR_TYPE_SEQUENCE 1 << 11
 
-#define RZ_ANALYSIS_ARCHINFO_MIN_OP_SIZE 0
-#define RZ_ANALYSIS_ARCHINFO_MAX_OP_SIZE 1
-#define RZ_ANALYSIS_ARCHINFO_ALIGN       2
-#define RZ_ANALYSIS_ARCHINFO_DATA_ALIGN  4
+typedef enum rz_analysis_arch_info_type_t {
```
XVilka, line 98: Please add Doxygen for these

XVilka (changes requested):
Logic looks good, please add a few tests also with this option turned off.

ret2libc:
I will review as well.

test/db/analysis/x86_32
```diff
@@ -42,7 +42,7 @@ aap
 afl.
 EOF
 EXPECT=<<EOF
-0x00401e80 fcn.00401e80
+0x00401e80 data.00401e80
```
ret2libc: This seems wrong

test/db/formats/elf/elf-riscv64
```diff
@@ -422,7 +422,7 @@ laddr    0x00000000
 lang     c
 machine  RISC V
 maxopsz  6
-minopsz  2
+minopsz  4
```
ret2libc, line 425: Is this alright? Just asking, I know nothing about RISC V arch.

librz/core/canalysis.c
```diff
@@ -5755,6 +5755,105 @@ RZ_IPI bool rz_core_analysis_var_rename(RzCore *core, const char *name, const ch
...
+		rz_list_foreach (func->bbs, it2, block) {
+			if (block->size < 1) {
+				continue;
+			}
+
+			bytes = malloc(block->size);
+			if (!bytes) {
+				RZ_LOG_ERROR("Failed allocate basic block bytes buffer\n");
+				goto end;
+			} else if (0 > rz_io_nread_at(core->io, block->addr, bytes, block->size)) {
```
ret2libc:
```suggestion
			}
			if (0 > rz_io_nread_at(core->io, block->addr, bytes, block->size)) {
```

librz/core/canalysis.c
ret2libc: I would remove it completely, if possible. Otherwise, comment it clearly as well

ret2libc:
> Before this fix, `rz_analysis_archinfo(RZ_ANALYSIS_ARCHINFO_DATA_ALIGN)` always returned `-1`, because the `RZ_ANALYSIS_ARCHINFO_DATA_ALIGN` value was not set in the switch; when the code was refactored, this broke `esil` because several checks that were always true, but now they are actually correctly executed. To resolve this the dead code has been removed.

Are you sure the code was dead? I see 
```c
static int archinfo(RzAnalysis *analysis, int q) {
	if (q == RZ_ANALYSIS_ARCHINFO_DATA_ALIGN) {
		return 4;
	}
```
in analysis_arm_cs.c

test/db/analysis/x86_32
wargio: no it is not. i checked this and is correct. the additional code adds the reference, then aap adds the function which is already named (because the flag exists) thus keeping the name.

test/db/formats/elf/elf-riscv64
wargio: yes it is, it was wrong before.

wargio:
> > Before this fix, `rz_analysis_archinfo(RZ_ANALYSIS_ARCHINFO_DATA_ALIGN)` always returned `-1`, because the `RZ_ANALYSIS_ARCHINFO_DATA_ALIGN` value was not set in the switch; when the code was refactored, this broke `esil` because several checks that were always true, but now they are actually correctly executed. To resolve this the dead code has been removed.
> 
> Are you sure the code was dead? I see
> 
> ```c
> static int archinfo(RzAnalysis *analysis, int q) {
> 	if (q == RZ_ANALYSIS_ARCHINFO_DATA_ALIGN) {
> 		return 4;
> 	}
> ```
> 
> in analysis_arm_cs.c

yes it is dead code because of this:
```c

RZ_API int rz_analysis_archinfo(RzAnalysis *analysis, int query) {
	rz_return_val_if_fail(analysis, -1);
	switch (query) {
	case RZ_ANALYSIS_ARCHINFO_MIN_OP_SIZE:
	case RZ_ANALYSIS_ARCHINFO_MAX_OP_SIZE:
	case RZ_ANALYSIS_ARCHINFO_ALIGN:
		if (analysis->cur && analysis->cur->archinfo) {
			return analysis->cur->archinfo(analysis, query);
		}
		break;
	}
	return -1;
}
```

librz/analysis/analysis.c
```diff
@@ -446,18 +446,24 @@ RZ_API void rz_analysis_purge(RzAnalysis *analysis) {
...
-	case RZ_ANALYSIS_ARCHINFO_MAX_OP_SIZE:
-	case RZ_ANALYSIS_ARCHINFO_ALIGN:
-		if (analysis->cur && analysis->cur->archinfo) {
-			return analysis->cur->archinfo(analysis, query);
-		}
-		break;
+		// Always consume at least 1 byte
+		return value > 0 ? value : 1;
+	case RZ_ANALYSIS_ARCHINFO_CAN_USE_POINTERS:
+		// When negative we assume the architecture does use them.
```
ret2libc: I don't understand this... For example, amd29k returns true for `RZ_ANALYSIS_ARCHINFO_CAN_USE_POINTERS`, which is positive, why do you say negative is true?

wargio: you can assume that  when a developer does not handle `RZ_ANALYSIS_ARCHINFO_CAN_USE_POINTERS` the return value can be `-1`, but we always want to return a boolean, thus i prefer to handle that case even tho -1 is "true"

librz/core/canalysis.c
wargio: No, this is not useless. there are a lot of false positives when it comes to addresses lower than 256, just imagine for example:
```
@@ --1,14 +-1,14 @@
 |       ,=< 0x000113bd      je    0x1144e
-|       |   0x000113c3      mov   esi, 0x2f                            ; '/' ; int c
+|       |   0x000113c3      mov   esi, 0x2f                            ; data.0000002f ; int c
 |       |   0x000113c8      call  sym.imp.strrchr                      ; char *strrchr(const char *s, int c)
 |       |   0x000113cd      test  rax, rax
 |      ,==< 0x000113d0      je    0x11424
 |      ||   0x000113d2      lea   rdx, [rax + 1]
 |      ||   0x000113d6      mov   rcx, rdx
 |      ||   0x000113d9      sub   rcx, rbx
-|      ||   0x000113dc      cmp   rcx, 6
+|      ||   0x000113dc      cmp   rcx, 6                               ; data.00000006
 |     ,===< 0x000113e0      jle   0x11424
 |     |||   0x000113e2      lea   rsi, [rax - 6]
 |     |||   0x000113e6      mov   ecx, 7
```

XVilka: Then please add a comment.

XVilka (changes requested):
LGTM, apart from two missing points:
- Comment for that function
- Tests with this feature disabled

librz/core/canalysis.c
wargio: i did.

wargio:
> LGTM, apart from two missing points:
> 
>     * Comment for that function
> 
>     * Tests with this feature disabled

I do not know yet how i should write the test for disabling the option, but what i can tell you is there is the comment.

librz/core/canalysis.c
```diff
@@ -5755,6 +5755,121 @@ RZ_IPI bool rz_core_analysis_var_rename(RzCore *core, const char *name, const ch
 	return true;
 }
 
+static bool is_in_data_map(RzCore *core, const ut64 address) {
+	if (address < 256 || address == UT64_MAX) {
+		// prevents a lot of false positives when it comes
+		// to invalid addresses, especially in x86 arch with
+		// mov and jmp instructions.
+		return false;
+	}
```
wargio, line 5764: @XVilka

XVilka (approved):

ret2libc (approved):

XVilka:
@wargio looks like it causes two fuzzing tests timeout:
```

[XX] TIMEOUT /Users/runner/work/rizin/rizin/test/bins/fuzzed <fuzz> /Users/runner/work/rizin/rizin/test/bins/fuzzed/a5ab54823e83703854378cb5bc1f587a
RZ_NOPLUGINS=1 /Users/runner/bin/rizin -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -Qc aaa /Users/runner/work/rizin/rizin/test/bins/fuzzed/a5ab54823e83703854378cb5bc1f587a
-- stdout

-- stderr
Copyright: SEGA MEGASIS    (C)2015 James L.
DomesticName: Liquid Space Dodger V3                          
OverseasName: Liquid Space Dodger V3                          
ProductCode: GM 01234567-89
Checksum: 0xadde
Peripherials: J               
SramCode:             
ModemCode:             
CountryCode: 
[ ] Analyze all flags starting with sym. and entry0 (aa)
[

[ ] Analyze function calls
[
[x] Analyze function calls
[ ] Analyze len bytes of instructions for references
[
[x] Analyze len bytes of instructions for references
[ ] Check for classes
[
[x] Check for classes

[x] Finding xrefs in noncode section with analysis.in=io.maps

[x] Analyze value pointers (aav)

[x] Value from 0x00000200 to 0x0001d4f0 (aav)

[x] 0x00000200-0x0001d4f0 in 0x200-0x1d4f0 (aav)
[ ] Emulate functions to find computed references
[
[x] Emulate functions to find computed references
[ ] Analyze local variables and arguments
[
[x] Analyze local variables and arguments
[ ] Applying signatures from sigdb
[
[x] Applied 0 FLIRT signatures via sigdb
[ ] Propagate noreturn information
[
[x] Propagate noreturn information
[ ] Resolve pointers to data sections
[
-- exit status: -1


[XX] TIMEOUT /Users/runner/work/rizin/rizin/test/bins/fuzzed <fuzz> /Users/runner/work/rizin/rizin/test/bins/fuzzed/9255cead7e2cb22611359fec200090da
RZ_NOPLUGINS=1 /Users/runner/bin/rizin -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -Qc aaa /Users/runner/work/rizin/rizin/test/bins/fuzzed/9255cead7e2cb22611359fec200090da
-- stdout

-- stderr
Copyright: SEGA MEGASIS    (C)2015 James L.
DomesticName: Liquid Space Dodger V3                          
OverseasName: Liquid Space Dodger V3                          
ProductCode: GM 01234567-89
Checksum: 0xadde
Peripherials: J               
SramCode:             
ModemCode:             
CountryCode: 
[ ] Analyze all flags starting with sym. and entry0 (aa)
[

[ ] Analyze function calls
[
[x] Analyze function calls
[ ] Analyze len bytes of instructions for references
[
[x] Analyze len bytes of instructions for references
[ ] Check for classes
[
[x] Check for classes

[x] Finding xrefs in noncode section with analysis.in=io.maps

[x] Analyze value pointers (aav)

[x] Value from 0x00000200 to 0x0001d503 (aav)

[x] 0x00000200-0x0001d503 in 0x200-0x1d503 (aav)
[ ] Emulate functions to find computed references
[
[x] Emulate functions to find computed references
[ ] Analyze local variables and arguments
[
[x] Analyze local variables and arguments
[ ] Applying signatures from sigdb
[
[x] Applied 0 FLIRT signatures via sigdb
[ ] Propagate noreturn information
[
[x] Propagate noreturn information
[ ] Resolve pointers to data sections
[
-- exit status: -1
[**]    /Users/runner/work/rizin/rizin/test/bins/fuzzed      544 OK         0 BR        2 XX        0 FX
```

PR #2892, opened 2022-08-08
block cleanup mechanism proposal.

devnexen:
works mainly with clang/libBlocksRuntime but has gcc implementation as well.
conceal in linux_desc_list for the moment as proof of concept.

 

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,4 +1,3 @@
-Language: Cpp
 MaxEmptyLinesToKeep: 1
 SpaceBeforeParens: ControlStatements
 SpaceInEmptyParentheses: false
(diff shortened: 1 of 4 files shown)


ret2libc:
Where are you going to use this new thing? What are the advantages and disadvantages? Why should we include it?

devnexen:
It would have been useful but I did not realized rizin was supposed to compile with tinycc and tinycc supports very few attributes :)

ret2libc:
> It would have been useful but I did not realized rizin was supposed to compile with tinycc and tinycc supports very few attributes :)

It is the first time I see this block thing, how is it useful?
We try to support as many compilers as possible, but in general this should not block us from using more updated things when they make sense (e.g. we can just do nothing on the compilers that don't support a given flag/option/feature).

devnexen:
It is a sort of golang defer mechanism, ie the block within is called when out of scope, handy for freeing resources, it s getting tricky when there is goto all over the place tough :)

ret2libc:
Ah ok it's like the clean up attribute, i see. Yeah that would be useful but it's not supported on all compilers systems unfortunately, so we can't really on that to free resources

devnexen:
In fact it s literally using cleanup attribute in a smart way, ie for each instruction there is a distinct cleanup attribute ;-) hopefully the next C iteration will provide a standard mechanism for that.

PR #2910, opened 2022-08-12
Avoid warning about unused endian for single byte

karliss:

**Detailed description**

Currently when building Cutter log gets spammed full with warnings about unused endianess in the 8bit versions of these functions/macros.

**Test plan**

Make sure all the CI platforms still build. I assume core utility functions like this have sufficient test coverage.

**Closing issues**

...

### DIFF
--- a/librz/include/rz_util/rz_buf.h
+++ b/librz/include/rz_util/rz_buf.h
@@ -234,8 +234,8 @@ DEFINE_RZ_BUF_WRITE_BLE(64)
 		return true; \
 	}
 
-#define rz_buf_read_ble8_at(b, addr, result, endian) rz_buf_read8_at(b, addr, result)
-#define rz_buf_write_ble8_at(b, addr, value, endian) rz_buf_write8_at(b, addr, value)
+#define rz_buf_read_ble8_at(b, addr, result, endian) ((void)endian, rz_buf_read8_at(b, addr, result))
+#define rz_buf_write_ble8_at(b, addr, value, endian) ((void)endian, rz_buf_write8_at(b, addr, value))
 
 /**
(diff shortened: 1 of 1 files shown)


thestr4ng3r (approved):
Change is fine ofc.

Have you tried removing the endian arg here completely?
Also, what compiler do you use that emits the warnings?

karliss:
> Have you tried removing the endian arg here completely?

It is called by `DEFINE_RZ_BUF_WRITE_OFFSET_BLE` macro which is parametrized by size and does stuff like 
```c
if (!rz_buf_write_ble##size##_at(b, *offset, value, big_endian)) { \
```
So the argument can't be easily removed without changing a lot more code.


> Also, what compiler do you use that emits the warnings?

g++ (GCC) 12.1.1 20220730

Note that it is from the included headers when compiling Cutter  which might have slightly different warning configuration than rizin. Didn't investigate what happens when compiling rizin separately.

librz/include/rz_util/rz_buf.h
```diff
@@ -234,8 +234,8 @@ DEFINE_RZ_BUF_WRITE_BLE(64)
 		return true; \
 	}
 
-#define rz_buf_read_ble8_at(b, addr, result, endian) rz_buf_read8_at(b, addr, result)
-#define rz_buf_write_ble8_at(b, addr, value, endian) rz_buf_write8_at(b, addr, value)
+#define rz_buf_read_ble8_at(b, addr, result, endian) ((void)endian, rz_buf_read8_at(b, addr, result))
+#define rz_buf_write_ble8_at(b, addr, value, endian) ((void)endian, rz_buf_write8_at(b, addr, value))
```
wargio, line 238: how about removing it?

PR #2918, opened 2022-08-14
Fix asm_riscv on big endian hosts

thestr4ng3r:

**Detailed description**

This fixes all tests in test/db/asm/riscv_32 on be, tested on
OpenBSD/sparc64.


XVilka (approved):
LGTM, but can't we read from buf directly? I thought we have all the necessary functions.

thestr4ng3r:
> LGTM, but can't we read from buf directly? I thought we have all the necessary functions.

The code reads only up to 8 bytes from buf, not necessarily 8. The real size is then determined by `riscv_insn_length()`. So if there are less than 8 bytes available and we read directly from buf, it would either read oob or the logic would be more complicated.

PR #2919, opened 2022-08-14
Fix rz_core_search_value_in_range (aav) for big endian hosts

thestr4ng3r:

**Detailed description**

The *(uut64 *)v and similar were naturally dependent on the host endian,
so all current tests assume little endian there, meaning we should use
`rz_read_le...` to fix it on big endian machines without touching the
behavior for now. Later it will be interesting to make this analysis
also respect the to-be-analyzed arch's endianness.

On top of that, the uut64 type's desired unaligned property does not
seem to work on OpenBSD/sparc64 here and it still crashed when the
pointer was unaligned, for example when doing aav on
test/bins/mach0/misaligned_data-iOS-armv7.

This fixes test/db/analysis/arm and possibly others too.


XVilka (approved):
Wow, that is the nasty bug.

PR #2944, opened 2022-08-20
Meson refactor: collect modules in dict and centralize config

thestr4ng3r:

**Detailed description**

Duplicated code for pkg-config and cmake configuration is avoided by
letting every rizin module register itself in a global dictionary
called "modules" which is then iterated in one place to perform any
shared logic, inspired by how qemu handles multiple targets in its meson
build system.

This is an almost pure refactor, so the resulting .pc/.cmake files
should be identical with the following exceptions:
- rz_reg.pc does not have plugindir anymore as rz_reg has no plugins.
- Some other modules have their dependencies in .pc files altered to be
  consistent with cmake.

**Test plan**

Compiling should work and installed `lib/cmake` and `lib/pkgconfig` contents should be identical to before except where noted above.

### DIFF
--- a/librz/io/meson.build
+++ b/librz/io/meson.build
@@ -152,36 +152,8 @@ rz_io_dep = declare_dependency(link_with: rz_io,
                               include_directories: platform_inc)
 meson.override_dependency('rz_io', rz_io_dep)
 
-pkgconfig_mod.generate(rz_io,
-  subdirs: 'librz',
-  version: rizin_version,
-  name: 'rz_io',
-  filebase: 'rz_io',
-  requires: [
-    'rz_util',
-    'rz_socket'
-  ],
-  description: 'rizin foundation libraries',
-  variables: [
-    'plugindir=@0@'.format(rizin_plugins),
-    'datdir=@0@'.format(rizin_datdir_rz),
-  ],
-)
-
-if not is_static_libs_only
-  conf = configuration_data()
-  conf.set('RZ_VERSION', rizin_version)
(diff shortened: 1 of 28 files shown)


ret2libc:
Re-submit from dist- branch.

librz/io/meson.build
```diff
@@ -152,36 +152,8 @@ rz_io_dep = declare_dependency(link_with: rz_io,
...
-    name: conf.get('rizin_libname'),
-    input: '../RzModulesConfig.cmake.in',
-    install_dir: rizin_cmakedir / conf.get('rizin_libname'),
-    configuration: conf,
-  )
-endif
+modules += { 'rz_io': {
+    'target': rz_io,
+    'dependencies': ['rz_util', 'rz_socket', 'rz_hash', 'rz_crypto', 'rz_cons'],
+    'plugins': [io_plugins]
```
ret2libc, line 158: shall `plugins` be a list? or is it enough to do `'plugins': io_plugins`?

ret2libc (approved):
Awesome, that's great Florian, I should have done this long ago :(
As said, re-submit from dist- first, but overall I think it's fine.

thestr4ng3r:
superseded by https://github.com/rizinorg/rizin/pull/2945

librz/io/meson.build
thestr4ng3r: `librz/bin/meson.build` currently needs it for `bin_xtr_...`.

PR #2874, opened 2022-08-05
Use GCC 12.x instead of GCC 11.x on CI job

XVilka:
**Your checklist for this pull request**
- [x] I've read the [guidelines for contributing](https://github.com/rizinorg/rizin/blob/master/DEVELOPERS.md) to this repository
- [x] I made sure to follow the project's [coding style](https://github.com/rizinorg/rizin/blob/master/DEVELOPERS.md#code-style)
- [ ] I've documented or updated the documentation of every function and struct this PR changes. If not so I've explained why.
- [ ] I've added tests that prove my fix is effective or that my feature works (if possible)
- [ ] I've updated the [rizin book](https://github.com/rizinorg/book) with the relevant information (if needed)

**Detailed description**

For some reason, `ubuntu:hirsuite` started to produce errors, lets's update to the latest available Ubuntu Kinetic (22.0) which will be released on October 2022.

**Test plan**

CI is green

PR #2998, opened 2022-09-01
Remove rz_str_mb_to_wc_l()

kazarmy:

**Detailed description**

This pr removes `rz_str_mb_to_wc_l()` since it uses `mbstowcs()` in an unsupported manner -- `len` there is actually ignored. From [man mbstowcs](https://linux.die.net/man/3/mbstowcs):

"If _dest_ is NULL, _n_ is ignored, and the conversion proceeds as above, except that the converted wide characters are not written out to memory, and that no length limit exists."

Apparently there's no replacement glibc function that can be used but `rz_str_mb_to_wc_l()` isn't used anywhere in Rizin. This is a cherry-pick from #2997. The use of 0 for the last mbstowcs parameter fixes https://github.com/rizinorg/rizin/pull/2874#issuecomment-1206506805.

**Test plan**

All builds are green.

**Closing issues**

...

### DIFF
--- a/librz/include/rz_util/rz_str.h
+++ b/librz/include/rz_util/rz_str.h
@@ -197,7 +197,6 @@ RZ_API void rz_str_filter(char *str);
 RZ_API const char *rz_str_tok(const char *str1, const char b, size_t len);
 RZ_API wchar_t *rz_str_mb_to_wc(const char *buf);
 RZ_API char *rz_str_wc_to_mb(const wchar_t *buf);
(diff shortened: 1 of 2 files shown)


XVilka (approved):

kazarmy:
> Apparently there's no replacement glibc function that can be used ...

There is actually the [mbsnrtowcs](https://linux.die.net/man/3/mbsnrtowcs) function that can be used but it's POSIX only and there's some ambiguity regarding its _nms_ parameter on whether it refers to single bytes or multibyte characters.

Issue #1590, opened 2021-09-03
Port `p` (print) to rzshell


ret2libc:
## Completed

- [x] `pd` https://github.com/rizinorg/rizin/pull/2552
- [x] `pc`
- [x] `pg`
- [x] `pm`
- [x] `ph`
- [x] `pt`
- [x] `p6` https://github.com/rizinorg/rizin/pull/2819
- [x] `pa` https://github.com/rizinorg/rizin/pull/2824
- [x] `pA` https://github.com/rizinorg/rizin/pull/3163
- [x] `pb`/`pB` https://github.com/rizinorg/rizin/pull/2828
- [x] `pC` https://github.com/rizinorg/rizin/pull/3213
- [x] `pD`
- [x] `pF` https://github.com/rizinorg/rizin/pull/3071
- [x] `pi` https://github.com/rizinorg/rizin/pull/3187
- [x] `pI` https://github.com/rizinorg/rizin/pull/3163
- [x] `pj` https://github.com/rizinorg/rizin/pull/4726
- [x] `pJ` (removed)
- [x] `pk`/`pK` https://github.com/rizinorg/rizin/pull/3178
- [x] `po` https://github.com/rizinorg/rizin/pull/3146
- [x] `pp` https://github.com/rizinorg/rizin/pull/3112
- [x] `pr` https://github.com/rizinorg/rizin/pull/3089
- [x] `ps` https://github.com/rizinorg/rizin/pull/3451
- [x] `px` https://github.com/rizinorg/rizin/pull/2853
- [x] `pv` https://github.com/rizinorg/rizin/pull/3069
- [x] `pu` https://github.com/rizinorg/rizin/pull/3163
- [x] `p-` https://github.com/rizinorg/rizin/pull/3198
- [x] `p=` https://github.com/rizinorg/rizin/pull/3324
- [x] `pz` (removed)
- [x] `pf` https://github.com/rizinorg/rizin/pull/3467
- [x]  `p2` https://github.com/rizinorg/rizin/pull/4250
- [x]  `p8` https://github.com/rizinorg/rizin/pull/4250

## Unfinished

- [x] `px/` blocked on https://github.com/rizinorg/rizin/issues/1586

PeiweiHu:
ICYMI, `Convert px[ dDhHwWqQol] to new shell` -> https://github.com/rizinorg/rizin/pull/2813

PeiweiHu:
`pD` is ported already.

PR #3069, opened 2022-10-07
Convert `pv` commands to rzshell

XVilka:

**Detailed description**

- Convert `pv` commands to rzshell
- Removed `pV` commands that are same as `pv` but with `useBytes == true`. No tests exists for those
- Removed `pvz` commands as undocumented and untested

**Test plan**

CI is green

**Closing issues**

Partially solves https://github.com/rizinorg/rizin/issues/1590

### DIFF
--- a/librz/core/cmd/cmd_print.c
+++ b/librz/core/cmd/cmd_print.c
@@ -394,17 +394,6 @@ static const char *help_msg_ps[] = {
 	NULL
 };
 
-static const char *help_msg_pv[] = {
-	"Usage: pv[j][1,2,4,8,z]", "", "",
-	"pv", "", "print bytes based on asm.bits",
-	"pv1", "", "print 1 byte in memory",
-	"pv2", "", "print 2 bytes in memory",
-	"pv4", "", "print 4 bytes in memory",
-	"pv8", "", "print 8 bytes in memory",
-	"pvz", "", "print value as string (alias for ps)",
-	NULL
-};
-
 static const char *help_msg_px[] = {
 	"Usage:", "px[0afoswqWqQ][f]", " # Print heXadecimal",
 	"px", "", "show hexdump",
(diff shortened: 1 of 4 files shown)


librz/core/cmd_descs/cmd_print.yaml
```diff
@@ -498,6 +498,78 @@ commands:
...
+        type: RZ_CMD_DESC_TYPE_ARGV_STATE
+        modes:
+          - RZ_OUTPUT_MODE_STANDARD
+          - RZ_OUTPUT_MODE_JSON
+          - RZ_OUTPUT_MODE_RIZIN
+        args:
+          - name: repeat
+            type: RZ_CMD_ARG_TYPE_NUM
+            optional: true
+      - name: pv1
```
wargio, line 558: i wonder if it makes more sense to just have `pv <bytes> <repeat>`

wargio:
```
[XX] db/cmd/cmd_print_misc pvj JSON indent - crash
RZ_NOPLUGINS=1 /usr/bin/rizin -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -Qc 'pvj ~{' =
-- stdout
--- expected
+++ actual
@@ -1,1 +1,1 @@
-[{"value":0,"string":""}]
+{"value":0,"string":""}
```

librz/core/cmd_descs/cmd_print.yaml
ret2libc: I think we should stick with `pvX` for now because we also have `wvX`.

librz/core/cmd_descs/cmd_print.yaml
```diff
@@ -498,6 +498,78 @@ commands:
...
+        summary: print bytes
+        cname: print_value
+        type: RZ_CMD_DESC_TYPE_ARGV_STATE
+        modes:
+          - RZ_OUTPUT_MODE_STANDARD
+          - RZ_OUTPUT_MODE_JSON
+          - RZ_OUTPUT_MODE_RIZIN
+        args:
+          - name: repeat
+            type: RZ_CMD_ARG_TYPE_NUM
```
ret2libc, line 556:
```suggestion
            type: RZ_CMD_ARG_TYPE_RZNUM
```

ret2libc: also in the other places

librz/core/cmd/cmd_print.c
```diff
@@ -6599,3 +6382,219 @@ RZ_IPI RzCmdStatus rz_print_byte_bitstream_handler(RzCore *core, int argc, const
...
+	if (repeat < 0) {
+		return RZ_CMD_STATUS_ERROR;
+	}
+	PrintValueOptions opts = {
+		.size = 2,
+		.repeat = repeat,
+		.useBytes = false
+	};
+	print_value(core, &opts, state);
+	return RZ_CMD_STATUS_OK;
```
ret2libc: this can be extracted in a function that just accept `size` as argument

librz/core/cmd/cmd_print.c
```diff
@@ -6599,3 +6382,219 @@ RZ_IPI RzCmdStatus rz_print_byte_bitstream_handler(RzCore *core, int argc, const
...
+	};
+	int n = core->rasm->bits / 8;
+	for (int i = 0; stack[i]; i++) {
+		if (!strcmp(argv[1], stack[i])) {
+			// TODO: Use the API
+			rz_core_cmdf(core, "ps @ [`drn sp`+%d]", n * i);
+		}
+	}
+	return RZ_CMD_STATUS_OK;
+}
```
ret2libc: can we kill this functionality? it is not tested, it is not explained in the doc and it is unclear :(

librz/core/cmd/cmd_print.c
```diff
@@ -6599,3 +6382,219 @@ RZ_IPI RzCmdStatus rz_print_byte_bitstream_handler(RzCore *core, int argc, const
...
+	print_value(core, &opts, state);
+	return RZ_CMD_STATUS_OK;
+}
+
+RZ_IPI RzCmdStatus rz_print_value_asciiz_handler(RzCore *core, int argc, const char **argv, RzCmdStateOutput *state) {
+	if (argc < 2) {
+		// TODO: Use the API
+		rz_core_cmdf(core, "ps");
+		return RZ_CMD_STATUS_OK;
+	}
```
ret2libc: This doesn't handle the JSON/RIZIN output modes.

librz/core/cmd/cmd_print.c
```diff
@@ -6599,3 +6382,219 @@ RZ_IPI RzCmdStatus rz_print_byte_bitstream_handler(RzCore *core, int argc, const
...
+static bool print_value(RzCore *core, PrintValueOptions *opts, RzCmdStateOutput *state) {
+	ut64 old_at = core->offset;
+	ut8 *block = core->block;
+	int blocksize = core->blocksize;
+	ut8 *block_end = core->block + blocksize;
+
+	if (block + 8 >= block_end) {
+		RZ_LOG_ERROR("core: block is truncated.\n");
+		return false;
+	}
```
ret2libc, line 6491: I think the block concept can be removed and we can use `rz_io_nread_at` instead. In any case, we can probably avoid the `rz_core_seek` in the do/while loop below and just keep track of `at` manually (you can use `ps @ {addr}` in the print_value_single function).

ret2libc (changes requested):

librz/core/cmd/cmd_print.c
XVilka: With pleasure.

librz/core/cmd/cmd_print.c
XVilka: Removed this code

librz/core/cmd/cmd_print.c
XVilka: I would prefer to postpone this until converting to the API calls, after `ps` are converted to the rzshell and a better API.

librz/core/cmd_descs/cmd_print.yaml
XVilka: `repeat` doesn't need to be resolved from flags or anything like that, just a number

librz/core/cmd/cmd_print.c
XVilka: Done

librz/core/cmd_descs/cmd_print.yaml
ret2libc: then let's not use core->num in rz_num_math

XVilka: Done

librz/core/cmd/cmd_print.c
```diff
@@ -6627,3 +6410,184 @@ RZ_IPI RzCmdStatus rz_cmd_print_axml_handler(RzCore *core, int argc, const char
...
+	return RZ_CMD_STATUS_OK;
+}
+
+RZ_IPI RzCmdStatus rz_print_value4_handler(RzCore *core, int argc, const char **argv, RzCmdStateOutput *state) {
+	int repeat = argc > 1 ? rz_num_math(NULL, argv[1]) : 1;
+	if (repeat < 0) {
+		return RZ_CMD_STATUS_ERROR;
+	}
+	print_value_size(core, state, 4, repeat);
+	return RZ_CMD_STATUS_OK;
```
```suggestion
	return print_value_size(core, argc, argv, state, 4, repeat);
```

I think all these functions should be no more than a call to another function where everything is done. They all do the same except from the size, so it's better to have just one place to handle everything IMO.

ret2libc (approved):
LGTM once that last comment is fixed.

librz/core/cmd/cmd_print.c
XVilka: Good point. Done

PR #3090, opened 2022-10-11
Threaded analysis matching

wargio:
# squash me
 

**Detailed description**

Optimize search by matching using also threads


test/db/tools/rz_diff
```diff
@@ -255,7 +255,7 @@ NAME=rz-diff functions comparison JSON
...
...5},"modified":{"name":"fcn.004037f0","addr":4208624,"size":35}},{"similarity":1.000000,"type":"COMPLETE","original":{"name":"fcn.00403810","addr":4208656,"size":67},"modified":{"name":"fcn.00403820","addr":4208672,"size":67}},{"similarity":0.972222,"type":"PARTIAL","original":{"name":"fcn.00403860","addr":4208736,"size":108},"modified":{"name":"fcn.00403870","addr":4208752,"size":108}},{"similarity":0.958621,"type":"PARTIAL","original":{"name":"fcn.004038e0","addr":4208864,"size":1015},"modified":{"name":"fcn.004038f0","addr":4208880,"size":1015}},{"similarity":0.932203,"type":"PARTIAL","original":{"name":"fcn.00403cf0","addr":4209904,"size":118},"modified":{"name":"fcn.00403d00","addr":4209920,"size":118}},{"similarity":0.949153,"type":"PARTIAL","original":{"name":"fcn.00403d70","addr":4210032,"size":59},"modified":{"name":"fcn.00403d80","addr":4210048,"size":59}},{"similarity":0.965116,"type":"PARTIAL","original":{"name":"fcn.00403db0","addr":4210096,"size":86},"modified":{"name":"fcn.00403dc0","addr":4210112,"size":86}},{"similarity":0.923077,"type":"PARTIAL","original":{"name":"fcn.00403e90","addr":4210320,"size":26},"modified":{"name":"fcn.00403ea0","addr":4210336,"size":26}}]
```
ret2libc: What if we pretty-print this JSON or we put something else here?
I think these long "blobs" are not good in tests as they become impossible to review so it's easier for wrong things to be introduced.

librz/main/rz-diff.c
```diff
@@ -1894,6 +1895,9 @@ static void diff_functions_result_as_json(RzAnalysisMatchResult *result, PJ *pj)
 	rz_list_foreach (result->matches, iter, pair) {
 		fcn_a = (RzAnalysisFunction *)pair->pair_a;
 		fcn_b = (RzAnalysisFunction *)pair->pair_b;
+		if (fcn_a->type != RZ_ANALYSIS_FCN_TYPE_FCN && fcn_a->type != RZ_ANALYSIS_FCN_TYPE_SYM) {
```
ret2libc: is this logic also in the other output formats?

The reason i usually prefer functions that do:
```
for x in l:
   [logic on x, filtering, whatever...]
   if output == json:
       print json format
   elif output == table:
       print table format
   [...]
```

instead of:
```
if output == json:
   for x in l:
       [some logic on x, filtering, whatever...]
       print json format
elif output == table:
   for x in l:
       [some logic on x, filtering, whatever... potentially different from the ones above...]
       print table format
[...]
```

I think the former is more robust as the logic is the same for all the output modes.

ret2libc (commented):
Didn't review completely yet.

librz/main/rz-diff.c
wargio: yes in all formats, but i disagree on this. i do not like seeing switches in loops.

wargio:
Just a small update, i have a threaded version of this ready, but kinda requires this to be merged first.
If you want i can push the new changes here directly and rename this PR to something else.

librz/main/rz-diff.c
ret2libc: You can setup a separate function then 😜

ret2libc:
Will this skip matching stuff if they have different names?

wargio:
> Will this skip matching stuff if they have different names?

Yes, but they could match later on in the second loop. in the threaded version i loop only once and is much more optimized.
i think i will just push it here directly

test/db/tools/rz_diff
wargio: the issue by doing that is that the file will become HUGE

XVilka (commented):
The code looks good to me. I am just not sure we should trust function names.

librz/analysis/similarity.c
```diff
@@ -328,7 +358,69 @@ RZ_API void rz_analysis_match_result_free(RZ_NULLABLE RzAnalysisMatchResult *res
  */
 RZ_API RZ_OWN RzAnalysisMatchResult *rz_analysis_match_basic_blocks(RZ_NONNULL RzAnalysis *analysis, RZ_NONNULL RzAnalysisFunction *fcn_a, RZ_NONNULL RzAnalysisFunction *fcn_b) {
 	rz_return_val_if_fail(analysis && fcn_a && fcn_b, NULL);
-	return analysis_match_result_new(analysis, analysis, fcn_a->bbs, fcn_b->bbs, (SignatureDataCb)signature_data_bb_new);
+	return analysis_match_result_new(analysis, analysis, fcn_a->bbs, fcn_b->bbs, (RzThreadFunction)analysis_match_basic_blocks, (AllocateBuffer)basic_block_data_new);
+}
+
+static bool function_cmp(RzAnalysisFunction *fcn_a, RzAnalysisFunction *fcn_b) {
```
XVilka: This is not a `function_cmp`, it's `function_name_cmp`

ret2libc (commented):
Code seems ok, but I'd like to test it a bit. Could you provide some instructions on how to test well the functionality?

wargio:
> Code seems ok, but I'd like to test it a bit. Could you provide some instructions on how to test well the functionality?

just `rz-diff -t functions bins/other/rz-diff/true bins/other/rz-diff/false` i have tested this with big libc files for calculating the time difference
also you can run (for diffing basic blocks) the graph one but you need to supply the function address `rz-diff -C -t graphs -0 0x401ae0 -1 0x401af0 bins/other/rz-diff/true bins/other/rz-diff/false`

ret2libc (approved):

PR #3133, opened 2022-10-26
replacing `"` with `'` in `ag* dot` functions 

### LINKED ISSUE #3132: Quotes escaping in `agf dot` string literals
### Work environment

| OS/arch/bits (mandatory)                             | Debian x64
| File format of the file you reverse (mandatory)      | any
| Architecture/bits of the file (mandatory)            | any
| `rizin -v` full output  | rizin 0.5.0 @ linux-x86-64 commit: 3e5217456daf186ca4b5afbf53b41d7c155c38f7

**agf dot**  just adds text from **agf**'s blocks in a **label** parameter of the corresponding digraph nodes. It also puts there disassembly annotations, e.g. string literals. But the label parameter itself is a string literal, therefore any double quotes there will break digraph syntax. So the fix should escape all possible quotes in the node's label.

### Expected behavior
```
digraph code {
...
[URL="0x79c", color="lightgray", label="0x79c\n; [0x7f4:4]=196
; data.000007f4
ldr r3, [str.lib_ld_linux.so.3]
; 0x86c
; \"/bin/ls\"
add r3, pc, r3
; const char *string
; 0x86c
; \"/bin/ls\"
mov r0, r3
; int system(const char *string)
bl sym.imp.system
"]
...
}
```

### Actual behavior
```
digraph code {
...
[URL="0x79c", color="lightgray", label="0x79c\n; [0x7f4:4]=196
; data.000007f4
ldr r3, [str.lib_ld_linux.so.3]
; 0x86c
; "/bin/ls"
add r3, pc, r3
; const char *string
; 0x86c
; "/bin/ls"
mov r0, r3
; int system(const char *string)
bl sym.imp.system
"]
...
}
```

```
rizin binary
aaaa; s main // main should be a function that uses string literals
agf dot > example.dot
xdot example.dot  // will throw а syntax error
```

ogre2007:

**Detailed description**

simple fix for functions that converts printable graph to digraph code. 
Changes are self-explanatory, issue is described in #3132 
https://graphviz.org/doc/info/lang.html -- specification, that doesnt permit raw unescaped double-quotes
...

**Test plan**

```
rizin bin
aaaa;s main
agf dot > graph
xdot graph
```

...

**Closing issues**

closes #3132 

...


wargio (changes requested):
please add a test.

librz/core/cagraph.c
```diff
@@ -112,6 +112,8 @@ RZ_IPI void rz_core_agraph_print_interactive(RzCore *core) {
 static void agraph_print_node_dot(RzANode *n, void *user) {
 	char *label = strdup(n->body);
 	// label = rz_str_replace (label, "\n", "\\l", 1);
+	label = rz_str_replace(label, "\"", "\\\"", 1);
```
ret2libc: Put it in the else branch

librz/util/graph_drawable.c
```diff
@@ -56,7 +56,9 @@ RZ_API RZ_OWN char *rz_graph_drawable_to_dot(RZ_NONNULL RzGraph /*<RzGraphNodeIn
 
 	rz_list_foreach (nodes, it, node) {
 		RzGraphNodeInfo *print_node = (RzGraphNodeInfo *)node->data;
-		const char *body = print_node->body;
+		char *body = print_node->body;
+		body = rz_str_replace(body, "\"", "\\\"", 1);
```
ret2libc: This as well

ret2libc (changes requested):
Thanks for your contribution! I left some comments.

ret2libc (approved):

ogre2007:
how to fix double-free there? asan test is failed

thestr4ng3r (changes requested):
I have not checked the asan issue yet, but you will also have to replace `\` by `\\`, otherwise a raw `\"` will be broken again.
Also, is there an official specification of which escape codes are used in the dot format? It will be important to document along with the change where the escaping scheme comes from exactly.

librz/util/graph_drawable.c
```diff
@@ -56,11 +56,13 @@ RZ_API RZ_OWN char *rz_graph_drawable_to_dot(RZ_NONNULL RzGraph /*<RzGraphNodeIn
...
+		char *body = print_node->body;
+
 		if (!body || !*body) {
 			rz_strbuf_appendf(&buf, "%d [URL=\"%s\", color=\"lightgray\", label=\"%s\"]\n",
 				node->idx, print_node->title, print_node->title);
 		} else {
+			body = rz_str_replace(body, "\"", "\\\"", 1);
 			rz_strbuf_appendf(&buf, "%d [URL=\"%s\", color=\"lightgray\", label=\"%s\\n%s\"]\n",
 				node->idx, print_node->title, print_node->title, body);
 		}
```
wargio: The double free issue is very simple:
`print_node->body` owns the buffer, then you replace a shorter string with a longer string but this makes the old pointer be freed and a new pointer is allocated with a longer buffer.
Since `print_node->body` is never updated, is then freed again by another method.
Essentially you are taking the ownership of something that you do not own.
I suggest to copy the string, call replace with the copy and after free the copy.
```c
char *body = strdup(print_node->body);
....
} else {
	body = rz_str_replace(body, "\"", "\\\"", 1);
	rz_strbuf_appendf(&buf, "%d [URL=\"%s\", color=\"lightgray\", label=\"%s\\n%s\"]\n",
				node->idx, print_node->title, print_node->title, body);
	free(body);
}
```

ogre2007:
> I have not checked the asan issue yet, but you will also have to replace `\` by `\\`, otherwise a raw `\"` will be broken again. Also, is there an official specification of which escape codes are used in the dot format? It will be important to document along with the change where the escaping scheme comes from exactly.

https://graphviz.org/doc/info/lang.html

I agree with you, but format doesn't allow `\\` either. Should I use some sort of regexp to find all `"` which are not prefixed with `\`, or just iterate over them?

wargio:
> > I have not checked the asan issue yet, but you will also have to replace `\` by `\\`, otherwise a raw `\"` will be broken again. Also, is there an official specification of which escape codes are used in the dot format? It will be important to document along with the change where the escaping scheme comes from exactly.
> 
> https://graphviz.org/doc/info/lang.html
> 
> I agree with you, but format doesn't allow `\\` either. Should I use some sort of regexp to find all `"` which are not prefixed with `\`, or just iterate over them?

the easiest way to escape that is to replace `"` with a `'`

XVilka (approved):

PR #3138, opened 2022-10-29
Rename only non-symbols, but add the FLIRT flag

wargio:

**Detailed description**

This PR ensures the FLIRT flags gets added only to non-symbols but also always adds flags.


librz/sign/flirt.c
```diff
@@ -356,6 +356,15 @@ static bool check_crc16(const RzFlirtModule *module, ut8 *b, ut32 b_size) {
 	return module->crc16 == flirt_crc16(b + RZ_FLIRT_MAX_PRELUDE_SIZE, module->crc_length);
 }
 
+static bool try_rename_function(RzAnalysis *analysis, RzAnalysisFunction *fcn, const char *name) {
+	if (fcn->type == RZ_ANALYSIS_FCN_TYPE_SYM) {
+		// do not rename if is a symbol but check if
+		// another function has the same name
```
XVilka, line 362: What this check is for? It's unclear to me why it should search for other functions with the same name.

XVilka (approved):
Code looks fine

librz/sign/flirt.c
wargio: because you could have some false positives which name is already being used but the function name must be unique, so this function is used to loop till a valid new name is found (i.e. unused)
In case of symbols we do not want to rename the function so we just check if the name is unique for the flags

XVilka: Ok, understood. Thanks.

ret2libc (changes requested):
I'm still reviewing and testing sorry

ret2libc (approved):

PR #3147, opened 2022-10-31
Fix gcc12-asan do_analysis_search -Wstringop-overread warning, take 2

kazarmy:

**Detailed description**

This pr redos the fix for the problem in #3137 so that it's more useful.

**Test plan**

All builds are green.

**Closing issues**

...

### DIFF
--- a/librz/core/cmd/cmd_search.c
+++ b/librz/core/cmd/cmd_search.c
@@ -1947,6 +1947,7 @@ static void do_ref_search(RzCore *core, ut64 addr, ut64 from, ut64 to, struct se
 }
 
 static bool do_analysis_search(RzCore *core, struct search_parameters *param, const char *input) {
+	rz_return_val_if_fail(input, false);
 	RzSearch *search = core->search;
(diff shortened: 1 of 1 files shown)


librz/core/cmd/cmd_search.c
```diff
@@ -2020,6 +2021,7 @@ static bool do_analysis_search(RzCore *core, struct search_parameters *param, co
 		pj_a(param->pj);
 	}
 	input = rz_str_trim_head_ro(input);
+	rz_return_val_if_fail(input, false);
```
wargio, line 2024: i think these should be only set for the input in a function. maybe is better to throw an error.

kazarmy: That line is there because otherwise gcc12-asan won't shut up, but yeah if `input` is NULL than `*input` should throw (technically [it's UB](https://en.cppreference.com/w/c/language/operator_member_access) but probably Rizin won't be run on any system where `*NULL` won't throw).

PR #3154, opened 2022-11-03
Fix pe string endianness

wargio:
# DO NOT SQUASH

**Detailed description**

Fixes the string endianness issue on PE files

### DIFF
--- a/librz/bin/format/pe/pe_rsrc.c
+++ b/librz/bin/format/pe/pe_rsrc.c
@@ -338,7 +338,7 @@ static String *Pe_r_bin_pe_parse_string(RzBinPEObj *bin, PE_DWord *curAddr) {
 		if (*curAddr > bin->size || *curAddr + sizeof(ut16) > bin->size) {
 			goto out_error;
 		}
-		if (!rz_buf_read_le16_at(bin->b, *curAddr, &utf16_char)) {
+		if (rz_buf_read_at(bin->b, *curAddr, (ut8 *)&utf16_char, sizeof(ut16)) != sizeof(ut16)) {
 			RZ_LOG_INFO("check (String szKey)\n");
 			goto out_error;
 		}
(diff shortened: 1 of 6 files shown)


librz/bin/p/bin_mdmp.c
```diff
@@ -57,7 +57,7 @@ static RzBinInfo *mdmp_info(RzBinFile *bf) {
 
 	obj = (MiniDmpObj *)bf->o->bin_obj;
 
-	ret->big_endian = obj->endian;
+	ret->big_endian = false;
```
ret2libc, line 60: why?

wargio: the obj->endian was never initialized. the only code that was initializing it was when `RZ_NEW0(MiniDmpObj)` was called

wargio: but then it never gets assigned in the whole code.

librz/bin/format/pe/pe_rsrc.c
```diff
@@ -338,7 +338,7 @@ static String *Pe_r_bin_pe_parse_string(RzBinPEObj *bin, PE_DWord *curAddr) {
 		if (*curAddr > bin->size || *curAddr + sizeof(ut16) > bin->size) {
 			goto out_error;
 		}
-		if (!rz_buf_read_le16_at(bin->b, *curAddr, &utf16_char)) {
+		if (rz_buf_read_at(bin->b, *curAddr, (ut8 *)&utf16_char, sizeof(ut16)) != sizeof(ut16)) {
```
XVilka, line 341: Note, that technically UTF16 could be also BE. We need to check if Windows PEs for big-endian platforms still used UTF16LE variant.

XVilka (changes requested):

librz/bin/format/pe/pe_rsrc.c
wargio: no, this is correct because the conversion (including endianness) is done by `rz_str_utf16_to_utf8`

PR #3168, opened 2022-11-07
Support for Xbox 360 PE32 architecture (PPC BE)

horsicq:
You can find a sample here: https://github.com/horsicq/Detect-It-Easy/issues/136


librz/bin/format/pe/pe_info.c
```diff
@@ -137,6 +138,7 @@ char *PE_(rz_bin_pe_get_arch)(RzBinPEObj *bin) {
 		break;
 	case PE_IMAGE_FILE_MACHINE_POWERPC:
 	case PE_IMAGE_FILE_MACHINE_POWERPCFP:
+	case PE_IMAGE_FILE_MACHINE_POWERPC_BE:
```
wargio: does BE stand for big endian? if yes, then should we expect different parsing for this variation?

horsicq: Yes, it's Big Endian. This should be noticed at least in the disassembler. (PPC Big Endian arch)

wargio: then maybe you want to set the `info->big_endian` boolean to true (unless this is already done by the code)

horsicq: I am not sure.

wargio:
i have just tested this, you indeed need to set the big endian flag.
please modify this line and add a check for that machine and return true 
https://github.com/rizinorg/rizin/blob/dev/librz/bin/format/pe/pe_info.c#L349

horsicq:
Thanks a lot for the testing!

XVilka:
@horsicq please fix the `clang-format` thing. One way to do that is to remove `_` in `POWERPC_BE`.

XVilka (approved):

PR #3213, opened 2022-11-28
Convert `pC` commands to rzshell

XVilka:

**Detailed description**

- Converted `pC` commands to rzshell
- Removed `pCc` command because there is no use for printing C-style array in columns

**Test plan**

CI is green

**Closing issues**

Partially addresses https://github.com/rizinorg/rizin/issues/1590

### DIFF
--- a/librz/core/cmd/cmd_print.c
+++ b/librz/core/cmd/cmd_print.c
@@ -727,53 +727,6 @@ static void cmd_prc(RzCore *core, const ut8 *block, int len) {
 	}
 }
 
-static void cmd_pCd(RzCore *core, const char *input) {
-	int h, w = rz_cons_get_size(&h);
-	int colwidth = rz_config_get_i(core->config, "hex.cols") * 2.5;
-	if (colwidth < 1) {
-		colwidth = 16;
-	}
-	int i, columns = w / colwidth;
-	int rows = h - 2;
-	int obsz = core->blocksize;
-	int user_rows = rz_num_math(core->num, input);
-	bool asm_minicols = rz_config_get_i(core->config, "asm.minicols");
-	char *o_ao = strdup(rz_config_get(core->config, "asm.offset"));
-	char *o_ab = strdup(rz_config_get(core->config, "asm.bytes"));
-	if (asm_minicols) {
-		rz_config_set(core->config, "asm.offset", "false");
-		// rz_config_set (core->config, "asm.bytes", "false");
-	}
-	rz_config_set(core->config, "asm.bytes", "false");
-	if (user_rows > 0) {
-		rows = user_rows + 1;
-	}
-	RzConsCanvas *c = rz_cons_canvas_new(w, rows);
-	ut64 osek = core->offset;
-	c->color = rz_config_get_i(core->config, "scr.color");
-	rz_core_block_size(core, rows * 32);
-	for (i = 0; i < columns; i++) {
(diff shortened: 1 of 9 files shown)


librz/core/cmd/cmd_print.c
```diff
@@ -6489,3 +6315,170 @@ RZ_IPI RzCmdStatus rz_print_minus_table_handler(RzCore *core, int argc, const ch
...
+		rz_config_set(core->config, "asm.offset", o_ao);
+		rz_config_set(core->config, "asm.bytes", o_ab);
+	}
+	rz_config_set(core->config, "asm.bytes", o_ab);
+	free(o_ao);
+	free(o_ab);
+	return RZ_CMD_STATUS_OK;
+}
+
+RZ_IPI RzCmdStatus rz_print_columns_Disassembly_handler(RzCore *core, int argc, const char **argv) {
```
ret2libc: pCD 20
ERROR: Failed to read registers.
ERROR: No such register or register type: ""
ERROR: core: Error while parsing command: `(null) 128@r:SP;pd@ 0x100003a3c`

ret2libc: also, this and the previous function are so similar that we should probably merge them IMHO.

ret2libc: and also the next one is somehow similar...

librz/core/cmd/cmd_print.c
```diff
@@ -6489,3 +6315,170 @@ RZ_IPI RzCmdStatus rz_print_minus_table_handler(RzCore *core, int argc, const ch
...
+	rz_config_set_i(core->config, "hex.cols", hex_cols);
+	return true;
+}
+
+RZ_IPI RzCmdStatus rz_print_columns_hex_annotated_handler(RzCore *core, int argc, const char **argv) {
+	int user_rows = argc > 1 ? rz_num_math(core->num, argv[1]) : -1;
+	return bool2status(print_hexdump_columns(core, user_rows, "pxa"));
+}
+
+RZ_IPI RzCmdStatus rz_print_columns_hex_op_colored_handler(RzCore *core, int argc, const char **argv) {
```
ret2libc, line 6467: pCA 10
ERROR: core: Invalid length
ERROR: core: Invalid length
ERROR: core: Invalid length

ret2libc (commented):
Any super simple test we can add for these commands? You can use scr.columns/scr.rows to force the size of the RzCons so it is reproducible.

librz/core/cmd/cmd_print.c
XVilka: This command works only in debug mode, I limited it to return the error message if not.

XVilka: They are not too similar - one returns disassembly, while other - debug context.

XVilka:
@ret2libc everything should be fixed now - the problem with `pCA` was that it was using `pxA` which changed behavior since - it limits itself to the block size in the short form (`pxA`) but works fine in the long form (`pxAl`).

ret2libc:
image

The problem is that the second columns starts at 0x100003a64, but the first one ends with 0x100003a5c. So there is an instruction (0x100003a60) that is not displayed.

However... this is not strictly a problem with your conversion, so if you want go ahead but let's create an issue for this at least.

XVilka:
OK: https://github.com/rizinorg/rizin/issues/3318

PR #3221, opened 2022-12-04
Add `*pty` API

### LINKED ISSUE #3212: Add support for `forkpty` in RzSubprocess
RzSubprocess API does not have a way to use `forkpty` or any of the `*pty` APIs at all, making it hard to use in scenarios where a new TTY/PTY may be needed.

DMaroo:

**Detailed description**

Add API to access `*pty` functions (`openpty`, `login_tty` and `forkpty`)

**Test plan**

Add unit tests, green CI

**Closing issues**

Closes #3212

### DIFF
--- a/librz/util/subprocess.c
+++ b/librz/util/subprocess.c
@@ -929,6 +929,13 @@ RZ_API RzSubprocess *rz_subprocess_start_opt(RzSubprocessOpt *opt) {
 	int stdin_pipe[2] = { -1, -1 };
 	int stdout_pipe[2] = { -1, -1 };
 	int stderr_pipe[2] = { -1, -1 };
+
+	/* There seems to be no reasonable use case for having pipes when forking
+	using forkpty, hence we will bypass all the pipe creation logic */
+	if (opt->fork_mode == RZ_SUBPROCESS_FORKPTY) {
+		goto no_pipes;
+	}
+
 	if (opt->stdin_pipe == RZ_SUBPROCESS_PIPE_CREATE) {
 		if (rz_sys_pipe(stdin_pipe, true) == -1) {
 			perror("pipe");
@@ -965,29 +972,23 @@ RZ_API RzSubprocess *rz_subprocess_start_opt(RzSubprocessOpt *opt) {
 		proc->stderr_fd = proc->stdout_fd;
 	}
 
+no_pipes:
 	// Let's create the environment for the child in the parent, with malloc,
 	// because we can't use functions that lock after fork
 	child_env = create_child_env(opt->envvars, opt->envvals, opt->env_size);
 
 	int master_fd = -1;
 	if (opt->fork_mode == RZ_SUBPROCESS_FORKPTY) {
-		if (opt->pty) {
-			/* fork normally for now, use login_tty later with opt->pty */
-			proc->pid = rz_sys_fork();
-			proc->master_fd = master_fd = opt->pty->master_fd;
-		} else {
-			void *term_params = NULL;
+		void *term_params = NULL;
 #if HAVE_FORKPTY && HAVE_OPENPTY && HAVE_LOGIN_TTY
-			struct termios t;
-			/* Needed to avoid reading back the writes again from the TTY */
-			tcgetattr(STDIN_FILENO, &t);
-			cfmakeraw(&t);
-			term_params = &t;
+		struct termios t;
+		/* Needed to avoid reading back the writes again from the TTY */
+		tcgetattr(STDIN_FILENO, &t);
+		cfmakeraw(&t);
+		term_params = &t;
 #endif
-			/* Fork to a new PTY if opt->pty is NULL */
-			proc->pid = rz_sys_forkpty(&master_fd, NULL, term_params, NULL);
-			proc->master_fd = master_fd;
-		}
+		proc->pid = rz_sys_forkpty(&master_fd, NULL, term_params, NULL);
+		proc->master_fd = master_fd;
 
 		proc->stderr_fd = (proc->stderr_fd == -1) ? master_fd : proc->stderr_fd;
 		proc->stdout_fd = (proc->stdout_fd == -1) ? master_fd : proc->stdout_fd;
@@ -1006,12 +1007,6 @@ RZ_API RzSubprocess *rz_subprocess_start_opt(RzSubprocessOpt *opt) {
 		goto error;
 	} else if (proc->pid == 0) {
 		// child
-		if (opt->fork_mode == RZ_SUBPROCESS_FORKPTY) {
-			if (opt->pty && !rz_subprocess_login_tty(opt->pty)) {
-				RZ_LOG_ERROR("login_tty failed, falling back to normal forking\n");
-			}
-		}
-
 		if (stderr_pipe[1] != -1) {
 			while ((dup2(stderr_pipe[1], STDERR_FILENO) == -1) && (errno == EINTR)) {
 			}
@@ -1405,8 +1400,7 @@ RZ_API RzSubprocess *rz_subprocess_start(
 		.stdin_pipe = RZ_SUBPROCESS_PIPE_CREATE,
 		.stdout_pipe = RZ_SUBPROCESS_PIPE_CREATE,
 		.stderr_pipe = RZ_SUBPROCESS_PIPE_CREATE,
-		.fork_mode = RZ_SUBPROCESS_FORK,
-		.pty = NULL
+		.fork_mode = RZ_SUBPROCESS_FORK
 	};
 	return rz_subprocess_start_opt(&opt);
 }
@@ -1451,11 +1445,11 @@ RZ_API bool rz_subprocess_login_tty(RZ_NONNULL RzPty *pty) {
  * envvars, so envvals[0] specifies the value of the environment variable named
  * envvars[0] and so on.
  * \param env_size Number of environment variables in arrays \p envvars and \p envvals
- * \param pty RzPty instance to use for the slave/child
- * pty == NULL implies a new PTY will be requested and used using openpty
+ *
+ * In forkpty mode, no pipes will be created, and std{in,out,err}_pipe values will be ignored
  */
 RZ_API RZ_OWN RzSubprocess *rz_subprocess_forkpty(const char *file, const char *args[], size_t args_size,
-	const char *envvars[], const char *envvals[], size_t env_size, RZ_NULLABLE RzPty *pty) {
+	const char *envvars[], const char *envvals[], size_t env_size) {
 	RzSubprocessOpt opt = {
 		.file = file,
 		.args = args,
@@ -1463,11 +1457,10 @@ RZ_API RZ_OWN RzSubprocess *rz_subprocess_forkpty(const char *file, const char *
 		.envvars = envvars,
 		.envvals = envvals,
 		.env_size = env_size,
-		.stdin_pipe = RZ_SUBPROCESS_PIPE_CREATE,
-		.stdout_pipe = RZ_SUBPROCESS_PIPE_CREATE,
-		.stderr_pipe = RZ_SUBPROCESS_PIPE_CREATE,
-		.fork_mode = RZ_SUBPROCESS_FORK,
-		.pty = pty
+		.stdin_pipe = RZ_SUBPROCESS_PIPE_NONE,
+		.stdout_pipe = RZ_SUBPROCESS_PIPE_NONE,
+		.stderr_pipe = RZ_SUBPROCESS_PIPE_NONE,
+		.fork_mode = RZ_SUBPROCESS_FORKPTY
 	};
 	return rz_subprocess_start_opt(&opt);
 }
--- a/librz/util/subprocess.c
+++ b/librz/util/subprocess.c
@@ -1039,24 +1039,6 @@ RZ_API RzSubprocess *rz_subprocess_start_opt(RZ_NONNULL const RzSubprocessOpt *o
 		goto error;
 	}
 
-#if HAVE_FORKPTY && HAVE_OPENPTY && HAVE_LOGIN_TTY
-	struct termios term_params;
-	/* Needed to avoid reading back the writes again from the TTY */
-	if (tcgetattr(STDIN_FILENO, &term_params) == -1) {
-		perror("tcgetattr");
-		goto no_term_change;
-	}
-	cfmakeraw(&term_params);
-
-	if (opt->make_raw && proc->slave_fd != -1) {
-		/* This avoids ECHO, so we don't read back whatever we wrote */
-		if (tcsetattr(proc->slave_fd, TCSANOW, &term_params) == -1) {
-			perror("tcsetattr");
-		}
-	}
-#endif
-
-no_term_change:
 	// Let's create the environment for the child in the parent, with malloc,
 	// because we can't use functions that lock after fork
 	child_env = create_child_env(opt->envvars, opt->envvals, opt->env_size);
@@ -1112,6 +1094,25 @@ no_term_change:
 	destroy_child_env(child_env);
 	free(argv);
 
+	if (!opt->make_raw || proc->slave_fd == -1) {
+		goto no_term_change;
+	}
+
+#if HAVE_FORKPTY && HAVE_OPENPTY && HAVE_LOGIN_TTY
+	struct termios term_params;
+	/* Needed to avoid reading back the writes again from the TTY */
+	if (tcgetattr(proc->slave_fd, &term_params) != 0) {
+		perror("tcgetattr");
+		goto no_term_change;
+	}
+	cfmakeraw(&term_params);
+	/* This avoids ECHO, so we don't read back whatever we wrote */
+	if (tcsetattr(proc->slave_fd, TCSANOW, &term_params) != 0) {
+		perror("tcsetattr");
+	}
+#endif
+
+no_term_change:
 	if (proc->slave_fd != -1 && close(proc->slave_fd) == -1) {
 		perror("close");
 	}
@@ -1572,7 +1573,7 @@ RZ_API RzSubprocess *rz_subprocess_start(
 		.stdout_pipe = RZ_SUBPROCESS_PIPE_CREATE,
 		.stderr_pipe = RZ_SUBPROCESS_PIPE_CREATE,
 		.pty = NULL,
-		.make_raw = /* does not matter */ true
+		.make_raw = /* does not matter */ false
 	};
 	return rz_subprocess_start_opt(&opt);
 }
--- a/librz/util/subprocess.c
+++ b/librz/util/subprocess.c
@@ -1,5 +1,6 @@
 // SPDX-FileCopyrightText: 2020 Florian Märkl <info@florianmaerkl.de>
 // SPDX-FileCopyrightText: 2021 ret2libc <sirmy15@gmail.com>
+// SPDX-FileCopyrightText: 2022 Dhruv Maroo <dhruvmaru007@gmail.com>
 // SPDX-License-Identifier: LGPL-3.0-only
 
 #include <rz_cons.h>
@@ -709,10 +710,6 @@ RZ_API bool rz_subprocess_login_tty(RZ_BORROW RZ_NONNULL const RzPty *pty) {
 	RZ_LOG_ERROR("login_tty: Not implemented for Windows!");
 }
 
-RZ_API RZ_OWN RzSubprocess *rz_subprocess_forkpty(const char *file, const char *args[], size_t args_size, const char *envvars[], const char *envvals[], size_t env_size, RZ_BORROW RZ_NULLABLE const RzPty *pty) {
-	RZ_LOG_ERROR("forkpty: Not implemented for Windows!");
-}
-
 #else // __WINDOWS__
 
 #include <errno.h>
@@ -727,7 +724,8 @@ struct rz_subprocess_t {
 	int ret;
 	RzStrBuf out;
 	RzStrBuf err;
-	int master_fd; ///< Needed to check whether PTY (user code should not access this directly)
+	int master_fd; ///< Needed to check whether PTY
+	int slave_fd;
 };
 
 static RzPVector subprocs;
@@ -904,16 +902,45 @@ static void destroy_child_env(char **child_env) {
 	free(child_env);
 }
 
-static bool init_pipes(RzSubprocess *proc, const RzSubprocessOpt *opt, int stdin_pipe[], int stdout_pipe[], int stderr_pipe[]) {
-	if (opt->stdin_pipe == RZ_SUBPROCESS_PIPE_CREATE) {
+static bool init_pipes(RzSubprocess *proc, const RzSubprocessOpt *opt, int stdin_pipe[], int stdout_pipe[], int stderr_pipe[], RzPty *new_pty) {
+	RzPty *pty = opt->pty;
+
+	/* If we need to use a PTY, we should create one right now */
+	if (!pty && (opt->stdin_pipe == RZ_SUBPROCESS_PIPE_PTY || opt->stdout_pipe == RZ_SUBPROCESS_PIPE_PTY || opt->stderr_pipe == RZ_SUBPROCESS_PIPE_PTY)) {
+		pty = rz_subprocess_openpty(NULL, NULL, NULL);
+		if (!pty) {
+			return false;
+		}
+		new_pty = pty;
+	}
+
+	if (pty) {
+		proc->master_fd = pty->master_fd;
+		proc->slave_fd = pty->slave_fd;
+	}
+
+	switch (opt->stdin_pipe) {
+	case RZ_SUBPROCESS_PIPE_CREATE:
 		if (rz_sys_pipe(stdin_pipe, true) == -1) {
 			perror("pipe");
 			return false;
 		}
 		proc->stdin_fd = stdin_pipe[1];
+		break;
+	case RZ_SUBPROCESS_PIPE_PTY:
+		stdin_pipe[0] = pty->slave_fd;
+		stdin_pipe[1] = pty->master_fd;
+		proc->stdin_fd = stdin_pipe[1];
+		break;
+	case RZ_SUBPROCESS_PIPE_STDOUT:
+		RZ_LOG_ERROR("Invalid pipe option 'RZ_SUBPROCESS_PIPE_STDOUT' for stdin. Ignoring...\n");
+		break;
+	case RZ_SUBPROCESS_PIPE_NONE:
+		break;
 	}
 
-	if (opt->stdout_pipe == RZ_SUBPROCESS_PIPE_CREATE) {
+	switch (opt->stdout_pipe) {
+	case RZ_SUBPROCESS_PIPE_CREATE:
 		if (rz_sys_pipe(stdout_pipe, true) == -1) {
 			perror("pipe");
 			return false;
@@ -923,9 +950,21 @@ static bool init_pipes(RzSubprocess *proc, const RzSubprocessOpt *opt, int stdin
 			return false;
 		}
 		proc->stdout_fd = stdout_pipe[0];
+		break;
+	case RZ_SUBPROCESS_PIPE_PTY:
+		stdout_pipe[0] = pty->master_fd;
+		stdout_pipe[1] = pty->slave_fd;
+		proc->stdout_fd = stdout_pipe[0];
+		break;
+	case RZ_SUBPROCESS_PIPE_STDOUT:
+		RZ_LOG_ERROR("Invalid pipe option 'RZ_SUBPROCESS_PIPE_STDOUT' for stdout. Ignoring...\n");
+		break;
+	case RZ_SUBPROCESS_PIPE_NONE:
+		break;
 	}
 
-	if (opt->stderr_pipe == RZ_SUBPROCESS_PIPE_CREATE) {
+	switch (opt->stderr_pipe) {
+	case RZ_SUBPROCESS_PIPE_CREATE:
 		if (rz_sys_pipe(stderr_pipe, true) == -1) {
 			perror("pipe");
 			return false;
@@ -935,10 +974,19 @@ static bool init_pipes(RzSubprocess *proc, const RzSubprocessOpt *opt, int stdin
 			return false;
 		}
 		proc->stderr_fd = stderr_pipe[0];
-	} else if (opt->stderr_pipe == RZ_SUBPROCESS_PIPE_STDOUT) {
+		break;
+	case RZ_SUBPROCESS_PIPE_PTY:
+		stdout_pipe[0] = pty->master_fd;
+		stdout_pipe[1] = pty->slave_fd;
+		proc->stdout_fd = stdout_pipe[0];
+		break;
+	case RZ_SUBPROCESS_PIPE_STDOUT:
 		stderr_pipe[0] = stdout_pipe[0];
 		stderr_pipe[1] = stdout_pipe[1];
 		proc->stderr_fd = proc->stdout_fd;
+		break;
+	case RZ_SUBPROCESS_PIPE_NONE:
+		break;
 	}
 
 	return true;
@@ -950,7 +998,7 @@ RZ_API RzSubprocess *rz_subprocess_start_opt(RZ_NONNULL const RzSubprocessOpt *o
 	RzSubprocess *proc = NULL;
 	char **child_env = NULL;
 	char **argv = calloc(opt->args_size + 2, sizeof(char *));
-	const RzPty *pty = NULL;
+	RzPty *new_pty = NULL;
 	if (!argv) {
 		return NULL;
 	}
@@ -970,6 +1018,7 @@ RZ_API RzSubprocess *rz_subprocess_start_opt(RZ_NONNULL const RzSubprocessOpt *o
 	proc->stdout_fd = -1;
 	proc->stderr_fd = -1;
 	proc->master_fd = -1;
+	proc->slave_fd = -1;
 	rz_strbuf_init(&proc->out);
 	rz_strbuf_init(&proc->err);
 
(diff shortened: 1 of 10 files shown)


DMaroo:
I need help in resolving the failing tests for MacOS and Windows. It would be really helpful if anyone could help in figuring out what exactly is the cause for the failing CI.

librz/util/subprocess.c
```diff
@@ -967,13 +969,54 @@ RZ_API RzSubprocess *rz_subprocess_start_opt(RzSubprocessOpt *opt) {
...
 
-	proc->pid = rz_sys_fork();
+	void *term_params = NULL;
+#if HAVE_FORKPTY && HAVE_OPENPTY && HAVE_LOGIN_TTY
+	struct termios t;
+	/* Needed to avoid reading back the writes again from the TTY */
+	tcgetattr(STDIN_FILENO, &t);
+	cfmakeraw(&t);
+	term_params = &t;
+#endif
```
ret2libc: Isn't this necessary only when RZ_SUBPROCESS_FORKPTY?

librz/util/subprocess.c
```diff
@@ -967,13 +969,54 @@ RZ_API RzSubprocess *rz_subprocess_start_opt(RzSubprocessOpt *opt) {
...
+			proc->master_fd = master_fd;
+		}
+	} else {
+		proc->pid = rz_sys_fork();
+	}
+
+	if (opt->fork_mode == RZ_SUBPROCESS_FORKPTY) {
+		proc->stderr_fd = (proc->stderr_fd == -1) ? master_fd : proc->stderr_fd;
+		proc->stdout_fd = (proc->stdout_fd == -1) ? master_fd : proc->stdout_fd;
+		proc->stdin_fd = (proc->stdin_fd == -1) ? master_fd : proc->stdin_fd;
```
ret2libc: you could just move these in the right branch of the `if` above

librz/util/subprocess.c
```diff
@@ -967,13 +969,54 @@ RZ_API RzSubprocess *rz_subprocess_start_opt(RzSubprocessOpt *opt) {
...
+		} else {
+			perror("fork");
+		}
 		goto error;
 	} else if (proc->pid == 0) {
 		// child
+		if (opt->fork_mode == RZ_SUBPROCESS_FORKPTY) {
+			if (opt->pty && !rz_subprocess_login_tty(opt->pty)) {
+				RZ_LOG_ERROR("login_tty failed, falling back to normal forking\n");
+				goto normal;
```
ret2libc: why the `goto`?

librz/util/subprocess.c
```diff
@@ -1096,6 +1141,13 @@ static RzSubprocessWaitReason subprocess_wait(RzSubprocess *proc, ut64 timeout_m
 	bool child_dead = false;
 	bool timedout = true;
 	bool bytes_enabled = n_bytes != 0;
+
+	/* Check if stdout and stderr are connected to a PTY
+	If yes, then set true if we n_bytes == 0,
+	i.e. waiing for subprocess to end */
+	bool stdout_pty = proc->stdout_fd == proc->master_fd && n_bytes == 0;
+	bool stderr_pty = proc->stderr_fd == proc->master_fd && n_bytes == 0;
```
ret2libc: Is `proc->master_fd` ever initialized when the subprocess is not in a new PTY?

ret2libc: Also, why are we interested in whether stdout/stederr are pty only when n_bytes == 0?

test/unit/test_subprocess.c
```diff
@@ -263,6 +263,120 @@ bool test_interactive(void) {
...
+bool test_interactive_pipe_pty(void) {
+	rz_subprocess_init();
+	const char *exe_path = get_auxiliary_path("subprocess-interactive");
+
+	RzSubprocessOpt opt = { 0 };
+	opt.file = exe_path;
+	opt.stdin_pipe = RZ_SUBPROCESS_PIPE_CREATE;
+	opt.stdout_pipe = RZ_SUBPROCESS_PIPE_CREATE;
+	opt.stderr_pipe = RZ_SUBPROCESS_PIPE_CREATE;
+	opt.fork_mode = RZ_SUBPROCESS_FORKPTY;
```
ret2libc: I don't remember anymore, but what is the diff between PIPE_NONE + FORKPTY and PIPE_CREATE + PTY? In other words, what is the diff between this test and test_interactive_pty?

ret2libc (commented):
Probably the failures are caused by changes in RzSubprocess?

librz/util/subprocess.c
DMaroo: Yes, indeed. I'll make the change.

librz/util/subprocess.c
DMaroo: Sorry, it is a remnant of some intermediate code I was trying. Removing it, thanks!

librz/util/subprocess.c
DMaroo: 1. `proc->master_fd` is set to -1 for when the subprocess is not in a new PTY.
2. Because `n_bytes == 0` only when we are waiting for the process to die (`n_bytes` are 0, when the caller is `rz_subprocess_wait`). I'll add a comment explaining this in the code.

DMaroo: This can be considered a sort of a hack-ish way to figure out whether we are waiting for the subprocess to die or not. The significance of `n_bytes` being zero is also stated in the function's (`subprocess_wait`) Doxygen:

```
 * ... If \p n_bytes is
 * not 0, the function terminates as soon as \p n_bytes bytes have been read or
 * the timeout expires.
```

test/unit/test_subprocess.c
DMaroo: In the test `test_interactive_pty`, the subprocess reads its `stdin` and writes its `stdout` to a single slave file descriptor, and the parent process reads the `stdout` and writes the `stdin` of the child to a single master file descriptor.

In the test `test_interactive_pipe_pty`, the subprocess reads its `stdin` and writes its `stdout` to a single slave file descriptor (this stays the same), however, there are two pipes, connecting the master file descriptor to two distinct file descriptors, one for the `stdin` and other for the `stdout`. Hence the parent process reads the `stdout` and writes the `stdin` of the child to two distinct file descriptors, both of which belong to pipes, whose other ends point to the master file descriptor.

DMaroo: But currently, `test_interactive_pipe_pty` passes but doesn't do what I specified above. It directly connects to the slave file descriptor. I'm working on changing this.

DMaroo: Upon further thinking, I can't seem to think of a use case where using pipes with forkpty would be any useful. I will just not use pipes if the `fork_mode` is set to `RZ_SUBPROCESS_FORKPTY`. This means that there is no need for the `test_interactive_pipe_pty` test. I'll remove that.

Please let me know if there are any issues with it. Thanks!

DMaroo:
Also, I don't think we can support invoking `rz_subprocess_start_opt` with an existing `RzPTY`, because the logic in `subprocess_wait` assumes contiguous file descriptors. To pass a PTY, we would need to create one before hand, henceforth breaking this assumption. Fixing this would require major changes in `subprocess_wait`, hence I am dropping the support for passing a `RzPTY`. I don't think there would be a scenario where it would be particularly useful.

ret2libc:
> Also, I don't think we can support invoking `rz_subprocess_start_opt` with an existing `RzPTY`, because the logic in `subprocess_wait` assumes contiguous file descriptors. To pass a PTY, we would need to create one before hand, henceforth breaking this assumption. Fixing this would require major changes in `subprocess_wait`, hence I am dropping the support for passing a `RzPTY`. I don't think there would be a scenario where it would be particularly useful.

Right now I don't remember exactly if that would be useful, but for ideas on how we plan to use RzSubprocess more look at run.c:handle_redirection_proc. A lot of the code in run.c should be replaced with the more uniform and standard RzSubprocess interface.

BTW, where does `subprocess_wait` assume that the files are contiguous?

librz/util/subprocess.c
```diff
@@ -927,6 +929,13 @@ RZ_API RzSubprocess *rz_subprocess_start_opt(RzSubprocessOpt *opt) {
 	int stdin_pipe[2] = { -1, -1 };
 	int stdout_pipe[2] = { -1, -1 };
 	int stderr_pipe[2] = { -1, -1 };
+
+	/* There seems to be no reasonable use case for having pipes when forking
+	using forkpty, hence we will bypass all the pipe creation logic */
+	if (opt->fork_mode == RZ_SUBPROCESS_FORKPTY) {
```
ret2libc: instead of `goto`, extract the pipe handling in a separate function, like

```
if (opt->for_mode != RZ_SUBPROCESS_FORKPTY && !start_handle_pipes(proc, opt)) {
   goto error;
}
```

librz/util/subprocess.c
```diff
@@ -963,14 +972,38 @@ RZ_API RzSubprocess *rz_subprocess_start_opt(RzSubprocessOpt *opt) {
...
+		tcgetattr(STDIN_FILENO, &t);
+		cfmakeraw(&t);
+		term_params = &t;
+#endif
+		proc->pid = rz_sys_forkpty(&master_fd, NULL, term_params, NULL);
+		proc->master_fd = master_fd;
+
+		proc->stderr_fd = (proc->stderr_fd == -1) ? master_fd : proc->stderr_fd;
+		proc->stdout_fd = (proc->stdout_fd == -1) ? master_fd : proc->stdout_fd;
+		proc->stdin_fd = (proc->stdin_fd == -1) ? master_fd : proc->stdin_fd;
```
ret2libc: I guess stderr_fd and the others are always -1 if we are in this branch. Maybe just put a `rz_warn_if_fail(proc->stderr_fd == 1);` etc. just to be sure, but it should never happen

librz/util/subprocess.c
ret2libc: Wouldn't this work well also without the `n_bytes == 0`? IN `read_to_strbuf` `read` would return EIO only when EOF.

test/unit/test_subprocess.c
```diff
@@ -263,6 +263,73 @@ bool test_interactive(void) {
...
+	/* This does NOT use pipes, this just checks whether forkpty ignores pipes */
+	rz_subprocess_init();
+	const char *exe_path = get_auxiliary_path("subprocess-interactive");
+
+	RzSubprocessOpt opt = { 0 };
+	opt.file = exe_path;
+	opt.stdin_pipe = RZ_SUBPROCESS_PIPE_NONE;
+	/* These won't matter since in forkpty no pipes are created */
+	opt.stdout_pipe = RZ_SUBPROCESS_PIPE_CREATE;
+	opt.stderr_pipe = RZ_SUBPROCESS_PIPE_CREATE;
```
ret2libc: I would prefer to have the API enforce that these are PIPE_NONE and fail if they aren't, otherwise it is not clear what a user can or cannot do.

librz/util/subprocess.c
DMaroo: Fair enough.

DMaroo:
I'm working on making `subprocess_wait` work with an existing `RzPTY` object.

DMaroo:
Now, we can use an existing `RzPTY` object as the master-slave pair for the PTY in `rz_subprocess_forkpty`. The test `test_interactive_custom_pty` in `test_subprocess.c` tests this.

ret2libc:
MacOS ARM does not fail for me locally, however I see it is failing in the CI. Also BSD :( Unsure what could be the problem though.

DMaroo:
The issue is that, in MacOS, the arguments `termp` and `winp` are not `const` qualified (https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/openpty.3.html), but in Linux they are (https://man7.org/linux/man-pages/man3/openpty.3.html). So passing the `const void *` gives a warning in MacOS, and since `-werror` is enabled, the build fails.

DMaroo:
Also, I'm more concerned why the Windows Appveyor job fails. It even fails to compile which seems very surprising. I hope it is not some MSVC specific thing (or worse, a MSVC bug).

DMaroo:
As for NetBSD, I will try researching it more. It should not really fail since Linux and OpenBSD have the same API for the PTY family of functions.

DMaroo:
Now the compilations finished successfully, but the `test_subprocess` unit test fails on MacOS. I have no idea why. I believe the reason may be similar to the one for NetBSD, but I don't know why.

@ret2libc It would be nice if you could also run the ninja tests on your MacOS machine. Thanks.

ret2libc:
> Now the compilations finished successfully, but the `test_subprocess` unit test fails on MacOS. I have no idea why. I believe the reason may be similar to the one for NetBSD, but I don't know why.
> 
> @ret2libc It would be nice if you could also run the ninja tests on your MacOS machine. Thanks.

I see MacOS working well now. Where is the failing job?

XVilka:
> Also, I'm more concerned why the Windows Appveyor job fails. It even fails to compile which seems very surprising. I hope it is not some MSVC specific thing (or worse, a MSVC bug).

It has nothing to do with MSVC, `clang-cl` fails too:
```
FAILED: librz/util/rz_util-0.5.dll.p/alloc.c.obj 
"clang-cl" "-Ilibrz\util\rz_util-0.5.dll.p" "-I." "-I.." "-Ilibrz" "-I..\librz" "-Ilibrz\include" "-I..\librz\include" "-Ilibrz\util\sdb\src" "-I..\librz\util\sdb\src" "-Ilibrz\util\sdb" "-I..\librz\util\sdb" "-Isubprojects\zlib-1.2.12" "-I..\subprojects\zlib-1.2.12" "-I..\subprojects\xz-5.2.5\src\liblzma\api" "/MD" "/nologo" "/showIncludes" "/utf-8" "/W2" "/O2" "-DPSAPI_VERSION=1" "-fcommon" "-D__STDC__=1" "-D_CRT_DECLARE_NONSTDC_NAMES " "-D_CRT_SECURE_NO_WARNINGS" "-D_CRT_NONSTDC_NO_DEPRECATE" "-DRZ_PLUGIN_INCORE=1" "-Werror=sizeof-pointer-memaccess" "-DLZMA_API_STATIC" "/Fdlibrz\util\rz_util-0.5.dll.p\alloc.c.pdb" /Folibrz/util/rz_util-0.5.dll.p/alloc.c.obj "/c" ../librz/util/alloc.c
In file included from ../librz/util/alloc.c:4:
In file included from ..\librz\include/rz_util.h:62:
..\librz\include/rz_util/rz_sys.h(47,8): error: unknown type name 'pid_t'
RZ_API pid_t rz_sys_forkpty(int *amaster, char *name, void /* const struct termios */ *termp, void /* const struct winsize */ *winp);
       ^
1 error generated.
```

DMaroo:
> It has nothing to do with MSVC, `clang-cl` fails too:
> 
> ```
> FAILED: librz/util/rz_util-0.5.dll.p/alloc.c.obj 
> "clang-cl" "-Ilibrz\util\rz_util-0.5.dll.p" "-I." "-I.." "-Ilibrz" "-I..\librz" "-Ilibrz\include" "-I..\librz\include" "-Ilibrz\util\sdb\src" "-I..\librz\util\sdb\src" "-Ilibrz\util\sdb" "-I..\librz\util\sdb" "-Isubprojects\zlib-1.2.12" "-I..\subprojects\zlib-1.2.12" "-I..\subprojects\xz-5.2.5\src\liblzma\api" "/MD" "/nologo" "/showIncludes" "/utf-8" "/W2" "/O2" "-DPSAPI_VERSION=1" "-fcommon" "-D__STDC__=1" "-D_CRT_DECLARE_NONSTDC_NAMES " "-D_CRT_SECURE_NO_WARNINGS" "-D_CRT_NONSTDC_NO_DEPRECATE" "-DRZ_PLUGIN_INCORE=1" "-Werror=sizeof-pointer-memaccess" "-DLZMA_API_STATIC" "/Fdlibrz\util\rz_util-0.5.dll.p\alloc.c.pdb" /Folibrz/util/rz_util-0.5.dll.p/alloc.c.obj "/c" ../librz/util/alloc.c
> In file included from ../librz/util/alloc.c:4:
> In file included from ..\librz\include/rz_util.h:62:
> ..\librz\include/rz_util/rz_sys.h(47,8): error: unknown type name 'pid_t'
> RZ_API pid_t rz_sys_forkpty(int *amaster, char *name, void /* const struct termios */ *termp, void /* const struct winsize */ *winp);
>        ^
> 1 error generated.
> ```

~~I have added a preprocessor define for this in `rz_sys.h`. The Windows CI should now not fail because of this.~~

I have changed the function signature for `rz_sys_forkpty` to return an `int` instead of `pid_t`. This will (hopefully) allow compilation on Windows.

XVilka:
Windows still has a few compilation problems with `clang-cl`:
```
"clang-cl" "-Ilibrz\util\rz_util-0.5.dll.p" "-I." "-I.." "-Ilibrz" "-I..\librz" "-Ilibrz\include" "-I..\librz\include" "-Ilibrz\util\sdb\src" "-I..\librz\util\sdb\src" "-Ilibrz\util\sdb" "-I..\librz\util\sdb" "-Isubprojects\zlib-1.2.12" "-I..\subprojects\zlib-1.2.12" "-I..\subprojects\xz-5.2.5\src\liblzma\api" "/MD" "/nologo" "/showIncludes" "/utf-8" "/W2" "/O2" "-DPSAPI_VERSION=1" "-fcommon" "-D__STDC__=1" "-D_CRT_DECLARE_NONSTDC_NAMES " "-D_CRT_SECURE_NO_WARNINGS" "-D_CRT_NONSTDC_NO_DEPRECATE" "-DRZ_PLUGIN_INCORE=1" "-Werror=sizeof-pointer-memaccess" "-DLZMA_API_STATIC" "/Fdlibrz\util\rz_util-0.5.dll.p\subprocess.c.pdb" /Folibrz/util/rz_util-0.5.dll.p/subprocess.c.obj "/c" ../librz/util/subprocess.c
../librz/util/subprocess.c(81,44): warning: incompatible pointer types passing 'RzThreadLock **' (aka 'struct rz_th_lock_t **') to parameter of type 'void *volatile *' [-Wincompatible-pointer-types]
                lock = InterlockedCompareExchangePointer(&subproc_mutex, INVALID_POINTER_VALUE, INVALID_POINTER_VALUE);
                                                         ^~~~~~~~~~~~~~
C:\Program Files (x86)\Windows Kits\10\include\10.0.22000.0\um\winnt.h(3219,43): note: passing argument to parameter 'Destination' here
    _Interlocked_operand_ PVOID volatile *Destination,
                                          ^
../librz/util/subprocess.c(92,31): warning: incompatible pointer types passing 'RzThreadLock **' (aka 'struct rz_th_lock_t **') to parameter of type 'void *volatile *' [-Wincompatible-pointer-types]
                        InterlockedExchangePointer(&subproc_mutex, INVALID_POINTER_VALUE);
                                                   ^~~~~~~~~~~~~~
C:\Program Files (x86)\Windows Kits\10\include\10.0.22000.0\um\winnt.h(3229,43): note: passing argument to parameter 'Target' here
    _Interlocked_operand_ PVOID volatile *Target,
                                          ^
../librz/util/subprocess.c(98,30): warning: incompatible pointer types passing 'RzThreadLock **' (aka 'struct rz_th_lock_t **') to parameter of type 'void *volatile *' [-Wincompatible-pointer-types]
                InterlockedExchangePointer(&subproc_mutex, lock);
                                           ^~~~~~~~~~~~~~
C:\Program Files (x86)\Windows Kits\10\include\10.0.22000.0\um\winnt.h(3229,43): note: passing argument to parameter 'Target' here
    _Interlocked_operand_ PVOID volatile *Target,
                                          ^
../librz/util/subprocess.c(147,37): warning: incompatible pointer types passing 'RzThreadLock **' (aka 'struct rz_th_lock_t **') to parameter of type 'void *volatile *' [-Wincompatible-pointer-types]
                lock = InterlockedExchangePointer(&subproc_mutex, NULL);
                                                  ^~~~~~~~~~~~~~
C:\Program Files (x86)\Windows Kits\10\include\10.0.22000.0\um\winnt.h(3229,43): note: passing argument to parameter 'Target' here
    _Interlocked_operand_ PVOID volatile *Target,
                                          ^
../librz/util/subprocess.c(150,30): warning: incompatible pointer types passing 'RzThreadLock **' (aka 'struct rz_th_lock_t **') to parameter of type 'void *volatile *' [-Wincompatible-pointer-types]
                InterlockedExchangePointer(&subproc_mutex, lock);
                                           ^~~~~~~~~~~~~~
C:\Program Files (x86)\Windows Kits\10\include\10.0.22000.0\um\winnt.h(3229,43): note: passing argument to parameter 'Target' here
    _Interlocked_operand_ PVOID volatile *Target,
                                          ^
../librz/util/subprocess.c(249,22): error: conflicting types for 'rz_subprocess_start_opt'
RZ_API RzSubprocess *rz_subprocess_start_opt(RzSubprocessOpt *opt) {
                     ^
..\librz\include/rz_util/rz_subprocess.h(102,22): note: previous declaration is here
RZ_API RzSubprocess *rz_subprocess_start_opt(RZ_NONNULL const RzSubprocessOpt *opt);
                     ^
5 warnings and 1 error generated.
[376/1967] Compiling C object librz/util/rz_util-0.5.dll.p/sys.c.obj
../librz/util/sys.c(302,9): warning: TODO: rz_sys_bt : unimplemented [-W#pragma-messages]
#pragma message("TODO: rz_sys_bt : unimplemented")
        ^
../librz/util/sys.c(487,9): warning: rz_sys_crash_handler : unimplemented for this platform [-W#pragma-messages]
#pragma message("rz_sys_crash_handler : unimplemented for this platform")
        ^
../librz/util/sys.c(673,9): warning: rz_sys_cmdbg is not implemented for this platform [-W#pragma-messages]
#pragma message("rz_sys_cmdbg is not implemented for this platform")
        ^
../librz/util/sys.c(1150,12): warning: unused variable 'val_utf16' [-Wunused-variable]
                wchar_t *val_utf16 = NULL;
                         ^
4 warnings generated.
```
And with Visual Studio 2022 MSVC:
```
"cl" "-Ilibrz\socket\rz_socket.lib.p" "-I." "-I.." "-Ilibrz" "-I..\librz" "-Ilibrz\include" "-I..\librz\include" "-Ilibrz\util\sdb\src" "-I..\librz\util\sdb\src" "-Ilibrz\util\sdb" "-I..\librz\util\sdb" "/MT" "/nologo" "/showIncludes" "/utf-8" "/W2" "/O2" "/Gw" "-DPSAPI_VERSION=1" "-DRZ_PLUGIN_INCORE=1" "/Fdlibrz\socket\rz_socket.lib.p\run.c.pdb" /Folibrz/socket/rz_socket.lib.p/run.c.obj "/c" ../librz/socket/run.c
TODO: handle_redirection_proc: Not implemented for this platform
string redirection handle not yet done
../librz/socket/run.c(791): warning C4311: 'type cast': pointer truncation from 'void *' to 'int'
../librz/socket/run.c(958): error C2065: 'pid_t': undeclared identifier
../librz/socket/run.c(958): error C2146: syntax error: missing ';' before identifier 'child_pid'
../librz/socket/run.c(958): error C2065: 'child_pid': undeclared identifier
../librz/socket/run.c(959): error C2065: 'child_pid': undeclared identifier
../librz/socket/run.c(964): error C2065: 'child_pid': undeclared identifier
../librz/socket/run.c(1139): warning C4312: 'type cast': conversion from 'int' to 'void *' of greater size
```

DMaroo:
I don't know why the `macos-meson-clang-tests` CI is failing. Can someone run and check it locally?

DMaroo:
Something weird is happening on MacOS platforms. The tests seem to pass sometimes, and fail sometimes. I don't know why there's this sort of inconsistency.

ret2libc:
I'm going to try it later tonight! I'll test with ASAN/UBSAN as well.

ret2libc:
@DMaroo I did this:
```
diff --git a/test/unit/test_subprocess.c b/test/unit/test_subprocess.c
index 1c2409314b..0469a2fda7 100644
--- a/test/unit/test_subprocess.c
+++ b/test/unit/test_subprocess.c
@@ -275,6 +275,7 @@ bool test_interactive_pty(void) {

        RzStrBuf *sb = rz_subprocess_stdout_readline(sp, UT_TIMEOUT);
        int c = atoi(rz_strbuf_get(sb));
+       eprintf("c = %d\n", c);
        char buf[100];
        snprintf(buf, sizeof(buf), "%d\n", 3 + 5 + c);
        rz_subprocess_stdin_write(sp, (const ut8 *)buf, strlen(buf));
@@ -334,17 +335,17 @@ bool test_interactive_custom_pty(void) {
 #endif // PTY functions
```
and also
```
@@ -1141,6 +1136,7 @@ static size_t read_to_strbuf(RzStrBuf *sb, int fd, bool *fd_eof, size_t n_bytes,
                to_read = n_bytes;
        }
        ssize_t sz = read(fd, buf, to_read);
+       printf("buf = %*s\n", sz, buf);
        if (sz == 0 || (is_pty && sz == -1 && errno == EIO)) {
                /* In case of PTY, EIO (input/output error) denotes EOF, hence the manual checking */
                *fd_eof = true;
```
and you can see that `c` is not correctly computed and that buf, at least in my case, has some things from stderr. Indeed if i execute the subprocess-interactive program I get the first line `subprocess-interactive(25345,0x10d292600) malloc: nano zone abandoned due to inability to preallocate reserved vm space.`

I *believe* (but you are now probably more into this than I am right now, so I may miss some important details) we should not use the `forkpty` but `openpty` + handling the master_fd so that we can split it into stdout/stderr/stdin according to `opt->std*_pipe` values and then do a regular `fork` (no `login_tty`, which seems to be the one merging in/out/err into one thing). Does this make sense?

DMaroo:
> I _believe_ (but you are now probably more into this than I am right now, so I may miss some important details) we should not use the `forkpty` but `openpty` + handling the master_fd so that we can split it into stdout/stderr/stdin according to `opt->std*_pipe` values and then do a regular `fork` (no `login_tty`, which seems to be the one merging in/out/err into one thing). Does this make sense?

Oh, so we dup the stream to `master_fd` and `slave_fd` (for the parent and child, respectively) if the pipe option for that stream has been set to `RZ_SUBPROCESS_PIPE_CREATE`. Is that what you meant? @ret2libc

ret2libc:
> > I _believe_ (but you are now probably more into this than I am right now, so I may miss some important details) we should not use the `forkpty` but `openpty` + handling the master_fd so that we can split it into stdout/stderr/stdin according to `opt->std*_pipe` values and then do a regular `fork` (no `login_tty`, which seems to be the one merging in/out/err into one thing). Does this make sense?
> 
> Oh, so we dup the stream to `master_fd` and `slave_fd` (for the parent and child, respectively) if the pipe option for that stream has been set to `RZ_SUBPROCESS_PIPE_CREATE`. Is that what you meant? @ret2libc

Not sure about `slave_fd`, but I think `master_fd` should be duplicated to be stdout/stderr/stdin only according to the `opt->std*_pipe` options. If `opt->stderr_pipe` is PIPE_NONE, you don't make it make `master_fd` the duplicate of `2`. If `opt->stdout_pipe` is PIPE_CREATE you duplicate `master_fd` to be `1`. I think this helps.

DMaroo:
> Not sure about slave_fd, but I think master_fd should be duplicated to be stdout/stderr/stdin only according to the opt->std*_pipe options. If opt->stderr_pipe is PIPE_NONE, you don't make it make master_fd the duplicate of 2. If opt->stdout_pipe is PIPE_CREATE you duplicate master_fd to be 1. I think this helps.

Understood. I have first obtained a pair of master-slave file descriptors through `openpty` (in case no `RzPTY` argument passed was `NULL`). I then only assign the `master_fd` to the streams which have `RZ_SUBPROCESS_PIPE_CREATE` as their pipe mode. This way, there is no way for the user (using the API) to access any of the streams blocked out, unless they go and manually access the `master_fd`. Then I fork normally, and use `rz_login_tty` for the child process.

ret2libc:
@DMaroo I think you broke something with the last changes

DMaroo:
~~Many of the tests are failing because of completely unrelated reasons. I have no idea why.~~

Alright, my bad. Changes have broken `rz-test` (since it uses the subprocess API). I'm looking into it.

Update: Should work now!

DMaroo:
Only the OpenBSD job is failing now. I don't know how to reproduce the failure to fix it.

librz/util/sys.c
```diff
@@ -603,6 +615,8 @@ RZ_API int rz_sys_cmd_str_full(const char *cmd, const char *input, char **output
 		.stdin_pipe = input ? RZ_SUBPROCESS_PIPE_CREATE : RZ_SUBPROCESS_PIPE_NONE,
 		.stdout_pipe = output ? RZ_SUBPROCESS_PIPE_CREATE : RZ_SUBPROCESS_PIPE_NONE,
 		.stderr_pipe = sterr ? RZ_SUBPROCESS_PIPE_CREATE : RZ_SUBPROCESS_PIPE_NONE,
+		.fork_mode = RZ_SUBPROCESS_FORK,
```
ret2libc: I see there are many of these cases. Now, or in the future, it might make sense to create a `rz_subprocess_opt_init` to avoid this and prevent errors from uninitialized fields.

ret2libc: If not now, please create an issue for this.

librz/util/subprocess.c
```diff
@@ -927,53 +985,67 @@ RZ_API RzSubprocess *rz_subprocess_start_opt(RzSubprocessOpt *opt) {
...
+		} else {
+			pty = rz_subprocess_openpty(NULL, term_params, NULL);
 		}
-		if (fcntl(stdout_pipe[0], F_SETFL, O_NONBLOCK) < 0) {
-			perror("fcntl");
+
+		if (pty) {
+			proc->master_fd = master_fd = pty->master_fd;
+			if (tcsetattr(pty->slave_fd, TCSANOW, term_params) == -1) {
+				perror("tcsetattr");
```
ret2libc: what is this? A comment like what's done above would help understand.

librz/util/subprocess.c
```diff
@@ -995,6 +1068,16 @@ RZ_API RzSubprocess *rz_subprocess_start_opt(RzSubprocessOpt *opt) {
...
 		}
 
+		if (opt->fork_mode == RZ_SUBPROCESS_FORKPTY) {
+			if (close(master_fd) == -1) {
+				perror("close");
+			}
+			if (rz_sys_login_tty(pty->slave_fd) == -1) {
+				perror("login_tty");
+				rz_sys_exit(-1, true);
+			}
```
ret2libc: Is this really needed? If yes, why? Maybe a comment.

ret2libc:
https://docs.pwntools.com/en/stable/tubes/processes.html is very interesting. I think what I initially wanted to do (loooong time ago) was to mimick something like this. Maybe we should take a similar approach and use `RzSubprocessPipeCreate` to specify whether a PTY is requested or not for stdin/stdout/stderr.

ret2libc:
Also, at least as a local test (not to push) it would be nice to run `python` with the new settings and see whether in one case it prints `>>> xxx` an in the other it doesn't. `python` will behave differently based on whether it's connected to PTY or not.

Another example would be to write an auxiliary program that uses libc stdout buffering and see if the new settings we can circument this (this was essentially the very main reason to do this).

librz/util/sys.c
DMaroo: To be fair, `RZ_SUBPROCESS_FORK` has the value 0. So even if we don't initialize the `fork_mode` and `pty` field explicitly, we would still get `fork_mode = RZ_SUBPROCESS_FORK` and `pty = NULL`. I just explicitly put them everywhere for the sake of it.

DMaroo: Creating an `rz_subprocess_opt_init` does not make sense to me because all it would do is just take every argument and add it to a `RzSubprocessOpt` struct. I would rather have struct initialization than a 9 argument function (which would be very ugly)>

librz/util/subprocess.c
DMaroo: Ah, sure. I'll add a comment.

librz/util/sys.c
ret2libc: I didn't mean to put all fields as args. What we have done for similar cases is to put common options as arguments, and then the caller will setup the extra/special fields. By doing so you can ensure all fields are initialized correctly.

librz/util/subprocess.c
DMaroo: pty

As labelled in the above chart, `login_tty` is responsible for connecting the child process' file descriptor to the slave file descriptor.

ret2libc: Awesome 👍

DMaroo:
@ret2libc

After going through `pwntools`'s source, I agree with your suggestion. It seems much more clearer. Thanks!

DMaroo:
Also, `pwntools`'s API has an argument `raw`, which is used to specify whether the opened PTY should be in raw mode or not. It is `True` by default, and I also directly make the PTY raw in case of `RzSubprocess`. Should I also add such an argument for our API?

ret2libc:
I think we can skip it for now, at least in this PR.

DMaroo:
> I think we can skip it for now, at least in this PR.

Oh, I'm sorry. I didn't read this, but I have added a `make_raw` option. It's only a single if-condition.

DMaroo:
Again, the  test results are quite unexpected. MacOS arm64 passes, but MacOS clang tests doesn't (lol).

librz/util/subprocess.c
```diff
@@ -890,10 +902,103 @@ static void destroy_child_env(char **child_env) {
...
+		stdin_pipe[0] = pty->slave_fd;
+		stdin_pipe[1] = pty->master_fd;
+		proc->stdin_fd = stdin_pipe[1];
+		break;
+	case RZ_SUBPROCESS_PIPE_STDOUT:
+		RZ_LOG_ERROR("Invalid pipe option 'RZ_SUBPROCESS_PIPE_STDOUT' for stdin. Ignoring...\n");
+		break;
+	case RZ_SUBPROCESS_PIPE_NONE:
+		break;
+	}
```
ret2libc, line 958: extract in a function

DMaroo: I think is better the way it is right now. Extracting it to a function makes the code very convoluted and contrived since in every case, we are doing something different. Also, the reduction in code size, if we use functions, is not significant.

librz/util/subprocess.c
```diff
@@ -246,7 +247,7 @@ static LPWCH override_env(const char *envvars[], const char *envvals[], size_t e
 	return ret;
 }
 
-RZ_API RzSubprocess *rz_subprocess_start_opt(RzSubprocessOpt *opt) {
+RZ_API RzSubprocess *rz_subprocess_start_opt(RZ_NONNULL const RzSubprocessOpt *opt) {
```
XVilka: Please add Doxygen

librz/util/subprocess.c
```diff
@@ -700,6 +701,15 @@ RZ_API void rz_subprocess_free(RzSubprocess *proc) {
 	CloseHandle(proc->proc);
 	free(proc);
 }
+
+RZ_API RZ_OWN RzPty *rz_subprocess_openpty(RZ_BORROW RZ_NULLABLE char *slave_name, RZ_NULLABLE void /* const struct termios */ *term_params, RZ_NULLABLE void /* const struct winsize */ *win_params) {
```
XVilka, line 716: Doxygen

librz/util/subprocess.c
```diff
@@ -714,6 +724,8 @@ struct rz_subprocess_t {
 	int ret;
 	RzStrBuf out;
 	RzStrBuf err;
+	int master_fd; ///< Needed to check whether PTY
```
XVilka, line 745: Maybe better to rename to something more neutral?

librz/util/subprocess.c
```diff
@@ -890,10 +902,103 @@ static void destroy_child_env(char **child_env) {
...
+		proc->stderr_fd = proc->stdout_fd;
+		break;
+	case RZ_SUBPROCESS_PIPE_NONE:
+		break;
+	}
+
+	return true;
+}
+
+RZ_API RzSubprocess *rz_subprocess_start_opt(RZ_NONNULL const RzSubprocessOpt *opt) {
```
XVilka: Doxygen and ownership mark of the return

librz/util/subprocess.c
```diff
@@ -1281,17 +1417,98 @@ RZ_API void rz_subprocess_free(RzSubprocess *proc) {
...
+	if (proc->stderr_fd != -1 && proc->stderr_fd != proc->stdout_fd && proc->stderr_fd != proc->master_fd) {
 		rz_sys_pipe_close(proc->stderr_fd);
 	}
 	free(proc);
 }
+
+/**
+ * \brief Call openpty(3) with the provided arguments
+ *
+ * \param slave_name The name of the slave PTY is stored in this
```
XVilka, line 1463: Please rename the parameter

librz/util/subprocess.c
```diff
@@ -1281,17 +1417,98 @@ RZ_API void rz_subprocess_free(RzSubprocess *proc) {
...
+RZ_API void rz_subprocess_close_pty(RZ_BORROW RZ_NONNULL const RzPty *pty) {
+	if (close(pty->master_fd) == -1) {
+		perror("close");
+	}
+	if (close(pty->slave_fd) == -1) {
+		perror("close");
+	}
+}
+
+/**
```
XVilka, line 1521: This Doxygen is unnecessary and can be removed

librz/util/sys.c
```diff
@@ -74,6 +74,18 @@ extern char **environ;
 #endif
 #endif
 
+/* For "openpty" family of funtcions */
+#if HAVE_OPENPTY && HAVE_FORKPTY && HAVE_LOGIN_TTY
+#if defined(__APPLE__) || defined(__NetBSD__) || defined(__OpenBSD__)
```
XVilka, line 79: Could it be converted to use just Meson? cc @ret2libc

librz/util/sys.c
```diff
@@ -1681,6 +1696,45 @@ RZ_API int rz_sys_fork(void) {
 }
 #endif
 
+RZ_API /* pid_t */ int rz_sys_forkpty(int *amaster, char *name, void /* const struct termios */ *termp, void /* const struct winsize */ *winp) {
```
XVilka, line 1709: Doxygen

librz/util/sys.c
```diff
@@ -1681,6 +1696,45 @@ RZ_API int rz_sys_fork(void) {
...
+		perror("forkpty");
+	}
+	return ret;
+#else
+	RZ_LOG_ERROR("forkpty() not found\n");
+	return -1;
+#endif
+}
+
+RZ_API int rz_sys_openpty(int *amaster, int *aslave, char *name, void /* const struct termios */ *termp, void /* const struct winsize */ *winp) {
```
XVilka, line 1733: Doxygen

librz/util/sys.c
```diff
@@ -1681,6 +1696,45 @@ RZ_API int rz_sys_fork(void) {
...
+		perror("openpty");
+	}
+	return ret;
+#else
+	RZ_LOG_ERROR("openpty() not found\n");
+	return -1;
+#endif
+}
+
+RZ_API int rz_sys_login_tty(int fd) {
```
XVilka, line 1752: Doxygen

XVilka (changes requested):

librz/util/subprocess.c
DMaroo: I agree about it not being neutral, but it is the standard term used for it. It will be much easier to understand if I also use the same term that the POSIX documentation uses.

librz/util/subprocess.c
DMaroo: Agreed, but I'll let it stay.

XVilka (approved):

PR #3237, opened 2022-12-09
Rewrite and extend type paths from offset

thestr4ng3r:
# DO NOT SQUASH
 

**Detailed description**

Tests and extends deriving paths into types at a given offset in the
following ways:
- Parameter for limiting the depth of the path
- Array support, e.g. `.member[5].anothermember`
- Typedef support
- Paths can be taken from arbitrary RzTypes, not only RzBaseType
- Removed the typename in the beginning of paths, which can be added by
  the caller if needed

see also the description of the first commit

**Test plan**

see unit tests

### DIFF
--- a/librz/type/path.c
+++ b/librz/type/path.c
@@ -122,91 +122,140 @@ RZ_API st64 rz_type_offset_by_path(const RzTypeDB *typedb, RZ_NONNULL const char
 	return path_walker(typedb, path);
 }
 
-// TODO: Handle arrays
-static bool structured_member_walker(const RzTypeDB *typedb, RzList /*<RzTypePath *>*/ *list, RzType *parent, RzType *type, char *path, ut64 offset) {
-	rz_return_val_if_fail(list && type, false);
-	if (type->kind != RZ_TYPE_KIND_IDENTIFIER) {
-		return false;
+static void collect_type_paths(const RzTypeDB *typedb, RzList /*<RzTypePath *>*/ *list, const RzType *type, char *prefix, ut64 offset, unsigned int depth);
+
+static void collect_base_type_paths(const RzTypeDB *typedb, RzList /*<RzTypePath *>*/ *list, const RzBaseType *btype, char *prefix, ut64 offset, unsigned int depth) {
+	rz_return_if_fail(typedb && list && btype && prefix);
+	if (!depth) {
+		return;
 	}
-	rz_return_val_if_fail(type->identifier.name, false);
-	bool result = true;
-	if (type->identifier.kind == RZ_TYPE_IDENTIFIER_KIND_STRUCT) {
-		// Get the base type
-		RzBaseType *btype = rz_type_db_get_base_type(typedb, type->identifier.name);
-		if (!btype) {
-			return false;
-		}
+	switch (btype->kind) {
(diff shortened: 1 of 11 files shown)


wargio (approved):

wargio:
looks like you broke various tests

thestr4ng3r:
yeah, needs a few adjustments and also contains a todo

thestr4ng3r:
Should hopefully work now

librz/type/path.c
```diff
@@ -122,91 +122,140 @@ RZ_API st64 rz_type_offset_by_path(const RzTypeDB *typedb, RZ_NONNULL const char
...
+		break;
 	}
-	return result;
 }
 
 /**
- * \brief Returns the list of all type paths matching the offset
+ * \brief Returns the list of all paths into the base type matching the offset
  *
- * \param typedb Types Database instance
```
XVilka, line 172: This one shouldn't have been removed since the `typedb` parameter is still here

librz/type/path.c
```diff
@@ -122,91 +122,140 @@ RZ_API st64 rz_type_offset_by_path(const RzTypeDB *typedb, RZ_NONNULL const char
...
- * \param btype The base type
- * \param offset The offset of the path to match against
+ * \param btype The root type from which paths are searched
+ * \param offset The offset into \p btype to match against
+ * \param max_depth Maximum number of components a path may have
  */
-RZ_API RZ_OWN RzList /*<RzTypePath *>*/ *rz_type_path_by_offset(const RzTypeDB *typedb, RzBaseType *btype, ut64 offset) {
+RZ_API RZ_OWN RzList /*<RzTypePath *>*/ *rz_base_type_path_by_offset(const RzTypeDB *typedb, RzBaseType *btype, ut64 offset, unsigned int max_depth) {
```
XVilka: Maybe worth writing that `RzBaseType *btype` is also `const`?

librz/type/path.c
```diff
@@ -122,91 +122,140 @@ RZ_API st64 rz_type_offset_by_path(const RzTypeDB *typedb, RZ_NONNULL const char
...
+		return NULL;
+	}
+	collect_base_type_paths(typedb, list, btype, "", offset, max_depth);
+	return list;
+}
+
+/**
+ * \brief Returns the list of all paths into the type matching the offset
+ *
+ * \param type The root type from which paths are searched
```
XVilka, line 248: Also missing `typedb` parameter?

XVilka: And also `RzType *type` is const?

XVilka (approved):
Two minor comments about the documentation, but the rest looks great.

librz/type/path.c
thestr4ng3r: The parameter exists, but `Types Database instance` adds no value to what the type already says. Doxygen will show the parameter anyway.

librz/type/path.c
thestr4ng3r: Good point

librz/type/path.c
thestr4ng3r: same as above, there isn't really anything valuable to document about the `typedb`.

PR #3266, opened 2023-01-05
Remove temporary type string conversion in DWARF integration

thestr4ng3r:

**Detailed description**

Due to the DWARF parsing having been implemented before proper RzTypes, it was still generating C-like strings that we then parsed into RzType. This is fragile and was identified to be a major performance bottleneck on OpenBSD/sparc64. Instead, we now generate RzTypes directly where possible.
There are still cases left where strings have to be generated from those types to be stored in an SDB. This should be changed in the future too.

Before:
```
sparc$ time rz -Qc '' test/bins/elf/ppc/emulateme-ppc64le
    0m30.31s real     0m28.03s user     0m01.73s system
```
After:
```
sparc$ time rz -Qc '' test/bins/elf/ppc/emulateme-ppc64le
    0m18.75s real     0m16.63s user     0m01.71s system
```
Still room for improvement.

### DIFF
--- a/librz/type/helpers.c
+++ b/librz/type/helpers.c
@@ -101,9 +101,6 @@ RZ_API RZ_OWN RzType *rz_type_pointer_of_base_type_str(const RzTypeDB *typedb, R
  */
 RZ_API RZ_OWN RzType *rz_type_pointer_of_type(const RzTypeDB *typedb, RZ_NONNULL RzType *type, bool is_const) {
 	rz_return_val_if_fail(typedb && type, NULL);
-	if (type->kind == RZ_TYPE_KIND_IDENTIFIER) {
-		return rz_type_pointer_of_base_type_str(typedb, type->identifier.name, is_const);
-	}
 	RzType *newtype = RZ_NEW0(RzType);
 	if (!newtype) {
 		return NULL;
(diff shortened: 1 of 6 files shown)


librz/type/helpers.c
```diff
@@ -101,9 +101,6 @@ RZ_API RZ_OWN RzType *rz_type_pointer_of_base_type_str(const RzTypeDB *typedb, R
  */
 RZ_API RZ_OWN RzType *rz_type_pointer_of_type(const RzTypeDB *typedb, RZ_NONNULL RzType *type, bool is_const) {
 	rz_return_val_if_fail(typedb && type, NULL);
-	if (type->kind == RZ_TYPE_KIND_IDENTIFIER) {
-		return rz_type_pointer_of_base_type_str(typedb, type->identifier.name, is_const);
-	}
 	RzType *newtype = RZ_NEW0(RzType);
```
ret2libc, line 104: Was this unused?

ret2libc (commented):
Nice!

librz/type/helpers.c
```diff
@@ -171,6 +168,22 @@ RZ_API RZ_OWN RzType *rz_type_array_of_type(const RzTypeDB *typedb, RZ_NONNULL R
 	return newtype;
 }
 
+/**
+ * \brief Creates a new array RzType of the given callable
```
XVilka: Misleading description?

XVilka (approved):
LGTM apart from that one thing.

librz/type/helpers.c
thestr4ng3r: yes

librz/type/helpers.c
thestr4ng3r: Probably needs clarification. I added this to the commit description:

    rz_type_pointer_of_type() is now also always creating a pointer type of
    the given type. Before, it would have failed when the pointee was an
    identifier that did not exist in the database. But this may be a valid
    case e.g. while still building the database from DWARF and such
    sanitization goes beyond what one would expect from
    rz_type_pointer_of_type().

XVilka (approved):

PR #3270, opened 2023-01-07
Make xnu reg profile depend on cpu type instead of RzDebug.bits

thestr4ng3r:

**Detailed description**

RzDebug.bits often has nonsensical contents set from all kinds of places. Specifically, when debugging fat binaries on arm64, it had a value not matching the debuggee when the reg profile was loaded, resulting in the arm32 profile being used.
The right way to determine the profile is to do it after attach, when the cpu type of the debuggee is known.

**Test plan**

see the added tests. The `test/db/archos/darwin-arm64/dbg` one would fail before.

### DIFF
--- a/librz/debug/dreg.c
+++ b/librz/debug/dreg.c
@@ -123,7 +123,14 @@ RZ_API ut64 rz_debug_num_callback(RzNum *userptr, const char *str, int *ok) {
 	return rz_reg_get_value(dbg->reg, ri);
 }
 
+/**
+ * Load the register profile from the current plugin into dbg->reg and
+ * if successful also sync all register contents into it.
+ *
+ * \return whether the register profile was provided by the plugin
+ */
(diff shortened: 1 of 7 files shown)


librz/debug/dreg.c
```diff
@@ -123,6 +123,12 @@ RZ_API ut64 rz_debug_num_callback(RzNum *userptr, const char *str, int *ok) {
 	return rz_reg_get_value(dbg->reg, ri);
 }
 
+/**
+ * Load the register profile from the current plugin into dbg->reg and
+ * if successful also sync all register contents into it.
+ *
+ * \return whether the register profile was provided by the plugin
+ */
 RZ_API bool rz_debug_reg_profile_sync(RzDebug *dbg) {
```
XVilka, line 132: could you please add `rz_return_if_fail(dbg, false)`?

XVilka (approved):

XVilka (approved):

Issue #2704, opened 2022-06-18
Rewrite GPL-only code


XVilka:
To become compatible with LGPL and other licenses, allowing Rizin library to be used anywhere.

# Disassemblers

- [x] SH [`librz/asm/arch/sh/gnu`](https://github.com/rizinorg/rizin/tree/dev/librz/asm/arch/sh/gnu) by @DMaroo, see https://github.com/rizinorg/rizin/pull/2518
- [x] Tricore [`librz/asm/arch/tricore/gnu`](https://github.com/rizinorg/rizin/tree/dev/librz/asm/arch/tricore/gnu) by @imbillow, see https://github.com/rizinorg/rizin/pull/3463
- [x] NIOS (**REMOVED**) [`librz/asm/arch/nios/gnu`](https://github.com/rizinorg/rizin/tree/dev/librz/asm/arch/nios/gnu) - could be done with capstone auto-sync and this unofficial LLVM port: https://github.com/ivanllopard/nios2-llvm/tree/master/lib/Target/Nios2
- [x] Xtensa [`librz/asm/arch/xtensa/gnu`](https://github.com/rizinorg/rizin/tree/dev/librz/asm/arch/xtensa/gnu) by @imbillow  https://github.com/rizinorg/rizin/pull/4654
- [x] HPPA (PARISC) [`librz/asm/arch/hppa/gnu`](https://github.com/rizinorg/rizin/tree/dev/librz/asm/arch/hppa/gnu) - by @XVilka (https://github.com/rizinorg/rizin/pull/4469)
- [x] SPARC [`librz/asm/arch/sparc/gnu`](https://github.com/rizinorg/rizin/tree/dev/librz/asm/arch/sparc/gnu) - by @Rot127  https://github.com/rizinorg/rizin/pull/5152 https://github.com/rizinorg/rizin/pull/5158
- [x] RISC-V [`librz/asm/arch/riscv/gnu`](https://github.com/rizinorg/rizin/tree/dev/librz/asm/arch/riscv/gnu) by @moste00 https://github.com/rizinorg/rizin/pull/5482
- [ ] ARC [`librz/asm/arch/arc/gnu`](https://github.com/rizinorg/rizin/tree/dev/librz/asm/arch/arc/gnu) (requires latest Capstone)
[...]

PR #3289, opened 2023-01-13
Add comment on load opcodes with the loaded value

bucienator:

**Detailed description**

First time user of Rizin here, I was doing some analysis on Xtensa code. (esp8266 firmware.) I found that in that architecture 32 bit constant values are loaded into registers by a PC relative load operation, loading the constant from the code segment.

For example:
`0x4000bf52      l32r  a5, 0x400003f4`

As 0x400003f4 is in the ROM as well, the loaded value is known at analysis time, and is usually needed to understand the analyzed function. So I was looking for ways to have the actual value shown in the disassembly, and ultimately came up with the change in this PR.

This modifies the disassembly output by adding a comment to it:

`0x4000bf52      l32r  a5, 0x400003f4                       ; refval: 0x0000ff00`

Actually I'm not yet happy with the change though, there are some things to refine on it:
* Check if the address is indeed in non-writable memory area, and so is useful in static analysis
* The loaded address is always treated as 32 bits, which is true for the Xtensa case, but not necessarily for other platform's load operations. I'll see how the size of the data could be determined
* I'm not sure if "refval" is the right term for the loaded value, maybe there's a more commonly known name for it
* I think it doesn't handle target endianness correctly
* And lastly, there might be better ways to have a similar outcome, maybe even without a code change.

I'm looking for some feedback if this change is deemed worthy for further improvements, or I should just drop it in favor of a different approach.

**Test plan**

(Tested manually so far, I'll look for unit tests for the disassebly functionality.)

### DIFF
--- a/librz/analysis/p/analysis_xtensa.c
+++ b/librz/analysis/p/analysis_xtensa.c
@@ -118,6 +118,7 @@ static void xtensa_shr_op(RzAnalysis *analysis, RzAnalysisOp *op, ut64 addr, con
 static void xtensa_l32r_op(RzAnalysis *analysis, RzAnalysisOp *op, ut64 addr, const ut8 *buf) {
 	op->type = RZ_ANALYSIS_OP_TYPE_LOAD;
 	op->ptr = ((addr + 3) & ~3) + ((buf[2] << 8 | buf[1]) << 2) - 0x40000;
+	op->refptr = 4;
 }
 
 static void xtensa_snm0_op(RzAnalysis *analysis, RzAnalysisOp *op, ut64 addr, const ut8 *buf) {
--- a/test/db/analysis/xtensa
+++ b/test/db/analysis/xtensa
@@ -26,3 +26,18 @@ var int32_t var_8h @ stack - 0x8
 var int32_t var_4h @ stack - 0x4
 EOF
 RUN
+
+
+NAME=xtensa l32r refptr
+FILE=malloc://512
+CMDS=<<EOF
+e asm.arch=xtensa
+e asm.bits=32
+wx 00000000ff11ff1121ffff0df0ff
+pD 3 @ 8
+EOF
+EXPECT=<<EOF
+            0x00000008      l32r  a2, 0x00000004                       ; [0x4:4]=0x11ff11ff
+EOF
+RUN
+


librz/core/disasm.c
```diff
@@ -4099,6 +4099,12 @@ static void ds_print_ptr(RzDisasmState *ds, int len, int idx) {
 			}
 		}
 		rz_io_read_at(core->io, refaddr, (ut8 *)msg, len - 1);
+		if (ds->analysis_op.type == RZ_ANALYSIS_OP_TYPE_LOAD) {
+			ut32 referenced_value = ((ut32 *)msg)[0];
+			ds_begin_comment(ds);
+			ds_comment(ds, true, "; refval: 0x%08" PFMT64x, referenced_value);
+		}
+
```
wargio: probably you want to ensure that the `refptr` value is initialized for load instructions in the arch. also there is a host endianness issue where msg can be little endian as big endian, making this code outputting different values if analyzed on hosts with different endianness.

wargio (changes requested):

librz/core/disasm.c
wargio: probably you want to add it here: https://github.com/rizinorg/rizin/blob/dev/librz/analysis/p/analysis_xtensa.c#L70

wargio: for example: https://github.com/rizinorg/rizin/blob/dev/librz/analysis/p/analysis_avr.c#L159

bucienator: Ok, so took a look at `refptr`, and actually setting that to 4 for op `l32r` has a perfect outcome:

`0x4000bf52      l32r  a5, 0x400003f4                       ; [0x400003f4:4]=0xff00`

I updated the PR, to become a much less intrusive change, which also solves much of the concerns with the original.

XVilka (approved):

XVilka:
Please add a test, especially in light of the future migration of the Xtensa support to capstone engine:
- https://github.com/rizinorg/rizin/issues/2704
- https://github.com/capstone-engine/capstone/pull/1949

wargio (approved):

bucienator:
> Please add a test, especially in light of the future migration of the Xtensa support to capstone engine:

I was considering that. My understanding is that I need to add a test binary to the rizin-testbins project, and then the test can refer to that. Is that the way? Should I add a binary only, or source code and build instructions too?

bucienator:
I've added a test case. Is that a good way to test this change?

XVilka (approved):
@bucienator looks good, thank you!

By the way, next time, I recommend using a different branch name. Usually, it's recommended to create a new "feature branch" for each PR separately.

PR #3292, opened 2023-01-14
Make afs respect the cc (Fix #3291)

### LINKED ISSUE #3291: Applying function signature swaps arguments with 'fastcall' cc.
*Issue transferred from Cutter to Rizin by Megabeets*

---

**Environment information**
* Operating System: Windows 10, 64bit
* Cutter version: 2.1.2
* Obtained from:
  - [ ] Built from source
  - [x] Distribution repository
* File format: PE x86 binary (not 64-bit)

**Describe the bug**

When assigning a function signature on a `fastcall` calling convention function, the ECX and EDX arguments are swapped.

**To Reproduce**

1. Open up an executable that contains a fastcall function with 2 or more arguments. Note that in my test example, the function is using a total of 4 arguments.
cutter_oTFSmsChzC

2. Mark the function as using 'fastcall' calling convention, using the dropdown menu 
cutter_BGyJX17YcJ

3. Run `af` in command, and then refresh the disassembler view using F5, verifying that the two arguments have been detected. At this point they are in the correct order.
cutter_YjRrkax2Ve

4. Apply a new function signature using `afs`. In this case I used `afs int32_t example_function(int32_t arg1, LPCSTR file_path, int32_t arg3, int32_t arg4);`
cutter_SxW2y0sB6E

5. After refreshing, the arguments are incorrectly flipped in the disassemble view.
cutter_FbAV0nrXxu

**Expected behavior**

Following fastcall, ECX and EDX should be assigned from left to right, thus, ECX should be argument 1, and EDX should be argument 2, as long as the parameters fit in registers.

**Additional context**

I have seen some similar behavior with cdecl, but was unable to quickly find reliable repro steps.

thestr4ng3r:

**Detailed description**

Applying a function signature with afs to a function was using a combination of the existing args and wrong stack addresses to apply the args to function variables. This generally resulted in incorrect locations. Instead, we now use the info from the cc.

**Test plan**

see the added test and issue noted below

**Closing issues**

Fix #3291


XVilka (approved):
Not a big fan of the `rz_analysis_function_delete_vars_by_arg_property` name though

thestr4ng3r:
suggestions welcome

XVilka:
No better idea either, TBH. Maybe other people could come with it?

librz/analysis/var.c
```diff
@@ -203,6 +203,24 @@ RZ_API void rz_analysis_function_delete_vars_by_storage_type(RzAnalysisFunction
 	}
 }
 
+/**
+ * Delete all variables from \p fcn that are either arguments or locals, depending on \p args
+ * \param args if true, all args will be deleted, otherwise all non-arg vars will be deleted
+ */
+RZ_API void rz_analysis_function_delete_vars_by_arg_property(RzAnalysisFunction *fcn, bool args) {
```
wargio: args is never used.

wargio (changes requested):

librz/analysis/var.c
LaylBongers: `args` parameter is not used, this function currently always deletes args, while the parameter says it will flip between args or not. This function is only used in one place where this is in fact the case.
There is the option of just removing the parameter, in which case the function is better called `rz_analysis_function_delete_all_args` or `rz_analysis_function_delete_arg_vars`.
Alternatively, if select specific filters are desired, it may be clearer to provide an enum or flags, allowing more filter options in the future. In this case a fitting name would be `rz_analysis_function_delete_vars` (as opposed to the "all_vars" version, though merging these functions to some degree would make sense), or `rz_analysis_function_delete_vars_matching`.

wargio: i think it would make more sense to just delete and recreate the definition of the function, having filters etc., might be counterproductive and make the code more complex

thestr4ng3r: Good catch, I will make it `rz_analysis_function_delete_arg_vars()`. That way the ugly name is also gone. The fact that the arg slipped through already shows this would have been more complexity than needed.

@wargio Recreating the entire function would not work for this use-case because you want to keep all (e.g. manually renamed and typed) local variables, control flow analysis, etc.

XVilka (approved):

PR #3306, opened 2023-01-21
Fix #3280 - Restores color on returning from Rh.

### LINKED ISSUE #3280: Uncolored command line after using rizin HTTP web server via `Rh`
### Work environment

| `rizin -v` full output  | rizin 0.4.1 @ linux-x86-64 commit: 9023f8b997db210cef3b9a25cf1748fbc94942ed

### Expected behavior
Return into previous colored radare command line mode as if nothing happened inbetween.

### Actual behavior
Command line remains uncolored.
image

- Open rizin
- Run `Rh` (HTTP web server)
- Abbort via SIGTERM (<kbd>Ctr+C</kbd>)
- See error

0xn4utilus:
# SQUASH ME

**Detailed description**

Restores colored radare command line after returning/aborting from rizin http server.  

**Test plan**

...

**Closing issues**

closes #3280
...


XVilka (changes requested):
This fix is wrong since it doesn't take in account previous `scr.color` values, which are set by the user commonly.

0xn4utilus:
Hi, sorry for late response. Hope this is the correct fix

librz/core/rtr_http.c
```diff
@@ -127,7 +127,7 @@ static int rz_core_rtr_http_run(RzCore *core, int launch, int browse, const char
 		so.accept_timeout = 1;
 	}
 
-	origcfg = core->config;
+	origcfg = rz_config_clone(core->config);
```
wargio: i think this is creating a memleak
the original allocated structure held by core->config always gets overwritten by a new pointer.

0xn4utilus: isn't cloning `core->config` necessary to store the original configs?

wargio: yes, but the issue is that the whole code is very messy. i strongly suggest to use instead `rz_config_hold_*` functions and remove this replacement of the `core->config`.

essentially what you would do is allocate a new `RzConfigHold` and then call `rz_config_hold_s|i(hold, "cfg.color", ...)` for each config variable that needs to be restored later, then at the end you will just call `rz_config_hold_restore` and `rz_config_hold_free` to restore the settings and free the allocated `RzConfigHold` structure.

As pure example on how to implement this, i suggest this function: https://github.com/rizinorg/rizin/blob/dev/librz/core/cmd/cmd_print.c#L5986

0xn4utilus: Thanks, it was really helpful.

0xn4utilus: Made the changes

librz/core/rtr_http.c
```diff
@@ -127,14 +126,20 @@ static int rz_core_rtr_http_run(RzCore *core, int launch, int browse, const char
...
-	origcfg = core->config;
-	newcfg = rz_config_clone(core->config);
-	core->config = newcfg;
+	// store current configs
+	RzConfigHold *hc = rz_config_hold_new(core->config);
+	rz_config_hold_i(hc, "scr.color", NULL);
+	rz_config_hold_i(hc, "scr.html", NULL);
+	rz_config_hold_i(hc, "scr.interactive", NULL);
+	rz_config_hold_i(hc, "asm.cmt.right", NULL);
+	rz_config_hold_i(hc, "asm.bytes", NULL);
```
wargio:
```suggestion
	rz_config_hold_i(hc, "scr.color", "scr.html", "scr.interactive", "asm.cmt.right", "asm.bytes", NULL);
```

librz/core/rtr_http.c
```diff
@@ -127,14 +126,20 @@ static int rz_core_rtr_http_run(RzCore *core, int launch, int browse, const char
 		so.accept_timeout = 1;
 	}
 
-	origcfg = core->config;
-	newcfg = rz_config_clone(core->config);
-	core->config = newcfg;
+	// store current configs
+	RzConfigHold *hc = rz_config_hold_new(core->config);
```
wargio, line 130: add a check in case `hc` is NULL.

librz/core/rtr_http.c
```diff
@@ -486,7 +483,7 @@ the_end : {
 	const char *port = rz_config_get(core->config, "http.port");
 	const char *cors = rz_config_get(core->config, "http.cors");
 	const char *allow = rz_config_get(core->config, "http.allow");
-	core->config = origcfg;
+
```
wargio: Probably you want to remove all the `rz_config_get` and `rz_config_set` above and below, because these calls makes no sense.

librz/core/rtr_http.c
```diff
@@ -127,14 +126,20 @@ static int rz_core_rtr_http_run(RzCore *core, int launch, int browse, const char
...
 
+	// set new configs
 	rz_config_set(core->config, "asm.cmt.right", "false");
 	rz_config_set_i(core->config, "scr.color", COLOR_MODE_DISABLED);
 	rz_config_set(core->config, "asm.bytes", "false");
 	rz_config_set(core->config, "scr.interactive", "false");
+
 	RZ_LOG_INFO("core: Starting http server...\n");
 	RZ_LOG_INFO("core: open http://%s:%d/\n", host, atoi(port));
 	RZ_LOG_INFO("core: rizin -C http://%s:%d/cmd/\n", host, atoi(port));
```
wargio:
```suggestion
	RZ_LOG_WARN("core: Starting http server...\nTo open a remote session, please use `rizin -C http://%s:%s/cmd/`\n", host, port);
```

wargio (changes requested):

wargio:
btw, i like the changes. good job :)
There are just some minor details missing.

librz/core/rtr_http.c
0xn4utilus: > btw, i like the changes. good job :) There are just some minor details missing.

Thanks a lot 😄. Here are the requested changes.

wargio:
please install `clang-format` and run `./sys/clang-format.py` then commit and push. for the rest looks good. great job

XVilka (approved):

wargio (approved):

PR #3313, opened 2023-01-22
librz: do not duplicate plugins

ret2libc:
Plugins structures are statically defined at the file scope and not dynamically allocated. Plugins should not allocate them either as RzLibPlugin keeps a reference to them anyway.

 

**Detailed description**

Using rz_list_delete_data was wrong because the plugin passed to rz_*_plugin_del was the one stored in RzLibPlugin, however before this patch the plugins were duplicated by rz_*_plugin_add. This was causing crashes  in some cases.

**Test plan**

...

**Closing issues**

...


librz/lang/lang.c
```diff
@@ -46,11 +45,16 @@ RZ_API RzLang *rz_lang_new(void) {
 
 RZ_API void rz_lang_free(RzLang *lang) {
 	if (lang) {
```
XVilka: Could you please flip the condition here while you are on it?

XVilka: Also add `RZ_NULLABLE`

librz/parse/parse.c
```diff
@@ -39,6 +39,13 @@ RZ_API RzParse *rz_parse_new(void) {
 }
 
 RZ_API void rz_parse_free(RzParse *p) {
```
XVilka, line 41: Add a quick exit if `p` is `NULL`, maybe?

XVilka (changes requested):

librz/lang/lang.c
ret2libc: I don't think we put NULLABLE in all _free methods. I *think* we have a convention that _free will always accept NULL and do nothing.

librz/util/list.c
```diff
@@ -126,7 +126,7 @@ RZ_API void rz_list_purge(RZ_NONNULL RzList *list) {
  * \brief Empties the list and frees the list pointer
  *
  **/
-RZ_API void rz_list_free(RZ_NONNULL RzList *list) {
+RZ_API void rz_list_free(RzList *list) {
 	if (list) {
```
XVilka, line 130: Maybe flip too?

XVilka (approved):

librz/util/list.c
ret2libc: it's just 2 instructions, that's why i didn't change it. I think it can be changed if we ever need to change rzlist(though i doubt).

PR #3321, opened 2023-01-25
Refactor srec reading to not use sscanf

thestr4ng3r:

**Detailed description**

Parsing with sscanf turned out to cause an extreme slowdown. As only hex values are being read, this can be done with RzUtil-provided functionality.

On OpenBSD/sparc64, before:
```
sparc$ time rz -Qc "" srec://test/bins/srec/arm1.bin.srec
   46m51.38s real    46m50.37s user     0m00.67s system
```
after:
```
sparc$ time rz -Qc "" srec://test/bins/srec/arm1.bin.srec
    0m04.92s real     0m03.85s user     0m00.91s system
```

**Test plan**

Relevant test cases exist in test/db/io/srec.

### DIFF
--- a/librz/io/p/io_srec.c
+++ b/librz/io/p/io_srec.c
@@ -134,6 +134,52 @@ static bool __plugin_open(RzIO *io, const char *pathname, bool many) {
 	return (!strncmp(pathname, SREC_PATH_PREFIX, strlen(SREC_PATH_PREFIX)));
 }
 
+/**
+ * Parse a value formatted like "%02x" of exactly 2 chars
+ */
+static bool parse_hex_byte(const char *str, RZ_OUT ut8 *out) {
+	if (!str[0] || !str[1]) {
+		return false;
+	}
+	*out = 0;
+	return !rz_hex_to_byte(out, str[0]) && !rz_hex_to_byte(out, str[1]);
+}
+
+/**
+ * Parse an n-byte big endian hex value of exactly 2*n chars
+ */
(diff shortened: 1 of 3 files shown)


librz/io/p/io_srec.c
```diff
@@ -157,8 +203,8 @@ static bool srecord_parse(RzBuffer *buf, char *str) {
 
 	do {
 		line++;
-		if (sscanf(str, "S%c%02x", &record_type, &byte_count) != 2) {
-			RZ_LOG_ERROR("srec:parse(): invalid data in motorola srecord file at line %d\n", line);
+		if (!parse_srecord_prefix(str, &record_type, &byte_count)) {
+			RZ_LOG_ERROR("srec:parse(): invalid prefix in motorola srecord file at line %d\n", line);
```
XVilka: Maybe `Motorola SRecord` instead of `motorola srecord`?

XVilka (approved):
I wonder why it is so slow with `sscanf()`. 46min! That's quite something.

thestr4ng3r:
I think it has to do with locales. It uses `isspace()` which is locale-dependent: https://cvsweb.openbsd.org/cgi-bin/cvsweb/~checkout~/src/lib/libc/stdio/vfscanf.c?rev=1.35&content-type=text/plain
But it is not only on OpenBSD, I see similar speedup on macOS.

librz/io/p/io_srec.c
thestr4ng3r: Wikipedia says "Motorola S-record", let's use that.

PR #3373, opened 2023-02-17
Add macros to generate inf/nan for different types

Heersin:
Since `strtold` sometimes return weird value, use a portable solution
- NaN = 0 / 0
- Inf = 1 / 0 

**Detailed description**

see above

**Test plan**

waiting for CI to check portable issues

**Closing issues**

...

### DIFF
--- a/librz/include/rz_types_base.h
+++ b/librz/include/rz_types_base.h
@@ -5,6 +5,7 @@
 #include <sys/types.h>
 #include <limits.h>
 #include <math.h>
+#include <fenv.h>
 
 #define cut8  const unsigned char
 #define ut64  unsigned long long
@@ -95,16 +96,6 @@ typedef struct _utX {
 #define UT8_MIN   0x00U
 #define ASCII_MIN 32
 #define ASCII_MAX 127
-
-#define F32_NAN   (strtof("NAN", NULL))
-#define F32_PINF  (strtof("INF", NULL))
-#define F32_NINF  (-strtof("INF", NULL))
-#define F64_NAN   (strtod("NAN", NULL))
-#define F64_PINF  (strtod("INF", NULL))
-#define F64_NINF  (-strtod("INF", NULL))
-#define F128_NAN  (strtold("NAN", NULL))
-#define F128_PINF (strtold("INF", NULL))
-#define F128_NINF (-strtold("INF", NULL))
 
 #if SSIZE_MAX == ST32_MAX
 #define SZT_MAX  UT32_MAX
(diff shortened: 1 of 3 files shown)


librz/include/rz_types_base.h
```diff
@@ -5,6 +5,7 @@
 #include <sys/types.h>
 #include <limits.h>
 #include <math.h>
+#include <fenv.h>
```
XVilka: Is this one necessary?

librz/include/rz_util/rz_float.h
```diff
@@ -10,6 +10,30 @@
...
+/**
+ *
+ * Portable float nums in C
+ */
+RZ_API float types_gen_f32_nan(void);
+RZ_API float types_gen_f32_inf(void);
+RZ_API double types_gen_f64_nan(void);
+RZ_API double types_gen_f64_inf(void);
+RZ_API long double types_gen_f128_nan(void);
+RZ_API long double types_gen_f128_inf(void);
```
wargio: since these are `RZ_API` the name should start with `rz_`.
I would say to name them `rz_f32/64/128_inf/nan`

XVilka (commented):
The code looks fine, maybe worth adding Doxygen brief for these functions if CI is green.

librz/util/float/float.c
```diff
@@ -20,6 +20,33 @@
...
+#define define_types_gen_nan(fname, ftype) \
+	RZ_API inline ftype types_gen_##fname##_nan() { \
+		static ftype zero = 0; \
+		ftype ret = zero / zero; \
+		feclearexcept(FE_ALL_EXCEPT); \
+		return ret; \
+	}
+
+#define define_types_gen_inf(fname, ftype) \
+	RZ_API inline ftype types_gen_##fname##_inf() { \
```
wargio: there is no real need to set the `inline` param since they are not in a header file.

librz/include/rz_types_base.h
wargio: yes for the `feclearexcept` function, but i guess that could be moved into `librz/util/float/float.c`

XVilka: Yes, better to move to the `librz/util/float/float.c` then.

librz/util/float/float.c
XVilka: It's possible to force inlining using both `extern` and `inline` though: https://gcc.gnu.org/onlinedocs/gcc/Inline.html
But it's very GCC-specific and likely will not work on other compilers.

Heersin: then I will just remove it.

wargio (approved):

XVilka (approved):

PR #3421, opened 2023-03-10
Replace `rz_core_cmd0` calls in `tui/panels.c` and port `cmd_help` commands to `RzShell` as `cmd_math`

### LINKED ISSUE #2834: Port `?` (help) to rzshell

brightprogrammer:
# SQUASH ME

**Detailed description**

- Replaces some of the `rz_core_cmd0` in `/core/tui/panels.c` with corresponding API/IPI calls as per the project (**Switch from command calls to the API**) requirements.
- Port `cmd_help` to RzShell.
- In lateast commits  `?` symbol used in `cmd_help` is changed to to `cmd_math` (new created command group) and `%` symbol. Most of the math commands now use `%` symbols instead of `?`.
- Some of the commands that made more sense in `cmd_shell` have been moved there, like the version info command or command excecution time calculator command.  

Closes https://github.com/rizinorg/rizin/issues/2834
Partially addresses https://github.com/rizinorg/rizin/issues/383


librz/core/tui/panels.c
```diff
@@ -3607,8 +3608,14 @@ int __writeValueCb(void *user) {
 	RzCore *core = (RzCore *)user;
 	char *res = __show_status_input(core, "insert number: ");
 	if (res) {
-		rz_core_cmdf(core, "\"wv %s\"", res);
+		ut64 value = rz_num_math(core->num, res);
+		if (core->num->nc.errors) {
+			RZ_LOG_ERROR("Could not convert argument to number");
+			free(res);
+			return RZ_CMD_STATUS_ERROR;
```
ret2libc: you want to return something else here. Like `-1`

librz/core/tui/panels.c
```diff
@@ -3607,8 +3608,14 @@ int __writeValueCb(void *user) {
...
+		}
 		free(res);
+		return rz_core_write_value_at(core, core->offset, value, 0) ? RZ_CMD_STATUS_OK : RZ_CMD_STATUS_ERROR;
```
ret2libc:
```suggestion
		return rz_core_write_value_at(core, core->offset, value, 0) ? 0 : -1;
```

librz/core/cmd_descs/cmd_descs.h
```diff
@@ -1374,6 +1374,8 @@ RZ_IPI RzCmdStatus rz_egg_reset_handler(RzCore *core, int argc, const char **arg
 RZ_IPI RzCmdStatus rz_egg_show_config_handler(RzCore *core, int argc, const char **argv);
 // "help"
 RZ_IPI RzCmdStatus rz_help_handler(RzCore *core, int argc, const char **argv);
+// "? <expr>"
+RZ_IPI void rz_help_calc_expr(RzCore* core, const char* input);
```
ret2libc: You don't want to place functions here. This is an auto-generated file as it should be written at the head of the file. Place this declaration in another header like rz_core.h

librz/core/cmd/cmd_help.c
```diff
@@ -430,6 +430,102 @@ RZ_IPI void rz_core_clippy_print(RzCore *core, const char *msg) {
 	}
 }
 
+RZ_IPI void rz_help_calc_expr(void* data, const char* input) {
```
ret2libc:
```suggestion
RZ_API void rz_help_calc_expr(RzCore *core, const char* input) {
```

ret2libc: Also, you may want to transform this into something like `rz_core_asm_plugins_print`, which accepts a `RzCmdStateOutput`.

Then you can call it with something like https://github.com/rizinorg/rizin/blob/dev/librz/core/cconfig.c#L426

ret2libc (changes requested):

librz/core/tui/panels.c
```diff
@@ -4238,7 +4245,7 @@ void __init_menu_saved_layout(void *_core, const char *parent) {
 void __init_menu_color_settings_layout(void *_core, const char *parent) {
 	RzCore *core = (RzCore *)_core;
 	const char *color = core->cons->context->pal.graph_box2;
-	char *now = rz_core_cmd_str(core, "eco.");
+	char *now = rz_core_theme_get(core);
```
XVilka: `const char*`

XVilka (changes requested):

librz/core/cmd/cmd_help.c
brightprogrammer: I used this function directly inside the `rz_cmd_help` function in `"? "` case. Would it be a good idea to modify it's signature? I'll be like copying the code? 

Since I needed only the part which calculates expression and prints it to screen, I removed that case from `rz_cmd_help` and created this new function and then used this function there directly.

librz/core/tui/panels.c
brightprogrammer: can't do `const char*` because there's a split below.
https://github.com/rizinorg/rizin/blob/fb71f368596879eb1664c5c2ac04c0613f425a0f/librz/core/tui/panels.c#L4242

brightprogrammer: I'll make a copy of string and pass it to this split.

brightprogrammer: I changed this, just forgot to free the duplicated string in this commit, I'll push the change in next commit, marking this as resolved.

librz/core/canalysis.c
```diff
@@ -6393,3 +6393,34 @@ RZ_API ut64 rz_core_prevop_addr_force(RzCore *core, ut64 start_addr, int numinst
 	}
 	return start_addr;
 }
+
+RZ_API void rz_core_perform_auto_analysis(RzCore *core, RzCoreAnalysisType type) {
```
brightprogrammer: had to move this function and related enum outside to this file in order to be able to use it directly instead of calling `rz_core_cmd("aa")` or `rz_core_cmd("aaa")` or `rz_core_cmd("aaaa")`.

librz/core/cmd/cmd_analysis.c
```diff
@@ -5894,55 +5898,18 @@ RZ_IPI RzCmdStatus rz_analysis_basic_block_find_paths_handler(RzCore *core, int
 	return status;
 }
 
-typedef enum {
```
brightprogrammer, line 5897: enum moved to `rz_core.h`

librz/core/canalysis.c
wargio: that sounds like a good idea.

wargio: before it was not an `RZ_API` so there was no need for this but you need to add these changes:
```suggestion
RZ_API void rz_core_perform_auto_analysis(RZ_NONNULL RzCore *core, RzCoreAnalysisType type) {
	rz_return_if_fail(core);
```

librz/core/cmd/cmd_help.c
```diff
@@ -430,6 +430,102 @@ RZ_IPI void rz_core_clippy_print(RzCore *core, const char *msg) {
 	}
 }
 
+RZ_API void rz_core_cmd_help_calc_expr(RzCore* data, const char* input) {
```
wargio: same thing here (as above)

librz/core/cmd/cmd_help.c
```diff
@@ -430,6 +430,102 @@ RZ_IPI void rz_core_clippy_print(RzCore *core, const char *msg) {
...
+	RzCore* core = (RzCore*)data;
```
wargio: why you need to cast something that is `RzCore` already into `RzCore` ? just rename `data` argument into `core` and delete this line.

librz/core/cmd/cmd_help.c
```diff
@@ -430,6 +430,102 @@ RZ_IPI void rz_core_clippy_print(RzCore *core, const char *msg) {
...
+	char *asnum, unit[8];
+	ut32 s, a;
+	double d;
+	float f;
+	char number[128], out[128] = RZ_EMPTY;
+	char *const inputs = strdup(input + 1);
```
wargio:
```suggestion
	const char *inputs = strdup(input + 1);
```

librz/core/cmd/cmd_help.c
```diff
@@ -430,6 +430,102 @@ RZ_IPI void rz_core_clippy_print(RzCore *core, const char *msg) {
...
+		pj_o(pj);
+	}
+	for (ut32 i = 0; i < list_len; i++) {
+		const char *str = rz_list_pop_head(list);
+		if (!*str) {
+			continue;
+		}
+		ut64 n = rz_num_math(core->num, str);
+		if (core->num->dbz) {
+			RZ_LOG_ERROR("core: RzNum ERROR: Division by Zero\n");
```
wargio:
```suggestion
			RZ_LOG_ERROR("core: RzNum ERROR: Division by Zero\n");
			return;
```

librz/include/rz_core.h
```diff
@@ -717,6 +723,8 @@ RZ_API bool rz_core_analysis_hint_set_offset(RZ_NONNULL RzCore *core, RZ_NONNULL
 RZ_API bool rz_core_analysis_continue_until_syscall(RZ_NONNULL RzCore *core);
 RZ_API bool rz_core_analysis_continue_until_call(RZ_NONNULL RzCore *core);
 
+RZ_API void rz_core_perform_auto_analysis(RzCore *core, RzCoreAnalysisType type);
```
wargio:
```suggestion
RZ_API void rz_core_perform_auto_analysis(RZ_NONNULL RzCore *core, RzCoreAnalysisType type);
```

librz/include/rz_core.h
```diff
@@ -1266,6 +1274,12 @@ typedef bool (*RzCmdRegSync)(RzCore *core, RzRegisterType type, bool write);
 RZ_API bool rz_core_reg_assign_sync(RZ_NONNULL RzCore *core, RZ_NONNULL RzReg *reg, RzCmdRegSync sync_cb, RZ_NONNULL const char *name, ut64 val);
 RZ_API RZ_OWN RzList /*<RzRegItem *>*/ *rz_core_reg_filter_items_sync(RZ_NONNULL RzCore *core, RZ_NONNULL RzReg *reg, RzCmdRegSync sync_cb, RZ_NULLABLE const char *filter);
 
+/* rz_core_cmd replacement function declarations */
+// "? <expr>"
+RZ_API void rz_core_cmd_help_calc_expr(RzCore* core, const char* input);
```
wargio:
```suggestion
RZ_API void rz_core_cmd_help_calc_expr(RZ_NONNULL RzCore* core, RZ_NONNULL const char* input);
```

librz/include/rz_core.h
```diff
@@ -1266,6 +1274,12 @@ typedef bool (*RzCmdRegSync)(RzCore *core, RzRegisterType type, bool write);
...
+// a? or aaaaa like help message cases
+RZ_API void rz_core_cmd_show_analysis_help(RzCore* core);
```
wargio:
```suggestion
RZ_API void rz_core_cmd_show_analysis_help(RZ_NONNULL RzCore* core);
```

wargio (changes requested):
For each of `RZ_NONNULL` added to the arguments, please add (if is not there) a `rz_return_if_fail(cond);` if the function returns `void` or `rz_return_val_if_fail(cond, value);` if returns anything else but `void`.

librz/core/cmd/cmd_help.c
brightprogrammer: I'm sorry, this was a side effect of copy pasting and forgot to change that. Will be changed in upcoming commit.

librz/core/canalysis.c
```diff
@@ -6394,7 +6394,9 @@ RZ_API ut64 rz_core_prevop_addr_force(RzCore *core, ut64 start_addr, int numinst
 	return start_addr;
 }
 
-RZ_API void rz_core_perform_auto_analysis(RzCore *core, RzCoreAnalysisType type) {
+RZ_API void rz_core_perform_auto_analysis(RZ_NONNULL RzCore *core, RzCoreAnalysisType type) {
+	rz_return_if_fail(core != NULL);
```
wargio:
```suggestion
	rz_return_if_fail(core);
```

librz/core/cmd/cmd_help.c
```diff
@@ -430,14 +430,15 @@ RZ_IPI void rz_core_clippy_print(RzCore *core, const char *msg) {
 	}
 }
 
-RZ_API void rz_core_cmd_help_calc_expr(RzCore* data, const char* input) {
-	RzCore* core = (RzCore*)data;
+RZ_API void rz_core_cmd_help_calc_expr(RzCore* core, const char* input) {
+	rz_return_if_fail(core != NULL);
```
wargio:
```suggestion
	rz_return_if_fail(core);
```

librz/include/rz_core.h
```diff
@@ -1276,9 +1276,9 @@ RZ_API RZ_OWN RzList /*<RzRegItem *>*/ *rz_core_reg_filter_items_sync(RZ_NONNULL
 
 /* rz_core_cmd replacement function declarations */
 // "? <expr>"
-RZ_API void rz_core_cmd_help_calc_expr(RzCore* core, const char* input);
+RZ_API void rz_core_cmd_help_calc_expr(RZ_NONNULL RzCore* core, const char* input);
```
wargio: input must be also `RZ_NONNULL` and also on the C file

librz/core/cmd/cmd_analysis.c
```diff
@@ -2028,7 +2028,7 @@ RZ_IPI RzCmdStatus rz_analysis_global_variable_retype_handler(RzCore *core, int
 }
 
 RZ_API void rz_core_cmd_show_analysis_help(RzCore *core) {
-	rz_return_if_fail(core != NULL);
+	rz_return_if_fail(core);
```
brightprogrammer, line 2031: forgot to change this and this commit is rebased with dev (as guided in `CONTRIBUTING.md`)

librz/core/cmd/cmd_help.c
brightprogrammer: this results in meson build failures for some compilers... Built on my system with a warning. Maybe I should remove const qualifier, because a function expects the input in that format
image

librz/main/rizin.c
```diff
@@ -851,12 +851,14 @@ RZ_API int rz_main_rizin(int argc, const char **argv) {
...
+			rz_core_rtr_add(r, uri);
 		} else {
 			argv[opt.ind] = rz_str_newf("http://%s/cmd/", argv[opt.ind]);
-			rz_core_cmdf(r, "R+ %s", argv[opt.ind]);
+			rz_core_rtr_add(r, argv[opt.ind]);
 		}
-		rz_core_cmd0(r, "R!=");
+		// rz_core_cmd0(r, "R!=");
+		RZ_FREE(r->cmdremote);
+		r->cmdremote = rz_str_trim_dup("0");
```
ret2libc: Please extract this into `rz_core_rtr_enable(RZ_NONNULL RzCore *, RZ_NONNULL char *cmdremote)` and use it in both cmd_remote.c and here.

librz/include/rz_core.h
```diff
@@ -1266,6 +1274,12 @@ typedef bool (*RzCmdRegSync)(RzCore *core, RzRegisterType type, bool write);
 RZ_API bool rz_core_reg_assign_sync(RZ_NONNULL RzCore *core, RZ_NONNULL RzReg *reg, RzCmdRegSync sync_cb, RZ_NONNULL const char *name, ut64 val);
 RZ_API RZ_OWN RzList /*<RzRegItem *>*/ *rz_core_reg_filter_items_sync(RZ_NONNULL RzCore *core, RZ_NONNULL RzReg *reg, RzCmdRegSync sync_cb, RZ_NULLABLE const char *filter);
 
+/* rz_core_cmd replacement function declarations */
```
ret2libc: no need for this, I think

librz/include/rz_core.h
```diff
@@ -1266,6 +1274,12 @@ typedef bool (*RzCmdRegSync)(RzCore *core, RzRegisterType type, bool write);
...
+// "? <expr>"
```
ret2libc: remove these comments as well.

ret2libc (commented):
Nice! Just few minor comments, but seems good to me :) Thanks!

librz/core/cmd/cmd_help.c
```diff
@@ -430,6 +430,105 @@ RZ_IPI void rz_core_clippy_print(RzCore *core, const char *msg) {
 	}
 }
 
+RZ_API void rz_core_cmd_help_calc_expr(RzCore *core, const char *input) {
+	rz_return_if_fail(core);
+	rz_return_if_fail(input);
```
wargio:
```suggestion
	rz_return_if_fail(core && input);
```

librz/core/cmd/cmd_help.c
```diff
@@ -430,6 +430,105 @@ RZ_IPI void rz_core_clippy_print(RzCore *core, const char *msg) {
 	}
 }
 
+RZ_API void rz_core_cmd_help_calc_expr(RzCore *core, const char *input) {
```
wargio:
```suggestion
RZ_API void rz_core_cmd_help_calc_expr(RZ_NONNULL RzCore *core, RZ_NONNULL const char *input) {
```

librz/core/cmd/cmd_analysis.c
```diff
@@ -2027,6 +2027,11 @@ RZ_IPI RzCmdStatus rz_analysis_global_variable_retype_handler(RzCore *core, int
 	return RZ_CMD_STATUS_OK;
 }
 
+RZ_API void rz_core_cmd_show_analysis_help(RzCore *core) {
```
wargio:
```suggestion
RZ_API void rz_core_cmd_show_analysis_help(RZ_NONNULL RzCore *core) {
```

librz/core/canalysis.c
```diff
@@ -6384,3 +6384,36 @@ RZ_API ut64 rz_core_prevop_addr_force(RzCore *core, ut64 start_addr, int numinst
 	}
 	return start_addr;
 }
+
+RZ_API void rz_core_perform_auto_analysis(RzCore *core, RzCoreAnalysisType type) {
```
wargio:
```suggestion
RZ_API void rz_core_perform_auto_analysis(RZ_NONNULL RzCore *core, RzCoreAnalysisType type) {
```

librz/core/cmd/cmd_help.c
```diff
@@ -430,6 +430,104 @@ RZ_IPI void rz_core_clippy_print(RzCore *core, const char *msg) {
...
```
XVilka: Since you are modifying this code, I suggest porting this particular command to the rzshell as well, and use `RzCmdStateOutput`, just add corresponding code into `librz/core/cmd_descs/cmd_help.yaml`. And extract evaluation function to the API, while keeping printing function as the IPI.

XVilka (changes requested):

librz/core/cmd/cmd_help.c
brightprogrammer: ok :smile:

librz/include/rz_core.h
```diff
@@ -1274,11 +1274,9 @@ typedef bool (*RzCmdRegSync)(RzCore *core, RzRegisterType type, bool write);
 RZ_API bool rz_core_reg_assign_sync(RZ_NONNULL RzCore *core, RZ_NONNULL RzReg *reg, RzCmdRegSync sync_cb, RZ_NONNULL const char *name, ut64 val);
 RZ_API RZ_OWN RzList /*<RzRegItem *>*/ *rz_core_reg_filter_items_sync(RZ_NONNULL RzCore *core, RZ_NONNULL RzReg *reg, RzCmdRegSync sync_cb, RZ_NULLABLE const char *filter);
 
-/* rz_core_cmd replacement function declarations */
-// "? <expr>"
 RZ_API void rz_core_cmd_help_calc_expr(RZ_NONNULL RzCore *core, RZ_NONNULL const char *input);
-// a? or aaaaa like help message cases
 RZ_API void rz_core_cmd_show_analysis_help(RZ_NONNULL RzCore *core);
+RZ_API void rz_core_rtr_enable(RZ_NONNULL RzCore *core, RZ_BORROW const char *cmdremote);
```
wargio: RZ_BORROW is only if the buffer is allocated by another function and held by a structure which should not be handled by the user (i.e. freed).

brightprogrammer:
I haven't written command handlers yet. That will be done in next two to three days :smile: Please review `cmd_help.yaml` till then. I'll be going through each entry again when implementing the handlers because I was confused about how arguments work in rzshell handlers. Will figure that out while working...

librz/core/cmd_descs/cmd_help.yaml
```diff
@@ -0,0 +1,396 @@
+# SPDX-FileCopyrightText: 2021 RizinOrg <info@rizin.re>
```
wargio:
```suggestion
# SPDX-FileCopyrightText: 2023 RizinOrg <info@rizin.re>
```

librz/core/cmd_descs/cmd_help.yaml
```diff
@@ -0,0 +1,396 @@
...
+    subcommands:
+      - name: "??"
+        summary: "Display value in $? register"
+        cname: display_core_num_value
+      - name: "???"
+        summary: "Display ? Help"
+        cname: show_help_question
+      - name: "????"
+        summary: "What are you doing?"
+        cname: what_are_you_doing?
```
wargio: remove that function and delete this command.

wargio (changes requested):
Just a couple of changes and then can be merged (once you implement the handlers).

librz/core/canalysis.c
```diff
@@ -6384,3 +6384,36 @@ RZ_API ut64 rz_core_prevop_addr_force(RzCore *core, ut64 start_addr, int numinst
 	}
 	return start_addr;
 }
+
+RZ_API void rz_core_perform_auto_analysis(RZ_NONNULL RzCore *core, RzCoreAnalysisType type) {
```
XVilka, line 6436: Please add Doxygen documentation

librz/core/canalysis.c
```diff
@@ -6384,3 +6384,36 @@ RZ_API ut64 rz_core_prevop_addr_force(RzCore *core, ut64 start_addr, int numinst
...
+	rz_cons_break_timeout(timeout);
+	rz_core_analysis_all(core);
+	rz_core_notify_done(core, "%s", notify);
+	rz_core_task_yield(&core->tasks);
+	if (type == RZ_CORE_ANALYSIS_SIMPLE || rz_cons_is_breaked()) {
+		goto finish;
+	}
+	// Run pending analysis immediately after analysis
+	// Usefull when running commands with ";" or via rizin -c,-i
+	debugger = core->dbg->cur ? strdup(core->dbg->cur->name) : strdup("esil");
```
XVilka: You can move definition of the variable here, no need to define everything at the beginning of the function - we support only >=C99 standard.

librz/core/cmd/cmd_analysis.c
```diff
@@ -5894,55 +5899,18 @@ RZ_IPI RzCmdStatus rz_analysis_basic_block_find_paths_handler(RzCore *core, int
...
-	rz_core_seek(core, old_offset, true);
-	// XXX this shouldnt be called. flags muts be created wheen the function is registered
-	rz_core_analysis_flag_every_function(core);
-	rz_cons_break_pop();
-	RZ_FREE(debugger);
-}
-
 RZ_IPI RzCmdStatus rz_analyze_simple_handler(RzCore *core, int argc, const char **argv) {
-	core_perform_auto_analysis(core, CORE_ANALYSIS_SIMPLE);
+	rz_core_perform_auto_analysis(core, RZ_CORE_ANALYSIS_SIMPLE);
```
XVilka, line 5903: For this new function - `rz_core_perform_auto_analysis()` I recommend to create a new unit/integration test, that will try all levels of the analysis for some target and compare the results between them - e.g. that some functions are missing on SIMPLE level but present on more advanced. You can see these as an example:
- https://github.com/rizinorg/rizin/blob/dev/test/integration/test_open_analyse_save_load_project.c
- https://github.com/rizinorg/rizin/blob/dev/test/integration/test_analysis_global_var.c

No need to save/load project, just opening file multiple times for every different analysis will suffice.

librz/core/cmd/cmd_help.c
```diff
@@ -430,6 +430,104 @@ RZ_IPI void rz_core_clippy_print(RzCore *core, const char *msg) {
 	}
 }
 
+RZ_API void rz_core_cmd_help_calc_expr(RZ_NONNULL RzCore *core, RZ_NONNULL const char *input) {
```
XVilka: 1. No need for it to be and API, IPI will suffice.
2. I recommend to use rzshell `intc argc, char **argv`-style arguments, also with `RzCmdStateOutput` for using JSON and other kinds of output

librz/core/cmd/cmd_remote.c
```diff
@@ -253,3 +255,10 @@ RZ_IPI RzCmdStatus rz_remote_tcp_handler(RzCore *core, int argc, const char **ar
 	}
 	return RZ_CMD_STATUS_ERROR;
 }
+
+RZ_API void rz_core_rtr_enable(RZ_NONNULL RzCore *core, RZ_BORROW const char *cmdremote) {
```
XVilka: Maybe no need for it to be an API?

XVilka (changes requested):

librz/core/cmd/cmd_help.c
```diff
@@ -529,6 +435,7 @@ RZ_API void rz_core_cmd_help_calc_expr(RZ_NONNULL RzCore *core, RZ_NONNULL const
 }
 
 RZ_IPI int rz_cmd_help(void *data, const char *input) {
```
brightprogrammer: Left out commented part are commands that were not ported. For `@`, `@@` and `@@@` the generation script is generating wrong output so compiler shows compile error.

This function is kept not deleted itself because it's used in `cmd.c` @ line 5597. I tried commenting that out but compiler again shows some errors. I can figure that out later. For now it's just this. There are also some other minor changes left out. Waiting your reviews! :smile:

librz/core/cmd/cmd_help.c
```diff
@@ -805,297 +462,6 @@ RZ_IPI int rz_cmd_help(void *data, const char *input) {
...
-		ut64 start = rz_time_now_mono();
-		rz_core_cmd(core, input + 1, 0);
-		ut64 end = rz_time_now_mono();
-		double seconds = (double)(end - start) / RZ_USEC_PER_SEC;
-		core->num->value = (ut64)seconds;
-		rz_cons_printf("%lf\n", seconds);
-		break;
-	}
 	case '?': // "??"
 		if (input[1] == '?') {
```
brightprogrammer: Some of the commands like `??` work differently now. Previously `??x` would have executed without any errors but now it's `?? x` (notice the SPACE in between)

librz/core/canalysis.c
brightprogrammer: Moving declaration here shows some compiler warning for non initialization of `debugger`. I'll need some time on that because I was focusing on rzshell.

librz/core/cmd/cmd_help.c
```diff
@@ -690,7 +690,7 @@ RZ_IPI RzCmdStatus rz_print_boundaries_prot_handler(RzCore *core, int argc, cons
 }
 
 RZ_IPI RzCmdStatus rz_print_djb2_hash_handler(RzCore *core, int argc, const char **argv) {
-	for (int i = i; i < argc; i++) {
+	for (int i = 1; i < argc; i++) {
```
wargio: nice catch. this means there is no test for this.

test/db/tools/rz
```diff
@@ -74,7 +74,7 @@ FILE==
 CMDS=!!rizin -hh~RZ_USER_PLUGINS?
 EXPECT=<<EOF
 1
-EOF
+EO
```
wargio: this is wrong.
```suggestion
EOF
```

librz/core/cmd/cmd_help.c
```diff
@@ -857,32 +857,24 @@ RZ_IPI RzCmdStatus rz_compare_and_set_core_num_value_handler(RzCore *core, int a
...
 }
 
 RZ_IPI RzCmdStatus rz_exec_cmd_if_core_num_value_zero_handler(RzCore *core, int argc, const char **argv) {
-	if (argc == 1) {
-		rz_cons_printf("%" PFMT64d "\n", core->num->value);
-	} else {
-		if (core->num->value == 0) {
-			core->num->value = rz_core_cmd(core, argv[1], 0);
-		}
+	if (core->num->value == 0) {
```
wargio:
```suggestion
	if (!core->num->value) {
```

test/db/cmd/cmd_help
```diff
@@ -58,9 +58,9 @@ CMDS=<<EOF
 # ?X
 # ?l
 # ?t is used for get the execution time, thus shall be ignored.
-?!
-?+
-?-
+#?!
+#?+
+#?-
```
wargio: why commenting?

brightprogrammer: These commands print help when wrong arguments (or no arguments) are passed instead of just printing the `core->num->value`. We have `?v` command group for that.

test/db/tools/rz
brightprogrammer: This was fixed in last commits

librz/core/cmd/cmd_help.c
brightprogrammer: This command will then do the same thing as `?+` That's why I changed it to check for zero value.

test/db/cmd/cmd_help
brightprogrammer: These checks were there to test whether they print the `core->num->value` or not

librz/core/canalysis.c
wargio: but `char *debugger = NULL;` at line 6398, so it should not.

librz/core/cmd/cmd_help.c
wargio: doesn't matter. when checking for `if zero` or `if NULL` just do the comparison as `if (!variable)`

brightprogrammer: Ok

librz/core/cmd/cmd_help.c
```diff
@@ -434,63 +435,8 @@ RZ_API void rz_core_cmd_help_calc_expr(RZ_NONNULL RzCore *core, RZ_NONNULL const
...
-				rz_cons_printf("%" PFMT64d "\n", core->num->value);
-			}
-			break;
-		case '\0': // "?"
-		default:
-			break;
-		}
-		*/
+	// TODO: I don't know how to remove this function.
+	// It's called in cmd.c file in rz_core_cmd_init
```
wargio: open an issue for this

brightprogrammer: I tried removing it this time and it worked. Maybe it failed before because not all commands were ported.

librz/core/tui/panels.c
```diff
@@ -5188,7 +5196,8 @@ bool __handle_console(RzCore *core, RzPanel *panel, const int key) {
 		__panel_prompt(prompt, cmd, sizeof(cmd));
 		if (*cmd) {
 			if (!strcmp(cmd, "clear")) {
-				rz_core_cmd0(core, ":>$console");
+				// rz_core_cmd0(core, ":>$console");
```
XVilka: Remove

librz/main/rizin.c
```diff
@@ -1221,7 +1221,13 @@ RZ_API int rz_main_rizin(int argc, const char **argv) {
 			const char *fstype = r->bin->cur->o->info->bclass;
 			rz_core_cmdf(r, "m /root %s @ 0", fstype);
 		}
-		rz_core_cmd0(r, "R!"); // initalize io subsystem
+		// initalize io subsystem
+		char *res = rz_io_system(r->io, NULL);
+		if (res) {
+			rz_cons_printf("%s\n", res);
```
XVilka: Could just use `rz_cons_println(res)`

test/db/cmd/feat_input
```diff
@@ -78,7 +78,7 @@ RUN
 
 NAME=at3 "?vi $$"@3;"?e bar"
 FILE=malloc://1024
-CMDS="?vi $$"@ 3;"?e bar"
+CMDS=?vi $$ @ 3; ?e bar
```
XVilka: Please change the name of the tests as well, to reflect the changes

test/integration/test_auto_analysis.c
```diff
@@ -0,0 +1,58 @@
...
+	mu_assert("deep analysis success", deep_fcn_list_length >= simple_fcn_list_length);
+
+	// aaa
+	rz_core_perform_auto_analysis(core, RZ_CORE_ANALYSIS_EXPERIMENTAL);
+	RzList *experimental_fcn_list = rz_analysis_function_list(core->analysis);
+	mu_assert_notnull(experimental_fcn_list, "simple function list");
+	ut32 experimental_fcn_list_length = rz_list_length(experimental_fcn_list);
+	eprintf("functions count = %d\n", experimental_fcn_list_length);
+
+	// again more in experimental
```
XVilka, line 54: I recommend also checking for some particular functions presence, not just simple count comparison.
Moreover, I recommend adding also ARM binary as a second test here, then you could extract the shared code in the static function and run over multiple different files.

test/integration/test_auto_analysis.c
```diff
@@ -0,0 +1,58 @@
...
+	RzList *deep_fcn_list = rz_analysis_function_list(core->analysis);
+	mu_assert_notnull(deep_fcn_list, "simple function list");
+	ut32 deep_fcn_list_length = rz_list_length(deep_fcn_list);
+	eprintf("functions count = %d\n", deep_fcn_list_length);
+
+	// more function should be detected in deep analysis
+	// not true for all binaries but for most it must be true
+	mu_assert("deep analysis success", deep_fcn_list_length >= simple_fcn_list_length);
+
+	// aaa
```
XVilka: `aaaa`

librz/include/rz_core.h
```diff
@@ -680,6 +680,12 @@ typedef struct rz_core_analysis_name_t {
 	ut64 offset;
 } RzCoreAnalysisName;
 
+typedef enum {
+	RZ_CORE_ANALYSIS_SIMPLE, ///< aa
```
XVilka, line 689: It needs better description, at least some context

librz/include/rz_core.h
XVilka: Yes, please remove it

librz/core/canalysis.c
XVilka: Don't forget this please

librz/core/cmd/cmd_help.c
XVilka: Just make it `IPI` for this PR, it can be converted later then.

XVilka (changes requested):

librz/core/canalysis.c
```diff
@@ -6384,3 +6384,41 @@ RZ_API ut64 rz_core_prevop_addr_force(RzCore *core, ut64 start_addr, int numinst
...
+	if (core->io && core->io->desc && core->io->desc->plugin && !core->io->desc->plugin->isdbg) {
+		// set it only if is debugging
+		RZ_FREE(debugger);
+	}
+	rz_cons_clear_line(1);
+	rz_core_analysis_everything(core, type == RZ_CORE_ANALYSIS_EXPERIMENTAL, debugger);
```
wargio, line 6417:
```suggestion
	if (!(core->io && core->io->desc && core->io->desc->plugin && !core->io->desc->plugin->isdbg)) {
		// set it only if is debugging
		debugger = core->dbg->cur ? strdup(core->dbg->cur->name) : strdup("esil");
	}
	rz_cons_clear_line(1);
	rz_core_analysis_everything(core, type == RZ_CORE_ANALYSIS_EXPERIMENTAL, debugger);
```

librz/core/canalysis.c
```diff
@@ -6384,3 +6384,41 @@ RZ_API ut64 rz_core_prevop_addr_force(RzCore *core, ut64 start_addr, int numinst
...
+	rz_core_task_yield(&core->tasks);
+	if (type == RZ_CORE_ANALYSIS_SIMPLE || rz_cons_is_breaked()) {
+		goto finish;
+	}
+	// Run pending analysis immediately after analysis
+	// Usefull when running commands with ";" or via rizin -c,-i
+	debugger = core->dbg->cur ? strdup(core->dbg->cur->name) : strdup("esil");
+	if (core->io && core->io->desc && core->io->desc->plugin && !core->io->desc->plugin->isdbg) {
+		// set it only if is debugging
+		RZ_FREE(debugger);
```
wargio:
```suggestion
	if (core->io && core->io->desc && core->io->desc->plugin && core->io->desc->plugin->isdbg) {
		// set it only if is debugging
		debugger = core->dbg->cur ? strdup(core->dbg->cur->name) : strdup("esil");
```

librz/core/cmd/cmd_help.c
wargio: set this as `RZ_IPI` and move the header definition under `core_private.h`

librz/core/cmd/cmd_help.c
```diff
@@ -430,696 +336,764 @@ RZ_IPI void rz_core_clippy_print(RzCore *core, const char *msg) {
...
+}
+
+RZ_IPI RzCmdStatus rz_print_ascii_table_handler(RzCore *core, int argc, const char **argv) {
+	rz_cons_printf("%s", ret_ascii_table());
+	return RZ_CMD_STATUS_OK;
+}
+
+RZ_IPI RzCmdStatus rz_print_binary_handler(RzCore *core, int argc, const char **argv) {
+	char out[128] = RZ_EMPTY;
+	ut64 n;
```
XVilka: fold these two lines into one

librz/core/cmd/cmd_help.c
```diff
@@ -430,696 +336,764 @@ RZ_IPI void rz_core_clippy_print(RzCore *core, const char *msg) {
...
+	return RZ_CMD_STATUS_OK;
+}
+
+RZ_IPI RzCmdStatus rz_flush_console_handler(RzCore *core, int argc, const char **argv) {
+	rz_cons_flush();
+	return RZ_CMD_STATUS_OK;
+}
+
+RZ_IPI RzCmdStatus rz_print_bitstring_handler(RzCore *core, int argc, const char **argv) {
+	ut64 n = 0;
```
XVilka: Ssame, fold

librz/core/cmd/cmd_help.c
```diff
@@ -430,696 +336,764 @@ RZ_IPI void rz_core_clippy_print(RzCore *core, const char *msg) {
...
+		RZ_LOG_ERROR("core: RzNum ERROR: Division by Zero\n");
+		return RZ_CMD_STATUS_ERROR;
+	}
+	rz_num_math(core->num, argv[1]);
+	return RZ_CMD_STATUS_OK;
+}
+
+RZ_IPI RzCmdStatus rz_show_value_handler(RzCore *core, int argc, const char **argv) {
+	ut64 n;
+	if (argc == 1) {
```
XVilka: You can use `ut64 n = argc == 1 ? core->num->value : rz_num_math()`

librz/core/cmd/cmd_help.c
```diff
@@ -430,696 +336,764 @@ RZ_IPI void rz_core_clippy_print(RzCore *core, const char *msg) {
...
+		RZ_LOG_ERROR("core: RzNum ERROR: Division by Zero\n");
+		return RZ_CMD_STATUS_ERROR;
+	}
+	rz_cons_printf("0x%" PFMT64x "\n", n);
+	return RZ_CMD_STATUS_OK;
+}
+
+RZ_IPI RzCmdStatus rz_show_value_hex_handler(RzCore *core, int argc, const char **argv) {
+	ut64 n;
+	if (argc == 1) {
```
XVilka: Same

librz/core/cmd/cmd_help.c
XVilka: Though I would use `argc < 2` to show the logic better instead of direct comparison.

librz/core/cmd/cmd_help.c
```diff
@@ -430,696 +336,764 @@ RZ_IPI void rz_core_clippy_print(RzCore *core, const char *msg) {
...
+		RZ_LOG_ERROR("core: RzNum ERROR: Division by Zero\n");
+		return RZ_CMD_STATUS_ERROR;
+	}
+	rz_cons_printf("0x%08" PFMT64x "\n", n); // differs from ?v here 0x%08
+	return RZ_CMD_STATUS_OK;
+}
+
+RZ_IPI RzCmdStatus rz_show_value_i1_handler(RzCore *core, int argc, const char **argv) {
+	ut64 n;
+	if (argc == 1) {
```
XVilka: Same, etc

librz/core/cmd/cmd_help.c
```diff
@@ -430,696 +336,764 @@ RZ_IPI void rz_core_clippy_print(RzCore *core, const char *msg) {
...
+}
+
+RZ_IPI RzCmdStatus rz_exec_cmd_if_core_num_value_zero_handler(RzCore *core, int argc, const char **argv) {
+	if (!core->num->value) {
+		core->num->value = rz_core_cmd(core, argv[1], 0);
+	}
+	return RZ_CMD_STATUS_OK;
+}
+
+/* RZ_IPI RzCmdStatus rz_show_help_at_handler(RzCore *core, int argc, const char **argv) { */
```
XVilka: Remove

librz/core/cmd/cmd_help.c
```diff
@@ -430,696 +336,764 @@ RZ_IPI void rz_core_clippy_print(RzCore *core, const char *msg) {
...
+	return RZ_CMD_STATUS_OK;
+}
+
+RZ_IPI RzCmdStatus rz_clippy_echo_handler(RzCore *core, int argc, const char **argv) {
+	rz_core_clippy_print(core, argv[1]);
+	return RZ_CMD_STATUS_OK;
+}
+
+RZ_IPI RzCmdStatus rz_echo_msg_newline_handler(RzCore *core, int argc, const char **argv) {
+	if (argc != 1) {
```
XVilka: `argc > 1` probably better

librz/core/cmd/cmd_help.c
```diff
@@ -430,696 +336,764 @@ RZ_IPI void rz_core_clippy_print(RzCore *core, const char *msg) {
...
+	}
+
+	for (; from <= to; from += step)
+		rz_cons_printf("%" PFMT64d " ", from);
+	rz_cons_newline();
+	return RZ_CMD_STATUS_OK;
+}
+
+RZ_IPI RzCmdStatus rz_phys2virt_handler(RzCore *core, int argc, const char **argv) {
+	ut64 n = argc == 2 ? rz_num_math(core->num, argv[1]) : core->offset;
```
XVilka: Just like you did here.

test/db/cmd/cmd_help
```diff
@@ -47,34 +47,25 @@ RUN
 FILE=malloc://1024
 CMDS=<<EOF
 ?v
-?=
-??
+# ?=
```
XVilka: So what do we do with those? Remove them maybe? They aren't valid commands in the rzshell anyways, since require additional argument, I suppose?

XVilka (changes requested):
Overall is good, see few minor changes necessary though.

librz/core/cmd/cmd_help.c
brightprogrammer: should I write test for this?

librz/core/cmd/cmd_analysis.c
XVilka: Could you make it `RZ_IPI`?

librz/core/cmd/cmd_help.c
```diff
@@ -430,696 +336,722 @@ RZ_IPI void rz_core_clippy_print(RzCore *core, const char *msg) {
...
+		core->curtab = 0;
+	}
+	core->curtab++;
+	return RZ_CMD_STATUS_OK;
+}
+
+RZ_IPI RzCmdStatus rz_generate_random_number_handler(RzCore *core, int argc, const char **argv) {
+	// TODO : Add support for 64bit random numbers
+	ut64 b = 0;
+	ut32 r = UT32_MAX;
```
XVilka: Could be folded with the following 3 lines as well using `?:` operator.

librz/core/cmd/cmd_help.c
XVilka: Would be nice, yes.

test/db/cmd/cmd_help
XVilka: Don't forget these

XVilka (changes requested):
Few last nitpicks, apart from that - LGTM. Please also rebase on top of the latest `dev`.

librz/core/cmd/cmd_analysis.c
brightprogrammer: this is used in `librz/main` module
image

XVilka (approved):
From the code PoV looks good. @ret2libc please check if all commands in YAML make sense.

ret2libc (changes requested):
There is a problem with this PR for now: How do you list all commands under `?`? So, I remember we discussed whether we wanted to keep the `?` prefix for these commands or actually use something else, considering that `?` is already used as the suffix to list sub-commands (e.g. `a?` list sub-commands of `a`).

@thestr4ng3r @XVilka @wargio @kazarmy @ITAYC0HEN WDYT?

brightprogrammer:
There was a command ??? To show help of all ? based commands but I had to remove it because someone requested it...

test/integration/test_auto_analysis.c
```diff
@@ -0,0 +1,130 @@
...
+	{"bins/arm/elf/hello-world-linaro-201201", "fcn.000082a4", RZ_CORE_ANALYSIS_DEEP},
+	{"bins/arm/elf/hello-world-linaro-201201", "fcn.000083bc", RZ_CORE_ANALYSIS_EXPERIMENTAL},
+
+	{"bins/elf/ls", "fcn.00016530", RZ_CORE_ANALYSIS_DEEP},
+	{"bins/elf/ls", "fcn.0000f1b0", RZ_CORE_ANALYSIS_DEEP},
+	{"bins/elf/ls", "fcn.00010be0", RZ_CORE_ANALYSIS_DEEP},
+	{"bins/elf/ls", "fcn.00016320", RZ_CORE_ANALYSIS_DEEP},
+};
+
+const ut32 bin_fcn_list_size = sizeof(bin_fcn_list)/sizeof(bin_fcn_list[0]);
```
ret2libc: Use `RZ_ARRAY_SIZE`

test/integration/test_auto_analysis.c
```diff
@@ -0,0 +1,130 @@
...
+
+bool test_fcn_list_size() {
+	for(ut32 i = 0; i < bin_fcn_list_size; ++i) {
+		test_detected_functions_list_size(bin_fcn_list[i].fpath);
+	}
+	mu_end;
+}
+
+bool test_new_fcn_detected() {
+	for(ut32 i = 0; i < bin_fcn_list_size; ++i) {
```
ret2libc, line 167: this is quite slow as it re-analyze the same binary multiple times for each function, but ok for now 👍

librz/core/cmd_descs/cmd_help.yaml
```diff
@@ -0,0 +1,559 @@
...
+      - name: Examples
+        entries:
+          - text: "?f"
+            arg_str: " 28 Hello"
+            comment: "LLO (00111b selected : big-endian bitstring)"
+          - text: "?f"
+            arg_str: " 12 Hello"
+            comment: "LL (00110b selected : big-endian bitstring)"
+  - name: "?o"
+    summary: Evalue expression and print value in octal.
```
ret2libc:
```suggestion
    summary: Evaluate expression and print value in octal.
```

librz/core/cmd_descs/cmd_help.yaml
```diff
@@ -0,0 +1,559 @@
...
+        cname: compute_string_length
+        args:
+          - name: str
+            type: RZ_CMD_ARG_TYPE_STRING
+      - name: "?lq"
+        summary: Calculate length of given string and store in $? register but print nothing (quietly).
+        cname: compute_string_length_quiet
+        args:
+          - name: str
+            type: RZ_CMD_ARG_TYPE_STRING
```
ret2libc: for these kinds of commands you can use something like https://github.com/rizinorg/rizin/blob/dev/librz/core/cmd_descs/cmd_write.yaml#L318 . You just have to specify both the STANDARD and QUIET mode and then implement them in the same handler (like https://github.com/rizinorg/rizin/blob/dev/librz/core/cio.c#L802).

test/integration/test_auto_analysis.c
brightprogrammer: how can I improve this test?

XVilka:
> There is a problem with this PR for now: How do you list all commands under `?`? So, I remember we discussed whether we wanted to keep the `?` prefix for these commands or actually use something else, considering that `?` is already used as the suffix to list sub-commands (e.g. `a?` list sub-commands of `a`).
> 
> @thestr4ng3r @XVilka @wargio @kazarmy @ITAYC0HEN WDYT?

I propose reusing `%` for these math evaluation commands. Currently `%` is used to access and set environment variables which is not so common use case, unlike evaluating something, also there is already `env` command for that. Thus, removing `%` for environment variables and using it for math expressions.

Moreover, `%` character is somewhat loosely related to the "mathematics" and "evaluation" semantically.

For commands like `?e` we can remove them, since there is already `echo`. If some non-math and non-evaluation command is there - it's better to move to the `shell` category, e.g. `time` instead of `?t`

test/integration/test_auto_analysis.c
XVilka: You could just make this array two-dimensional. One dimension are files, another dimension are functions/analysis levels.

wargio:
i'm ok with `%`, but it is used currently for environment vars.
i propose instead `^` which is not used

brightprogrammer:
If we eventually need to move non mathematical commands out of this group, then I guess what ret2libc suggested is a good idea. We can use `m` for mathematical expression and `m` isn't used. Then other commands can be moved to a new misc command group by `^` or `%` or any other symbol not in use.

But I guess that should be done in another PR because that'd mean changing tests too.

ret2libc:
> If we eventually need to move non mathematical commands out of this group, then I guess what ret2libc suggested is a good idea. We can use `m` for mathematical expression and `m` isn't used. Then other commands can be moved to a new misc command group by `^` or `%` or any other symbol not in use.
> 
> But I guess that should be done in another PR because that'd mean changing tests too.

The problem is that right now this can't be merged with `?` because `???` does not work.

XVilka:
```
pylint: Command line or configuration file:1: UserWarning: Specifying exception names in the overgeneral-exceptions option without module name is deprecated and support for it will be removed in pylint 3.0. Use fully qualified name (maybe 'builtins.BaseException' ?) instead.
pylint: Command line or configuration file:1: UserWarning: Specifying exception names in the overgeneral-exceptions option without module name is deprecated and support for it will be removed in pylint 3.0. Use fully qualified name (maybe 'builtins.Exception' ?) instead.
************* Module cmd_descs_generate
librz/core/cmd_descs/cmd_descs_generate.py:188:8: W0[7](https://github.com/rizinorg/rizin/actions/runs/4546398351/jobs/8020421131?pr=3421#step:5:8)19: Raising too general exception: Exception (broad-exception-raised)
librz/core/cmd_descs/cmd_descs_generate.py:731:[8](https://github.com/rizinorg/rizin/actions/runs/4546398351/jobs/8020421131?pr=3421#step:5:9): W071[9](https://github.com/rizinorg/rizin/actions/runs/4546398351/jobs/8020421131?pr=3421#step:5:10): Raising too general exception: Exception (broad-exception-raised)
************* Module fuzz_rz_asm
test/fuzz/scripts/fuzz_rz_asm.py:85:[11](https://github.com/rizinorg/rizin/actions/runs/4546398351/jobs/8020421131?pr=3421#step:5:12): R1735: Consider using '{"cause": cause, "ins": ins, "inpairs": inpairs, "oins": oins, "case": inskey, ... }' instead of a call to 'dict'. (use-dict-literal)

-----------------------------------
```

brightprogrammer:
> ```
> pylint: Command line or configuration file:1: UserWarning: Specifying exception names in the overgeneral-exceptions option without module name is deprecated and support for it will be removed in pylint 3.0. Use fully qualified name (maybe 'builtins.BaseException' ?) instead.
> pylint: Command line or configuration file:1: UserWarning: Specifying exception names in the overgeneral-exceptions option without module name is deprecated and support for it will be removed in pylint 3.0. Use fully qualified name (maybe 'builtins.Exception' ?) instead.
> ************* Module cmd_descs_generate
> librz/core/cmd_descs/cmd_descs_generate.py:188:8: W0[7](https://github.com/rizinorg/rizin/actions/runs/4546398351/jobs/8020421131?pr=3421#step:5:8)19: Raising too general exception: Exception (broad-exception-raised)
> librz/core/cmd_descs/cmd_descs_generate.py:731:[8](https://github.com/rizinorg/rizin/actions/runs/4546398351/jobs/8020421131?pr=3421#step:5:9): W071[9](https://github.com/rizinorg/rizin/actions/runs/4546398351/jobs/8020421131?pr=3421#step:5:10): Raising too general exception: Exception (broad-exception-raised)
> ************* Module fuzz_rz_asm
> test/fuzz/scripts/fuzz_rz_asm.py:85:[11](https://github.com/rizinorg/rizin/actions/runs/4546398351/jobs/8020421131?pr=3421#step:5:12): R1735: Consider using '{"cause": cause, "ins": ins, "inpairs": inpairs, "oins": oins, "case": inskey, ... }' instead of a call to 'dict'. (use-dict-literal)
> 
> -----------------------------------
> ```

How do I fix this?

XVilka:
@brightprogrammer:
1. Make a commit fixing these files (you can run `pylint` and `prettier` locally)
2. Push it here, if everything is green - send a separate PR with those fixes, wait for it to pass all tests, we will merge if everything is fine
3. Rebase this PR

test/db/tools/rz_ax
```diff
@@ -354,7 +354,7 @@ RUN
 
 NAME=Python string '\x33\x44\x88'
 FILE==
-CMDS=echo "foo = '\x33\x44\x88'" | rz-ax -F
+CMDS=echo foo = "'\x33\x44\x88'" | rz-ax -F
```
ret2libc, line 357: is this required? I thought `echo "foo = '\x33\x44\x88'"` would work too.

test/db/esil/x86_32
```diff
@@ -188,9 +188,9 @@ aes
 .arf
 s 0x10000000
 cX 0x20000000
-ar?edi
-ar?esi
-ar?ecx
+arechodi
+arechosi
+arechocx
```
ret2libc: wrong replacement

test/db/cmd/feat_grep
```diff
@@ -165,10 +165,10 @@ i~:..3
...
 i~:-2
 i~:-1
-?e 0\n1\n2\n3\n4~:-2
-?e 0\n1\n2\n3\n4~:-2..
-?e 0\n1\n2~:0..
-?e 0\n1\n2~:..
+echo "0\n1\n2\n3\n4"~:-2
+echo "0\n1\n2\n3\n4"~:-2..
+echo "0\n1\n2"~:0..
+echo "0\n1\n2"~:..
```
ret2libc: TODO: check if `echo` unescape characters like `?e` did.

librz/core/cmd_descs/cmd_shell.yaml
```diff
@@ -109,9 +109,13 @@ commands:
...
     cname: cmd_shell_echo
     args:
-      - name: argument
+      - name: str
+        type: RZ_CMD_ARG_TYPE_STRING
+        optional: true
+      - name: strs
         type: RZ_CMD_ARG_TYPE_STRING
         flags: RZ_CMD_ARG_FLAG_ARRAY
+        optional: true
```
ret2libc: I think you can just use one single argument named `strs` with FLAG_ARRAY and optional:true.

librz/core/cmd_descs/cmd_shell.yaml
```diff
@@ -187,3 +191,45 @@ commands:
...
+          - RZ_OUTPUT_MODE_STANDARD
+          - RZ_OUTPUT_MODE_QUIET
+      - name: vernum
+        summary: Show numeric version
+        cname: show_version_numeric
+        args: []
+      - name: verjson
+        summary: Show version info in JSON format
+        cname: show_version_json
+        args: []
```
ret2libc: remove this and add RZ_OUTPUT_MODE_JSON in `ver` command

librz/core/cmd/cmd_print.c
```diff
@@ -732,10 +733,12 @@ RZ_API void rz_core_set_asm_configs(RzCore *core, char *arch, ut32 bits, int seg
 	rz_config_set_i(core->config, "asm.segoff", segoff);
 }
 
-static void helpCmdTasks(RzCore *core) {
-	// TODO: integrate with =h& and bg analysis/string/searches/..
-	rz_core_cmd_help(core, help_msg_amper);
-}
+/* unused so compiler shows warning */
```
ret2libc: remove if unused.

librz/core/cmd/cmd_print.c
```diff
@@ -14,21 +14,22 @@
...
-	"&=", " 3", "show output of task 3",
-	"&b", " 3", "break task 3",
-	"&-", " 1", "delete task #1 or schedule for deletion when it is finished",
-	"&", "-*", "delete all done tasks",
-	"&?", "", "show this help",
-	"&&", " 3", "wait until task 3 is finished",
-	"&&", "", "wait until all tasks are finished",
-	NULL
-};
+/* used by helpCmdTasks function */
```
ret2libc: remove if unused.

test/db/json/json1
```diff
@@ -7,7 +7,7 @@
 /Rj
 /wj
 ?j 0x10
-?Vj
+%vj
```
ret2libc:
```suggestion
verj
```

librz/core/cmd/cmd_math.c
```diff
@@ -0,0 +1,806 @@
...
+
+RZ_IPI void rz_core_clippy_print(RzCore *core, const char *msg) {
+	char *string = rz_core_clippy(core, msg);
+	if (string) {
+		rz_cons_print(string);
+		free(string);
+	}
+}
+
+/* keeping these functions because they can be useful in future */
```
ret2libc: remove if unused

librz/core/cmd/cmd_math.c
```diff
@@ -0,0 +1,806 @@
...
+}
+
+RZ_IPI RzCmdStatus rz_exec_cmd_if_core_num_value_positive2_handler(RzCore *core, int argc, const char **argv) {
+	if (core->num->value) {
+		core->num->value = rz_core_cmd(core, argv[1], 0);
+	}
+	return RZ_CMD_STATUS_OK;
+}
+
+RZ_IPI RzCmdStatus rz_help_handler(RzCore *core, int argc, const char **argv) {
```
ret2libc: move this in `cmd.c`

subprojects/rizin-shell-parser/grammar.js
ret2libc, line 1: you changed all the style in this file and it becomes hard for me to review and know what actually changed :( please re-push with just the actual changes you made

ret2libc (commented):
Terrific work! That's a big change and a very useful one, thanks!

Just few things missing to adjust, but overall things look good, good job!

test/db/tools/rz_ax
brightprogrammer: It makes more sense to be like this. There are two python tests for different python string syntax.
- `"string"` which is tested by `'"string"'`
- `'string'` which is tested by `"'string'"`

test/db/cmd/feat_grep
brightprogrammer: ~~No that doesn't work, because I replaced the old echo code with code in `?e` command in order to be able to replace flags in string like the `?e` command did. Only then I could have removed `?e` command group and kept from failing lots of regression tests.~~

Actually that does work! Will change all places where it'll work!

subprojects/rizin-shell-parser/grammar.js
brightprogrammer: I ran prettier on this :sweat_smile: Which formatter do you use?
Changes made : 
- `math_arged_stmt` used `?` as identifier for starting of command. Now we have a choice between `%` (`percent_mark_identifier`) and `math_mark_identifier` (a regex) and all of these accept optional arguments (don't know whether it should be like that but if it's working then what can we say :smile:)
```js
    percent_mark_identifier: ($) => "%",
    percent_percent_mark_identifier: ($) => "%%",
    math_mark_identifier: ($) => /%[a-zA-Z0-9]*[!@#$&\-.+=]*/,
    _math_arged_stmt: ($) =>
      prec.left(
        1,
        choice(
          seq(
            field("command", $.percent_mark_identifier),
            field("args", optional($.args))
          ),
          seq(
            field("command", $.percent_percent_mark_identifier),
            field("args", optional($.args))
          ),
          seq(
            field("command", alias($.math_mark_identifier, $.cmd_identifier)),
            field("args", optional($.args))
          )
        )
      ),
```
- `_env_stmt_identifier` used a choice between `%` or `env`. Now it's just the `env` identifier.
```js
_env_stmt_identifier: ($) => "env",
```

Sorry for all the trouble :sweat_smile:

brightprogrammer: I also changed the `simple_commands.txt` and `special_commands.txt`. You should check in `simple_commands.txt` for changed tests.

test/db/cmd/feat_grep
ret2libc: Sorry, what do you mean? which places are you going to replace?

brightprogrammer: I mean I'll remove `"` where it's present with the `echo` command and not needed (as your TODO states)

XVilka:
```
../librz/core/cmd/cmd_shell.c:379: error: identifier expected
```
```
librz/core/cmd/cmd_shell.c:357:45: error: code should be clang-formatted [-Wclang-format-violations]
        rz_cons_printf("plug.init = %" PFMT64d "\n"
                                                   ^
librz/core/cmd/cmd_shell.c:358:36: error: code should be clang-formatted [-Wclang-format-violations]
                                   "plug.load = %" PFMT64d "\n"
                                                               ^
librz/core/cmd/cmd_shell.c:377:8: error: code should be clang-formatted [-Wclang-format-violations]
        switch(state->mode) {
              ^
librz/core/cmd/cmd_shell.c:387:5: error: code should be clang-formatted [-Wclang-format-violations]
                PJ* pj = state->d.pj;
                  ^
librz/core/cmd/cmd_shell.c:387:6: error: code should be clang-formatted [-Wclang-format-violations]
                PJ* pj = state->d.pj;
                   ^
```

librz/core/cmd/cmd_print.c
XVilka: @brightprogrammer you forgot to remove these

librz/core/cmd/cmd_math.c
```diff
@@ -0,0 +1,597 @@
...
+		if (wideOffsets) {
+			rz_cons_printf("%s %s 0x%016" PFMT64x "\n", vars[i], pad, rz_num_math(core->num, vars[i]));
+		} else {
+			rz_cons_printf("%s %s 0x%08" PFMT64x "\n", vars[i], pad, rz_num_math(core->num, vars[i]));
+		}
+		i++;
+	}
+}
+
+/// If third parameter (PJ) is NULL then standard output will be displayed
```
XVilka: Convert into Doxygen

librz/core/cmd/cmd_math.c
```diff
@@ -0,0 +1,597 @@
...
+/// If it's non-NULL then it's treated as a borrowd PJ and used and then
+/// printing will be done in JSON format and not standard one.
+/// So basically it's the third paramter that decides in which paramter
```
XVilka: parameter

librz/core/cmd/cmd_math.c
```diff
@@ -0,0 +1,597 @@
...
+			rz_cons_printf("%s %s 0x%016" PFMT64x "\n", vars[i], pad, rz_num_math(core->num, vars[i]));
+		} else {
+			rz_cons_printf("%s %s 0x%08" PFMT64x "\n", vars[i], pad, rz_num_math(core->num, vars[i]));
+		}
+		i++;
+	}
+}
+
+/// If third parameter (PJ) is NULL then standard output will be displayed
+/// If it's non-NULL then it's treated as a borrowd PJ and used and then
```
XVilka: borrowed

librz/core/cmd_descs/cmd_descs.h
```diff
@@ -4,11 +4,9 @@
 // WARNING: This file was auto-generated by cmd_descs_generate.py script. Do not
 // modify it manually. Look at cmd_descs.yaml if you want to update commands.
 //
-
```
XVilka: This change is completely unnecessary, caused by the removal of some empty strings in the python script below

XVilka (changes requested):
A few more things left, before it's ready to be merged

librz/core/cmd_descs/cmd_descs_generate.py
```diff
@@ -47,14 +44,11 @@
 // WARNING: This file was auto-generated by cmd_descs_generate.py script. Do not
 // modify it manually. Look at cmd_descs.yaml if you want to update commands.
 //
-
```
XVilka: You forgot this part - otherwise generated header/C file will be out-of-sync

XVilka (changes requested):

librz/core/cmd/cmd_math.c
```diff
@@ -0,0 +1,603 @@
...
+			rz_cons_printf("%s %s 0x%016" PFMT64x "\n", vars[i], pad, rz_num_math(core->num, vars[i]));
+		} else {
+			rz_cons_printf("%s %s 0x%08" PFMT64x "\n", vars[i], pad, rz_num_math(core->num, vars[i]));
+		}
+		i++;
+	}
+}
+
+/**
+ * \brief If @p pj is NULL then standard output will be displayed
```
XVilka: `@p` -> `\p`

librz/core/cmd/cmd_math.c
```diff
@@ -0,0 +1,603 @@
...
+ * If it's non-NULL then it's treated as a borrowed PJ and used and
+ * printing will be done in JSON format and not standard one.
+ * Hence the the third parameter decides in which format data will be displayed.
+ *
+ * \param core RzCore*.
```
XVilka: Just `RzCore`, no need to add pointer character

librz/core/cmd/cmd_math.c
```diff
@@ -0,0 +1,603 @@
...
+ * \param input Input mathematical expression.
+ * \param pj Borrowed PJ* if calculated value is to be printed in JSON format.
```
XVilka: Same

librz/core/cmd/cmd_math.c
```diff
@@ -0,0 +1,603 @@
...
+ * \return true on success, false otherwise.
+ * */
```
XVilka: Unnecessary space between `* *`

XVilka (changes requested):

XVilka (approved):

librz/core/cmd_descs/cmd_descs_generate.py
```diff
@@ -29,9 +29,7 @@
 //
 
 #include <cmd_descs.h>
-
```
ret2libc: keep these, they don't hurt and they help splitting the output.

librz/core/cmd_descs/cmd_descs_generate.py
```diff
@@ -29,9 +29,7 @@
...
 {helps_declarations}
-
```
ret2libc: also this

librz/core/cmd_descs/cmd_descs_generate.py
```diff
@@ -54,7 +52,6 @@
 
 // Command handlers, manually defined somewhere else
 {handlers_declarations}
-
```
ret2libc: and this

XVilka (approved):

ret2libc (approved):
Few wrong help messages (e.g. missing spaces) in `%?` but that's fine, let's do separate PR. Good job! :)

PR #3425, opened 2023-03-19
Add analysis for Objective-C selector stubs + arm64_32 support

thestr4ng3r:
# DO NOT SQUASH

**Detailed description**

Add analysis for Objective-C selector stubs

aalo is renamed to aalor to make room for the new aalos command, which
analyzes the __objc_stubs section and assigns names to all detected
selector stubs there. The stub patterns currently supported are all that
are hardcoded in ld64 version 820.
This analysis is also run automatically in aaa if the binary is detected
to potentially having such stubs.
To cleanly rename functions which have been created previously at
unnamed stub locations, rz_core_analysis_function_rename() has been
enhanced to not leave behind flags called fcn.0x<offset>.

Add support for DYLD_CHAINED_PTR_32

arm64_32 is AArch64, but using 32bit pointers, which is primarily used
on watchOS. Handling DYLD_CHAINED_PTR_32 fixups is necessary to
correctly load such binaries.

**Test plan**

See the added tests. Stripped binaries that have these stubs should get nice function names for them after aaa.


librz/bin/format/mach0/mach0_defines.h
```diff
@@ -1663,6 +1663,33 @@ static inline void dyld_chained_ptr_arm64e_auth_bind24_read(struct dyld_chained_
...
+	READ_BITS(dst->next, 5);
+	READ_BITS(dst->bind, 1);
+}
+
+struct dyld_chained_ptr_32_bind {
+	ut32 ordinal : 20,
+		addend : 6,
+		next : 5,
+		bind : 1; // == 1
+};
```
wargio, line 1684: isn't better if we avoid these kinda of structures with bitfields since they are not compatible with BE archs?

thestr4ng3r: This is only an issue if they are plainly copied over. The `dyld_chained_ptr_32_bind_read()` works on be too (see s390x test success).
Still keeping the structure itself as a bitfield is useful to have a reference for the original layout when reading the code.

librz/core/analysis_objc.c
ret2libc, line 1: do you have links/references from where you took these stubs from?

thestr4ng3r: I took them from the current ld64 binary in Xcode 14. Unfortunately the source for that version is not published yet, but it should show up soon here: https://opensource.apple.com/releases/ under Developer Tools 14.
But the contents of these patterns can also be seen when just compiling binaries with the right options with ld64-820.1.

wargio (approved):

PR #3428, opened 2023-03-21
Fix few sscanf issues reported by @yeggor

ret2libc:

**Detailed description**

`scanf("%s", buf)` is not safe, limit the number of characters in strings.

**Test plan**

...

**Closing issues**

...


wargio:
i think also the scanf in `librz/asm/arch/sh/assembler.c` are vulnerable

XVilka (approved):

ret2libc:
> i think also the scanf in `librz/asm/arch/sh/assembler.c` are vulnerable

I'm not sure, those seems to do a weird `var=strdup(param)` and then `sscanf(param, "%s", var)`, thus I think the array has the right size. That said, it should/can be improved. However, at this point, I believe it's better if we make rz-bindgen detect these kind of things across the whole codebase and make it a requirement for CI.

wargio (approved):

PR #3442, opened 2023-03-31
Remove TCP_NODELAY

wargio:

**Detailed description**

...

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/librz/socket/socket.c
+++ b/librz/socket/socket.c
@@ -334,8 +334,6 @@ RZ_API bool rz_socket_connect(RzSocket *s, const char *host, const char *port, i
 			return false;
 		}
 		for (rp = res; rp != NULL; rp = rp->ai_next) {
-			int flag = 1;
-
 			s->fd = socket(rp->ai_family, rp->ai_socktype, rp->ai_protocol);
 			if (s->fd == -1) {
 				perror("socket");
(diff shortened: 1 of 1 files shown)


XVilka (commented):
What is the reason behind this change? 

https://holmeshe.me/network-essentials-setsockopt-TCP_NODELAY/

Is it this?
> Use TCP_QUICKACK, not TCP_NODELAY
>
>Turning on TCP_NODELAY has similar effects, but can make throughput worse for small writes. If you write a loop which sends just a few bytes (worst case, one byte) to a socket with "write()", and the Nagle algorithm is disabled with TCP_NODELAY, each write becomes one IP packet. This increases traffic by a factor of 40, with IP and TCP headers for each payload. Tinygram prevention won't let you send a second packet if you have one in flight, unless you have enough data to fill the maximum sized packet. It accumulates bytes for one round trip time, then sends everything in the queue. That's almost always what you want. If you have TCP_NODELAY set, you need to be much more aware of buffering and flushing issues. None of this matters for bulk one-way transfers, which is most HTTP today. (I've never looked at the impact of this on the SSL handshake, where it might matter.) Short version: set TCP_QUICKACK.

XVilka:
Please use `dist-` branch for such PRs btw, to check all platforms aren't broken.

wargio:
because it is a bad idea. i have a huge amount of fail delivery on my own internal network. let the packages be handled by the OS.
this fixed all the issues i had.

wargio:
https://github.com/rizinorg/rizin/pull/3443

PR #3444, opened 2023-03-31
build: add "extra" prefix to load plugins from

ret2libc:

**Detailed description**

Homebrew on MacOS would like to load plugins (and maybe other things) from a separate "system" prefix because of how they configure packages with the actual package being in `/opt/homebrew/Cellar/<package>/<version>/<prefix>` and then a global prefix at `/opt/homebrew/<prefix>`. Rizin is installed in the version-specific dir, but if a user wants to install packages for the whole system, he needs to install them in `/opt/homebrew/<prefix>` so that they are not overwritten on a rizin upgrade.

To do that we introduce a `EXTRA_PREFIX` from where we load plugins, and few other things. I didn't really re-implemented everything to load from `EXTRA_PREFIX` because not everything is loaded from both system and home prefixes so it probably does not make sense to have the additional flexibility.

**Test plan**

...

**Closing issues**

...


XVilka:
@ret2libc what is the status of this PR? This kind of change it better to merge earlier, to catch possible problems, rather then later, before the release.

ret2libc:
This is in theory ok, but I wanted to first fix `dist` branch so that it builds correctly. The red CI is annoying :(

XVilka (approved):

PR #3496, opened 2023-05-03
Rewrite how binaries are loaded

### LINKED ISSUE #3404: Language should be inferred before demangling the symbols
Currently, the demangling happens in the

```c
RzList /*<RzBinSymbol *>*/ *symbols(RzBinFile *bf)
```

function, which has to be implemented for every executable format by its plugin. The interface can be found at

https://github.com/rizinorg/rizin/blob/f4e15d232a407a72ff8d393f22ff06daf4205c6c/librz/include/rz_bin.h#L531

It is invoked in `rz_bin_object_set_items` as following

https://github.com/rizinorg/rizin/blob/f4e15d232a407a72ff8d393f22ff06daf4205c6c/librz/bin/bobj.c#L455

The invocation of `symbols` happens before the invocation of `rz_bin_language_detect`

https://github.com/rizinorg/rizin/blob/f4e15d232a407a72ff8d393f22ff06daf4205c6c/librz/bin/bobj.c#L527

This implies that `symbols` tries to guess the language for every single symbol instead of using the language detected in `rz_bin_language_detect`. This language is then used to resolve the demangler and demangle the symbol. As expected, the guesses can be incorrect and hence many symbols don't even get demangled by the correct demangler (the one associated with the executable's language).

The solution to this is not just to put the `rz_bin_language_detect` before the `symbols` function is called. This is because `rz_bin_language_detect` uses the symbols in the binfile (`binfile->o->symbols`) to deduce the language of the executable. And the symbols list is empty until the `symbols` function gets executed, which populates the list.

One way would be to populate the symbols list, but not demangle it yet. Then detect the language. Once we know the language, we can now demangle the symbols, without guessing the language for every symbol. This would require us to implement two functions (on which populates, other which demangles) for all the executable plugins (not so good).

I would like to hear about any neater and less disruptive solutions.

wargio:
# DO NOT SQUASH

**Detailed description**

This is a rework of how the RzBinPlugins are being used by RzBin to load RzBinObject data.

The whole rewrite of the function is due how many times we re-iterate to add little information to the single element.

Also this allows us to demangle symbols immediately instead doing it on-the fly.

Thanks to @DMaroo for the original PR (i had to rewrite some stuff)

Fix #3404


test/db/formats/qnx
```diff
@@ -18,10 +18,10 @@ NAME=sections of qnx
...
 CMDS=iS
 EXPECT=<<EOF
-paddr      size  vaddr      vsize align perm name     type flags 
------------------------------------------------------------------
-0x00000044 0xf6  0x00000044 0xf6  0x0   ---- LMF_LOAD      
-0x00000152 0x6c3 0x00000152 0x6c3 0x0   ---- LMF_LOAD      
+paddr      size  vaddr      vsize align perm name              type flags 
+--------------------------------------------------------------------------
+0x00000044 0xf6  0x00000044 0xf6  0x0   ---- LMF_LOAD               
+0x00000152 0x6c3 0x00000152 0x6c3 0x0   ---- LMF_LOAD_00000152
```
XVilka: I wonder, maybe better remove leading zeroes? also use `0x` prefix to mark it's hex.

librz/bin/bin.c
```diff
@@ -136,10 +136,11 @@ RZ_API RzBinImport *rz_bin_import_clone(RzBinImport *o) {
 
 RZ_API void rz_bin_import_free(RzBinImport *imp) {
 	if (imp) {
```
XVilka: Reverse the condition while you are on it, please

librz/bin/bin.c
```diff
@@ -1316,188 +1317,74 @@ RZ_API const RzBinXtrPlugin *rz_bin_xtrplugin_get(RZ_NONNULL RzBin *bin, RZ_NONN
...
-	if (sym && sym->vaddr == 0) {
-		sym->vaddr = vaddr;
-	}
-	*name = ':';
-	return demangled;
-}
-
 /**
- * \brief Demangles a symbol based on the language or the RzBinFile data
+ * \brief Demangles a symbol based on the language or by iterating all demanglers.
```
XVilka: For all `RzBinFiles` that are in the `RzBin` or just a current pne? Please specify that in a doxygen doc

librz/bin/bobj_process.c
```diff
@@ -0,0 +1,199 @@
...
+	}
+
+	char *name = get_swift_field(demangled, classname);
+	if (name) {
+		rz_bin_object_add_field(o, classname, name, paddr, vaddr);
+		free(name);
+		return;
+	}
+
+#if 0
```
XVilka: Remove this?

librz/core/cbin.c
```diff
@@ -1284,105 +1296,91 @@ RZ_API bool rz_core_bin_apply_imports(RzCore *core, RzBinFile *binfile, bool va)
...
+	if (!flagname) {
 		return NULL;
 	}
-	rz_name_filter(r, len, true); // maybe unnecessary..
-	char *R = __filterQuotedShell(r);
-	free(r);
-	return R;
+	rz_name_filter(flagname, -1, true);
+
+	// TODO: probably we should remove this due to newshell.
```
XVilka: @ret2libc take a look at this this thing in particular, please

XVilka (changes requested):
Great new design and optimization!

XVilka:
See also some broken tests:
```
[XX] db/cmd/cmd_acll acll list class detailed
RZ_NOPLUGINS=1 /private/var/folders/bc/yvvw2g2s4jlg9r8x7tf61xv800009d/T/woodpecker-local-693079402/rizinorg/rizin/prefix/bin/rizin -escr.utf8=0 -escr.color=0 -escr.interactive=0 -eflirt.sigdb.load.system=false -eflirt.sigdb.load.home=false -N -Qc 'aaa
acll
' bins/elf/analysis/elf-virtualtable
-- stdout
--- expected
+++ actual
@@ -22,24 +22,26 @@
 2   printValue      0x400bf8 0x8       VIRTUAL
 3   method.A.greet  0x400ac8 0x0       VIRTUAL
 
+[std]
 [std::basic_ostream_char__std::char_traits_char_____std::endl_char__std]
 nth name                     addr vt_offset type    
 ----------------------------------------------------
-1   char_traits<char> >  0x6012a8 -1        DEFAULT
+1   char_traits<char> >  0x4008b0 -1        DEFAULT
 
 [std::basic_ostream_char__std::char_traits_char_____std::operator____std]
 nth name                     addr vt_offset type    
 ----------------------------------------------------
-1   char_traits<char> >  0x601298 -1        DEFAULT
+1   char_traits<char> >  0x400890 -1        DEFAULT
 
 [std::ios_base::Init]
 nth name       addr vt_offset type    
 --------------------------------------
-1   Init   0x601278 -1        DEFAULT
-2   ~Init  0x601290 -1        DEFAULT
+1   Init   0x400850 -1        DEFAULT
+2   ~Init  0x400880 -1        DEFAULT

' bins/elf/analysis/classes_Polygon
-- stdout
--- expected
+++ actual
@@ -28,19 +28,21 @@
 3   Triangle   0x15d4 -1        CONSTRUCTOR
 4   ~Triangle  0x1602 -1        DESTRUCTOR
 
+[std]
 [std::basic_ostream_char__std::char_traits_char_____std::endl_char__std]
-nth name                   addr vt_offset type    
---------------------------------------------------
-1   char_traits<char> >  0x3fd0 -1        DEFAULT
+nth name                               addr vt_offset type    
+--------------------------------------------------------------
+1   char_traits<char> >  0xffffffffffffffff -1        DEFAULT
 
 [std::ios_base::Init]
-nth name     addr vt_offset type    
-------------------------------------
-1   Init   0x3fb0 -1        DEFAULT
-2   ~Init  0x3ff8 -1        DEFAULT
+nth name                 addr vt_offset type    
+------------------------------------------------
+1   ~Init  0xffffffffffffffff -1        DEFAULT
+2   Init               0x10d0 -1        DEFAULT
 
 [std::ostream]
 nth name          addr vt_offset type    
 -----------------------------------------
-1   operator<<  0x3fb8 -1        DEFAULT
+1   operator<<  0x10b0 -1        DEFAULT
+2   operator<<  0x10e0 -1        DEFAULT
```

Not sure if new output is more valid than the old, please recheck.

Regarding the new disassembly output, is this change intended?
```
[XX] db/cmd/lea_intel lea_intel att subrel
RZ_NOPLUGINS=1 /private/var/folders/bc/yvvw2g2s4jlg9r8x7tf61xv800009d/T/woodpecker-local-693079402/rizinorg/rizin/prefix/bin/rizin -escr.utf8=0 -escr.color=0 -escr.interactive=0 -eflirt.sigdb.load.system=false -eflirt.sigdb.load.home=false -N -Qc 'e asm.lines.bb=0
e asm.bytes=0
e asm.comments=false
s 0x00402c43
pd 30
' bins/elf/ezpz
-- stdout
--- expected
+++ actual
@@ -0,19 +0,19 @@
   0x00402c43      lea   rax, [rbp - 0xa0]                              ; ezpz.cpp:265
   0x00402c4a      mov   rdi, rax
-  0x00402c4d      call  sym MD5_Init(MD5_CTX*)
+  0x00402c4d      call  sym.MD5_Init_MD5_CTX
   0x00402c52      mov   eax, dword [rbp - 0xb4]                        ; ezpz.cpp:266
   0x00402c58      add   eax, 1
   0x00402c5b      movsxd rdx, eax
   0x00402c5e      lea   rax, [rbp - 0xa0]
   0x00402c65      mov   esi, obj.Password
   0x00402c6a      mov   rdi, rax
-  0x00402c6d      call  sym MD5_Update(MD5_CTX*, void const*, unsigned long)
+  0x00402c6d      call  sym.MD5_Update_MD5_CTX___void_const___unsigned_long
   0x00402c72      lea   rdx, [rbp - 0xa0]                              ; ezpz.cpp:267
   0x00402c79      lea   rax, [rbp - 0xb0]
   0x00402c80      mov   rsi, rdx
   0x00402c83      mov   rdi, rax
-  0x00402c86      call  sym MD5_Final(unsigned char*, MD5_CTX*)
+  0x00402c86      call  sym.MD5_Final_unsigned_char___MD5_CTX
   0x00402c8b      mov   eax, dword [rbp - 0xb4]                        ; ezpz.cpp:269
   0x00402c91      cdqe
   0x00402c93      shl   rax, 4
```
I think the old output might have been more helpful for the reverser than the new one?
Same with the Pascal one:
```
[XX] db/cmd/dwarf "function info integration freepascal"
RZ_NOPLUGINS=1 /private/var/folders/bc/yvvw2g2s4jlg9r8x7tf61xv800009d/T/woodpecker-local-693079402/rizinorg/rizin/prefix/bin/rizin -escr.utf8=0 -escr.color=0 -escr.interactive=0 -eflirt.sigdb.load.system=false -eflirt.sigdb.load.home=false -N -Qc 'e analysis.jmp.cref=true
e asm.bytes=true
e asm.dwarf=false
aaa
pd 20 @  dbg.palya
' bins/efi/freepascal_test_dwarf
-- stdout
--- expected
+++ actual
@@ -16,7 +16,7 @@
 |           0x00401999      b801000000     mov   eax, 1
 |           0x0040199e      bf01000000     mov   edi, 1                ; int64_t arg2
 |           0x004019a3      89c6           mov   esi, eax              ; int64_t arg_8h
-|           0x004019a5      e8b6a10200     call  sym unit crt gotoxy(tcrtcoord,tcrtcoord) ; sym.unit_crt_gotoxy_tcrtcoord_tcrtcoord
+|           0x004019a5      e8b6a10200     call  sym.unit_crt_gotoxy_tcrtcoord_tcrtcoord
 |           0x004019aa      66448b65f0     mov   r12w, word [szel]
 |           0x004019af      66c745ec0100   mov   word [i], 1
 |           0x004019b5      66443b65ec     cmp   r12w, word [i]
@@ -24,6 +24,6 @@
 |       |   0x004019bc      66816dec0100   sub   word [i], 1
 |       |   0x004019c2      66666690       nop
 |       |   0x004019c6      6690           nop
-|       |   ; CODE XREF from unit p$mozgkigyo palya(smallint,smallint) @ 0x4019ff
+|       |   ; CODE XREF from dbg.palya @ 0x4019ff
 |       |   0x004019c8      668145ec0100   add   word [i], 1
 |       |   0x004019ce      e82d280200     call  sym.fpc_get_output
```

test/db/formats/qnx
wargio: i think is better with them, but yes, i can always remove them.

librz/bin/bin.c
wargio: `RzBinFiles` is not used anymore. there was no need, it was a huge hack to add classes and methods.
This function should do one and only thing: demangle symbols.

kazarmy:
`rz_name_filter()` really shouldn't be removing trailing (and leading) underscores.

ret2libc:
I will have a look!

wargio:
> I think the old output might have been more helpful for the reverser than the new one?

I disagree. i would prefer to have the actual flag instead of the demangled string.
i think there is a huge mess on the printing side which should be cleanup.
I strongly think that we should show the demangled output as a comment (like other tool does)

The issue with the XREFS in the comment is on the flag side, where there is a huge mess on their output.
Probably what i will do is refactor how flags are created and have `name` which is the flag you normally see and maybe `demangled_string` which could be filled for those structures which can be demangled (like imports, symbols, relocs, etc..).

librz/core/canalysis.c
```diff
@@ -1591,18 +1591,7 @@ RZ_API int rz_core_analysis_fcn_clean(RzCore *core, ut64 addr) {
...
-	char *name = strdup(fcn->name ? fcn->name : "");
-	if (demangle) {
-		char *tmp = rz_bin_demangle(core->bin->cur, lang, name, fcn->addr, keep_lib);
-		if (tmp) {
-			free(name);
-			name = tmp;
-		}
-	}
-	return name;
+	return strdup(fcn->name ? fcn->name : "");
```
wargio: Note to myself: probably this should return `fcn.XXXXXX` if `fcn->name` is null

librz/core/cbin.c
```diff
@@ -31,6 +31,25 @@
...
+	char *symbolname; ///< display symbol name
+	char *libname; ///< name of the lib this symbol is specific to, if any
+	char *nameflag; ///< flag name for symbol
+	char *demname; ///< demangled raw symbol name
+	char *demflag; ///< flag name for demangled symbol
+	char *classname; ///< classname
+	char *classflag; ///< flag for classname
+	char *methname; ///< methods [class]::[method]
+	char *methflag; ///< methods flag sym.[class].[method]
+} BinSymNames;
```
wargio: Note to myself: this is being used in cutter. i will revert it back how it was before.

librz/core/cbin.c
wargio: i doubt that now we suffer of code injections with newshell like we did before. but needs to be confirmed

librz/core/cbin.c
```diff
@@ -3507,7 +3466,7 @@ static void classdump_objc(RzCore *r, RzBinClass *c) {
 		rz_cons_printf("@interface %s\n{\n", c->name);
 	}
 	RzListIter *iter2, *iter3;
-	RzBinClassField *f;
+	RzBinField *f;
```
wargio: Note to myself: this was correct previously.

librz/core/cbin.c
```diff
@@ -3934,7 +3901,7 @@ RZ_API bool rz_core_bin_classes_print(RZ_NONNULL RzCore *core, RZ_NONNULL RzBinF
 	RzListIter *iter, *iter2, *iter3;
 	RzBinSymbol *sym;
 	RzBinClass *c;
-	RzBinClassField *f;
+	RzBinField *f;
```
wargio: Note to myself: this was correct previously.

wargio:
@ret2libc i'm waiting for your inputs before i start fixing things

librz/bin/bin.c
```diff
@@ -1316,188 +1317,74 @@ RZ_API const RzBinXtrPlugin *rz_bin_xtrplugin_get(RZ_NONNULL RzBin *bin, RZ_NONN
...
 
-	RzBinLanguage type = RZ_BIN_LANGUAGE_UNKNOWN;
-	RzBin *bin = bf ? bf->rbin : NULL;
-	RzBinObject *o = bf ? bf->o : NULL;
+	RzBinLanguage type = language ? rz_bin_language_to_id(language) : RZ_BIN_LANGUAGE_UNKNOWN;
+	type = RZ_BIN_LANGUAGE_MASK(type);
 
-	if (!language && o && o->info && o->info->lang) {
-		language = o->info->lang;
+	if (!strncmp(mangled, "reloc.", 6)) {
```
ret2libc: I wonder if we can assume that `mangled` *always* has a prefix like `reloc.`, `sym.`, etc. What happens if a symbol in the binary actually starts with "sym" or other prefixes we use here.

librz/bin/bin.c
```diff
@@ -1316,188 +1317,74 @@ RZ_API const RzBinXtrPlugin *rz_bin_xtrplugin_get(RZ_NONNULL RzBin *bin, RZ_NONN
...
- * \param libs Append the library name to the demangled symbol, if set to true
- * \return char* Demangled name of the \p symbol
+ * \param   bin       The RzBin context to used for demangling
+ * \param   language  The language to be used for demangling
+ * \param   mangled   The mangled string to be demangled
+ * \return  On success returns a valid pointer, otherwise NULL.
  */
-RZ_API RZ_OWN char *rz_bin_demangle(RZ_NULLABLE RzBinFile *bf, RZ_NULLABLE const char *language, RZ_NULLABLE const char *symbol, ut64 vaddr, bool libs) {
-	if (RZ_STR_ISEMPTY(symbol)) {
+RZ_API RZ_OWN char *rz_bin_demangle(RZ_NULLABLE RzBin *bin, RZ_NULLABLE const char *language, RZ_NULLABLE const char *mangled) {
```
ret2libc: Do you think this can be aggregated with the functions in `bin_demangle.c`? I see many similarities.

librz/bin/p/bin_avr.c
```diff
@@ -197,7 +197,7 @@ static RzList /*<RzBinSymbol *>*/ *symbols(RzBinFile *bf) {
 
 static RzList /*<RzBinString *>*/ *strings(RzBinFile *bf) {
 	// we dont want to find strings in avr bins because there are lot of false positives
```
ret2libc, line 199: shall this be removed? are there still a lot of false positives?

librz/bin/p/bin_bf.c
```diff
@@ -19,7 +19,7 @@ static ut64 baddr(RzBinFile *bf) {
 }
 
 static RzList /*<RzBinString *>*/ *strings(RzBinFile *bf) {
-	return NULL;
+	return rz_list_newf((RzListFree)rz_bin_string_free);
```
ret2libc, line 22: what about this?

librz/bin/bobj_process.c
```diff
@@ -0,0 +1,199 @@
...
+		name = strstr(demangled, classname);
+	}
+	if (name && name[strlen(classname)] == '.') {
+		rz_bin_object_add_method(o, classname, name, paddr, vaddr);
+	}
+#endif
+}
+#endif /* WITH_SWIFT_DEMANGLER */
+
+#include "bobj_process_class.c"
```
ret2libc: can we make these functions RZ_IPI and just use regularly instead of including .c files?

librz/bin/bobj_process_class.c
```diff
@@ -0,0 +1,92 @@
...
+	}
+
+	if (!ht_pp_find(o->glue_to_class_method, key, NULL)) {
+		ht_pp_insert(o->glue_to_class_method, key, method);
+	}
+	free(key);
+
+	if (!ht_up_find(o->vaddr_to_class_method, method->vaddr, NULL)) {
+		ht_up_insert(o->vaddr_to_class_method, method->vaddr, method);
+	}
```
ret2libc: you can just insert, it should ignore the insert if the key is already present.

librz/bin/bobj_process_class.c
```diff
@@ -0,0 +1,92 @@
...
+
+	char *key = rz_str_newf(RZ_BIN_FMT_CLASS_HT_GLUE, method->classname, method->name);
+	if (!key) {
+		RZ_LOG_ERROR("bin: failed to allocate class-method string\n");
+		return;
+	}
+
+	if (!ht_pp_find(o->glue_to_class_method, key, NULL)) {
+		ht_pp_insert(o->glue_to_class_method, key, method);
+	}
```
ret2libc: same and everywhere else

librz/bin/bin.c
wargio: tbh, i expect nothing like that.

librz/bin/bin.c
wargio: i can actually move it there. makes more sense.

librz/bin/p/bin_avr.c
wargio: actually it does find the strings correctly for our tests.

librz/bin/p/bin_bf.c
wargio: bf cannot have strings, because the code itself is a string.

librz/bin/bobj_process_class.c
wargio: no, this cannot use the insert directly, the issue is that sometimes the key might exists and we do not want to overwrite dups

librz/bin/bobj_process_class.c
wargio: i guess this is ok to switch as you asked.

librz/bin/bin.c
ret2libc: What do you mean?

librz/bin/bobj_process_class.c
ret2libc: I will check but I don't think it overwrites. The update function overwrites , the insert returns false if there is already an element

librz/core/canalysis.c
wargio: rz_core_analysis_fcn_name has been removed.

librz/bin/p/bin_avr.c
wargio: I will keep this  as is, since it does not break the tests (without it some stuff gets auto-detected and therefore some tests "fails").

librz/bin/bin.c
wargio: i have removed the checks. what i meant is that i should never expect those prefixes.

librz/core/cbin.c
ret2libc: Yes we can likely remove this, all the names can be wrapped in quotes or double quotes or bad chars can be escaped in rzshell, so it should be fine. You could also try to remove it in separate PR to see how things change in a more manageable way.

librz/bin/bobj_process_class.c
wargio: when i have used the insert function, it was weird the output due to those updates.

librz/core/cbin.c
wargio: i think i will do this in another PR

XVilka:
By the way, since we moved to rzshell, quoting whole line now should be deprecated:
```
"f method._JNIEnv.CallVoidMethod__jobject____jmethodID___... @ 0x60c0"
+"f method._JNIEnv.GetStaticMethodID__jclass___char_const___char_const @ 0x60dc"
+"f method._JNIEnv.CallStaticObjectMethod__jclass____jmethodID___... @ 0x60ea"
 "f method._JNIEnv.CallStaticIntMethod__jclass____jmethodID___... @ 0x695e"
-"f method._JNIEnv.CallStaticObjectMethod__jclass____jmethodID___... @ 0x60ea"
```
Not a part of this PR but a think to keep in mind.

test/db/formats/demangling_bin
```diff
@@ -17,12 +17,12 @@ EXPECT=<<EOF
 0x00001e24 0  stub helpers
 0x00002000 0 _pvars
 0x00002064 0 std::__ioinit
-0x00001dee 0 std::string::size() const
-0x00001df4 0 std::string::operator[](unsigned long) const
-0x00001dfa 0 std::ostream::operator<<(std::ostream& (*)(std::ostream&))
-0x00001e00 0 std::ios_base::Init::Init()
-0x00001e06 0 std::ios_base::Init::~Init()
-0x00001e12 0 std::basic_ostream<char, std::char_traits<char> >& std::operator<< <std::char_traits<char> >(std::basic_ostream<char, std::char_traits<char> >&, char const*)
+0x00001dee 0 imp.std::string::size() const
```
XVilka, line 20: A regression?

test/db/formats/demangling_bin
```diff
@@ -176,27 +174,24 @@ pd 1
...
-|   ; CALL XREF from QString::~QString() @ 0x80493e5           |
-|     ; method.QString._QString                                |
-|   ; CALL XREF from sym._fini @ 0x80495d4                     |
-| sym.__x86.get_pc_thunk.bx ();                                |
-| mov ebx, dword [esp]                                         |
-| ret                                                          |
-`--------------------------------------------------------------'
+.------------------------------------------------------------------.
+|  0x8049140                                                       |
+|   ; CALL XREF from sym._init @ 0x8049008                         |
```
XVilka, line 186: Looks much more readable now, awesome.

test/db/formats/demangling_bin
```diff
@@ -229,8 +224,8 @@ EOF
 EXPECT=<<EOF
             ; CALL XREF from main @ +0x222
             ; CALL XREF from fcn.00001717 @ +0xbb
-            ; CALL XREF from std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::basic_string<char*, void>(char*, char*, std::allocator<char> const&) @ +0x66 ; sym.std::__cxx11::basic_string_char__std::char_traits_char___std::allocator_char___::basic_string_char___void__char___char___std::allocator_char__const
-            ; CALL XREF from void std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::_M_construct<char*>(char*, char*, std::forward_iterator_tag) @ +0x116 ; sym.void_std::__cxx11::basic_string_char__std::char_traits_char___std::allocator_char___::_M_construct_char___char___char___std::forward_iterator_tag
+            ; CALL XREF from sym.std::__cxx11::basic_string_char__std::char_traits_char___std::allocator_char___::basic_string_char___void__char___char___std::allocator_char__const @ +0x66
```
XVilka, line 227: I think previous output made more sense

test/db/formats/demangling_bin
wargio: no, is not. just re-sorted.

librz/analysis/class.c
```diff
@@ -623,7 +611,6 @@ RZ_API void rz_analysis_class_method_fini(RzAnalysisMethod *meth) {
 RZ_API void rz_analysis_class_method_recover(RzAnalysis *analysis, RzBinClass *cls, RzList /*<RzBinSymbol *>*/ *methods) {
```
XVilka, line 614: Could you please add a Doxygen comment why it substitutes certain things?

XVilka: Could you please also add `RZ_NULLABLE meth` and quick exit into `rz_analysis_class_method_fini()`?

librz/bin/bin.c
XVilka: Don't forget this one

librz/bin/bobj.c
```diff
@@ -36,16 +36,6 @@ RZ_API void rz_bin_string_decode_base64(RZ_NONNULL RzBinString *bstr) {
...
-	rz_return_if_fail(strings);
-
-	RzBinString *bstr;
-	RzListIter *iter;
-	rz_list_foreach (strings, iter, bstr) {
-		rz_bin_string_decode_base64(bstr);
-	}
-}
-
 RZ_API void rz_bin_mem_free(void *data) {
```
XVilka: Please add `RZ_NULLABLE` and quick exit

librz/bin/bobj.c
```diff
@@ -190,88 +180,263 @@ RZ_IPI void rz_bin_object_free(RzBinObject *o) {
...
+ * \param[in]  vaddr   The virtual address of the method
+ *
+ * \return     On success returns a valid pointer, otherwise NULL.
+ */
+RZ_API RzBinSymbol *rz_bin_object_find_method_by_vaddr(RZ_NONNULL RzBinObject *o, ut64 vaddr) {
+	rz_return_val_if_fail(o, NULL);
+	return (RzBinSymbol *)ht_up_find(o->vaddr_to_class_method, vaddr, NULL);
+}
+
+RZ_API RZ_BORROW RzBinSymbol *rz_bin_object_add_method(RZ_NONNULL RzBinObject *o, RZ_NONNULL const char *klass, RZ_NONNULL const char *method, ut64 paddr, ut64 vaddr) {
```
XVilka, line 359: Add Doxygne please

librz/bin/bobj.c
```diff
@@ -190,88 +180,263 @@ RZ_IPI void rz_bin_object_free(RzBinObject *o) {
...
+	char *key = rz_str_newf(RZ_BIN_FMT_CLASS_HT_GLUE, klass, field);
+	if (!key) {
+		return NULL;
+	}
+	RzBinClassField *sym = (RzBinClassField *)ht_pp_find(o->glue_to_class_field, key, NULL);
+	free(key);
+	return sym;
+}
+
+RZ_API RZ_BORROW RzBinClassField *rz_bin_object_add_field(RZ_NONNULL RzBinObject *o, RZ_NONNULL const char *klass, RZ_NONNULL const char *name, ut64 paddr, ut64 vaddr) {
```
XVilka, line 438: Doxygen please

test/db/formats/dyldcache
```diff
@@ -657,11 +655,11 @@ canary   false
...
 RELROCS  false
 NX       false
-nth paddr      vaddr       bind   type size lib name                      
---------------------------------------------------------------------------
+nth paddr      vaddr       bind   type size lib name                            
+--------------------------------------------------------------------------------
 0   0x00007fb4 0x180007fb4 GLOBAL FUNC 0        _i_am_libSystem
 0   0x0000bfb0 0x18000bfb0 GLOBAL FUNC 0        _say_hello
-0   0x0000ff38 0x18000ff38 GLOBAL FUNC 0        dyld3::entryVectorForDyld
+0   0x0000ff38 0x18000ff38 GLOBAL FUNC 0        __ZN5dyld318entryVectorForDyldE
```
XVilka: This one looks like a regression?

XVilka (changes requested):
Great refactoring! Much cleaner code and better output. Apart from few comments I left - LGTM.

test/db/formats/dyldcache
thestr4ng3r: I see similar differences locally with real-world dyldcaches, does look like a regression to me. `is` output:
```diff
  10693  0x3555ebe8 0x21a4d2be8 LOCAL  FUNC 0        __OBJC_PROTOCOL_$_NSObject
 10694  0x3555ec48 0x21a4d2c48 LOCAL  FUNC 0        __OBJC_PROTOCOL_$_NSXPCListenerDelegate
 10695  0x3555eca8 0x21a4d2ca8 LOCAL  FUNC 0        __OBJC_PROTOCOL_$_DIIOClientProtocol
-10696  0x3555ed08 0x21a4d2d08 LOCAL  FUNC 0        boost::icl::identity_element<std::__1::optional<udif::details::block_info> >::value()::_value
-10697  0x3555ed98 0x21a4d2d98 LOCAL  FUNC 0        boost::icl::identity_element<unsigned long long>::value()::_value
-10698  0x3555eda0 0x21a4d2da0 LOCAL  FUNC 0        boost::icl::identity_element<udif::details::run_info>::value()::_value
-10699  0x3555edb8 0x21a4d2db8 LOCAL  FUNC 0        guard variable for boost::icl::identity_element<udif::details::run_info>::value()::_value
-10700  0x3555edc0 0x21a4d2dc0 LOCAL  FUNC 0        ref::details::_get_tagged_ptr_lock(void const*)::mutexs
-10701  0x3555f1c0 0x21a4d31c0 LOCAL  FUNC 0        ref::details::_get_tagged_ptr_cond(void const*)::conds
-10702  0x3555f4c0 0x21a4d34c0 LOCAL  FUNC 0        guard variable for make_error_code(DIErrorEnum)::di_err_category_instance
+10696  0x3555ed08 0x21a4d2d08 LOCAL  FUNC 0        __ZZN5boost3icl16identity_elementINSt3__18optionalIN4udif7details10block_infoEEEE5valueEvE6_value
+10697  0x3555ed98 0x21a4d2d98 LOCAL  FUNC 0        __ZZN5boost3icl16identity_elementIyE5valueEvE6_value
+10698  0x3555eda0 0x21a4d2da0 LOCAL  FUNC 0        __ZZN5boost3icl16identity_elementIN4udif7details8run_infoEE5valueEvE6_value
+10699  0x3555edb8 0x21a4d2db8 LOCAL  FUNC 0        __ZGVZN5boost3icl16identity_elementIN4udif7details8run_infoEE5valueEvE6_value
+10700  0x3555edc0 0x21a4d2dc0 LOCAL  FUNC 0        __ZZN3ref7details20_get_tagged_ptr_lockEPKvE6mutexs
+10701  0x3555f1c0 0x21a4d31c0 LOCAL  FUNC 0        __ZZN3ref7details20_get_tagged_ptr_condEPKvE5conds
+10702  0x3555f4c0 0x21a4d34c0 LOCAL  FUNC 0        __ZGVZ15make_error_code11DIErrorEnumE24di_err_category_instance
 10703  0x3555f4c8 0x21a4d34c8 LOCAL  FUNC 0        _gErrorContextKey
-10704  0x3555f4d0 0x21a4d34d0 LOCAL  FUNC 0        -[DIEncryptionUnlocker lookupLegacyKeychainWithXpcHandler:error:]::pDIHLUpgradeKeychainItem
-10705  0x3555f4d8 0x21a4d34d8 LOCAL  FUNC 0        -[DIEncryptionUnlocker lookupLegacyKeychainWithXpcHandler:error:]::once
-10706  0x3555f4e0 0x21a4d34e0 LOCAL  FUNC 0        system_properties::get_logical_cpus()::flag
-10707  0x3555f4e8 0x21a4d34e8 LOCAL  FUNC 0        system_properties::get_logical_cpus()::num_cpus
-10708  0x3555f4f0 0x21a4d34f0 LOCAL  FUNC 0        +[DIError frameworkBundle]::_bundle
-10709  0x3555f4f8 0x21a4d34f8 LOCAL  FUNC 0        +[DIError frameworkBundle]::onceToken
+10704  0x3555f4d0 0x21a4d34d0 LOCAL  FUNC 0        __ZZ65-[DIEncryptionUnlocker lookupLegacyKeychainWithXpcHandler:error:]E24pDIHLUpgradeKeychainItem
+10705  0x3555f4d8 0x21a4d34d8 LOCAL  FUNC 0        __ZZ65-[DIEncryptionUnlocker lookupLegacyKeychainWithXpcHandler:error:]E4once
+10706  0x3555f4e0 0x21a4d34e0 LOCAL  FUNC 0        __ZZN17system_properties16get_logical_cpusEvE4flag
+10707  0x3555f4e8 0x21a4d34e8 LOCAL  FUNC 0        __ZZN17system_properties16get_logical_cpusEvE8num_cpus
+10708  0x3555f4f0 0x21a4d34f0 LOCAL  FUNC 0        __ZZ26+[DIError frameworkBundle]E7_bundle
+10709  0x3555f4f8 0x21a4d34f8 LOCAL  FUNC 0        __ZZ26+[DIError frameworkBundle]E9onceToken
 10710  0x3555f500 0x21a4d3500 LOCAL  FUNC 0        _getDIOSLog.log
 10711  0x3555f508 0x21a4d3508 LOCAL  FUNC 0        _getDIOSLog.onceToken
 10712  0x3555f510 0x21a4d3510 LOCAL  FUNC 0        _forwardLogs
```

wargio: It's because if fails to guess the language of the dyldcache

thestr4ng3r: It is at least C++, Objective-C and Swift in the same binary here. Can we handle that somehow?

thestr4ng3r (changes requested):
Still some issues in dyldcaches, e.g.
```diff
 10694  0x3555ec48 0x21a4d2c48 LOCAL  FUNC 0        __OBJC_PROTOCOL_$_NSXPCListenerDelegate
 10695  0x3555eca8 0x21a4d2ca8 LOCAL  FUNC 0        __OBJC_PROTOCOL_$_DIIOClientProtocol
-10696  0x3555ed08 0x21a4d2d08 LOCAL  FUNC 0        boost::icl::identity_element<std::__1::optional<udif::details::block_info> >::value()::_value
-10697  0x3555ed98 0x21a4d2d98 LOCAL  FUNC 0        boost::icl::identity_element<unsigned long long>::value()::_value
-10698  0x3555eda0 0x21a4d2da0 LOCAL  FUNC 0        boost::icl::identity_element<udif::details::run_info>::value()::_value
-10699  0x3555edb8 0x21a4d2db8 LOCAL  FUNC 0        guard variable for boost::icl::identity_element<udif::details::run_info>::value()::_value
-10700  0x3555edc0 0x21a4d2dc0 LOCAL  FUNC 0        ref::details::_get_tagged_ptr_lock(void const*)::mutexs
-10701  0x3555f1c0 0x21a4d31c0 LOCAL  FUNC 0        ref::details::_get_tagged_ptr_cond(void const*)::conds
-10702  0x3555f4c0 0x21a4d34c0 LOCAL  FUNC 0        guard variable for make_error_code(DIErrorEnum)::di_err_category_instance
+10696  0x3555ed08 0x21a4d2d08 LOCAL  FUNC 0        __ZZN5boost3icl16identity_elementINSt3__18optionalIN4udif7details10block_infoEEEE5valueEvE6_value
+10697  0x3555ed98 0x21a4d2d98 LOCAL  FUNC 0        __ZZN5boost3icl16identity_elementIyE5valueEvE6_value
+10698  0x3555eda0 0x21a4d2da0 LOCAL  FUNC 0        __ZZN5boost3icl16identity_elementIN4udif7details8run_infoEE5valueEvE6_value
+10699  0x3555edb8 0x21a4d2db8 LOCAL  FUNC 0        __ZGVZN5boost3icl16identity_elementIN4udif7details8run_infoEE5valueEvE6_value
+10700  0x3555edc0 0x21a4d2dc0 LOCAL  FUNC 0        __ZZN3ref7details20_get_tagged_ptr_lockEPKvE6mutexs
+10701  0x3555f1c0 0x21a4d31c0 LOCAL  FUNC 0        __ZZN3ref7details20_get_tagged_ptr_condEPKvE5conds
+10702  0x3555f4c0 0x21a4d34c0 LOCAL  FUNC 0        __ZGVZ15make_error_code11DIErrorEnumE24di_err_category_instance
 10703  0x3555f4c8 0x21a4d34c8 LOCAL  FUNC 0        _gErrorContextKey
```

I can share the file privately if you need it

librz/bin/bin.c
```diff
@@ -135,23 +135,25 @@ RZ_API RzBinImport *rz_bin_import_clone(RzBinImport *o) {
 }
 
 RZ_API void rz_bin_import_free(RzBinImport *imp) {
```
XVilka: You forgot `RZ_NULLABLE` here

librz/bin/bin.c
```diff
@@ -135,23 +135,25 @@ RZ_API RzBinImport *rz_bin_import_clone(RzBinImport *o) {
...
 	}
+	free(imp->name);
+	free(imp->dname);
+	free(imp->libname);
+	free(imp->classname);
+	free(imp->descriptor);
+	free(imp);
 }
 
 RZ_API void rz_bin_resource_free(RzBinResource *res) {
```
XVilka: Same, `RZ_NULLABLE`

XVilka (approved):
Please fix tests output and LGTM

wargio:
> Still some issues in dyldcaches, e.g.
> 
> ```diff
>  10694  0x3555ec48 0x21a4d2c48 LOCAL  FUNC 0        __OBJC_PROTOCOL_$_NSXPCListenerDelegate
>  10695  0x3555eca8 0x21a4d2ca8 LOCAL  FUNC 0        __OBJC_PROTOCOL_$_DIIOClientProtocol
> -10696  0x3555ed08 0x21a4d2d08 LOCAL  FUNC 0        boost::icl::identity_element<std::__1::optional<udif::details::block_info> >::value()::_value
> -10697  0x3555ed98 0x21a4d2d98 LOCAL  FUNC 0        boost::icl::identity_element<unsigned long long>::value()::_value
> -10698  0x3555eda0 0x21a4d2da0 LOCAL  FUNC 0        boost::icl::identity_element<udif::details::run_info>::value()::_value
> -10699  0x3555edb8 0x21a4d2db8 LOCAL  FUNC 0        guard variable for boost::icl::identity_element<udif::details::run_info>::value()::_value
> -10700  0x3555edc0 0x21a4d2dc0 LOCAL  FUNC 0        ref::details::_get_tagged_ptr_lock(void const*)::mutexs
> -10701  0x3555f1c0 0x21a4d31c0 LOCAL  FUNC 0        ref::details::_get_tagged_ptr_cond(void const*)::conds
> -10702  0x3555f4c0 0x21a4d34c0 LOCAL  FUNC 0        guard variable for make_error_code(DIErrorEnum)::di_err_category_instance
> +10696  0x3555ed08 0x21a4d2d08 LOCAL  FUNC 0        __ZZN5boost3icl16identity_elementINSt3__18optionalIN4udif7details10block_infoEEEE5valueEvE6_value
> +10697  0x3555ed98 0x21a4d2d98 LOCAL  FUNC 0        __ZZN5boost3icl16identity_elementIyE5valueEvE6_value
> +10698  0x3555eda0 0x21a4d2da0 LOCAL  FUNC 0        __ZZN5boost3icl16identity_elementIN4udif7details8run_infoEE5valueEvE6_value
> +10699  0x3555edb8 0x21a4d2db8 LOCAL  FUNC 0        __ZGVZN5boost3icl16identity_elementIN4udif7details8run_infoEE5valueEvE6_value
> +10700  0x3555edc0 0x21a4d2dc0 LOCAL  FUNC 0        __ZZN3ref7details20_get_tagged_ptr_lockEPKvE6mutexs
> +10701  0x3555f1c0 0x21a4d31c0 LOCAL  FUNC 0        __ZZN3ref7details20_get_tagged_ptr_condEPKvE5conds
> +10702  0x3555f4c0 0x21a4d34c0 LOCAL  FUNC 0        __ZGVZ15make_error_code11DIErrorEnumE24di_err_category_instance
>  10703  0x3555f4c8 0x21a4d34c8 LOCAL  FUNC 0        _gErrorContextKey
> ```
> 
> I can share the file privately if you need it

i want to fix this via a change on libdemangler side.

XVilka (approved):

thestr4ng3r:
> This is a rework of how the RzBinPlugins are being used by RzBin to load RzBinObject data.
> The whole rewrite of the function is due how many times we re-iterate to add little information to the single element.
> Also this allows us to demangle symbols immediately instead doing it on-the fly.

Could you please go more in-depth in the description? Some questions that I currently have and will be good for the commit description to answer:
* "immediately" in the above means when exactly?
* Did demangling happen on-the-fly multiple times for the same symbol before?
* `rz_bin_demangle()` had some heuristics in there like checking for `'T'` to see if it is a swift symbol. Was this just moved somewhere else or are there any actual semantic changes in the demangling logic? If yes, which.

XVilka:
```
Mismatched function annotation for rz_bin_xtrdata_free at <rizin/librz/include/rz_bin.h:394:13> : None (was RZ_NULLABLE at <rizin/librz/bin/bin.c:61:13>
```

wargio:
> > This is a rework of how the RzBinPlugins are being used by RzBin to load RzBinObject data.
> > The whole rewrite of the function is due how many times we re-iterate to add little information to the single element.
> > Also this allows us to demangle symbols immediately instead doing it on-the fly.
> 
> Could you please go more in-depth in the description? Some questions that I currently have and will be good for the commit description to answer:
> 
>     * "immediately" in the above means when exactly?
> 
>     * Did demangling happen on-the-fly multiple times for the same symbol before?
> 
>     * `rz_bin_demangle()` had some heuristics in there like checking for `'T'` to see if it is a swift symbol. Was this just moved somewhere else or are there any actual semantic changes in the demangling logic? If yes, which.

Before the rz_bin_demangle was doing many things besides demangling (like adding methods to classes etc..).
The demangling before was also done on the fly when printing or accessing the symbol and the action was performed multiple times even over the demangled string itself.

That "checking for `T` is swift" was done only for guessing the language and it was moved already in a unique place.
On top of that the code that handles how to generate methods/fields ([here](https://github.com/rizinorg/rizin/blob/b943e1d05f0585e3600f940f00a5b3c24a4f46c8/librz/bin/bobj.c#L258)) was already questionable, but it is kept ([rz_bin_process_swift](https://github.com/rizinorg/rizin/blob/4f7b1fd5ad9ee69ea4c5c7e81abf8e1d6e1a2f0e/librz/bin/bobj_process.c#L124)).
The good plan is to actually rely on the apple demangler, but i will try to add support on it later.

XVilka:
> > > This is a rework of how the RzBinPlugins are being used by RzBin to load RzBinObject data.
> > > The whole rewrite of the function is due how many times we re-iterate to add little information to the single element.
> > > Also this allows us to demangle symbols immediately instead doing it on-the fly.
> > 
> > 
> > Could you please go more in-depth in the description? Some questions that I currently have and will be good for the commit description to answer:
> > ```
> > * "immediately" in the above means when exactly?
> > 
> > * Did demangling happen on-the-fly multiple times for the same symbol before?
> > 
> > * `rz_bin_demangle()` had some heuristics in there like checking for `'T'` to see if it is a swift symbol. Was this just moved somewhere else or are there any actual semantic changes in the demangling logic? If yes, which.
> > ```
> 
> Before the rz_bin_demangle was doing many things besides demangling (like adding methods to classes etc..). The demangling before was also done on the fly when printing or accessing the symbol and the action was performed multiple times even over the demangled string itself.
> 
> That "checking for `T` is swift" was done only for guessing the language and it was moved already in a unique place. On top of that the code that handles how to generate methods/fields ([here](https://github.com/rizinorg/rizin/blob/b943e1d05f0585e3600f940f00a5b3c24a4f46c8/librz/bin/bobj.c#L258)) was already questionable, but it is kept ([rz_bin_process_swift](https://github.com/rizinorg/rizin/blob/4f7b1fd5ad9ee69ea4c5c7e81abf8e1d6e1a2f0e/librz/bin/bobj_process.c#L124)). The good plan is to actually rely on the apple demangler, but i will try to add support on it later.

@wargio could you please then squash all those commits into one (or multiple) and put that explanation in the commit message as @thestr4ng3r requested?

wargio:
The issue is that everything was mixed together (i'm talking about the old code), so just have commits about this is not exactly possible.

Let me see what i can do.

XVilka:
Then just squash into one with the good commit message. Please also rebase.

XVilka:
Regarding `asm.demangle` and inline names I created a separate issue: https://github.com/rizinorg/rizin/issues/3543 Feel free to add your suggestions and ideas there.

XVilka:
```
[XX] db/tools/rz rizin project with explicit file
RZ_NOPLUGINS=1 /private/var/folders/bc/yvvw2g2s4jlg9r8x7tf61xv800009d/T/woodpecker-local-693079402/rizinorg/rizin/prefix/bin/rizin -escr.utf8=0 -escr.color=0 -escr.interactive=0 -eflirt.sigdb.load.system=false -eflirt.sigdb.load.home=false -N -Qc '!rizin -escr.utf8=0 -easm.bytes=1 -escr.color=0 -escr.interactive=0 -N -Qc "f easy @ 0x080483fe; Ps .crackme.rzdb" bins/elf/ioli/crackme0x00
echo Same project, different file
!rizin -escr.utf8=0 -easm.bytes=1 -escr.color=0 -escr.interactive=0 -N -p .crackme.rzdb -Qc "ol; pdq 3 @ easy" bins/elf/ioli/crackme0x01
!rm .crackme.rzdb
' =
-- stdout
--- expected
+++ actual
@@ -5,4 +5,4 @@
 0x080483fe   easy:
 0x080483fe                 29c4  sub esp, eax
 0x08048400       c7042428850408  mov dword [esp], str.IOLI_Crackme_Level_0x01
-0x08048407           e810ffffff  call sym.imp.printf
+0x08048407           e810ffffff  call sym.imp.printf_0x804831c
```
Please also rebase

test/db/formats/demangling_bin
```diff
@@ -176,27 +174,24 @@ pd 1
...
-|     ; method.QString._QString                                |
-|   ; CALL XREF from sym._fini @ 0x80495d4                     |
-| sym.__x86.get_pc_thunk.bx ();                                |
-| mov ebx, dword [esp]                                         |
-| ret                                                          |
-`--------------------------------------------------------------'
+.------------------------------------------------------------------.
+|  0x8049140                                                       |
+|   ; CALL XREF from sym._init @ 0x8049008                         |
+|   ; CALL XREF from method.QtPrivate::RefCount.deref @ 0x8049349  |
```
XVilka, line 187: Why some `::` are substituted, e.g. in `QString`, but here it doesn't happen?

wargio: what are you talking about?

wargio: oh is because the class name is `QtPrivate::RefCount` meanwhile for QString is just `QString`

XVilka: It was `QtPrivate::RefCount::deref()` became `method.QtPrivate::RefCount.deref`, but what was `QString::QString()` became `method.QString.QString_char_const`. Looks inconsistent to me - when it decides to remove `::` and when it doesn't?

wargio: you are mixing flags with actual demangled strings.
Flags gets created as `"method." + class_name + "." + method` and then the flag is stripped of special chars

wargio: also as @kazarmy said here https://github.com/rizinorg/rizin/pull/3496#issuecomment-1534118795 the filter trims underscores.

XVilka: Ok, then it's better to sort these things out in a separate PR/whatever, see https://github.com/rizinorg/rizin/issues/3543

XVilka (approved):

XVilka (approved):

wargio:
# don't squash

librz/bin/bin_demangle.c
```diff
@@ -0,0 +1,117 @@
...
+
+	return RZ_STR_ISEMPTY(mangled) ? NULL : mangled;
+}
+#undef skip_prefix_s
+#undef skip_prefix_n
+
+RZ_IPI bool rz_bin_demangle_symbol(RzBinSymbol *bsym, const RzDemanglerPlugin *plugin) {
+	if (!plugin || bsym->dname) {
+		return false;
+	}
```
ret2libc, line 38: should this ever happen? or maybe it's a programming error when this happens and it's better to use `rz_return_`?

librz/bin/bin_demangle.c
```diff
@@ -0,0 +1,117 @@
...
+
+	bsym->dname = plugin->demangle(mangled);
+	return bsym->dname != NULL;
+}
+
+RZ_IPI bool rz_bin_demangle_import(RzBinImport *import, const RzDemanglerPlugin *plugin) {
+	if (!plugin) {
+		return false;
+	}
+
```
ret2libc, line 53: same

test/db/cmd/lea_intel
```diff
@@ -10,19 +10,19 @@ EOF
 EXPECT=<<EOF
   0x00402c43      lea   rax, [rbp - 0xa0]                              ; ezpz.cpp:265
   0x00402c4a      mov   rdi, rax
-  0x00402c4d      call  sym MD5_Init(MD5_CTX*)
+  0x00402c4d      call  sym.MD5_Init_MD5_CTX
```
ret2libc, line 13: mmm why this change? I think before it looked slightly better.

XVilka: I am glad someone agrees with me. But we already decided to address this later and properly, I created an issue to track this: https://github.com/rizinorg/rizin/issues/3543

XVilka: While minor inconvenience now, I think this should not block this PR

librz/bin/bin_demangle.c
wargio: no, it can happen. on java it does. we avoid demangling stuff we have demangled, and we don't demangle if we do not have a plugin.

librz/bin/bin_demangle.c
wargio: same as above.

test/db/cmd/lea_intel
wargio: i disagree and strongly. the output should be showing a flag, not a string with a demangled line.

XVilka: This is why our disagreement can be easily solved by a good configuration option

thestr4ng3r (approved):
Very nice and meaningful descriptions, thanks!

PR #3501, opened 2023-05-06
Fix visual debug view + %% help update

thestr4ng3r:
# DO NOT SQUASH

**Detailed description**

Fix visual debug view:
Regression from https://github.com/rizinorg/rizin/commit/16e722cffcaf49c3fec4243477e7aba24a7ea6e6:
Visual debug (Vpp) was empty because of broken commands.
While fixing this I also noticed that the help for `%%` was suboptimal:

Correct help for %% command:
The term "positive" is less meaningful for %% as it is dealing only with
unsigned values. It belongs to $! and uses the exact opposite condition,
which is the value being non-zero.

**Test plan**

`rz -c 'Vp'` should display a little hexdump, registers and disassembly

### DIFF
--- a/librz/core/tui/visual.c
+++ b/librz/core/tui/visual.c
@@ -3716,7 +3716,7 @@ RZ_IPI int rz_core_visual(RzCore *core, const char *input) {
 
 			if (cmdvhex && *cmdvhex) {
 				rz_strf(debugstr,
-					"?0 ; f+ tmp ; sr %s @e: cfg.seek.silent=true ; %s ; ?1 ; %s ; ?1 ; "
+					"%%0 ; f+ tmp ; sr %s @e: cfg.seek.silent=true ; %s ; %%1 ; %s ; %%1 ; "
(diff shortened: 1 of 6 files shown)


ret2libc (approved):

librz/core/tui/visual.c
```diff
@@ -3716,7 +3716,7 @@ RZ_IPI int rz_core_visual(RzCore *core, const char *input) {
 
 			if (cmdvhex && *cmdvhex) {
 				rz_strf(debugstr,
-					"?0 ; f+ tmp ; sr %s @e: cfg.seek.silent=true ; %s ; ?1 ; %s ; ?1 ; "
```
XVilka, line 3719: We really should use the API here

XVilka (approved):

librz/core/tui/visual.c
thestr4ng3r: Agree 100%

PR #3521, opened 2023-05-21
Remove useless index and addr from RzBinClass

wargio:

**Detailed description**

Removes some useless info from class.
So `addr` was mostly used to store either a phy or virt addr from where the info was parsed in the bin but the actual value for this is zero to the reverser.

IDA, BN and Ghidra do not have nor provide info about classes.


test/db/cmd/cmd_tc
```diff
@@ -168,8 +166,6 @@ FILE=bins/mach0/objc-employee-ios14-arm64
 CMDS=ic*
 EXPECT=<<EOF
 fs classes
-"f class.public.Employee @ 0x100007958"
```
XVilka, line 171: Why this disappears?

XVilka (changes requested):

test/db/cmd/cmd_tc
wargio: because before there was a flag for each `addr`, now since we remove it, we don't anymore.

ret2libc:
> Removes some useless info from class.
> So `addr` was mostly used to store either a phy or virt addr from where the info was parsed in the bin but the actual value for this is zero to the reverser.

What was this address like usually?

wargio:
usually for any bin besides mach0 with objc compiled code is either 0 or UT64_MAX.
seems that this address is a virtual address used only for objc classes to create xrefs.
See chat over mm.

XVilka:
As @thestr4ng3r determined it's still necessary for some cases, should be this one closed?

PR #3554, opened 2023-06-04
Add all the missing SPDX headers

wargio:
# Squash ME


XVilka (changes requested):
Aren't they defined in `.reuse/dep5` already?

wargio:
> Aren't they defined in `.reuse/dep5` already?

yes, but these are the correct ones.

sys/header-check.py
ret2libc, line 1: Isn't `reuse` enough?

PR #3611, opened 2023-06-26
Fix il vm aezi and aezs sync behavior

Heersin:

**Detailed description**

1. issue about reg sync 
found aezv [reg] [value] command change the value of register, but those changes would be discard after aezs, I guess it's not the expected behavior. 

that's due to the reg sync, aezv modifies the value of il_vm, but didn't sync back to rz_reg
 the il_vm_step work as follows :
```C
sync_il_reg_from_rz_reg();
evaluate_il()
sync_il_reg_to_rz_reg();
```
if we don't sync those changes to `rz_reg`, `il_vm` will load unchanged reg value from `rz_reg` first when execute `aezs`

- also, sync PC with il_vm in vm initialization, which makes aezv print correct PC value after aezi cmd.

2. Fix a bug in validation
3. Fix a bug of rzil float macro, use `INV` instead of `NOT` 
...

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/librz/core/cil.c
+++ b/librz/core/cil.c
@@ -436,6 +436,9 @@ RZ_IPI void rz_core_analysis_il_reinit(RzCore *core) {
 		// initialize the program counter with the current offset
 		rz_reg_set_value_by_role(core->analysis->reg, RZ_REG_NAME_PC, core->offset);
 		rz_core_reg_update_flags(core);
+
+		// sync back to il vm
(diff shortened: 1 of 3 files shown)


XVilka (approved):

wargio (approved):

wargio (approved):

thestr4ng3r:
Keep in mind aezv is really just for low-level debugging and users should always only change the registers. Likely we should remove aezv completely in the future.

PR #3687, opened 2023-07-22
Make rz_core_analysis_var_display work with stack vars when debugging

frmdstryr:

**Detailed description**

The `rz_analysis_var_addr` does not pull from the correct registers when debugging causing the stack address to be incorrect. This inlines the address calculation in `rz_core_analysis_var_display` and uses `rz_core_reg_getv_by_role_or_name` to read it correctly.

**Test plan**

For https://github.com/rizinorg/cutter/pull/3206

Before
Screenshot_20230721_084938

After

Screenshot_20230721_123759

**Closing issues**

N/A


XVilka (commented):
Not sure if the fix is correct. Could you please add a test in `test/db/`? I remember that function is used in `afvd`, right?

frmdstryr:
Yes `afvd` is `rz_analysis_function_vars_display_handler` which calls `rz_core_analysis_var_display`. 

Is there an example of testing cmds while in a debug session somewhere?

XVilka:
> Yes `afvd` is `rz_analysis_function_vars_display_handler` which calls `rz_core_analysis_var_display`.
> 
> Is there an example of testing cmds while in a debug session somewhere?

Yes, see `test/db/archos/linux-x64/` for example - these are debug tests that will run only on Linux x86_64

librz/core/canalysis.c
```diff
@@ -5123,7 +5123,10 @@ RZ_API RZ_OWN char *rz_core_analysis_var_display(RZ_NONNULL RzCore *core, RZ_NON
 		break;
 	}
 	case RZ_ANALYSIS_VAR_STORAGE_STACK: {
-		ut64 addr = rz_analysis_var_addr(var);
+		// rz_analysis_var_addr does not work when debugging
+		const char *regname = rz_reg_get_name(analysis->reg, RZ_REG_NAME_BP);
+		ut64 stack = rz_core_reg_getv_by_role_or_name(core, regname);
+		ut64 addr = stack + var->fcn->bp_off + var->storage.stack_off;
```
wargio: probably this should be enabled only when debugging.

wargio: And probably this should be done within `rz_analysis_var_addr`

ret2libc:
Nice fix, thanks! I agree with @XVilka though that the fix looks a bit out of place there though.

What if we rename/move `rz_analysis_var_addr` into `core/rz_core_analysis_var_addr(RzCore *, RzAnalysisVar *)` that uses `rz_core_reg_getv_by_role_or_name` instead of `rz_reg_getv`?

BTW, if you use `rz_core_reg_getv_by_role_or_name` you should be able to directly use `RZ_REG_NAME_BP` as argument.

frmdstryr:
Well, something is still not correct.. the test I added is broken. The rbp register still returns UT64_MAX for some reason. 

When debugging bin/cat (the test I added) the address is still incorrect but it seems to work for the other binary I was trying (`structs` from https://github.com/wrongbaud/hackaday-u/tree/master/session-three/exercises). I'm not sure what the difference is?

XVilka:
> Well, something is still not correct.. the test I added is broken. The rbp register still returns UT64_MAX for some reason.
> 
> When debugging bin/cat (the test I added) the address is still incorrect but it seems to work for the other binary I was trying (`structs` from https://github.com/wrongbaud/hackaday-u/tree/master/session-three/exercises). I'm not sure what the difference is?

You seem to debug the binary from the system, you need to use the binary that wouldn't change during the CI image upgrade - e.g. from rizin-testbins. This way you have full control on what is being run and avert breakage in case of the changed binary on the CI.

XVilka (changes requested):
It seems some CI jobs get slightly different value:
```
 [XX] db/archos/linux-x64/dbg_afvd dbg.afvd
RZ_NOPLUGINS=1 /home/runner/bin/rizin -escr.utf8=0 -escr.color=0 -escr.interactive=0 -eflirt.sigdb.load.system=false -eflirt.sigdb.load.home=false -N -d -A -Qc 'aa
dcu main
ds 8
afvd var_ch
' bins/elf/struct64
-- stdout
--- expected
+++ actual
@@ -1,1 +1,1 @@
-(qword)0x0000000100005661
+(qword)0x0000000100000061
```

frmdstryr:
I tried a few different bins but couldn't really find a good one to use for a test, any suggestions?

For some reason the var type is int64_t but it should be a single char/byte with just 0x61. I guess it seems to be using the padding or something. 

image

XVilka:
> I tried a few different bins but couldn't really find a good one to use for a test, any suggestions?
> 
> For some reason the var type is int64_t but it should be a single char/byte with just 0x61. I guess it seems to be using the padding or something.

you can change the variable type with `afvt`

XVilka:
As for autodetection - please open a new issue, I will check after the release

librz/core/canalysis.c
```diff
@@ -5097,6 +5097,17 @@ RZ_IPI bool rz_core_analysis_function_delete_var(RzCore *core, RzAnalysisFunctio
 	return true;
 }
 
+RZ_API ut64 rz_core_analysis_var_addr(RZ_NONNULL RzCore *core, RZ_NONNULL RzAnalysisVar *var) {
```
XVilka, line 5109: Please add Doxygen

XVilka (approved):

librz/include/rz_core.h
```diff
@@ -1281,6 +1281,16 @@ RZ_API void rz_core_rtr_enable(RZ_NONNULL RzCore *core, const char *cmdremote);
 
 RZ_API RZ_OWN char *rz_core_analysis_var_to_string(RZ_NONNULL RzCore *core, RZ_NONNULL RzAnalysisVar *var);
 RZ_API RZ_OWN char *rz_core_analysis_var_display(RZ_NONNULL RzCore *core, RZ_NONNULL RzAnalysisVar *var, bool add_name);
+
+/**
+ * @brief Get the address of a stack variable.
```
XVilka: We use `\brief` and `\param` syntax. `\ref` also

XVilka: See https://github.com/rizinorg/rizin/blob/dev/DEVELOPERS.md#documentation

XVilka (changes requested):

librz/include/rz_core.h
```diff
@@ -1282,6 +1282,17 @@ RZ_API void rz_core_rtr_enable(RZ_NONNULL RzCore *core, const char *cmdremote);
 RZ_API RZ_OWN char *rz_core_analysis_var_to_string(RZ_NONNULL RzCore *core, RZ_NONNULL RzAnalysisVar *var);
 RZ_API RZ_OWN char *rz_core_analysis_var_display(RZ_NONNULL RzCore *core, RZ_NONNULL RzAnalysisVar *var, bool add_name);
 
+/**
+ * \brief Get the address of a stack variable.
```
XVilka: Please move the comment to the function in the C file,not in header

XVilka (changes requested):

XVilka (approved):

PR #3690, opened 2023-07-26
Fix #3647

### LINKED ISSUE #3647: Some default panels broken
### Work environment

| OS/arch/bits (mandatory)                             | Linux 6.4.1-arch2-1 x86_64
| File format of the file you reverse (mandatory)      | any
| Architecture/bits of the file (mandatory)            | any
| `rizin -v` full output, **not truncated** (mandatory)  | rizin 0.6.0 @ linux-x86-64 (not printed: commit 83e40b0c0cc76fc344dd778ec89c5eb43a2eaa8a)

### Expected behavior

The default panels in visual panels mode should work.

### Actual behavior

"Stack" and any other panels using `px` commands are broken for some reason. The default command for "Stack" is `pxq 256@r:SP` which runs fine on its own but not in the panel. Changing the command (using the `"` key) to `pxq 256@r:SP` again makes the window say `pxq 256@r:SP (xc 256@r:SP)` which seems to work fine but also prints the results of `xc`, not `pxq`. Changing the command to `xc 256@r:SP` works as intended.

The "Registers" command is out of date and should be ``%== true `e cfg.debug`; %! dr; %% ar``. Changing it to this works fine.

Open anything with `-d` and check `V!`

image

wargio:

**Detailed description**
Fix #3647

### DIFF
--- a/librz/core/tui/panels.c
+++ b/librz/core/tui/panels.c
@@ -28,8 +28,8 @@
 
 #define PANEL_CMD_SYMBOLS       "isq"
 #define PANEL_CMD_XREFS         "axl"
-#define PANEL_CMD_STACK         "px"
-#define PANEL_CMD_REGISTERS     "?== true `e cfg.debug`; ?! dr; ?? ar" // select dr or ar depending on cfg.debug
+#define PANEL_CMD_STACK         "xc"
+#define PANEL_CMD_REGISTERS     "%== true `e cfg.debug`; %! dr; %% ar"
 #define PANEL_CMD_DISASSEMBLY   "pd"
 #define PANEL_CMD_DISASMSUMMARY "pdsf"
 #define PANEL_CMD_FUNCTION      "afl"


librz/core/tui/panels.c
```diff
@@ -28,8 +28,8 @@
 
 #define PANEL_CMD_SYMBOLS       "isq"
 #define PANEL_CMD_XREFS         "axl"
-#define PANEL_CMD_STACK         "px"
-#define PANEL_CMD_REGISTERS     "?== true `e cfg.debug`; ?! dr; ?? ar" // select dr or ar depending on cfg.debug
+#define PANEL_CMD_STACK         "xc"
+#define PANEL_CMD_REGISTERS     "drt:value/cols/name/role"
```
ret2libc: doesn't this work in debug mode only?

wargio: no also esil

XVilka (approved):

XVilka (approved):

librz/core/tui/panels.c
ret2libc: then why is there that hack later with "if the command starts with 'drt', replace it with 'art'"?

wargio: because the user made me notice that drt does not output anything if not in debug/emu mode

ret2libc: Then it was just ok to do as before, I think:
```
%== true `e cfg.debug` ; %! dr; %% ar
```

wargio: no it is not because there is a function that cuts by ` ` (white space) and splits the command in tokens.
That function breaks some stuff. the whole panel code needs to be refactored to use the C api. this is a huge mess of commands and weird ways of swapping how commands are executed.

ret2libc: Are you sure? It seems the user opening the issue has tried it and it worked. I also tried it briefly and it seems to work fine even in panels.

wargio: doesn't work for me. only for debug mode

wargio: seems that my not-working issue was caused by a broken layout

Issue #3650, opened 2023-07-10
Warn about or remove 256-byte limit for pf


imyxh:
### Work environment

| Questions                                            | Answers
|------------------------------------------------------|--------------------
| OS/arch/bits (mandatory)                             | Archlinux x86_64
| File format of the file you reverse (mandatory)      | ELF
| Architecture/bits of the file (mandatory)            | x86_64
| `rizin -v` full output, **not truncated** (mandatory)  | rizinrizin 0.6.0 @ linux-x86-64

### Expected behavior

`pf [$N]$fmt` should warn me if `$N*sizeof($fmt) > 256`, or it should actually print all `$N` occurences.

### Actual behavior

Seems like there is a fixed-size buffer that makes zeroes show up after the first 256 bytes of an array from the `pf` command.

### Steps to reproduce the behavior
```
$ ./build/binrz/rizin/rizin /bin/ls
[0x00005f70]> pf [0x20+0x04]q @ 0x1000
0x00001000 = (qword)[ 0x657473696765725f, 0x656e6f6c434d5472, 0x5f5f00656c626154, 0x6174735f6362696c, 0x006e69616d5f7472, 0x69665f6178635f5f, 0x5f00657a696c616e, 0x6574615f6178635f, 0x6372747300746978, 0x617473626f00706d, 0x636f6c6c615f6b63, 0x5f64656c6961665f, 0x0072656c646e6168, 0x5f0074756f647473, 0x6f6c667265766f5f, 0x6574697277660077, 0x656b636f6c6e755f, 0x5f73747570660064, 0x64656b636f6c6e75, 0x746e6972705f5f00, 0x6261006b68635f66, 0x6c6c616d0074726f, 0x787461747300636f, 0x6b636174735f5f00, 0x6961665f6b68635f, 0x7063706d656d006c, 0x6e616c5f6c6e0079, 0x7473006f666e6967, 0x635f5f0072686372, 0x6c5f625f65707974, 0x656c72747300636f, 0x707974635f5f006e, 0x0000000000000000, 0x0000000000000000, 0x0000000000000000, 0x0000000000000000 ]
[0x00005f70]> pf [0x20+0x04]q @ 0x1020
[...]

PR #3693, opened 2023-07-27
core: fix 'pxj' command to not use core->block

ret2libc:

**Detailed description**

`pxj` was using `core->block`, while `px` is correctly reading the length as requested. Fixing it by using the same code as `px`.

**Test plan**

```
$ rizin /bin/ls
> px 272
> pxj 272
```
output should be the same (no 0 values at the end of the `pxj` output).

**Closing issues**

Address the problem in https://github.com/rizinorg/rizin/issues/3650#issuecomment-1652237530

### DIFF
--- a/librz/core/cprint.c
+++ b/librz/core/cprint.c
@@ -346,32 +346,39 @@ RZ_API RZ_OWN char *rz_core_print_hexdump_or_hexdiff_str(RZ_NONNULL RzCore *core
 	RzPrint *print = core->print;
 	bool old_use_comments = print->use_comments;
 	print->use_comments = use_comment ? print->flags & RZ_PRINT_FLAGS_COMMENT : false;
-	switch (mode) {
-	case RZ_OUTPUT_MODE_STANDARD: {
-		ut64 from = rz_config_get_i(core->config, "diff.from");
-		ut64 to = rz_config_get_i(core->config, "diff.to");
-		if (from == to && !from) {
-			len_fixup(core, &addr, &len);
-			ut8 *buffer = malloc(len);
-			if (!buffer) {
-				return NULL;
-			}
-			rz_io_read_at(core->io, addr, buffer, len);
-			string = rz_print_hexdump_str(core->print, rz_core_pava(core, addr), buffer, len, 16, 1, 1);
-			free(buffer);
-		} else {
-			string = rz_core_print_hexdump_diff_str(core, addr, addr + to - from, len);
+	ut64 from = rz_config_get_i(core->config, "diff.from");
+	ut64 to = rz_config_get_i(core->config, "diff.to");
+	if (from == to && !from) {
+		len_fixup(core, &addr, &len);
+		ut8 *buffer = malloc(len);
+		if (!buffer) {
+			return NULL;
 		}
-		core->num->value = len;
-		break;
+		rz_io_read_at(core->io, addr, buffer, len);
+		switch (mode) {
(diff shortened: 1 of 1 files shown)


librz/core/cprint.c
```diff
@@ -346,32 +346,39 @@ RZ_API RZ_OWN char *rz_core_print_hexdump_or_hexdiff_str(RZ_NONNULL RzCore *core
...
-			rz_io_read_at(core->io, addr, buffer, len);
+	ut64 from = rz_config_get_i(core->config, "diff.from");
+	ut64 to = rz_config_get_i(core->config, "diff.to");
+	if (from == to && !from) {
+		len_fixup(core, &addr, &len);
+		ut8 *buffer = malloc(len);
+		if (!buffer) {
+			return NULL;
+		}
+		rz_io_read_at(core->io, addr, buffer, len);
```
XVilka, line 357: It can fail, please handle that

XVilka (changes requested):

wargio:
I think pxj was correctly behaving, and px is incorrect. we should not ask the user to provide the length but we should ask the user to use the `@!<size>` modifier.

ret2libc:
> I think pxj was correctly behaving, and px is incorrect. we should not ask the user to provide the length but we should ask the user to use the `@!<size>` modifier.

Good point... I wonder what's our take on core->block though.

wargio:
i think that all the commands should work in the same way. if i start adding length etc.. it might get messy

XVilka:
Agree on that, let's try to be uniform, especially inside one command "family".

ret2libc:
But length is already present in *all* commands under `px`:
```
Usage: px[?]   # Show hexdump
| px[j] [<len>]  # show hexdump
| pxa [<len>]    # show annotated hexdump
| pxA[l] [<len>] # show op analysis color map
| pxb [<len>]    # dump bits in hexdump form
| pxc [<len>]    # show hexdump with comments
| pxd[jhwq]      # show signed integer dump
| pxe [<len>]    # emoji hexdump! :)
| pxf [<len>]    # show hexdump of current function
| pxi [<len>]    # HexII compact binary representation
| pxr[?]         # show hexword references
| pxs [<len>]    # show hexadecimal in sparse mode
| pxt[j]         # show delta pointer table in rizin commands
| pxx            # show <N> bytes of hex-less hexdump
| pxX [<N>]      # show <N> words of hex-less hexdump
| px0 [<N>]      # 8bit hexpair list of bytes until zero byte
| pxh[j] [<len>] # show 2-bytes hexadecimal integers dump
| pxH[j] [<len>] # show 2-bytes hexadecimal integers dump, one per line
| pxw[j] [<len>] # show 4-bytes hexadecimal integers dump
| pxW[j] [<len>] # show 4-bytes hexadecimal integers dump, one per line
| pxq[j] [<len>] # show 8-bytes hexadecimal integers dump
| pxQ[j] [<len>] # show 8-bytes hexadecimal integers dump, one per line
| pxo [<len>]    # show 1-byte octal integers dump
| pxl[j] [<N>]   # display <N> lines of hexdump
```

So in this case it was `pxj` that was not respecting the length. I know however we have used the blocksize in a few places, but I do also think in some others it make sense to keep the length. Like, `px 10` is easier to write and use than `px @!10`.

XVilka:
Lets go with length then

wargio:
agreed.

wargio (approved):

PR #3718, opened 2023-08-02
Fix arrow keys on some terminals

McKael:

**Detailed description**

Some terminals (xterm, tmux, screen...) use the escape sequences ^[{A,B,C,D} for the arrow keys.
This M.R. adds support for those escape sequences.

Additionally, it fixes the behavior of Ctrl-L which does a cursor move before clearing the screen.

**Test plan**

Tested on multiple terminals (xterm, urxvt, wezterm, tmux), in command line and in visual mode (for example in graph view).

**Closing issues**


librz/cons/dietline.c
```diff
@@ -1680,7 +1680,6 @@ RZ_API const char *rz_line_readline_cb(RzLineReadCallback cb, void *user) {
 			__move_cursor_right();
 			break;
 		case 12: // ^L -- right
-			__move_cursor_right();
```
XVilka, line 1683: Why is that change? According to the comment it should move right, no?

XVilka (changes requested):
```
Formatting [624/1600] librz/cons/input.c
librz/cons/input.c:112:30: error: code should be clang-formatted [-Wclang-format-violations]
                switch (ch) { // Arrow keys
                                           ^
librz/cons/input.c:113:30: error: code should be clang-formatted [-Wclang-format-violations]
                        case 'A': ch = 'k'; break;
                                                  ^
librz/cons/input.c:114:30: error: code should be clang-formatted [-Wclang-format-violations]
                        case 'B': ch = 'j'; break;
                                                  ^
librz/cons/input.c:115:30: error: code should be clang-formatted [-Wclang-format-violations]
                        case 'C': ch = 'l'; break;
                                                  ^
librz/cons/input.c:116:30: error: code should be clang-formatted [-Wclang-format-violations]
                        case 'D': ch = 'h'; break;
                                                  ^
librz/cons/input.c:117:12: error: code should be clang-formatted [-Wclang-format-violations]
                        default:  ch = 0xf1 + (ch & 0xf); break;
                                ^
Formatting [625/1600] librz/cons/grep.c
```

librz/cons/dietline.c
McKael: Indeed, I forgot to update the comment.  I'll fix that.

Or do you disagree about Ctrl-L not being supposed to move the cursor?

librz/cons/input.c
```diff
@@ -109,7 +109,13 @@ RZ_API int rz_cons_arrow_to_hjkl(int ch) {
 			ch = rz_cons_readchar();
 		}
 #else
-		ch = 0xf1 + (ch & 0xf);
+		switch (ch) { // Arrow keys
+		case 'A': ch = 'k'; break;
+		case 'B': ch = 'j'; break;
+		case 'C': ch = 'l'; break;
+		case 'D': ch = 'h'; break;
```
ret2libc, line 116: is this CTRL-D? Because afaik CTRL-D has a special meaning in rizin/shell.

McKael: No, the full sequence is actually `\x27 \x4f D` (or `^[OD`).

McKael: @ret2libc I have opened the same PR in the radare2 project (22123) and it turns out the key codes sent are caused by my terminal being stuck in _application mode_, which seems to be a configuration issue (especially if no one but me has seen this problem).

The ^L bug, I think, is still real.  Please let me know if you can reproduce and if you want me to open a separate PR.

Here are the steps to reproduce the Ctrl-L issue:
- launch `rizin`
- type a string, e.g. "foobar"
- move the cursor to the left, e.g. ^A to go to the first column
- hit ^L one or several times

ret2libc: > Here are the steps to reproduce the Ctrl-L issue:
> 
> * launch `rizin`
> * type a string, e.g. "foobar"
> * move the cursor to the left, e.g. ^A to go to the first column
> * hit ^L one or several times

I can reproduce this 👍  and it is fixed in your PR

ret2libc (commented):
Could you provide a full reproducer also for the other issue you are trying to fix?

McKael:
> Could you provide a full reproducer also for the other issue you are trying to fix?

I've found out it is an issue with Debian' system zshrc configuration and my own setup.
I can prevent Debian from messing with my terminal setup by exporting `DEBIAN_PREVENT_KEYBOARD_CHANGES=1` in my zshenv.

I'm not sure how good it is, but here's a way to reproduce it:

```
% xterm -e zsh -f                   # launch zsh without configuration file in an xterm
% printf '%s' ${terminfo[smkx]}     # switch to application mode (in the new xterm)
% rizin                             # run rizin
# ... try to use the arrow keys
```

In application mode, with screen or tmux, if you run `cat -v` and type an arrow key
you should get sequences like `^[OD^[OA`.

All other applications behave correctly (I assume they use terminfo) so I suppose
adding support of those key sequences to rizin would still make sense.

ret2libc:
Ok I could see some problems with arrow by using your instructions, however I notice that also `CTRL-A` does not work in this way (it does not bring my cursor to the beginning of the line). So I'm not sure it's something we really need to fix in rizin... I'm tempted to just call it a "configuration issue". What do you think? Can you use CTRL-A/E?

McKael:
> Ok I could see some problems with arrow by using your instructions, however I notice that also `CTRL-A` does not work in this way (it does not bring my cursor to the beginning of the line). So I'm not sure it's something we really need to fix in rizin... I'm tempted to just call it a "configuration issue". What do you think? Can you use CTRL-A/E?

Yes, ctrl-a/ctrl-e definitely work for me (and using `cat -v`, I can see they still send ^A/^E).

I guess it is a configuration issue, but on the other hand I'm not sure supporting these sequences would be a bad thing since every other application does work and these sequences are in the terminfo database.

Using `infocmp -1 | grep cuu1`, for example, should display both sequences for the up arrow key (in some terminals they are the same):
```
        cuu1=\E[A,
        kcuu1=\EOA,
```

So I don't really know what the proper solution is (except a full switch to terminfo and no more hard-coded sequences), I think whether we should support those key sequences or not is really up to you.

McKael:
@ret2libc  If you don't think it's worth changing, I can create another PR with only the Ctrl-L fix and we can close this one.

ret2libc:
@McKael is the current PR enough to fix the issue for you? If so, we can just double check on a bunch of different settings and if no regressions, we could merge! :)

McKael:
@ret2libc Yes I've been running the patch for a while and it works fine for me.

ret2libc (approved):

ret2libc:
cc @XVilka

XVilka (approved):
LGTM but we need to test it on Windows first.

XVilka:
I tested it - it didn't break Windows consoles, so I merged. Thank you!

PR #3750, opened 2023-08-13
Gameboy RzIL Lifting

thestr4ng3r:
**squash this**

This adds RzIL lifting for all Game Boy (SM83) instructions, including asm tests for each.

Tracetesting was done with [SameBoy](https://github.com/rizinorg/SameBoy), which is one of the most accurate emulators (https://daid.github.io/GBEmulatorShootout/), on the following roms:
* A short play of Super Mario Land
* `bins/gb/emulateme.gb`
* blargg's cpu_instrs test rom, which is meant to cover pretty all legal instructions (except STOP) including edge cases https://gbdev.gg8.se/files/roms/blargg-gb-tests/

A small emulation example (`test/db/rzil/gb`) where a short string is decrypted in memory is added as well.


omasanori:
Bikeshed: how about renaming `gb` to `sm83`?

thestr4ng3r:
The plugin also contains some Gameboy-specific logic, like for memory mapping, which is not necessarily related to the sm83 itself so I find gb to be slightly more accurate here.

omasanori:
Thanks for explanation. It makes sense then.

XVilka (approved):
Looks good!

PR #3757, opened 2023-08-15
visual: mention `o` key in help

XVilka:

**Detailed description**

`o/O` rotates between different print modes - in hexadecimal output

**Test plan**

1. Open any binary
2. Open visual hex mode with `VP`
3. Press `o` and `O` to rotate between different hex representations

CI is green

### DIFF
--- a/librz/core/tui/visual.c
+++ b/librz/core/tui/visual.c
@@ -298,7 +298,7 @@ static const char *help_msg_visual[] = {
 	"mK/'K", "mark/go to Key (any key)",
 	"n/N", "seek next/prev function/flag/hit (scr.nkey)",
 	"g", "go/seek to given offset (g[g/G]<enter> to seek begin/end of file)",
-	"O", "toggle asm.pseudo and asm.esil",
+	"o/O", "rotate between different formats (next/prev)",
 	"p/P", "rotate print modes (hex, disasm, debug, words, buf)",
 	"q", "back to rizin shell",
 	"r", "toggle call/jmp/lea hints",


librz/core/tui/visual.c
```diff
@@ -298,7 +298,8 @@ static const char *help_msg_visual[] = {
 	"mK/'K", "mark/go to Key (any key)",
 	"n/N", "seek next/prev function/flag/hit (scr.nkey)",
 	"g", "go/seek to given offset (g[g/G]<enter> to seek begin/end of file)",
-	"O", "toggle asm.pseudo and asm.esil",
+	"o/O", "rotate between different formats (next/prev)",
+	"O", "toggle asm.pseudo and asm.esil in disasm mode",
```
ret2libc: Why keeping this line? The same key can't do two different things.

ret2libc (changes requested):

librz/core/tui/visual.c
XVilka: The problem - it behaves differently in different visual "submodes"

ret2libc: I think even in disasm mode it doesn't just "toggle asm.pseudo and asm.esil", but it does rotate between several formats (e.g. it also enables asm.bytes, etc.). So I would keep only the first description and remove the second.

XVilka: You are right - I tried and it is indeed rotates between modes.

ret2libc (approved):

PR #3764, opened 2023-08-16
Fix blake3 build with incompatible assemblers

thestr4ng3r:

**Detailed description**

Apple Clang 4.2 on Mac OS X 10.7 Lion does support avx2, but it has issues with some of the asm syntax used, in particular jc and some mov syntax.
So we add an additional check for a short assembly snippet.

**Test plan**

Build on x86_64 platforms should recognize simd extensions as before (can be checked in the CI output), but building on Lion should not fail.

### DIFF
--- a/subprojects/packagefiles/blake3/unix_syntax_check.S
+++ b/subprojects/packagefiles/blake3/unix_syntax_check.S
@@ -1,0 +1,30 @@
+#if defined(__ELF__) && defined(__linux__)
+.section .note.GNU-stack,"",%progbits
+#endif
+
+#if defined(__ELF__) && defined(__CET__) && defined(__has_include)
+#if __has_include(<cet.h>)
+#include <cet.h>
+#endif
+#endif
+
+#if !defined(_CET_ENDBR)
+#define _CET_ENDBR
+#endif
+
+.intel_syntax noprefix
+.global _blake3_hash_many_avx2
+.global blake3_hash_many_avx2
+#ifdef __APPLE__
+.text
+#else
+.section .text
+#endif
+        .p2align  6
+        mov r10d, dword ptr [rsp+0x110+8*rax]
+        jc 3f
+        nop
+        nop
+3:
+        nop
+        nop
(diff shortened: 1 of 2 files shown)


subprojects/packagefiles/blake3/unix_syntax_check.S
```diff
@@ -0,0 +1,30 @@
...
+#ifdef __APPLE__
+.text
+#else
+.section .text
+#endif
+        .p2align  6
+        mov r10d, dword ptr [rsp+0x110+8*rax]
+        jc 3f
+        nop
+		nop
```
XVilka: Is broken alignment on purpose?

XVilka (changes requested):

subprojects/packagefiles/blake3/unix_syntax_check.S
thestr4ng3r: No, just me typing tabs, fixed.

XVilka (approved):
LGTM if CI will be green

wargio:
I throw it just in a simpler way. how about we just have an option? we try with auto, otherwise we provide other optional values and we set macos on a specific config if with clang version X.Y ?

thestr4ng3r:
> I throw it just in a simpler way. how about we just have an option? we try with auto, otherwise we provide other optional values and we set macos on a specific config if with clang version X.Y ?

Sorry I don't understand at all what you mean.

wargio:
> Sorry I don't understand at all what you mean.

Just an option like `meson -Dblake_cpufeat=avx2` and when is not set (i.e. on `auto`) we try to detect like you did in the code.

thestr4ng3r:
Should be possible. But is it really worth it?
If the build with auto works, I don't think there is any good reason to explicitly specify only a strict subset of possible simd implementations (that would be compiled by auto).
If the build with auto fails, specifying `meson -Dblake_cpufeat=` would be an easy workaround, but at the same time it means there is a bug and we should just fix our checking code.

PR #3788, opened 2023-08-25
Refactor the pointer command to use the API

### LINKED ISSUE #3780: Pointer read with `*` command not working
### Work environment

| OS/arch/bits (mandatory)                             | Android/Termux arm 64
| Architecture/bits of the file (mandatory)            | arm 64
| `rizin -v` full output, **not truncated** (mandatory)  | rizin 0.7.0 @ linux-arm-64 commit: 03dd2d222aa217346d81dcc8e6fcaf966bc2eb30

### Expected behavior

The pointer `*` command behaves as documented (at least for reads, I didn't attempt writes):

```
*entry0             # read the value contained at the
                      entrypoint
```

### Actual behavior

```
[0x00001d90]> *entry0
ERROR: Command '?v' does not exist.
```

Since `rz_pointer_handler` just execs a `?v` command for reads, I also tried directly using `%v` (since IIUC `%` replaces radare2's `?` in rizin):

```
[0x00001d90]> %v entry0
0x1d90
[0x00001d90]> %v *entry0
0x0
[0x00001d90]> pxw 4 @ entry0
0x00001d90  0xd503249f
```

For this binary I expected `*entry0` to return `0xd503249f` (or a 64-bits value, whatever).

Any binary exhibits the (presumably) faulty behavior. For the above snippets I used the `rizin` binary itself. No other action was taken in rizin prior to running the above commands.

No rizinrc was in use for the above snippets.

laomaiweng:
- [X] I've updated the [rizin book](https://github.com/rizinorg/book) with the relevant information (if needed)

**Detailed description**

See issue #3780. The `*` command still uses `rz_core_cmdf`, and the read case even invokes an no-longer-existent command (`?v` instead of `%v`).

This PR fixes the `*` command for read (and I'll also try to do writes shortly) by using the API.

This is a fix, no documentation or book updates needed.

**Test plan**

I've included a _pointer read_ test case in the newly-created `test/db/cmd/cmd_pointer`.

**Closing issues**

closes #3780

### DIFF
--- a/librz/core/cmd/cmd.c
+++ b/librz/core/cmd/cmd.c
@@ -908,12 +908,24 @@ RZ_IPI RzCmdStatus rz_push_escaped_handler(RzCore *core, int argc, const char **
 	return res;
 }
 
+static RzCmdStatus pointer_read(RzCore *core, const char *expr) {
+	ut64 n = rz_num_math(core->num, expr);
+	if (core->num->dbz) {
+		RZ_LOG_ERROR("core: RzNum ERROR: Division by Zero\n");
+		return RZ_CMD_STATUS_ERROR;
+	}
+	if (!rz_io_read_i(core->io, n, &n, core->rasm->bits / 8, core->print->big_endian)) {
+		return RZ_CMD_STATUS_ERROR;
+	}
+	rz_cons_printf("0x%" PFMT64x "\n", n);
+	return RZ_CMD_STATUS_OK;
+}
+
 RZ_IPI RzCmdStatus rz_pointer_handler(RzCore *core, int argc, const char **argv) {
 	int ret;
 	switch (argc) {
 	case 2:
-		ret = rz_core_cmdf(core, "?v [%s]", argv[1]);
-		return rz_cmd_int2status(ret);
+		return pointer_read(core, argv[1]);
 	case 3:
 		if (rz_str_startswith(argv[2], "0x")) {
 			ret = rz_core_cmdf(core, "wv %s @ %s", argv[2], argv[1]);
--- a/librz/core/core.c
+++ b/librz/core/core.c
@@ -576,21 +576,10 @@ static ut64 num_callback(RzNum *userptr, const char *str, int *ok) {
 		if (ok) {
 			*ok = 1;
 		}
-		ut8 buf[sizeof(ut64)] = RZ_EMPTY;
-		(void)rz_io_read_at(core->io, n, buf, RZ_MIN(sizeof(buf), refsz));
-		switch (refsz) {
-		case 8:
-			return rz_read_ble64(buf, core->print->big_endian);
-		case 4:
-			return rz_read_ble32(buf, core->print->big_endian);
-		case 2:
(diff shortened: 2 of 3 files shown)


ret2libc:
@thestr4ng3r @wargio Any idea why this comment?

```
// this is wrong, there is more than big and little endian
RZ_API bool rz_io_read_i(RzIO *io, ut64 addr, ut64 *val, int size, bool endian) {
	ut8 buf[8];
	rz_return_val_if_fail(io && val, false);
	size = RZ_DIM(size, 1, 8);
	if (!rz_io_read_at(io, addr, buf, size)) {
		return false;
	}
	// size says the number of bytes to read transform to bits for rz_read_ble
	*val = rz_read_ble(buf, endian, size * 8);
	return true;
}
```
This code does not seem to be used too much around the codebase.

laomaiweng:
> Any idea why this comment?

(FWIW I thought this was because _rz_endian.h_ also supports middle-endian?)

ret2libc:
Yeah that's one thing I noticed too. Maybe it's just for that.

XVilka (changes requested):
The last case fails:
```
[XX] db/cmd/feat_arithmetic pointer read
RZ_NOPLUGINS=1 /usr/bin/rizin -escr.utf8=0 -escr.color=0 -escr.interactive=0 -eflirt.sigdb.load.system=false -eflirt.sigdb.load.home=false -N -Qc 'wv8 0xdeadbeefdeadbeef
*0
*0 @e:asm.bits=64
*0 @e:asm.bits=32
*0 @e:asm.bits=16
*0 @e:asm.bits=8
' malloc://1024
-- stdout
--- expected
+++ actual
@@ -2,4 +2,4 @@
 0xdeadbeefdeadbeef
 0xdeadbeef
 0xbeef
-0xef
+0xdeadbeefdeadbeef
```

laomaiweng:
> The last case fails

Yeah as mentioned I don't know why this happens:

> The test case does **not** pass currently, being broken for `asm.bits=8`, for reasons I can't fathom. I debugged it and upon calling `pointer_read`, `asm.bits` is still 64 instead of the expected 8. This could be a separate bug, I don't know. Please advise.

Rot127:
@laomaiweng Didn't looked too deep into it. But try to set a break point in `rz_asm_set_bits()` and check if it gets set. It maybe doesn't, because the default arch doesn't support 8bits and `has_bits()` returns false.

laomaiweng:
@Rot127 Yes that was it, thanks! I've amended the test to use an 8-bit compatible arch. Should pass fine now.

Rot127:
Nice! While you are at it. Could you please add a warning if the setting of bits fails? A simple `RZ_LOG_WARN(...);` should do it. For the setting I mean. Not within the `rz_asm_set_bits()` function.

laomaiweng:
@Rot127 Not sure what you mean, I'm not calling `rz_asm_set_bits` directly? The test just runs with `rz-test` and uses `@e:asm.bits=<8|16|32|64>`.

XVilka (approved):

PR #3793, opened 2023-08-26
Refactor pointer write command `*` to use the API

laomaiweng:

**Detailed description**

This PR follows #3788 and refactors the write path for the `*` pointer command, so that it uses the API instead of `rz_core_cmdf`.

I also went a bit further and changed the pointer command's behavior wrt. the value to write:

* previously, the value would be interpreted as a word if it started with _0x_, or as a hexstring otherwise
* with this PR, the value will be interpreted as a hexstring if it's a valid hexstring, or as an expr (eval'ed with `rz_num_math` to a word of the current bitness) otherwise

This doesn't change the behavior for hexstrings, but allows more elaborate commands such as `*entry0=$$` or `*foo=[bar]` (which would previously have been rejected as invalid hexstring values).

This new behavior is fully contained in the last commit of the PR (along with all required changes). I can discard it if you feel it's not appropriate.

**Test plan**

I expanded the `cmd_pointer` regression test file to cover pointer writes.

**Closing issues**

N/A

### DIFF
--- a/librz/core/cmd_descs/cmd_descs.yaml
+++ b/librz/core/cmd_descs/cmd_descs.yaml
@@ -124,17 +124,15 @@ commands:
     cname: pointer
     summary: Pointer read/write data/values
     description: >
-      Read or write values at a given address. When the value starts with `0x`,
-      a 4-bytes value or 8-bytes value is written in the memory at address,
-      depending on the size of the value. When value does not start with `0x`
-      an hexstring with arbitrary length is expected and it is written starting
-      from the specified address.
-    args_str: "<addr>[=<0xvalue>|<hexstring>]"
+      Read or write values at a given address. When the value is a hexstring,
+      it is decoded and written as raw bytes. Otherwise, it is evaluated as an
+      expression and written using the current endianness and bitness settings.
+    args_str: "<addr>[=<expr>|<hexstring>]"
     args:
       - name: addr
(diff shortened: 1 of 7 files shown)


librz/core/cmd_descs/cmd_descs.yaml
```diff
@@ -144,7 +142,7 @@ commands:
             comment: write trap in entrypoint
           - text: "*"
             arg_str: "entry0+10=0x804800"
-            comment: write 0x804800 as a 4-byte value at 10 bytes from the entrypoint
+            comment: write 0x804800 with the current bitness, 10 bytes from the entrypoint
```
wargio, line 145: probably is worth adding an example to write with different bitness.

laomaiweng:
Amended the last commit to update the command description files from YAML.

librz/util/hex.c
```diff
@@ -551,10 +551,10 @@ RZ_API st64 rz_hex_bin_truncate(ut64 in, int n) {
 }
 
 // Check if str contains only hexadecimal characters and return length of bytes
```
XVilka: Convert this comment to Doxygen and update it please

XVilka (changes requested):
LGTM when feedback is addressed

laomaiweng:
@XVilka @wargio Applied, thanks for the feedback.

XVilka (approved):

PR #3794, opened 2023-08-26
Restore original seek offset after iterating over flags with `@@f`

laomaiweng:

**Detailed description**

Using the flags iterator `@@f` does not restore the original seek offset after processing the last flag.
This PR simply fixes that. No doc changes needed.

To avoid bloat in the _test_ directory, I didn't add a regression test as this doesn't seem like something that's likely to ever regress in isolation, but I can do so if needed.

**Test plan**

Before:

```
$ rizin malloc://1024
[0x00000000]> f foo @ 16
[0x00000000]> %v $$ @@f
0x10
[0x00000010]> s
0x10
```

After:

```
$ rizin malloc://1024
[0x00000000]> f foo @ 16
[0x00000000]> %v $$ @@f
0x10
[0x00000000]> s
0x0
```

**Closing issues**

N/A

### DIFF
--- a/librz/core/cmd/cmd.c
+++ b/librz/core/cmd/cmd.c
@@ -4392,6 +4392,7 @@ DEFINE_HANDLE_TS_FCN_AND_SYMBOL(iter_flags_stmt) {
 	if (!ts_node_is_null(arg)) {
 		arg_str = ts_node_handle_arg(state, node, arg, 1);
 	}
+	ut64 offorig = core->offset;
 	const RzSpace *flagspace = rz_flag_space_cur(core->flags);
 	RzFlagItem *flag;
 	RzListIter *iter;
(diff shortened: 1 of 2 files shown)


laomaiweng:
Hmm not sure what's going on with all the Golang strings analysis failures. 🤔

wargio:
> Hmm not sure what's going on with all the Golang strings analysis failures. 🤔

On the core/golang.c we actually call `rz_core_cmd0(core, "aF @@f:sym.go.*");` and for some weird reasons the last address was used also to start recovering the strings.

XVilka:
We also should use the API for that Go scan as well BTW...

laomaiweng:
Ok so it looks like just doing `s main` before the `pdr~"` command should fix the failing tests: after `@@f:sym.go.*`, the seek offset ended up in `main` previously, which is no longer the case in this PR.

I'll also take a shot at using the API instead.

XVilka (approved):
LGTM. The API change should be in a separate PR.

XVilka:
@wargio take look please, and if is good - merge.

wargio (approved):
that does make sense.

PR #3823, opened 2023-09-02
Fix use after free in pyc due missing return while error handling

ifyGecko:

**Detailed description**

Found a Use After Free bug in the pyc format code, if ret-> data is null we free ret but instead of returning we progress onward which will cause the now free'd ret to dereference its data field for the call to add_string_to cache.


wargio (approved):

wargio:
Looks correct. nice find!

XVilka (approved):

PR #3838, opened 2023-09-08
Serialize seek history

laomaiweng:

**Detailed description**

This PR implements seek history serialization/deserialization to/from the Rizin project file.

This is something that I feel is worth preserving across sessions, as seek history is often the logical path the user took to reach some location within the binary. I know in my case I often use it as a lifeline to go back up when I go down some rabbit hole in a binary, and having that lifeline lost if I close Rizin is annoying, hence the PR.

Since I'm new to this code base, there are a number of implementation details that I feel I should mention:

* SDB format
  * the seek history items are serialized as SDB entries with the item index as the entry key, which means the entries
    * may not be ordered numerically by key in the SDB (I figured that doesn't really matter)
    * have to be sorted numerically upon load before parsing the items into the seek history
  * this could have been avoided by storing the items as a JSON array but that felt less legible
  * also using the SDB entry key for a numerical sort means the sort will behave badly if (for some reason, e.g. hand-crafted project file) some key isn't a number, but I figured if the project file is malformed it's not my problem as long as rizin doesn't crash
* current offset
  * the current seek offset is saved as part of the seek history (with key 0 & `"is_current": true`), which is somewhat redundant with the `core/offset` entry (not fully redundant since the current seek history item also holds the cursor position which is otherwise not saved)
  * for now the offset of the current item in the seek history is just ignored upon load, I figured at some point in the future it would probably make sense to drop `core/offset` and use exclusively the offset from the current seek history item
* I added a new vector API for swapping two items in a vector, but it's only used by the seek history deserialization code so if that's not enough users for it to be an API just yet I can remove it

**Test plan**

Green CI, incl. added unit tests (for seek history serialization & the new `rz_vector_swap` API) and integration tests (for the bumped project version).

**Closing issues**

N/A


laomaiweng:
Updated the `db/cmd/project` test to account for the project version bump.

librz/core/serialize_core_seek.c
```diff
@@ -0,0 +1,199 @@
+// SPDX-FileCopyrightText: 2023 Quentin Minster <quentin@minster.io>
+// SPDX-License-Identifier: LGPL-3.0-only
+
+#include <rz_util/rz_serialize.h>
+#include <rz_core.h>
+
+RZ_API void rz_serialize_core_seek_save(RZ_NONNULL Sdb *db, RZ_NONNULL RzCore *core) {
```
wargio: Add doxygen.

librz/core/serialize_core_seek.c
```diff
@@ -0,0 +1,199 @@
...
+	SEEK_FIELD_OFFSET,
+	SEEK_FIELD_CURSOR,
+	SEEK_FIELD_CURRENT
+};
+
+/**
+ * \brief Create and initialize a JSON parser for seek history items
+ * \return a RzKeyParser* for seek history items
+ **/
+RzKeyParser *seek_parser_new(void) {
```
wargio: Missing `static`

librz/core/serialize_core_seek.c
```diff
@@ -0,0 +1,199 @@
...
+
+static int __cmp_num_asc(const void *a, const void *b) {
+	const SdbKv *ka = a, *kb = b;
+	// Parse as signed ints but don't bother witb error detection, it'll sort bad and that's it
+	long ia = strtol(sdbkv_key(ka), NULL, 10);
+	long ib = strtol(sdbkv_key(kb), NULL, 10);
+	return ia > ib;
+}
+
+RZ_API bool rz_serialize_core_seek_load(RZ_NONNULL Sdb *db, RZ_NONNULL RzCore *core, RZ_NULLABLE RzSerializeResultInfo *res) {
```
wargio: Add doxygen comment

librz/include/rz_core.h
```diff
@@ -1279,6 +1279,24 @@ RZ_API void rz_serialize_core_save(RZ_NONNULL Sdb *db, RZ_NONNULL RzCore *core,
 RZ_API bool rz_serialize_core_load(RZ_NONNULL Sdb *db, RZ_NONNULL RzCore *core, bool load_bin_io,
 	RZ_NULLABLE const char *prj_file, RZ_NULLABLE RzSerializeResultInfo *res);
 
+/**
+ * \brief Serialize seek history state and save to a sdb
+ *
+ * \param db sdb to save the state
+ * \param core RzCore instance to save from
+ */
```
wargio: move these in the `.c` file.

librz/util/vector.c
```diff
@@ -226,6 +226,16 @@ RZ_API void *rz_vector_push_front(RzVector *vec, void *x) {
 	return rz_vector_insert(vec, 0, x);
 }
 
+RZ_API void rz_vector_swap(RzVector *vec, size_t index_a, size_t index_b) {
+	rz_return_if_fail(vec && index_a < vec->len && index_b < vec->len);
+	char tmp[vec->elem_size];
```
wargio:
```suggestion
	ut8 tmp[vec->elem_size];
```

wargio (changes requested):
Looks good, just some small details.

laomaiweng:
Fixed these, thanks for the review!

laomaiweng:
Update for the `rz-bindgen` lint check failure.

wargio (approved):
LGTM

laomaiweng:
CI fails on MSVC because it doesn't support VLAs (in `rz_vector_swap`), can I `#ifdef` my way around it and use `alloca` instead on MSVC only ?

librz/core/serialize_core_seek.c
```diff
@@ -0,0 +1,213 @@
...
+			if (child->type != RZ_JSON_INTEGER) {
+				break;
+			}
+			seek_item.offset = child->num.u_value;
+			break;
+		case SEEK_FIELD_CURSOR:
+			if (child->type != RZ_JSON_INTEGER) {
+				break;
+			}
+			seek_item.cursor = child->num.s_value;
```
ret2libc: why `s_value` instead of `u_value` here?

librz/include/rz_vector.h
```diff
@@ -142,6 +142,13 @@ RZ_API void *rz_vector_push(RzVector *vec, void *x);
 // like rz_vector_insert for the beginning of vec
 RZ_API void *rz_vector_push_front(RzVector *vec, void *x);
 
+/**
+ * \brief Swap to elements of the vector
```
ret2libc:
```suggestion
 * \brief Swap two elements of the vector
```

librz/core/serialize_core_seek.c
laomaiweng: Because `cursor` in `RzCoreSeekItem` is an `int`, so I assumed it may be negative sometimes?

laomaiweng:
Update for typo & MSVC build failure.

librz/core/meson.build
```diff
@@ -60,6 +60,7 @@ rz_core_sources = [
   #'rtr_shell.c',
   'seek.c',
   'serialize_core.c',
+  'serialize_core_seek.c',
```
thestr4ng3r: The seek serialization can be put into `serialize_core.c` too so it is all together.

test/prj/v15-seek-history.rzdb
```diff
@@ -0,0 +1,5847 @@
...
+/core
+blocksize=0x100
+offset=0x5af0
+
+/core/analysis
+
+/core/analysis/blocks
+
+/core/analysis/callables
+_Exit=func
```
thestr4ng3r: A lot of info that is not tested for in this file can be stripped away to reduce the file size. See for example `v7-zignatures.rzdb`, which contains almost only empty namespaces except where info is needed.

librz/util/vector.c
```diff
@@ -226,6 +230,21 @@ RZ_API void *rz_vector_push_front(RzVector *vec, void *x) {
...
 }
 
+RZ_API void rz_vector_swap(RzVector *vec, size_t index_a, size_t index_b) {
+	rz_return_if_fail(vec && index_a < vec->len && index_b < vec->len);
+#ifndef _MSC_VER
+	ut8 tmp[vec->elem_size];
+#else
+	// MSVC is a C++ compiler and doesn't support C99 VLAs
+	ut8 *tmp = _alloca(vec->elem_size);
+#endif
```
thestr4ng3r: I think it will be easier to just use `malloc` instead of dynamic stack here. A boolean return value will be necessary to indicate if the swap succeeded or failed because of a failed malloc.

thestr4ng3r (changes requested):
Overall a sensible addition and good code!

laomaiweng:
Updated with the suggested changes, thanks for the review!
I realize I forgot to even include the rationale for the change in the PR, I'll edit it to add that.

thestr4ng3r (approved):

laomaiweng:
Not sure what's wrong with the failed `db/formats/dmp/dmp` test, there's only been a single failure in CI and I can't reproduce locally. Perhaps a network error? It seems to be fetching a PDB.

XVilka (approved):

PR #3867, opened 2023-09-16
Various Breakpoint Changes

thestr4ng3r:
# DO NOT SQUASH

**Detailed description**

Fix dbg.hwbp config var:
The var was created as an int node, but read as boolean, which is always
false. Since it is used as boolean, creating it as such fits best.

Avoid unnecessary "hw breakpoints not yet supported" warning:
This warning could happen when debug calls into bp for restoring
software breakpoints and hardware breakpoints exist at the same time.
These do not need memory restoration so we can avoid calling into
rz_bp_restore_one() at all.

Add specifiable size for watchpoints and tests:
rz_bp_size_at() is not always the right size for a watchpoint, so we add
the possibility to specify the size as an extra argument to dmw.

**Test plan**

For dbg.hwbp:

Enter debug, set `dbg.hwbp=1` and try to set a breakpoint with `db`. It should be created as a hw breakpoint, not sw.
This becomes especially apparent in e.g. https://git.sr.ht/~thestr4ng3r/rz-commodore with vice:// where sw breakpoints are not supported.

For watchpoint size:

Set a watchpoint with `dbw <perm> <size>`. It should then have the specified size

### DIFF
--- a/librz/core/cconfig.c
+++ b/librz/core/cconfig.c
@@ -3340,7 +3340,7 @@ RZ_API int rz_core_config_init(RzCore *core) {
 
 	SETCB("dbg.libs", "", &cb_dbg_libs, "If set stop when loading matching libname");
 	SETBPREF("dbg.skipover", "false", "Make dso perform a dss (same goes for esil and visual/graph");
-	SETI("dbg.hwbp", 0, "Set HW or SW breakpoints");
+	SETB("dbg.hwbp", false, "Use hardware instead of software breakpoints by default");
 	SETCB("dbg.unlibs", "", &cb_dbg_unlibs, "If set stop when unloading matching libname");
 	SETCB("dbg.verbose", "false", &cb_dbg_verbose, "Verbose debug output");
 	SETBPREF("dbg.slow", "false", "Show stack and regs in visual mode in a slow but verbose mode");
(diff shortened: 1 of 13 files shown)


librz/core/cconfig.c
```diff
@@ -3340,7 +3340,7 @@ RZ_API int rz_core_config_init(RzCore *core) {
 
 	SETCB("dbg.libs", "", &cb_dbg_libs, "If set stop when loading matching libname");
 	SETBPREF("dbg.skipover", "false", "Make dso perform a dss (same goes for esil and visual/graph");
-	SETI("dbg.hwbp", 0, "Set HW or SW breakpoints");
+	SETB("dbg.hwbp", false, "Set HW or SW breakpoints");
```
XVilka: Maybe "use" instead of "set"?

XVilka (approved):

librz/core/cconfig.c
thestr4ng3r: What about "Use hardware instead of software breakpoints by default"? I am planning to automatically fallback to hardware breakpoints if software fails (bc not possible on a certain arch for example) on `dbg.hwbp=false` and then this description would not contradict that.

XVilka: Sounds good too.

PR #3890, opened 2023-09-26
Make it clear the "details" of command are available

### LINKED ISSUE #3879: Make it clear the "details" of command are available.
There is no way for user to infer that there is a "details" help message available just looking at the normal help message. This somehow should be reflected in the corresponding help messages or any other way:
```
rizin /bin/ls
 -- If you want to open the file in read-write mode, invoke rizin with '-w'

[0x100003a38]> w?
Usage: w[?]   # Write commands
| w <string>        # Write string
| wB[-]             # Set or unset bits with given value
| wv[1248]          # Write value of given size
| w0 <len>          # Write <len> bytes with value 0x00
| w<1248><+-> [<n>] # increment/decrement byte, word, ..
| w6<de>            # Write base64 [d]ecoded or [e]ncoded string
| we<nsx>           # Extend write operations (insert bytes instead of replacing)
| wu <file>         # Apply unified hex patch (see output of cu)
| wr <len>          # Write <len> random bytes
| wc[j*-+ip?]       # Write cache commands
| wz <string>       # Write zero-terminated string
| wf[xfs]           # Write data from file, socket, offset
| ww <string>       # Write wide 16 little endian string
| wx[f]             # Write hexadecimal data
| wa[ifo]           # Write opcodes
| wb <hex>          # Write in current block an hexstring cyclically
| wm[-]             # Set binary mask hexpair to be used as cyclic write mask
| wo<?>             # Write a block with a special operation
| wD[/]             # Write De Bruijn Pattern
| wd <src> <len>    # Duplicate <len> bytes from <src> offset to current seek
| ws <string>       # Write 1 byte for length and then the string

[0x100003a38]> w??
Usage: w <string>   # Write string

Examples:
| w 123\n    # Write the chars '1', '2', '3' and a newline
| w ab\0cd\0 # Write the chars 'a', 'b', a NUL, 'c', 'd' and another NUL

Escape sequences:
| \0   # NUL (0x0)
| \a   # Bell (0x7)
| \b   # Backspace (0x8)
| \e   # Escape (0x1b)
| \f   # Form feed (0xc)
| \n   # Newline (0xa)
| \r   # Carriage return (0xd)
| \t   # Tab (0x9)
| \v   # Vertical tab (0xb)
| \\   # Backslash ('\')
| \xhh # Byte in hexadecimal
| \nnn # Byte in octal (eg. \033 for the escape char)
[0x100003a38]>
```

byteninjaa0:
- [ yes] I've read the [guidelines for contributing](https://github.com/rizinorg/rizin/blob/master/DEVELOPERS.md) to this repository

**Detailed description**
now when we run the w? command it will show that w?? option is also available

...

**Test plan**

...

**Closing issues**

Closes #3879

### DIFF
--- a/librz/core/cmd/cmd_api.c
+++ b/librz/core/cmd/cmd_api.c
@@ -1140,7 +1140,28 @@ static void do_print_child_help(RzCmd *cmd, RzStrBuf *sb, const RzCmdDesc *cd, c
 }
 
 static void print_child_help(RzCmd *cmd, RzStrBuf *sb, RzCmdDesc *cd, size_t max_len, bool use_color) {
-	do_print_child_help(cmd, sb, cd, cd->name, cd->help->summary ? cd->help->summary : "", true, max_len, use_color);
+	void **it_cd;
+	bool check = false;
+
+	rz_cmd_desc_children_foreach(cd, it_cd) {
+		RzCmdDesc *child = *(RzCmdDesc **)it_cd;
+		if (!strcmp(child->name, cd->name)) {
+			check = true;
+		}
+	}
+
+	const char *summary = strdup(cd->help->summary);
+	size_t summary_len = strlen(summary);
+	char *newSummary = malloc(strlen(summary) + 32);
+
+	if (check) {
+		snprintf(newSummary, summary_len + 32, "%s (see %s?? for more details)", summary, cd->name);
+		summary = newSummary;
+	}
+
+	do_print_child_help(cmd, sb, cd, cd->name, summary ? summary : "", true, max_len, use_color);
+
+	free(newSummary);
 }
 
 static char *group_get_help(RzCmd *cmd, RzCmdDesc *cd, bool use_color) {
--- a/librz/core/cmd_descs/cmd_descs.c
+++ b/librz/core/cmd_descs/cmd_descs.c
@@ -16679,7 +16679,7 @@ static const RzCmdDescArg write_args[] = {
 	{ 0 },
 };
 static const RzCmdDescHelp write_help = {
-	.summary = "Write string (see 'w??' for more help)",
+	.summary = "Write string",
 	.details = write_details,
 	.args = write_args,
 };
--- a/librz/core/cmd_descs/cmd_descs.c
+++ b/librz/core/cmd_descs/cmd_descs.c
@@ -16679,7 +16679,7 @@ static const RzCmdDescArg write_args[] = {
 	{ 0 },
 };
 static const RzCmdDescHelp write_help = {
-	.summary = "Write string",
+	.summary = "Write string (see 'w??' for more help)",
 	.details = write_details,
 	.args = write_args,
 };
--- a/test/unit/test_cmd.c
+++ b/test/unit/test_cmd.c
@@ -661,7 +661,7 @@ bool test_cmd_argv_state(void) {
 	RzCmdParsedArgs *pa = rz_cmd_parsed_args_newcmd("?");
 	char *h = rz_cmd_get_help(cmd, pa, false);
 	char *exp_h = "Usage: [.][times][cmd][~grep][@[@iter]addr][|>pipe] ; ...\n"
-		      "| z[jqJ] # group summary\n";
+		      "| z[jqJ] # group summary (see z?? for more details)\n";
 	mu_assert_streq(h, exp_h, "zj, zJ and zq are considered in the help");
 	free(h);
 	rz_cmd_parsed_args_free(pa);
@@ -709,7 +709,7 @@ bool test_cmd_group_argv_modes(void) {
 	RzCmdParsedArgs *pa = rz_cmd_parsed_args_newcmd("?");
 	char *h = rz_cmd_get_help(cmd, pa, false);
 	char *exp_h = "Usage: [.][times][cmd][~grep][@[@iter]addr][|>pipe] ; ...\n"
-		      "| z[jqd] # z group summary\n";
+		      "| z[jqd] # z group summary (see z?? for more details)\n";
 	mu_assert_streq(h, exp_h, "zd is considered in the help");
 	free(h);
 	rz_cmd_parsed_args_free(pa);


ret2libc (changes requested):
Hi and thanks for the contribution! Some comments on the PR:

- I don't think this is exactly what @XVilka meant in his bug. We would like something that would be generic and not specific to the `w` command. There are many other commands that have `??` available and we want to show a similar help message for all of them. That's why we have [rzshell](https://github.com/rizinorg/rizin/blob/dev/doc/rzshell.md)
- these `cmd_descs.c`/`cmd_descs.h` files, as you can read at the head of the files, are automatically generated from the `.yaml` files you can find at https://github.com/rizinorg/rizin/tree/dev/librz/core/cmd_descs . You should modify the .yaml files and re-build the project (this will update the `cmd_descs.c`/`cmd_descs.h` files).
- However, since this change should be done for all commands that have some "details" available, I believe this change should be done somewhere in cmd_api.c

byteninjaa0:
show i do the same for other commands too?

ret2libc:
> show i do the same for other commands too?

Yes but in a general way. You should not modify the summary of each command manually.

librz/core/cmd_descs/cmd_descs.c
```diff
@@ -16679,7 +16679,7 @@ static const RzCmdDescArg write_args[] = {
 	{ 0 },
 };
 static const RzCmdDescHelp write_help = {
-	.summary = "Write string",
+	.summary = "Write string (see 'w??' for more help)",
```
wargio: Please modify the yaml file not, the generated c file.

librz/core/cmd/cmd_api.c
```diff
@@ -338,6 +338,9 @@ static RzCmdDesc *cmd_get_desc_best(RzCmd *cmd, const char *cmd_identifier, bool
 	while (*cmdid) {
 		RzCmdDesc *cd = ht_pp_find(cmd->ht_cmds, cmdid, NULL);
 		if (cd) {
+			if (cd->help->details || cd->help->details_cb) {
+      		cd->has_extended_details = true;
+    	}
```
wargio: format.

wargio:
please install pyaml and clang-format, then rebuild and run ./sys/clang-format.py before pushing.

byteninjaa0:
when i am run w? it is showing the message four times .

wargio (approved):

ret2libc (changes requested):
image

I don't like how this looks. 
IMHO something like `(%s?? for more details)` should be appended to the summary of the top command

image

ret2libc:
Please show a screenshot of how it looks like now!

byteninjaa0:
image

ret2libc:
I think it is not working as expected yet. Like, `wv?` does not show the message. Or `we?` says to use `we??` but actually only `wex` has details...

Now, I think given that when a command is a leaf (there are no subcommands) the details are automatically displayed (try `wex?`, for example), I think we can just show the message `<x>?? for more details` near the summary of the subcommand with the same name as the group name.

For example:
- `w` is both a group (`w?` displays the sub-commands) and a command (`w??` shows the detail of the `w` command). In this case, it would make sense to have in the `w?` screen the message `(w?? for more details)` near the `w` command.

- `we` is the name of a group (`we?` shows the sub-commands), but there is no sub-command with the same name (there is no `we` command), thus nothing should be shown.
- `wex` is a leaf command, thus nothing should be shown. When you do `wex?` it already shows the details
- `wv` is both a group (`wv?` displays the sub-commands) and a command (`wv?? shows the details of the `wv` command). In this case it would make sense to have in the `wv?` screen the message `(wv?? for more details)` near the `wv` command.

I hope it make sense.

byteninjaa0:
thank you for the detailed explanation, making a commit soon!

wargio:
what if we do something like tree view where if there are more subcommands to it we have a `+` instead of `|` at the beginning of the line?
@ret2libc @XVilka

ret2libc:
> what if we do something like tree view where if there are more subcommands to it we have a `+` instead of `|` at the beginning of the line? @ret2libc @XVilka

Interesting ideas... We'd need to see how it looks like, but in theory I don't mind it. It would make things clearer immediately. However, it's a slightly different problem. The `+`/`|` says whether a command has some subcommands or not. The `(use <cmd>?? for more details)` tells you whether you should use `<cmd>??` to get more info about a subcommand with the same name as the name of a group (`w` group and `w` command, for example).

wargio:
> The `(use <cmd>?? for more details)` tells you whether you should use `<cmd>??` to get more info about a subcommand with the same name as the name of a group (`w` group and `w` command, for example).

The issue i see is that this might make the command line long.

byteninjaa0:
Screenshot 2023-10-22 115942
Screenshot 2023-10-22 120009
Screenshot 2023-10-22 120033
Screenshot 2023-10-22 120056

librz/core/cmd/cmd_api.c
```diff
@@ -1059,13 +1058,13 @@ static void fill_usage_strbuf(RzCmd *cmd, RzStrBuf *sb, RzCmdDesc *cd, bool use_
...
 		rz_cmd_desc_children_foreach(cd, it_cd) {
-		RzCmdDesc *child = *(RzCmdDesc **)it_cd;
-			if (child->help->details && !already_printed) {
-				rz_strbuf_appendf(summary_sb," (%s?? for more details)", cd->name);
-				already_printed=true;
-			
-			}
-		}
+      		RzCmdDesc *child = *(RzCmdDesc **)it_cd;
+      		if (strcmp(child->name, cd->name) == 0) {
```
wargio:
```suggestion
      		if (!strcmp(child->name, cd->name)) {
```

wargio (changes requested):
run clang format (requires clang format 16 and then run `./sys/clang-format.py`)

ret2libc:
Sorry for requesting changes again :(

> Now, I think given that when a command is a leaf (there are no subcommands) the details are automatically displayed (try `wex?`, for example), I think we can just show the message `<x>?? for more details` near the summary of the subcommand with the same name as the group name.

As mentioned above, I think it makes more sense to have the `<x>?? for more details` message near the *subcommand* and not near the group summary. Right now you are displaying the msg in the summary of the command group, but I think it makes more sense on the *subcommand*.

byteninjaa0:
No problem I'll make a commit as soon as I figure out the solution btw thanks for your support and patience.

XVilka:
@byteninjaa0 have you had a chance to tackle this?

byteninjaa0:
image
it is not show the "see w?? for more details " for w subcommand other than that its working fine i guess.

XVilka:
You need to update the unit test:
```
 test_cmd_argv_state ERR
[XX] Fail at line 665: zj, zJ and zq are considered in the help: expected Usage: [.][times][cmd][~grep][@[@iter]addr][|>pipe] ; ...
| z[jqJ] # group summary
, got Usage: [.][times][cmd][~grep][@[@iter]addr][|>pipe] ; ...
| z[jqJ] # group summary (see z?? for more details)
.

test_cmd_group_argv_modes ERR
[XX] Fail at line 713: zd is considered in the help: expected Usage: [.][times][cmd][~grep][@[@iter]addr][|>pipe] ; ...
| z[jqd] # z group summary
, got Usage: [.][times][cmd][~grep][@[@iter]addr][|>pipe] ; ...
| z[jqd] # z group summary (see z?? for more details)
.
```

byteninjaa0:
same test is failed again i am seeing into the issue i think we have to make our unit test flexible for the edited summary of subcommand

byteninjaa0:
is there anything else left in this?

XVilka:
> is there anything else left in this?

green CI

XVilka (approved):

PR #3910, opened 2023-10-04
Move Hashtable and Set primitives to RzUtil

### LINKED ISSUE #3894: Move Hashtable and Set primitives from SDB to RzUtil
Currently they are located in these files:
- `librz/util/sdb/src/ht_*.{c,h}`
- `librz/util/sdb/src/set.{c,h}`

They should be moved upward, to `librz/util/` and properly documented with Doxygen.

vanshjangir:

**Detailed description**

Move Hashtable from `librz/util/sdb/src/ht_*.c` to `librz/util/ht` and 
`librz/util/sdb/src/ht_*.h` to `librz/include/rz_util` and
Set primitives from `librz/util/sdb/src/set.{c,h}` to `librz/util/`.
Update `librz/util/meson.build` , `librz/util/sdb/src/meson.build` and `librz/include/meson.build` for changed hashtables and set primitives file paths.

**Test plan**

...

**Closing issues**

Closes https://github.com/rizinorg/rizin/issues/3894


librz/util/meson.build
```diff
@@ -112,7 +117,7 @@ else
   rz_util_sources += ['big.c']
 endif
 
-rz_util_includes = [platform_inc, include_directories('sdb/src', 'sdb')]
+rz_util_includes = [platform_inc, include_directories('sdb/src', 'sdb', '.')]
```
ret2libc: Is this really necessary?

ret2libc:
I thought we would have encountered some issues in the moving, but it seems all is fine, nice :D

librz/util/meson.build
vanshjangir: without this ninja -C build test is failing. I guess the failure is because the path to hashtable ht_*.h is not specified. So I added '.'

XVilka:
I wonder if it makes sense to put them into a subdirectory since there are many files? Maybe something like `librz/util/ht/`?

vanshjangir:
Should I move only hastables to a new subdirectory `librz/util/ht` or move both hashtables and set primitives

XVilka (changes requested):
Better to move header files to `librz/include/rz_util/` though.

librz/include/meson.build
```diff
@@ -117,6 +117,7 @@ rz_util_files = [
   'rz_util/rz_utf8.h',
   'rz_util/rz_version.h',
   'rz_util/rz_x509.h',
+
```
XVilka: Unnecessary change

XVilka: But you need to install them too in fact.

XVilka (changes requested):

librz/include/meson.build
vanshjangir: Do I need to install the hashtables in `librz/include/meson.build`. Because the hashtables path is present in `librz/util/meson.build`

XVilka: After you move includes with exporting APIs to the `librz/include/rz_util`, of course.

librz/util/meson.build
```diff
@@ -112,7 +117,7 @@ else
   rz_util_sources += ['big.c']
 endif
 
-rz_util_includes = [platform_inc, include_directories('sdb/src', 'sdb')]
+rz_util_includes = [platform_inc, include_directories('sdb/src', 'sdb', '../include/rz_util', '.')]
```
ret2libc, line 120: I do not think these additions are needed. Why did you add them?

vanshjangir: when I was testing (ninja -C build test), some error were there, showing
`FAILED: librz/util/librz_util.so.0.7.0.p/astr.c.o` and 
`/librz/include/rz_list.h:5:10: fatal error: ht_pp.h: No such file or directory`
    `5 | #include <ht_pp.h>`
    
so i thought adding `'.'` and `../include/rz_util` will specify the path to look for the header files (ht_*.h and set.h)

wargio:
i'm ok with these changes, but tbh, i think the original headers should be internals and we should provide explicit function headers somewhere.

@XVilka @ret2libc @thestr4ng3r @kazarmy @Rot127

vanshjangir:
@wargio Can you describe in detail, if there is something left to change.

wargio:
> @wargio Can you describe in detail, if there is something left to change.

before you do this, i would prefer to see if the others are also ok with my suggestion. sorry. i should have had explicitly said this to avoid you to do extra useless work :/

ret2libc:
> i'm ok with these changes, but tbh, i think the original headers should be internals and we should provide explicit function headers somewhere.
> 
> @XVilka @ret2libc @thestr4ng3r @kazarmy @Rot127

Do you mean instead of having a generic ht_inc.h, just "unroll" the various macros and provide a regular .h? I was thinking about that, but I'm unsure... also, I think it is ok to merge as-is anyway, because it is no worse than what we have right now (those headers are just under the sdb directory right now).

XVilka:
> > i'm ok with these changes, but tbh, i think the original headers should be internals and we should provide explicit function headers somewhere.
> > @XVilka @ret2libc @thestr4ng3r @kazarmy @Rot127
> 
> Do you mean instead of having a generic ht_inc.h, just "unroll" the various macros and provide a regular .h? I was thinking about that, but I'm unsure... also, I think it is ok to merge as-is anyway, because it is no worse than what we have right now (those headers are just under the sdb directory right now).

I agree with @ret2libc, let's merge this as is first. Having just one HT header would be nice but it's a separate thing and can be done later.

vanshjangir:
The changes requested are done, so I thought I might ask if there is anything left to do.

XVilka (approved):
LGTM after that Include path problem is solved.

wargio:
> > i'm ok with these changes, but tbh, i think the original headers should be internals and we should provide explicit function headers somewhere.
> > @XVilka @ret2libc @thestr4ng3r @kazarmy @Rot127
> 
> Do you mean instead of having a generic ht_inc.h, just "unroll" the various macros and provide a regular .h? I was thinking about that, but I'm unsure... also, I think it is ok to merge as-is anyway, because it is no worse than what we have right now (those headers are just under the sdb directory right now).

exactly.

ret2libc (approved):

ret2libc (approved):

XVilka:
@vanshjangir could you please fix also this error during the rz-ghidra compilation?

https://github.com/rizinorg/rz-ghidra/issues/332

vanshjangir:
> @vanshjangir could you please fix also this error during the rz-ghidra compilation?
> 
> [rizinorg/rz-ghidra#332](https://github.com/rizinorg/rz-ghidra/issues/332)

@XVilka sure, I will fix that

Issue #3656, opened 2023-07-14
Show mangled and demangled names in symbols/flags


XVilka:
Currently there is no way to show both mangled and demangled symbols in the Rizin binary, from the shell.

The possible solution is to extend `is` command to show both:

```
rizin test/bins/demangle-test-cpp
[0x000010d0]> is~vector
46  0x000012fc 0x000012fc WEAK   FUNC   27       std::_Vector_base<int, std::allocator<int> >::_Vector_impl::~_Vector_impl()
47  0x00001a48 0x00001a48 WEAK   FUNC   51       std::_Vector_base<int, std::allocator<int> >::_M_allocate(unsigned long)
52  0x00001462 0x00001462 WEAK   FUNC   62       std::_Vector_base<int, std::allocator<int> >::_Vector_impl::_Vector_impl()
60  0x00001286 0x00001286 WEAK   FUNC   68       std::vector<int, std::allocator<int> >::~vector()
62  0x000012ca 0x000012ca WEAK   FUNC   49       std::vector<int, std::allocator<int> >::push_back(int&&)
67  0x00001a08 0x00001a08 WEAK   FUNC   64       __gnu_cxx::__normal_iterator<int*, std::vector<int, std::allocator<int> > >::difference_type __gnu_cxx::operator-<int*, std::vector<int, std::allocator<int> > >(__gnu_cxx::__normal_iterator<int*, std::vector<int, std::allocator<int> > > const&, __gnu_cxx::__normal_iterator<int*, std::vector<int, std::allocator<int> > > const&)
69  0x00001318 0x00001318 WEAK   FUNC   27       std::_Vector_base<int, std::allocator<int> >::_Vector_base()
70  0x000019c0 0x000019c0 WEAK   FUNC   72       std::vector<int, std::allocator<int> >::begin()
76  0x000013c8 0x000013c8 WEAK   FUNC   154      void std::vector<int, std::allocator<int> >::emplace_back<int>(int&&)
```
We should extend `is` to show also demangled names and/or `fl`/`fl.`:
```
[0x000010d0]> s 0x12fc
[0x000012fc]> fl.
0x000012fc 27 sym.std::_Vector_base_int__std::allocator_int___::_Vector_impl::__Vector_impl
0x000012fc 1 method.std::_Vector_base_int__std::allocator_int___::_Vector_impl.__Vector_impl
[0x000012fc]>
```

Issue #3543, opened 2023-06-01
Change how functions/flags shown in disassembly (`pd`, etc)


XVilka:
Currently we have `asm.demangle` config variable which forces (or should to do so) to show the demangled name inline instead of the flag, otherwise demangled name shown as a comment. I propose to expand this idea further:

1. Rename `asm.demangle` into something like `asm.inlinenames` to extend the concept of better names - not only demangled, but also shortened somehow, if necessary, e.g. without long unnecessary module paths, etc
2. When disabled, it should show flag name, while showing the demangled and/or beautified name in the comment
3. When enabled, it should show the demangled and simplified/beautified name inline, while not making any comment, also remove any duplication of the same information in other places (this is what I suspect most reversers would be expected to use)
4. Anything else I forgot?

It should substitute the name in:
- [ ] disassembly, e.g. after `CALL`/`J`/`BR`/etc opcodes
- [ ] CALL XREF

See also related/relevant config `asm.flags.real` variable.

ret2libc:
Mmm, is asm.demangle useless right now? I see it's only set but never read by any part of the code.
Moreover, I think what you wanted is already done by `asm.flags.real`, right?

PR #3915, opened 2023-10-11
Demangle flags symbols

ret2libc:

**Detailed description**

I'd like to have `asm.flags.real=true` by default, however one of the problems with having that enabled is that the users cannot easily know what is the flag name they should use to reference the symbol.

For example:
```
call  std::vector<int, std::allocator<int> >::vector()
```
It is not immediately clear that the flag name (which should be used in commands like `s`) is `method.std::vector_int__std::allocator_int___.vector`.

The idea is that we want the flag name to be displayed as comment (but only when the real names are displayed!):
```
-> asm.flags.real=true

call  std::vector<int, std::allocator<int> >::vector() ; method.std::vector_int__std::allocator_int___.vector
```
```
-> asm.flags.real=false
call  method.std::vector_int__std::allocator_int___.vector
```

These comments are not displayed in visual mode because the user can easily use the "jumpkeys"(the numbers near the calls e.g. `[1]`) to move between functions.

**Test plan**

See attached test.
Moreover, ensure that in visual mode the comments are not displayed.

**Closing issues**

...

### DIFF
--- a/librz/core/disasm.c
+++ b/librz/core/disasm.c
@@ -3418,8 +3418,10 @@ static void ds_print_fcn_name(RzDisasmState *ds) {
 			if (ds->core->vmode && (rz_str_startswith(flag->name, "sym.") || (flag_sym = rz_flag_get_by_spaces(ds->core->flags, ds->analysis_op.jump, RZ_FLAGS_FS_SYMBOLS, NULL))) && flag_sym->demangled) {
 				return;
 			}
-			ds_begin_comment(ds);
-			ds_comment(ds, true, "; %s", flag->name);
+			if (ds->core->flags->realnames && flag->realname) {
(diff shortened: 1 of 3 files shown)


wargio (approved):

XVilka:
See also these related bugs: https://github.com/rizinorg/rizin/issues/3656 and https://github.com/rizinorg/rizin/issues/3543

ret2libc:
@XVilka about the test, please see https://github.com/rizinorg/rizin/pull/3915/commits/98f94497c1b2ccfb120eb584e2a4c21999756e88 . Also, i shared some concern in mattermost, but i don't know anything about RISCV.

XVilka (approved):

PR #3919, opened 2023-10-12
Replace RZ_ASM_NOPLUGINS and RZ_BIN_NOPLUGINS with more general RZ_NOPLUGINS

Rot127:

**Detailed description**

...

**Test plan**

All green (since the variable is used for all tests, stuff will explode if something is wrong).

**Closing issues**

...

### DIFF
--- a/librz/main/rz-asm.c
+++ b/librz/main/rz-asm.c
@@ -203,7 +203,7 @@ static int rasm_show_help(int v) {
 		       " If '-l' value is greater than output length, output is padded with nops\n"
 		       " If the last argument is '-' reads from stdin\n");
 		printf("Environment:\n"
-		       " RZ_ASM_NOPLUGINS  do not load shared plugins (speedup loading)\n"
+		       " RZ_NOPLUGINS      do not load shared plugins (speedup loading)\n"
 		       " RZ_ASM_ARCH       same as rz-asm -a\n"
 		       " RZ_ASM_BITS       same as rz-asm -b\n"
 		       " RZ_DEBUG          if defined, show error messages and crash signal\n"
@@ -477,7 +477,7 @@ static int print_assembly_output(RzAsmState *as, const char *buf, ut64 offset, u
 }
 
 static void __load_plugins(RzAsmState *as) {
-	char *tmp = rz_sys_getenv("RZ_ASM_NOPLUGINS");
+	char *tmp = rz_sys_getenv("RZ_NOPLUGINS");
 	if (tmp) {
 		free(tmp);
 		return;
--- a/librz/main/rz-bin.c
+++ b/librz/main/rz-bin.c
@@ -216,8 +216,8 @@ static int rabin_show_help(int v) {
 	}
 	if (v) {
 		printf("Environment:\n"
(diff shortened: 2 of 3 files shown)


XVilka (approved):

XVilka:
@ret2libc take a look at this one please, and if you don't have any objections - merge.

wargio (approved):

librz/main/rz-asm.c
```diff
@@ -203,7 +203,7 @@ static int rasm_show_help(int v) {
 		       " If '-l' value is greater than output length, output is padded with nops\n"
 		       " If the last argument is '-' reads from stdin\n");
 		printf("Environment:\n"
-		       " RZ_ASM_NOPLUGINS  do not load shared plugins (speedup loading)\n"
+		       " RZ_NOPLUGINS  do not load shared plugins (speedup loading)\n"
```
ret2libc: does this display aligned?

ret2libc (commented):
Mh, If the others are ok i'm ok, though I would have gone with the opposite solution of keeping one NOPLUGIN for each kind of plugins.

Rot127:
@ret2libc 
> though I would have gone with the opposite solution of keeping one NOPLUGIN for each kind of plugins.

We can do this as well. Though, if we want to have it properly done, we would need to:
- Add checks for `RZ_NOPLUGIN` where ever the other `NOPLUGINS` are.
- Define somewhere what a bin plugin and what an asm plugin is.

Idk, is it worth the trade? We can also open an issue about it for a better plugin loading module.
This is maybe even a nice task for beginners with C knowledge.

Something where each plugin needs to set its category and we have a simply call to enable a certain plugin

```c
load_plugin(plugin) { if (do_not_load_plugin_group(X) return; .... }
```

ret2libc:
I'm not sure I follow you.

> Define somewhere what a bin plugin and what an asm plugin is.

What do you mean? We already have separate bin plugins and separate asm plugins.

Rot127:
> What do you mean? We already have separate bin plugins and separate asm plugins.

Sorry, I was not clear with this.

The flag `BIN/ASM_NOPLUGINS` suggests that only ASM/BIN plugins don't get loaded. But when the flag is checked and is set, no plugin is loaded at all (at the location of the env check).
Also there are no checks for `RZ_HASH_NOPLUGINS` or `RZ_ANALYSIS_NOPLUGINS`, although plugins of this type are loaded as well.

A more general problem is, plugin libraries get loaded via `rz_lib_opendir()`. But this loads always all libraries from this directory (at least this is what the doc string says).
Though the plugin is only usable if the handler was added via `rz_lib_add_handler()`.

This appears inconsistent to me and hard to manage, because there is no single module which does the plugin loading.
`RZ_NOPLUGINS` (or `RZ_XXX_NOPLUGINS`) should be only checked at a single location in the code base IMHO. So if for example ASM plugins should be loaded, a single plugin load handler is called, it checks if `RZ_ASM_NOPLUGINS` or `RZ_NOPLUGINS` is set, loads only the needed libs and adds only the needed handlers.

ret2libc:
I understand now, thanks for the explanation! If so, yes then it makes sense how you did it :) Thanks for the patience :D

Issue #3775, opened 2023-08-21
Switch from `RzList` to `RzPVector` in `RzBinPlugin`


XVilka:
Currently it uses a lot of `RzList` which is slower to traverse and takes more memory. The idea is to substitute them for `RzPVector` instead.
These are members that could be switched:
- [x] `RzList /*<RzBinImport *>*/ *(*imports)(RzBinFile *bf);`
- [x] `RzList /*<RzBinSymbol *>*/ *(*symbols)(RzBinFile *bf);`
- [x] `RzList /*<RzBinString *>*/ *(*strings)(RzBinFile *bf);`
- [x] `RzList /*<RzBinVirtualFile *>*/ *(*virtual_files)(RzBinFile *bf);`
- [x] `RzList /*<RzBinMap *>*/ *(*maps)(RzBinFile *bf);`
- [ ] `RzList /*<RzBinAddr *>*/ *(*entries)(RzBinFile *bf);`
- [x] `RzList /*<RzBinSection *>*/ *(*sections)(RzBinFile *bf);`
- [x] `RzList /*<RzBinField *>*/ *(*fields)(RzBinFile *bf);`
- [x] `RzList /*<char *>*/ *(*libs)(RzBinFile *bf);`
- [x] `RzList /*<RzBinReloc *>*/ *(*relocs)(RzBinFile *bf);`
- [x] `RzList /*<RzBinTrycatch *>*/ *(*trycatch)(RzBinFile *bf);`
- [x] `RzList /*<RzBinClass *>*/ *(*classes)(RzBinFile *bf);`
- [x] `RzList /*<RzBinMem *>*/ *(*mem)(RzBinFile *bf);`
- [x] `RzList /*<RzBinReloc *>*/ *(*patch_relocs)(RzBinFile *bf);`
- [x] `RzList /*<RzBinFileHash *>*/ *(*hashes)(RzBinFile *bf);`
- [x] `RzList /*<RzBinResource *>*/ *(*resources)(RzBinFile *bf);`

XVilka:
@PeiweiHu hi, do you plan to continue working on this?

PeiweiHu:
> @PeiweiHu hi, do you plan to continue working on this?

I just came across this message since the last month was too busy. I will continue working on this and also other stuffs these days.

Issue #3755, opened 2023-08-14
Remove use of deprecated `rz_bin_get_*()` APIs


XVilka:
```c
// XXX: those accessors are redundant
RZ_DEPRECATE RZ_API RZ_BORROW RzList /*<RzBinAddr *>*/ *rz_bin_get_entries(RZ_NONNULL RzBin *bin) {
	rz_return_val_if_fail(bin, NULL);
	RzBinObject *o = rz_bin_cur_object(bin);
	return o ? (RzList *)rz_bin_object_get_entries(o) : NULL;
}

RZ_DEPRECATE RZ_API RZ_BORROW RzList /*<RzBinField *>*/ *rz_bin_get_fields(RZ_NONNULL RzBin *bin) {
	rz_return_val_if_fail(bin, NULL);
	RzBinObject *o = rz_bin_cur_object(bin);
	return o ? (RzList *)rz_bin_object_get_fields(o) : NULL;
}

RZ_DEPRECATE RZ_API RZ_BORROW RzList /*<RzBinImport *>*/ *rz_bin_get_imports(RZ_NONNULL RzBin *bin) {
	rz_return_val_if_fail(bin, NULL);
	RzBinObject *o = rz_bin_cur_object(bin);
	return o ? (RzList *)rz_bin_object_get_imports(o) : NULL;
}

RZ_DEPRECATE RZ_API RZ_BORROW RzBinInfo *rz_bin_get_info(RzBin *bin) {
	rz_return_val_if_fail(bin, NULL);
	RzBinObject *o = rz_bin_cur_object(bin);
	return o ? (RzBinInfo *)rz_bin_object_get_info(o) : NULL;
}

RZ_DEPRECATE RZ_API RZ_BORROW RzList /*<char *>*/ *rz_bin_get_libs(RZ_NONNULL RzBin *bin) {
	rz_return_val_if_fail(bin, NULL);
	RzBinObject *o = rz_bin_cur_object(bin);
	return o ? (RzList *)rz_bin_object_get_libs(o) : NULL;
}

RZ_DEPRECATE RZ_API RZ_BORROW RzList /*<RzBinSection *>*/ *rz_bin_get_sections(RZ_NONNULL RzBin *bin) {
	rz_return_val_if_fail(bin, NULL);
	RzBinObject *o = rz_bin_cur_object(bin);
	return o ? (RzList *)rz_bin_object_get_sections_all(o) : NULL;
}
```

[...]

PR #3961, opened 2023-11-06
Refactor <libs> from RzList to RzPVector in RzBinPlugin

PeiweiHu:

**Detailed description**

...

**Test plan**

CI is green.

**Closing issues**

partially address https://github.com/rizinorg/rizin/issues/3775 and https://github.com/rizinorg/rizin/issues/3755

### DIFF
--- a/librz/bin/format/dex/dex.c
+++ b/librz/bin/format/dex/dex.c
@@ -1498,6 +1498,10 @@ RZ_API RZ_OWN RzPVector /*<RzBinImport *>*/ *rz_bin_dex_imports(RZ_NONNULL RzBin
 	return imports;
 }
 
+static int compare_strings(const void *a, const void *b) {
+	return strcmp((const char *)a, (const char *)b);
+}
+
 /**
  * \brief Returns a RzPVector<char*> containing the dex libraries
  */
@@ -1560,7 +1564,7 @@ RZ_API RZ_OWN RzPVector /*<char *>*/ *rz_bin_dex_libraries(RZ_NONNULL RzBinDex *
 			object = p;
 		}
 
-		if (rz_pvector_contains(libraries, object)) {
+		if (rz_pvector_find(libraries, object, compare_strings)) {
 			free(object);
 			continue;
 		}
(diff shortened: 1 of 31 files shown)


librz/bin/format/dex/dex.c
```diff
@@ -1564,12 +1560,12 @@ RZ_API RZ_OWN RzList /*<char *>*/ *rz_bin_dex_libraries(RZ_NONNULL RzBinDex *dex
 			object = p;
 		}
 
-		if (rz_list_find(libraries, object, compare_strings)) {
+		if (rz_pvector_contains(libraries, object)) {
```
ret2libc: does this work well? I think this just compares the pointers, but the old behaviour compares the content, so this may creates problems.

librz/bin/format/java/class_bin.c
```diff
@@ -1929,14 +1926,14 @@ RZ_API RZ_OWN RzList /*<char *>*/ *rz_bin_java_class_as_libraries(RZ_NONNULL RzB
 				// arg0 is name_index
 				tmp = java_class_constant_pool_stringify_at(bin, arg0);
 			}
-			if (tmp && !rz_list_find(list, tmp, compare_strings)) {
-				rz_list_append(list, tmp);
+			if (tmp && !rz_pvector_contains(vec, tmp)) {
```
ret2libc: same

librz/bin/format/dex/dex.c
wargio: yeah, this is wrong

wargio: looks like we don't have a real comparator. i suggest to rename `rz_pvector_contains` to `rz_pvector_contains_ptr` and have `rz_pvector_contains` using `RzPVectorComparator`

PeiweiHu: My fault. Will revise it later. Thanks.

wargio: don't forget to rename also the rz_vector_contains in the same way

PeiweiHu: Decide to use `rz_pvector_find` instead of `rz_pvector_contains` for comparator-based searching considering there is also `rz_list_contains` performing the similar semantic as the current `rz_pvector_contains`. What we want for `rz_pvector_contains` is actually similar to `rz_list_find`. What do you think?

PeiweiHu:
The broken test is weird. I cannot reproduce it.

librz/bin/format/dex/dex.c
wargio: sounds perfect.

wargio (approved):
LGTM <3

PR #3978, opened 2023-11-15
Fix arm64 ldr mask in objc_stubs analysis

thestr4ng3r:

**Detailed description**

This ldr is "LDR (immediate), Unsigned offset" as described in the ISA manual. The mask was covering data bits and thus could lead to false negative matches.

**Test plan**

Tests exist in `test/db/formats/mach0/objc` that cover this code already. They did not catch this bug because the data values in these files did not have any 1s in the respective bits. Adding more tests to cover every bit combination is impractical though.


wargio (approved):

librz/core/analysis_objc.c
```diff
@@ -575,6 +575,7 @@ static void analyze_objc_stubs(RzCore *core, ut64 start, ut64 size) {
 			break;
 		}
 		if (!stride) {
+			RZ_LOG_ERROR("Failed to match any known pattern against __objc_stubs contents\n");
```
wargio: maybe is worth adding to open an issue on github ?

thestr4ng3r: Added a note.

PR #3979, opened 2023-11-15
`rz_io_write_at`: Return silently if len <= 0.

Rot127:

**Detailed description**

Return silently if `len <= 0` and add docs for function.

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/librz/io/io.c
+++ b/librz/io/io.c
@@ -353,11 +353,24 @@ RZ_API int rz_io_nread_at(RzIO *io, ut64 addr, ut8 *buf, int len) {
 	return ret;
 }
 
+/**
+ * \brief Writes \p len bytes of data from \p buf to \p addr into the given \p io.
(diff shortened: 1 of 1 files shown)


ret2libc:
Why this? Where was the _assert_ being triggered?

XVilka:
Why would we ever want to write negative amount of bytes? Shouldn't it be solved at the root of the problem?

Rot127:
\*sign\* I was a little stupid. I thought for whatever reason that those `rz_return_val_if_fail` get removed by a release build. Ignore it please.

XVilka:
@Rot127 in fact, that `int len` could be converted into `ut64 len`, since negative length doesn't make sense, and we should be able to write chunks bigger than 4Gb size. Then your addition of the documentation could be a good PR too.

Issue #3996, opened 2023-11-21
Extended RzIL VM setup/IO cache handling


Rot127:
**Is your feature request related to a problem? Please describe.**

Opening a binary and emulating it with RzIL fails out of the box if the binary writes to memory.
Because writable memory regions are not mapped by default. So each memory write of the VM is ignored.

Hence the emulation is incorrect.

**Describe the solution you'd like**

We need a way to initialize the VM and with it the required memory regions (if any are known).

**Describe alternatives you've considered**

A current work-around is enabling io.cache with `e io.cache=1`. This makes all memory writable.
But this is obviously not nice, since we cannot detect access violations this way.

**Additional context**

Whatever the concrete implementation is, it should allow to:
- Initialize the VM without any additional configuration
  - Have commands to do this manually (like `aeim` for stack setup).
- Initialize the VM and setup as much as possible automatially (e.g. the memory regions + access rights)
  - If a single setup step fails, notify the user.

PR #4032, opened 2023-12-13
[RzIL] Warn the user about `io.cache`

Rot127:

**Detailed description**

It took me again way to long to resolve a "bug" which simply was a disabled `io.cache` setting.

I hope this warning helps at least somewhat until we address https://github.com/rizinorg/rizin/issues/3996

**Test plan**

All green

**Closing issues**

...

### DIFF
--- a/librz/analysis/il/analysis_il.c
+++ b/librz/analysis/il/analysis_il.c
@@ -91,12 +91,28 @@ static void setup_vm_from_config(RzAnalysis *analysis, RzAnalysisILVM *vm, RzAna
 static void setup_vm_init_state(RzAnalysisILVM *vm, RZ_NULLABLE RzAnalysisILInitState *is, RZ_NULLABLE RzReg *reg);
 
 /**
+ * \brief Checks if io.cache = true and warns the user if it is unset.
+ * Wthout it IL memory writes are ignored.
+ *
+ * See https://github.com/rizinorg/rizin/issues/3996 for a general
+ * problem description.
+ */
+static void check_io_cache_setting(const RzAnalysis *a) {
+	rz_return_if_fail(a);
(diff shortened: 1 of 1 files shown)


librz/analysis/il/analysis_il.c
```diff
@@ -90,13 +90,29 @@ RZ_API void rz_analysis_il_config_add_label(RZ_NONNULL RzAnalysisILConfig *cfg,
...
+}
+
 /**
  * Create and initialize an RzAnalysisILVM with the current arch/cpu/bits configuration and plugin
  * \p init_state_reg optional RzReg to take variable values from, unless the plugin overrides them using RzAnalysisILInitState
  * \return RzAnalysisRzil* a pointer to RzAnalysisILVM instance
  */
 RZ_API RZ_OWN RzAnalysisILVM *rz_analysis_il_vm_new(RzAnalysis *a, RZ_NULLABLE RzReg *init_state_reg) {
 	rz_return_val_if_fail(a, NULL);
+	check_io_cache_setting(a);
```
thestr4ng3r, line 115: Better put this only in the command that calls it, `rz_analysis_il_vm_new()` may be used outside an interactive context.

thestr4ng3r (changes requested):

thestr4ng3r:
My 2 cents, I don't think it is a good idea to suggest users unconditionally to use io.cache since the correct way is to set up appropriate writable maps. This warning would show even if the user has set up these maps by themselves already, in which case enabling io.cache would be wrong.

Rot127:
> This warning would show even if the user has set up these maps by themselves already,

Would it be fine for you if we add this to the note? Meaning that users can setup the maps themselves?

Rot127:
-.- Forgot of cause about all the test we would need to update. This doesn't makes sense of cause.

PR #4078, opened 2024-01-06
Adding missing operation in bitvector lib and core theory

Heersin:

**Detailed description**

1. Add some basic operations in util/bitvector

some of them are defined and implemented in rzil vm, but has no implementation in util/bitvector lib

```C
RZ_API RZ_OWN RzBitVector *rz_bv_pred(RZ_NONNULL RzBitVector *bv);
RZ_API RZ_OWN RzBitVector *rz_bv_succ(RZ_NONNULL RzBitVector *bv);
RZ_API bool rz_bv_arshift(RZ_NONNULL RzBitVector *bv, ut32 dist);
RZ_API RZ_OWN RzBitVector *rz_bv_cast(RZ_NONNULL RzBitVector *bv, ut32 to_size, bool fill_bit);
RZ_API RZ_OWN RzBitVector *rz_bv_signed_cast(RZ_NONNULL RzBitVector *bv, ut32 to_size);
RZ_API RZ_OWN RzBitVector *rz_bv_unsigned_cast(RZ_NONNULL RzBitVector *bv, ut32 to_size);

RZ_API bool rz_bv_slt(RZ_NONNULL RzBitVector *x, RZ_NONNULL RzBitVector *y);
RZ_API bool rz_bv_ult(RZ_NONNULL RzBitVector *x, RZ_NONNULL RzBitVector *y);
RZ_API bool rz_bv_sgt(RZ_NONNULL RzBitVector *x, RZ_NONNULL RzBitVector *y);
RZ_API bool rz_bv_ugt(RZ_NONNULL RzBitVector *x, RZ_NONNULL RzBitVector *y);
RZ_API bool rz_bv_sge(RZ_NONNULL RzBitVector *x, RZ_NONNULL RzBitVector *y);
RZ_API bool rz_bv_uge(RZ_NONNULL RzBitVector *x, RZ_NONNULL RzBitVector *y);
```

<del> 2. Add two new operation `pred` and `succ` to rzil bitvector theory </del>
After basic bitvector theory implemented in rizin, two operation were added to core theory basic, defined in [basic](http://binaryanalysisplatform.github.io/bap/api/master/bap-core-theory/Bap_core_theory/Theory/module-type-Basic/index.html#val-pred) 

similar to `inc` and `dec` operation in machine instruction
`pred x` get the predecessor of x in 2^n modulo
`succ x` get the successor of x in 2^n modulo

**Test plan**

unit test
`test_bitvector`

**Closing issues**

...

### DIFF
--- a/librz/util/bitvector.c
+++ b/librz/util/bitvector.c
@@ -1509,7 +1509,7 @@ RZ_API bool rz_bv_is_all_one(RZ_NONNULL const RzBitVector *x) {
 }
 
 /**
- * get predecessor of bv (dec 1)
+ * get predecessor of bv (dec 1) in 2^n modulo
  * \param bv
  * \return predecessor of bv
  */
@@ -1530,12 +1530,33 @@ RZ_API RZ_OWN RzBitVector *rz_bv_pred(RZ_NONNULL RzBitVector *bv) {
 }
 
 /**
+ * get successor of bv (inc 1) in 2^n modulo
+ * \param bv
+ * \return successor of bv
+ */
+RZ_API RZ_OWN RzBitVector *rz_bv_succ(RZ_NONNULL RzBitVector *bv) {
+	rz_return_val_if_fail(bv, NULL);
+	ut32 len = bv->len;
+	if (len <= 64) {
+		ut64 val = rz_bv_to_ut64(bv);
+		val += 1;
+		return rz_bv_new_from_ut64(len, val);
+	}
+
+	RzBitVector *one = rz_bv_new_one(len);
+	RzBitVector *result = rz_bv_sub(bv, one, NULL);
+	rz_bv_free(one);
+
+	return result;
+}
+
+/**
  * Arithmetic right shift of bv, shift right with (msb bv) bit filled
  * \param bv
  * \param dist shift distance
  * \return true if success
  */
-RZ_API bool rz_arshift(RZ_NONNULL RzBitVector *bv, ut32 dist) {
+RZ_API bool rz_bv_arshift(RZ_NONNULL RzBitVector *bv, ut32 dist) {
 	rz_return_val_if_fail(bv, false);
 	bool msb = rz_bv_msb(bv);
 	return rz_bv_rshift_fill(bv, dist, msb);
--- a/librz/util/bitvector.c
+++ b/librz/util/bitvector.c
@@ -1507,3 +1507,141 @@ RZ_API bool rz_bv_is_all_one(RZ_NONNULL const RzBitVector *x) {
 	}
 	return true;
 }
+
+/**
+ * get predecessor of bv (dec 1)
+ * \param bv
+ * \return predecessor of bv
+ */
+RZ_API RZ_OWN RzBitVector *rz_bv_pred(RZ_NONNULL RzBitVector *bv) {
+	rz_return_val_if_fail(bv, NULL);
+	ut32 len = bv->len;
+	if (len <= 64) {
+		ut64 val = rz_bv_to_ut64(bv);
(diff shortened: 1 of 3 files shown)


librz/util/bitvector.c
```diff
@@ -1507,3 +1507,162 @@ RZ_API bool rz_bv_is_all_one(RZ_NONNULL const RzBitVector *x) {
...
+ * get predecessor of bv (dec 1) in 2^n modulo
+ * \param bv
+ * \return predecessor of bv
+ */
+RZ_API RZ_OWN RzBitVector *rz_bv_pred(RZ_NONNULL RzBitVector *bv) {
+	rz_return_val_if_fail(bv, NULL);
+	ut32 len = bv->len;
+	if (len <= 64) {
+		ut64 val = rz_bv_to_ut64(bv);
+		val -= 1;
```
b1llow, line 1521: So what should pred 0 be?

DMaroo: I suppose it would be `0xfff...` (for `len` bits), given that we are doing `pred` modulo `2^n`.

DMaroo (approved):
Looks good to me! But I just have a question, why do we need these as ops? Is there a specific use-case? I don't see a reason why one cannot use the `ADD` and `SUB` ops by themselves to get these values. It'd be nicer to not have multiple redundant ways of doing the same thing.

b1llow:
I also have a feeling that what's more needed might be something like `(leading|tailing)_(one|zero|sign)s`? Though there are none of those in bap either.

librz/util/bitvector.c
Heersin: > So what should pred 0 be?

pred 0 gets 0xffff... according to bap description and its implementation

Heersin:
> I also have a feeling that what's more needed might be something like `(leading|tailing)_(one|zero|sign)s`? Though there are none of those in bap either.

yeah, but bap does not choose to provide them. Maybe it reserves `pred` and `succ` to generate (n + 1), (n - 1), and some sequence like (pred (pred ( ... 0)))... I guess.

DMaroo:
I feel that the `pred` and `succ` ops are there just due to their significance in functional programming per se. They are just residues of that paradigm, probably. I don't think they are any more useful than the ops we already have.

In any case, I wouldn't be against adding these two ops, but I just wanted to know the motivation behind them.

Heersin:
> I feel that the `pred` and `succ` ops are there just due to their significance in functional programming per se. They are just residues of that paradigm, probably. I don't think they are any more useful than the ops we already have.
> 
> In any case, I wouldn't be against adding these two ops, but I just wanted to know the motivation behind them.

yes, I also feel that they are here due to keep a consistence with functional programming. In fact, I just need the functions in util/bitvector to assist in implementing some floating point operations. The two omitted operations of rzil were just added by the way, just feel they maybe used in functional programming style without any strong motivation. As for the implementation motivation in bap, I will ask again.

XVilka:
I agree with @DMaroo and @imbillow, while these ops might be legit so far there are no instructions that require them and all logic could be expressed without them. Let's keep in mind that leaner VM is slightly faster, easier to test, and easier to keep "sound". I would say if you don't need it them for anything in particular - let's not add them.

XVilka (changes requested):
Please keep the bitvector changes but remove new ops, then we can merge.

librz/util/bitvector.c
wargio: add a test

wargio (changes requested):
add a test for pred & succ on limit cases.

XVilka (approved):

PR #4141, opened 2024-01-24
[RzIL] Unreachable effects

Rot127:

**Detailed description**

The assumptions made doesn't need to be true.
For Hexagon every write to a register first happens to a temporary
register file. Once the whole packet is checked and is valid,
the operations are performed.
The last step is to sync the the temporary register file with the default one.

So there is the possibility to have effects after a jump. Namely anything a processor would do to update its state after an instruction is executed.

**Test plan**

All green.

**Closing issues**

...

### DIFF
--- a/librz/il/il_validate.c
+++ b/librz/il/il_validate.c
@@ -980,11 +980,6 @@ VALIDATOR_EFFECT(seq) {
 	VALIDATOR_DESCEND_EFFECT(args->x, &tx, ctx, {});
 	RzILTypeEffect ty;
 	VALIDATOR_DESCEND_EFFECT(args->y, &ty, ctx, {});
-	// Code after a jmp/goto makes no sense because the jmp naturally jumps somewhere else already.
-	// Intuitively, this could be considered just dead code and valid, but because it is not practically useful,
-	// we reject such code completely for now, which gives us more freedom if in the future we do want to define
-	// semantics for code after ctrl in some way.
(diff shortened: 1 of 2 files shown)


wargio (commented):
@thestr4ng3r you may want to look at this one.

Rot127:
Answer by @thestr4ng3r in mattermost:

>  the "no effects after jump" rule is a consequence of how bap core theory is designed. There you have a distinction between data effects and control effects which are different types in the ocaml code so it is impossible to mix them at all.
> Then there is the "block" effect val blk : label -> data eff -> ctrl eff -> unit eff, which combines effects of both types with the semantics that first the data effect is performed and then the ctrl effect, passing control flow to another block to be executed.
> The result of this is that once you lifted all ISA instrutions, where each results in a single block, these blocks will be linked together via their jumps to each others' labels. You then have an entire program purely consisting of IL basic blocks and you can apply formal analyses and forget about any "informal" ISA details, like instruction addresses, program counters, etc. which is a huge step towards more formal and provable analysis.
> This is why I enforced this rule. In other words, the semantics of a jump in RzIL are not "update the ISA pc and continue executing more IL as if nothing happened" but really a jump to other IL code, so in that sense this unreachable code really would never be executed anyway and emitting such code is almost certainly a bug in the lifter.
> So the way to approach this would be to adjust your lifter to put the jump after all data effects, even if technically in the ISA the program counter is updated before.

PR #4167, opened 2024-01-29
Add support to Windows NUMA core count

wargio:

**Detailed description**

This is not testable normally but it "should work".

### DIFF
--- a/librz/util/thread_pool.c
+++ b/librz/util/thread_pool.c
@@ -14,6 +14,66 @@ struct rz_th_pool_t {
 	RzThread **threads;
 };
 
+#ifdef __WINDOWS__
+/// defines the GetLogicalProcessorInformationEx function
+typedef BOOL (*WINAPI glpie_t)(
+	LOGICAL_PROCESSOR_RELATIONSHIP,
+	PSYSTEM_LOGICAL_PROCESSOR_INFORMATION_EX,
+	PDWORD);
+
+// Classic way to get number of cores in windows.
+static size_t get_windows_compatible_n_cores() {
+	SYSTEM_INFO sysinfo;
+	GetSystemInfo(&sysinfo);
+	return sysinfo.dwNumberOfProcessors;
+}
+
+/*
+ * Windows NUMA support is particular
+ * https://learn.microsoft.com/en-us/windows/win32/api/sysinfoapi/nf-sysinfoapi-getlogicalprocessorinformationex#remarks
+ *
+ * New api here:
+ * https://learn.microsoft.com/en-us/windows/win32/procthread/numa-support
+ */
+static size_t get_windows_numa_n_cores() {
+	ut8 *buffer = NULL;
+	size_t n_cores = 0;
+	DWORD length = 0;
+	HMODULE kernel32 = GetModuleHandleW(L"kernel32.dll");
+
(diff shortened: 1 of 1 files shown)


karliss:
As was already pointed out to GermanAizek multiple times in other issues/PR that he spammed . The NUMA behavior on windows is a bit more complicated than simply bigger number goes faster.  Even if you ignore the implications of memory access from multiple sockets, there is windows scheduler which isn't necesarily going to spread the threads across all CPUs (with behavior differing depending on Windows version). Actually utilizing multiple CPUs requires more work than returning bigger number in `get_thread_count()`. 

The smaller number that you get by default from windows is intentional.

I don't have the required hardware to confirm how exactly each windows version behaves.

See 
https://github.com/llvm/llvm-project/pull/72270#pullrequestreview-1734790096
https://developercommunity.visualstudio.com/t/hardware-concurrency-returns-an-incorrect-result/350854#T-N354351
https://github.com/Xaymar/obs-StreamFX/issues/1093

wargio:
that is a good point, now the issue is that even if we wanted to implement that, we would need to have a NUMA system (which we don't).

GermanAizek:
> that is a good point, now the issue is that even if we wanted to implement that, we would need to have a NUMA system (which we don't).

I can test branches, but I need time to build as in CI under Windows, it is easy to building on Linux. Why is there no cmake build, because it is so convenient to change C CXX compiler for building via `-DCXX_COMPILER`, but how do I change MinGW to MSVC using meson build?

wargio:
just run the build from the cmd prompt or powershell.

wargio:
https://github.com/rizinorg/cutter/issues/3286#issuecomment-1973444166

As per cutter thread, this works but numa support requires some better implementation of the multhreading code, which we cannot right now implement.

GermanAizek:
> [rizinorg/cutter#3286 (comment)](https://github.com/rizinorg/cutter/issues/3286#issuecomment-1973444166)
> 
> As per cutter thread, this works but numa support requires some better implementation of the multhreading code, which we cannot right now implement.

@wargio, you can take this, it really quickly check NUMA support. I can helped with NUMA portability.

```cpp
bool IsNUMA() noexcept
{
    ULONG HighestNodeNumber;
    return !(!GetNumaHighestNodeNumber(&HighestNodeNumber) || HighestNodeNumber == 0);
}
```

wargio:
the issue mentioned above is that in order to use all the numa capabilities, we need a more numa-friendly code, which requires specific malloc etc.. on windows. we don't have that yet. potentially we could implement a piece of code that detects if numa and allows to run with more cores, but i'm not sure.
@XVilka what do you think?

XVilka:
I don't think it worth the trouble at this point, especially since currently file parsing and analysis aren't multithreaded, thus benefits of this will be quite limited.

GermanAizek:
> we need a more numa-friendly code, which requires specific malloc etc.. on windows.

@wargio,
Are you talking about WinAPI allocation?
Their implementation has malloc functions for working with NUMA, previously I tried to send PR to x64dbg
https://github.com/x64dbg/x64dbg/pull/3307

wargio:
yes, but the issue is related on how rizin works. we do not have a centralized `malloc` function.

notxvilka:
I am closing this, because 1) no updates 2) the solution should be different

Issue #904, opened 2021-03-25
Nested `~{0}~{}` doesn't work


XVilka:
### Work environment

| Questions                                            | Answers
|------------------------------------------------------|--------------------
| OS/arch/bits (mandatory)                             | -
| File format of the file you reverse (mandatory)      | -
| Architecture/bits of the file (mandatory)            | -
| `rizin -v` full output, **not truncated** (mandatory)  | https://github.com/rizinorg/rizin/commit/bed31a6173a7a3632395498e0c7386ca72505ba0

### Expected behavior

Nested `~{}~{}` should still format JSON and provide syntax highlight similar to how it's done in the unnested case.

```
[0x00004020]> iij~{0} # extract 0th JSON array element
{"ordinal":1,"bind":"GLOBAL","type":"FUNC","name":"__ctype_toupper_loc","plt":0}
[0x00004020]> iij~{2} # extract 2nd JSON array element
{"ordinal":3,"bind":"GLOBAL","type":"FUNC","name":"free","plt":0}
[0x00004020]> iij~{0}~{}
{
  "ordinal":1,
  "bind":"GLOBAL",
  "type":"FUNC",
  "name":"__ctype_toupper_loc",
  "plt":0
 }
[0x00004020]> iij~{2}~{}
{
  "ordinal":3,
  "bind":"GLOBAL",
  "type":"FUNC",
  "name":"free",
   "plt":0
}
[0x00004020]> 
```

### Actual behavior

```
[0x00004020]> iij~{0}
{"ordinal":1,"bind":"GLOBAL","type":"FUNC","name":"__ctype_toupper_loc","plt":0}
[0x00004020]> iij~{2}
{"ordinal":3,"bind":"GLOBAL","type":"FUNC","name":"free","plt":0}
[0x00004020]> iij~{0}~{}
{"ordinal":1,"bind":"GLOBAL","type":"FUNC","name":"__ctype_toupper_loc","plt":0}
[0x00004020]> iij~{2}~{}
{"ordinal":3,"bind":"GLOBAL","type":"FUNC","name":"free","plt":0}
[0x00004020]> 
```

ret2libc:
I agree. I think grep syntax should be completely moved and handled by the parser. Right now almost everything after `~` is passed to rz_cons_grep, which does not support many of the nested things (even multi-word grep are supported just through a hack that rewrites them into a single grep with `,`).

XVilka:
[...]

PR #4200, opened 2024-02-09
Rz-find -E "filename" to execute command on match

### LINKED ISSUE #4192: Add `rz-find -E` (Execute) option to execute command on match
One of the most common uses of UNIX `find` command is the use together with the `-exec` option, that allows to run some command given the output of the match. In case of the `rz-find` it would be search among directories, and when this option provided, it should run the command on every filename/path matched.

```
find [path] [arguments] -exec [command] {} \;
```
For example:
```sh
$ find . -name "*.mp3" -exec echo {} \;
```

See
- `binrz/rz-find/rz-find.c`
- `librz/main/rz-find.c`

HN026:

**Detailed description**

...

**Test plan**

...

**Closing issues**

closes #4192 

**Book Changes**
https://github.com/rizinorg/book/pull/116


XVilka (changes requested):
Add a simple test in `test/db/tools/rz_find`. Using `echo <filename>` as a command, it should be supported on all platforms.

wargio (changes requested):
please update the man page (under docs), the help message and also update the rizin book and open a PR.

librz/main/rz-find.c
```diff
@@ -633,6 +633,15 @@ RZ_API int rz_main_rz_find(int argc, const char **argv) {
 			return 1;
 		}
 		rzfind_open(&ro, file);
+
+		if (ro.exec_command) {
+			char *command = rz_str_newf("%s %s", ro.exec_command, file);
+			int status = system(command);
```
wargio: should we use execv* ? anyway use `rz_sys_system` and not `system`

HN026:
> please update the man page (under docs), the help message and also update the rizin book and open a PR.

Sure

librz/main/rz-find.c
HN026: I thought of that but went with using `system`, will make the changes soon.

librz/main/rz-find.c
```diff
@@ -181,6 +182,7 @@ static int show_help(const char *argv0, int line) {
 		"-a",    "[align]", "Only accept aligned hits",
 		"-b",    "[size]",  "Set block size",
 		"-e",    "[regex]", "Search for regex matches (can be used multiple times)",
+        "-E",    "[cmd]",   "Execute command for each file found",
```
wargio: this does not look correctly formatted. run clang-format.

wargio (commented):
LGTM.

test/db/tools/rz_find
```diff
@@ -260,3 +260,17 @@ EXPECT=<<EOF
 EOF
 RUN
 
+NAME=rz-find -E with echo as a command
+FILE==
+CMDS=!rz-find -E "echo" bins/arm/README
```
XVilka: Change the test so that it woudl search for a string `README`, like in previous tests, like the recursive test, on a directory. It should also only print the filename with the `echo`, I think.

XVilka (changes requested):

test/db/tools/rz_find
HN026: Suree

XVilka (approved):

wargio (approved):

XVilka (changes requested):
It doesn't work as expected. For some reason it finds "README" string even in files without it:
```
+File: bins/arm/README
 0x0
+File: bins/arm/arm64_rzil_test.rz
+File: bins/arm/crackme.arm32.bin
+File: bins/arm/elf/hello_world-linaro-201201
+File: bins/arm/elf/hello_world
+File: bins/arm/elf/hello_world-linaro-201902
+File: bins/arm/elf/hello_world-buildroot-201402
+File: bins/arm/src/hello_world.c
```

librz/main/rz-find.c
```diff
@@ -633,6 +634,15 @@ RZ_API int rz_main_rz_find(int argc, const char **argv) {
 			return 1;
 		}
 		rzfind_open(&ro, file);
+
+		if (ro.exec_command) {
```
wargio: this should be probably executed within `rzfind_open`

HN026: Okayy

librz/main/rz-find.c
```diff
@@ -498,6 +498,14 @@ static int rzfind_open(RzfindOptions *ro, const char *file) {
...
 		return res;
 	}
+	if (ro->exec_command) {
+		char *command = rz_str_newf("%s %s", ro->exec_command, file);
+		int status = rz_sys_system(command);
+		if (status == -1) {
+			RZ_LOG_ERROR("Failed to execute command: %s", command);
+		}
+		free(command);
+	}
```
wargio: no. the function that actually does the search is `rzfind_open_file`. please try to understand what the request is and what the function does. the command needs to be executed only when there is a hit with the search. this means that each time the program finds something that matches what it is looking for it should either print (current behaviour) or execute a program (new feature you are implementing).

Also you have removed the previous command `-E` which was used for ESIL search (even tho there was no help message info on the util) and didn't remove all the old usages.

HN026: It's clear now, will fix it. thanks

XVilka (changes requested):
Much better now! 
The only remaining problem (not in your code), is that Windows shows paths differently from *nix systems, thus causing this test to fail on AppVeyor:
```
[XX] C:\projects\rizin\test\db\tools\rz_find rz-find -E with echo as a command
ANSICON=1 RZ_NOPLUGINS=1 C:\projects\rizin\rizin-vs2022_64-v0.8.0\bin\rizin.exe -escr.utf8=0 -escr.color=0 -escr.interactive=0 -eflirt.sigdb.load.system=false -eflirt.sigdb.load.home=false -N -Qc '!rz-find -s "README" -E "echo" bins/arm' =
-- stdout
--- expected
+++ actual
@@ -1,2 +1,2 @@
-bins/arm/README
+bins/arm\README
 0x0
-- stderr
WARNING: Neither hash nor gnu_hash exist. Falling back to heuristics for deducing the number of dynamic symbols...
WARNING: Neither hash nor gnu_hash exist. Falling back to heuristics for deducing the number of dynamic symbols...
WARNING: Neither hash nor gnu_hash exist. Falling back to heuristics for deducing the number of dynamic symbols...
ERROR: Cannot determine entrypoint, using 0x000003fc.
ERROR: Cannot determine entrypoint, using 0x000082fc.
ERROR: Cannot determine entrypoint, using 0x00010320.
```

librz/main/rz-find.c
```diff
@@ -541,8 +554,13 @@ RZ_API int rz_main_rz_find(int argc, const char **argv) {
...
 		case 'E':
-			ro.mode = RZ_SEARCH_ESIL;
-			rz_list_append(ro.keywords, (void *)opt.arg);
+			ro.quiet = true;
+			if (opt.arg) {
+				ro.exec_command = opt.arg;
+				rz_list_append(ro.keywords, (void *)opt.arg);
+			} else {
+				ro.mode = RZ_SEARCH_ESIL;
+			}
```
wargio: no. this command cannot be both.

wargio: as i told you previously remove the esil search from here.

librz/main/rz-find.c
```diff
@@ -51,13 +53,15 @@ static void rzfind_options_init(RzfindOptions *ro) {
...
 	ro->keywords = rz_list_newf(NULL);
+	ro->exec_command = NULL;
 }
 
 static int rzfind_open(RzfindOptions *ro, const char *file);
 
 static int hit(RzSearchKeyword *kw, void *user, ut64 addr) {
 	RzfindOptions *ro = (RzfindOptions *)user;
 	int delta = addr - ro->cur;
+	ro->match_found = true;
```
wargio: just execute the command here.

wargio:
```c
if (exec_command) {
		char *command = rz_str_newf("%s %s", ro->exec_command, ro->curfile);
		int status = rz_sys_system(command);
		if (status == -1) {
			RZ_LOG_ERROR("Failed to execute command: %s", command);
		}
		free(command);
		return 1;
}
```
and then return

librz/main/rz-find.c
HN026: My bad, I misunderstood it since I had removed the ESIL search earlier, I thought you want me to have both. Will Fix it.

wargio:
> Much better now! The only remaining problem (not in your code), is that Windows shows paths differently from *nix systems, thus causing this test to fail on AppVeyor:
> 
\> ```
\> -bins/arm/README
> +bins/arm\README
\> ```

it's caused by `rz_file_path_join` but the question we should have is, how do we check for the actual filename since the paths are definitely going to be always bugged? maybe we can grep and count ?

librz/main/rz-find.c
```diff
@@ -143,7 +143,7 @@ static int hit(RzSearchKeyword *kw, void *user, ut64 addr) {
 		}
 	}
 	if (ro->exec_command) {
-		char *command = rz_str_newf("%s %s", ro->exec_command, ro->curfile);
+		char *command = rz_str_newf("%s ; grep -c . %s", ro->exec_command, ro->curfile);
```
wargio: no.

wargio: you could change the test file like this `CMDS=!rz-find -s "README" -E "echo" bins/arm~README~?`

test/db/tools/rz_find
```diff
@@ -260,3 +260,10 @@ EXPECT=<<EOF
 EOF
 RUN
 
+NAME=rz-find -E with echo as a command
+FILE==
+CMDS=!rz-find -s "README" -E "echo" bins/arm~README~?
+EXPECT=<<EOF
+0
```
XVilka: That test is bad, to be honest. Why zero? It should have one match

XVilka (changes requested):

test/db/tools/rz_find
HN026: It's actually Incorrect. The command `!rz-find -s "README" -E "echo" bins/arm` correctly finds the addresses of 'README'. However, when I filter these results with `~README~` and count lines with `?`, it returns `0`. Despite 'README' being found, the filter seems to fail.

XVilka: I think it's because we don't have support for nested greps yet: https://github.com/rizinorg/rizin/issues/904
Please try to make a test with just one internal grep

HN026: Okay Sure.

test/db/tools/rz_find
```diff
@@ -262,10 +262,8 @@ RUN
 
 NAME=rz-find -E with echo as a command
 FILE==
-CMDS=!rz-find -s "README" -E "echo" bins/arm
+CMDS=!rz-find -s "README" -E "echo" bins/arm~README~?
 EXPECT=<<EOF
-
-5
-0x0
+0
```
wargio: this should return 1, not 0

HN026: I fixed this, it was because rizin doesn't have support for nested Greps so I fixed it to one internal grep but now issue is something else, it works as expected on Linux but it's returning `2` on windows.

test/db/tools/rz_find
HN026: @wargio you can see this

librz/main/rz-find.c
```diff
@@ -140,6 +142,15 @@ static int hit(RzSearchKeyword *kw, void *user, ut64 addr) {
 			}
 		}
 	}
+	if (ro->exec_command) {
+		char *command = rz_str_newf("%s", ro->exec_command);
+		int status = rz_sys_system(command);
+		if (status == -1) {
+			RZ_LOG_ERROR("Failed to execute command: %s", command);
```
wargio: shouldn't this be executed with the found file?

wargio: i think that is why you had the issue before where the double `grep` didn't work.

HN026: It is but ig we are filtering the output w `~?` to return the number of times the string has occurred because without `~?`, I do get all instances of the string I'm searching.

wargio: yes, but the command should be executed with the file name, that is why you got `0` hit with the double grep.

HN026:
```sh
[XX] C:\projects\rizin\test\db\tools\rz_find rz-find -E with echo as a command
ANSICON=1 RZ_NOPLUGINS=1 C:\projects\rizin\rizin-vs2022_64-v0.8.0\bin\rizin.exe -escr.utf8=0 -escr.color=0 -escr.interactive=0 -eflirt.sigdb.load.system=false -eflirt.sigdb.load.home=false -N -Qc '!rz-find -s "README" -E "echo" bins/arm~?' =
-- stdout
--- expected
+++ actual
@@ -1,1 +1,1 @@
-1
+2
-- stderr
```
I've for some reason, this returning me `2` on windows instead of `1`.

wargio: https://github.com/rizinorg/rizin/pull/4200/commits/4eb0d284adde63abcd8e0c39522d6585df5610f4
This failed to return `1` because you removed the file name.

wargio: The code should be
```c
char *command = rz_str_newf("%s %s", ro->exec_command, ro->curfile);
```

wargio: Then this should work `CMDS=!rz-find -s "README" -E "echo" bins/arm~README~?` and return `1`

HN026: Okayyyy got it. Thanks

wargio:
@XVilka looks like my double grep works. :D

librz/main/rz-find.c
```diff
@@ -140,6 +142,15 @@ static int hit(RzSearchKeyword *kw, void *user, ut64 addr) {
 			}
 		}
 	}
+	if (ro->exec_command) {
+		char *command = rz_str_newf("%s %s", ro->exec_command, ro->curfile);
```
XVilka, line 146: Filename should be put in quotes here to avoid (likely unintended) command injection and properly handle filenames with spaces.

XVilka (changes requested):
LGTM apart from one thing

librz/main/rz-find.c
HN026: Okay Sure

wargio: > Filename should be put in quotes here to avoid (likely unintended) command injection and properly handle filenames with spaces.

actually the only way to patch this is to use the `rz_sys_execve` but i would prefer to merge this as is since command injection is not exploitable from the prospective of the os. you will always have to provide the filename before hand so it's kinda impossible to exploit.

XVilka (approved):

PR #4202, opened 2024-02-10
Refactor how ASLR and implement rz_sys_execv using RzSubprocess

### LINKED ISSUE #4147: ASLR handling on windows is not supported in rz_sys_aslr
### Work environment

| OS/arch/bits (mandatory)                             | windows 10
| File format of the file you reverse (mandatory)      | PE,
| Architecture/bits of the file (mandatory)            | x64
| `rizin -v` full output, **not truncated** (mandatory)  |rizin 0.7.0 @ windows-x86-64 commit: b2112ddcdf6be4fc292a699d0f77ed3add50d682

### Expected behavior

When I rerun the program, I would like to have the same base address for any function.

### Actual behavior

The base address of roblox change sometimes.

I did:

```
.\OUT\bin\rizin.exe -R "aslr=no" -e "pdb.autoload=true" -e scr.color=2 -c "dc" -e "search.in=analysis.fcn" -d C:\Users\yoshi\AppData\Local\Roblox\Versions\version-88cfc23f4e7d4e4b\RobloxPlayerBeta.exe
```

wargio:

**Detailed description**

Refactors the code and moves the code into the same function.

Adds support to windows via `SetProcessMitigationPolicy` but works only when [ASLR is disable and we want to enable it](https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-setprocessmitigationpolicy#remarks).

Also now on linux we try to disable ASLR programmatically using the `personality` function instead of OS wide which works also when the user is not root.

These changes also implements non-native `rz_sys_execv` by using `RzSubprocess`.

**Test plan**

Unfortunately this cannot be tested easily in the regression suite.

- Build the following C code by `gcc example.c`
```c
#include <stdio.h>

const char *test = "test";

int main() {
  printf("%p\n", (void*)test);
  return 0;
}
```

- Then create the following rz-run file `noaslr.rz`:

```bash
aslr=no
program=a.out
```

Then run `rz-run noaslr.rz` multiple times.

If the value output is the same then ASLR is correctly disabled.

**Closing issues**

Fix #4147

### DIFF
--- a/librz/util/sys.c
+++ b/librz/util/sys.c
@@ -1710,7 +1710,7 @@ RZ_API int rz_sys_execv(RZ_NONNULL const char *pathname, RZ_NONNULL char *const 
  * \return Returns -1 on failure otherwise the exit(n) value returned by the executed program.
  */
 RZ_API int rz_sys_execv(RZ_NONNULL const char *pathname, RZ_NONNULL char *const argv[]) {
-	rz_return_val_if_fail(!pathname || !argv || RZ_STR_ISEMPTY(argv[0]), -1);
+	rz_return_val_if_fail(RZ_STR_ISNOTEMPTY(pathname) && argv && RZ_STR_ISNOTEMPTY(argv[0]), -1);
 
 	if (!rz_subprocess_init()) {
 		RZ_LOG_ERROR("Cannot initialize subprocess in execv.\n");
--- a/librz/util/sys.c
+++ b/librz/util/sys.c
@@ -609,53 +609,53 @@ RZ_API bool rz_sys_chdir(RZ_NONNULL const char *s) {
 
 /**
  * \brief Enable or disable ASLR for the calling process
- * 
+ *
  * Host is Linux:
- * 
+ *
  * We can disable ASLR using the personlity flags (does not require root) or
  * system-wide by setting /proc/sys/kernel/randomize_va_space to 0.
  * If we want to enable it we must write 1 or 2 into /proc/sys/kernel/randomize_va_space
  * ASLR support was added in linux 2.6 (year 2005)
- * 
+ *
  * Host is NetBSD:
- * 
+ *
  * We can enable/disable ASLR by calling sysctlbyname("security.pax.aslr.enabled").
  * May require root/super-user priviledges.
- * 
+ *
  * Host is DragonFly:
- * 
- * We can enable/disable ASLR by calling sysctlbyname("vm.randomize_mmap") 
+ *
+ * We can enable/disable ASLR by calling sysctlbyname("vm.randomize_mmap")
  * May require root/super-user priviledges.
- * 
+ *
  * Host is FreeBSD (since version 13.0):
- * 
+ *
  * We can enable/disable ASLR by calling sysctlbyname("kern.elf32.aslr.enable") and
  * sysctlbyname("kern.elf64.aslr.enable")
  * May require root/super-user priviledges.
- * 
+ *
  * Host is Windows:
- * 
+ *
  * We can only enable ASLR and not disable it.
  * To enable it we just need to initialize PROCESS_MITIGATION_ASLR_POLICY by setting to true
  * EnableBottomUpRandomization, EnableForceRelocateImages and EnableHighEntropy, and then call
  * SetProcessMitigationPolicy(ProcessASLRPolicy).
  * https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-setprocessmitigationpolicy
- * 
+ *
  * This feature is available since _WIN32_WINNT >= 0x0602 (Windows 8, Windows Server 2012)
  * It is possible to manually disable ASLR on a binary by removing the PE flags called DLL Characteristics:
  * - IMAGE_DLLCHARACTERISTICS_DYNAMIC_BASE
  * - IMAGE_DLLCHARACTERISTICS_HIGH_ENTROPY_VA
  * or as admin from PowerShell by running `Set-ProcessMitigation -Name file.exe -Disable ForceRelocateImages`
- * 
+ *
  * Host is Apple:
- * 
+ *
  * We can enable/disable ASRL when we spawn a process by adding the flag called _POSIX_SPAWN_DISABLE_ASLR (0x0100) to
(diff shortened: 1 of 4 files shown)


wargio:
@gogo2464 if you want to try this on windows, it could be useful.

wargio:
i'm currently testing this on windows.

wargio:
So i have tested this extensively on windows and you cannot disable ASLR programmatically, only enable it.

XVilka:
> So i have tested this extensively on windows and you cannot disable ASLR programmatically, only enable it.

This makes perfect sense from the security PoV.

librz/util/sys.c
```diff
@@ -556,48 +610,154 @@ RZ_API bool rz_sys_chdir(RZ_NONNULL const char *s) {
 /**
```
XVilka, line 610: Add a note about Windows limitation here

librz/util/sys.c
```diff
@@ -556,48 +610,154 @@ RZ_API bool rz_sys_chdir(RZ_NONNULL const char *s) {
...
+	}
+	if (!setProcessMitigationPolicy(ProcessASLRPolicy, &policy, sizeof(policy))) {
+		DWORD dw = GetLastError();
+		char *err = sys_windows_error_msg(dw);
+		RZ_LOG_ERROR("Cannot set ProcessASLRPolicy: %s (0x%x)\n", err, dw);
+		free(err);
+		return false;
+	}
+#elif __APPLE__
+	// Apple supports this by using the spawn api.
```
XVilka: Most other multiline comments are `/**/`. Lets be consistent.

librz/util/sys.c
```diff
@@ -1501,7 +1643,45 @@ RZ_API int rz_sys_execv(const char *pathname, char *const argv[]) {
 }
 #elif !HAVE_EXECV
 RZ_API int rz_sys_execv(const char *pathname, char *const argv[]) {
```
XVilka: Could you please add a Doxygen to this function while you are at it?

XVilka (changes requested):
These files are an absolute mess; thanks for making them a bit more readable. As for other platforms, maybe @arrowd and @devnexen could help us?

XVilka:
It breaks one unit test on Windows:
```
 66/119 rizin:unit / run                                   FAIL             0.02s   (exit status 4294967295 or signal 4294967167 SIGinvalid)                 
```

devnexen:
> These files are an absolute mess; thanks for making them a bit more readable. As for other platforms, maybe @arrowd and @devnexen could help us?

it builds and has the same succesful unit tests than `dev` on haiku.

XVilka:
Static build fails subprocess unit test:
```
Summary of Failures:

 90/119 rizin:unit / subprocess                            FAIL            0.01s   exit status 1
```

XVilka (approved):

gogo2464:
> So i have tested this extensively on windows and you cannot disable ASLR programmatically, only enable it.

does `-R aslr=no` works now?

arrowd:
Is `rz_sys_aslr()` intended to change a global system-wide ASLR setting?

gogo2464:
> Is `rz_sys_aslr()` intended to change a global system-wide ASLR setting?

only for the current process, check: https://github.com/rizinorg/rizin/blob/d86e8aebade8bdc232d09d1699ad5969e794f9c1/librz/util/sys.c#L762

wargio:
> > So i have tested this extensively on windows and you cannot disable ASLR programmatically, only enable it.
> 
> does `-R aslr=no` works now?

on windows no and never will.

```c
		RZ_LOG_ERROR("On Windows, ASLR mitigation policies cannot be made less restrictive if they are already enabled.\n");
		RZ_LOG_ERROR("It is possible to disable ASLR by modifying the PE header and removing the following flags:\n");
		RZ_LOG_ERROR("- IMAGE_DLLCHARACTERISTICS_DYNAMIC_BASE\n");
		RZ_LOG_ERROR("- IMAGE_DLLCHARACTERISTICS_HIGH_ENTROPY_VA.\n");
		RZ_LOG_ERROR("Or as Administrator from PowerShell by running: `Set-ProcessMitigation -Name file.exe -Disable ForceRelocateImages`\n");
```

wargio:
> Is `rz_sys_aslr()` intended to change a global system-wide ASLR setting?

it is supposed to be used for spawning processes with aslr or not.
that function will change in the future and will be only used in the debugger.

arrowd:
Then the FreeBSD case is wrong. Changing sysctl is global and affects all processes that are started after the switch.

On FreeBSD it is possible to do `elfctl -e +noaslr file` to force an executable file to be started without ASLR. Pretty much, just like the Windows case.

gogo2464:
> > > So i have tested this extensively on windows and you cannot disable ASLR programmatically, only enable it.
> > 
> > 
> > does `-R aslr=no` works now?
> 
> on windows no and never will.
> 
> ```c
> 		RZ_LOG_ERROR("On Windows, ASLR mitigation policies cannot be made less restrictive if they are already enabled.\n");
> 		RZ_LOG_ERROR("It is possible to disable ASLR by modifying the PE header and removing the following flags:\n");
> 		RZ_LOG_ERROR("- IMAGE_DLLCHARACTERISTICS_DYNAMIC_BASE\n");
> 		RZ_LOG_ERROR("- IMAGE_DLLCHARACTERISTICS_HIGH_ENTROPY_VA.\n");
> 		RZ_LOG_ERROR("Or as Administrator from PowerShell by running: `Set-ProcessMitigation -Name file.exe -Disable ForceRelocateImages`\n");
> ```

might we consider patching the PEB for disabling aslr?

arrowd:
> might we consider patching the PEB for disabling aslr?

I'd say it makes sense, because it follows all other plaforms except Linux and MacOS for which it is possible to set ASLR state on the program startup.

devnexen:
> Then the FreeBSD case is wrong. Changing sysctl is global and affects all processes that are started after the switch.
> 
> On FreeBSD it is possible to do `elfctl -e +noaslr file` to force an executable file to be started without ASLR. Pretty much, just like the Windows case.

or you can use [the procctl api with PROC_ASLR_CTL](https://man.freebsd.org/cgi/man.cgi?procctl(2)).

PR #4207, opened 2024-02-12
Use the no-pie version (no ASLR) for the test

wargio:
This may fix the annoying mach0 debug test from failing

### DIFF
--- a/test/db/archos/darwin-arm64/dbg
+++ b/test/db/archos/darwin-arm64/dbg
@@ -85,7 +85,7 @@ EOF
 RUN
 
 NAME=maps as flags
-FILE=bins/mach0/hello-macos-arm64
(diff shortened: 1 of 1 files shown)


XVilka (approved):

wargio:
@XVilka can you rebuild the hello example for arm and remove the pie flag while building it?

XVilka:
> @XVilka can you rebuild the hello example for arm and remove the pie flag while building it?

What file exactly do you mean?

wargio:
that file was built with the -pie flag on macos. we need this to be without the ASLR support (https://stackoverflow.com/questions/23897963/documented-way-to-disable-aslr-on-os-x)

wargio:
The binary i have added is broken. i will revert changes on the test repo

PR #4223, opened 2024-02-13
Rewrite how rz_cons_pipe_open/close behave to remove globals

wargio:
**Detailed description**

Removed the old code because it was just wrong and replaced with a proper implementation.

This removes also some globals.

Tested on linux and windows.


XVilka (approved):
LGTM if green

PR #4227, opened 2024-02-14
Trim grep expression before processing it.

Rot127:

**Detailed description**

Fixes issue that `<command>~expr > /tmp/somewhere` is differently parsed than `<command>~expr> /tmp/somewhere`

**Test plan**

All green

**Closing issues**

...

### DIFF
--- a/librz/cons/grep.c
+++ b/librz/cons/grep.c
@@ -433,11 +433,13 @@ RZ_API char *rz_cons_grep_strip(char *cmd, const char *quotestr) {
 	return ptr;
 }
 
-RZ_API void rz_cons_grep_process(char *grep) {
-	if (grep) {
-		parse_grep_expression(grep);
-		free(grep);
+RZ_API void rz_cons_grep_process(RZ_OWN char *grep) {
+	if (!grep) {
+		return;
 	}
+	rz_str_trim_tail(grep);
+	parse_grep_expression(grep);
+	free(grep);
 }
 
 static int cmp(const void *a, const void *b, void *user) {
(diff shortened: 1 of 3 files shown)


Rot127:
Maybe we should abandon this and just wait until we refactor grep to use the PCRE2? Because currently it doesn't

librz/cons/grep.c
```diff
@@ -433,11 +433,13 @@ RZ_API char *rz_cons_grep_strip(char *cmd, const char *quotestr) {
...
 
-RZ_API void rz_cons_grep_process(char *grep) {
-	if (grep) {
-		parse_grep_expression(grep);
-		free(grep);
+RZ_API void rz_cons_grep_process(RZ_OWN char *grep) {
+	if (!grep) {
+		return;
 	}
+	rz_str_trim(grep);
```
wargio: i think this should be just a right trim.

ret2libc (approved):

XVilka (approved):

librz/cons/grep.c
```diff
@@ -433,11 +433,13 @@ RZ_API char *rz_cons_grep_strip(char *cmd, const char *quotestr) {
...
 
-RZ_API void rz_cons_grep_process(char *grep) {
-	if (grep) {
-		parse_grep_expression(grep);
-		free(grep);
+RZ_API void rz_cons_grep_process(RZ_OWN char *grep) {
+	if (!grep) {
+		return;
 	}
+	rz_str_trim_head(grep);
```
wargio: this should be `rz_str_trim_tail`

Rot127: True, agitated day.

ret2libc:
Damn I also had the same feeling as @wargio but then I looked at the docstring and said:
```
/**
 * \brief Removes whitespace characters (space, tab, newline etc.)
 * from the end of a string.
 * The string is changed in place.
 *
 * \param str The string to trim.
 */
RZ_API void rz_str_trim_head(RZ_NONNULL char *str) {
```

Could you fix that. Of course trim_head is removing spaces from the beginning of the string, not the end.

PR #4251, opened 2024-02-17
Fix initialization and conversion for 80-bit floats

DMaroo:

**Detailed description**

Now we can _correctly_ initialize an 80-bit `RzFloat` and also convert floats from and to 80-bit floats.

It is kind of unfortunate to deal with the idiosyncrasy of 80-bit float this way (through if conditions checking for 80-bit float). There must be cleaner ways, but they would require more refactor, and are probably not worth it since we wouldn't be changing the float code that often.

**Test plan**

Added a new unit test for 80-bit float conversion. Added 80-bit float initialization in an existing unit test.

The `#if` check in the tests is necessary since ARM does not support a wide `long double` type. Windows also resorts to 64-bit float when using `long double`. I'm not sure if these are the only platforms which don't support it. But in any case, we'll be able to find out by the failing test on other platforms as well.

**Closing issues**

None.

### DIFF
--- a/librz/util/float/float.c
+++ b/librz/util/float/float.c
@@ -2639,6 +2639,10 @@ RZ_API RZ_OWN RzFloat *rz_float_convert(RZ_NONNULL RzFloat *f, RzFloatFormat for
 	RzFloatFormat old_format = f->r;
 	bool sign = get_sign(f->s, old_format);
 	ut32 man_len = rz_float_get_format_info(old_format, RZ_FLOAT_INFO_MAN_LEN);
+	if (old_format == RZ_FLOAT_IEEE754_BIN_80) {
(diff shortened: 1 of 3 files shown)


wargio (approved):
nice patch.

wargio:
```
[XX] Fail at line 627: float80 bit value: expected +100000000000010|1101000010001001001101110100101111000110101001111110111110011110, got +100000000000010|1101000010001001001101110100101111000110101001111111000000000000.

EXP: +100000000000010|110100001000100100110111010010111100011010100111111 0111110011110
GOT: +100000000000010|110100001000100100110111010010111100011010100111111 1000000000000.
```
on ARM64. i have splitted where differs

XVilka (approved):

PR #4260, opened 2024-02-18
Proof HtPP anomaly

Rot127:

**Detailed description**

...

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/test/unit/test_util.c
+++ b/test/unit/test_util.c
@@ -83,7 +83,7 @@ bool test_htpp_anomaly(void) {
 }
 
 int all_tests() {
-	mu_run_test(test_htpp_anomaly);
+	// mu_run_test(test_htpp_anomaly);
 	mu_run_test(test_file_slurp);
 	mu_run_test(test_leading_zeros);
 	return tests_passed != tests_run;
--- a/test/unit/test_util.c
+++ b/test/unit/test_util.c
@@ -62,7 +62,28 @@ bool test_leading_zeros(void) {
 	mu_end;
 }
 
+
+bool test_htpp_anomaly(void) {
+	HtPP *ht = ht_pp_new0();
+	void *p1 = malloc(0x10);
+	void *p2 = malloc(0x20);
+	void *p3 = malloc(0x30);
+	ut64 p4 = 0x50505050505;
+	ht_pp_insert(ht, p1, (void *) 1);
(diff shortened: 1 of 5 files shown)


test/unit/test_util.c
```diff
@@ -62,7 +62,28 @@ bool test_leading_zeros(void) {
...
 }
 
+
+bool test_htpp_anomaly(void) {
+	HtPP *ht = ht_pp_new0();
+	void *p1 = malloc(0x10);
+	void *p2 = malloc(0x20);
+	void *p3 = malloc(0x30);
+	ut64 p4 = 0x50505050505;
+	ht_pp_insert(ht, p1, (void *) 1);
```
wargio, line 72: this doesn't work because it will threat `p1`, `p2`, `p3` and `p4` as strings instead of raw pointers.
You need to implement a custom hash function in order to calculate the hash.

wargio: to use raw pointers as actual input you need to use `HtUP` and cast `void*` to `ut64`

Rot127: Yeah, I got to it from `SetP`. Which uses it (incorrectly) like this. I would say to get rid of the `ht_xx_new0()` functions. First of all, they are not documented. And figuring out how the hash table works exactly, is pretty hard because of all the macros used.
So it is in-transparent, if it works or not, hence one would just assume everything is fine.

Rot127:
Some tests:

| commit | tests (no asan) |
|--------|-----------------|
| https://github.com/rizinorg/rizin/pull/4260/commits/fdd5aaa1c3a0f882a6121bd94ec24748809d2c70 | green |
| https://github.com/rizinorg/rizin/pull/4260/commits/901faeb80526f676bd5c90e44b972c2338e43e41 | green |

At least the one occasion `SetP` is used, doesn't seem to suffer from it.

test/unit/test_util.c
wargio: i actually suggest to make proper headers and implement correctly the methods and provide a ht_ss/ht_su/ht_sp types.

p371jah: Or predefined set of options

XVilka:
@Rot127 I suppose this one isn't needed anymore?

Rot127:
Indeed.

PR #4262, opened 2024-02-18
Extend `SetU`

Rot127:

**Detailed description**

Did not add the doxygen yet. Wanted to get some feedback first.

- Add a getter for length of SetU.
- Add a foreach macro.

**Test plan**

Added

**Closing issues**

...

### DIFF
--- a/librz/include/rz_util/set.h
+++ b/librz/include/rz_util/set.h
@@ -48,8 +48,8 @@ RZ_API void advance_set_u_iter(SetU *s, SetUIter *it);
  * So our condition checks if ti <= set->size.
  */
 #define set_u_foreach(set, iter) \
-	if (set) \
-		for (advance_set_u_iter(set, &iter); iter.ti <= set->size && set_u_size(set) > 0; advance_set_u_iter(set, &iter))
+	if (set && set_u_size(set) > 0) \
+		for (advance_set_u_iter(set, &iter); iter.ti <= set->size; advance_set_u_iter(set, &iter))
 
 #ifdef __cplusplus
 }
--- a/librz/include/rz_util/set.h
+++ b/librz/include/rz_util/set.h
@@ -37,19 +37,13 @@ typedef struct {
 
 RZ_API void advance_set_u_iter(SetU *s, SetUIter *it);
 
-#define set_u_iter_reset(iter) \
-	do { \
-		iter.ti = 0; \
-		iter.bi = 0; \
-		iter.v = 0; \
-	} while (0)
 /**
  * The adcvance_set_u_iter() sets iter.ti always to the entry of the next table to check.
  * So our condition checks if ti <= set->size.
  */
 #define set_u_foreach(set, iter) \
 	if (set && set_u_size(set) > 0) \
-		for (advance_set_u_iter(set, &iter); iter.ti <= set->size; advance_set_u_iter(set, &iter))
+		for (iter.ti = 0, iter.bi = 0, iter.v = 0, advance_set_u_iter(set, &iter); iter.ti <= set->size; advance_set_u_iter(set, &iter))
 
 #ifdef __cplusplus
 }
--- a/librz/include/rz_util/set.h
+++ b/librz/include/rz_util/set.h
@@ -24,9 +24,32 @@ typedef HtUP SetU;
 
 RZ_API RZ_OWN SetU *set_u_new(void);
 RZ_API void set_u_add(RZ_NONNULL SetU *set, ut64 u);
+RZ_API ut64 set_u_size(SetU *s);
 RZ_API bool set_u_contains(RZ_NONNULL SetU *set, ut64 u);
 RZ_API void set_u_delete(RZ_NONNULL SetU *set, ut64 u);
 RZ_API void set_u_free(RZ_NULLABLE SetU *set);
+
+typedef struct {
+	st64 ti; ///< Table index
+	ut64 bi; ///< Bucket index
+	ut64 v; ///< Current value of the iteration
+} SetUIter;
+
+RZ_API void advance_set_u_iter(SetU *s, SetUIter *it);
+
+#define set_u_iter_reset(iter) \
+	do { \
+		iter.ti = 0; \
+		iter.bi = 0; \
+		iter.v = 0; \
+	} while (0)
+/**
+ * The adcvance_set_u_iter() sets iter.ti always to the entry of the next table to check.
+ * So our condition checks if ti <= set->size.
+ */
+#define set_u_foreach(set, iter) \
+	if (set) \
+		for (advance_set_u_iter(set, &iter); iter.ti <= set->size && set_u_size(set) > 0; advance_set_u_iter(set, &iter))
 
 #ifdef __cplusplus
 }
--- a/librz/util/set.c
+++ b/librz/util/set.c
@@ -84,7 +84,15 @@ RZ_API void set_u_free(RZ_NULLABLE SetU *set) {
 	ht_up_free((HtUP *)set);
 }
 
+/**
+ * \brief Advances an SetUIter to the next element in
+ * the set \p s and saves it in \p it->v.
+ *
+ * \param s The set to iterate over.
+ * \param it The iterator over the set.
+ */
 RZ_API void advance_set_u_iter(SetU *s, SetUIter *it) {
+	rz_return_if_fail(s && it);
 	if (it->ti >= s->size) {
 		it->ti++;
 		return;
@@ -93,7 +101,7 @@ RZ_API void advance_set_u_iter(SetU *s, SetUIter *it) {
 		if (s->table[it->ti].count == 0) {
 			continue;
 		}
-		for (; it->bi < s->table[it->ti].count;) {
+		if (it->bi < s->table[it->ti].count) {
 			it->v = s->table[it->ti].arr[it->bi].key;
 			it->bi++;
 			return;
--- a/librz/util/set.c
+++ b/librz/util/set.c
@@ -2,6 +2,7 @@
 // SPDX-FileCopyrightText: 2024 pelijah
 // SPDX-License-Identifier: MIT
 
+#include <rz_util/rz_assert.h>
 #include <rz_util/set.h>
 #include <rz_util/rz_assert.h>
 
@@ -57,6 +58,14 @@ RZ_API void set_u_add(RZ_NONNULL SetU *set, ut64 u) {
 }
 
 /**
+ * \brief Get the size of set \s.
+ */
+RZ_API ut64 set_u_size(SetU *s) {
+	rz_return_val_if_fail(s, 0);
+	return s->count;
+}
+
+/**
  * \brief Check if hash set \p set contains element \p u.
  */
 RZ_API bool set_u_contains(RZ_NONNULL SetU *set, ut64 u) {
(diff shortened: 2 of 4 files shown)


librz/util/set.c
```diff
@@ -35,6 +36,11 @@ RZ_API void set_u_add(SetU *s, ut64 u) {
 	ht_up_insert(s, u, (void *)1);
 }
 
+RZ_API ut64 set_u_len(SetU *s) {
```
XVilka: Sets do not have length, they have size

XVilka (changes requested):

librz/include/rz_util/set.h
```diff
@@ -14,6 +14,7 @@ extern "C" {
 typedef HtPP SetP;
 
 RZ_API SetP *set_p_new(void);
```
wargio: i don't even know what this is and why we have them nor where we use them.

Rot127:
How about we change this PR to re-implement the sets properly? `SetU` works fine, if it uses `HtUU` (not in this implementation thoug, push some commits later). And we can implement `SetP` by simply using `SetU` but casting the pointers to `ut64` before.

Rot127:
Also I couldn't find any unit tests specifically for the hash tables. Which should be definitely added.

ret2libc:
> Also I couldn't find any unit tests specifically for the hash tables. Which should be definitely added.

There are some in `test/unit/test_sdb_hash.c` (that's becase `Ht` was originally in SDB, then we moved the code to the main rizin repo but some stuff are still tied to SDB :( )

librz/include/rz_util/set.h
ret2libc: and btw they should have `rz_`... these definitely needs some refactoring!

ret2libc: btw, `set_p` does not seem to be used anywhere and i'm not sure it would really make sense to have, unless we define a very specific set (e.g. a set with strings as key). Otherwise, we are probably better of by leaving the developer with the choice of the type of hashtable he wants. 

IMHO we can remove `set_p`.

Rot127:
@ret2libc Any opinion on the `foreach` macro construction? I found the implemented `foreach` loop via callbacks really unpractical. But It looks a little wonky. But if you in general approve it, we could have something like this for hash tables as well.

librz/include/rz_util/set.h
```diff
@@ -19,14 +19,38 @@ RZ_API bool set_p_contains(SetP *s, const void *u);
...
+
+RZ_API void advance_set_u_iter(SetU *s, SetUIter *it);
+
+#define set_u_iter_init(iter) SetUIter iter = { 0 }
+#define set_u_iter_reset(iter) \
+	do { \
+		iter.ti = 0; \
+		iter.bi = 0; \
+		iter.v = 0; \
+	} while (0)
```
ret2libc: I think we should just have a `_reset` API and always use it, no matter if init or a reset later.

librz/include/rz_util/set.h
```diff
@@ -19,14 +19,38 @@ RZ_API bool set_p_contains(SetP *s, const void *u);
...
+/**
+ * The adcvance_set_u_iter() sets iter.ti always to the entry of the next table to check.
```
ret2libc, line 41:
```suggestion
 * The advance_set_u_iter() sets iter.ti always to the entry of the next table to check.
```

ret2libc (commented):
I like the approach with the iter if it works well ;) 
Before using it here, though, i'd like to have it in the hashtable as well so we can really test if it works well (with ASAN).

XVilka (changes requested):
Blocked until https://github.com/rizinorg/rizin/pull/4415 is merged.

XVilka:
```
 test/unit/test_util.c:128:28: error: code should be clang-formatted [-Wclang-format-violations]
        set_u_foreach(set_u, it) {
                                  ^
test/unit/test_util.c:129:8: error: code should be clang-formatted [-Wclang-format-violations]
                        x++;
                            ^
```

librz/include/rz_util/set.h
Rot127: Removed the init and changed the unit test in a way, so it can be used as an example.

librz/include/rz_util/set.h
```diff
@@ -24,10 +24,33 @@ typedef HtUP SetU;
...
+ * So our condition checks if ti <= set->size.
+ */
+#define set_u_foreach(set, iter) \
+	if (set) \
+		for (advance_set_u_iter(set, &iter); iter.ti <= set->size && set_u_size(set) > 0; advance_set_u_iter(set, &iter))
```
p371jah: Wouldn't it be better to call `set_u_size()` just once?
```suggestion
	if (set && set_u_size(set) > 0) \
		for (advance_set_u_iter(set, &iter); iter.ti <= set->size; advance_set_u_iter(set, &iter))
```

librz/util/set.c
```diff
@@ -2,6 +2,7 @@
 // SPDX-FileCopyrightText: 2024 pelijah
 // SPDX-License-Identifier: MIT
 
+#include <rz_util/rz_assert.h>
 #include <rz_util/set.h>
 #include <rz_util/rz_assert.h>
```
p371jah: Duplicate
```suggestion
#include <rz_util/set.h>
#include <rz_util/rz_assert.h>
```

librz/util/set.c
```diff
@@ -75,3 +83,22 @@ RZ_API void set_u_delete(RZ_NONNULL SetU *set, ut64 u) {
 RZ_API void set_u_free(RZ_NULLABLE SetU *set) {
 	ht_up_free((HtUP *)set);
 }
+
+RZ_API void advance_set_u_iter(SetU *s, SetUIter *it) {
```
XVilka, line 94: Doxygen, please. Also check for NULL, etc

XVilka (changes requested):

XVilka (approved):

p371jah:
Ideally `foreach` macro should be implemented for HT and be adapted for sets then. But it might be a tough task to implement universal macro considering custom HTs like SdbHt. So for now it's ok to have `foreach` for sets only. And yet we also need the same for SetS.

librz/include/rz_util/set.h
```diff
@@ -24,10 +24,33 @@ typedef HtUP SetU;
...
+		iter.bi = 0; \
+		iter.v = 0; \
+	} while (0)
+/**
+ * The adcvance_set_u_iter() sets iter.ti always to the entry of the next table to check.
+ * So our condition checks if ti <= set->size.
+ */
+#define set_u_foreach(set, iter) \
+	if (set && set_u_size(set) > 0) \
+		for (advance_set_u_iter(set, &iter); iter.ti <= set->size; advance_set_u_iter(set, &iter))
```
p371jah: Just wondering can it be done without `set_u_iter_reset`?
```suggestion
	if (set && set_u_size(set) > 0) \
		for (iter.ti = 0, iter.bi = 0, iter.v = 0, advance_set_u_iter(set, &iter); iter.ti <= set->size; advance_set_u_iter(set, &iter))
```

librz/include/rz_util/set.h
```diff
@@ -24,10 +24,33 @@ typedef HtUP SetU;
...
+	do { \
+		iter.ti = 0; \
+		iter.bi = 0; \
+		iter.v = 0; \
+	} while (0)
+/**
+ * The adcvance_set_u_iter() sets iter.ti always to the entry of the next table to check.
+ * So our condition checks if ti <= set->size.
+ */
+#define set_u_foreach(set, iter) \
```
p371jah, line 44: Add to `ForEachMacros` list in clang-format config

librz/include/rz_util/set.h
Rot127: > Just wondering can it be done without `set_u_iter_reset`?

The `set_u_iter_reset` is only there, so the same iter can be used twice in a function, without the need to init another one.
Also it is supposed to signal, that the `iter` can not just be reused without any extra steps (resetting it).

I tried in the beginning to implement it with a single pointer for value `v`, but failed.
I think the problem was, that the logic of `bi/ti` index tracking of `advance_set_u_iter()`, doesn't work in the for loop alone.

Coming up with this version already took a while. But then again, I am not a C magician. So maybe there is a more elegant solution.

p371jah: So have you tried the suggested one?

Rot127: Ah, where is my head these days -.- Of course it works. Thanks for the suggestion.

XVilka (approved):

XVilka:
@Rot127 clang-format:
```
 Formatting [1762/1785] test/unit/test_util.c
test/unit/test_util.c:88:15: error: code should be clang-formatted [-Wclang-format-violations]
        set_u_foreach(set_u, it) {
                     ^
test/unit/test_util.c:100:15: error: code should be clang-formatted [-Wclang-format-violations]
        set_u_foreach(set_u, it) {
                     ^
test/unit/test_util.c:106:15: error: code should be clang-formatted [-Wclang-format-violations]
        set_u_foreach(set_u, it) {
                     ^
test/unit/test_util.c:125:15: error: code should be clang-formatted [-Wclang-format-violations]
        set_u_foreach(set_u, it) {
                     ^
```
Please also rebase on latest `dev`

p371jah (changes requested):
`foreach` and size method for `SetS`

test/unit/test_util.c
```diff
@@ -62,7 +63,76 @@ bool test_leading_zeros(void) {
...
+	set_u_delete(set_u, 0x7070707);
+	mu_assert_eq(set_u_size(set_u), 2, "Length wrong.");
+
+	size_t x = 0;
+	SetUIter it;
+	set_u_foreach (set_u, it) {
+		x++;
+		bool matches = it.v == 0x5050505 || it.v == 0x6060606;
+		mu_assert_true(matches, "Set contained ill-formed value.");
+	}
```
p371jah, line 92: I would prefer `rz_list_foreach` like variant with separate `value` variable:
```suggestion
	SetUIter it;
	ut64 val;
	set_u_foreach (set_u, it, val) {
		x++;
		bool matches = val == 0x5050505 || val == 0x6060606;
		mu_assert_true(matches, "Set contained ill-formed value.");
	}
```

p371jah: @ret2libc your opinion?

Rot127: Don't really have strong feelings about it. Separated variable means the type is more explicit, which is probably worth the trade against the extra line. But as long as we have a `foreach` without a callback, I am happy.

p371jah: https://github.com/rizinorg/rizin/blob/ea460e54c28d7e44d123fe59bf123fda12e3bcd2/librz/include/rz_list.h#L35

Adapted for your case: `(val = iter.v, 1)`

XVilka: I think the one with the explicit value is better.

XVilka: It will be more important for the SetP, where we have a room for type ambiguity.

wargio: i think the value should always be a pointer. so you can read and write the value.

Rot127:
Then let me extend the PR to include more foreach macros:

So the todo list would be:

- [ ] Use @pelijah suggested changes
- Add foreach for:
  - [ ] HtUU (and in extend SetU)
  - [ ] HtUS (and in extend SetS)
  - [ ] HtSS, HtSU
  - [ ] Testing
- [ ] Value should be always be a pointer, so writes are possible.

test/unit/test_util.c
p371jah: > i think the value should always be a pointer. so you can read and write the value.

For HT, yes. But in case of sets value is a key.

p371jah:
If you are going to implement `foreach` for HTs, it would be better to rename `ht_xx_foreach` APIs to avoid code refactoring.
Here an idea of iterator type for `ht_inc.h` based on yours SetU iterator:
```c
typedef struct Ht_(iter_t) {
	st64 ti; ///< Table index
	ut64 bi; ///< Bucket index
	HT_(Kv) *kv; ///< Current KV-pair.  Stop iteration if kv is NULL. 
} HT(Iter);
```
Also you need to provide in `ht_inc.c` some sort of `Ht_(iter_init)` and `Ht_(iter_next)` so you can just use the expression below in `foreach` macros:
```c
for (Ht_(iter_init)(&iter); iter.kv != NULL && ... ; Ht_(iter_next)(&iter))
```

Rot127:
Implemented by https://github.com/rizinorg/rizin/pull/4500

PR #4280, opened 2024-02-21
Weaken f80 assertions in test_float for non-x86

thestr4ng3r:

**Detailed description**

OpenBSD/sparc64 uses 128-bit floats as long double by default while our tests seeminly assumed either 64-bit or 80-bit, both of which produce different results than 128 here. Since 80-bit is x86-specific, we check only the bits we know will be identical between other architectures.

**Test plan**

`test/unit/test_float` passes on all platforms

### DIFF
--- a/test/unit/test_float.c
+++ b/test/unit/test_float.c
@@ -618,16 +618,23 @@ bool float_print_num(void) {
 	rz_float_free(f64);
 
 	RzFloat *f80 = rz_float_new_from_f80(13.0335l);
-	/* Need the check since ARM does not support 128-bit long doubles and Windows ignores it. */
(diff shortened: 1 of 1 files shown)


wargio (approved):

DMaroo:
Why have extra code to mask out final bytes/bits? I would prefer if we had this test where we can keep track of how the architectures deal with 80-bit floats. If someone wants to have a look at what happens to their 80-bit floats in, say ARM64 or x86 Windows, they can see the test and figure out that the last few bits get zeroed out. This would be better than just seeing a masked representation.

Moreover, removing the masking code = less code = (IMHO) better and simpler.

DMaroo (approved):
Thanks for the catch!

PR #4289, opened 2024-02-23
Fix mismatch 'rz_cons_break_push' and 'rz_cons_break_pop' calls.

### LINKED ISSUE #2552: no graph/diassembly/nothing when opening raspberry pi kernel firmware
**Environment information**
* Operating System: Ubuntu 20.04
* Cutter version: 1.12.0
* Obtained from:
  - [] Built from source
  - [ ] Distribution repository
* File format:

**Describe the bug**

Opening this or many other kernel modules from raspberry pi firmware repo: https://github.com/raspberrypi/firmware/tree/master/modules/5.10.4%2B/kernel/drivers/char/broadcom will give no disassembly, or nothing at all. I click on the functions on the left and on the right I see nothing.

**To Reproduce**

1. Open .ko file
2. Double click one of the functions

**Expected behavior**

Disassembly, graph, etc for the module

**Screenshots**

cutter

### LINKED ISSUE #3275: Problem Empty disassembly
**Environment information**
* Operating System: Windows 10 22H2
* Cutter version: 2.3.2-HEAD-1d07ebc
* Obtained from: https://github.com/rizinorg/cutter/releases
* File format: pe64
* Type: EXEC (Executable file)

**Describe the bug**

My friend, after analyzing the large files, we cannot see anything in the disassembly , the disassembly is empty as in the picture. This error only occurs when large files are analyzed.

**Expected behavior**

The Disassembly section should not be empty, regardless of whether the file is large or small.

**Screenshots**

image

**Additional context**

There are two things I would like to add: When I save the project with an empty disassembly section and close the application, it continues to run in the background with full power consumption.
When I kill the application and load the saved project, the empty disassembly section appears and the problem disappears.

Rot127:

**Detailed description**
If a function calls `rz_cons_break_push()` but never calls `rz_cons_break_pop()` before return, the stack count of `RzConsContext->break_stack` contains too many elements (each time one too much).

This in turn will lead to not resetting `RzConsContext->breaked` flag. Because the flag is only set to false, if `rz_stack_is_empty(context->break_stack) == true` (in `rz_cons_context_break_push()`).

This wasn't a problem so far, because `RzConsContext->breaked` is simply never set to true (exceptions are some timeout cases as far as I can see). Also these cases when `rz_cons_break_pop()` was forgotten to be called, were edge error cases. So not often hit.

But if Rizin is usd by Cutter `RzConsContext->breaked` is set to `true`, if an `AnalysisTask` interrupt is handled (in `AnalysisTask::interrupt()`). This interrupt is triggered for example, when the introduction dialog is closed and the main Cutter window opens (after the optional `aaa`).

Now, if the binary file was analysed with `aaa`, and a lot of error cases were hit, those error cases sometimes never called `rz_cons_break_pop()` before returning from their function. Although, of course, they should have to the `RzConsContext->break_stack` is in a proper state.

This means, when the main Cutter window opens binary files which trigger many error edge cases, the `RzConsContext->break_stack` is not empty
(because of the not executed `rz_cons_break_pop()`).

This also means, that the last thing done, was setting `RzConsContext->breaked = true` (by `AnalysisTask::interrupt()`).

If Cutter wants to show some disassembly, it calls `rz_core_print_disasm()` which checks `RzConsContext == false` via `rz_cons_is_breaked()`. This condition is never true, because the flag was not reset to `false` because the stack was never empty. So it returns before anything was disassembled.

Hence Cutter gets no disassembly text.

**Test plan**

All green

**Closing issues**

Fixes https://github.com/rizinorg/cutter/issues/2552
Fixes https://github.com/rizinorg/cutter/issues/3275

It should at least.

### DIFF
--- a/librz/analysis/reflines.c
+++ b/librz/analysis/reflines.c
@@ -258,6 +258,7 @@ RZ_API RzList /*<RzAnalysisRefline *>*/ *rz_analysis_reflines_get(RzAnalysis *an
 
 sten_err:
 list_err:
+	rz_cons_break_pop();
 	rz_list_free(sten);
 	rz_list_free(list);
 	return NULL;
--- a/librz/analysis/rtti_msvc.c
+++ b/librz/analysis/rtti_msvc.c
@@ -190,6 +190,7 @@ static RZ_OWN RzList /*<rtti_base_class_descriptor *>*/ *rtti_msvc_read_base_cla
 			ut8 tmp[4] = { 0 };
 			if (!context->analysis->iob.read_at(context->analysis->iob.io, addr, tmp, 4)) {
 				rz_list_free(ret);
+				rz_cons_break_pop();
 				return NULL;
 			}
 			ut32 (*read_32)(const void *src) = context->analysis->big_endian ? rz_read_be32 : rz_read_le32;
--- a/librz/core/canalysis.c
+++ b/librz/core/canalysis.c
@@ -2801,6 +2801,7 @@ RZ_API RzList /*<RzAnalysisCycleHook *>*/ *rz_core_analysis_cycles(RzCore *core,
 			if (!ch) {
 				rz_analysis_cycle_frame_free(cf);
 				rz_list_free(hooks);
+				rz_cons_break_pop();
 				return NULL;
 			}
 			ch->addr = addr;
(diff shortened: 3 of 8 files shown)


XVilka (approved):
It fixes both problems for me. Amazing job!

XVilka:
Long-term though, we should think how to avoid this kind of situation, since "counting" RzCons breaks every time isn't sustainable, especially if we are going to make analysis multithreaded.

wargio (approved):

Rot127:
The good thing is though, that the push and pops where always contained in a single function. So if we have RzCons properly implemented with mutexes or even can synchronize it somehow over different analysis tasks, it would work.
Point is, the more thought we put into RzCons, the less trouble those things make.

PR #4294, opened 2024-02-24
docs(PACKAGERS): fix meson option names

chenrui333:
the option is not valid, thus removing it

---

relates to https://github.com/Homebrew/homebrew-core/pull/164020

### DIFF
--- a/doc/PACKAGERS.md
+++ b/doc/PACKAGERS.md
@@ -78,7 +78,7 @@ At the time of writing, these are:
 * `use_sys_zstd`
 * `use_sys_xxhash`
 * `use_sys_openssl`
-* `use_sys_mspack`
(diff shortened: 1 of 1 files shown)


XVilka (changes requested):
It's actually `use_sys_libmspack`, see full list at `meson_options.txt`. So it was just a typo. Same, with few others, e.g. `use_sys_libzstd`.

chenrui333:
oh, ok, I actually did not look further, I can update the PR to use libmsgpack

chenrui333:
should be good now

XVilka (approved):

XVilka:
Thank you!

chenrui333:
Thanks @XVilka!

Issue #3786, opened 2023-08-25
Provide hashes and signatures for releases


XVilka:
Just a pretty standard GPG and SHA256 information for all assets attached to the GitHub release.

- https://github.com/rnpgp/release-sign
- https://gist.github.com/cyphar/21a1b213901a0a0ec19764ac03646ebe
- https://sequoia-pgp.org/blog/2023/07/05/202307-sq-commandline-improvements/

ret2libc:
I'm not sure I'd trust https://github.com/rnpgp/release-sign . I'd just go with a manual process following the second link.

We should also add it to the RELEASE issue template.

ret2libc:
I think we should use https://www.sigstore.dev/ .

XVilka:
@ret2libc last time I checked it mainly targeted containers.

ret2libc:
Nah, it works for all kind of releases.

XVilka:
No time for this in this release. Moving to the next one

PR #4337, opened 2024-03-06
Add new release procedure for Rizin with manual tagging and release creation.

Rot127:

**Detailed description**

Github releases doesn't seem to work reliably. So this adds the manual release process.

Still missing:

- [ ] Add a script to download the packages and source code to attach to the release.
- [ ] Add script to checklist.

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/doc/release-template.md
+++ b/doc/release-template.md
@@ -26,10 +26,8 @@ content. Go step-by-step through it.
 - [ ] Finalize release notes.
 - [ ] Wait until all runs completed and succeeded on `stable`.
 - [ ] Tag procedure (assumes remote `upstream` is set to `git@github.com:rizinorg/rizin.git`)
-  - [ ] Make sure local `git` is set up with a GPG key for signing, and it is known to Github (sign a test commit on a dummy branch, push and check for `verified` label on Github).
   - [ ] Ensure you have the latest `stable` branch *locally*: `git fetch upstream && git checkout upstream/stable`
-  - [ ] Use git *locally* to create an annotated tag: `git tag -a vX.Y.Z` (editor will open for tag message).
-    - [ ] Add the release notes as tag message.
+  - [ ] Use git *locally* to create a tag: `git tag vX.Y.Z`.
   - [ ] Push tag `git push upstream vX.Y.Z`
 - [ ] Repeat for Cutter
 - [ ] Check on Github: tags point to the correct commit in the stable branch.
--- a/doc/release-template.md
+++ b/doc/release-template.md
@@ -24,7 +24,16 @@ content. Go step-by-step through it.
 - [ ] Replace `RZ_PIPE_PY_VERSION` and `RZ_GHIDRA_VERSION` in [`Dockerfile`](https://github.com/rizinorg/rizin/blob/dev/Dockerfile) with the right commits from those repositories. Merge the commit with the Rizin changes in the `stable` branch.
 - [ ] Quick test again that Rizin and Cutter work together.
 - [ ] Finalize release notes.
-- [ ] Release Rizin and Cutter by making their release public on GitHub.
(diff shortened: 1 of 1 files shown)


ret2libc:
What's the issue with GitHub releases?

doc/release-template.md
```diff
@@ -24,7 +24,19 @@ content. Go step-by-step through it.
...
+- [ ] Tag procedure (assumes remote `upstream` is set to `git@github.com:rizinorg/rizin.git`)
+  - [ ] Make sure local `git` is set up with a GPG key for signing, and it is known to Github (sign a test commit on a dummy branch, push and check for `verified` label on Github).
+  - [ ] Ensure you have the latest `stable` branch *locally*: `git fetch upstream && git checkout upstream/stable`
+  - [ ] Use git *locally* to create an annotated tag: `git tag -a vX.Y.Z` (editor will open for tag message).
+    - [ ] Add the release notes as tag message.
+  - [ ] Push tag `git push upstream vX.Y.Z`
+- [ ] Repeat for Cutter
+- [ ] Check on Github: tags point to the correct commit in the stable branch.
+  - [ ] If incorrect, delete tag on `upstream` and do it again.
+- [ ] Make public release
```
XVilka: This should be removed or changed, since the CI makes the draft release and uploads artifacts automatically upon push to `stable`.

XVilka (changes requested):

Rot127:
@ret2libc It now assigned twice in a row the tag to the `dev` branch. And not to the `stable`.
Also the signing doesn't work reliably. So we can just as well sign it on our own.

ret2libc:
> @ret2libc It now assigned twice in a row the tag to the `dev` branch. And not to the `stable`. Also the signing doesn't work reliably. So we can just as well sign it on our own.

Where exactly? In the CI? Maybe we should use `target_commitish` in the `softprops/action-gh-release@v1` action. With regards to signing, there is https://sigstore.dev/ which allows automatic signatures from GH CI without any manual operation and without us having to deal with GPG keys, rotating them, updating them, etc.

XVilka:
> > @ret2libc It now assigned twice in a row the tag to the `dev` branch. And not to the `stable`. Also the signing doesn't work reliably. So we can just as well sign it on our own.
> 
> Where exactly? In the CI? Maybe we should use `target_commitish` in the `softprops/action-gh-release@v1` action. With regards to signing, there is https://sigstore.dev/ which allows automatic signatures from GH CI without any manual operation and without us having to deal with GPG keys, rotating them, updating them, etc.

Not CI - tag was created on `dev` if we create it via GitHub interface from the Draft Release.

XVilka:
Please remove instructions about signing, since we need first to decide on how exactly we do that. GitHub "standard" verification isn't enough, especially since they changed ZIP archives once. This is tracked in https://github.com/rizinorg/rizin/issues/3786

ret2libc:
> > > @ret2libc It now assigned twice in a row the tag to the `dev` branch. And not to the `stable`. Also the signing doesn't work reliably. So we can just as well sign it on our own.
> > 
> > 
> > Where exactly? In the CI? Maybe we should use `target_commitish` in the `softprops/action-gh-release@v1` action. With regards to signing, there is https://sigstore.dev/ which allows automatic signatures from GH CI without any manual operation and without us having to deal with GPG keys, rotating them, updating them, etc.
> 
> Not CI - tag was created on `dev` if we create it via GitHub interface from the Draft Release.

Yeah I understand, I'm afraid that it's caused by the wrong commit being referenced in the draft release when we create it in the CI. IIRC I "recently" changed that part of the CI as we had to switch GH action for release, so it might behave slightly different and cause this behaviour.

Rot127:
Guess this is obsolete now.

Issue #4340, opened 2024-03-07
Function `rz_type_offset_by_path` is broken


rockrid3r:
## System info
* System: `debian 12, x86-64`.
* `rizin -v` output: 
```
rizin 0.8.0 @ linux-x86-64
commit: 29f9c8a9da7fada66271b8fb0b380c03b1034752
```

## TL; DR for bug

Function `rz_type_offset_by_path` does not work. It is marked with `RZ_API` so it should work. For example, the code:
```c
rz_type_offset_by_path(typedb, "a.b[20].c");
```
fails with `"Type path: unexpected char (\"%s\")", path - 1)"` in stderr.

This function is a wrapper for `path_walker`. And  actually `path_walker` is broken.

## Digging into it
See this in `rz_type_offset_by_path`:
```c
/**
 * \brief Returns the offset of the member given path
 *
 * Resolves the path in the form of "a.b[20].c" where "b" is
 * a member of "a" and "c" is a member of "b" array and located
 * inside the 20-th element, and calculates the offset.
 * Opposite function of "rz_type_path_by_offset"
 *
 * \param type RzTypePath
 */
RZ_API st64 rz_type_offset_by_path(const RzTypeDB *typedb, RZ_NONNULL const char *path) {
	rz_return_val_if_fail(typedb && path, -1);
	return path_walker(typedb, path);
}
```
It means the function `path_walker` should accept lines like "a.b[20].c".

And what it does:
```c
static st64 path_walker(const RzTypeDB *typedb, const char *path) {
	rz_return_val_if_fail(typedb && path, -1);
	const char *member;
	size_t membsize;
	ut64 index;
	st64 offset = 0;
	RzType *parent = NULL;
	const char *path_begin = path;
	while (*path) {
		switch (*path++) {
		case '\0':
			break;
		case '[':
			member = path;
			index = (ut64)strtoull(member, (char **)&path, 10);
			if (member == path || *path != ']') {
				eprintf("Type path: expected ] (\"%s\")", path - 1);
				return -1;
			}
			++path;
			if (!parent || parent->kind != RZ_TYPE_KIND_ARRAY) {
				return -1;
			}
			offset += rz_type_db_get_bitsize(typedb, parent) * index;
			break;
		case '.':
			member = path;
			for (membsize = 0; member[membsize]; ++membsize) {
				if (strchr(".[", member[membsize])) {
					break;
				}
			}
			if (membsize == 0) {
[...]

PR #4341, opened 2024-03-07
fix path_walker & add unit tests for offset_by_path

rockrid3r:
# SQUASH ME

- [x] ~~I've updated the [rizin book](https://github.com/rizinorg/book) with the relevant information (if needed)~~ there is no need to update rizin book.

**Detailed description**

Just rewrote the function `path_walker`. Added unit tests for it.

**Test plan**

Created unit tests for structs/unions/arrays.

**Closing issues**

#4340


test/unit/test_type.c
```diff
@@ -1705,6 +1705,146 @@ bool test_path_by_offset_typedef(void) {
...
+	char *error_msg = NULL;
+	RzType *ttype = rz_type_parse_string_single(typedb->parser, "struct Hello { int32_t a; uint32_t b; };", &error_msg);
+	mu_assert_notnull(ttype, "type parse successful");
+	mu_assert_eq(ttype->kind, RZ_TYPE_KIND_IDENTIFIER, "parsed type");
+	mu_assert_streq(ttype->identifier.name, "Hello", "parsed type");
+
+	/**
+	 * Set offsets manually.
+	 * Since parse_struct_node() in librz/type/parser/types_parser.c sets them to 0.
+	 */
```
ret2libc: Is this behaviour correct? cc @XVilka

test/unit/test_type.c
```diff
@@ -1705,6 +1705,146 @@ bool test_path_by_offset_typedef(void) {
...
+	btype = rz_type_get_base_type(typedb, ttype);
+	mu_assert_notnull(btype, "btype get successful");
+	RzTypeUnionMember *umemb_it;
+	rz_vector_foreach(&btype->struct_data.members, umemb_it) {
+		if (!strcmp(umemb_it->name, "ulu")) {
+			umemb_it->offset = 0;
+		} else if (!strcmp(umemb_it->name, "mulu")) {
+			umemb_it->offset = 0;
+		} else if (!strcmp(umemb_it->name, "urshak")) {
+			umemb_it->offset = 0;
```
ret2libc: Weren't they already set all to 0 according to the previous comment?

librz/type/path.c
```diff
@@ -289,10 +375,18 @@ RZ_API RZ_OWN RzList /*<RzTypePath *>*/ *rz_type_db_get_by_offset(const RzTypeDB
...
+	eprintf("here\n");
 	if (!btype || btype->kind != RZ_BASE_TYPE_KIND_STRUCT) {
+		if (!btype) {
+			eprintf("return offset 0: !btype\n");
+		} else {
+			eprintf("return offset 0: btype->kind != RZ_BASE_TYPE_KIND_STRUCT\n");
+		}
 		return 0;
 	}
+	eprintf("here\n");
```
ret2libc: just to remember to remove various debug printf.

librz/type/path.c
```diff
@@ -36,74 +36,160 @@ RZ_API void rz_type_path_free(RZ_NULLABLE RzTypePath *tpath) {
...
-	ut64 index;
+
+	size_t i;
+	for (i = 0; isalnum(path[i]); ++i)
+		;
+
+	if (path[i] != '.') {
+		eprintf("Unexpected character '%c' at position %lu\n", path[i], i);
+		return -1;
+	}
```
ret2libc: I think you can use `rz_str_strchr`

librz/type/path.c
```diff
@@ -36,74 +36,160 @@ RZ_API void rz_type_path_free(RZ_NULLABLE RzTypePath *tpath) {
...
+
+	RzType *parent = rz_type_identifier_of_base_type_str(typedb, strndup(path, i));
```
ret2libc: Should the strndup string be assigned to a var and freed later?

librz/type/path.c
```diff
@@ -36,74 +36,160 @@ RZ_API void rz_type_path_free(RZ_NULLABLE RzTypePath *tpath) {
...
+
+static RzType *path_walker_parse_dot(const RzTypeDB *typedb, RzType *parent, const char *path, size_t *i, st64 *offset) {
+	++*i;
+
+	size_t tok_beg = *i;
+	for (; isalnum(path[*i]); ++*i)
+		;
+
+	const char *tok = strndup(&path[tok_beg], *i - tok_beg);
+
```
ret2libc: again, i feel like there are too many "low level" string operations. Are you trying to find all the dot-separated strings? If so, we may want to use `rz_str_split_list` or a function like that and then handle each token in a cleaner way.

librz/type/path.c
```diff
@@ -36,74 +36,160 @@ RZ_API void rz_type_path_free(RZ_NULLABLE RzTypePath *tpath) {
...
-	const char *member;
-	size_t membsize;
-	ut64 index;
+
+	size_t i;
+	for (i = 0; isalnum(path[i]); ++i)
+		;
+
+	if (path[i] != '.') {
+		eprintf("Unexpected character '%c' at position %lu\n", path[i], i);
```
wargio: use RZ_LOG_ERROR for errors. eprintf are only for dev debugging purposes.

test/unit/test_type.c
XVilka: It was planned to remove offsets completely, because they should be calculated on the fly, depending on the architecture, CPU, OS selected. So this part could probably be omitted, if not necessary.

librz/type/path.c
rockrid3r: fixed for errors, thanks. I left some `eprintf` for debug output.

librz/type/path.c
rockrid3r: removed :)

test/unit/test_type.c
rockrid3r: Hello. Yes, they are. Fixed, thanks

librz/type/path.c
rockrid3r: I don't think `rz_str_strchr` will work if the code above is not rethinked & reprogrammed fully. The code above produces error if char is not alphanumeric or a dot or a bracket. But `rz_strchr(path, ".")` will let dirty-symbols to be in place.

librz/type/path.c
rockrid3r: Fixed here, thanks

librz/type/path.c
rockrid3r: What I am doing is that:
> go through string and parse it on the fly.

What you want me to do:
> Pre-parse the string: split it on dots, verify that all bracket sequences are ok, verify that there is nothing but alphanumeric characters, dots and brackets.

I've chosen the first approach because
1. it uses less memory.
2. it is easier to code
3. on the fly it is easier to verify that passed string is valid than to pre-check that string is valid once and for all.

I agree that your approach will look more clean, yeah. If you really want the code to be perfect, I can redo it.

rockrid3r: So what do you think?

librz/type/path.c
```diff
@@ -68,7 +68,10 @@ static RzType *path_walker_parse_bracket(const RzTypeDB *typedb, RzType *parent,
 			return NULL;
 		}
 
-		size_t idx = strtoull(strndup(&path[tok_beg], *i - tok_beg), NULL, 10);
+		char *idx_str = strndup(&path[tok_beg], *i - tok_beg);
+		size_t idx = strtoull(idx_str, NULL, 10);
```
wargio: use `rz_str_ndup()` and `rz_num_math(NULL, idx_str)`

rockrid3r: fixed, thanks

librz/type/path.c
```diff
@@ -291,6 +386,11 @@ RZ_API ut64 rz_type_db_struct_member_packed_offset(RZ_NONNULL const RzTypeDB *ty
 	rz_return_val_if_fail(typedb && name && member, 0);
 	RzBaseType *btype = rz_type_db_get_base_type(typedb, name);
 	if (!btype || btype->kind != RZ_BASE_TYPE_KIND_STRUCT) {
+		if (!btype) {
+			eprintf("return offset 0: !btype\n");
+		} else {
+			eprintf("return offset 0: btype->kind != RZ_BASE_TYPE_KIND_STRUCT\n");
+		}
```
ret2libc: if this should be surfaced to users, it's better to use `RZ_LOG_ERROR` or similar.

librz/type/path.c
ret2libc: it's ok as you are doing, thanks

librz/type/path.c
rockrid3r: Nope, it shouldn't :) 
Deleted.

test/unit/test_type.c
```diff
@@ -1705,6 +1705,136 @@ bool test_path_by_offset_typedef(void) {
...
 }
 
+bool test_offset_by_path_struct(void) {
+	RzTypeDB *typedb = rz_type_db_new();
+	const char *types_dir = TEST_BUILD_TYPES_DIR;
+	rz_type_db_init(typedb, types_dir, "x86", 64, "linux");
+
+	// simple
+
+	// eprintf("=============== SIMPLE ===============\n");
```
XVilka: Remove this

test/unit/test_type.c
XVilka: Please remove this comment also

test/unit/test_type.c
```diff
@@ -1705,6 +1705,136 @@ bool test_path_by_offset_typedef(void) {
...
+
+	long long offset;
+	offset = rz_type_offset_by_path(typedb, "Hello.a");
+	mu_assert_eq(offset, 0, "offset");
+	offset = rz_type_offset_by_path(typedb, "Hello.b");
+	mu_assert_eq(offset, 32, "offset");
+
+	// recursive
+
+	// eprintf("============= RECURSIVE ==============\n");
```
XVilka: Same, remove

test/unit/test_type.c
```diff
@@ -1705,6 +1705,136 @@ bool test_path_by_offset_typedef(void) {
...
+bool test_offset_by_path_array(void) {
+	RzTypeDB *typedb = rz_type_db_new();
+	const char *types_dir = TEST_BUILD_TYPES_DIR;
+	rz_type_db_init(typedb, types_dir, "x86", 64, "linux");
+	RzBaseType *btype;
+	RzType *ttype;
+
+	// simple
+
+	// eprintf("=============== SIMPLE ===============\n");
```
XVilka: You got the idea

test/unit/test_type.c
```diff
@@ -1705,6 +1705,136 @@ bool test_path_by_offset_typedef(void) {
...
+	mu_assert_streq(ttype->identifier.name, "Hello", "parsed type");
+
+	/**
+	 * Set offsets manually.
+	 * Since parse_struct_node() in librz/type/parser/types_parser.c sets them to 0.
+	 */
+	btype = rz_type_get_base_type(typedb, ttype);
+	mu_assert_notnull(btype, "btype get successful");
+	RzTypeStructMember *memb_it;
+	rz_vector_foreach(&btype->struct_data.members, memb_it) {
```
XVilka, line 1778: Remove these as well

test/unit/test_type.c
```diff
@@ -1705,6 +1705,136 @@ bool test_path_by_offset_typedef(void) {
...
+	// simple
+
+	// eprintf("=============== SIMPLE ===============\n");
+	char *error_msg = NULL;
+	ttype = rz_type_parse_string_single(typedb->parser, "struct Hello { int32_t a; uint32_t b; };", &error_msg);
+	mu_assert_notnull(ttype, "type parse successful");
+	mu_assert_eq(ttype->kind, RZ_TYPE_KIND_IDENTIFIER, "parsed type");
+	mu_assert_streq(ttype->identifier.name, "Hello", "parsed type");
+
+	/**
```
XVilka: And this

test/unit/test_type.c
```diff
@@ -1705,6 +1705,136 @@ bool test_path_by_offset_typedef(void) {
...
+	}
+	btype->size = 64;
+
+	ttype = rz_type_parse_string_single(typedb->parser, "struct HelloWrap { int32_t a; Hello harr[20]; };", &error_msg);
+	mu_assert_notnull(ttype, "type parse successful");
+	mu_assert_eq(ttype->kind, RZ_TYPE_KIND_IDENTIFIER, "parsed type");
+	mu_assert_streq(ttype->identifier.name, "HelloWrap", "parsed type");
+
+	/**
+	 * Set offsets manually.
```
XVilka: And this

librz/type/path.c
```diff
@@ -36,74 +36,169 @@ RZ_API void rz_type_path_free(RZ_NULLABLE RzTypePath *tpath) {
...
+		++nd;
+		curd_off *= typd->array.count; // TODO: overflow
+
+		typd = typd->array.type;
+	}
+	curd_off *= rz_type_get_base_type(typedb, typd)->size; // elem size in bits
+
+	typd = parent;
+	for (size_t id = 0; id < nd; ++id) {
+
```
XVilka: Please remove unnecessary empty lines

librz/type/path.c
```diff
@@ -36,74 +36,169 @@ RZ_API void rz_type_path_free(RZ_NULLABLE RzTypePath *tpath) {
 	free(tpath);
 }
 
+static RzType *path_walker_parse_bracket(const RzTypeDB *typedb, RzType *parent, const char *path, size_t *i, st64 *offset) {
+	size_t nd = 0;
+	RzType *typd = parent;
+	st64 curd_off = 1; // prod of all dim sizes
```
XVilka: Why curd? Please try to give more meaningful names. Also the same in the comment

librz/type/path.c
```diff
@@ -36,74 +36,169 @@ RZ_API void rz_type_path_free(RZ_NULLABLE RzTypePath *tpath) {
...
+	while (typd->kind == RZ_TYPE_KIND_ARRAY) {
+		++nd;
+		curd_off *= typd->array.count; // TODO: overflow
```
XVilka: Well, ideally it shouldn't be a TODO but a real check

librz/type/path.c
```diff
@@ -36,74 +36,169 @@ RZ_API void rz_type_path_free(RZ_NULLABLE RzTypePath *tpath) {
...
+
+		typd = typd->array.type;
+	}
+	curd_off *= rz_type_get_base_type(typedb, typd)->size; // elem size in bits
```
XVilka: And here too

librz/type/path.c
```diff
@@ -36,74 +36,169 @@ RZ_API void rz_type_path_free(RZ_NULLABLE RzTypePath *tpath) {
...
+		}
+	}
+
+	if (!cur_type) {
+		RZ_LOG_ERROR("Invalid member '%s' for parent type\n", tok);
+		*offset = -1;
+		free(tok);
+		return NULL;
+	}
+
```
XVilka, line 151: Same, too many unnecessary empty lines

XVilka (changes requested):

librz/type/path.c
rockrid3r: "current dimesion". Fixed :)

librz/type/parser/c_cpp_parser.c
```diff
@@ -302,7 +302,8 @@ RZ_API void rz_type_parse_reset(RzTypeDB *typedb) {
 }
 
 /**
- * \brief Parses the single C type definition
+ * \brief Parses the single C type definition.
```
XVilka:
```
librz/type/parser/c_cpp_parser.c:305:47: error: code should be clang-formatted [-Wclang-format-violations]
 * \brief Parses the single C type definition. 
                                              ^
```

XVilka (changes requested):
Much better now, I have a nitpick

librz/type/parser/c_cpp_parser.c
rockrid3r: Fixed :)

XVilka (approved):

XVilka:
@wargio @ret2libc @thestr4ng3r @Rot127  please take a look again before merging

librz/type/path.c
```diff
@@ -36,74 +36,167 @@ RZ_API void rz_type_path_free(RZ_NULLABLE RzTypePath *tpath) {
...
 }
 
+static RzType *path_walker_parse_bracket(const RzTypeDB *typedb, RzType *parent, const char *path, size_t *i, st64 *offset) {
+	size_t nd = 0;
+	RzType *typd = parent;
+	st64 cur_dim_off = 1; // product of all dimension sizes
+	while (typd->kind == RZ_TYPE_KIND_ARRAY) {
+		++nd;
+
+		if (typd->array.count != 0 && cur_dim_off <= ST64_MAX / typd->array.count) { // check for overflow
```
Rot127: Move the overflow check in a `static inline` function. Saves the comment and makes it more readable.

librz/type/path.c
```diff
@@ -36,74 +36,167 @@ RZ_API void rz_type_path_free(RZ_NULLABLE RzTypePath *tpath) {
 	free(tpath);
 }
 
+static RzType *path_walker_parse_bracket(const RzTypeDB *typedb, RzType *parent, const char *path, size_t *i, st64 *offset) {
```
Rot127: They are no API functions, but because they are pretty big and do complicated things I would document them nonetheless (Also please add `RZ_OWN` and `RZ_OUT`). Also below.

librz/type/path.c
```diff
@@ -36,74 +36,167 @@ RZ_API void rz_type_path_free(RZ_NULLABLE RzTypePath *tpath) {
...
+
+		if (path[*i] != ']') {
+			RZ_LOG_ERROR("Expected ']' got '%c'\n", path[*i]);
+			*offset = -1;
+			return NULL;
+		}
+
+		char *idx_str = rz_str_ndup(&path[tok_beg], *i - tok_beg);
+		size_t idx = rz_num_math(NULL, idx_str);
+		free(idx_str);
```
Rot127, line 90: Maybe you are able to just pass `path + tok_beg` to `rz_num_math` or `rz_num_get` (to save the allocation/free)? 
At least `atol` stops and returns the number, when it finds the first non-digit character. But of cause the `rz_num_X` API is not documented properly \*sigh*, so I have no idea if it behaves the same (please add the doxygen to them if you look at it).

librz/type/path.c
```diff
@@ -36,74 +36,167 @@ RZ_API void rz_type_path_free(RZ_NULLABLE RzTypePath *tpath) {
...
+
+	RzBaseType *parent_btype = rz_type_get_base_type(typedb, parent);
+	if (!parent_btype) {
+		RZ_LOG_ERROR("Could not found btype for parent '%s'\n", parent->identifier.name);
+		*offset = -1;
+		free(tok);
+		return NULL;
+	}
+
+	RzTypeStructMember *memb_it;
```
Rot127:
```suggestion
	RzTypeStructMember *memb_it = NULL;
```

librz/type/path.c
```diff
@@ -36,74 +36,167 @@ RZ_API void rz_type_path_free(RZ_NULLABLE RzTypePath *tpath) {
...
+			return NULL;
+		}
+
+		path_walker_parse_bracket(typedb, parent, path, i, offset);
+		if (*offset == -1) {
+			return NULL;
+		}
+
+		// unwrap array
+		while (parent->kind == RZ_TYPE_KIND_ARRAY) {
```
Rot127: Sorry, haven't looked at the semantic. But is this guaranteed to terminate? Meaning, will in the `parent->array.type` sequence always be a node with `RZ_TYPE_KIND_ARRAY`? Or are malfunctions possible?

librz/type/path.c
```diff
@@ -36,74 +36,167 @@ RZ_API void rz_type_path_free(RZ_NULLABLE RzTypePath *tpath) {
...
+		// so just continue.
+
+		++*i; // skip ']'
+	}
+
+	return typd;
+}
+
+static RzType *path_walker_parse_dot(const RzTypeDB *typedb, RzType *parent, const char *path, size_t *i, st64 *offset) {
+	++*i;
```
Rot127: Add `rz_return_val_if_fail()`

Rot127: Also above

librz/type/path.c
```diff
@@ -36,74 +36,167 @@ RZ_API void rz_type_path_free(RZ_NULLABLE RzTypePath *tpath) {
...
+		}
+
+		// unwrap array
+		while (parent->kind == RZ_TYPE_KIND_ARRAY) {
+			parent = parent->array.type;
+		}
+		return parent;
+	} else if (path[*i] == '.' || path[*i] == '\0') {
+		return cur_type;
+	} else {
```
Rot127: Remove `else` to save indention.

Rot127 (changes requested):

librz/type/path.c
Rot127: As @XVilka mentioned, no need to document the `rz_num` stuff.

rockrid3r: Probably it is not the best idea. 
* `path + tok_beg` is not 0-terminated. 
* using functions like `atol` on non-0-terminated strings does not look good (even if we know what char goes after digit sequence).

librz/type/path.c
rockrid3r: fixed, thanks

librz/type/path.c
rockrid3r: fixed

librz/type/path.c
rockrid3r: added

librz/type/path.c
rockrid3r: fixed

librz/type/path.c
rockrid3r: It seems I don't really understand what you've meant. The loop breaks if current node is not `RZ_TYPE_KIND_ARRAY`.

librz/type/path.c
rockrid3r: fixed.

librz/type/path.c
Rot127: Ah, ignore it. I was obviously diluted :D

librz/type/path.c
```diff
@@ -36,74 +36,199 @@ RZ_API void rz_type_path_free(RZ_NULLABLE RzTypePath *tpath) {
...
+
+/**
+ * \brief Parses the array subscription 'arr[3]'
+ * \param typedb
+ * \param elem_type Type of array element.
+ * \param path String where struct path is stored.
+ * \param i Index in path pointing to the subscription bracket '['.
+ * \param offset Value to place resulting offset. Set to -1 on error.
+ */
+static RzType *path_walker_parse_bracket(const RzTypeDB *typedb, RZ_NONNULL RzType *elem_type, RZ_NONNULL const char *path, RZ_NONNULL RZ_INOUT size_t *i, RZ_OUT st64 *offset) {
```
Rot127:
```suggestion
static RZ_BORROW RzType *path_walker_parse_bracket(const RzTypeDB *typedb, RZ_NONNULL RZ_BORROW RzType *elem_type, RZ_NONNULL const char *path, RZ_NONNULL RZ_INOUT size_t *i, RZ_OUT st64 *offset) {
```

And document the return value.

librz/type/path.c
Rot127: It doesn't need to be terminated, it is guaranteed that `tok_beg <= *i` and `!is_digit(path[*i])`. At least in your current code.
So the checks already happened, alternatively add an assert before with `rz_return_val_if_fail()`. Allocations are pretty resource extensive and not using them, if there is a simpler solution, is better.

librz/type/path.c
```diff
@@ -36,74 +36,199 @@ RZ_API void rz_type_path_free(RZ_NULLABLE RzTypePath *tpath) {
...
+
+/**
+ * \brief Walks through the given struct path and returns offset from base.
+ * \param typedb
+ * \param path Path to walk through (like "Struct.arr[30].member")
+ *
+ * \brief Splits struct path on tokens and iterates over them.
+ * \brief Calls `path_walker_parse_dot` on each.
+ */
+static st64 path_walker(const RzTypeDB *typedb, RZ_NONNULL const char *path) {
```
Rot127:
```suggestion
static st64 path_walker(RZ_NONNULL const RzTypeDB *typedb, RZ_NONNULL const char *path) {
```

librz/type/path.c
```diff
@@ -36,74 +36,199 @@ RZ_API void rz_type_path_free(RZ_NULLABLE RzTypePath *tpath) {
...
+
+/**
+ * \brief Parses the dot-separated token.
+ * \param typedb
+ * \param parent Previous token type.
+ * \param path String where struct path is stored.
+ * \param i Index in path pointing to the dot '.'.
+ * \param offset Value to place resulting offset on success. Set to -1 on error.
+ */
+static RzType *path_walker_parse_dot(const RzTypeDB *typedb, RZ_NONNULL RzType *parent, RZ_NONNULL const char *path, RZ_NONNULL RZ_INOUT size_t *i, RZ_NONNULL RZ_OUT st64 *offset) {
```
Rot127:
```suggestion
static RZ_BORROW RzType *path_walker_parse_dot(RZ_NONNULL const RzTypeDB *typedb, RZ_NONNULL RZ_BORROW RzType *parent, RZ_NONNULL const char *path, RZ_NONNULL RZ_INOUT size_t *i, RZ_NONNULL RZ_OUT st64 *offset) {
```

Rot127: Depending how the `RzType` free function works, this can lead to a double free. If the given `parent` is freed and all it's children are freed as well, the returned pointer from this function is no longer valid.
Check the free function of `RzType` please and mark the return value as either `RZ_OWN` (if doesn't free the children) or `RZ_BORROW` (if the children get freed).

Rot127 (changes requested):

librz/type/path.c
Rot127: The fix above assumes the `RzType` free function frees all the children as well.

librz/type/path.c
```diff
@@ -36,74 +36,199 @@ RZ_API void rz_type_path_free(RZ_NULLABLE RzTypePath *tpath) {
 	free(tpath);
 }
 
-static st64 path_walker(const RzTypeDB *typedb, const char *path) {
+/**
+ * \brief False if overflow detected.
+ */
+static inline bool _st64_mult_check_overflow(st64 a, st64 b) {
```
XVilka: Please use macroses from `librz/include/rz_types_overflow.h`

XVilka (changes requested):

librz/type/path.c
rockrid3r: fixed

librz/type/path.c
rockrid3r: fixed

librz/type/path.c
rockrid3r: Reasoning makes sense now. Thanks for explaining man

librz/type/path.c
rockrid3r: fixed, thanks

librz/type/path.c
rockrid3r: About the `tok_beg`. Here is how I will save the allocation.

The `rz_num_math` function internally determines the base of passed number. So using it allows to specify such paths: `"Struct.arr[0x20]"` which is useful from user's perspective. And it also frees me from determining the base myself.

librz/type/path.c
rockrid3r: Here is how I will save the allocation.

The `rz_num_math` function internally determines the base of passed number. So using it allows to specify such paths: `"Struct.arr[0x20]"` which is useful from user's perspective. And it also frees me from determining the base myself.

XVilka (approved):
LGTM from my side

XVilka:
The ASAN problem isn't specific to this PR, I saw it happened also on https://github.com/rizinorg/rizin/pull/4360
So, if @Rot127 doesn't have any objections, I think we can merge this one. ASAN problem should be addressed separately.

Rot127 (approved):
LGTM

PR #4343, opened 2024-03-07
Remove global variables from GNU z80 disassembler

moste00:

**Detailed description**

This change removes global variables from librz\arch\isa_gnu\z80. This is a fairly massive change, so sorry in advance for any reviewer, most of the diff is trivial.

The global variables that were removed were shared state that multiple functions in the file could modify, this is in essence the "OO" pattern, the functions that modified the variables were "methods" that operated on the same implicit "object". This pattern were made explicit by moving the variables into a structure, and all the functions were changed so that they accept a pointer to the structure as the first argument. The function that instantiates the structure is the one that starts the assembly process, `assemble`. This function outlives all other functions, ensuring the pointer it passes to the others is valid.

In addition to that change, several minor cleanups were also made:

-  Dead code, presumably added for debugging, was removed (see: the variable `cont`)

-  Dead code branching on the value of the variable `define_macro` was removed, since the variable has one value that no other code modifies.

- Dead code that was nested under an `#if 0` preprocessor directive. The function that used to contain this code is now empty, to be removed in a future cleanup (it's an error logging function that was presumably useful in the original context of the code)

More cleanups to this file is later needed, to remove the function `printerr` and to free the memory returned by `strdup` in the initialization of `z80buffer`.

**Test plan**

No new functionality was added so no new tests are necessary. The branch builds and all the tests in db/asm/z80 pass.

**Closing issues**

...

### DIFF
--- a/librz/arch/isa_gnu/z80/expressions.c
+++ b/librz/arch/isa_gnu/z80/expressions.c
@@ -23,7 +23,7 @@
    along with this program.  If not, see <http://www.gnu.org/licenses/>.
 */
 
-//#include "z80asm.h"
+#include "z80asm.h"
 
 /* reading expressions. The following operators are supported
  * in order of precedence, with function name:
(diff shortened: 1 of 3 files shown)


librz/arch/isa_gnu/z80/expressions.c
```diff
@@ -39,723 +39,658 @@
...
-  if (endp)
-    *endp = *p;
-  *p = delspc (*p);
-  if (verbose >= 7)
-    fprintf (stderr, "%5d (0x%04x): rd_number returned %d (%04x).\n",
-	     stack[sp].line, addr, result, result);
-  return result;
+/* print an error message, including current line and file */
+static void printerr(int error, const char *fmt, ...) {
 }
```
wargio: is this empty?

wargio (commented):
awesome job.

librz/arch/isa_gnu/z80/expressions.c
moste00: Yes, in the original code it was also empty but there was code in its body guarded under an `#if 0` preprocessor directive. I saw there was no good reason to keep maintaining that code since it's essentially comments, and I deleted it.

This function in some sense "deserves" deletion anyway because it's not following the same general approach to error reporting followed in the rest of the project, it prints directly to stderr and uses a weird `va_list` type to manipulate variadic arguments. I leave it empty instead of completely removing it because other functions in `expression.c` use it a lot, I plan to replace it with multiple functions or by writing to an error buffer in the state structure in a future PR. I didn't do any of this in this PR because it's already massive as it is.

wargio:
something is wrong when you compile it.

moste00:
> something is wrong when you compile it.

Yeah very sorry for that. I previously blamed it on the original code but it turns out that I did remove a few `const` qualifiers from the function signatures here and there and all those warnings are because of that. I put the 3 qualifiers I removed back and hopefully those are all the ones I deleted.

I moved the function `printerrr` into the .h file in an attempt to make the diff of `expressions.c` cleaner and smaller, alas git still thinks I removed the entire file and wrote it again. I'm open to any suggestions that make git see that I didn't change anything except the function signature and a few lines in the body of each function.

wargio:
if `printerrr` is an empty function, i strongly suggest to remove it and its usages.

wargio (approved):
so much better.

PR #4369, opened 2024-03-18
Add RzIL info to rz-asm and La

wargio:

**Detailed description**

Adds info about which plugin has RzIL support or not.

### DIFF
--- a/librz/core/casm.c
+++ b/librz/core/casm.c
@@ -69,11 +69,23 @@ static const char *has_esil(RzCore *core, const char *name) {
 	RzAnalysisPlugin *h;
 	rz_return_val_if_fail(core && core->analysis && name, NULL);
 	rz_list_foreach (core->analysis->plugins, iter, h) {
-		if (h->name && !strcmp(name, h->name)) {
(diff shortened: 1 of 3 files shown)


librz/core/casm.c
```diff
@@ -99,7 +111,7 @@ RZ_API RzCmdStatus rz_core_asm_plugin_print(RzCore *core, RzAsmPlugin *ap, const
 			strcat(bits, "32 ");
 		}
 		if (ap->bits & 64) {
-			strcat(bits, "64 ");
+			strcat(bits, "64");
```
XVilka, line 114: Why this change? Other bitness values have space as well.

XVilka (changes requested):
The rest LGTM

librz/core/casm.c
XVilka: Nevermind, I see that is the last value, so the space is unnecessary.

XVilka (approved):

librz/core/casm.c
wargio: because is always the last one and doesn't need the extra space.

PR #4377, opened 2024-03-22
$: Use rz_flag_get to check if flag exists

kazarmy:

**Detailed description**

This pr makes `cmd_alias` use `rz_flag_get` instead of `rz_num_get` to check whether a flag exists, so that the test works as expected.

**Test plan**

All builds are green.

**Closing issues**

...

### DIFF
--- a/librz/core/cmd/cmd_alias.c
+++ b/librz/core/cmd/cmd_alias.c
@@ -117,9 +117,9 @@ RZ_IPI RzCmdStatus rz_alias_handler(RzCore *core, int argc, const char **argv) {
 				rz_core_cmd0(core, v);
 			}
 		} else {
-			ut64 at = rz_num_get(core->num, buf + 1);
-			if (at != UT64_MAX) {
-				rz_core_seek(core, at, true);
+			RzFlagItem *flag = rz_flag_get(core->flags, buf + 1);
+			if (flag) {
+				rz_core_seek(core, flag->offset, true);
 			} else {
 				RZ_LOG_ERROR("core: unknown alias '%s'\n", buf + 1);
 				free(buf);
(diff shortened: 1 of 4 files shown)


XVilka (approved):

librz/core/cmd/cmd_alias.c
```diff
@@ -117,9 +117,9 @@ RZ_IPI RzCmdStatus rz_alias_handler(RzCore *core, int argc, const char **argv) {
 				rz_core_cmd0(core, v);
 			}
 		} else {
-			ut64 at = rz_num_get(core->num, buf + 1);
-			if (at != UT64_MAX) {
-				rz_core_seek(core, at, true);
+			RzFlagItem *flag = rz_flag_get(core->flags, buf + 1);
+			if (flag) {
+				rz_core_seek(core, flag->offset, true);
```
ret2libc, line 122: what is the purpose of this whole piece of code?

kazarmy: $-flag
(from `$??`)

librz/core/cmd/cmd_alias.c
```diff
@@ -117,9 +117,9 @@ RZ_IPI RzCmdStatus rz_alias_handler(RzCore *core, int argc, const char **argv) {
 				rz_core_cmd0(core, v);
 			}
 		} else {
-			ut64 at = rz_num_get(core->num, buf + 1);
```
wargio, line 120: doesn't removing this forbid doing `flag.something+0x123` ?

kazarmy: no, that should still work fine

flag+off

it does forbid you from doing `$0x123` to mean `s 0x123` but the latter definitely seems better

wargio (approved):

librz/core/cmd/cmd_alias.c
kazarmy: I don't know whether there's a use case for allowing `$foo+1` to mean `s foo+1`, unless one wants everything to be $-prefixed for the bling ... also not sure whether it's wise for `$` to be overloaded as shown below:

$-flag
$-alias

but that's not my concern right now.

wargio: no, i think is ok.

PR #4379, opened 2024-03-22
Lift MSP430 machine language to RzIL

moste00:

**Detailed description**

This PR is the first step towards lifting of the MSP430 instruction set to the RzIL representation. MSP430 is a quite simple architecture with comprehensive references here (http://elec327.github.io/assets/documents/slau144j_userguide.pdf), here (https://www.ti.com/sc/docs/products/micro/msp430/userguid/ag_b.pdf) and (https://www.ti.com/sc/docs/products/micro/msp430/userguid/ag_04.pdf). 

In this PR, only lifting of the following 3 single-operand instruction is implemented. This is deliberate to keep the diff small. The rest of the lifting could be either new commits on this same branch and PR or future PRs.

Instructions lifted:
 
- RRC (Rotate Right through Carry): takes a single operand which is also the destination, shifts it right by 1 position. The MSB becomes the carry flag, and the carry flag becomes the discarded LSB (from before the shift). Other flags are updated as specified in the references. Instruction has word-sized and byte-sized variants.

- SWPB (Swap Bytes): takes a single operand which is also the destination, and exchanges the 2 bytes in it. Doesn't update the flags or any other CPU state. Always a word-sized operation.

- SXT (Sign eXTension): takes a single operand which is also the destination, and replicates the MSB of the lower byte of the operand to the entire upper byte (e.g. if the lower byte is less than 128, the upper byte is 255, else the upper byte is 0). Always a word-sized operation that reads the lower byte and writes the upper byte of the operand.

Each instruction in the MSP430 has seven addressing modes, only one of them is implemented in this PR, namely the Register Addressing Mode, where the operand(s) are a register index. Other indexing modes including memory access and immediates should be straightforward to add later. 

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/librz/arch/isa/msp430/msp430_disas.c
+++ b/librz/arch/isa/msp430/msp430_disas.c
@@ -7,26 +7,11 @@
 #include <rz_types.h>
 #include <rz_util.h>
 
-#include "msp430_disas.h"
+#include <msp430/msp430_disas.h>
 
-static const char *msp430_register_names[] = {
-	"pc",
-	"sp",
-	"sr",
-	"cg",
-	"r4",
-	"r5",
-	"r6",
-	"r7",
-	"r8",
-	"r9",
-	"r10",
-	"r11",
-	"r12",
-	"r13",
-	"r14",
-	"r15",
-};
+#include <msp430/msp430_register_names.h>
+
+#define WRITE_INSTR_STR(cmd, str) snprintf(cmd->instr, sizeof(cmd->instr), "%s", str)
 
 static const char *two_op_instrs[] = {
 	[MSP430_MOV] = "mov",
@@ -88,14 +73,14 @@ static int get_dst(ut16 instr) {
 	return instr & 0xF;
 }
 
-static void remove_first_operand(struct msp430_cmd *cmd) {
+static void remove_first_operand(Msp430Cmd *cmd) {
 	if (strchr(cmd->operands, ',')) {
 		memmove(cmd->operands, strchr(cmd->operands, ',') + 2,
 			strlen(strchr(cmd->operands, ',') + 2) + 1);
 	}
 }
 
-static void remove_second_operand(struct msp430_cmd *cmd) {
+static void remove_second_operand(Msp430Cmd *cmd) {
 	if (strchr(cmd->operands, ',')) {
 		{
 			*strchr(cmd->operands, ',') = '\0';
@@ -103,9 +88,7 @@ static void remove_second_operand(struct msp430_cmd *cmd) {
 	}
 }
 
-/* TODO: This code is messy, needs to be refactored. */
-static int decode_emulation(ut16 instr, struct msp430_cmd *cmd) {
-	int ret = -1;
+static void decode_emulation(ut16 instr, Msp430Cmd *cmd) {
 	ut8 as, ad, src, dst, bw, opcode;
 
 	as = get_as(instr);
@@ -116,87 +99,129 @@ static int decode_emulation(ut16 instr, struct msp430_cmd *cmd) {
 	opcode = get_twoop_opcode(instr);
 
 	if (opcode == MSP430_ADDC && as == 0 && src == MSP430_R3) {
-		snprintf(cmd->instr, sizeof(cmd->instr), "%s", bw ? "adc.b" : "adc");
+		WRITE_INSTR_STR(cmd, bw ? "adc.b" : "adc");
 		snprintf(cmd->operands, sizeof(cmd->operands), "%s", msp430_register_names[dst]);
+		cmd->type = MSP430_EMULATE;
+		cmd->opcode = MSP430_ADC;
 	} else if (opcode == MSP430_MOV && as == 0 && src == MSP430_R3) {
+		cmd->type = MSP430_EMULATE;
+
 		if (ad == 0 && dst == MSP430_R3) {
 			snprintf(cmd->instr, sizeof(cmd->instr), "nop");
 			cmd->operands[0] = '\0';
+			cmd->opcode = MSP430_NOP;
 		} else {
-			snprintf(cmd->instr, sizeof(cmd->instr), "%s", bw ? "clr.b" : "clr");
+			WRITE_INSTR_STR(cmd, bw ? "clr.b" : "clr");
 			remove_first_operand(cmd);
+			cmd->opcode = MSP430_CLR;
 		}
 	} else if (opcode == MSP430_MOV && as == 3 && src == MSP430_SP) {
+		cmd->type = MSP430_EMULATE;
+
 		if (dst == MSP430_PC) {
 			snprintf(cmd->instr, sizeof(cmd->instr), "ret");
-			cmd->type = MSP430_ONEOP;
-			cmd->opcode = MSP430_RETI;
 			cmd->operands[0] = '\0';
+			cmd->opcode = MSP430_RET;
 		} else {
-			snprintf(cmd->instr, sizeof(cmd->instr), "%s", bw ? "pop.b" : "pop");
+			WRITE_INSTR_STR(cmd, bw ? "pop.b" : "pop");
 			remove_first_operand(cmd);
+			cmd->opcode = MSP430_POP;
 		}
 	} else if (opcode == MSP430_MOV && ad == 0 && dst == MSP430_PC) {
-		snprintf(cmd->instr, sizeof(cmd->instr), "%s", "br");
+		WRITE_INSTR_STR(cmd, "br");
 		remove_second_operand(cmd);
+		cmd->type = MSP430_EMULATE;
+		cmd->opcode = MSP430_BR;
 	} else if (opcode == MSP430_BIC && as == 2 && src == MSP430_SR && dst == MSP430_SR && ad == 0) {
-		snprintf(cmd->instr, sizeof(cmd->instr), "%s", "clrn");
+		WRITE_INSTR_STR(cmd, "clrn");
 		cmd->operands[0] = '\0';
+		cmd->type = MSP430_EMULATE;
+		cmd->opcode = MSP430_CLRN;
 	} else if (opcode == MSP430_BIC && as == 2 && src == MSP430_R3 && dst == MSP430_SR && ad == 0) {
-		snprintf(cmd->instr, sizeof(cmd->instr), "%s", "clrz");
+		WRITE_INSTR_STR(cmd, "clrz");
 		cmd->operands[0] = '\0';
+		cmd->type = MSP430_EMULATE;
+		cmd->opcode = MSP430_CLRZ;
 	} else if (opcode == MSP430_BIC && as == 3 && src == MSP430_SR && dst == MSP430_SR && ad == 0) {
-		snprintf(cmd->instr, sizeof(cmd->instr), "%s", "dint");
+		WRITE_INSTR_STR(cmd, "dint");
 		cmd->operands[0] = '\0';
+		cmd->type = MSP430_EMULATE;
+		cmd->opcode = MSP430_DINT;
 	} else if (opcode == MSP430_BIS && as == 3 && src == MSP430_SR && dst == MSP430_SR && ad == 0) {
-		snprintf(cmd->instr, sizeof(cmd->instr), "%s", "eint");
+		WRITE_INSTR_STR(cmd, "eint");
 		cmd->operands[0] = '\0';
+		cmd->type = MSP430_EMULATE;
+		cmd->opcode = MSP430_EINT;
 	} else if (opcode == MSP430_DADD && as == 0 && src == MSP430_R3) {
-		snprintf(cmd->instr, sizeof(cmd->instr), "%s", bw ? "dadc.b" : "dadc");
+		WRITE_INSTR_STR(cmd, bw ? "dadc.b" : "dadc");
 		remove_first_operand(cmd);
+		cmd->type = MSP430_EMULATE;
+		cmd->opcode = MSP430_DADC;
 	} else if (opcode == MSP430_SUB && as == 1 && src == MSP430_R3) {
-		snprintf(cmd->instr, sizeof(cmd->instr), "%s", bw ? "dec.b" : "dec");
+		WRITE_INSTR_STR(cmd, bw ? "dec.b" : "dec");
 		remove_first_operand(cmd);
+		cmd->type = MSP430_EMULATE;
+		cmd->opcode = MSP430_DEC;
 	} else if (opcode == MSP430_SUB && as == 2 && src == MSP430_R3) {
-		snprintf(cmd->instr, sizeof(cmd->instr), "%s", bw ? "decd.b" : "decd");
+		WRITE_INSTR_STR(cmd, bw ? "decd.b" : "decd");
 		remove_first_operand(cmd);
+		cmd->type = MSP430_EMULATE;
+		cmd->opcode = MSP430_DECD;
 	} else if (opcode == MSP430_ADD && as == 1 && src == MSP430_R3) {
-		snprintf(cmd->instr, sizeof(cmd->instr), "%s", bw ? "inc.b" : "inc");
+		WRITE_INSTR_STR(cmd, bw ? "inc.b" : "inc");
 		remove_first_operand(cmd);
+		cmd->type = MSP430_EMULATE;
+		cmd->opcode = MSP430_INC;
 	} else if (opcode == MSP430_ADD && as == 2 && src == MSP430_R3) {
-		snprintf(cmd->instr, sizeof(cmd->instr), "%s", bw ? "incd.b" : "incd");
+		WRITE_INSTR_STR(cmd, bw ? "incd.b" : "incd");
 		remove_first_operand(cmd);
+		cmd->type = MSP430_EMULATE;
+		cmd->opcode = MSP430_INCD;
 	} else if (opcode == MSP430_XOR && as == 3 && src == MSP430_R3) {
-		snprintf(cmd->instr, sizeof(cmd->instr), "%s", bw ? "inv.b" : "inv");
+		WRITE_INSTR_STR(cmd, bw ? "inv.b" : "inv");
 		remove_first_operand(cmd);
+		cmd->type = MSP430_EMULATE;
+		cmd->opcode = MSP430_INV;
 	} else if (opcode == MSP430_ADD && src == dst) {
-		snprintf(cmd->instr, sizeof(cmd->instr), "%s", bw ? "rla.b" : "rla");
+		WRITE_INSTR_STR(cmd, bw ? "rla.b" : "rla");
 		remove_second_operand(cmd);
+		cmd->type = MSP430_EMULATE;
+		cmd->opcode = MSP430_RLA;
 	} else if (opcode == MSP430_ADDC && src == dst) {
-		snprintf(cmd->instr, sizeof(cmd->instr), "%s", bw ? "rlc.b" : "rlc");
+		WRITE_INSTR_STR(cmd, bw ? "rlc.b" : "rlc");
 		remove_second_operand(cmd);
+		cmd->type = MSP430_EMULATE;
+		cmd->opcode = MSP430_RLC;
 	} else if (opcode == MSP430_SUBC && as == 0 && src == MSP430_R3) {
-		snprintf(cmd->instr, sizeof(cmd->instr), "%s", bw ? "sbc.b" : "sbc");
+		WRITE_INSTR_STR(cmd, bw ? "sbc.b" : "sbc");
 		remove_first_operand(cmd);
+		cmd->type = MSP430_EMULATE;
+		cmd->opcode = MSP430_SBC;
 	} else if (opcode == MSP430_BIS && as == 1 && src == MSP430_R3 && dst == MSP430_SR && ad == 0) {
 		snprintf(cmd->instr, sizeof(cmd->instr), "setc");
 		cmd->operands[0] = '\0';
+		cmd->type = MSP430_EMULATE;
+		cmd->opcode = MSP430_SETC;
 	} else if (opcode == MSP430_BIS && as == 2 && src == MSP430_SR && dst == MSP430_SR && ad == 0) {
 		snprintf(cmd->instr, sizeof(cmd->instr), "setn");
 		cmd->operands[0] = '\0';
+		cmd->type = MSP430_EMULATE;
+		cmd->opcode = MSP430_SETN;
 	} else if (opcode == MSP430_BIS && as == 2 && src == MSP430_R3 && dst == MSP430_SR && ad == 0) {
 		snprintf(cmd->instr, sizeof(cmd->instr), "setz");
 		cmd->operands[0] = '\0';
+		cmd->type = MSP430_EMULATE;
+		cmd->opcode = MSP430_SETZ;
 	} else if (opcode == MSP430_CMP && as == 0 && src == MSP430_R3) {
-		snprintf(cmd->instr, sizeof(cmd->instr), "%s", bw ? "tst.b" : "tst");
+		WRITE_INSTR_STR(cmd, bw ? "tst.b" : "tst");
 		remove_first_operand(cmd);
+		cmd->type = MSP430_EMULATE;
+		cmd->opcode = MSP430_TST;
 	}
-
-	return ret;
 }
 
 /* return #byte of instruction */
-static int decode_addressing_mode(ut16 instr, ut16 op1, ut16 op2, struct msp430_cmd *cmd) {
+static int decode_addressing_mode(ut16 instr, ut16 op1, ut16 op2, Msp430Cmd *cmd) {
 	int ret = 0, srcOperInCodeWord = 0;
 	ut8 as, ad, src, dst;
 	ut16 op;
@@ -215,9 +240,13 @@ static int decode_addressing_mode(ut16 instr, ut16 op1, ut16 op2, struct msp430_
 		switch (src) {
 		case MSP430_R3: /* CG2 */
 			snprintf(cmd->operands, sizeof(cmd->operands), "#0");
+			cmd->src_mode = MSP430_IMM;
+			cmd->src = 0;
 			break;
 		default: /* register mode */
 			snprintf(cmd->operands, sizeof(cmd->operands), "%s", msp430_register_names[src]);
+			cmd->src_mode = MSP430_REG;
+			cmd->src = src;
 		}
 		ret = 2;
 		break;
@@ -226,18 +255,26 @@ static int decode_addressing_mode(ut16 instr, ut16 op1, ut16 op2, struct msp430_
 		switch (src) {
 		case MSP430_PC: /* symbolic mode */
 			snprintf(cmd->operands, sizeof(cmd->operands), "0x%04x", op1);
+			cmd->src_mode = MSP430_SYM;
+			cmd->src = op1;
 			srcOperInCodeWord = 1;
 			break;
 		case MSP430_R3: /* CG2 */
 			snprintf(cmd->operands, sizeof(cmd->operands), "%s", "#1");
+			cmd->src_mode = MSP430_IMM;
+			cmd->src = 1;
 			ret = 2;
 			break;
 		case MSP430_SR: /* absolute mode */
 			snprintf(cmd->operands, sizeof(cmd->operands), "&0x%04x", op1);
+			cmd->src_mode = MSP430_ABS;
+			cmd->src = op1;
 			srcOperInCodeWord = 1;
 			break;
 		default: /* indexed mode */
 			snprintf(cmd->operands, sizeof(cmd->operands), "0x%x(%s)", op1, msp430_register_names[src]);
+			cmd->src_mode = MSP430_INDX;
+			cmd->src = (op1 << 8) | src;
 			srcOperInCodeWord = 1;
 		}
 		break;
@@ -245,12 +282,18 @@ static int decode_addressing_mode(ut16 instr, ut16 op1, ut16 op2, struct msp430_
 		switch (src) {
 		case MSP430_SR: /* CG1 */
 			snprintf(cmd->operands, sizeof(cmd->operands), "#4");
+			cmd->src_mode = MSP430_IMM;
+			cmd->src = 4;
 			break;
 		case MSP430_R3: /* CG2 */
 			snprintf(cmd->operands, sizeof(cmd->operands), "#2");
+			cmd->src_mode = MSP430_IMM;
+			cmd->src = 2;
 			break;
 		default: /* indirect register mode */
 			snprintf(cmd->operands, sizeof(cmd->operands), "@%s", msp430_register_names[src]);
+			cmd->src_mode = MSP430_IND_REG;
+			cmd->src = src;
 		}
 		ret = 2;
 		break;
@@ -259,25 +302,36 @@ static int decode_addressing_mode(ut16 instr, ut16 op1, ut16 op2, struct msp430_
 		switch (src) {
 		case MSP430_SR: /* CG1 */
 			snprintf(cmd->operands, sizeof(cmd->operands), "#8");
+			cmd->src_mode = MSP430_IMM;
+			cmd->src = 8;
 			break;
 		case MSP430_R3: /* CG2 */
 			snprintf(cmd->operands, sizeof(cmd->operands), "#-1");
+			cmd->src_mode = MSP430_IMM;
+			cmd->src = -1;
 			break;
 		case MSP430_PC: /* immediate mode */
 			snprintf(cmd->operands, sizeof(cmd->operands), "#0x%04x", op1);
+			cmd->src_mode = MSP430_IMM;
+			cmd->src = op1;
 			srcOperInCodeWord = 1;
 			ret = 4;
 			break;
 		default: /* indirect autoincrement mode */
 			snprintf(cmd->operands, sizeof(cmd->operands), "@%s+", msp430_register_names[src]);
+			cmd->src_mode = MSP430_IND_AUTOINC;
+			cmd->src = src;
 		}
 		break;
+	default: RZ_LOG_WARN("UNREACHABLE");
 	}
 
 	/* addressing mode of destination operand */
 	switch (ad) {
 	case 0: /* register mode */
 		snprintf(dstbuf, sizeof(dstbuf), ", %s", msp430_register_names[dst]);
+		cmd->dst_mode = MSP430_REG;
+		cmd->dst = dst;
 		break;
 	case 1:
 		/* check addr. mode of source operand */
@@ -291,14 +345,21 @@ static int decode_addressing_mode(ut16 instr, ut16 op1, ut16 op2, struct msp430_
 		switch (get_dst(instr)) {
 		case MSP430_PC: /* symbolic mode */
 			snprintf(dstbuf, sizeof(dstbuf), ", 0x%04x", op);
+			cmd->dst_mode = MSP430_SYM;
+			cmd->dst = op;
 			break;
 		case MSP430_SR: /* absolute mode */
 			snprintf(dstbuf, sizeof(dstbuf), ", &0x%04x", op);
+			cmd->dst_mode = MSP430_ABS;
+			cmd->dst = op;
 			break;
 		default: /* indexed mode */
 			snprintf(dstbuf, sizeof(dstbuf), ", 0x%x(%s)", op, msp430_register_names[dst]);
+			cmd->dst_mode = MSP430_INDX;
+			cmd->dst = (op << 8) | dst;
 		}
 		break;
+	default: RZ_LOG_WARN("UNREACHABLE");
 	}
 
 	strncat(cmd->operands, dstbuf, sizeof(cmd->operands) - 1 - strlen(cmd->operands));
@@ -306,11 +367,12 @@ static int decode_addressing_mode(ut16 instr, ut16 op1, ut16 op2, struct msp430_
 	return ret;
 }
 
-static int decode_twoop_opcode(ut16 instr, ut16 op1, ut16 op2, struct msp430_cmd *cmd) {
+static int decode_twoop_opcode(ut16 instr, ut16 op1, ut16 op2, Msp430Cmd *cmd) {
 	ut8 opcode = get_twoop_opcode(instr);
 
-	snprintf(cmd->instr, sizeof(cmd->instr), "%s", two_op_instrs[opcode]);
-	if (get_bw(instr)) {
+	WRITE_INSTR_STR(cmd, two_op_instrs[opcode]);
+	cmd->is_byte = get_bw(instr);
+	if (cmd->is_byte) {
 		strncat(cmd->instr, ".b", sizeof(cmd->instr) - 1 - strlen(cmd->instr));
 	}
 
@@ -326,10 +388,10 @@ static ut8 get_jmp_cond(ut16 instr) {
 	return (instr >> 10) & 7;
 }
 
-static void decode_jmp(ut16 instr, struct msp430_cmd *cmd) {
+static void decode_jmp(ut16 instr, Msp430Cmd *cmd) {
 	ut16 addr;
 
-	snprintf(cmd->instr, sizeof(cmd->instr), "%s", jmp_instrs[get_jmp_cond(instr)]);
+	WRITE_INSTR_STR(cmd, jmp_instrs[get_jmp_cond(instr)]);
 
 	addr = instr & 0x3FF;
 
@@ -346,7 +408,7 @@ static int get_oneop_opcode(ut16 instr) {
 	return (instr >> 7) & 0x7;
 }
 
-static int decode_oneop_opcode(ut16 instr, ut16 op, struct msp430_cmd *cmd) {
+static int decode_oneop_opcode(ut16 instr, ut16 op, Msp430Cmd *cmd) {
 	int ret = 2;
 	ut8 as, opcode;
 
@@ -354,7 +416,7 @@ static int decode_oneop_opcode(ut16 instr, ut16 op, struct msp430_cmd *cmd) {
 
 	as = get_as(instr);
 
-	snprintf(cmd->instr, sizeof(cmd->instr), "%s", one_op_instrs[opcode]);
+	WRITE_INSTR_STR(cmd, one_op_instrs[opcode]);
 
 	cmd->opcode = get_oneop_opcode(instr);
 
@@ -370,10 +432,15 @@ static int decode_oneop_opcode(ut16 instr, ut16 op, struct msp430_cmd *cmd) {
 			switch (get_dst(instr)) {
 			case MSP430_R3:
 				snprintf(cmd->operands, sizeof(cmd->operands), "#0");
+				cmd->dst = 0;
+				cmd->dst_mode = MSP430_IMM;
 				break;
 			default:
 				snprintf(cmd->operands, sizeof(cmd->operands),
 					"%s", msp430_register_names[get_dst(instr)]);
+				cmd->dst = get_dst(instr);
+				cmd->dst_mode = MSP430_REG;
+				break;
 			}
 			ret = 2;
 			break;
@@ -383,18 +450,26 @@ static int decode_oneop_opcode(ut16 instr, ut16 op, struct msp430_cmd *cmd) {
 			switch (get_dst(instr)) {
 			case MSP430_R3:
 				snprintf(cmd->operands, sizeof(cmd->operands), "#1");
+				cmd->dst = 1;
+				cmd->dst_mode = MSP430_IMM;
 				/* this is an unusual encoding in that there's no index word */
 				ret = 2;
 				break;
 			case MSP430_PC:
 				snprintf(cmd->operands, sizeof(cmd->operands), "0x%04x", op);
+				cmd->dst = op;
+				cmd->dst_mode = MSP430_SYM;
 				break;
 			case MSP430_SR:
 				snprintf(cmd->operands, sizeof(cmd->operands), "&0x%04x", op);
+				cmd->dst = op;
+				cmd->dst_mode = MSP430_ABS;
 				break;
 			default:
 				snprintf(cmd->operands, sizeof(cmd->operands),
 					"0x%x(%s)", op, msp430_register_names[get_dst(instr)]);
+				cmd->dst = (op << 8) | get_dst(instr);
+				cmd->dst_mode = MSP430_INDX;
 			}
 
 			break;
@@ -402,19 +477,27 @@ static int decode_oneop_opcode(ut16 instr, ut16 op, struct msp430_cmd *cmd) {
 			switch (get_dst(instr)) {
 			case MSP430_SR:
 				snprintf(cmd->operands, sizeof(cmd->operands), "#4");
+				cmd->dst = 4;
+				cmd->dst_mode = MSP430_IMM;
 				break;
 			case MSP430_R3:
 				snprintf(cmd->operands, sizeof(cmd->operands), "#2");
+				cmd->dst = 2;
+				cmd->dst_mode = MSP430_IMM;
 				break;
 			default:
 				snprintf(cmd->operands, sizeof(cmd->operands),
 					"@%s", msp430_register_names[get_dst(instr)]);
+				cmd->dst = get_dst(instr);
+				cmd->dst_mode = MSP430_IND_REG;
 			}
 
 			ret = 2;
 			break;
 		case 3:
 			snprintf(cmd->operands, sizeof(cmd->operands), "#0x%04x", op);
+			cmd->dst = op;
+			cmd->dst_mode = MSP430_IMM;
 			ret = 4;
 			break;
 		default:
@@ -427,6 +510,7 @@ static int decode_oneop_opcode(ut16 instr, ut16 op, struct msp430_cmd *cmd) {
 	}
 
 	cmd->type = MSP430_ONEOP;
+	cmd->is_byte = get_bw(instr);
 
 	return ret;
 }
@@ -438,7 +522,7 @@ enum {
 	MSP430_TWOOP_OPCODE_JUMP3,
 };
 
-int msp430_decode_command(const ut8 *in, int len, struct msp430_cmd *cmd) {
+int msp430_decode_command(const ut8 *in, int len, Msp430Cmd *cmd) {
 	int ret = -1;
 	ut16 operand1 = 0, operand2 = 0;
 	if (len < 2) {
@@ -483,7 +567,7 @@ int msp430_decode_command(const ut8 *in, int len, struct msp430_cmd *cmd) {
 	/* if ret < 0, it's an invalid opcode.Say so and return 2 since
 	 * all MSP430 opcodes are of 16 bits,valid or invalid */
 	if (ret < 0) {
-		cmd->type = MSP430_INV;
+		cmd->type = MSP430_INVALID;
 		snprintf(cmd->instr, sizeof(cmd->instr), "invalid");
 		cmd->operands[0] = '\0';
 		ret = 2;
--- a/librz/arch/isa/msp430/msp430_disas.c
+++ b/librz/arch/isa/msp430/msp430_disas.c
@@ -359,7 +359,7 @@ static int decode_addressing_mode(ut16 instr, ut16 op1, ut16 op2, Msp430Cmd *cmd
 			cmd->dst = (op << 8) | dst;
 		}
 		break;
-	default: RZ_LOG_WARN("UNREACHABLE");
+	default: rz_warn_if_reached();
 	}
 
 	strncat(cmd->operands, dstbuf, sizeof(cmd->operands) - 1 - strlen(cmd->operands));
--- a/librz/arch/isa/msp430/msp430_disas.h
+++ b/librz/arch/isa/msp430/msp430_disas.h
@@ -7,7 +7,7 @@
 #ifndef MSP430_DISAS_H
 #define MSP430_DISAS_H
 
-enum msp430_oneop_opcodes {
+typedef enum msp430_oneop_opcodes {
 	MSP430_RRC,
 	MSP430_SWPB,
 	MSP430_RRA,
@@ -16,9 +16,9 @@ enum msp430_oneop_opcodes {
 	MSP430_CALL,
 	MSP430_RETI,
 	MSP430_UNUSED,
-};
+} Msp430OneopOpcodes;
 
-enum msp430_jumps {
+typedef enum msp430_jumps {
 	MSP430_JNE,
 	MSP430_JEQ,
 	MSP430_JNC,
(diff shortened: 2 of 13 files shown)


librz/arch/isa/msp430/msp430_il.c
```diff
@@ -0,0 +1,199 @@
...
+}
+
+RzILOpBitVector *update_sr_nz_flags(const RzILOpBitVector *new_value, const RzILOpBitVector *old_sr_value) {
+    // the general idea is that we zero out the N and Z bits in the old SR value 
+    // (by ANDing with a mask of all 1s except in those positions)
+    // and then we OR that resulting value with the new bits in the relevant positions and all 0s everywhere else
+
+    // index of the N flag in the SR register: 2
+    RzILOpBool *n_flag_value = MSB(new_value);
+    RzILOpBitVector *n_or_mask = SHIFTL(IL_FALSE, UNSIGNED(16, n_flag_value), UN8(2));
```
DMaroo: You can use `SHIFTL0` here.

librz/arch/isa/msp430/msp430_il.c
```diff
@@ -0,0 +1,199 @@
...
+
+RzILOpBitVector *update_sr_c_flag(const RzILOpBool *new_carry, const RzILOpBitVector *old_sr_value) {
+    return MSP430_SETR(MSP430_SR, OR(
+        AND(old_sr_value, UN16(1)), // zero out the 0th position (i.e. the carry bit)
+        UNSIGNED(16, new_carry)    // OR with the new carry 
+    ));
+}
+
+RzILOpEffect *rz_msp430_lift_single_operand_instr(RZ_NONNULL RzAnalysis *analysis, const Msp430Instruction *op) {
+	switch (op->iopcode) {
```
DMaroo: It would be neater if you have a separate IL lifter function for each of the instructions. You can have an instruction-to-function map and use it as a dispatch table. For example, have a look at:

https://github.com/rizinorg/rizin/blob/c52750ccfeb493b59929300276a57a901b1f4765/librz/arch/isa/sh/sh_il.c#L1665

DMaroo (commented):
Excited about this! Feel free to reach out if you need any help.

XVilka (changes requested):
1. You need to update Meson files to include them to be built
2. It is necessary to add per-instruction RzIL tests in `test/db/asm/msp430`

librz/arch/isa/msp430/msp430_il.c
```diff
@@ -0,0 +1,199 @@
+// SPDX-FileCopyrightText: 2023 Mostafa Mahmoud <ubermenchun@gmail.com>
+// SPDX-License-Identifier: LPGL-3.0-only
+
+#include "msp430_il.h"
```
Rot127: Please include them with `<msp430/msp430_il.h>`

librz/arch/isa/msp430/msp430_il.c
```diff
@@ -0,0 +1,199 @@
...
+    RzILOpBool *z_flag_value = IS_ZERO(new_value);
+    RzILOpBitVector *z_or_mask = SHIFTL(IL_FALSE, UNSIGNED(16, z_flag_value), UN8(1));
+    
+    RzILOpBitVector *and_mask = UN16(~(1 << 2 | 1 << 1));
+    RzILOpBitVector *or_mask = OR(n_or_mask, z_or_mask);
+    
+    return OR(AND(old_sr_value, and_mask), or_mask);
+}
+
+RzILOpBitVector *update_sr_v_flag_rcc(const RzILOpBitVector *old_value, const RzILOpBool *old_carry, const RzILOpBitVector *old_sr_value) {
```
Rot127:
```suggestion
RZ_OWN RzILOpBitVector *update_sr_v_flag_rcc(const RzILOpBitVector *old_value, const RzILOpBool *old_carry, const RzILOpBitVector *old_sr_value) {
```

It is only a draft. But don't forget to add those `RZ_OWN`, `RZ_BORROW` annotations later. They help quite a lot to force onceself to think about which component is responsible for what.

librz/arch/isa/msp430/msp430_il.c
```diff
@@ -0,0 +1,199 @@
...
+
+RzILOpEffect *rz_msp430_lift_double_operand_instr(RZ_NONNULL RzAnalysis *analysis, const Msp430Instruction *op) {
+	return NOP(); // TODO
+}
+
+RzILOpEffect *rz_msp430_lift_jump_instr(RZ_NONNULL RzAnalysis *analysis, const Msp430Instruction *op) {
+	return NOP(); // TODO
+}
+
+RZ_IPI RzILOpEffect *rz_msp430_lift_instr(RZ_NONNULL RzAnalysis *analysis, const Msp430Instruction *op) {
```
Rot127:
```suggestion
RZ_IPI RzILOpEffect *rz_msp430_lift_instr(RZ_NONNULL RzAnalysis *analysis, const Msp430Instruction *op) {
    rz_return_val_if_fail(analysis && op, NULL);
```

These kind of `NULL` checks at the beginning will help you a lot later to locate errors.
It is advisable to add them in every function.

librz/arch/isa/msp430/msp430_il.c
```diff
@@ -0,0 +1,199 @@
...
+
+RzILOpEffect *rz_msp430_lift_single_operand_instr(RZ_NONNULL RzAnalysis *analysis, const Msp430Instruction *op) {
+	switch (op->iopcode) {
+    // rotation is just a shift with filling
+    case MSP430_RRC: {
+        // 1- get the carry (to use later as the filling for the MSB of the operand)
+        RzILOpBitVector *sr = MSP430_GETR(MSP430_SR);
+        RzILOpBool *carry = LSB(sr);
+        // 2- get LSB of the operand (register, memory location, ...)
+        RzILOpBitVector *operand = get_destination(op);
```
Rot127: This `operand` variable gets passed to three different functions later. Mind that once a `RzILOpPure` was used in a single operation like `ADD`, you have to use `DUP(operand)` if you want to use it again.
When a rzil operation is freed, it frees all its operands it consumes recursively. So if the same rzil operand (like this one) is used at multiple places without `DUP()`, it will be double freed.

librz/arch/isa/msp430/msp430_il.c
```diff
@@ -0,0 +1,199 @@
...
+        RzILOpBool *lsb = LSB(operand);
+        // 3- Perform the Rotate Right through Carry operation: 
+        //      a- Shift the operand by 1 to the right and fill with carry
+        RzILOpBitVector *result = SHIFTR(carry, operand, U8(1));
+        
+        // ...  b- and set the operand to the value of the previous computation 
+        RzILOpEffect *set_operand = set_destination(op, operand, result);
```
Rot127:
```suggestion
        RzILOpEffect *set_operand = set_destination(op, DUP(operand), result);
```
See above

librz/arch/isa/msp430/msp430_il.c
```diff
@@ -0,0 +1,199 @@
...
+        // ...  c- and finally set the flags: NZ as usual, v espeically for RCC, and the carry flag from the discarded LSB
+        RzILOpBitVector *sr_1 = update_sr_nz_flags(result, sr);
+        RzILOpBitVector *sr_2 = update_sr_v_flag_rcc(operand, carry, sr_1);
```
Rot127:
```suggestion
        RzILOpBitVector *sr_2 = update_sr_v_flag_rcc(DUP(operand), carry, sr_1);
```

librz/arch/isa/msp430/msp430_il.c
```diff
@@ -0,0 +1,199 @@
...
+    }
+    // TODO add the rest of instructions
+    default:
+        rz_warn_if_reached();
+        return rz_msp430_dummy();
+    }
+}
+
+RzILOpEffect *rz_msp430_lift_double_operand_instr(RZ_NONNULL RzAnalysis *analysis, const Msp430Instruction *op) {
+	return NOP(); // TODO
```
Rot127:
```suggestion
	NOT_IMPLEMENTED;
```
The macro returns NULL. Architecturally it is better IMHO, to handle the `NOT_IMPLEMENTED` cases at one place.

Rot127 (commented):
Very nice! Would be awesome if you finish it!

It is advisable to think about how to group several instructions. E.g. for ARM, PPC etc. we grouped them by arithmetic, shift, branch etc instructions. Mostly because the operands of those groups are almost always the same. So it is easier to initialize the operands at the start of the function and than use them to build the operations.
Check the ISA, if they group instructions into classes or something. Usually the grouping from the ISA makes the most sense.

librz/arch/isa/msp430/msp430_il.c
DMaroo: Adding onto this, in such a case, I would suggest using local variables. Check out the `SETL` and the `VARL` macro. `DUP` literally copies the entire expression, leading to bigger IL. Moreover, it's better if a local variable is used for the same expression rather than duplicating the entire expression, since the former will allow for better analysis.

Rot127: Actually in this case it is even better to use `LET("operand", operand, <expr>)` IMHO.

moste00:
> Very nice! Would be awesome if you finish it!

Thanks a lot! I do intend to finish it, fingers crossed. The architecture is a bit obscure and the only available emulator (https://msp430.online/) requires a compilation toolchain because it expects an ELF or something like that, on the one hand this is good because adding support for obscure architectures is precisely why ESIL and RzIL were introduced in the first place, but on the other hand that means it's very easy to come up with bugs and - even worse - subtle deviations from the behavior of the CPU due to vague English descriptions. Add to that I'm not very familiar with RzIL API (e.g. the DUP bug above never crossed my mind), and that's why I'm taking it very slow and careful here.  

> It is advisable to think about how to group several instructions. E.g. for ARM, PPC etc. we grouped them by arithmetic, shift, branch etc instructions. 

It appears that MSP430 instructions are very orthogonal, there is almost no difference at all between instructions of the same arity, the only 2 differences is whether instructions update the flags and whether they're emulated (syntactic sugar for specific uses of other instructions) or not.

I'm generally not familiar enough yet with the architecture to split by myself, so I'm just grouping by arity and whether the instruction is a jump instruction, this is the same approach taken by the disassembler, so I'm following it now but keeping my mind open if there are any patterns that hint if there is a better grouping. 

Unlike the disassembler, which naturally doesn't mention emulated instructions (since those instructions don't exist in the byte stream input to the disassembler), it might make sense in my case to add the emulated instruction because it's more efficient. For example it's inefficient to implement a SETC (set carry flag) using the architecture's emulation, which uses a more complex instruction BIS with 2 operands and more general logic. SETC is much simpler. But I'm implementing the bare basics and worrying about efficiency and auxiliaries later.

librz/arch/isa/msp430/msp430_il.c
moste00: Do you mean the copyright notices? I did write them in the `<msp430/msp430_il.h>`. Or did you mean that I should remove them from the `.c` and rely on the `#include` to copy them by itself?

XVilka:
Don't forget to rebase when you do your next change.

librz/arch/isa/msp430/msp430_il.c
moste00: Good Catch! 

Well, I do agree that `DUP`ing is inefficient, but using `LET(...)` is a bit hard because it expects a pure expression as both the value and the body, this turned out to not get along well with what the function is trying to do, which is constructing and returning 2 effects. I ended up using a bunch of `SETL(...)`s in a `SEQ(...)` instead, that was straightforward enough.

Thanks a lot for the free bug-spotting.

librz/arch/isa/msp430/msp430_disas.h
```diff
@@ -7,6 +7,25 @@
 #ifndef MSP430_DISAS_H
 #define MSP430_DISAS_H
 
+static const char *msp430_register_names[] = {
```
XVilka:
```
 gcc -Ilibrz/arch/librz_arch.so.0.8.0.p -I. -I.. -Ilibrz -I../librz -Ilibrz/include -I../librz/include -I../librz/arch/isa -I../librz/arch/isa_gnu -Ilibrz/arch -I../librz/arch -Ilibrz/util/sdb/src -I../librz/util/sdb/src -I../librz/bin/format -I../librz/type/parser -I../subprojects/capstone-next/include -I../subprojects/capstone-next/include/capstone -Isubprojects/rzspp -I../subprojects/rzspp -fdiagnostics-color=always -D_FILE_OFFSET_BITS=64 -Wall -Winvalid-pch -Werror -O3 -Wimplicit-fallthrough=3 -DRZ_PLUGIN_INCORE=1 -DSUPPORTS_PCRE2_JIT -D_GNU_SOURCE --std=gnu99 -Werror=sizeof-pointer-memaccess -fvisibility=hidden -DRZ_ASSERT_STDOUT=1 -Wno-cpp -fPIC -MD -MQ librz/arch/librz_arch.so.0.8.0.p/p_arch_msp430.c.o -MF librz/arch/librz_arch.so.0.8.0.p/p_arch_msp430.c.o.d -o librz/arch/librz_arch.so.0.8.0.p/p_arch_msp430.c.o -c ../librz/arch/p/arch_msp430.c
In file included from ../librz/arch/p/analysis/analysis_msp430.c:11,
                 from ../librz/arch/p/arch_msp430.c:7:
../librz/arch/isa/msp430/msp430_disas.h:10:20: error: ‘msp430_register_names’ defined but not used [-Werror=unused-variable]
   10 | static const char *msp430_register_names[] = {
      |                    ^~~~~~~~~~~~~~~~~~~~~
```

XVilka (changes requested):
1. Please add a test
2. See the warning.

librz/arch/isa/msp430/msp430_disas.h
moste00: Sorry for the hiatus, on it now.

librz/arch/isa/msp430/msp430_instruction_parse.inc
```diff
@@ -0,0 +1,195 @@
...
+	return m;
+}
+
+static Msp430DestinationAddressingMode parse_dst(char *dst_str, RZ_OUT ut32 *dst) {
+	switch (dst_str[0]) {
+	case '&':
+		*dst = parse_4digit_hex(dst_str + 1);
+		return DST_ABS;
+	case '0':
+		int end_offset;
```
XVilka:
```
 tcc -Ilibrz/arch/librz_arch.so.0.8.0.p -I. -I.. -Ilibrz -I../librz -Ilibrz/include -I../librz/include -I../librz/arch/isa -I../librz/arch/isa_gnu -Ilibrz/arch -I../librz/arch -Ilibrz/util/sdb/src -I../librz/util/sdb/src -I../librz/bin/format -I../librz/type/parser -I../subprojects/capstone-next/include -I../subprojects/capstone-next/include/capstone -Isubprojects/rzspp -I../subprojects/rzspp -Wall -O2 -g -Wimplicit-fallthrough=3 -DRZ_PLUGIN_INCORE=1 -D_GNU_SOURCE -Werror=sizeof-pointer-memaccess -fvisibility=hidden -MD -MF librz/arch/librz_arch.so.0.8.0.p/isa_msp430_msp430_il.c.o.d -o librz/arch/librz_arch.so.0.8.0.p/isa_msp430_msp430_il.c.o -c ../librz/arch/isa/msp430/msp430_il.c
In file included from ../librz/arch/isa/msp430/msp430_il.c:208:
../librz/arch/isa/msp430/msp430_instruction_parse.inc:113: error: identifier expected
```

XVilka (changes requested):
Please also rebase on top of the latest `dev`

librz/arch/isa/msp430/msp430_il.c
```diff
@@ -0,0 +1,208 @@
+// SPDX-FileCopyrightText: 2023 Mostafa Mahmoud <ubermenchun@gmail.com>
```
XVilka: Isn't it 2024?

librz/arch/isa/msp430/msp430_instruction_parse.inc
```diff
@@ -0,0 +1,195 @@
+#include <rz_types.h>
```
XVilka: Missing SPDX

librz/arch/isa/msp430/msp430_instruction_parse.inc
```diff
@@ -0,0 +1,195 @@
...
+	rz_warn_if_fail(*(curr + 1) == 'b');
+	return 0; // byte sized read and write
+}
+
+static ut16 parse_4digit_hex(char *num);
+static ut16 parse_immediate(char *imm);
+static ut16 parse_reg(char *reg);
+static ut16 parse_hex(char *num, int *end_offset);
+
+static Msp430SourceAddressingMode parse_src(char *src_str, RZ_OUT ut32 *src, RZ_OUT int *dst_offset) {
```
XVilka: Instead of parsing string, I wonder if it's better to modify the `librz/arch/isa/msp430/msg430_disas.c` instead and provide this kind of information as a C structure, similarly how other disassemblers do it, e.g. capstone-based ones.

XVilka (changes requested):

XVilka:
@moste00 any updates on this PR?

moste00:
> @moste00 any updates on this PR?

Really sorry for being held up this long, many many interviews and disruptions in my irl life.

What remains isn't that much. Just a rebase into a newer dev, and debugging the issue of why `git-clang-format` gives the green on my machine but somehow fails in the CI even though the 2 versions match. There is a complaint by tinyCC about a switch case somewhere, but I think it's a nitpick and I have a fix in mind. I will push later today and ask for review.

All in all, most of the major work is done IMO, the supported instructions pass into the lifter and come out as IL trees. It's just a matter of seeing why the CI doesn't think so from now on.

librz/arch/isa/msp430/msp430_instruction_parse.inc
moste00: Parsing the operands is definitely pretty ugly, but it's way easier to do than messing with the internals of the disassembler.

Someone, very likely me, will probably touch the MSP430 plugin soon to add assembly anyway (there is no MSP430 assembler at the moment, I'm using an external Python script I found somewhere on Github to obtain the binary of instructions), that will be a suitable moment to rethink `disas.c`..

XVilka: I recommend focusing on improving the disassembly, analysis, and RzIL first and foremost before spending time adding the assembler. These are higher priority, and improving disassembler to provide instruction details for the analysis is much more important than writing an assembler.

XVilka:
@moste00 please check Mattermost.

librz/arch/isa/msp430/msp430_il.c
Rot127:
```suggestion
#include <msp430/msp430_il.h>
```

Sorry, missed this one. This is the suggestion I had.

librz/arch/isa/msp430/msp430_il.c
```diff
@@ -0,0 +1,228 @@
+// SPDX-FileCopyrightText: 2024 Mostafa Mahmoud <ubermenchun@gmail.com>
+// SPDX-License-Identifier: LPGL-3.0-only
```
Rot127:
```suggestion
// SPDX-License-Identifier: LGPL-3.0-only
```

librz/arch/isa/msp430/msp430_il.h
```diff
@@ -0,0 +1,51 @@
+// SPDX-FileCopyrightText: 2024 Mostafa Mahmoud <ubermenchun@gmail.com>
+// SPDX-License-Identifier: LPGL-3.0-only
```
Rot127:
```suggestion
// SPDX-License-Identifier: LGPL-3.0-only
```

librz/arch/isa/msp430/msp430_instruction_parse.inc
```diff
@@ -0,0 +1,209 @@
+// SPDX-FileCopyrightText: 2024 Mostafa Mahmoud <ubermenchun@gmail.com>
+// SPDX-License-Identifier: LPGL-3.0-only
```
Rot127:
```suggestion
// SPDX-License-Identifier: LGPL-3.0-only
```

librz/arch/isa/msp430/msp430_instruction_parse.inc
```diff
@@ -0,0 +1,209 @@
...
+
+#include <rz_types.h>
+#include "msp430_disas.h"
+#include "msp430_il.h"
```
Rot127:
```suggestion
#include <msp430/msp430_disas.h>
#include <msp430/msp430_il.h>
```

Rot127 (commented):
The typo in the license identifier was likely the reason for the failing clang-format

librz/arch/isa/msp430/msp430_instruction_parse.inc
moste00: You're actually right, the parsing code is a constant source of overflow headache and bugs, it keeps making AddressSanitizer complain about illegal reads. I decided to remove the problem from its root and delete the code, I will instead make the disassembler bubble up the information I need.

XVilka:
@moste00 please rebase

XVilka (changes requested):
Doesn't build:
```
FAILED: librz/arch/librz_arch.so.0.8.0.p/isa_msp430_msp430_il.c.o 
tcc -Ilibrz/arch/librz_arch.so.0.8.0.p -I. -I.. -Ilibrz -I../librz -Ilibrz/include -I../librz/include -I../librz/arch/isa -I../librz/arch/isa_gnu -Ilibrz/arch -I../librz/arch -Ilibrz/util/sdb/src -I../librz/util/sdb/src -I../librz/bin/format -I../librz/type/parser -I../subprojects/capstone-next/include -I../subprojects/capstone-next/include/capstone -Isubprojects/rzspp -I../subprojects/rzspp -Wall -O2 -g -Wimplicit-fallthrough=3 -DRZ_PLUGIN_INCORE=1 -D_GNU_SOURCE -Werror=sizeof-pointer-memaccess -fvisibility=hidden -MD -MF librz/arch/librz_arch.so.0.8.0.p/isa_msp430_msp430_il.c.o.d -o librz/arch/librz_arch.so.0.8.0.p/isa_msp430_msp430_il.c.o -c ../librz/arch/isa/msp430/msp430_il.c
In file included from ../librz/arch/isa/msp430/msp430_il.c:16:
../librz/arch/isa/msp430/msp430_il_getset.inc:48: error: 'address' undeclared
```
Please after that finish those instructions that aren't implemented and add tests for them in `test/db/asm/msp430`

librz/arch/isa/msp430/msp430_il.c
```diff
@@ -0,0 +1,333 @@
...
+#define MEM_ADDR_SIZE 16U
+
+#include <msp430/msp430_register_names.h>
+#include <msp430/msp430_il_getset.inc>
+#include <msp430/msp430_il_flags.inc>
+#include <msp430/msp430_il_jmp_utils.inc>
+
+#include "rz_il/rz_il_opbuilder_begin.h"
+
+// ************************************* ------------------- ********************************* //
```
XVilka: Instead of this I recommend Doxygen grouping, see how it's done in other files as well. See https://www.doxygen.nl/manual/grouping.html

test/db/asm/msp430
```diff
@@ -1,5 +1,7 @@
...
+d "and.b #0xf0f0, r12" 7cf0f0f0 0x0 (seq (set old_destination (cast 8 false (var r12))) (set source (bv 8 0xf0)) (set result (& (var source) (var old_destination))) (set r12 (cast 16 false (var result))) (set sr (| (| (| (& (var sr) (bv 16 0xfef8)) (| (<< (ite (msb (var result)) (bv 16 0x1) (bv 16 0x0)) (bv 8 0x2) false) (<< (ite (is_zero (var result)) (bv 16 0x1) (bv 16 0x0)) (bv 8 0x1) false))) (ite (! (is_zero (var result))) (bv 16 0x1) (bv 16 0x0))) (<< (ite false (bv 16 0x1) (bv 16 0x0)) (bv 8 0x8) false))))
 d "bic #2, r10" 2ac3
```
XVilka: See all those unlifted instructions

XVilka (changes requested):

test/db/asm/msp430
```diff
@@ -1,54 +1,64 @@
...
+d "clr 0x0033" 80433300 0x0 (storew 0 (+ (bv 16 0x33) (bv 16 0x0)) (bv 16 0x0))
+d "clr 0xfffa(r4)" 8443faff 0x0 (storew 0 (+ (bv 16 0xfffa) (var r4)) (bv 16 0x0))
+d "clrn" 22c2 0x0 (set sr (& (var sr) (bv 16 0xfffb)))
+d "clrz" 22c3 0x0 (set sr (& (var sr) (bv 16 0xfffd)))
+d "cmp.b #0x003b, 0x0(r11)" fb903b000000 0x0 (seq (set op0 (bv 8 0x3b)) (set op1 (+ (bv 8 0x1) (~ (var op0)))) (set op2 (loadw 0 8 (+ (bv 16 0x0) (var r11)))) (set result (+ (var op1) (var op2))) (set sr (| (| (| (| (| (| (& (var sr) (bv 16 0xfef8)) (<< (ite (is_zero (var result)) (bv 16 0x1) (bv 16 0x0)) (bv 8 0x1) false)) (<< (ite (msb (var result)) (bv 16 0x1) (bv 16 0x0)) (bv 8 0x2) false)) (ite (|| (|| (&& (msb (var op1)) (msb (var op2))) (&& (! (msb (var result))) (msb (var op1)))) (&& (msb (var op2)) (! (msb (var result))))) (bv 16 0x1) (bv 16 0x0))) (ite (== (var op0) (bv 8 0x0)) (bv 16 0x1) (bv 16 0x0))) (<< (ite (&& (! (^^ (msb (var op1)) (msb (var op2)))) (^^ (msb (var result)) (msb (var op1)))) (bv 16 0x1) (bv 16 0x0)) (bv 8 0x8) false)) (<< (ite (== (var op0) (bv 8 0x80)) (bv 16 0x1) (bv 16 0x0)) (bv 8 0x8) false))))
 d "dadc r11" 0ba3
```
XVilka: `dadc` instruction still left unlifted

XVilka (changes requested):
This PR is quite close to being complete. Only few steps left to do.

XVilka:
@moste00 check also this ASAN (UBSAN) error:
```
[XX] db/asm/msp430 <asm> clr 0xfffa(r4)
-- <asm> clr 0xfffa(r4) <--- 8443faff ---> <IL>
-- IL
(storew 0 (+ (bv 16 0xfffa) (var r4)) (bv 16 0x0))
../librz/arch/isa/msp430/msp430_disas.c:359:19: runtime error: left shift of 65530 by 16 places cannot be represented in type 'int'
```

librz/arch/isa/msp430/msp430_disas.c
```diff
@@ -104,8 +89,7 @@ static void remove_second_operand(struct msp430_cmd *cmd) {
 }
 
 /* TODO: This code is messy, needs to be refactored. */
-static int decode_emulation(ut16 instr, struct msp430_cmd *cmd) {
-	int ret = -1;
+static void decode_emulation(ut16 instr, struct msp430_cmd *cmd) {
```
Rot127: The `TODO` above can be removed, right? Code looks petty clean to me.
But I would remove the `ut8 as, ad, src, dst, bw, opcode;` declaration.
Also add an `else` case which gives an error message (or whatever is appropriate) and fill `cmd` with invalid data.

librz/arch/isa/msp430/msp430_disas.c
```diff
@@ -278,6 +330,8 @@ static int decode_addressing_mode(ut16 instr, ut16 op1, ut16 op2, struct msp430_
 	switch (ad) {
 	case 0: /* register mode */
 		snprintf(dstbuf, sizeof(dstbuf), ", %s", msp430_register_names[dst]);
+		cmd->dst_mode = MSP430_REG;
```
Rot127, line 333: Please add a default case for the `switch(ad)`

librz/arch/isa/msp430/msp430_disas.c
```diff
@@ -370,10 +431,14 @@ static int decode_oneop_opcode(ut16 instr, ut16 op, struct msp430_cmd *cmd) {
...
 			case MSP430_R3:
 				snprintf(cmd->operands, sizeof(cmd->operands), "#0");
+				cmd->dst = 0;
+				cmd->dst_mode = MSP430_IMM;
 				break;
 			default:
 				snprintf(cmd->operands, sizeof(cmd->operands),
 					"%s", msp430_register_names[get_dst(instr)]);
+				cmd->dst = get_dst(instr);
+				cmd->dst_mode = MSP430_REG;
```
Rot127, line 442:
```suggestion
				cmd->dst = get_dst(instr);
				cmd->dst_mode = MSP430_REG;
				break;
```

Rot127: Also below.

librz/arch/isa/msp430/msp430_disas.h
```diff
@@ -78,22 +99,48 @@ enum msp430_registers {
 	MSP430_R15,
 };
 
+typedef enum Msp430AddressingMode {
+	// register: Rn, contents of Rn
+	MSP430_REG,
```
Rot127:
```suggestion
	MSP430_REG, ///< register: Rn, contents of Rn
```

Also below

librz/arch/isa/msp430/msp430_disas.h
```diff
@@ -45,18 +45,39 @@ enum msp430_twoop_opcodes {
 	MSP430_AND,
 };
 
-enum msp430_addr_modes {
-	MSP430_DIRECT,
-	MSP430_INDEXED,
-	MSP430_INDIRECT,
-	MSP430_INDIRECT_INC,
+enum msp430_emulated_opcodes {
```
Rot127: Please `typedef` this enum.

Rot127: Also `msp430_cmd_type` while you are at it.

librz/arch/isa/msp430/msp430_disas.h
```diff
@@ -78,22 +99,48 @@ enum msp430_registers {
...
 	// Length of array: 2 * ('0', 'x', 4-digit hexadecimal numeral, '(', 'r', 2-digit
 	// decimal numeral, ')'), ',', ' ', '\0'
 	char operands[2 * (2 + 4 + 2 + 3) + 2 + 1];
+
+	// The source and the dst of the operands, along with their modes
+	// This info is contained in the strings above, but parsing strings to obtain it is ugly so we replicate it here
+	ut32 src;
+	ut32 dst;
+	Msp430AddressingMode src_mode;
+	Msp430AddressingMode dst_mode;
```
Rot127: For the whole `msp430_cmd` struct:
- Please `typedef` it.
- Document each member separately with `///< ...`
- Add a doxygen to it with
  ```
  /**
   * \brief ...
   */
  ```

librz/arch/isa/msp430/msp430_il.c
```diff
@@ -0,0 +1,433 @@
+// SPDX-FileCopyrightText: 2024 Mostafa Mahmoud <ubermenchun@gmail.com>
+// SPDX-License-Identifier: LGPL-3.0-only
+
+#include <msp430/msp430_il.h>
+
+// ``The 16-bit program counter (PC/R0) points to the next instruction to be executed.``
```
Rot127:
```suggestion
// The 16-bit program counter (PC/R0) points to the next instruction to be executed.
```

Also below. Or is this valid doxygen syntax I am not aware of?

librz/arch/isa/msp430/msp430_il.c
```diff
@@ -0,0 +1,433 @@
...
+
+#include "rz_il/rz_il_opbuilder_begin.h"
+
+// ************************************* ------------------- ********************************* //
+// ************************************* ------------------- ********************************* //
+// ************************************* One-Operand Lifters ********************************* //
+// ************************************* ------------------- ********************************* //
+// ************************************* ------------------- ********************************* //
+
+RZ_OWN RzILOpEffect *rz_msp430_lift_call_instr(RzAnalysis *analysis, const Msp430Instruction *op, ut64 current_addr, int instr_size) {
```
Rot127:
```suggestion
RZ_OWN RzILOpEffect *rz_msp430_lift_call_instr(RZ_BORROW RzAnalysis *analysis, const Msp430Instruction *op, ut64 current_addr, int instr_size) {
```

Also below

librz/arch/isa/msp430/msp430_il.c
```diff
@@ -0,0 +1,433 @@
...
+	return set_destination(op, result, current_addr);
+}
+
+// ************************************* ---------------------------- ******************************** //
+// ************************************* ---------------------------- ******************************** //
+// ************************************* Emulated Instruction Lifters ******************************** //
+// ************************************* ---------------------------- ******************************** //
+// ************************************* ---------------------------- ******************************** //
+
+RzILOpEffect *rz_msp430_lift_ret_instr(RzAnalysis *analysis, const Msp430Instruction *op, ut64 current_addr, int instr_size) {
```
Rot127: `Rz_OWN` and `RZ_NONNULL`/`RZ_NONNULL` missing

Rot127: Also below

librz/arch/isa/msp430/msp430_il.c
```diff
@@ -0,0 +1,433 @@
...
+RZ_OWN RzILOpEffect *rz_msp430_dummy() {
+	RZ_LOG_ERROR("UNREACHABLE CONTROL FLOW");
+	return NOP();
+}
+
+RzILOpEffect *rz_msp430_lift_todo(RzAnalysis *analysis, const Msp430Instruction *op, ut64 current_addr, int instr_size) {
+	NOT_IMPLEMENTED;
+}
+
+#include "rz_il/rz_il_opbuilder_end.h"
```
Rot127:
```suggestion
#include <rz_il/rz_il_opbuilder_end.h>
```

librz/arch/isa/msp430/msp430_il.c
```diff
@@ -0,0 +1,433 @@
...
+
+static const MSP430InstructionLifter one_op_lifters[] = {
+	[MSP430_RRC] = rz_msp430_lift_rrc_instr,
+	[MSP430_SWPB] = rz_msp430_lift_swpb_instr,
+	[MSP430_RRA] = rz_msp430_lift_todo, // TODO
```
Rot127:
```suggestion
	[MSP430_RRA] = rz_msp430_lift_todo,
```

Also below

Rot127: Also, this syntax is C20 standard, right? Could conflict with some compilers. cc @XVilka

librz/arch/isa/msp430/msp430_il.c
```diff
@@ -0,0 +1,433 @@
...
+	[MSP430_SXT] = rz_msp430_lift_sxt_instr,
+	[MSP430_PUSH] = rz_msp430_lift_push_instr,
+	[MSP430_CALL] = rz_msp430_lift_call_instr,
+	[MSP430_RETI] = rz_msp430_lift_todo, // TODO
+	[MSP430_UNUSED] = NULL
+};
```
Rot127, line 418: Newline missing

librz/arch/isa/msp430/msp430_il_flags.inc
```diff
@@ -0,0 +1,247 @@
+// SPDX-FileCopyrightText: 2024 Mostafa Mahmoud <ubermenchun@gmail.com>
+// SPDX-License-Identifier: LGPL-3.0-only
+
+#include <msp430/msp430_disas.h>
+#include "rz_il/rz_il_opbuilder_begin.h"
```
Rot127:
```suggestion
#include <rz_il/rz_il_opbuilder_begin.h>
```

librz/arch/isa/msp430/msp430_il_flags.inc
```diff
@@ -0,0 +1,247 @@
...
+	return LOGOR(old_sr_value, SHIFTL0(BOOL_TO_BV(new_value, 16), U8(1)));
+}
+
+RZ_OWN RzILOpBitVector *update_sr_n_flag(RzILOpBool *new_value, RzILOpBitVector *old_sr_value) {
+	return LOGOR(old_sr_value, SHIFTL0(BOOL_TO_BV(new_value, 16), U8(2)));
+}
+
+RZ_OWN RzILOpBitVector *update_sr_nz_flags(RzILOpBitVector *new_value, RzILOpBitVector *old_sr_value) {
+	RzILOpBool *n_flag_value = MSB(new_value);
+	RzILOpBool *z_flag_value = IS_ZERO(DUP(new_value)); // must DUP new_value because it has been used before
```
Rot127:
```suggestion
	RzILOpBool *z_flag_value = IS_ZERO(DUP(new_value));
```

librz/arch/isa/msp430/msp430_il_flags.inc
```diff
@@ -0,0 +1,247 @@
...
+
+RZ_OWN RzILOpEffect *set_incb_flags(const char *result_name, const char *old_destination_name) {
+	return do_set_inc_flags(result_name, old_destination_name, U8(0xFF), U8(0x7F));
+}
+
+RZ_OWN RzILOpEffect *set_inc_flags(const char *result_name, const char *old_destination_name) {
+	return do_set_inc_flags(result_name, old_destination_name, U16(0xFFFF), U16(0x7FFF));
+}
+
+RZ_OWN RzILOpEffect *do_set_dec_flags(const char *result_name, const char *old_destination_name, RzILOpBitVector *zero_const, RzILOpBitVector *carry_const, RzILOpBitVector *overflow_const) {
```
Rot127: Please mark all `RzILOpBitVector` parameters of those helper functions as `RZ_OWN` (if they are not duplicated before use, then `RZ_BORROW` of course).

librz/arch/isa/msp430/msp430_il_flags.inc
```diff
@@ -0,0 +1,247 @@
...
+
+RzILOpEffect *set_clrz_flags() {
+	return MSP430_SETR(MSP430_SR, LOGAND(MSP430_GETR(MSP430_SR), U16(~(1 << 1))));
+}
+
+RzILOpBool *check_if_zero_carry() {
+	return IS_ZERO(LOGAND(MSP430_GETR(MSP430_SP), U16(1)));
+}
+
+#include "rz_il/rz_il_opbuilder_end.h"
```
Rot127:
```suggestion
#include <rz_il/rz_il_opbuilder_end.h>
```

Rot127: Also below

librz/arch/isa/msp430/msp430_il_jmp_utils.inc
```diff
@@ -0,0 +1,60 @@
...
+RzILOpBool *mk_jge_cond(RzILOpBitVector *curr_sr) {
+	int mask = 4 | (1 << 8);
+	RzILOpBitVector *nv = LOGAND(curr_sr, U16(mask));
+	// test if both the N and V flags are set or reset
+	return OR(IS_ZERO(nv), EQ(DUP(nv), U16(mask)));
+}
+
+#include "rz_il/rz_il_opbuilder_end.h"
+
+typedef RzILOpBool *(*JmpConditionConstructor)(RzILOpBitVector *curr_sr);
```
Rot127: Missing newlines. Also below

librz/arch/isa/msp430/msp430_il_jmp_utils.inc
```diff
@@ -0,0 +1,60 @@
+// SPDX-FileCopyrightText: 2024 Mostafa Mahmoud <ubermenchun@gmail.com>
```
Rot127: Is there a specific reason this is an `inc` file? Can't it be its own header file with `static inline` functions?
`inc` files tend to break the syntax highlighting and is really just useful for huge amounts of generated code.

librz/arch/isa/msp430/msp430_register_names.h
```diff
@@ -0,0 +1,28 @@
+// SPDX-FileCopyrightText: 2024 Mostafa Mahmoud <ubermenchun@gmail.com>
+// SPDX-License-Identifier: LGPL-3.0-only
+
+#ifndef MSP430_REGISTER_NAMES_H
+#define MSP430_REGISTER_NAMES_H
+
+#include <rz_analysis.h>
```
Rot127:
```suggestion
```

Rot127: The `include` belongs to `rmsp430_il.h`

Rot127 (changes requested):
Great job!

Besides the review comments above (mostly small syntax things), please ensure that the code is leak free. This is important because those will become hot paths when analysis algos are build on top of RzIL.

You can test it with valgrind: `valgrind --leak-check=yes rz-test -j 1 test/db/asm/msp430` or with ASAN (by enabling an option, can't find currently which one though).

librz/arch/isa/msp430/msp430_disas.c
moste00: Well, it's defintely a bit repetitive and could use some abstraction into a more declarative form.

Every condition has the same general form ` if (opcode == ... && src == ... [&& dst == ...])`, so I'm thinking of abstracting this into an array that declaratively specifies those tuples of values (so `{ {val1, val2, val3}, {val4, val5, val6}, ...}`) and then a single loop that will interpret each tuple in the array with the same if condition. Worth it or too much of an overengineering gymnastics ?

Also, this condition **should** have no `else` clause, it's totally fine if none of the conditions match, that just means the instruction isn't an emulated instructions. We can't know beforehand if the instruction is emulated, the MSP430 spec just lists a bunch of usage patterns of 2-operand instructions and says they're "emulations" for specific "instructions", those are the patterns checked by this unusually large `if` statement. If none of the conditinos match, then the instruction isn't an emulated instruction.

moste00: For example, there is an instruction `INV` that bitwise-inverts its operand, but that's just a human-readable illusion over a specific usage of the `XOR` instruction with its `src` being the literal `0xFFFF`. `INV` doesn't really exist in the architecture, the assembler transparently replaces it with its equivalent `XOR`  "emulation" (a really fancy word for what's essentially a very trivial macro construct). The disassembler needs to be able to infer `INV` usages from the bitstream, so after every successful 2-op decode op it simply looks if the instruction we have just decoded is a XOR and if its operand is a literal equa to `0xFFFF` or `0xFF` if it's a byte instruction, if so, it replaces the string representation of a 2-op `XOR` with a 1-op `INV`. And so on for about 23 other specific usage patterns. If none of those patterns match, then we're not looking at an emulation, just a normal use.

librz/arch/isa/msp430/msp430_il_jmp_utils.inc
moste00: Hmmm, I usually think of  ` .h`s as files that contain "non-logic" code: declarations, type definitions, constants, and so on. I think of `.c`s as files that contain logic and should be seperately compiled, but `inc`s are those that contain logic but shouldn't be seperately compilable, as they're considered just fragment of a single "owning" `.c` file.

That's my "head cannon", so to speak. But I can change it anyway if it conflicts with any convention or aesthetic preference.

librz/arch/isa/msp430/msp430_disas.c
Rot127: > this condition should have no else clause

Ok, wasn't sure if it was supposed to be like this.

Regarding cleaness: IMHO this is fine. No need to overengineer here.

librz/arch/isa/msp430/msp430_il_jmp_utils.inc
Rot127: I would vote to move it to a normal .c .h form. In my experience working with .inc files is always a little annoying in one or another aspect. @XVilka @DMaroo

librz/arch/isa/msp430/msp430_il.c
moste00: No, it's not valid doxygen. It's just  a quote from the manual.

librz/arch/isa/msp430/msp430_disas.c
moste00: There is an assignment below it that sets the return value, so it shouldn't break.

moste00:
@Rot127 Thanks a lot for your kind words and for the comprehensive review.

> Besides the review comments above (mostly small syntax things), please ensure that the code is leak free. This is important because those will become hot paths when analysis algos are build on top of RzIL.
 
> You can test it with valgrind: `valgrind --leak-check=yes rz-test -j 1 test/db/asm/msp430` or with ASAN (by enabling an option, can't find currently which one though).

Done. Valgrind output below indicate no leaks.

image

XVilka (changes requested):
```
Run python3 rz-bindgen/src/lint.py \
Mismatched function annotation for rz_msp430_lift_instr at <rizin/librz/arch/isa/msp430/msp430_il.h:14:29> : RZ_NONNULL (was RZ_BORROW at <rizin/librz/arch/isa/msp430/msp430_il.c:493:29>
Mismatched function annotation for rz_msp430_lift_instr at <rizin/librz/arch/isa/msp430/msp430_il.h:14:29> : None (was RZ_NONNULL at <rizin/librz/arch/isa/msp430/msp430_il.c:493:29>
Mismatched function annotation for rz_msp430_il_config at <rizin/librz/arch/isa/msp430/msp430_il.h:16:35> : RZ_NONNULL (was RZ_BORROW at <rizin/librz/arch/isa/msp430/msp430_il.c:517:35>
Mismatched function annotation for rz_msp430_il_config at <rizin/librz/arch/isa/msp430/msp430_il.h:16:35> : None (was RZ_NONNULL at <rizin/librz/arch/isa/msp430/msp430_il.c:517:35>
```

librz/arch/isa/msp430/msp430_il.c
```diff
@@ -0,0 +1,521 @@
...
+	return set_destination(op, result, current_addr);
+}
+
+RZ_OWN RzILOpEffect *rz_msp430_lift_bic_instr(RZ_BORROW RZ_NONNULL RzAnalysis *analysis, const Msp430Instruction *op, ut64 current_addr, int instr_size) {
+	RzILOpBitVector *result = LOGAND(LOGNOT(get_source(op, current_addr)), get_destination(op, current_addr));
+	return set_destination(op, result, current_addr);
+}
+/* @} */
+
+// ************************************* ---------------------------- ******************************** //
```
XVilka: You forgot to remove non-Doxygen comment lines then

librz/arch/isa/msp430/msp430_il.c
```diff
@@ -0,0 +1,521 @@
...
+}
+
+RZ_OWN RzILOpEffect *rz_msp430_lift_pop_instr(RZ_BORROW RZ_NONNULL RzAnalysis *analysis, const Msp430Instruction *op, ut64 current_addr, int instr_size) {
+	return SEQ2(
+		(op->is_byte) ? set_destination(op, LOAD(MSP430_GETR(MSP430_SP)), current_addr) : set_destination(op, LOADW(16, MSP430_GETR(MSP430_SP)), current_addr),
+		MSP430_SETR(MSP430_SP, ADD(MSP430_GETR(MSP430_SP), U16(2))));
+}
+/** @} */
+
+// ************************************* ------------------- ********************************* //
```
XVilka: Same

librz/arch/isa/msp430/msp430_il.c
XVilka: Please convert to Doxygent then.

XVilka (approved):
LGTM but please fix these comments

librz/arch/isa/msp430/msp430_il.c
moste00: Can we leave those comment markers ? They're a very obvious way to demarcate the file into the conceptually seperate "Regions". Doxygen groups don't look as visually distinct. 

With those comments, I can look at the file overview in the small window in VSCode and immeditately know that the file is seperated into independent chunks.

librz/arch/isa/msp430/msp430_disas.c
Rot127: This gets set anyways, the `break;` only exits the `switch`. But it is optional anyways. I just like to have any `case/default` with a `break`. Maybe someone changes the order of the `default` and other `cases` in the future and forgets it the `break`. This was my thinking here

librz/arch/isa/msp430/msp430_il.c
XVilka: I don't strongly oppose it but it will be the only file that uses such a way to group, so it will be inconsistent with other code.

librz/arch/isa/msp430/msp430_disas.h
```diff
@@ -43,23 +42,44 @@ enum msp430_twoop_opcodes {
...
 
-enum msp430_cmd_type {
+typedef enum msp430_cmd_type {
 	MSP430_ONEOP,
 	MSP430_TWOOP,
 	MSP430_JUMP,
-	MSP430_INV,
-};
+	MSP430_EMULATE,
+	MSP430_INVALID,
```
Rot127, line 79: Careful with the `INVALID` at the last position.
If some struct is initialized with all `0`, the value won't be `INVALId`. Is this on purpose?

librz/arch/isa/msp430/msp430_disas.h
```diff
@@ -76,25 +96,45 @@ enum msp430_registers {
...
+	st16 jmp_addr; ///< for jumps, what's the address to jump to
+	ut8 jmp_cond; ///< for jumps, when will the jump be taken
 
-	// Null-delimited string representation of an assembly operation mnemonic.
 	// Length of array: 'i', 'n', 'v', 'a', 'l', 'i', 'd', '\0'
 	// (This is longer than any real assembly mnemonic.)
-	char instr[7 + 1];
+	char instr[7 + 1]; ///< Null-delimited string representation of an assembly operation mnemonic.
+
+	bool is_byte; ///< does it have a .b ?
```
Rot127:
```suggestion
	bool is_byte; ///< does it have a ".b" postfix?
```

librz/arch/isa/msp430/msp430_il_getset.h
```diff
@@ -0,0 +1,142 @@
...
+			return STORE(address, new_value);
+		}
+
+		rz_warn_if_reached();
+		return NOP();
+	}
+	}
+}
+
+ut8 word_sizedness(const Msp430Instruction *op) {
```
Rot127:
```suggestion
static inline ut8 word_sizedness(const Msp430Instruction *op) {
```

Rot127: The ones above should also be `static`, since they are in a header file?

Rot127: @XVilka Forgot what the usual decisions on theses are (since they are quite long). `static` or moving to `.c`?

librz/arch/isa/msp430/msp430_il.c
Rot127: @moste00 Don't forget this one, please.

librz/arch/isa/msp430/msp430_il.c
```diff
@@ -0,0 +1,521 @@
...
+
+#include <msp430/msp430_il.h>
+
+// ``The 16-bit program counter (PC/R0) points to the next instruction to be executed.``
+#define PC_SIZE 16
+// ``The low byte of a word is always an even address. The high byte is at the next odd address.
+//   For example, if a data word is located at address xxx4h, then the low byte of that data word
+//   is located at address xxx4h, and the high byte of that word is located at address xxx5h.``
+#define IS_BIG_ENDIAN false
+// implied by the width of the PC and other registers (which are used as pointers in the relevant addressing modes)
```
Rot127: Just reacently learned, that you can make those `doxygen` comments as well!

```
/**
 * \brief ...
 */
```

Might be usefult here

Rot127: Also, please `#undef` them at the end of the file. The names are very generic and might be used somewhere else.

librz/arch/isa/msp430/msp430_il.c
Rot127: Agree partially with @XVilka here. But I like these comment blocks as well. But one line is enough, IMHO.

Maybe without those ones?
```
// ************************************* ---------------------------- ******************************** //
```

Rot127 (commented):
More nitpicks :D Sorry, always happens when looking at stuff a second time. But again, in general really nice.

librz/arch/isa/msp430/msp430_il_getset.h
XVilka: If it's not used across multiple files - just `static` (and `inline` optionally, though often that's reduntant) and move to a C file directly, if it is used across multiple files - keep it in a header file, and mark as `static inline`, see e.g. `librz/include/rz_vector.h`.

librz/arch/isa/msp430/msp430_il.c
moste00: Hmmm, the generic names are probably not a problem because that's a `.c ` so it will never be included anywhere.

I could simply prefix a `MSP430_` to each one of them but it just feels redundant when the entire file is called `msp430_il.c` so of course any constant defined in it is about MSP430.

librz/arch/isa/msp430/msp430_disas.h
moste00: Hmm, I actually don't know, I'm not the one who wrote the disassembler. Porbably the original writer just didn't care enough.

From what I have seen the code do, that's a non-problem. Nobody relies on the fact that an uninitialized command struct is invalid, the entrypoint of the diassembler explicitly sets it on every control path.

librz/arch/isa/msp430/msp430_il_getset.h
moste00: In this case the move to a different file is intentional to conceptually seperate main logic from helpers. The helpers are used by every lifter and will clutter and derail the flow of the main `.c` if moved to it, not to mention make it balloon in size to almost double. 

So that's why I seperate them into coherent nice pieces with names like `msp430_il_getset.h` (memory/registers accessors that implements addressing modes) and `msp430_il_flags.h` (flags setters).

XVilka (approved):

librz/arch/isa/msp430/msp430_il.c
Rot127: >  so it will never be included anywhere.

ah yes, true.

librz/arch/isa/msp430/msp430_disas.c
```diff
@@ -291,26 +345,34 @@ static int decode_addressing_mode(ut16 instr, ut16 op1, ut16 op2, struct msp430_
...
+			cmd->dst_mode = MSP430_ABS;
+			cmd->dst = op;
 			break;
 		default: /* indexed mode */
 			snprintf(dstbuf, sizeof(dstbuf), ", 0x%x(%s)", op, msp430_register_names[dst]);
+			cmd->dst_mode = MSP430_INDX;
+			cmd->dst = (op << 8) | dst;
 		}
 		break;
+	default: RZ_LOG_WARN("UNREACHABLE");
```
Rot127:
```suggestion
	default: rz_warn_if_reached();
```

Rot127 (approved):
~~Please just replace this one `rz_warn_if_reached()` change.~~

PR #4402, opened 2024-03-30
Fix unescaping of `\\` esc seq

kazarmy:

**Detailed description**

This pr fixes the unescaping of some uses of the `\\` esc seq in the shell as can be seen in [this diff](https://github.com/rizinorg/rizin/commit/4af99675a49ee2ed9ec33e2a0ff7e4515b79b2e2).

**Test plan**

All builds are green.

**Closing issues**

...


XVilka (approved):

librz/core/cmd/cmd_api.c
```diff
@@ -2327,6 +2327,12 @@ static char *unescape_special_chars(const char *s, const char *special_chars) {
 	int i, j = 0;
 
 	for (i = 0; s[i]; i++) {
+		if (s[i] == '\\' && s[i + 1] == '\\') {
+			dst[j++] = '\\';
+			dst[j++] = '\\';
+			i++;
+			continue;
+		}
```
wargio, line 2335: maybe i'm wrong, but shouldn't you add a `\` for each `\` you find ?

wargio: also, judging as is, the second `j++` might generate a OOB write.

kazarmy: > maybe i'm wrong, but shouldn't you add a `\` for each `\` you find ?

(afaik) unescaping is for command input and it's done due to the existence of rzshell special chars like `#` and the backtick ... what it does is that it converts something like `\#` into just `#` (the `#` is escaped here since it's the rzshell comment char) ... this fix guarantees that `\\` stays as `\\` and not become `\`.

> also, judging as is, the second j++ might generate a OOB write.

since the code is copying to the destination `\\` (2 chars) for a source `\\` (2 chars), and on line 2326, there is
https://github.com/rizinorg/rizin/blob/8fec79ab6114b24f1c69fffbec21059b33516834/librz/core/cmd/cmd_api.c#L2326
if the allocation succeeds and assuming the rest of the function is correct, no OOB write can happen

wargio (approved):

PR #4417, opened 2024-04-05
.(: Don't unwrap quoted macro args

kazarmy:

**Detailed description**

The `.(` command (macro call) basically works by replacing itself with another rzshell command, so `.(` macro args should not be modified by any means. This pr prevents `.(` from unwrapping quoted macro args (i.e. remove their quotes), thus making the test work with regards to the macro arg `"a \\ b"`.

**Test plan**

All builds are green.

**Closing issues**

...

### DIFF
--- a/librz/core/cmd/cmd.c
+++ b/librz/core/cmd/cmd.c
@@ -3332,15 +3332,25 @@ static char *do_handle_ts_unescape_arg(struct tsr2cmd_state *state, TSNode arg, 
 	}
 }
 
-static RzCmdParsedArgs *parse_args(struct tsr2cmd_state *state, TSNode args, bool do_unwrap) {
+static bool is_arg_raw(const RzCmdDesc *cd, size_t i) {
+	const RzCmdDescArg *arg = rz_cmd_desc_get_arg(cd, i);
+	return arg && arg->type == RZ_CMD_ARG_TYPE_RAW;
+}
+
+static RzCmdParsedArgs *parse_args(struct tsr2cmd_state *state, TSNode args, bool do_unwrap, const RzCmdDesc *cd) {
 	if (ts_node_is_null(args)) {
 		return rz_cmd_parsed_args_newargs(0, NULL);
 	} else if (is_ts_args(args)) {
+		bool old_do_unwrap = do_unwrap;
 		uint32_t n_children = ts_node_named_child_count(args);
 		uint32_t i;
 		char **unescaped_args = RZ_NEWS0(char *, n_children);
 		for (i = 0; i < n_children; i++) {
 			TSNode arg = ts_node_named_child(args, i);
+			do_unwrap = old_do_unwrap;
+			if (!do_unwrap && cd && cd->type != RZ_CMD_DESC_TYPE_OLDINPUT && !is_arg_raw(cd, i)) {
+				do_unwrap = true;
+			}
 			unescaped_args[i] = do_handle_ts_unescape_arg(state, arg, do_unwrap);
 		}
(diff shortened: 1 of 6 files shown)


ret2libc:
Not a fan of special casing commands this way. Can we add a `flags` field to `RzCmdDescHelp` which can be a set of values (similar to what we have for `RzCmdDescArg`) so in the YAML files we can set `.(` (and maybe other commands) as `RZ_CMD_FLAG_RAW_ARGS`.

What do you think?

kazarmy:
I agree but that will take some time.

kazarmy:
After looking more closely at the proposed and then seeing https://github.com/rizinorg/rizin/blob/6280243ff251ab54207331eb06a9c9edb7546463/librz/core/cmd_descs/cmd_interpret.yaml#L51-L54 I think having a new `RzCmdArgType` of `RZ_CMD_ARG_TYPE_RAW` is enough -- `RZ_CMD_ARG_TYPE_RAW` is basically `RZ_CMD_ARG_TYPE_STRING` but unprocessed.

ret2libc:
> After looking more closely at the proposed and then seeing
> 
> https://github.com/rizinorg/rizin/blob/6280243ff251ab54207331eb06a9c9edb7546463/librz/core/cmd_descs/cmd_interpret.yaml#L51-L54
> 
> I think having a new `RzCmdArgType` of `RZ_CMD_ARG_TYPE_RAW` is enough -- `RZ_CMD_ARG_TYPE_RAW` is basically `RZ_CMD_ARG_TYPE_STRING` but unprocessed.

Yes I thought about that, however you likely want *all* arguments unprocessed. If you want to put this at the arg level, then you must support unquoting some arguments, while keeping others raw. I'm ok with both approaches though.

kazarmy:
Putting this at the arg level seems to be more correct (e.g. the macro name shouldn't really be treated as a raw argument) but the implementation probably would require multiple calls to `rz_cmd_desc_get_arg()`.

kazarmy:
Ok it's done (at least for the goal stated in the title). Some possible points of interest:

1. `rz_cmd_desc_get_arg()` is not exactly O(1), but I don't think there are any commands with a significantly long argument descriptor list.
2. The API for `ts_node_handle_arg_prargs()` has become somewhat uglier, but I couldn't find a more elegant way -- I don't think it's wise to cast integers to pointers and vice versa.

librz/core/cmd/cmd.c
```diff
@@ -3332,15 +3332,25 @@ static char *do_handle_ts_unescape_arg(struct tsr2cmd_state *state, TSNode arg,
...
 	if (ts_node_is_null(args)) {
 		return rz_cmd_parsed_args_newargs(0, NULL);
 	} else if (is_ts_args(args)) {
+		bool old_do_unwrap = do_unwrap;
 		uint32_t n_children = ts_node_named_child_count(args);
 		uint32_t i;
 		char **unescaped_args = RZ_NEWS0(char *, n_children);
 		for (i = 0; i < n_children; i++) {
 			TSNode arg = ts_node_named_child(args, i);
+			do_unwrap = old_do_unwrap;
```
ret2libc: very minor thing, but i think it's better to define in the loop a `do_unwrap_arg` and set it to `do_unwrap` and change it in the `if` below, instead of changing the global `do_unwrap` and restoring it.

kazarmy: ok 632c51f2c33eb7734f12eec6a3173ec19f75353e

ret2libc (approved):

PR #4446, opened 2024-04-18
Remove globals and refactor to complete implementation of logging.

wargio:

**Detailed description**

- Removes almost all the globals from the logging api
- Logging C API are now `thread-safe`.
- Implements colored logs and more.
- `RZ_LOG_SILLY` has been removed.
- Should be a bit faster since `fprintf` has been converted to `fputs`.
- Renamed `log.traplevel` to `log.abortlevel` and hide it in release builds.

### DIFF
--- a/librz/core/cconfig.c
+++ b/librz/core/cconfig.c
@@ -2880,21 +2880,18 @@ static bool cb_log_config_traplevel(void *coreptr, void *nodeptr) {
 
 static bool cb_log_config_file(void *coreptr, void *nodeptr) {
 	RzConfigNode *node = (RzConfigNode *)nodeptr;
-	const char *value = node->value;
-	rz_log_set_file(value);
-	return true;
+	return rz_log_set_file(node->value);
 }
 
-static bool cb_log_config_srcinfo(void *coreptr, void *nodeptr) {
+static bool cb_log_config_show_sources(void *coreptr, void *nodeptr) {
 	RzConfigNode *node = (RzConfigNode *)nodeptr;
 	const char *value = node->value;
-	switch (value[0]) {
-	case 't':
-	case 'T':
-		rz_log_set_srcinfo(true);
-		break;
-	default:
-		rz_log_set_srcinfo(false);
+	if (rz_str_is_true(value)) {
+		rz_log_set_show_sources(true);
+	} else if (rz_str_is_false(value)) {
+		rz_log_set_show_sources(false);
+	} else {
+		return false;
 	}
 	return true;
 }
@@ -2902,13 +2899,12 @@ static bool cb_log_config_srcinfo(void *coreptr, void *nodeptr) {
 static bool cb_log_config_colors(void *coreptr, void *nodeptr) {
 	RzConfigNode *node = (RzConfigNode *)nodeptr;
 	const char *value = node->value;
-	switch (value[0]) {
-	case 't':
-	case 'T':
+	if (rz_str_is_true(value)) {
 		rz_log_set_colors(true);
-		break;
-	default:
+	} else if (rz_str_is_false(value)) {
 		rz_log_set_colors(false);
+	} else {
+		return false;
 	}
 	return true;
 }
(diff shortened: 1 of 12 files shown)


librz/util/log.c
```diff
@@ -5,133 +5,207 @@
...
 	[RZ_LOGLVL_VERBOSE] = "VERBOSE",
 	[RZ_LOGLVL_DEBUG] = "DEBUG",
 	[RZ_LOGLVL_INFO] = "INFO",
 	[RZ_LOGLVL_WARN] = "WARNING",
 	[RZ_LOGLVL_ERROR] = "ERROR",
-	[RZ_LOGLVL_FATAL] = "FATAL"
+	[RZ_LOGLVL_FATAL] = "FATAL",
 };
 
+static const char *level_tags_colors[] = {
```
XVilka, line 33: We should make this configurable in the future

librz/util/log.c
```diff
@@ -5,133 +5,207 @@
...
+	ret = true;
+
+end:
+	rz_th_lock_leave(logcfg.lock);
+	return ret;
 }
 
-RZ_API void rz_log_set_srcinfo(bool show_info) {
-	cfg_logsrcinfo = show_info;
+RZ_API void rz_log_set_show_sources(bool show_sources) {
```
XVilka, line 131: Could you please add Doxygen to the API functions that miss it while you are on it?

XVilka (changes requested):
Nice finding. Apart from Doxygen - LGTM

librz/util/log.c
wargio: yes, but for now, let's make it simple.

Rot127 (commented):
While you are at it. Could you add a simple `getter` for the current log level? Plugins can check it, when they print log messages.

Or, even better, add a logging function which takes the message as a string (not with `vargs`). Because `vargs` cannot be used from a few languages.

XVilka (approved):

Rot127:
Actually, if you do spent the time to add the logging API for plugins, it might be a good idea add an additional argument, where plugins can pass their name.
So the log messages can be distinguished from main Rizin.

```
[INFO] [Plugin] ...
```

Rot127 (approved):

librz/core/cconfig.c
```diff
@@ -3338,21 +3334,24 @@ RZ_API int rz_core_config_init(RzCore *core) {
...
 	SETICB("log.level", p ? atoi(p) : RZ_DEFAULT_LOGLVL, cb_log_config_level, "Target log level/severity"
-										  " (0:SILLY, 1:DEBUG, 2:VERBOSE, 3:INFO, 4:WARN, 5:ERROR, 6:FATAL)");
+										  " (0:DEBUG, 1:VERBOSE, 2:INFO, 3:WARN, 4:ERROR, 5:FATAL)");
 	free(p);
 	// RZ_LOGTRAP_LEVEL / log.traplevel
 	p = rz_sys_getenv("RZ_LOGTRAPLEVEL");
-	SETICB("log.traplevel", p ? atoi(p) : RZ_LOGLVL_FATAL, cb_log_config_traplevel, "Log level for trapping rizin when hit"
-											" (0:SILLY, 1:VERBOSE, 2:DEBUG, 3:INFO, 4:WARN, 5:ERROR, 6:FATAL)");
+	SETICB("log.traplevel", p ? atoi(p) : RZ_DEFAULT_LOGLVL_TRAP, cb_log_config_traplevel, "Log level for trapping rizin when hit"
+											       " (0:VERBOSE, 1:DEBUG, 2:INFO, 3:WARN, 4:ERROR, 5:FATAL)");
```
thestr4ng3r: VERBOSE/DEBUG swapped here?

wargio:
> Actually, if you do spent the time to add the logging API for plugins, it might be a good idea add an additional argument, where plugins can pass their name. So the log messages can be distinguished from main Rizin.
> 
> ```
> [INFO] [Plugin] ...
> ```

tbh, there is an option which tells you the function, file and line where that occurs.

librz/core/cconfig.c
wargio: nice catch.

PR #4503, opened 2024-05-19
Support ARM ELF THM_PC22 / THM_JUMP24 relocs

timiv:
This revision adds proper support for the following ARM ELF relocations:

- `RZ_ARM_THM_PC22`
- `RZ_ARM_THM_JUMP24`

These require a rather special encoding as some bits of the relative offset are encoded in a non-common fashion.

 

**Detailed description**

As of now, rizin implements a rather rudimentary handling of ARM ELF Relocation: it just overwrites the the patch address with the value calculated for the symbol. While this works in some cases, there are others that do not, most notably `ARM_THM_PC22` and `ARM_THM_JUMP24` that use a quite _different_ encoding.

With this change, the relocations are correctly handled and lead to properly encoded values. Note that this change doesn't handle possible overflows for which some thunk would need to be inserted.

The implementation was heavily inspired by [lld's implementation](https://github.com/llvm/llvm-project/blob/597ac471cc7da97ccf957362a7e9f7a52d6910ee/lld/ELF/Arch/ARM.cpp#L601), although the special cases (e.g. BL/BLX patching, missing J1J2 support) are not considered in this change.

I did not extend the documentation as it is not that detailed / does not even exist here.

**Test plan**

To test the change, it would be sufficient to check that after relocation, the `bl` instruction to the proper target exists and is not replaced by some unrelated instruction.

Example:

Previously:

    ┌ sym.mbedtls_md5_init();
    │           0x08000034      push  {r3, lr}                             ; RELOC TARGET 32 .text.mbedtls_md5_init @ 0x08000034 ; [04] -r-x section size 12 named .text.mbedtls_md5_init
    │           0x08000036      movs  r1, 0
    │           0x08000038      movs  r2, 0x58                             ; 'X'
    │           ;-- memset:
    │           0x0800003a      ldr   r0, [r1, 0xc]                        ; RELOC 32 memset
    │           0x0800003c      lsrs  r0, r0, 0x20
    └           0x0800003e      pop   {r3, pc}

With this change:

    ┌ sym.mbedtls_md5_init();
    │           0x08000034      push  {r3, lr}                             ; RELOC TARGET 32 .text.mbedtls_md5_init @ 0x08000034 ; [04] -r-x section size 12 named .text.mbedtls_md5_init
    │           0x08000036      movs  r1, 0
    │           0x08000038      movs  r2, 0x58                             ; 'X'
    │           ;-- memset:
    │           0x0800003a      bl    memset                               ; RELOC 24 memset
    └           0x0800003e      pop   {r3, pc}

I added a test for a binary that contains THM_PC22.

**Closing issues**

-

### DIFF
--- a/librz/bin/p/bin_elf.inc
+++ b/librz/bin/p/bin_elf.inc
@@ -1023,7 +1023,7 @@ static void patch_reloc_hexagon(RZ_INOUT RzBuffer *buf_patched, const ut64 patch
 // AARCH64-specific defines
 // Take the PAGE component of an address or offset.
 #define PG(x)         ((x) & ~0xFFFULL)
-#define PG_OFFSET(x)  ((x)&0xFFFULL)
+#define PG_OFFSET(x)  ((x) & 0xFFFULL)
 #define ADR_IMM_MASK1 (((1U << 2) - 1) << 29)
 #define ADR_IMM_MASK2 (((1U << 19) - 1) << 5)
 #define ADR_IMM_MASK3 (((1U << 19) - 1) << 2)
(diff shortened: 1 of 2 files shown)


XVilka:
To add a test you need:
1. Add a binary to https://github.com/rizinorg/rizin-testbins
2. Create a test in `test/db/formats/elf/` files, e.g. in `test/db/formats/elf/elf-relarm`

timiv:
Thanks @XVilka!

I added a test bin with rizinorg/rizin-testbins#151 as a prerequisite for this PR.

XVilka (approved):

PR #4522, opened 2024-05-29
[MSP430] Removed code which deletes indicator characters for certain asm syntaxes

### LINKED ISSUE #3324: MSP430 disassembly, missing indicator for Immediate Value or Memory Location
* Operating System: Windows 10
* Cutter version: 2.3.4
* Obtained from: Downloaded release from Cutter website or GitHub 
* File format: RAW Binary

After I have disassembled a MSP430 binary, I do not see an "#" or "(0x1234)" or "@" to show it is a Immediate Value or Memory Location.  It just shows the plain hex numbers.
There are some parenthesis that appear.  But they show only when Memory offsets are used.

Image5

moste00:

**Detailed description**

This should fix [issue #3324](https://github.com/rizinorg/cutter/issues/3324) in Cutter's bug log. 

...

**Test plan**

I have searched in the git history and tracked the introduction of the deleted code to commit hash [c83d5f4664f222f318f8e4ae3910aeb9e5515902](https://github.com/rizinorg/rizin/commit/c83d5f4664f222f318f8e4ae3910aeb9e5515902), I have tried to find any PR or detailed message associated with this hash, explaining why the code was needed, but I found none, the commit was probably added sometimes in the project's pre-history (back when it was radare2). 

As far as I can see, the deleted code has no other effect or purpose except the undesirable effect of deleting indicator characters from the disassembly when the asm syntax isn't AT&T. Any tests breaking due to the removal of this code are wrong tests that should be removed as well.

**To verify**

Run `rz-asm -a msp430 -d 3f401000` on dev and then on this PR's branch. On dev, it produces `mov 0x0010, r15` as output, which is wrong, the correct output should include the character '#' before the constant, as it appears on this PR's branch.

### DIFF
--- a/test/db/asm/msp430
+++ b/test/db/asm/msp430
@@ -1,20 +1,20 @@
 d "adc r11" 0b63
 d "adc.b r11" 4b63
-d "bic 2, r10" 2ac3
-d "bic 4, r10" 2ac2
-d "bic.b 2, 0x0026" e2c32600
-d "bic.b 4, 0x002e" e2c22e00
-d "bis.b 2, 0x0021" e2d32100
-d "br 0x0021" 30402100
-d "br 0x0003" 10420300
+d "bic #2, r10" 2ac3
+d "bic #4, r10" 2ac2
+d "bic.b #2, &0x0026" e2c32600
+d "bic.b #4, &0x002e" e2c22e00
+d "bis.b #2, &0x0021" e2d32100
+d "br #0x0021" 30402100
+d "br &0x0003" 10420300
 d "br sp" 0041
 d "br r10" 004a
-d "call 0xc096" b01296c0
+d "call #0xc096" b01296c0
 d "clr 0x0033" 80433300
 d "clr 0xfffa(r4)" 8443faff
 d "clrn" 22c2
 d "clrz" 22c3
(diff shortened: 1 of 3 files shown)


test/db/asm/msp430
XVilka, line 1: Could you please add an instruction test with the `%` indicator as well?

XVilka (changes requested):
LGTM with a small request

test/db/asm/msp430
moste00: Hmmm, I don't think there is a `%` indicator in MSP430 assembly. If you Ctrl-F "%" in the disassembler source code (`librz\arch\isa\msp430\msp430_disas.c`), you will find only its usages as a format control character to the printing function, it's never used as an actual output of the disassembler.

moste00: The indicators are only "#", "&", "@", and "@Rn+". The disassembler also seems to use the "$" indicator to imply a relative jump, but that usage is actually not-documented elsewhere in MSP430 resources, a relative jump is implied by just a plain address, which implicitly means the (ADDR + PC) addressing mode according the MSP430 manual.

XVilka (approved):

PR #4535, opened 2024-06-05
SoftFloat integration

### LINKED ISSUE #4252: Fix RzIL operations/handlers for 80-bit floats
80-bit floats are handled incorrectly by Rizin. Partial fixes (like #4251) help in fixing some of the IL operations. But many other IL operations are still broken.

Example:

```
x = 1.l (0x3fff8000000000000000)
y = -0.384677154767621320985l (0xbffdc4f4676e7d934000)

ret = rz_float_div(x, y, RZ_FLOAT_RMODE_RNE)

ret != -2.59958250082224806804l (0xc000d907403a7256641c)
```

`x`, `y`, and `div` are 80-bit floats here (`RZ_FLOAT_IEEE754_BIN_80`). Here `ret != x / y`. Also, the rounding mode does not matter, the result is incorrect regardless.

This issue is a blocker for tracetesting and running floating-point emulateme tests (rizinorg/rizin-testbins#147) for x86 floating-point RzIL (#3865).

DMaroo:

**Detailed description**

I have added failing tests when using 80-bit floats (similar to the one outlined in #4252). I have also added a new meson subproject which clones SoftFloat version ~2c~ 3e. It is hosted at https://github.com/rizinorg/softfloat. I have made minor modifications to the original source which include the following:

- Restructure the directories/files and remove any unnecessary files.
- Add a `meson.build` file.
- Modify the source to remove unused variables (so as to silence all warnings).

## Why use version 2c over 3e?

The complete changelog can be found [here](http://www.jhauser.us/arithmetic/SoftFloat-3/doc/SoftFloat-history.html), but it has mainly to do with bugfixes in the square root algorithm, support for [bfloat](https://en.wikipedia.org/wiki/Bfloat16_floating-point_format) and a new rounding mode (round to nearest, ties away: `RZ_FLOAT_RMODE_RNA` in Rizin and `softfloat_round_near_maxMag` in SoftFloat).

**Test plan**

Added new tests for 80-bit floats. Newly added `f80_ieee_{add,sub,mul,div}_test` fail and `f80_ieee_sqrt_test` times out with the old implementation, and passes with the new implementation.

**Closing issues**

Closes #4252.


DMaroo:
As a proof of concept, the previously failing test case `f80_div_test` now passes when using SoftFloat functions to calculate the result.

XVilka (commented):
If that's the only code changes necessary, not bad, not bad.

wargio (commented):
That is not bad. Small repo

wargio (commented):
It would be better to have this under rizin.

Rot127:
~~Could you push it to a `dist-fuzz-` branch, so we have an idea how many builds potentially fail?~~ https://github.com/rizinorg/rizin/tree/dist-fuzz-softfloat-integration

Builds are ok, tests fail of course. Would be for this change then.

librz/util/float/float.c
```diff
@@ -1555,157 +1556,155 @@ RZ_API RZ_OWN RzFloat *rz_float_mul_ieee_bin(RZ_NONNULL RzFloat *left, RZ_NONNUL
...
-			return rz_float_new(left->r);
-		}
-	} else {
-		if (r_is_zero) {
-			return rz_float_new_inf(left->r, sign);
-		}
+	ut16 upper = 0;
+	for (ut8 i = 0; i < 16; i++) {
+		upper <<= 1;
+		upper |= rz_bv_get(f80->s, 80 - i - 1);
```
wargio: maybe instead of doing math just iterate from 64 to 80?

librz/util/float/float.c
```diff
@@ -1555,157 +1556,155 @@ RZ_API RZ_OWN RzFloat *rz_float_mul_ieee_bin(RZ_NONNULL RzFloat *left, RZ_NONNUL
...
-	// similar to multiplication
-	if (l_exp_val == 0) {
-		// is sub-normal
-		shift_dist = rz_bv_clz(l_mantissa) - (1 + exp_len) + 1 - extra_len;
-		l_exp_val = 1 - shift_dist;
-		rz_bv_lshift(l_mantissa, shift_dist);
+	ut64 upper = 0;
+	for (ut8 i = 0; i < 64; i++) {
+		upper <<= 1;
+		upper |= rz_bv_get(f128->s, 128 - i - 1);
```
wargio: same here. 64 to 128 ?

test/unit/test_float.c
```diff
@@ -1469,6 +1482,9 @@ bool all_tests() {
 	mu_run_test(f32_ieee_fround_test);
 	mu_run_test(f32_ieee_cast_test);
 	mu_run_test(f80_round_test);
+#if (__i386__ || __x86_64__) && !__WINDOWS__
+	mu_run_test(f80_ieee_div_test);
```
wargio: add the `#if` also to the actual function.

wargio (dismissed):

test/unit/test_float.c
```diff
@@ -558,9 +558,10 @@ bool f32_ieee_special_num_test(void) {
 }
 
 bool f32_ieee_rem_test(void) {
+	/* This test should return a different result in mod. */
 	RzFloat *a1 = rz_float_new_from_f32(4.0f);
 	RzFloat *b1 = rz_float_new_from_f32(1.5f);
-	RzFloat *expect1 = rz_float_new_from_f32(1.0f);
+	RzFloat *expect1 = rz_float_new_from_f32(-0.5f);
```
wargio, line 566: this doesn't look correct

XVilka (commented):
LGTM apart from the @wargio's comment

test/unit/test_float.c
DMaroo: So, according to my understanding, the remainder function works as follows.

```
rem(x, y) = x - rem_quot * y
```

where `rem_quot` is the result of `x/y` rounded toward the nearest integral value (with halfway cases rounded toward the even number). Hence, in our case, `x = 4.0` and `y = 1.5`.

```
rem_quot = round(4.0/1.5) = round(2.66666...) = 3
rem(x, y) = 4.0 - 3 * 1.5 = -0.5
```

Source: https://stackoverflow.com/questions/25734144/difference-between-c-functions-remainder-and-fmod

wargio: Add a comment saying one is `rounded towards zero` and one `rounded nearest`

wargio: or rename the test, because is not clear.

thestr4ng3r (changes requested):
Code looks good at a first glance. Requesting changes for myself to make sure I have tested this on ppc and sparc before we merge.

DMaroo:
> Code looks good at a first glance. Requesting changes for myself to make sure I have tested this on ppc and sparc before we merge.

@thestr4ng3r Thank you! But 80-bit floats are an x87 thing and I don't think SPARC and/or PPC have them, hence the newly added 80-bit float tests wouldn't be run there. However, it is definitely worth testing whether the existing tests pass.

wargio:
> > Code looks good at a first glance. Requesting changes for myself to make sure I have tested this on ppc and sparc before we merge.
> 
> @thestr4ng3r Thank you! But 80-bit floats are an x87 thing and I don't think SPARC and/or PPC have them, hence the newly added 80-bit float tests wouldn't be run there. However, it is definitely worth testing whether the existing tests pass.

if is an endian problem or similar it could still work especially since i expect this to be software floats.

test/unit/test_float.c
```diff
@@ -1414,7 +1472,86 @@ bool f32_ieee_cast_test(void) {
 	mu_end;
 }
 
-bool f80_round_test(void) {
+bool f80_ieee_add_test(void) {
+#if (__i386__ || __x86_64__) && !__WINDOWS__
+	RzFloat *x_f80 = rz_float_new_from_f80(2.59958250082224806804l);
```
thestr4ng3r: Could those tests be written with `rz_float_new_from_bv()` instead of `from_f80`? Then they would be portable. `rz_float_new_from_f80()` could be (or is already perhaps) tested individually.

DMaroo: I see, that makes sense. I'll do this.

test/unit/test_float.c
DMaroo: You're right, I'll add a comment explaining this.

XVilka (approved):

thestr4ng3r (approved):

thestr4ng3r:
To whoever merges this, please squash and add a good commit description like "Replace RzFloat implementation by SoftFloat 3e"

PR #4550, opened 2024-06-13
Enable subproject tree-sitter-c in librz/type

Heersin:

**Detailed description**

If the tree-sitter-c library is already installed on the system, then during the linking process of the rz_type library, /usr/lib/tree-sitter-c.so will be used preferentially, which leads to multiple `undefined reference` error about `ts_*` function family. 

this pr enable the `tree-sitter-c.wrap` under subproject according to `use_sys_treesitter` option

**Test plan**

**Closing issues**


XVilka (approved):
Good catch! Since we use the modified fork, we don't need the system-wide `tree-sitter-c`. I need to update the parser btw.

PR #4552, opened 2024-06-13
Ropchain constraint syntax parser

### LINKED ISSUE #2080: ESIL to RZIL conversion

giridharprasath:
- Parse the rop constraints
- Update esil output to rzil sdb
- Add /Rg gadget command

 

Current ROP chain parser misses some gadget when compared with the ROPGadget
Asm:
```
section .text
global _start
_start:
mov dword [esp+0x0],0x80484b0
ret
```
ROPgadget output:
ScreenShot_2024-06-13_at_06:15:20-PM

Rizin output before
ScreenShot_2024-06-13_at_06:14:47-PM

Rizin output now:
ScreenShot_2024-06-13_at_06:15:54-PM

These ROP gadgets are needed for crafting better rop chains.

Partially fix: #2080 

**Detailed description**

...

**Test plan**

...

**Closing issues**

...


librz/core/cmd/cmd_search_rop.c
```diff
@@ -727,3 +730,389 @@ static void rop_classify(RzCore *core, Sdb *db, RzList /*<char *>*/ *ropList, co
 
 	free(str);
 }
+
+RZ_API void rz_rop_constraint_free(void *data) {
```
XVilka: What is not related to the command handling, I recommend moving to a separate file in `librz/core`. E.g. `librz/core/crop.c` or something like that, where the main logic could be moved to.

librz/core/cmd/cmd_search_rop.c
```diff
@@ -727,3 +730,389 @@ static void rop_classify(RzCore *core, Sdb *db, RzList /*<char *>*/ *ropList, co
...
+	RzRopConstraint *constraint = data;
+	if (constraint) {
```
XVilka: In such cases you could use the "return early" pattern and reverse the condition, it will make the code slightly easier to read.

librz/core/cmd/cmd_search_rop.c
XVilka: For all `*_free()` functions we usually allow the argument to be NULL and add a notation `RZ_NULLABLE` to that argument.

librz/core/cmd/cmd_search_rop.c
```diff
@@ -727,3 +730,389 @@ static void rop_classify(RzCore *core, Sdb *db, RzList /*<char *>*/ *ropList, co
...
+		return false;
+	}
+
+	rop_constraint->type = MOV_REG;
+	rop_constraint->args[DST_REG] = dst_reg;
+	rop_constraint->args[SRC_REG] = src_reg;
+	return true;
+}
+
+static bool parse_il_op(RzList *args, const char *str, int *idx) {
```
XVilka: This kind of code is better to be implemented in a general way and moved into `librz/il`, also covered with tests. @wargio @thestr4ng3r @DMaroo @Rot127  any thoughts?

XVilka (changes requested):

librz/core/cmd/cmd_search_rop.c
XVilka: Or into `librz/core/cil.c` where printing is already implemented.

giridharprasath: I have already started writing unit tc for these APIs

DMaroo: I don't get why we are parsing the IL as a string. Why can't we just pass an `RzILOpEffect *` for every instruction instead of parsing the stringified IL?

giridharprasath: This is not completely parsed to IL yet. I have just stringified `RzILInstructionType`, the plan is to store the registers and op as `RzILOpEffect` as you have mentioned.

librz/core/rop.c
```diff
@@ -0,0 +1,774 @@
...
+				rz_list_append(ropList, opstr_n);
+			}
+			pj_o(state->d.pj);
+			pj_kn(state->d.pj, "offset", hit->addr);
+			pj_ki(state->d.pj, "size", hit->len);
+			pj_ks(state->d.pj, "opcode", rz_asm_op_get_asm(asmop));
+			pj_ks(state->d.pj, "type", rz_analysis_optype_to_string(aop.type));
+			pj_end(state->d.pj);
+			free(buf);
+			rz_analysis_op_fini(&aop);
```
wargio: this is a lot of code within the switch. probably is better to make a static function which handles only this form of printing.

librz/core/rop.c
```diff
@@ -0,0 +1,774 @@
...
+			break;
+		case RZ_OUTPUT_MODE_QUIET:
+			// Print gadgets in a 'linear manner', each sequence on one line.
+			buf = malloc(hit->len);
+			if (!buf) {
+				goto cleanup;
+			}
```
wargio: this can be defined before the switch to avoid repetition and bugs

librz/core/cmd/cmd_search_rop.c
DMaroo: No, I would like if there wasn't any parsing at all. Rizin is a single application and it should be able to pass around data without needing to stringify it and then (re)parse it. I must be missing something, because I don't get why any sort of parsing is necessary.

librz/core/rop.c
```diff
@@ -0,0 +1,774 @@
...
+				continue;
+			}
+			if (is_end_gadget(&end_gadget, crop)) {
+#if 0
+				if (search->maxhits && rz_list_length (end_list) >= search->maxhits) {
+					// limit number of high level rop gadget results
+					rz_analysis_op_fini (&end_gadget);
+					break;
+				}
+#endif
```
wargio: if this is not used, you can remove it

librz/core/cmd/cmd_search_rop.c
Rot127: I agree with @XVilka. Better move this logic to `librz/il`.

@DMaroo The ROP gadget constraints are given as string by the user. And need to be translated to their RzIL representation for matching. @giridharprasath This was the design you had in mind, correct?

librz/core/cmd/cmd_search_rop.c
```diff
@@ -4,9 +4,11 @@
 #include <stddef.h>
 
 #include "rz_core.h"
-#include "rz_io.h"
 #include "rz_list.h"
 #include "rz_types_base.h"
+#include "rz_rop.h"
+
+#include <subprojects/capstone-next/include/capstone/x86.h>
```
Rot127:
```suggestion
#include <capstone/x86.h>
```

librz/il/meson.build
```diff
@@ -21,6 +21,7 @@ rz_il_sources = [
    'il_validate.c',
    'il_vm.c',
    'il_vm_eval.c',
+    'il_rop_gadget.c',
```
Rot127:
```suggestion
   'il_rop_gadget.c',
```

librz/include/rz_rop.h
```diff
@@ -0,0 +1,70 @@
...
+	ut64 address;
+	ut64 stack_change;
+	RzList *modified_registers;
+	RzRopMemoryOp memory_write;
+	RzRopMemoryOp memory_read;
+} RzRopGadgetInfo;
+
+typedef enum rzil_instr_type {
+	// Register to register
+	MOV_CONST, // reg <- const
```
Rot127:
```suggestion
	MOV_CONST, ///< reg <- const
```

Also below

librz/include/rz_rop.h
```diff
@@ -0,0 +1,70 @@
+// SPDX-FileCopyrightText: 2020-2021 z3phyr <giridh1337@gmail.com>
```
Rot127:
```suggestion
// SPDX-FileCopyrightText: 2024 z3phyr <giridh1337@gmail.com>
```

librz/core/cmd/cmd_search_rop.c
giridharprasath: Yes, that's right 👍

giridharprasath:
```
  0x00400565               0000  add byte [rax], al
  0x00400567           007cffff  add byte [rdi + rdi*8 - 1], bh
  0x0040056b           ff6c0000  jmp [rax + rax]
Gadget size: 10

Current /Rg output:
Gadget 0x400565
Stack change: 0x8
Changed registers: rax rdi rbx 
Register dependencies:
rax rax rax rdi rdi rdi rdi rbx rax rax 
```
TODO:
Update the changed registers based on the `init_val` and `new_val`
The dependencies list(still in progress, has to be  verbose) is memory read and write done on the registers in the gadget.
Stack change should be updated.
Show jump or syscall
Fix leaks

giridharprasath:
Current output:

```
  0x004003d0                 a6  cmpsb byte [rsi], byte ptr [rdi]
  0x004003d1               0440  add al, 0x40
  0x004003d3               00ff  add bh, bh
  0x004003d5         15160c2000  adc eax, 0x200c16
  0x004003da                 f4  hlt
Gadget size: 11
Gadget 0x4003d0
Stack change: 0x0
Changed registers: rsi rdi rax rbx 
Register dependencies:
Memory Read: rsi 5
Memory Read: rdi 4195495
Var write: rax 4195648 4195712
Var write: rbx 0 0
Var write: rbx 0 0
Var write: rax 4195648 6295958
```

Added support for stack change
Add the memory read and write to dependencies list
Track the register state in `modified_registers`

Another eg:

```
  0x0040052d                 5c  pop rsp
  0x0040052e               415d  pop r13
  0x00400530               415e  pop r14
  0x00400532               415f  pop r15
  0x00400534                 c3  ret
Gadget size: 8
Gadget 0x40052d
Stack change: 0x28
Changed registers: rsp r13 r14 r15 
Register dependencies:
```

giridharprasath:
Current output:

ScreenShot_2024-06-27_at_12:26:01-AM

## TODO:
Refactor `construct_rop_gadget` API and move the il event handling code to `Rg` handler
Tidy up the `Rg` cmd output
Fix Unit and regression TC

librz/arch/analysis.c
```diff
@@ -240,6 +240,22 @@ RZ_API char *rz_analysis_get_reg_profile(RzAnalysis *analysis) {
 		: NULL;
 }
 
+RZ_API bool rz_analysis_is_reg_in_profile(RzAnalysis *analysis, const char *str) {
```
XVilka: When you add new API function, please add Doxygen for it as well

librz/include/rz_rop.h
```diff
@@ -0,0 +1,74 @@
...
+// SPDX-License-Identifier: LGPL-3.0-only
+
+#ifndef RZ_ROP_H
+#define RZ_ROP_H
+
+#endif // RZ_ROP_H
+
+#include <rz_cmd.h>
+
+typedef struct rz_reg_info_t {
```
XVilka: When you create new include with the API, please document types (structures, enums) using Doxygen syntax as well.

XVilka: Also I recommend a file-level note what this header for

librz/include/rz_rop.h
```diff
@@ -0,0 +1,74 @@
...
+} RzRegInfo;
+
+typedef struct rz_rop_gadget_info_t {
+	ut64 address;
+	ut64 stack_change;
+	ut64 curr_pc_val;
+	bool is_pc_write;
+	bool is_syscall;
+	RzPVector /*<RzRegInfo*>*/ *modified_registers;
+	RzList /*<RzRegInfo*>*/ *dependencies;
```
XVilka: I think you need spaces in those for `rz-bindgen` linter not to complain

librz/core/cmd/cmd_search_rop.c
XVilka, line 1: Since you remove the ESIL code, please mark the PR as partially solving https://github.com/rizinorg/rizin/issues/2080

XVilka (changes requested):

test/db/cmd/cmd_rop
```diff
@@ -336,19 +478,16 @@ Gadget size: 12
   0x0000012e           0e946b04  call 0x8d6
 Gadget size: 12
 
-  0x0000013a               6de6  ldi r22, 0x6d
   0x0000013c               71e0  ldi r23, 0x01
   0x0000013e               8fe1  ldi r24, 0x1f
   0x00000140               92e0  ldi r25, 0x02
   0x00000142           0e946b04  call 0x8d6
-Gadget size: 12
+Gadget size: 10
```
giridharprasath: Updated this TC as it is a duplicate and not printed out, hence the Gadget starts from `13c`

test/db/cmd/cmd_rop
```diff
@@ -336,19 +478,16 @@ Gadget size: 12
...
 
-  0x0000015a               6fe7  ldi r22, 0x7f
-  0x0000015c               71e0  ldi r23, 0x01
   0x0000015e               8fe1  ldi r24, 0x1f
   0x00000160               92e0  ldi r25, 0x02
   0x00000162           0e946b04  call 0x8d6
-Gadget size: 12
```
giridharprasath, line 351: Same here

XVilka (changes requested):
```
Run python3 rz-bindgen/src/lint.py \
Missing type comment at <rizin/librz/core/rop.c:71:77> (token is not a comment)
Missing type comment at <rizin/librz/core/rop.c:84:77> (token is not a comment)
Missing type comment at <rizin/librz/core/rop.c:110:80> (token is not a comment)
Missing type comment at <rizin/librz/core/rop.c:149:76> (token is not a comment)
Missing type comment at <rizin/librz/core/rop.c:392:142> (token is not a comment)
Type comment at <rizin/librz/core/rop.c:509:77> for unknown type RzCoreAsmHit
Missing type comment at <rizin/librz/core/rop.c:821:102> (token is not a comment)
Missing type comment at <rizin/librz/core/cmd/cmd_search_rop.c:155:33> (token is not a comment)
Missing type comment at <rizin/librz/core/cmd/cmd_search_rop.c:382:16> (token is not a comment)
```

giridharprasath:
I think we can merge this PR(once review is over) and we can improve and add direct ROP gadget search in separate PR @XVilka @wargio

librz/core/rop.c
```diff
@@ -538,7 +538,7 @@ static void analyze_gadget(RzCore *core, RzCoreAsmHit /*<RzCoreAsmHit *>*/ *hit,
 }
 
 void print_rop_gadget_info(RzCore *core, RzRopGadgetInfo *gadget_info) {
-	rz_cons_printf("Gadget 0x%" PFMT64x "\n", gadget_info->address);
+	rz_cons_printf("\nGadget 0x%" PFMT64x "\n", gadget_info->address);
```
wargio: why the additional newline ?

giridharprasath: Separator between gadget output.
`\Rg` output:
```
Gadget 0x401004
Stack change: 0x8
Changed registers: rax rsp 
Register dependencies:
Var write: rax 128 140
Var write: rax 128 140

Gadget 0x401005
Stack change: 0x8
Changed registers: rax rsp 
Register dependencies:
Memory Read: rax 0
```

wargio: ok, it makes sense.

librz/arch/analysis.c
```diff
@@ -240,6 +240,31 @@ RZ_API char *rz_analysis_get_reg_profile(RzAnalysis *analysis) {
...
+ */
+RZ_API bool rz_analysis_is_reg_in_profile(RzAnalysis *analysis, const char *str) {
+	rz_return_val_if_fail(analysis, false);
+
+	char *reg_prof = rz_analysis_get_reg_profile(analysis);
+	if (!reg_prof) {
+		return false;
+	}
+
+	if (strstr(reg_prof, str) != NULL) {
```
wargio:
```suggestion
	if (strstr(reg_prof, str)) {
```

librz/core/cmd/cmd_search_rop.c
```diff
@@ -4,726 +4,402 @@
...
 #include "rz_types_base.h"
+#include "rz_rop.h"
 
-static RzList /*<char *>*/ *parse_list(const char *str) {
-	char *line, *data, *str_n;
-
-	if (!str) {
-		return NULL;
+static void skip_whitespace(const char *str, int *idx) {
+	while (str[*idx] == ' ' || str[*idx] == '\t' || str[*idx] == '\n' || str[*idx] == '\r') {
```
wargio:
```suggestion
	while (IS_WHITECHAR(str[*idx])) {
```

librz/core/cmd/cmd_search_rop.c
```diff
@@ -4,726 +4,402 @@
...
+		neg = 1;
+		(*idx)++;
 	}
-	if (!strcmp(op, "*=")) {
-		*result = old_src1 * src2;
-		return true;
+
+	skip_whitespace(str, idx);
+
+	if (str[*idx] == '0' && (str[*idx + 1] == 'x' || str[*idx + 1] == 'X')) {
```
wargio: this might be an out of bound read. @XVilka ?

librz/core/cmd/cmd_search_rop.c
```diff
@@ -4,726 +4,402 @@
...
+		free(dst_reg);
+		return false;
 	}
-	return false;
+
+	rop_constraint->type = MOV_CONST;
+	rop_constraint->args[DST_REG] = dst_reg;
+	rop_constraint->args[SRC_REG] = NULL;
+	char value_str[256];
+	snprintf(value_str, sizeof(value_str), "%llu", const_value);
```
wargio:
```suggestion
	rz_strf(value_str, "%" PFMT64u, const_value);
```

wargio: `%llu` is not portable, you need to use `"%" PFMT64u` which gets converted to the correct format.

librz/core/cmd/cmd_search_rop.c
```diff
@@ -4,726 +4,402 @@
...
-		if (!head) {
-			goto out_error;
-		}
-		rz_list_foreach (head, iter_dst, item_dst) {
-			ut64 diff_dst, value_dst;
-			if (!rz_list_find(reg_write, item_dst->name,
-				    (RzListComparator)strcmp, NULL)) {
-				continue;
-			}
+	RzILOpPureCode *op_ptr = malloc(sizeof(RzILOpPureCode));
```
wargio:
```suggestion
	RzILOpPureCode *op_ptr = RZ_NEW0(RzILOpPureCode);
```

librz/core/cmd/cmd_search_rop.c
```diff
@@ -4,726 +4,402 @@
...
-		if (!rz_list_find(ops_list, "=", (RzListComparator)strcmp, NULL)) {
-			goto continue_error;
-		}
+	char *src_reg = parse_register(core, str, &idx);
+	if (!src_reg) {
+		free(dst_reg);
+		return false;
+	}
+	RzList *args = rz_list_new();
+	if (!parse_il_op(args, str, &idx)) {
```
wargio, line 238:
```suggestion
	if (!args || !parse_il_op(args, str, &idx)) {
```

librz/core/cmd/cmd_search_rop.c
```diff
@@ -4,726 +4,402 @@
...
-	return mov;
-out_error:
-	FREE_ROP;
-	return NULL;
+
+	char op_str[16];
+	snprintf(op_str, sizeof(op_str), "%s", rz_il_op_pure_code_stringify(*op));
+	rop_constraint->args[OP] = strdup(op_str);
+	char value_str[256];
+	snprintf(value_str, sizeof(value_str), "%llu", const_value);
```
wargio: use `rz_strf` like above and convert that `%llu` like above.

librz/core/cmd/cmd_search_rop.c
```diff
@@ -4,726 +4,402 @@
...
+		return false;
 	}
-	return mov;
-out_error:
-	FREE_ROP;
-	return NULL;
+
+	char op_str[16];
+	snprintf(op_str, sizeof(op_str), "%s", rz_il_op_pure_code_stringify(*op));
+	rop_constraint->args[OP] = strdup(op_str);
```
wargio: there is no need to copy the output of `rz_il_op_pure_code_stringify` twice. just

```c
const char *op_str = rz_il_op_pure_code_stringify(*op);
rop_constraint->args[OP] = rz_str_dup(op_str);
```

librz/core/cmd/cmd_search_rop.c
```diff
@@ -4,726 +4,402 @@
...
+		free(dst_reg2);
+		rz_list_free(args);
+		return false;
 	}
 
-	rz_list_foreach (ropList, iter_r, esil_str) {
-		fillRegisterValues(core);
+	char op_str[16];
+	snprintf(op_str, sizeof(op_str), "%s", rz_il_op_pure_code_stringify(*op));
+	rop_constraint->args[OP] = strdup(op_str);
```
wargio: same here as above. this is useless. just follow the guidelines above.

librz/core/rop.c
```diff
@@ -0,0 +1,1248 @@
...
+		const RzILEventVarWrite *var_write = &evt->data.var_write;
+		RzBitVector *init_val = rz_il_value_to_bv(var_write->old_value);
+		RzBitVector *new_val = rz_il_value_to_bv(var_write->new_value);
+		if (!init_val || !new_val) {
+			rz_bv_free(init_val);
+			rz_bv_free(new_val);
+			break;
+		}
+		// reg_info_dup->init_val = rz_bv_to_ut64(init_val);
+		reg_info_dup->new_val = rz_bv_to_ut64(new_val);
```
wargio, line 431: this works, but is also kinda limited to 64 bits. @XVilka what do you think?

librz/core/rop.c
```diff
@@ -0,0 +1,1248 @@
...
+			break;
+		}
+		if (reg_info) {
+			break;
+		}
+		RzBitVector *val = rz_il_value_to_bv(var_read->value);
+		if (!val) {
+			break;
+		}
+		reg_info = rz_reg_info_new(core, event, rz_bv_to_ut64(val), rz_bv_to_ut64(val));
```
wargio: define a `ut64 val64 = rz_bv_to_ut64(val);` then pass `val64` as argument for `rz_reg_info_new`

wargio (changes requested):
Overall looks very well thought. just some minor changes needed for me (see comments).
That said, well done.

librz/core/cmd/cmd_search_rop.c
wargio: after re-reading the code since you are going to use `strdup` anyway, then use `rz_str_newf("%" PFMT64u, const_value);`. apply this to all the `snprintf(value_str, sizeof(value_str), "%llu", const_value);` which you have defined.

librz/core/cmd_descs/cmd_descs.c
```diff
@@ -1397,10 +1398,26 @@ static const RzCmdDescArg cmd_query_gadget_args[] = {
...
+	{
+		.name = "Gadget address",
+		.type = RZ_CMD_ARG_TYPE_STRING,
+		.flags = RZ_CMD_ARG_FLAG_LAST,
+		.optional = true,
+
+	},
+	{ 0 },
+};
+static const RzCmdDescHelp cmd_detail_gadget_help = {
```
Rot127, line 1416: Missing newline

librz/core/rop.c
```diff
@@ -0,0 +1,1248 @@
+// SPDX-FileCopyrightText: 2010-2021 z3phyr <giridh1337@gmail.com>
```
Rot127: Year should be 2024

librz/core/rop.c
```diff
@@ -0,0 +1,1248 @@
...
+#include <rz_asm.h>
+#include <rz_util/rz_log.h>
+#include <rz_util/rz_regex.h>
+#include <rz_rop.h>
+
+static bool is_end_gadget(const RzAnalysisOp *aop, const ut8 crop) {
+	if (aop->family == RZ_ANALYSIS_OP_FAMILY_SECURITY) {
+		return false;
+	}
+	switch (aop->type) {
```
Rot127, line 14: `default` case missing

Rot127: Also below

librz/core/rop.c
```diff
@@ -0,0 +1,1248 @@
...
+	}
+	const RzList *head = rz_reg_get_list(core->analysis->reg, RZ_REG_TYPE_GPR);
+	if (!head) {
+		free(reg_info);
+		return NULL;
+	}
+	RzListIter *iter_dst;
+	RzRegItem *item_dst;
+	rz_list_foreach (head, iter_dst, item_dst) {
+		if (!strcmp(name, item_dst->name) && item_dst->type == RZ_REG_TYPE_GPR) {
```
Rot127: There is a `RZ_STR_EQ` macro somewhere. Better use this one.

librz/core/rop.c
```diff
@@ -0,0 +1,1248 @@
...
+
+RzRegInfo *rz_rop_gadget_info_get_modified_register(RzRopGadgetInfo *gadget_info, const char *name) {
+	if (!gadget_info) {
+		return NULL;
+	}
+	RzRegInfo *reg_info;
+	void **it;
+	rz_pvector_foreach (gadget_info->modified_registers, it) {
+		reg_info = *it;
+		if (strcmp(reg_info->name, name) == 0) {
```
Rot127: Also here `RZ_STR_EQ`

librz/core/rop.c
```diff
@@ -0,0 +1,1248 @@
...
+		existing_reg_info->is_mem_read = new_reg_info->is_mem_read;
+		existing_reg_info->is_pc_write = new_reg_info->is_pc_write;
+		existing_reg_info->is_mem_write = new_reg_info->is_mem_write;
+		existing_reg_info->is_var_read = new_reg_info->is_var_read;
+	} else {
+		rz_pvector_push(gadget_info->modified_registers, new_reg_info);
+	}
+}
+
+RzRegInfo *rz_reg_info_dup(RzRegInfo *src) {
```
Rot127: These and other functions should be marked as `RZ_IPI` if not static.

librz/core/rop.c
```diff
@@ -0,0 +1,1248 @@
...
+	if (core->analysis->bits == 64) {
+		return !strcmp(reg_item->name, "rsp") || !strcmp(reg_item->name, "esp");
+	}
+	if (core->analysis->bits == 64 && !strcmp(core->analysis->cpu, "arm")) {
+		return !strcmp(reg_item->name, "r13");
+	}
+	return reg_item->name;
+}
+
+bool is_base_pointer(const RzCore *core, const char *name) {
```
Rot127: If only used here, please make them static. The name is pretty generic. Otherwise add a prefix to the name

librz/core/rop.c
```diff
@@ -0,0 +1,1248 @@
...
+	} break;
+	default:
+		break;
+	}
+}
+
+static void analyze_gadget(RzCore *core, const RzCoreAsmHit *hit, RzRopGadgetInfo *rop_gadget_info) {
+	if (!core->analysis) {
+		return;
+	}
```
Rot127: Shouldn't those checks be `rz_return_if_fail()`? Or is `core->analysis` being NULL a legal value? Also above.

librz/core/rop.c
```diff
@@ -0,0 +1,1248 @@
...
+		ret = false;
+		goto bad;
+	}
+	// TODO full address cannot be represented, shrink 1 byte to [0, UT64_MAX)
+	if (search_from == UT64_MAX && search_to == UT64_MAX) {
+		search_itv.addr = 0;
+		search_itv.size = UT64_MAX;
+	}
+
+	Sdb *gadgetSdb = NULL;
```
Rot127: Did not review far enough, but is a `Sdb` really necessary here? We want to remove it soon. So using it at a place like this is not a good idea. Would be a `ht_ss` enough?

librz/core/cmd/cmd_search_rop.c
XVilka: Right. Also `idx` pointer should be marked as `RZ_NONNULL` and have an assert.

librz/core/rop.c
XVilka: For now it's ok, but could be improved as more architectures will be added for working with the upcoming ROP generator.

librz/core/rop.c
```diff
@@ -0,0 +1,1248 @@
...
+		rz_list_free(end_list);
+		RZ_LOG_ERROR("core: ROP length (rop.len) must be greater than 1.\n");
+		if (max_instr == 1) {
+			RZ_LOG_ERROR("core: For rop.len = 1, use /c to search for single "
+				     "instructions. See /c? for help.\n");
+		}
+		return false;
+	}
+
+	if (!strcmp(arch, "mips")) { // MIPS has no jump-in-the-middle
```
Rot127: RZ_STR_EQ also below

librz/core/rop.c
```diff
@@ -0,0 +1,1248 @@
...
+	if (grep_arg && regexp) {
+		if (!rx_list) {
+			rx_list = rz_list_newf(free);
+		}
+		gregexp = strdup(grep_arg);
+		tok = strtok(gregexp, ";");
+		while (tok) {
+			rx = strdup(tok);
+			rz_list_append(rx_list, rx);
+			tok = strtok(NULL, ";");
```
Rot127: I don't get this line, isn't `tok` always `NULL` after a single iteration?

librz/core/rop.c
```diff
@@ -0,0 +1,1248 @@
...
+		if (!rz_itv_overlap(search_itv, map->itv)) {
+			continue;
+		}
+		RzInterval itv = rz_itv_intersect(search_itv, map->itv);
+		ut64 from = itv.addr, to = rz_itv_end(itv);
+		if (rz_cons_is_breaked()) {
+			break;
+		}
+		delta = to - from;
+		buf = calloc(1, delta);
```
Rot127: calloc -> `RZ_NEW0`

librz/core/rop.c
```diff
@@ -0,0 +1,1248 @@
...
+		if (!buf) {
+			result = false;
+			goto bad;
+		}
+		(void)rz_io_read_at(core->io, from, buf, delta);
```
Rot127: Handle errors in read here.

librz/core/rop.c
```diff
@@ -0,0 +1,1248 @@
...
+
+		// Find the end gadgets.
+		for (i = 0; i < delta; i += increment) {
```
Rot127:
```suggestion
		for (size_t i = 0; i < delta; i += increment) {
```

No need to define it above. Always has the risk that it is not reset properly.

librz/core/rop.c
Rot127: Specifically check the `read_at` and `read_at_mapped` (or a similar name). You want the `mapped` version here and don't search for gadgets in non-mapped regions.

librz/core/rop.c
```diff
@@ -0,0 +1,1248 @@
...
+			// Disassemble one.
+			rz_analysis_op_init(&end_gadget);
+			if (rz_analysis_op(core->analysis, &end_gadget, from + i, buf + i,
+				    delta - i, RZ_ANALYSIS_OP_MASK_BASIC) < 1) {
+				rz_analysis_op_fini(&end_gadget);
+				continue;
+			}
+			if (is_end_gadget(&end_gadget, crop)) {
+				RzRopEndListPair *epair = RZ_NEW0(RzRopEndListPair);
+				if (epair) {
```
Rot127: You can do a `rz_return_if_fail(epair)` and save indentation. If `calloc` fails everything is over anyways.

librz/core/rop.c
Rot127: Can this whole loop be move to it's own `static` function? It would make the code way more readable.

librz/core/rop.c
```diff
@@ -0,0 +1,1248 @@
...
+			rz_analysis_op_fini(&end_gadget);
+			if (rz_cons_is_breaked()) {
+				break;
+			}
+			// Right now we have a list of all of the end/stop gadgets.
+			// We can just construct gadgets from a little bit before them.
+		}
+		rz_list_reverse(end_list);
+		// If we have no end gadgets, just skip all of this search nonsense.
+		if (!rz_list_empty(end_list)) {
```
Rot127: This whole `if` statement is quite deep (meaning many indentations). This is hard to read. I would suggest to:

- Move some code to their own small `static` functions
- Add one or two clean-up gotos, so you can do error handling like `if (somethign_invalid) { goto cleanup }`. This is one of the still valid `goto` use cases in C.

Before you refactor with `gotos` check again that actually everything is freed properly in case of error.
Then add them. Added suggestions below.

librz/core/rop.c
```diff
@@ -0,0 +1,1248 @@
...
+			prev = 0;
+			// Start at just before the first end gadget.
+			for (i = 0; i < delta && max_count; i += increment) {
+				if (increment == 1) {
+					// give in-boundary instructions a shot
+					if (i < prev - max_inst_size_x86) {
+						i = prev - max_inst_size_x86;
+					}
+				} else {
+					if (i < prev) {
```
Rot127: `else if`

librz/core/rop.c
```diff
@@ -0,0 +1,1248 @@
...
+						i = prev;
+					}
+				}
+				if (i < 0) {
+					i = 0;
+				}
```
Rot127: In what case is `i` smaller then 0?

librz/core/rop.c
```diff
@@ -0,0 +1,1248 @@
...
+	RzListIter *itermap = NULL;
+	char *grep_arg = NULL;
+	char *tok, *gregexp = NULL;
+	char *rx = NULL;
+	RzAsmOp *asmop = NULL;
+	RzList *boundaries = NULL;
+	int delta = 0;
+	ut8 *buf;
+	RzIOMap *map;
+	HtSU *unique_hitlists = ht_su_new(HT_STR_DUP);
```
Rot127: Quite a lot of those declarations can be made in place with definition. There is no need for C89 compatibility. And this pattern makes code harder to read and is error prone.

librz/core/rop.c
```diff
@@ -0,0 +1,1248 @@
...
+						rz_asm_op_free(asmop);
+						asmop = NULL;
+						rz_strbuf_free(sb);
+						continue;
+					}
+					if (align && 0 != (from + i) % align) {
+						rz_asm_op_free(asmop);
+						asmop = NULL;
+						rz_strbuf_free(sb);
+						continue;
```
Rot127: These things would belong to a `goto` block

Rot127: It repeats below again.

librz/core/rop.c
```diff
@@ -0,0 +1,1248 @@
...
+						ht_su_insert(unique_hitlists, asm_op_hex, 1);
+					} else {
+						rz_list_free(hitlist);
+						rz_asm_op_free(asmop);
+						asmop = NULL;
+						rz_strbuf_free(sb);
+						continue;
+					}
+					rz_strbuf_free(sb);
+					if (gadgetSdb) {
```
Rot127: This check should have been handled before. It doesn't make sense that the db is undefined, but the logic went on so far, doesn't it?

librz/core/rop.c
```diff
@@ -0,0 +1,1248 @@
...
+						end_gadget = (RzRopEndListPair *)rz_list_pop(end_list);
+						next = end_gadget->instr_offset;
+						i = next - ropdepth;
+						if (i < 0) {
+							i = 0;
+						}
+					} else {
+						end_gadget = NULL;
+						break;
+					}
```
Rot127: Could this be maybe move to a `set_next_end_gadget()`?

librz/core/rop.c
```diff
@@ -0,0 +1,1248 @@
...
+ * \param core Pointer to the RzCore object.
+ * \param input Input string for ROP search.
+ * \param state Pointer to the command state output.
+ * \return RZ_CMD_STATUS_OK on success.
+ *
+ * Displays ROP gadgets from the gadgetSdb.
+ * If unavailable, performs a ROP search with the input.
+ */
+RZ_API RzCmdStatus rz_core_rop_gadget_info(RzCore *core, const char *input, RzCmdStateOutput *state) {
+	Sdb *gadgetSdb = sdb_ns(core->sdb, "gadget_sdb", false);
```
Rot127: Please check also here if a `ht_ss` wouldn't be enough. @XVilka How forbidden is `SDB` usage these days?

librz/core/rop.c
```diff
@@ -0,0 +1,1248 @@
...
+
+	if (!gadgetSdb) {
+		rz_core_search_rop(core, input, 0, RZ_ROP_GADGET_PRINT, state);
+		return RZ_CMD_STATUS_OK;
```
Rot127: Shouldn't this be an error code?

librz/core/rop.c
```diff
@@ -0,0 +1,1248 @@
...
+		if (subchain) {
+			do {
+				print_rop(core, hitlist, state);
+				hitlist->head = hitlist->head->next;
+			} while (hitlist->head->next);
+		} else {
+			print_rop(core, hitlist, state);
+		}
+	}
+	if (type & RZ_ROP_GADGET_DETAIL) {
```
Rot127: Negate check and return early.

librz/core/rop.c
```diff
@@ -0,0 +1,1248 @@
...
+	if (regex && rx) {
+		rz_list_free(hitlist);
+		ht_uu_free(localbadstart);
+		return NULL;
+	}
+	if (!valid || (grep && end)) {
+		rz_list_free(hitlist);
+		ht_uu_free(localbadstart);
+		return NULL;
+	}
```
Rot127: Combine these two if conditions. Saves a few lines of duplicate code.

librz/core/rop.c
```diff
@@ -0,0 +1,1248 @@
...
+		if (error < 0 || (nb_instr == 0 && aop.type == RZ_ANALYSIS_OP_TYPE_NOP)) {
+			valid = false;
+			goto ret;
+		}
+
+		if (is_end_gadget(&aop, 0)) {
+			end_gadget_cnt++;
+		}
+		const int opsz = aop.size;
+		// opsz = rz_strbuf_length (asmop.buf);
```
Rot127:
```suggestion
```

librz/core/rop.c
```diff
@@ -0,0 +1,1248 @@
...
+		char *opst = aop.mnemonic;
+		RzAsmOp asmop;
```
Rot127:
```suggestion
		RzAsmOp asmop = RZ_EMPTY;
```

librz/core/rop.c
```diff
@@ -0,0 +1,1248 @@
...
+		int asm_ret = rz_asm_disassemble(core->rasm, &asmop, buf + idx, buflen - idx);
+		if (!opst) {
```
Rot127: Because you passed `RZ_ANALYSIS_OP_MASK_DISASM` to the `rz_analysis_op` it already called `rz_asm_disassemble` and if it succeeded the `opst` is definitely set. So this test could be skipped.

librz/core/rop.c
```diff
@@ -0,0 +1,1248 @@
...
+			opst = strdup(rz_asm_op_get_asm(&asmop));
+		}
+		if (!rz_str_ncasecmp(opst, "invalid", strlen("invalid")) ||
+			!rz_str_ncasecmp(opst, ".byte", strlen(".byte")) || end_gadget_cnt > 1) {
+			valid = false;
+			goto ret;
+		}
+
+		hit = rz_core_asm_hit_new();
+		if (hit) {
```
Rot127: `rz_return_if_fail()` and remove `if (hit)`. If initialization fails, there is no need to continue.

librz/core/rop.c
```diff
@@ -0,0 +1,1248 @@
...
+		if (asm_ret >= 0) {
+			rz_asm_op_fini(&asmop);
+		}
+
+		// Move on to the next instruction
+		idx += opsz;
+		addr += opsz;
+		if (rx) {
+			grep_find = rz_regex_contains(rx, opst, RZ_REGEX_ZERO_TERMINATED, RZ_REGEX_EXTENDED, RZ_REGEX_DEFAULT);
+			search_hit = (end && grep && grep_find);
```
Rot127: Define `search_hit` over the `if (rx)` as `bool search_hit`. It is defined as `int` above, but this gives the wrong idea what the variable does.

librz/core/rop.c
```diff
@@ -0,0 +1,1248 @@
...
+			end = start + strlen(grep);
+		}
+		grep_str = calloc(1, end - start + 1);
+		strncpy(grep_str, start, end - start);
+		if (regex) {
+			// get the first regexp.
+			if (rz_list_length(rx_list) > 0) {
+				rx = rz_list_get_n(rx_list, count++);
+			}
+		}
```
Rot127: Move grep parsing to static function.

librz/core/rop.c
```diff
@@ -0,0 +1,1248 @@
...
+	}
+
+	bool found;
```
Rot127:
```suggestion
	bool found = false;
```

librz/core/rop.c
```diff
@@ -0,0 +1,1248 @@
...
+	RzList *hitlist = rz_core_asm_hit_list_new();
+	ut8 nb_instr = 0;
+	const ut8 max_instr = rz_config_get_i(core->config, "rop.len");
+	bool valid = false;
+	int grep_find;
+	int search_hit;
+	char *rx = NULL;
+	HtUUOptions opt = { 0 };
+	HtUU *localbadstart = ht_uu_new_opt(&opt);
+	int count = 0;
```
Rot127: Please define variables in place.

librz/core/rop.c
```diff
@@ -0,0 +1,1248 @@
...
+	rz_list_free(ropList);
+}
+
+static bool insert_into(void *user, const ut64 k, const ut64 v) {
+	HtUU *ht = (HtUU *)user;
+	ht_uu_insert(ht, k, v);
+	return true;
+}
+
+// TODO: follow unconditional jumps
```
Rot127: Don't forget to open an issue about this one. Would suggest to remove the comment, and add this as a task to the PR description. After merge you can open the issue.

librz/core/rop.c
```diff
@@ -0,0 +1,1248 @@
...
+		if (reg_info->is_var_write) {
+			rz_cons_printf("Var write: %s %llu %llu\n", reg_info->name, reg_info->init_val, reg_info->new_val);
+		} else if (reg_info->is_mem_read) {
+			rz_cons_printf("Memory Read: %s %llu\n", reg_info->name, reg_info->new_val);
+		} else if (reg_info->is_mem_write) {
+			rz_cons_printf("Memory Write: %s %llu %llu\n", reg_info->name, reg_info->init_val, reg_info->new_val);
+		}
+	}
+}
+static void print_rop(const RzCore *core, const RzList /*<RzCoreAsmHit *>*/ *hitlist, RzCmdStateOutput *state) {
```
Rot127: Missing newline

librz/core/rop.c
```diff
@@ -0,0 +1,1248 @@
...
+				rz_bv_free(new_val);
+				break;
+			}
+			reg_info = rz_reg_info_new(core, event, rz_bv_to_ut64(old_val),
+				rz_bv_to_ut64(new_val));
+			rz_rop_gadget_info_add_register(gadget_info, reg_info, is_dependency);
+			rz_bv_free(old_val);
+			rz_bv_free(new_val);
+		}
+	} break;
```
Rot127, line 536: Move the breaks within the `{}`

librz/core/rop.c
```diff
@@ -0,0 +1,1248 @@
...
+			if (!old_val || !new_val) {
+				rz_bv_free(old_val);
+				rz_bv_free(new_val);
+				break;
+			}
+			reg_info = rz_reg_info_new(core, event, rz_bv_to_ut64(old_val),
+				rz_bv_to_ut64(new_val));
+			rz_rop_gadget_info_add_register(gadget_info, reg_info, is_dependency);
+			rz_bv_free(old_val);
+			rz_bv_free(new_val);
```
Rot127: You can invert the `if (!old_val || !new_val)` condition and do 
```c
rz_bv_free(old_val);
rz_bv_free(new_val);
```

only once.

Rot127: Above as well.

librz/include/rz_rop.h
```diff
@@ -0,0 +1,100 @@
...
+typedef struct rz_reg_info_t {
+	char *name;
+	bool is_mem_read; ///< Register involved in Memory read.
+	bool is_pc_write; ///< PC write flag.
+	bool is_var_read; ///< Register involved in Variable read.
+	bool is_var_write; ///< Register involved in Variable write.
+	bool is_mem_write; ///< Register involved in Memory write.
+	ut64 init_val;
+	ut64 new_val;
+} RzRegInfo;
```
Rot127:
```suggestion
} RzRopRegInfo;
```

Otherwise it follows the naming in `rz_reg.h`.
Alos please document the last three fields.

librz/include/rz_rop.h
```diff
@@ -0,0 +1,100 @@
...
+/**
+ * \brief Types of IL instructions for ROP constraints.
+ */
+typedef enum rzil_instr_type {
+	MOV_CONST, ///< reg <- const
+	MOV_REG, ///< reg <- reg
+	MOV_OP_CONST, ///< reg <- reg OP const
+	MOV_OP_REG, ///< reg <- reg OP reg
+	SYSCALL, ///< syscall
+} RzILInstructionType;
```
Rot127: Naming is off here as well. Please use a `RzRop` prefix

librz/include/rz_rop.h
```diff
@@ -0,0 +1,100 @@
...
+} RzRopConstraint;
+
+// Command APIs
+RZ_API int rz_core_search_rop(RzCore *core, const char *greparg, int regexp, RzRopRequestMask type, RzCmdStateOutput *state);
+RZ_API RzCmdStatus rz_core_rop_gadget_info(RzCore *core, const char *input, RzCmdStateOutput *state);
+RZ_API bool analyze_constraint(RzCore *core, char *str, RzRopConstraint *rop_constraint);
+
+// ROP Constraint APIs
+RZ_API void rz_rop_constraint_free(RZ_NULLABLE void *data);
+RZ_API RzList /*<RzRopConstraint *>*/ *rz_rop_constraint_list_new(void);
```
Rot127: Naming should be also `rz_core_<something with rop>()`

test/unit/test_rop_constraint.c
```diff
@@ -0,0 +1,142 @@
+// SPDX-FileCopyrightText: 2021 z3phyr <giridh1337@gmail.com>
```
Rot127: 2024

librz/include/rz_rop.h
```diff
@@ -0,0 +1,100 @@
...
+ */
+typedef struct rz_rop_constraint_t {
+	RzILInstructionType type; ///< IL instruction type.
+	char *args[NUM_ARGS]; ///< Arguments.
+} RzRopConstraint;
+
+// Command APIs
+RZ_API int rz_core_search_rop(RzCore *core, const char *greparg, int regexp, RzRopRequestMask type, RzCmdStateOutput *state);
+RZ_API RzCmdStatus rz_core_rop_gadget_info(RzCore *core, const char *input, RzCmdStateOutput *state);
+RZ_API bool analyze_constraint(RzCore *core, char *str, RzRopConstraint *rop_constraint);
```
Rot127: Prefix in name missing.

librz/core/rop.c
```diff
@@ -0,0 +1,1248 @@
...
+		}
+	} break;
+	case RZ_IL_EVENT_MEM_WRITE: {
+		while (!rz_pvector_empty(vec)) {
+			RzILEvent *evt = rz_pvector_pop(vec);
+			fill_rop_gadget_info_from_events(core, gadget_info, event, evt, vec, true);
+		}
+	} break;
+	case RZ_IL_EVENT_PC_WRITE: {
+		// RzILEventPCWrite *pc_write = &event->data.pc_write;
```
Rot127:
```suggestion
```

librz/core/rop.c
```diff
@@ -0,0 +1,1248 @@
...
+		return !strcmp(reg_item->name, "ebp");
+	}
+	if (core->analysis->bits == 64) {
+		return !strcmp(reg_item->name, "rbp") || !strcmp(reg_item->name, "ebp");
+	}
+	if (core->analysis->bits == 64 && !strcmp(core->analysis->cpu, "arm")) {
+		return !strcmp(reg_item->name, "r11");
+	}
+	return reg_item->name;
+}
```
Rot127: These `is_stack/base_ptr` should query the register profile and compare the register name against the `SP`/`BP` defined there.
Otherwise this code is only valid for `x86` and `arm`.
Also, returning `reg_item->name` as `bool` is not the way we do it here :)

Rot127: Check out `rz_reg_role_by_name()` (or adjacent functions in `rz_reg.h`. These two functions become way smaller this way as well.

Rot127 (changes requested):
In general really nice! I have a few remarks though (don't be too discouraged by the number of comments, most of them are suggestions and syntax stuff).

- My highest priority change request is: Please definitely refactor `rz_core_search_rop` and `construct_rop_gadget`. They are very long and have a lot of indentations. This is hard to read, maintain and debug. I left you some suggestions how to refactor. In general though: With so many variables in each function, I'd suggest to think about, what steps can be moved to separated `static` functions. This will help everyone enormously in the future to maintain this code and find bugs.
- Please add `RZ_OWN`/`RZ_BORROW` to at least the `API` functions. And to the longer `IPI` functions. And definitely if a parameter is freed by a function.
- Then there are a lot of code duplicates in error handling (`fee()` mostly). Please try remove them by refactoring the code. I made some suggestions.
- Besides, there is really no need to define all variables at the beginning of functions. Rizin is not a C89 project (and if it is and I just didn't see the memo, it definitely doesn't work :D ).

Awesome job though!

librz/core/rop.c
Rot127: But please let's do the `instruction word` abstraction first. It is a pain to work with Hexagon and these kind of analysis algos.

librz/core/cmd_descs/cmd_descs.c
wargio: i believe this is auto-gen.

giridharprasath:
Phew :sweat_smile:  Thanks for the review, @Rot127  and @wargio . I didn't refactor `rz_core_search_rop` and `construct_rop_gadget` because I was concerned about potentially introducing bugs. I will definitely try to refactor both of these and address the other comments.

Rot127:
> I was concerned about potentially introducing bugs. 

Then you should refactor as well :)
It is way easier to find bugs, if everything is brief and concise. Because less stuff to look at and less stuff to remember.

librz/core/rop.c
wargio: i think that step should be handled by the disassembler (RzArch) and should be defined by the arch, so we could have a packet with 1 or N instructions.

test/integration/test_project_migrate.c
```diff
@@ -1020,6 +1028,7 @@ int all_tests() {
 	mu_run_test(test_load_v15_seek_history);
 	mu_run_test(test_load_v15_str_config);
 	mu_run_test(test_load_v16);
+	mu_run_test(test_load_v17);
```
XVilka, line 1049: You need to add a test also for a migration itself to the v17

librz/core/rop.c
```diff
@@ -0,0 +1,1344 @@
...
+RZ_API void rz_core_rop_gadget_info_free(RzRopGadgetInfo *gadget_info) {
+	if (!gadget_info) {
+		return;
+	}
+	rz_pvector_free(gadget_info->modified_registers);
+	rz_list_free(gadget_info->dependencies);
+	free(gadget_info);
+}
+
+RZ_API void rz_core_rop_gadget_info_add_register(RzRopGadgetInfo *gadget_info, RzRopRegInfo *reg_info, const bool is_dependency) {
```
XVilka: For every `RZ_API` that you add or change, except maybe obvious basic `*_free()` and `*_new()` functions, please add a Doxygen documentation. Moreover, for the external API it's inmportant to provide hints like `RZ_BORROW`, `RZ_OUT`, `RZ_NULLABLE`, `RZ_NONNULL` to the corresponding parameters, and for those marked with `RZ_NONNULL` add `rz_return_*()` asserts at the beginning of the function.

librz/core/rop.c
```diff
@@ -0,0 +1,1344 @@
...
+	if (!gadget_info || !reg_info) {
+		return;
+	}
+	if (!is_dependency) {
+		rz_pvector_push(gadget_info->modified_registers, reg_info);
+	}
+}
+
+RZ_API RzRopRegInfo *rz_core_rop_gadget_info_get_modified_register(RzRopGadgetInfo *gadget_info, const char *name) {
```
XVilka: If function returns pointer - add an onwership (`RZ_OWN`/`RZ_BORROW`) as appropriate.

librz/core/rop.c
XVilka: @DMaroo @thestr4ng3r do you have an opinion on this?

XVilka (changes requested):

librz/core/rop.c
giridharprasath: Noted :+1:

librz/core/rop.c
giridharprasath: I will be opening a new issue for tracking this.

librz/core/rop.c
giridharprasath: `RZ_NEW0` is for struct right, retaining this

librz/core/rop.c
giridharprasath: Changed this to different logic

librz/core/rop.c
giridharprasath: I will be opening an issue and finish that in a separate PR

XVilka:
```
librz/include/rz_project.h:64:84: error: code should be clang-formatted [-Wclang-format-violations]
RZ_API bool rz_project_migrate_v16_v17(RzProject *prj, RzSerializeResultInfo *res);
```

test/prj/v15-seek-history.rzdb
```diff
@@ -55,7 +55,7 @@ prj.file=./test/prj/v15-seek-history.rzdb
 /core/debug/breakpoints
 
 /core/file
-relative=../bins/elf/ls
+relative=../bins/elf/crackme0x05
```
XVilka: Old test projects shouldn't be changed

XVilka (changes requested):

librz/core/rop.c
```diff
@@ -0,0 +1,1238 @@
...
+		char *asmop_hex = rz_asm_op_get_hex(asmop);
+		*asmop_hex_str = rz_str_append(*asmop_hex_str, asmop_hex);
+		free(asmop_hex);
+	}
+
+	free(buf);
+	return 0;
+}
+
+static int rz_rop_print_table_mode(const RzCore *core, const RzCoreAsmHit *hit, const RzList /*<RzCoreAsmHit *>*/ *hitlist,
```
XVilka: This one could be converted to returning `bool` since only two values are possible

librz/core/rop.c
```diff
@@ -0,0 +1,1238 @@
...
+	case RZ_ANALYSIS_OP_TYPE_CCALL:
+	case RZ_ANALYSIS_OP_TYPE_UCCALL:
+	case RZ_ANALYSIS_OP_TYPE_CRET:
+		return true;
+	default:
+		return false;
+	}
+}
+
+static int rz_rop_process_asm_op(const RzCore *core, const RzCoreAsmHit *hit, RzAsmOp *asmop, RzAnalysisOp *aop, unsigned int *size, char **asmop_str, char **asmop_hex_str) {
```
XVilka: Same

librz/core/rop.c
```diff
@@ -0,0 +1,1238 @@
...
+	const ut64 addr_last = ((RzCoreAsmHit *)rz_list_last(hitlist))->addr;
+	if (addr_last != hit->addr) {
+		*asmop_str = rz_str_append(*asmop_str, "; ");
+	}
+	rz_asm_op_free(asmop);
+	rz_analysis_op_fini(&aop);
+	return 0;
+}
+
+static int rz_rop_print_quiet_mode(const RzCore *core, const RzCoreAsmHit *hit, unsigned int *size, const bool colorize) {
```
XVilka: Same

librz/core/rop.c
```diff
@@ -0,0 +1,1238 @@
...
+		rz_strbuf_free(bw_str);
+	} else {
+		rz_cons_printf(" %s;", rz_asm_op_get_asm(asmop));
+	}
+	rz_asm_op_free(asmop);
+	rz_analysis_op_fini(&aop);
+	return 0;
+}
+
+static int rz_rop_print_standard_mode(const RzCore *core, const RzCoreAsmHit *hit,
```
XVilka: Same

librz/core/rop.c
```diff
@@ -0,0 +1,1238 @@
...
+				hit->addr, asm_op_hex, rz_asm_op_get_asm(asmop));
+		}
+	}
+	free(asm_op_hex);
+	rz_analysis_op_fini(&aop);
+	rz_asm_op_free(asmop);
+	return 0;
+}
+
+static int rz_rop_print_json_mode(const RzCore *core, const RzCoreAsmHit *hit, unsigned int *size, PJ *pj) {
```
XVilka: Same

XVilka (changes requested):
Looks better! Only few small nitpicks left, apart from that - LGTM.

XVilka (approved):
LGTM. Next time please use different branch name, by the way.

librz/core/cconfig.c
```diff
@@ -3735,8 +3735,7 @@ RZ_API int rz_core_config_init(RzCore *core) {
 
 	/* rop */
 	SETI("rop.len", 5, "Maximum ROP gadget length");
-	SETBPREF("rop.sdb", "false", "Cache results in sdb (experimental)");
-	SETBPREF("rop.db", "true", "Categorize rop gadgets in sdb");
+	SETBPREF("rop.cache", "false", "Cache rop gadget results(experimental)");
```
wargio, line 3738:
```suggestion
	SETBPREF("rop.cache", "false", "Cache rop gadget results (experimental)");
```

librz/arch/analysis.c
```diff
@@ -240,6 +243,31 @@ RZ_API char *rz_analysis_get_reg_profile(RzAnalysis *analysis) {
...
+ * \brief Check if a register is in the analysis profile.
+ * \param analysis Pointer to the RzAnalysis object.
+ * \param str The register name to check.
+ * \return true if the register name is found, false otherwise.
+ *
+ * This function checks if the given register name is present
+ * in the register profile of the given RzAnalysis.
+ */
+RZ_API bool rz_analysis_is_reg_in_profile(RzAnalysis *analysis, const char *str) {
+	rz_return_val_if_fail(analysis, false);
```
wargio: Add `RZ_NONNULL` to both args and change the assert to `rz_return_val_if_fail(analysis && str, false);`

librz/core/cmd/cmd_search_rop.c
```diff
@@ -4,726 +4,457 @@
 #include <stddef.h>
```
wargio, line 5: Add your name as Copyright by adding a new `// SPDX-FileCopyrightText: ` on top of the old one.

librz/core/cmd/cmd_search_rop.c
```diff
@@ -4,726 +4,457 @@
...
-	if (!strcmp(op, ">>")) {
-		*result = src1 >> src2;
-		return true;
+
+	rop_constraint->type = MOV_CONST;
+	rop_constraint->args[DST_REG] = dst_reg;
+	rop_constraint->args[SRC_REG] = NULL;
+	char value_str[256];
+	rz_strf(value_str, "%" PFMT64u, const_value);
+	rop_constraint->args[SRC_CONST] = strdup(value_str);
```
wargio: use `rz_str_newf` directly as `rop_constraint->args[SRC_CONST] = rz_str_newf("%" PFMT64u, const_value);`

librz/core/rop.c
wargio: i think a comment here is enough. just say that this is limited to max 64 bit val but it could be expanded in the future.

librz/core/rop.c
```diff
@@ -0,0 +1,1238 @@
...
+}
+
+static void set_increment_based_on_arch(const RzCore *core, const char *arch, int *increment) {
+	if (RZ_STR_EQ(arch, "mips")) { // MIPS has no jump-in-the-middle
+		*increment = 4;
+	} else if (RZ_STR_EQ(arch, "arm")) { // ARM has no jump-in-the-middle
+		*increment = rz_config_get_i(core->config, "asm.bits") == 16 ? 2 : 4;
+	} else if (RZ_STR_EQ(arch, "avr")) { // AVR is halfword aligned.
+		*increment = 2;
+	}
```
wargio, line 971: @XVilka it would be nice to have an issue for this. seems that this should be an info that RzArch should be able to provide.

librz/include/rz_rop.h
```diff
@@ -0,0 +1,136 @@
+// SPDX-FileCopyrightText: 2024 z3phyr <giridh1337@gmail.com>
+// SPDX-License-Identifier: LGPL-3.0-only
+
+#ifndef RZ_ROP_H
+#define RZ_ROP_H
+
+#include <rz_cmd.h>
+
```
wargio: Add the cpp barriers.

```c
#ifdef __cplusplus
extern "C" {
#endif
```

librz/include/rz_rop.h
```diff
@@ -0,0 +1,136 @@
...
+RZ_API void rz_core_rop_gadget_info_free(RZ_NULLABLE RzRopGadgetInfo *gadget_info);
+RZ_API void rz_core_rop_gadget_info_add_register(const RZ_NONNULL RZ_OUT RzRopGadgetInfo *gadget_info,
+	RZ_NONNULL RzRopRegInfo *reg_info, bool is_dependency);
+RZ_API int rz_core_rop_gadget_info_update_register(RZ_INOUT RzRopGadgetInfo *gadget_info, RZ_NONNULL RzRopRegInfo *new_reg_info);
+RZ_API RZ_OWN RzRopGadgetInfo *rz_core_rop_gadget_info_new(ut64 address);
+RZ_IPI RzRopRegInfo *rz_core_rop_reg_info_dup(RzRopRegInfo *src);
+RZ_IPI void rz_core_rop_reg_info_free(RzRopRegInfo *reg_info);
+RZ_IPI RzRopRegInfo *rz_core_rop_reg_info_new(const RzCore *core, const RzILEvent *evt, ut64 init_val, ut64 new_val);
+RZ_BORROW RZ_API RzRopRegInfo *rz_core_rop_gadget_info_get_modified_register(const RZ_NONNULL RzRopGadgetInfo *gadget_info, RZ_NONNULL const char *name);
+
```
wargio, line 139: Add the cpp barrier.

```c
#ifdef __cplusplus
}
#endif
```

wargio (changes requested):
Huge improvements. fix the comments i left and i think we can merge.

XVilka (approved):

librz/include/rz_rop.h
```diff
@@ -0,0 +1,141 @@
...
+#include <rz_cmd.h>
+#ifdef __cplusplus
+extern "C" {
+#endif
+/**
+ * \file rz_rop.h
+ * \brief Return-Oriented Programming (ROP) related APIs and structures..
+ *
+ * This file contains definitions, structures, and function prototypes for handling ROP gadgets and constraints.
+ */
```
wargio:
```suggestion
/**
 * \file rz_rop.h
 * \brief Return-Oriented Programming (ROP) related APIs and structures..
 *
 * This file contains definitions, structures, and function prototypes for handling ROP gadgets and constraints.
 */

#include <rz_cmd.h>

#ifdef __cplusplus
extern "C" {
#endif
```

librz/core/cmd/cmd_search_rop.c
```diff
@@ -1,729 +1,458 @@
-// SPDX-FileCopyrightText: 2009-2016 Alexandru Caciulescu <alex.darredevil@gmail.com>
+// SPDX-FileCopyrightText: 2024 z3phyr <giridh1337@gmail.com>
```
wargio, line 1:
```suggestion
// SPDX-FileCopyrightText: 2024 z3phyr <giridh1337@gmail.com>
// SPDX-FileCopyrightText: 2009-2016 Alexandru Caciulescu <alex.darredevil@gmail.com>
```

librz/arch/analysis.c
```diff
@@ -240,6 +243,31 @@ RZ_API char *rz_analysis_get_reg_profile(RzAnalysis *analysis) {
...
+		return false;
+	}
+
+	if (strstr(reg_prof, str)) {
+		free(reg_prof);
+		return true;
+	}
+	free(reg_prof);
+	return false;
+}
```
wargio:
```suggestion
 * \param name The register name to check.
 * \return true if the register name is found, false otherwise.
 *
 * This function checks if the given register name is present
 * in the register profile of the given RzAnalysis.
 */
RZ_API bool rz_analysis_is_reg_in_profile(RZ_NONNULL RzAnalysis *analysis, RZ_NONNULL const char *name) {
	rz_return_val_if_fail(analysis && name, false);

	char *reg_prof = rz_analysis_get_reg_profile(analysis);
	if (!reg_prof) {
		return false;
	}

	if (strstr(reg_prof, name)) {
		free(reg_prof);
		return true;
	}
	free(reg_prof);
	return false;
}
```

librz/include/rz_analysis.h
```diff
@@ -1573,6 +1574,7 @@ RZ_API int rz_analysis_archinfo(RzAnalysis *analysis, RzAnalysisInfoType query);
 RZ_API bool rz_analysis_use(RzAnalysis *analysis, const char *name);
 RZ_API bool rz_analysis_set_reg_profile(RzAnalysis *analysis);
 RZ_API char *rz_analysis_get_reg_profile(RzAnalysis *analysis);
+RZ_API bool rz_analysis_is_reg_in_profile(RzAnalysis *analysis, const char *str);
```
wargio:
```suggestion
RZ_API bool rz_analysis_is_reg_in_profile(RZ_NONNULL RzAnalysis *analysis, RZ_NONNULL const char *name);
```

wargio (approved):

Issue #4576, opened 2024-07-25
Incorrect handling of OpenBSD specific code


cjeker:
The OpenBSD handling in bsd_debug.c is incorrect. The inspected flags have nothing to do with the execution state of the process. In OpenBSD -current the PS_STOPPED field was removed and so this code no longer compiles.

https://github.com/rizinorg/rizin/blob/2b4aa5351eb1f6721a629bd20a8d659d5c09d99e/librz/debug/p/native/bsd/bsd_debug.c#L223-L231

That code should do the same as the FreeBSD and NetBSD code and inspect p_stat. This will return the scheduler state (SSLEEP, SSTOP, SRUN, SONPROC and SDEAD) of the main thread.

I do not use rizin but was notified about the build failure on OpenBSD-current. This is why I open this issue.

See also https://github.com/radareorg/radare2/issues/23158 for the same issue in radare2.

Rot127:
@thestr4ng3r Do you have a working OpenBSD machine? I tried to set up a VM quickly, but it takes forever to get it to a state to compile Rizin.

cjeker:
I created a PR #4579 with the change that was committed to the OpenBSD ports repository to fix this issue.

cjeker:
I guess this can be closed now. Right?

PR #4579, opened 2024-07-26
Adjust OpenBSD code to set the process state to be less wrong.

cjeker:
Use the p_stat of the main thread to decide in which state the process is. For multi-threaded applications this is still not quite correct but at least this compiles on OpenBSD-current.

See also https://github.com/openbsd/ports/commit/5d4f90ee105757eda1cd9b31108c3ef6df67c48b

Fix for #4576

### DIFF
--- a/librz/debug/p/native/bsd/bsd_debug.c
+++ b/librz/debug/p/native/bsd/bsd_debug.c
@@ -218,16 +218,19 @@ RzDebugInfo *bsd_info(RzDebug *dbg, const char *arg) {
 		rdi->gid = kp->p__pgid;
 		rdi->exe = strdup(kp->p_comm);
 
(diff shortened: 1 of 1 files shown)


Rot127:
@XVilka If it is not working, builds but is not tested, we could remove it all together.

XVilka:
@Rot127 idea was to rewrite debugging from scratch, it's broken in so many ways on so many platforms. But for now we don't have capacity for that as our priorities lie in the realm of static analysis and RzIL-based emulation.

Rot127 (approved):
Ok, if we rewrite it from scratch anyways, I would say it is not worth the effort to remove it.

wargio (approved):

PR #4588, opened 2024-08-15
DWARF: fixes for loading big Zlib-compressed sections

XVilka:
# DO NOT SQUASH

**Detailed description**

In real binaries DWARF sections could be much bigger than just 50Mb, could be gigabytes (e.g. for Linux kernel), allow loading these. Also update Zlib version to address some stream inflating issues. Improve error messages.

Zlib-1.3 fails to build with the ASAN on macOS 14, but 1.3.1 succeeds.

Version 1.3.1 has these key updates from 1.3:

-    Reject overflows of zip header fields in minizip.
-    Fix bug in inflateSync() for data held in bit buffer.
-    Add LIT_MEM define to use more memory for a small deflate speedup.
-    Fix decision on the emission of Zip64 end records in minizip.
-    Add bounds checking to ERR_MSG() macro, used by zError().
-    Neutralize zip file traversal attacks in miniunz.
-    Fix a bug in ZLIB_DEBUG compiles in check_match(). 

**Test plan**

CI is green

### DIFF
--- a/librz/util/compression.c
+++ b/librz/util/compression.c
@@ -12,8 +12,8 @@
 #include <lzma.h>
 #endif
 
-// set a maximum output buffer of 50MB
-#define MAXOUT 50000000
+// set a maximum output buffer of 5000MB
+#define MAXOUT 5000000000
 
 /**
  * \brief inflate zlib compressed or gzipped, automatically accepts either the zlib or gzip format, and use MAX_WBITS as the window size logarithm.
@@ -66,7 +66,7 @@ RZ_API ut8 *rz_inflatew(RZ_NONNULL const ut8 *src, int srcLen, int *srcConsumed,
 	rz_return_val_if_fail(srcLen > 0, NULL);
 
 	int err = 0;
-	int out_size = 0;
+	ut64 out_size = 0;
 	ut8 *dst = NULL;
 	ut8 *tmp_ptr;
 	z_stream stream;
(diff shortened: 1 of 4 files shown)


librz/util/compression.c
```diff
@@ -12,8 +12,8 @@
 #include <lzma.h>
 #endif
 
-// set a maximum output buffer of 50MB
-#define MAXOUT 50000000
+// set a maximum output buffer of 5000MB
+#define MAXOUT 5000000000
```
wargio, line 16: this should be configurable.
maybe a get/set api is enough.

XVilka: WDYM? We don't store any context for these inflate functions, thus get/set API is not suitable. What I could do is to create some kind of "DecompressOpts" or similar.

wargio: for now is ok, but these seems to be in place to avoid to alloc too much mem

XVilka: True. But it made me puzzled at first why decompression failed on some debug files and why it didn't. It took a while to realize this comes from this limitation. This is why I also extended error messages.

XVilka: It didn't affect CI in any way. Could we merge this now, and maybe do a proper API change afterwards? Because this blocks many big ELF files with debug information compiled with relatively recent compilers (but before they introduced ZSTD).

wargio (approved):

PR #4602, opened 2024-09-02
meson.build: more specific flag for tree-sitter-c

ajakk:

**Detailed description**

We want the ability to use the system tree-sitter with the bundled tree-sitter-c, so we need a more granular switch for the latter rather than configuring them together.

See-Also: https://bugs.gentoo.org/928301

**Test plan**

Tested building with all combinations and (expectedly) hit a linking failure with the system tree-sitter-c due to the above bug.


XVilka (commented):
I need to think about this one. While you are right, and `tree-sitter` should be separate from `tree-sitter-c`, we ship modified `tree-sitter-c`, and extracting it into a system library might break other programs, or, if system ships unmodified `tree-sitter-c`, Rizin will break. 

@Rot127 @wargio @thestr4ng3r @ret2libc @kazarmy  opinions?

ret2libc:
> I need to think about this one. While you are right, and `tree-sitter` should be separate from `tree-sitter-c`, we ship modified `tree-sitter-c`, and extracting it into a system library might break other programs, or, if system ships unmodified `tree-sitter-c`, Rizin will break.
> 
> @Rot127 @wargio @thestr4ng3r @ret2libc @kazarmy opinions?

I think it is already possible to run Rizin with a system-provided tree-sitter-c, isn't it? If so, what this PR does is adding the possibility to have system tree-sitter but embedded tree-sitter-c (or viceversa). I *think* it should be fine.

wargio:
> > I need to think about this one. While you are right, and `tree-sitter` should be separate from `tree-sitter-c`, we ship modified `tree-sitter-c`, and extracting it into a system library might break other programs, or, if system ships unmodified `tree-sitter-c`, Rizin will break.
> > @Rot127 @wargio @thestr4ng3r @ret2libc @kazarmy opinions?
> 
> I think it is already possible to run Rizin with a system-provided tree-sitter-c, isn't it? If so, what this PR does is adding the possibility to have system tree-sitter but embedded tree-sitter-c (or viceversa). I _think_ it should be fine.

Worse case we have breakage because they used the system version. i think is acceptable, but i would add a note into the build script for gentoo, which explains that using that dependency as system instead of the modified version can make rizin malfunction. @ajakk

ajakk:
> i would add a note into the build script for gentoo, which explains that using that dependency as system instead of the modified version can make rizin malfunction

I think that the status quo is that the system tree-sitter is supported as much as any other `use_sys`-able package, especially since there's no downstream changes in the `tree-sitter` subproject. If that ever changes or there has to be downstream changes, I think Rizin is the right place to make that kind of dependency clear. Gentoo isn't unique in preferring system dependenceis to bundled dependencies.

Arguably, because of the downstream changes, the bundled `tree-sitter-c` should be forced to be used because of the downstream changes.

XVilka:
Using system `tree-sitter-c` should be prohibited. It will break most of the Rizin features related to the analysis and decompilation. I *strongly advise* you against using system `tree-sitter-c` since it will  ship broken programs to the users. Using system `tree-sitter` itself is fine, we also should update it before the release: https://github.com/rizinorg/rizin/issues/4623

You can see for yourself why the fork of `tree-sitter-c` is *absolutely necessary*:
- https://github.com/rizinorg/rizin/blob/dev/librz/type/parser/c_cpp_parser.c#L322
- https://github.com/rizinorg/tree-sitter-c/commit/d668edc6cbfb62f703ee4bc613fb0726bf2a8bc5

wargio:
Then we must not allow this.

ret2libc:
Yeah I guess the change should be to make the tree-sitter-c non-configurable (bundled as the default and only option) instead of configurable. Right @XVilka ?

wargio (changes requested):
Then this must NOT be merged, since it is not compatible.

XVilka:
Now is irrelevant because of https://github.com/rizinorg/rizin/commit/6f0d6750d14a1f3be16cada0ab351bc64f188db4

PR #4651, opened 2024-09-26
Check for jumps to preludes by default.

Rot127:

**Detailed description**

Remove the `tailcall` option and always check for it.
It also adds the possibility to mark instructions as `RZ_ANALYSIS_OP_TAIL`, if they are part of a tail call.

CFGs and iCFGs are not touched, they are flow wise still correct. But ignore function structures. So a `tail call` is not categorized as return for them.
This is better though, than attempting to encode abstract concepts (tail calls) into them which don't match instruction properties (jump != return, but it is conceptually). 

Others:

- Remove invalid code.
- Fix golang tests: Accept to find less strings, to not increase analysis depth.
  Because tail jumps are now correctly marked as end of a function,
  they are not always followed deeper with the default analysis depth of 1.
  Hence, less strings are found.
- Fix `pde` tests because they were wrong
  `rz_analysis_op_is_eob()` did not applied the mask onto op.type,
  so it checked against the wrong flags (especially for conditional jumps).
  Because the `eob` check was faulty the execution flow was different.

**Test plan**

All green and for Hexagon, x86 added one.

**Closing issues**

...


test/prj/v12-types-empty.rzdb
```diff
@@ -206,7 +206,6 @@ analysis.jmp.indir=false
 analysis.jmp.mid=true
 analysis.jmp.ref=true
 analysis.jmp.retpoline=true
-analysis.jmp.tailcall=0
```
XVilka: Don't do that - implement migration and a new project version bump instead.

XVilka (changes requested):
I like the idea but we definitely need at least a couple more tests for different architectures - ARM64 and x86_64, for example.

XVilka (approved):

wargio (approved):

Issue #4334, opened 2024-03-05
[RzArch] Progress tracker


Rot127:
## Tracking issue for `RzArch`

**Refactor old plugin**

- [x] Merge `asm` and `analysis`
  - [x] Merge `asm` and `analysis` code (keep plugin structure) (https://github.com/rizinorg/rizin/pull/4332)
  - [x] Mark `asm`/`analysis` plugin API as deprecated.
  - [ ] Rename existing `rz_arch_` functions to something else.
- [x] Move pseudo code generator code from `RzParse` into the merged `asm`/`analysis` directories.

**Extensions and new `RzArch` plugin**

- [ ] Add new `RzArch` plugin (see: https://github.com/rizinorg/rizin/pull/2557)
- [ ] Add compatibility functions (since we need to support `asm`/`analysis`/`arch` for a while all at once):
  - [ ] Wrap old `RzAnalysis` and `RzAsm` API functions into some new functions. They should check if the plugin is a `RzArch` plugin and call the new API if yes.
  - [ ] If `RzArch` returns different objects/data/values whatever, translate them for `core` 
- [ ] Refactor each archs `asm` and `analysis` plugin into the new `RzArch` plugin
  - [ ] `6502`
  - [ ] `8051`
  - [ ] `amd29k`
  - [ ] `arc`
  - [ ] `arm`
  - [ ] `avr`
  - [ ] `cil`
  - [ ] `cr16`
  - [ ] `cris`
  - [ ] `dalvik`
  - [ ] `dcpu16`
  - [ ] `ebc`
  - [ ] `gb`
  - [ ] `h8300`
  - [ ] `hexagon`
  - [ ] `hppa`
  - [ ] `i4004`
  - [ ] `i8080`
  - [ ] `include`
  - [ ] `java`
  - [ ] `lanai`
  - [ ] `lh5801`
  - [ ] `lm32`
  - [ ] `luac`
  - [ ] `mcore`
  - [ ] `mcs96`
  - [ ] `mips`
  - [ ] `msp430`
  - [ ] `nios`
  - [ ] `or1k`
  - [ ] `pic`
  - [ ] `ppc`
  - [ ] `propeller`
  - [ ] `pyc`
  - [ ] `riscv`
  - [ ] `rl78`
  - [ ] `rsp`
  - [ ] `rx`
  - [ ] `sh`
  - [ ] `snes`
  - [ ] `sparc`
  - [ ] `spc700`
  - [ ] `tms320`
[...]

PR #4656, opened 2024-09-30
RzArch and plugins

wargio:

**Detailed description**

Initial proposal and conversion of all the RzAsm/RzAnalysis plugins towards the unified RzArch* structures

Partially addresses https://github.com/rizinorg/rizin/issues/4334

### DIFF
--- a/librz/include/rz_arch.h
+++ b/librz/include/rz_arch.h
@@ -37,9 +37,8 @@ typedef enum {
 typedef ut64 RzAddress;
 
 typedef struct rz_arch_xcode_t {
-	RzArchXCodeMember member; ///< Describes the content of array
-	void *array; ///< Array of variable size, containing one or more structures of type member.
-	size_t length; ///< Length of the data as an array
+	RzArchXCodeMember member_type; ///< Describes the content of array
+	RzPVector /*<void>*/ *members; ///< Array of variable size, containing one or more structures of type member.
 } RzArchXCode;
 
 /// This needs to be redone, copied for reference from RzAnalysisOp
--- a/librz/include/rz_arch.h
+++ b/librz/include/rz_arch.h
@@ -5,6 +5,7 @@
 #ifndef RZ_ARCH_H
 #define RZ_ARCH_H
 
+#include <rz_util/rz_bitvector.h>
 #include <rz_util/rz_iterator.h>
 #include <rz_types.h>
 #include <rz_vector.h>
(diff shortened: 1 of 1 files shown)


librz/include/rz_arch.h
```diff
@@ -16,12 +20,140 @@ extern "C" {
...
+} RzArchHint;
+
+typedef enum {
+	RZ_ARCH_DETAIL_ACCESS_UNDEF = 0, ///< Undefined access
+	RZ_ARCH_DETAIL_ACCESS_READ, ///< Read access
+	RZ_ARCH_DETAIL_ACCESS_WRITE, ///< Write access
+} RzArchDetailAccess;
+
+typedef enum {
+	RZ_ARCH_DETAIL_MEMBER_REGISTER = 0, ///< Member is raw bytes
```
XVilka: Comment is wrong

XVilka: Same for other members

XVilka (changes requested):

librz/include/rz_arch.h
```diff
@@ -16,12 +20,140 @@ extern "C" {
...
+	bool (*can_xcode_out)(RZ_NONNULL RzArchXCodeMember output); ///< Returns true if the plugin can support the given RzArchXCodeMember in ouput.
+	bool (*context_init)(RZ_NONNULL RzConfig *config, RZ_OUT RzArchPluginContext** context); ///< Create a new context for a given configuration
+	void (*context_fini)(RZ_NULLABLE RzArchPluginContext *context); ///< Free the given context
+	void (*context_update)(RZ_NULLABLE RzArchPluginContext *context, RZ_NONNULL RzConfig *config); ///< Updates the given context with the given configuration.
+	bool (*context_xcode)(RZ_NULLABLE RzArchPluginContext *context, RZ_NONNULL RzArchXCode *input, RZ_NONNULL RzArchXCode *output); ///< Updates the given context with the given configuration.
 } RzArchPlugin;
 
+typedef struct rz_arch_t {
+	RzPVector /*<RzArchPlugin*>*/ *plugins;
+	HtSP /*<char*, RzConfig*>*/ plugins_config;
```
XVilka: Is it a plugin-specific configuration?

XVilka (changes requested):
One big thing is missing in here - a target description, e.g.
```c
typedef struct {
    char *arch;
    char *cpu;
    ut16 bits;
} RzArchTarget;
```
This is where we can pack also the "features", e.g. particular ISA features or anything like that. Maybe also have a separate `address_width` member, e.g. see function `rz_analysis_get_address_bits()`

librz/include/rz_arch.h
```diff
@@ -16,12 +20,140 @@ extern "C" {
 
 RZ_LIB_VERSION_HEADER(rz_arch);
 
+typedef enum {
+	RZ_ARCH_XCODE_MEMBER_BYTES = 0, ///< Member is raw bytes
```
Rot127: ~~Maybe add an invalid value so we can initialize it with `RZ_NEW0`~~
As discussed in Mattermost, we should consider just passing an `RzBuffer`.

librz/include/rz_arch.h
```diff
@@ -16,12 +20,140 @@ extern "C" {
...
+	RZ_ARCH_XCODE_MEMBER_TOKEN, ///< Member is RzArchToken
+	RZ_ARCH_XCODE_MEMBER_PARSE, ///< [Deprecated] Member is parsed string (pseudo code)
+	RZ_ARCH_XCODE_MEMBER_ESIL, ///< [Deprecated] Member is ESIL
+} RzArchXCodeMember;
+
+#define RZ_ARCH_BUFFER_SIZE 128
+
+typedef struct rz_arch_xcode_t {
+	RzArchXCodeMember member; ///< Describes the content of array
+	void *array; ///< Array of variable size, containing one or more structures of type member.
```
Rot127: Better a `(P)Vector` or `Set`? Because we get iterators and such for free.

librz/include/rz_arch.h
```diff
@@ -16,12 +20,140 @@ extern "C" {
...
+	RZ_ARCH_DETAIL_MEMBER_SIGNED, ///< Member is RzArchPacket
+} RzArchDetailMember;
+
+typedef struct rz_arch_detail_value_t {
+	RzArchDetailMember member;
+	union {
+		size_t register_id;
+		ut64 imm_unsigned;
+		st64 imm_signed;
+	};
```
Rot127: Also needs a bitvector for arbitrary sized values.

librz/include/rz_arch.h
```diff
@@ -16,12 +20,140 @@ extern "C" {
...
+	RzStrBuf esil;
+	RzStrBuf opex;
+	const char *reg; /* destination register */
+	const char *ireg; /* register used for indirect memory computation*/
+	int scale;
+	ut64 disp;
+	RzAnalysisSwitchOp *switch_op;
+	RzAnalysisHint hint;
+	RzAnalysisDataType datatype;
+} RzArchPacket;
```
Rot127, line 244: If `RzArchPacket` is supposed to be the "instruction packet" of VLIW, please move the this content into an `RzArchInsn` and make `RzArchPacket` a struct which holds multiple `RzArchInsn` in a vector (so we allow for order in them).

Would propose to implement iterators for `RzArchPacket` to access the single instructions:

- `RzIterator *rz_arch_packet_insn()` - Over all instructions
- `RzIterator *rz_arch_packet_calls()` - Over call instructions
- `RzIterator *rz_arch_packet_jumps()` - Over jumps
- `RzIterator *rz_arch_packet_mem()` - Over memory referencing instructions.
- `RzIterator *rz_arch_packet_specific(RzAnalysisOpType *atype)`
...

And getters for values with certain properties:

- `RzIterator *rz_arch_packet_code_refs(..., RzInsnRefType *ref_type)`
- `RzIterator *rz_arch_packet_data_refs(..., RzInsnRefType *ref_type)`

```
typedef enum {
  REF_TYPE_DATA = 1, ///< All data references
  REF_TYPE_CODE = 2, ///< All code references
  // Now more detailed characteristics
  REF_TYPE_STACK, ///< ref is stack related
  REF_TYPE_HEAP, ///< ref is heap related
  REF_TYPE_GLOBAL,
  REF_TYPE_CACHE,
  // Deprecated
  REF_TYPE_PTR, ///< returns former RzAnalysisOp.ptr
  REF_TYPE_STACKPTR, ///< returns former RzAnalysisOp.stackptr
  REF_TYPE_JUMP, ///< returns former RzAnalysisOp.jump
...
} RzRefType
```

Some or all of the should also be accessible via `xcode()`. `RZ_XCODE_MEMBER_STACK_REFS`, `RZ_XCODE_MEMBER_JMP_TARGETS`, `RZ_XCODE_MEMBER_CALL_TARGES` etc.

Because if we keep it like this, we starting coding again on the wrong abstraction layer of RzAnalysisOp.

librz/include/rz_arch.h
```diff
@@ -16,12 +20,140 @@ extern "C" {
...
+		size_t register_id;
+		ut64 imm_unsigned;
+		st64 imm_signed;
+	};
+} RzArchDetailValue;
+
+typedef struct rz_arch_detail_t {
+	RzArchDetailValue source[6];
+	RzArchDetailValue destination;
+} RzArchDetail;
```
Rot127: I struggle with the term `Detail`. Does it mean detail of operands? Or other details of the instruction? It looks like all operands of an instruction?
Also vectors are needed here again. Because we cannot assume a single destination or a just 6 sources etc.

If it is supposed to mean "RzArchOperand", as in "operand of an instruction", we should probably introduce a new object for them as well. Because they can become quite complex as well (e.g. memory operands consist of one reg, one imm. Or one reg, one implicit operand etc).

librz/include/rz_arch.h
```diff
@@ -16,12 +20,140 @@ extern "C" {
...
-	RZ_DEPRECATE RzAnalysisPlugin *p_analysis; ///< Analysis Plugin
-	RZ_DEPRECATE RzParsePlugin *p_parse; ///< Parse Plugin
+	RZ_DEPRECATE RzAsmPlugin *p_asm; ///< [Deprecated] Assembly Plugin
+	RZ_DEPRECATE RzAnalysisPlugin *p_analysis; ///< [Deprecated] Analysis Plugin
+	RZ_DEPRECATE RzParsePlugin *p_parse; ///< [Deprecated] Parse Plugin
+
+    bool (*init)(RZ_NONNULL RzConfig *config); ///< Global constructor for the plugin to fill the configuration values.
+    bool (*fini)(); ///< Global destructor for the plugin
+	bool (*can_xcode_in)(RZ_NONNULL RzArchXCodeMember input); ///< Returns true if the plugin can support the given RzArchXCodeMember in input.
+	bool (*can_xcode_out)(RZ_NONNULL RzArchXCodeMember output); ///< Returns true if the plugin can support the given RzArchXCodeMember in ouput.
```
Rot127, line 279: An `can_xcode_transform` would be nice. To test if the plugin supports a given `in/out` combination. Because `can_xcode_in() && can_xcode_out() == true` but `context_xcode(in, out)` could be unsupported.

Rot127 (commented):
Added quite some stuff which probably shouldn't be done in the first implementation.
I think necessary is though:
- The `RzArchPacket`/`RzArchInsn` distinction.
- Starting enums always with `invalid = 0` to prevent false positives.

librz/include/rz_arch.h
Rot127: Ouh wait. I am stupid.
The packet can contain both. A vector with instructions (for VLIW archs) and single `RzAnalysisOp`/`RzAsm`/`RzArchInsn`.

Single instruction focused archs use always the last one. VLIW can implement everything over the vector of instructions.
And we only allow access to the details like `references`, `jump/call targets`, `asm_text` over the `xcode()`.

But the Packet itself should never be made available to the core.
But the iterators from above can be implemented just for plugin internal stuff.

Rot127:
If we allow to query for specific properties via `xcode()`, we should implement also a common buffer mechanism for disassembled instructions of a plugin.
So the plugin doesn't need to disassemble everything again and again, just because the code asks for `call refs`, `stack refs`, `asm_text` in a row.

librz/include/rz_arch.h
wargio: i don't think this is an issue because we can define the transform, for example:
`asm > details` could become internally (in rzarch not plugins) `asm > bytes > details`.

librz/include/rz_arch.h
wargio: maybe operand makes more sense

librz/include/rz_arch.h
wargio: Yes pvec is a better choice.

XVilka:
This needs to be updated on top of the latest `dev` because of the recent changes in RzAnalysis and RzAsm

PR #4667, opened 2024-10-13
Aweful RzAsm <-> RzAnalysis pointer sharing and Hexagon unify

Rot127:

**Detailed description**

**Hacks**

Attempt to share pointers between `RzAsm` and `RzAnalysis` without any `RzArch`.

The assumption made here is that `RzAsm` is always initialized. Also without any `RzAnalysis`. But `RzAnalysis` is never initialized without a companion `RzAsm`.
Hence a plugin requiring a common state, always accesses `RzAsm->plugin_data`, no matter if it was called from `RzAnalysis` or `RzAsm`.

`RzAnalysis` holds the file buffer. So if this one is present, the plugin should initialize a buffer around it. If it is not there, it should make a byte buffer around the bytes passed to the function.

This mimics future behavior of `RzArch` which will get either an `IO` or `Bytes` buffer.

**Plugin configs**

With this change there is also the need to refactor the plugin specific configs.
I designed them badly before. And they couldn't be used by a plugin without a static variable.

Plugin specific configs are hold now in a hash mpa in RzCore. The reason to have them separately and not in the `RzCore->config` is because plugins can be the _owner_ of them. Meaning the `config_set` functions will pass a pointer with the plugin state. Not the core state. This is required if plugins have specific configuration options they need to know, but have no access to `RzCore` to get them.

It adds luckily only a few more lines to handle them in `cmd_eval.c` and for the auto-complete.

**Hexagon changes**

- Moves the state from a static variable into `RzAsm->plugin_data`.
- Always disassembles the context around an instruction to know the packet properties the instruction is in.

**Test plan**

For plugin configs
Otherwise: all green with fuzz-dist.

**Closing issues**

None. Probably opens a few more.


XVilka (commented):
The changes are good. I am only in doubt about those hacks. Let's see what others say.

librz/core/core.c
```diff
@@ -1,6 +1,8 @@
 // SPDX-FileCopyrightText: 2009-2020 pancake <pancake@nopcode.org>
 // SPDX-License-Identifier: LGPL-3.0-only
 
+#include "rz_config.h"
+#include "rz_util/ht_sp.h"
```
wargio:
```suggestion
#include <rz_config.h>
#include <rz_util/ht_sp.h>
```

wargio (approved):
LGTM. Hopefully this hack will go away with RzArch

XVilka (approved):
LGTM but only as a temporary measure. We should not make a new release with this hack.

wargio (approved):

PR #4692, opened 2024-10-28
Print the max number of lines as default in pd

wargio:

**Detailed description**

Seems that `pd` had a regression. `pd` has an **optional** argument but we always fail when not given, thus this reverts back the changes and allows to print with a default value equivalent of the max rows number of the terminal (when this fails, then we use 16).


Jookia:
As a little context, currently (dev & stable) visual mode disassembly is broken because its pd invocation expects this behaviour.
These changes fix the issue

XVilka (approved):

PR #4695, opened 2024-11-02
Use relocation types name defined by each bin format.

Roeegg2:

**Detailed description**

* Modified `bin_reloc_type_name` to return correct naming if possible, otherwise return the old `ADD`/`SET` naming.
* Add print_name field to `RzBinReloc`

Every relocation is being assigned a type based off whether it needs an addend added to it or not (`SET` for no addend, `ADD` for with). This is what `RzBinReloc.additive` was for.
`RzBinReloc.type` has just been the bit size of the relocation
And the final name of the relocation printed was just a combination of the both (which is wrong)

In order to not break all the other relocation types, I haven't removed the old code completely yet. When all the other executable formats support the new printing I'll remove it completely.  
...

**Test plan**

All tests green
...

**Closing issues**

...

### DIFF
--- a/librz/bin/p/bin_elf.inc
+++ b/librz/bin/p/bin_elf.inc
@@ -1023,7 +1023,7 @@ static void patch_reloc_hexagon(RZ_INOUT RzBuffer *buf_patched, const ut64 patch
 // AARCH64-specific defines
 // Take the PAGE component of an address or offset.
 #define PG(x)         ((x) & ~0xFFFULL)
-#define PG_OFFSET(x)  ((x) & 0xFFFULL)
+#define PG_OFFSET(x)  ((x)&0xFFFULL)
 #define ADR_IMM_MASK1 (((1U << 2) - 1) << 29)
 #define ADR_IMM_MASK2 (((1U << 19) - 1) << 5)
 #define ADR_IMM_MASK3 (((1U << 19) - 1) << 2)
--- a/librz/bin/p/bin_elf.inc
+++ b/librz/bin/p/bin_elf.inc
@@ -1023,7 +1023,7 @@ static void patch_reloc_hexagon(RZ_INOUT RzBuffer *buf_patched, const ut64 patch
 // AARCH64-specific defines
 // Take the PAGE component of an address or offset.
 #define PG(x)         ((x) & ~0xFFFULL)
-#define PG_OFFSET(x)  ((x)&0xFFFULL)
+#define PG_OFFSET(x)  ((x) & 0xFFFULL)
 #define ADR_IMM_MASK1 (((1U << 2) - 1) << 29)
 #define ADR_IMM_MASK2 (((1U << 19) - 1) << 5)
 #define ADR_IMM_MASK3 (((1U << 19) - 1) << 2)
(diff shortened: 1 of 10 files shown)


librz/bin/p/bin_elf.inc
```diff
@@ -1023,7 +1023,7 @@ static void patch_reloc_hexagon(RZ_INOUT RzBuffer *buf_patched, const ut64 patch
 // AARCH64-specific defines
 // Take the PAGE component of an address or offset.
 #define PG(x)         ((x) & ~0xFFFULL)
-#define PG_OFFSET(x)  ((x)&0xFFFULL)
+#define PG_OFFSET(x)  ((x) & 0xFFFULL)
```
Rot127: You might need to use `clang-format-16` or just revert these changes to macro formatting.

librz/bin/p/bin_elf.inc
```diff
@@ -1389,33 +1389,34 @@ static RzBinReloc *reloc_convert(ELFOBJ *bin, RzBinElfReloc *rel, ut64 GOT) {
 		r->section_vaddr = section->rva;
 	}
 
-#define SET(T) \
+#define SET(T, N) \
```
Rot127:
```suggestion
#define SET(T, RELOC_NAME) \
```

Nitpicky, though it saves one more lookup.
Same below

librz/bin/p/bin_elf.inc
```diff
@@ -1389,33 +1389,34 @@ static RzBinReloc *reloc_convert(ELFOBJ *bin, RzBinElfReloc *rel, ut64 GOT) {
...
+		case RZ_386_RELATIVE: ADD(32, B, "R_386_RELATIVE");
+		case RZ_386_GOTOFF: ADD(32, -GOT, "R_386_GOTOFF");
+		case RZ_386_GOTPC: ADD(32, GOT - P, "R_386_GOTPC");
+		case RZ_386_16: ADD(16, 0, "R_386_16");
+		case RZ_386_PC16: ADD(16, -P, "R_386_PC16");
+		case RZ_386_8: ADD(8, 0, "R_386_8");
+		case RZ_386_PC8: ADD(8, -P, "R_386_PC8");
+		case RZ_386_COPY: ADD(32, 0, "R_386_COPY"); // copy symbol at runtime
+		case RZ_386_IRELATIVE: r->is_ifunc = true; SET(32, "R_386_IRELATIVE");
+		// FIXME: Quite a few relocatons missing here.
```
Rot127, line 1418: Please open an issue about it. Otherwise we will forget.

librz/core/cbin.c
```diff
@@ -1854,6 +1854,9 @@ RZ_API void rz_core_bin_print_source_line_info(RzCore *core, const RzBinSourceLi
 }
 
 static const char *bin_reloc_type_name(RzBinReloc *reloc) {
+	if (reloc->additive == 2) { // if we can use '.print_name'.
+		return reloc->print_name;
+	}
```
Rot127: Please use another indicator then `additive == 2`. `additive` has no relation to the name printing and it gets confusing. Couldn't you just check if `print_name` is `NULL`? Or am I missing something?

Rot127 (changes requested):
Very much like these changes!
Please address the comments though and don't delete half of the PR description template. It is ok to fill out `All tests green` in the testing paragraph. If this is actually sufficient to test the changes (as it does here).

librz/core/cbin.c
Roeegg2: You could check if `print_name` is `NULL`, it's just adding `print_name = NULL` on every other format now. I thought one line changing `additive` would be better. It's adding a single line to the `ELF` processing. 
We will be deleting the line either way soon so I didn't think it mattered much, but I could change it

librz/bin/p/bin_elf.inc
Roeegg2: You're right, changing it now

librz/bin/p/bin_elf.inc
Roeegg2: This was the result of running `/sys/clang-format.py`... 
I'll split the formatting changes into a separate commit

librz/core/cbin.c
Rot127: The `reloc` struct should be initilized with `RZ_NEW0` before. So it should be already 0 by default.

Roeegg2: Either way I'll change the naming

Roeegg2: > The `reloc` struct should be initilized with `RZ_NEW0` before. So it should be already 0 by default.

Oh OK, I'll change it then

wargio: i don't know why 2 is there, but maybe define a macro which can be used to understand why the constant 2 is there.

wargio (commented):
looks so much better. wonderful job.

librz/bin/p/bin_elf.inc
wargio: you can download clang format 16 (precompiled) here https://github.com/muttleyxd/clang-tools-static-binaries/releases/ and pass the path of it as argument to the python script.

Roeegg2:
@Rot127 @wargio did I miss anything?

librz/core/cbin.c
Roeegg2: > Either way I'll change the naming

Ended up not changing the name since when used in the old way, it makes sense. Added a comment though to clarify what it is.

librz/core/cbin.c
```diff
@@ -1854,6 +1854,9 @@ RZ_API void rz_core_bin_print_source_line_info(RzCore *core, const RzBinSourceLi
 }
 
 static const char *bin_reloc_type_name(RzBinReloc *reloc) {
+	if (reloc->print_name != NULL) { // print_name is set to NULL by default. If it's not null, that means we can use the new naming
```
Rot127:
```suggestion
	if (reloc->print_name) {
```

This is a very common pattern in C code. No need to document it in code :)
Do it in the struct as suggested below.

librz/include/rz_bin.h
```diff
@@ -666,7 +666,8 @@ typedef struct rz_bin_reloc_t {
 	ut64 target_vaddr; ///< the target address that the patched reloc points to
 	ut64 section_vaddr; ///< the subsection address
 	ut32 visibility;
-	bool additive;
+	bool additive; ///< Specifies whether the relocation needs to have an addend added to it.
+	const char *print_name;
```
Rot127:
```suggestion
	const char *print_name; ///< Name of the relocation type. NULL if none is specified. Not setting this field is deprecated.
```

Rot127 (changes requested):

Roeegg2:
@Rot127 done

wargio (approved):
LGTM

librz/core/cbin.c
```diff
@@ -1854,6 +1854,9 @@ RZ_API void rz_core_bin_print_source_line_info(RzCore *core, const RzBinSourceLi
 }
 
 static const char *bin_reloc_type_name(RzBinReloc *reloc) {
+	if (reloc->print_name != NULL) {
```
wargio:
```suggestion
	if (RZ_STR_ISNOTEMPTY(reloc->print_name)) {
```

librz/include/rz_bin.h
```diff
@@ -666,7 +666,8 @@ typedef struct rz_bin_reloc_t {
 	ut64 target_vaddr; ///< the target address that the patched reloc points to
 	ut64 section_vaddr; ///< the subsection address
 	ut32 visibility;
-	bool additive;
+	bool additive; ///< Name of the relocation type. NULL if none is specified. Not setting this field is deprecated.
```
wargio:
```suggestion
	bool additive;
```

wargio:
@XVilka muon seems broken?

XVilka:
> @XVilka muon seems broken?

No, just a small name change:
- https://git.sr.ht/~lattis/muon/commit/6ec469bb42b28f3a1943d58c0e15a8493220c2f7
- https://github.com/rizinorg/rizin/pull/4700

XVilka (approved):
LGTM

PR #4703, opened 2024-11-05
Force Function Rename

brightprogrammer:
New function to forcefully rename an analysis function. Multiple retries will be performed by adding a numeric suffix to function name until the number of retries max's out

 

**Detailed description**

...

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/librz/arch/function.c
+++ b/librz/arch/function.c
@@ -259,17 +259,17 @@ RZ_API bool rz_analysis_function_relocate(RzAnalysisFunction *fcn, ut64 addr) {
 }
 
 /**
- * @b Rename given function.
+ * \b Rename given function.
  * This method will fail if a function with same name as `name`
  * already exists in rizin's analysis.
  *
- * @param fcn Rizin analysis function.
- * @param name A unique name
+ * \param fcn Rizin analysis function.
+ * \param name A unique name
  *
- * @return `true` on success.
- * @return `false` otherwiese.
+ * \return `true` on success.
+ * \return `false` otherwiese.
  *
- * @sa rz_analysis_function_force_rename
+ * \sa rz_analysis_function_force_rename
  * */
 RZ_API bool rz_analysis_function_rename(RZ_NONNULL RzAnalysisFunction *fcn, RZ_NONNULL const char *name) {
 	rz_return_val_if_fail(fcn && name, false);
@@ -298,17 +298,17 @@ RZ_API bool rz_analysis_function_rename(RZ_NONNULL RzAnalysisFunction *fcn, RZ_N
 }
 
 /**
- * @b Force rename a function.
+ * \b Force rename a function.
  * This will make sure a function with non-unique name gets renamed by
  * adding a numeric suffix to it.
  *
- * @param fcn
- * @param name
+ * \param fcn
+ * \param name
  *
- * @return New name on success.
- * @return `NULL` otherwise.
+ * \return New name on success.
+ * \return `NULL` otherwise.
  *
- * @sa rz_analysis_function_rename
+ * \sa rz_analysis_function_rename
  * */
 RZ_API RZ_BORROW const char *rz_analysis_function_force_rename(RZ_NONNULL RzAnalysisFunction *fcn, RZ_NONNULL const char *name) {
 	rz_return_val_if_fail(fcn && name, NULL);
--- a/librz/arch/function.c
+++ b/librz/arch/function.c
@@ -258,7 +258,22 @@ RZ_API bool rz_analysis_function_relocate(RzAnalysisFunction *fcn, ut64 addr) {
 	return true;
 }
 
+/**
+ * @b Rename given function.
+ * This method will fail if a function with same name as `name`
+ * already exists in rizin's analysis.
+ *
+ * @param fcn Rizin analysis function.
+ * @param name A unique name
+ *
+ * @return `true` on success.
+ * @return `false` otherwiese.
+ *
+ * @sa rz_analysis_function_force_rename
+ * */
 RZ_API bool rz_analysis_function_rename(RzAnalysisFunction *fcn, const char *name) {
+	rz_return_val_if_fail(fcn && name, false);
+
 	RzAnalysis *analysis = fcn->analysis;
 	RzAnalysisFunction *existing = ht_sp_find(analysis->ht_name_fun, name, NULL);
 	if (existing) {
@@ -280,6 +295,38 @@ RZ_API bool rz_analysis_function_rename(RzAnalysisFunction *fcn, const char *nam
 		ht_sp_insert(analysis->ht_name_fun, fcn->name, fcn);
 	}
 	return true;
+}
+
+/**
+ * @b Force rename a function.
+ * This will make sure a function with non-unique name gets renamed by
+ * adding a numeric suffix to it.
+ *
+ * @param fcn
+ * @param name
+ *
+ * @return New name on success.
+ * @return `NULL` otherwise.
+ *
+ * @sa rz_analysis_function_rename
+ * */
+RZ_API const char *rz_analysis_function_force_rename(RzAnalysisFunction *fcn, const char *name) {
+	rz_return_val_if_fail(fcn && name, NULL);
+
+	const char *final_name = strdup(name);
+	ut32 retries = 100;
+	ut32 suffix = 0;
+	while ((retries--) && !rz_analysis_function_rename(fcn, final_name)) {
+		if (final_name) {
+			free((void *)final_name);
+			final_name = NULL;
+		}
+
+		ut32 strsz = snprintf(NULL, 0, "%s_%u", name, suffix) + 1;
+		final_name = malloc(strsz);
+		snprintf((char *)final_name, strsz, "%s_%u", name, suffix++);
+	}
+	return retries ? fcn->name : NULL;
 }
 
 RZ_API void rz_analysis_function_add_block(RzAnalysisFunction *fcn, RzAnalysisBlock *bb) {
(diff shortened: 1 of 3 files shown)


brightprogrammer:
About [canalysis.c:3336](https://github.com/rizinorg/rizin/blob/64a4bb491a7cedb6381b1d0b0f34974ad856e13d/librz/core/canalysis.c#L3336) TODO?

librz/arch/function.c
```diff
@@ -258,7 +258,22 @@ RZ_API bool rz_analysis_function_relocate(RzAnalysisFunction *fcn, ut64 addr) {
...
+ *
+ * @param fcn Rizin analysis function.
+ * @param name A unique name
+ *
+ * @return `true` on success.
+ * @return `false` otherwiese.
+ *
+ * @sa rz_analysis_function_force_rename
+ * */
 RZ_API bool rz_analysis_function_rename(RzAnalysisFunction *fcn, const char *name) {
```
wargio: add RZ_NOTNULL

librz/arch/function.c
```diff
@@ -282,6 +297,38 @@ RZ_API bool rz_analysis_function_rename(RzAnalysisFunction *fcn, const char *nam
...
+ *
+ * @param fcn
+ * @param name
+ *
+ * @return New name on success.
+ * @return `NULL` otherwise.
+ *
+ * @sa rz_analysis_function_rename
+ * */
+RZ_API const char *rz_analysis_function_force_rename(RzAnalysisFunction *fcn, const char *name) {
```
wargio: same here. RZ_NOT_NULL.

librz/arch/function.c
```diff
@@ -282,6 +297,38 @@ RZ_API bool rz_analysis_function_rename(RzAnalysisFunction *fcn, const char *nam
...
+		if (final_name) {
+			free((void *)final_name);
+			final_name = NULL;
+		}
+
+		ut32 strsz = snprintf(NULL, 0, "%s_%u", name, suffix) + 1;
+		final_name = malloc(strsz);
+		snprintf((char *)final_name, strsz, "%s_%u", name, suffix++);
+	}
+	return retries ? fcn->name : NULL;
```
wargio: instead of doing retries, just do `name_offset` which is granted to be unique. also use `rz_strf` directly instead of calling snprintf.

wargio (changes requested):
See above

wargio:
> About [canalysis.c:3336](https://github.com/rizinorg/rizin/blob/64a4bb491a7cedb6381b1d0b0f34974ad856e13d/librz/core/canalysis.c#L3336) TODO?

ignore it.

wargio:
Add a test also for this.

librz/arch/function.c
```diff
@@ -310,23 +310,18 @@ RZ_API bool rz_analysis_function_rename(RzAnalysisFunction *fcn, const char *nam
  *
  * @sa rz_analysis_function_rename
  * */
-RZ_API const char *rz_analysis_function_force_rename(RzAnalysisFunction *fcn, const char *name) {
+RZ_API const char *rz_analysis_function_force_rename(RZ_NONNULL RzAnalysisFunction *fcn, RZ_NONNULL const char *name) {
```
wargio:
```suggestion
RZ_API RZ_BORROW const char *rz_analysis_function_force_rename(RZ_NONNULL RzAnalysisFunction *fcn, RZ_NONNULL const char *name) {
```

librz/arch/function.c
```diff
@@ -310,23 +310,18 @@ RZ_API bool rz_analysis_function_rename(RzAnalysisFunction *fcn, const char *nam
...
-		snprintf((char *)final_name, strsz, "%s_%u", name, suffix++);
+	if (rz_analysis_function_rename(fcn, name)) {
+		return fcn->name;
+	} else {
+		// assuming final name can never be bigger than 255 bytes
+		char final_name[256];
+		// {name}_{addr} is guaranteed to be unique
+		rz_strf(final_name, "%s_%llx", name, fcn->addr);
+		return rz_analysis_function_rename(fcn, final_name) ? fcn->name : NULL;
 	}
```
wargio: remove the `else` since it is not needed.
Also do not use fixed size buffer for this but instead use `rz_str_newf` (sorry my bad, i mixed up names when i asked the previous changes) and free it after calling `bool ok = rz_analysis_function_rename(...)`, then `return ok ? fcn->name : NULL;`

wargio: also use `PFMT64x` instead of `llx`, for example: `rz_str_newf("%s_%" PFMT64x, name, fcn->addr);` because `llx` is not portable.

librz/arch/function.c
brightprogrammer: This is already done [here](https://github.com/rizinorg/rizin/pull/4703/commits/057500dc42364527c5db3116e2e729d57d5109b3#diff-79ead514aa1e847124523b7f25695cb7e436de2494b5b09d7533a0ae1b3332baR313)

test/unit/test_analysis_function.c
```diff
@@ -468,6 +468,31 @@ bool test_noreturn_functions_list() {
...
+
+	// we know b does exist, so rename must fail
+	RzAnalysisFunction *c = rz_analysis_create_function(analysis, "c", 0xbbbb, RZ_ANALYSIS_FCN_TYPE_FCN);
+	mu_assert_false(rz_analysis_function_rename(c, "b"), "rename c to b");
+
+	mu_end;
+}
+
+bool test_analysis_function_force_rename() {
+	RzAnalysis *analysis = rz_analysis_new();
```
wargio, line 491: free rz_analysis

test/unit/test_analysis_function.c
```diff
@@ -468,6 +468,31 @@ bool test_noreturn_functions_list() {
 	mu_end;
 }
 
+bool test_analysis_function_rename() {
+	RzAnalysis *analysis = rz_analysis_new();
```
wargio, line 473: free rz_analysis

test/unit/test_analysis_function.c
```diff
@@ -468,6 +468,31 @@ bool test_noreturn_functions_list() {
...
+}
+
+bool test_analysis_function_force_rename() {
+	RzAnalysis *analysis = rz_analysis_new();
+
+	RzAnalysisFunction *a = rz_analysis_create_function(analysis, "a", 0xcafe, RZ_ANALYSIS_FCN_TYPE_FCN);
+	rz_analysis_create_function(analysis, "b", 0xbabe, RZ_ANALYSIS_FCN_TYPE_FCN);
+
+	// rename a to b, but b already exists, so we force rename
+	mu_assert_notnull(rz_analysis_function_force_rename(a, "b"), "function force rename");
```
wargio: maybe check the output name.

wargio (changes requested):
there is no real need to call `RZ_FREE` since `free` was enough, but for me this is almost ok (besides the missing frees on rz_analysis and the missing check on the output name).
@thestr4ng3r

test/unit/test_analysis_function.c
```diff
@@ -490,6 +495,11 @@ bool test_analysis_function_force_rename() {
 
 	// rename a to b, but b already exists, so we force rename
 	mu_assert_notnull(rz_analysis_function_force_rename(a, "b"), "function force rename");
+	// check whether the name begins with originally provided one
+	mu_assert_true(rz_str_startswith(a->name, "b"), "function renamed incorrectly");
```
wargio: i'm not yet ok with this test.
I think it should be two functions defined `a` and `b` (at different addresses), then i would force rename `a` to be `b` and check that is `b_xxxx` (since you know the address, it should be easy to guess the name).

wargio: also check the return value matches with what is in  `a->name`

brightprogrammer: I first thought about doing it that way, but then it assumes many things about the internal implementation. I just want to assume the function gets renamed with a unique suffix. This allows the implementation to be changed in future without breaking the test

wargio: no. the function is meant to return a certain result which we know being name_offset, we should test for this, not for anything else. also if in the future we change something we can always change the test.

brightprogrammer: Ok, I'll make the change :-)

wargio (approved):
LGTM (as long as is green)

librz/arch/function.c
```diff
@@ -258,7 +258,22 @@ RZ_API bool rz_analysis_function_relocate(RzAnalysisFunction *fcn, ut64 addr) {
 	return true;
 }
 
-RZ_API bool rz_analysis_function_rename(RzAnalysisFunction *fcn, const char *name) {
+/**
+ * @b Rename given function.
```
XVilka: Use `\p`, etc Doxygen syntax

librz/arch/function.c
```diff
@@ -282,6 +297,34 @@ RZ_API bool rz_analysis_function_rename(RzAnalysisFunction *fcn, const char *nam
 	return true;
 }
 
+/**
+ * @b Force rename a function.
+ * This will make sure a function with non-unique name gets renamed by
+ * adding a numeric suffix to it.
+ *
+ * @param fcn
```
XVilka: Same

XVilka (changes requested):

librz/arch/function.c
brightprogrammer: Isn't this doxygen syntax as well?
> Structural commands (like [all other commands](https://doxygen.nl/manual/commands.html#cmd_intro)) start with a backslash (\), or an at-sign (@) if you prefer Javadoc style, followed by a command name and one or more parameters.

XVilka: Yes, but we use the `\p` one: https://github.com/rizinorg/rizin/blob/dev/DEVELOPERS.md#documentation

brightprogrammer: 👍 sorry for not checking that 😅 , I'll make the change

brightprogrammer: I have a habit of using `@` everywhere :-)

XVilka (approved):

PR #4711, opened 2024-11-11
Implement PE & MDMP base relocations

Roeegg2:

**Detailed description**

Up until now PE relocations were implemented as imports, which is incorrect. The PE format uses base relocations for image files (ie executable files) and classic COFF relocations for object files.

Implemented so far:
- PE base relocation parsing + assigning
- PE base relocation tests

Still need to implement:
- ~~relocation patching~~ - Rizin always loads the binary in it's expected base address, so no patching is needed (I added a detailed comment about this in `relocs` function in `bin_pe.inc`)
- ~~make MDMP use the new base relocation~~ MDMP doesn't have relocations, only imports.
- ~~add COFF relocations for object files~~ technically PE can have COFF relocations since it's based off of COFF, but COFF relocations are only used for for object files, and there is no reason to use PE for object files, so it's never actually used. 

Notes:
- There are no names printed with base relocations, since they don't have a symbol/name field associated with them. Simply a `type` and `offset` to apply the relocation in.
...

**Test plan**

All tests green.

...

**Closing issues**

...


Roeegg2:
Some things I wasn't sure about:
* The comment in `has_canary` seems to be from r2, and I don't see what leaks might be caused from this...
* Should I add licensing to the new created files?
* I deliberately didn't add COFF base relocations yet, since we could use the COFF format code for this. There is actually quite a handful of code we could use from COFF plugin instead of completely re implementing it for PE. It might be better to add that to a separate issue & PR since this seems like an issue on its own. This might take a lot of work though since it would require changing a lot of the internal structures used by PE plugin, as well as re implementing handling of many things in COFF to make it compatible with PE as well.

librz/bin/format/pe/pe_relocs.c
```diff
@@ -0,0 +1,73 @@
+#include "pe.h"
```
XVilka, line 4: Add SPDX headers

librz/bin/format/pe/pe64_relocs.c
```diff
@@ -0,0 +1,2 @@
+#define RZ_BIN_PE64 1
```
XVilka, line 4: SPDX

librz/bin/format/pe/pe_relocs.c
```diff
@@ -0,0 +1,73 @@
...
+
+		rz_vector_push(relocs, &reloc);
+
+	} while (bytes_read < block_size);
+
+	return bytes_read;
+}
+
+static bool get_relocs_from_data_dir(RZ_NONNULL RzBinPEObj *bin, RZ_NONNULL RzVector /*<RzBinPeRelocEnt>*/ *relocs) {
+	// XXX: not sure if I need this here or not
```
XVilka: Not sure why the comment here, you can safely remove

librz/bin/p/bin_pe.inc
```diff
@@ -554,30 +638,16 @@ err:
...
-				imp = *it;
-				if (!strcmp(imp->name, "__security_init_cookie")) {
-					return true;
-				}
+	const RzPVector *imports_vec = imports(bf);
+	RzBinImport *imp;
+	if (imports_vec) {
+		rz_pvector_foreach (imports_vec, it) {
+			imp = *it;
+			// __security_init_cookie is a function imported from msvcrt.dll (libc) that when called initiliazes the stack canary. So if the function is imported, we can assume the binary uses stack canary.
```
XVilka: Split long comments into multiple lines

librz/bin/p/bin_pe.inc
```diff
@@ -554,30 +638,16 @@ err:
 }
 
 static int has_canary(RzBinFile *bf) {
-	// XXX: We only need imports here but this causes leaks, we need to wait for the below. This is a horrible solution!
 	// TODO: use O(1) when imports sdbized
```
XVilka: Remove this comment too

XVilka (changes requested):

XVilka:
> I deliberately didn't add COFF base relocations yet, since we could use the COFF format code for this. There is actually quite a handful of code we could use from COFF plugin instead of completely re implementing it for PE. It might be better to add that to a separate issue & PR since this seems like an issue on its own. This might take a lot of work though since it would require changing a lot of the internal structures used by PE plugin, as well as re implementing handling of many things in COFF to make it compatible with PE as well.

For now, I think some code duplication is fine. Sharing the code between two plugins is non-trivial at this point.

librz/bin/format/pe/pe_relocs.c
```diff
@@ -0,0 +1,74 @@
+// SPDX-FileCopyrightText: 2024 Roee Toledano <roeetoledano10@gmail.com>
+// SPDX-License-Identifier: LGPL-3.0-only
+#include "pe.h"
```
XVilka, line 4: Please leave 1 empty line between SPDX headers and first `#include`

librz/bin/format/pe/pe64_relocs.c
```diff
@@ -0,0 +1,4 @@
+// SPDX-FileCopyrightText: 2024 Roee Toledano <roeetoledano10@gmail.com>
+// SPDX-License-Identifier: LGPL-3.0-only
+#define RZ_BIN_PE64 1
```
XVilka, line 4: Same

librz/bin/format/pe/pe.h
```diff
@@ -69,6 +69,16 @@
 
 #define PE_SCN_ALIGN_MASK 0x00F00000
 
+typedef struct rz_bin_pe_reloc_block_t {
+	ut32 page_rva;
+	ut32 block_size;
+} RzBinPeRelocBlock;
+
+typedef struct rz_bin_pe_reloc_ent_t {
```
XVilka, line 83: Bitfields aren't portable. Please use masks instead. It's ok for keeping them in definition, but they should not be used for reading. It's especially important for reading files of different endianess from the host Rizin compiled for.

XVilka (changes requested):

librz/bin/format/pe/pe.h
wargio: use the le/be methods from `rz_util/rz_buf.h`

wargio (changes requested):
This needs to be tested via `distro-` branch before merge. @XVilka

librz/bin/format/pe/pe.h
Roeegg2: > use the le/be methods from `rz_util/rz_buf.h`

I've used `ble`. Is there something I did wrong?

librz/bin/format/pe/pe_relocs.c
```diff
@@ -0,0 +1,72 @@
...
+
+static unsigned long read_reloc_ent_from_block(RZ_NONNULL RzVector /*<RzBinPeRelocEnt>*/ *relocs, RzBuffer *b, int big_endian, ut32 block_size) {
+	unsigned int bytes_read = sizeof(RzBinPeRelocBlock);
+
+	do {
+		bytes_read += 2;
+		RzBinPeRelocEnt reloc;
+		ut8 ent_buf[sizeof(RzBinPeRelocEnt)];
+		rz_buf_read(b, ent_buf, sizeof(ent_buf));
+		reloc = rz_read_ble16(ent_buf, big_endian);
```
wargio: maybe use the `rz_buf_read_ble16_offset` which is a special api which updates automatically the offset/address parameter and returns true on successful read.

librz/bin/format/pe/pe_relocs.c
```diff
@@ -0,0 +1,72 @@
...
+	}
+	unsigned long total_bytes_to_read = bin->nt_headers->optional_header.DataDirectory[PE_IMAGE_DIRECTORY_ENTRY_BASERELOC].Size;
+	unsigned long total_bytes_read = 0;
+
+	do {
+		ut8 block_buf[sizeof(RzBinPeRelocBlock)];
+		rz_buf_read(b, block_buf, sizeof(block_buf));
+		RzBinPeRelocBlock block;
+		PE_READ_STRUCT_FIELD_L(block_buf, block, RzBinPeRelocBlock, page_rva, 32);
+		PE_READ_STRUCT_FIELD_L(block_buf, block, RzBinPeRelocBlock, block_size, 32);
```
wargio: you can do the same here using `rz_buf_read_bleXX_offset` and checking the return value.
you can do this in a static function like [this](https://github.com/rizinorg/rizin/blob/dev/librz/bin/p/bin_qnx.c#L7) and fail when returns false.

librz/bin/format/pe/pe_relocs.c
wargio: also be sure to check the boolean return value to fail gracefully.

librz/bin/format/pe/pe_relocs.c
```diff
@@ -0,0 +1,72 @@
...
+
+#include "pe.h"
+#include <rz_util/ht_uu.h>
+
+static unsigned long read_reloc_ent_from_block(RZ_NONNULL RzVector /*<RzBinPeRelocEnt>*/ *relocs, RzBuffer *b, int big_endian, ut32 block_size) {
+	unsigned int bytes_read = sizeof(RzBinPeRelocBlock);
+
+	do {
+		bytes_read += 2;
+		RzBinPeRelocEnt reloc;
```
wargio: always init this to 0.
```suggestion
		RzBinPeRelocEnt reloc = 0;
```

librz/bin/format/pe/pe_relocs.c
```diff
@@ -0,0 +1,72 @@
...
+			return false;
+		}
+	}
+	unsigned long total_bytes_to_read = bin->nt_headers->optional_header.DataDirectory[PE_IMAGE_DIRECTORY_ENTRY_BASERELOC].Size;
+	unsigned long total_bytes_read = 0;
+
+	do {
+		ut8 block_buf[sizeof(RzBinPeRelocBlock)];
+		rz_buf_read(b, block_buf, sizeof(block_buf));
+		RzBinPeRelocBlock block;
```
wargio:
```suggestion
		RzBinPeRelocBlock block = { 0 };
```

wargio (changes requested):
Please always initialize the variables to `0` or `NULL` so we can avoid random crashes and weird unintentional behavior.

librz/bin/format/pe/pe.h
```diff
@@ -69,6 +69,17 @@
...
 
+typedef struct rz_bin_pe_reloc_block_t {
+	ut32 page_rva;
+	ut32 block_size;
+} RzBinPeRelocBlock;
+
+// encoding is 12 bits for type and 4 bits for offset. 
+// See: https://learn.microsoft.com/en-us/windows/win32/debug/pe-format#the-reloc-section-image-only
+#define PE_RELOC_ENT_TYPE(val)   ((val) >> 12)
+#define PE_RELOC_ENT_OFFSET(val) ((val)&0b111111111111)
```
wargio: i'm not sure how `0b111111111111` is portable. use hex (`0xfff`).

Roeegg2: You're right, totally forgot about that. I'll fix it

Roeegg2:
Sorry for the ugly commit history. I could rebase interactive + squash with the old ones so it's easier to review.

Roeegg2:
All tests are green except flagspace tests. Because there is no symbol associated with base relocations, they have no name associated with them, so I'm not sure how we can use flags here

librz/bin/p/bin_pe.inc
```diff
@@ -535,48 +535,42 @@ static const char *get_reloc_type_name(ut32 reloc_type, ut32 machine_type) {
...
+	// NOTE: No patching is needed because base relocations (as their name suggests) patch relocations based off the delta
+	// between the executable expected load address, and the actual load address. And since Rizin always loads
+	// the binaries at their preferred base address, the delta is always 0 and so no patching is needed.
 	rz_return_val_if_fail(bf && bf->o && bf->o->bin_obj, NULL);
-	RzPVector *ret = NULL;
-
 	RzBinPEObj *bin = bf->o->bin_obj;
 
-	// TODO: patch relocs
+	if (bin->relocs == NULL) {
```
wargio:
```suggestion
	if (!bin->relocs) {
```

librz/bin/p/bin_pe.inc
```diff
@@ -535,48 +535,42 @@ static const char *get_reloc_type_name(ut32 reloc_type, ut32 machine_type) {
...
+	}
 
+	RzPVector *ret = NULL;
 	if (!(ret = rz_pvector_new((RzPVectorFree)rz_bin_reloc_free))) {
 		return NULL;
 	}
 
-	// stripped relocs means that the binary has no base relocations
-	if (!PE_(rz_bin_pe_is_stripped_relocs)((RzBinPEObj *)bin)) {
+	if (PE_(bin_pe_has_base_relocs)(bin)) {
```
wargio: invert this if and return early.

librz/bin/format/pe/pe.h
```diff
@@ -183,6 +187,7 @@ struct PE_(rz_bin_pe_obj_t) {
 	RzVector /*<RzBinPeRelocEnt>*/ *relocs;
```
wargio, line 188: convert this to a pvector.

Roeegg2: Why is regular vector not good here?

librz/bin/p/bin_pe.inc
```diff
@@ -464,37 +451,130 @@ static RzPVector /*<RzBinImport *>*/ *imports(RzBinFile *bf) {
...
+
+	if (!PE_(bin_pe_has_base_relocs)(bin)) {
+		return NULL;
+	}
+
+	RzBinPeRelocEnt *pe_reloc;
+	rz_vector_foreach (bin->relocs, pe_reloc) {
+		RzBinReloc *general_reloc = RZ_NEW0(RzBinReloc);
+		if (!general_reloc) {
+			RZ_LOG_WARN("Failed to convert PE reloc to general reloc\n");
```
wargio:
```suggestion
			RZ_LOG_ERROR("Failed to allocate PE RzBinReloc\n");
```

librz/bin/p/bin_pe.inc
```diff
@@ -464,37 +451,130 @@ static RzPVector /*<RzBinImport *>*/ *imports(RzBinFile *bf) {
...
+			rz_bin_reloc_free(general_reloc);
+			continue;
```
wargio: if alloc fails, then there is no memory left.
You should probably do the following instead of calling `continue`:

```
rz_vector_free(bin->relocs);
return ret;
```

librz/bin/format/pe/pe_relocs.c
```diff
@@ -0,0 +1,67 @@
...
+#include "pe.h"
+
+bool PE_(bin_pe_has_base_relocs)(RZ_NONNULL RzBinPEObj *bin) {
+	rz_return_val_if_fail(bin, false);
+
+	return bin->relocs && (rz_vector_len(bin->relocs) > 0);
+}
+
+static bool read_reloc_ent_from_block(RZ_NONNULL RzVector /*<RzBinPeRelocEnt>*/ *relocs, RzBuffer *b, RzBinPeRelocBlock *block, ut64 *offset, const int big_endian) {
+	const ut32 reloc_block_end = *offset + block->block_size - 8; // block size includes the size of the next blocks entry, which is 8 bytes long
```
XVilka: Move comment to another line (before probably) to avoid such a long line

librz/bin/format/pe/pe_relocs.c
```diff
@@ -0,0 +1,67 @@
...
+
+	return true;
+}
+
+static bool get_relocs_from_data_dir(RZ_NONNULL RzBinPEObj *bin, RZ_BORROW RZ_NONNULL RzVector /*<RzBinPeRelocEnt>*/ *relocs) {
+	RzBuffer *b = bin->b;
+	const st64 o_addr = rz_buf_tell(b);
+
+	// get offset in file of first reloc block
+	ut64 offset = PE_(bin_pe_rva_to_paddr)(bin, bin->nt_headers->optional_header.DataDirectory[PE_IMAGE_DIRECTORY_ENTRY_BASERELOC].VirtualAddress);
```
XVilka: Check first if `nt_headers` and `optional_header` aren't NULL. Also, make sure that the array access doesn't go out of bounds.

librz/bin/format/pe/pe_relocs.c
```diff
@@ -0,0 +1,67 @@
...
+	rz_buf_seek(b, o_addr, RZ_BUF_SET);
+
+	return true;
+}
+
+int PE_(bin_pe_init_relocs)(RZ_NONNULL RzBinPEObj *bin) {
+	rz_return_val_if_fail(bin, false);
+
+	RzVector *ret = rz_vector_new(sizeof(RzBinPeRelocEnt), NULL, NULL);
+	bin->relocs = ret;
```
XVilka: Probably better to put it after the NULL check, and specify `bin->relocs = NULL` in the NULL case to avoid misunderstanding for people and static analyzers.

librz/bin/p/bin_pe.inc
```diff
@@ -464,37 +451,129 @@ static RzPVector /*<RzBinImport *>*/ *imports(RzBinFile *bf) {
...
+		RZ_LOG_WARN("Unknown PE relocation type number: %d\n", reloc_type);
+		return "IMAGE_REL_BASED_UNKNOWN";
+	}
+}
+
 static RzPVector /*<RzBinReloc *>*/ *relocs(RzBinFile *bf) {
-	struct PE_(rz_bin_pe_obj_t) *obj = bf->o->bin_obj;
-	if (obj) {
-		return obj->relocs;
+	// NOTE: No patching is needed because base relocations (as their name suggests) patch relocations based off the delta
```
XVilka: Should it be a comment *before* the function?

librz/bin/p/bin_pe.inc
```diff
@@ -464,37 +451,129 @@ static RzPVector /*<RzBinImport *>*/ *imports(RzBinFile *bf) {
...
+		return NULL;
 	}
-	return NULL;
+
+	RzPVector *ret = NULL;
+	if (!(ret = rz_pvector_new((RzPVectorFree)rz_bin_reloc_free))) {
+		return NULL;
+	}
+
+	if (!PE_(bin_pe_has_base_relocs)(bin)) {
```
XVilka, line 554: Memory leak of `ret`?

XVilka (changes requested):
Good changes, but why did you remove MDMP relocs completely?

XVilka:
I think it should be in 0.8.0 release, assuming it will be finished in time.

wargio:
Ok, i checked the documentation and saw how other tools shows this, so the changes are ok, just give it a name like `pe_<va offset>` so rizin can create the flags `reloc.pe_<va offset>`, when you define the name use `rz_str_newf("pe_%08" PFMT64x", offset)` (change "offset" to the correct variable holding the vaddr).

please also fix the mdmp relocs that you have removed.

librz/bin/format/pe/pe_relocs.c
Roeegg2: Fixed. I just didn't add the bounds checking, since the `DataDirectory` array is always `PE_IMAGE_DIRECTORY_ENTRIES` length, so it's never out of bounds.

XVilka (approved):

librz/bin/format/mdmp/mdmp_pe.c
```diff
@@ -121,24 +116,10 @@ RzPVector /*<RzBinImport *>*/ *PE_(rz_bin_mdmp_pe_get_imports)(struct PE_(rz_bin
...
-		}
-#ifdef RZ_BIN_PE64
-		rel->type = RZ_BIN_RELOC_64;
-#else
-		rel->type = RZ_BIN_RELOC_32;
-#endif
 		offset = imports[i].vaddr;
 		if (offset > pe_bin->vaddr) {
 			offset -= pe_bin->vaddr;
 		}
```
wargio: this seems to be used only by the removed code. probably can be deleted too, including the variable `offset`

librz/bin/format/pe/pe.h
```diff
@@ -69,6 +69,21 @@
 
 #define PE_SCN_ALIGN_MASK 0x00F00000
 
+typedef struct rz_bin_pe_reloc_block_t {
+	ut32 page_rva;
+	ut32 block_size;
```
wargio: please give a short description of what these fields are.
Example of formatting

```suggestion
	ut32 page_rva; ///< PE RVA see doc https://bla
	ut32 block_size; ///< PE reloc block size for xxxxx
```

librz/bin/format/pe/pe.h
```diff
@@ -165,14 +180,14 @@ struct PE_(rz_bin_pe_obj_t) {
...
-	int endian;
 	bool verbose;
 	int big_endian;
 	RzList /*<Pe_image_rich_entry *>*/ *rich_entries;
-	RzPVector /*<RzBinReloc *>*/ *relocs;
 	RzList /*<rz_pe_resource *>*/ *resources;
+	RzVector /*<RzBinPeRelocEnt>*/ *relocs;
 	const char *file;
 	RzBuffer *b;
+	RzBuffer *buf_patched; // overlay over the original file with relocs patched
```
wargio: is this used? cannot find any reference anywhere to its usage.

librz/bin/p/bin_pe.inc
```diff
@@ -554,30 +638,18 @@ err:
 }
 
 static int has_canary(RzBinFile *bf) {
-	// XXX: We only need imports here but this causes leaks, we need to wait for the below. This is a horrible solution!
 	// TODO: use O(1) when imports sdbized
```
wargio: this comment about sdb is still here.

librz/bin/format/pe/pe.h
Roeegg2: It's not, I added it to implement patching (which we ended up not needing)

librz/bin/format/pe/pe.h
Roeegg2: I added a link to the base relocations documentation section of the specification, I will add detailed comments to each struct field too though

PR #4740, opened 2024-11-25
Improved performance of rz_bv_copy_nbits and rz_bv_set_range

### LINKED ISSUE #4716: Enhance performance of `rz_bv_copy_nbits` and `rz_bv_set_range`

Performance could be better and we save energy which would make us better humans.

Both functions do something like this:
```c
for (ut32 i = 0; i < nbit; ++i) {
   bool c = rz_bv_get(src, src_start_pos + i); // Not in case of rz_bv_set_range()
   rz_bv_set(dst, dst_start_pos + i, c);
}
```

At least if we deal with small bit vectors (`<= 64bit`) we could do this with shift, AND + OR.
This would save a bunch of calls and other shifts and ORs.

For larger bit vectors we could check if the range is large enough (e.g. 8 bits or more).
Then set the first bits until the rest of them is aligned to a multiple of a byte (for `RzBitVector->large_a`) and assign the rest of them via `memcopy`. 

None really.

**Additional context**

This would give us quite some computation back. In my use case this loop consumes around 2.3%.
I also think this is a required improvement before we merge resource intensive analysis algorithms based on it.

rajRishi22:

**Detailed description**
This pull request optimizes the performance of the rz_bv_set_range function. The original implementation iterated through bits one at a time, leading to inefficiencies for large ranges. The updated implementation:

Processes aligned chunks of bits using system word size for faster operations.
Dynamically adjusts chunk size for different architectures (e.g., 32-bit, 64-bit).
Handles unaligned prefix and suffix bits separately while optimizing the main loop.
Adds robust boundary validation to ensure correctness.
This change reduces iteration overhead and improves performance while maintaining compatibility and correctness.

**Test plan**

1. Verify functionality for small ranges, large ranges, and edge cases (unaligned ranges).
2. Test on various architectures to confirm portability (32-bit, 64-bit).
3. Use unit tests to ensure the results match the original functionality.
4. Check for any regressions with existing test suites.

**Closing issues**
Partially addresses #4716 
...

### DIFF
--- a/librz/util/bitvector.c
+++ b/librz/util/bitvector.c
@@ -4,6 +4,7 @@
 #include "rz_util.h"
 #include <stdlib.h>
 #include <stdio.h>
+#include <limits.h>
 
 #define NELEM(N, ELEMPER) ((N + (ELEMPER)-1) / (ELEMPER))
 #define BV_ELEM_SIZE      8U
@@ -209,25 +210,62 @@ RZ_API ut32 rz_bv_copy(RZ_NONNULL const RzBitVector *src, RZ_NONNULL RzBitVector
  * \param nbit ut32, control the size of copy (in bits)
  * \return copied_size ut32, Actual copied size
  */
+
 RZ_API ut32 rz_bv_copy_nbits(RZ_NONNULL const RzBitVector *src, ut32 src_start_pos, RZ_NONNULL RzBitVector *dst, ut32 dst_start_pos, ut32 nbit) {
-	rz_return_val_if_fail(src && dst, 0);
+    rz_return_val_if_fail(src && dst, 0);
 
-	ut32 max_nbit = RZ_MIN((src->len - src_start_pos),
-		(dst->len - dst_start_pos));
+    // Determine the chunk size (word size) dynamically
+    const ut32 chunk_size = sizeof(unsigned long) * CHAR_BIT; // Word size in bits
+    ut32 max_nbit = RZ_MIN((src->len - src_start_pos), (dst->len - dst_start_pos));
 
-	// prevent overflow
-	if (max_nbit < nbit) {
-		return 0;
-	}
+    if (max_nbit < nbit) {
+        return 0;
+    }
 
-	// normal case here
-	for (ut32 i = 0; i < nbit; ++i) {
-		bool c = rz_bv_get(src, src_start_pos + i);
-		rz_bv_set(dst, dst_start_pos + i, c);
-	}
+    ut32 bits_copied = 0;
 
-	return nbit;
+    // Handle unaligned prefix
+    while ((src_start_pos % chunk_size != 0 || dst_start_pos % chunk_size != 0) && nbit > 0) {
+        bool bit = rz_bv_get(src, src_start_pos++);
+        rz_bv_set(dst, dst_start_pos++, bit);
+        --nbit;
+        ++bits_copied;
+    }
+
+    // Process aligned chunks
+    while (nbit >= chunk_size) {
+        // Get chunks from the source and destination
+        unsigned long src_chunk = rz_bv_get_chunk(src, src_start_pos / chunk_size);
+        unsigned long dst_chunk = rz_bv_get_chunk(dst, dst_start_pos / chunk_size);
+
+        // Create a mask for the bits to copy
+        unsigned long mask = (1UL << chunk_size) - 1;
+        if (nbit < chunk_size) {
+            mask = (1UL << nbit) - 1;
+        }
+
+        // Merge chunks using the optimized approach
+        unsigned long result = dst_chunk ^ ((dst_chunk ^ src_chunk) & mask);
+        rz_bv_set_chunk(dst, dst_start_pos / chunk_size, result);
+
+        src_start_pos += chunk_size;
+        dst_start_pos += chunk_size;
+        nbit -= chunk_size;
+        bits_copied += chunk_size;
+    }
+
+    // Handle remaining unaligned suffix bits
+    while (nbit > 0) {
+        bool bit = rz_bv_get(src, src_start_pos++);
+        rz_bv_set(dst, dst_start_pos++, bit);
+        --nbit;
+        ++bits_copied;
+    }
+
+    return bits_copied;
 }
+
+
 
 /**
  * Return a new bitvector prepended with bv with n zero bits
@@ -1480,17 +1518,42 @@ RZ_API ut64 rz_bv_to_ut64(RZ_NONNULL const RzBitVector *x) {
  * \return return true if success, else return false
  */
 RZ_API bool rz_bv_set_range(RZ_NONNULL RzBitVector *bv, ut32 pos_start, ut32 pos_end, bool b) {
-	rz_return_val_if_fail(bv, false);
-	if (pos_start > bv->len - 1 || pos_end > bv->len - 1) {
-		return false;
-	}
+    rz_return_val_if_fail(bv, false);
 
-	for (ut32 i = pos_start; i <= pos_end; ++i) {
-		rz_bv_set(bv, i, b);
-	}
+    if (pos_start > bv->len - 1 || pos_end > bv->len - 1 || pos_start > pos_end) {
+        return false;
+    }
 
-	return true;
+    // Determine the chunk size dynamically
+    const ut32 chunk_size = sizeof(unsigned long) * CHAR_BIT;
+
+    // Handle unaligned prefix bits
+    while (pos_start < pos_end && pos_start % chunk_size != 0) {
+        rz_bv_set(bv, pos_start++, b);
+    }
+
+    // Process aligned chunks
+    if (pos_start < pos_end) {
+        ut32 chunk_start = pos_start / chunk_size;
+        ut32 chunk_end = pos_end / chunk_size;
+
+        unsigned long fill_value = b ? ~0UL : 0UL;
+
+        for (ut32 i = chunk_start; i < chunk_end; ++i) {
+            rz_bv_set_chunk(bv, i, fill_value);
+        }
+
+        pos_start = chunk_end * chunk_size;
+    }
+
+    // Handle remaining unaligned suffix bits
+    while (pos_start <= pos_end) {
+        rz_bv_set(bv, pos_start++, b);
+    }
+
+    return true;
 }
+
 
 /**
  * check if bitvector's bits are all set to bit 1
--- a/librz/util/bitvector.c
+++ b/librz/util/bitvector.c
@@ -6,11 +6,9 @@
 #include <stdlib.h>
 #include <stdio.h>
 
-
 #define NELEM(N, ELEMPER) ((N + (ELEMPER)-1) / (ELEMPER))
 #define BV_ELEM_SIZE      8U
-#define SIZE_OF_UNSIGNED_LONG (sizeof(unsigned long))
-
+#define RZ_BV_CHUNK_SIZE  (sizeof(unsigned long) * CHAR_BIT)
 
 // optimization for reversing 8 bits which uses 32 bits
 // https://graphics.stanford.edu/~seander/bithacks.html#ReverseByteWith32Bits
@@ -141,7 +139,6 @@ RZ_API RZ_OWN char *rz_bv_as_hex_string(RZ_NONNULL const RzBitVector *bv, bool p
 	if (!str) {
 		return NULL;
 	}
-
 	str[0] = '0';
 	str[1] = 'x';
 	ut32 j = 2;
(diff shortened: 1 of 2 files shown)


librz/util/bitvector.c
```diff
@@ -209,26 +210,63 @@ RZ_API ut32 rz_bv_copy(RZ_NONNULL const RzBitVector *src, RZ_NONNULL RzBitVector
  * \param nbit ut32, control the size of copy (in bits)
  * \return copied_size ut32, Actual copied size
  */
+
 RZ_API ut32 rz_bv_copy_nbits(RZ_NONNULL const RzBitVector *src, ut32 src_start_pos, RZ_NONNULL RzBitVector *dst, ut32 dst_start_pos, ut32 nbit) {
-	rz_return_val_if_fail(src && dst, 0);
+    rz_return_val_if_fail(src && dst, 0);
```
Rot127: Please run `clang-format-16` on those files. The indentation gets fixed this way.

librz/util/bitvector.c
```diff
@@ -209,26 +210,63 @@ RZ_API ut32 rz_bv_copy(RZ_NONNULL const RzBitVector *src, RZ_NONNULL RzBitVector
...
+        nbit -= chunk_size;
+        bits_copied += chunk_size;
+    }
+
+    // Handle remaining unaligned suffix bits
+    while (nbit > 0) {
+        bool bit = rz_bv_get(src, src_start_pos++);
+        rz_bv_set(dst, dst_start_pos++, bit);
+        --nbit;
+        ++bits_copied;
```
Rot127: No need to increment `bits_copied`. You can simply save the original `nbit` value and return it.
As I see it there is no other return value then `nbit` or `0`.

librz/util/bitvector.c
```diff
@@ -4,6 +4,7 @@
 #include "rz_util.h"
 #include <stdlib.h>
 #include <stdio.h>
+#include <limits.h>
```
Rot127:
```suggestion
#include <rz_types.h>
```

`limits.h` should be included already there.

librz/util/bitvector.c
```diff
@@ -209,26 +210,63 @@ RZ_API ut32 rz_bv_copy(RZ_NONNULL const RzBitVector *src, RZ_NONNULL RzBitVector
...
  */
+
 RZ_API ut32 rz_bv_copy_nbits(RZ_NONNULL const RzBitVector *src, ut32 src_start_pos, RZ_NONNULL RzBitVector *dst, ut32 dst_start_pos, ut32 nbit) {
-	rz_return_val_if_fail(src && dst, 0);
+    rz_return_val_if_fail(src && dst, 0);
 
-	ut32 max_nbit = RZ_MIN((src->len - src_start_pos),
-		(dst->len - dst_start_pos));
+    // Determine the chunk size (word size) dynamically
+    const ut32 chunk_size = sizeof(unsigned long) * CHAR_BIT; // Word size in bits
```
Rot127: This is fine for now. But we should eventually change it against `uint_fast8_t, uint_fast16_t, etc.`.
But these need to be added in `rz_base_types.h` as `typedef`.

librz/util/bitvector.c
```diff
@@ -209,26 +210,63 @@ RZ_API ut32 rz_bv_copy(RZ_NONNULL const RzBitVector *src, RZ_NONNULL RzBitVector
...
 
-	// normal case here
-	for (ut32 i = 0; i < nbit; ++i) {
-		bool c = rz_bv_get(src, src_start_pos + i);
-		rz_bv_set(dst, dst_start_pos + i, c);
-	}
+    ut32 bits_copied = 0;
+
+    // Handle unaligned prefix
+    while ((src_start_pos % chunk_size != 0 || dst_start_pos % chunk_size != 0) && nbit > 0) {
```
Rot127: You can make `(src_start_pos % chunk_size != 0 || dst_start_pos % chunk_size != 0)` an `if` condition and then loop around `nbit > 0`.
This way you can return early within the `if` body.

librz/util/bitvector.c
```diff
@@ -209,26 +210,63 @@ RZ_API ut32 rz_bv_copy(RZ_NONNULL const RzBitVector *src, RZ_NONNULL RzBitVector
...
+
+    // Process aligned chunks
+    while (nbit >= chunk_size) {
+        // Get chunks from the source and destination
+        unsigned long src_chunk = rz_bv_get_chunk(src, src_start_pos / chunk_size);
+        unsigned long dst_chunk = rz_bv_get_chunk(dst, dst_start_pos / chunk_size);
 
-	return nbit;
+        // Create a mask for the bits to copy
+        unsigned long mask = (1UL << chunk_size) - 1;
```
Rot127: This is equivalent to `UINT32_MAX` (or whatever type we settle on later).

librz/util/bitvector.c
Rot127: `sizeof()` is a macro. So you can move this to a constant.

librz/util/bitvector.c
```diff
@@ -1480,18 +1518,43 @@ RZ_API ut64 rz_bv_to_ut64(RZ_NONNULL const RzBitVector *x) {
...
-	for (ut32 i = pos_start; i <= pos_end; ++i) {
-		rz_bv_set(bv, i, b);
-	}
+    if (pos_start > bv->len - 1 || pos_end > bv->len - 1 || pos_start > pos_end) {
+        return false;
+    }
 
-	return true;
+    // Determine the chunk size dynamically
+    const ut32 chunk_size = sizeof(unsigned long) * CHAR_BIT;
```
Rot127: Same as above. Please move it to a constant/`#define`.

Rot127 (changes requested):
Nice start!
I can imagine we can even optimize the unaligned cases. But let's do this later.

I also changed your PR message. Because it closes https://github.com/rizinorg/rizin/issues/4716 only partially (missing unaligned cases).

You can add test cases in `test/unit/test_bitvector.c`.
Once everything is implemented and passes, we can run the Travis CI to test it on big endian machines.

librz/util/bitvector.c
Rot127: cc @XVilka @thestr4ng3r Regarding the `fast` types.

XVilka: First we need to make sure all our supported systems have it - at least running on `dist-` branch is required.

Rot127: It is in the `C99` standard. But if they don't support it we could just default to a normal `uint` type.

rajRishi22: Hi @Rot127  and @XVilka  , first of all thanks a lot for such a detailed review of my code ,as a beginner  it was extremely helpful for me . I have tried to address all problems mentioned in this review.
Do i need to create rz_base_types.h for now , as i could not find that file ?

librz/util/bitvector.c
```diff
@@ -209,26 +213,62 @@ RZ_API ut32 rz_bv_copy(RZ_NONNULL const RzBitVector *src, RZ_NONNULL RzBitVector
...
  */
+
 RZ_API ut32 rz_bv_copy_nbits(RZ_NONNULL const RzBitVector *src, ut32 src_start_pos, RZ_NONNULL RzBitVector *dst, ut32 dst_start_pos, ut32 nbit) {
-	rz_return_val_if_fail(src && dst, 0);
+    rz_return_val_if_fail(src && dst, 0);
 
-	ut32 max_nbit = RZ_MIN((src->len - src_start_pos),
-		(dst->len - dst_start_pos));
+    // Determine the chunk size (word size) dynamically
+    const ut32 chunk_size = SIZE_OF_UNSIGNED_LONG * CHAR_BIT; // Word size in bits
```
Rot127: You can move this whole expression into a `#define`. Macros allow some basic math

```
#define RZ_BV_CHUNK_SIZE (sizeof(unsigned long) * CHAR_BIT)
```

librz/util/bitvector.c
```diff
@@ -209,26 +213,62 @@ RZ_API ut32 rz_bv_copy(RZ_NONNULL const RzBitVector *src, RZ_NONNULL RzBitVector
...
+        unsigned long src_chunk = rz_bv_get_chunk(src, src_start_pos / chunk_size);
+        unsigned long dst_chunk = rz_bv_get_chunk(dst, dst_start_pos / chunk_size);
+
+        // Create a mask for the bits to copy
+		unsigned long mask = UINT32_MAX;
+        if (nbit < chunk_size) {
+            mask = (1UL << nbit) - 1;
+        }
+
+        // Merge chunks using the optimized approach
```
Rot127: Add the link here.

librz/util/bitvector.c
```diff
@@ -209,26 +213,62 @@ RZ_API ut32 rz_bv_copy(RZ_NONNULL const RzBitVector *src, RZ_NONNULL RzBitVector
...
+        unsigned long result = dst_chunk ^ ((dst_chunk ^ src_chunk) & mask);
+        rz_bv_set_chunk(dst, dst_start_pos / chunk_size, result);
+
+        src_start_pos += chunk_size;
+        dst_start_pos += chunk_size;
+        nbit -= chunk_size;
```
Rot127: `nbit` underflows here if `nbit < chunk_size`.

Rot127 (changes requested):

librz/util/bitvector.c
```diff
@@ -209,24 +211,60 @@ RZ_API ut32 rz_bv_copy(RZ_NONNULL const RzBitVector *src, RZ_NONNULL RzBitVector
...
+		// Merge chunks using the optimized approach , reference : https://graphics.stanford.edu/~seander/bithacks.html#MaskedMerge
+		unsigned long result = dst_chunk ^ ((dst_chunk ^ src_chunk) & mask);
+		rz_bv_set_chunk(dst, dst_start_pos / RZ_BV_CHUNK_SIZE, result);
+
+		src_start_pos += RZ_BV_CHUNK_SIZE;
+		dst_start_pos += RZ_BV_CHUNK_SIZE;
+		if (nbit >= RV_BV_CHUNK_SIZE)
+			nbit -= RZ_BV_CHUNK_SIZE;
+		else
+			break;
```
wargio:
```suggestion
		if (nbit < RV_BV_CHUNK_SIZE) {
			break;
		}
		nbit -= RZ_BV_CHUNK_SIZE;
```

librz/util/bitvector.c
```diff
@@ -209,24 +211,60 @@ RZ_API ut32 rz_bv_copy(RZ_NONNULL const RzBitVector *src, RZ_NONNULL RzBitVector
...
+			rz_bv_set(dst, dst_start_pos++, bit);
+			--nbit;
+		}
+	}
+
+	// Process aligned chunks
+	while (nbit >= RZ_BV_CHUNK_SIZE) {
+		// Get chunks from the source and destination
+		unsigned long src_chunk = rz_bv_get_chunk(src, src_start_pos / RZ_BV_CHUNK_SIZE);
+		unsigned long dst_chunk = rz_bv_get_chunk(dst, dst_start_pos / RZ_BV_CHUNK_SIZE);
```
wargio: maybe i'm missing something, but this function seems not in the code. also please use `ut32` as type.

librz/util/bitvector.c
```diff
@@ -209,24 +211,60 @@ RZ_API ut32 rz_bv_copy(RZ_NONNULL const RzBitVector *src, RZ_NONNULL RzBitVector
...
+
+		// Create a mask for the bits to copy
+		unsigned long mask = UINT32_MAX;
```
wargio:
```suggestion
		ut32 mask = UT32_MAX;
```

librz/util/bitvector.c
```diff
@@ -2,11 +2,14 @@
...
 
 #include "rz_util.h"
+#include <rz_types.h>
 #include <stdlib.h>
 #include <stdio.h>
 
 #define NELEM(N, ELEMPER) ((N + (ELEMPER)-1) / (ELEMPER))
 #define BV_ELEM_SIZE      8U
+#define RZ_BV_CHUNK_SIZE  (sizeof(unsigned long) * CHAR_BIT)
+#define SIZE_OF_UNSIGNED_LONG sizeof(unsigned long)
```
wargio: ut32 not unsigned long

wargio (changes requested):
missing implementation of `rz_bv_get_chunk`

librz/util/bitvector.c
```diff
@@ -202,6 +202,79 @@ RZ_API ut32 rz_bv_copy(RZ_NONNULL const RzBitVector *src, RZ_NONNULL RzBitVector
...
+		// The chunk is aligned to a 32-bit boundary
+		bv->data[word_idx] = chunk;
+	} else {
+		// The chunk spans across two 32-bit words
+		ut32 first_word = bv->data[word_idx];
+		ut32 second_word = bv->data[word_idx + 1];
+
+		// Clear the bits in the current chunk positions
+		first_word &= ~(0xFFFFFFFF >> bit_offset); // Clear the upper bits
+		second_word &= (0xFFFFFFFF >> (32 - bit_offset)); // Clear the lower bits
```
wargio, line 266:
```suggestion
		first_word &= ~(UT32_MAX >> bit_offset); // Clear the upper bits
		second_word &= (UT32_MAX >> (32 - bit_offset)); // Clear the lower bits
```

wargio (changes requested):
You are touching something very critical within rizin, so please provide benchmark and C test for correctness and edge cases.

Rot127:
Are you still interested in finishing this? If yes, please fix the build and add tests.
Otherwise, I would close it in the next days.

rajRishi22:
Sorry , i had my college exams so i was busy , i will start working on this issue again from today.

XVilka:
This code looks like AI-generated. Closing it.

PR #4742, opened 2024-11-25
Add / (search) commands to newshell + rewrite of rz_search

### LINKED ISSUE #1909: Arbitrarily high hit counts / high allocation counts hurts search performance, refactor search hits to use rz_vector?
May be more related to cutter's gui because it looks like the api has a maximum search limit / count and perhaps that'd solve the issue in commandline. However, without knowing that in advance you can hang the application by running a search which may have too many results. The issue stems from usage of a list structure which allocates for every saved result in `rz_search_t`.

The problem function in question would be `rz_search_hit_new`. 

Edit: on second look it appears the api is being used with a callback which terminates early and skips the allocation I thought was the original issue. This might impact uses of this part of the api without a provided callback, they're likely slower than it should be. Whatever's causing the application hang eludes me.

Key part here would be to refactor the search functionality to use a vector like (or probably better rope like) structure. This should improve search timings considerably since time won't be wasted waiting on the system for memory. There appears to be a header for vectors in the library, it probably could be put to use here. A well designed rope structure would probably be better.

EG: instead of
```c
typedef struct rz_search_t {
	int n_kws; // hit${n_kws}_${count}
	int mode;
	ut32 pattern_size;
	ut32 string_min; // max length of strings for RZ_SEARCH_STRING
	ut32 string_max; // min length of strings for RZ_SEARCH_STRING
	void *data; // data used by search algorithm
	void *user; // user data passed to callback
	RzSearchCallback callback;
	ut64 nhits;
	ut64 maxhits; // search.maxhits
	RzList *hits;
	int distance;
	int inverse;
	bool overlap; // whether two matches can overlap
	int contiguous;
	int align;
	int (*update)(struct rz_search_t *s, ut64 from, const ut8 *buf, int len);
	RzList *kws; // TODO: Use rz_search_kw_new ()
	RzIOBind iob;
	char bckwrds;
} RzSearch;
```

something like:
```c
typedef struct rz_search_t {
	int n_kws; // hit${n_kws}_${count}
	int mode;
	ut32 pattern_size;
	ut32 string_min; // max length of strings for RZ_SEARCH_STRING
	ut32 string_max; // min length of strings for RZ_SEARCH_STRING
	void *data; // data used by search algorithm
	void *user; // user data passed to callback
	RzSearchCallback callback;
	ut64 nhits;
	ut64 maxhits; // search.maxhits
	RzVector *hits; // this probably doesn't need to be a pointer? also nhits would be hits->len
	int distance;
	int inverse;
	bool overlap; // whether two matches can overlap
	int contiguous;
	int align;
	int (*update)(struct rz_search_t *s, ut64 from, const ut8 *buf, int len);
	RzList *kws; // TODO: Use rz_search_kw_new ()
	RzIOBind iob;
	char bckwrds;
} RzSearch;
```

Since results are addresses there's no need to allocate for every 8 byte address, and since any information related to pointers is auxiliary to the pointer there's and pointers are by definition unique  there's plenty of opportunity to use hashtables to grab any interesting auxiliary information about the pointer later. It seems like in addition to the matched address currently a pointer to the search used is saved. There's probably a good amount of memory to save by not storing that pointer for each hit.

Another idea would be a custom allocator, seems like that would considerably more difficult, and I suspect a better data structure is the better approach.

**Additional context**
A fairly easy way to test this is to do what I did by accident, which is to debug a program and search for 0 effectively everywhere.

It also looks as if the maxhits variable can be used to reserve space in `rz_search_begin`. This in theory should reduce allocations wherever this api is used.

wargio:

**Detailed description**

This is a complete rewrite of the legacy code which is too complicated to import into new-shell.

The goal is to refactor `rz_search` and implement the search command properly using pre-existing code like search of strings via rz_utils and more.

closes https://github.com/rizinorg/rizin/issues/1909


librz/include/rz_search.h
```diff
@@ -12,117 +12,46 @@ extern "C" {
...
+RZ_API bool rz_search_collection_bytes_add(RZ_NONNULL RzSearchCollection *col, RZ_NONNULL const char *metadata, RZ_NONNULL const ut8 *bytes, RZ_NULLABLE const ut8 *mask, size_t length);
+RZ_API bool rz_search_collection_bytes_add_pattern(RZ_NONNULL RzSearchCollection *col, RZ_NONNULL const char *hex_pattern);
+
+RZ_API RZ_OWN RzSearchCollection *rz_search_collection_magic(RZ_NONNULL const char *magic_dir);
+
+RZ_API bool rz_search_collection_match_any(RZ_NULLABLE RzSearchCollection *sc, RZ_NONNULL const ut8 *buffer, size_t length);
+RZ_API void rz_search_collection_free(RZ_NULLABLE RzSearchCollection *sc);
+RZ_API void rz_search_hit_free(RZ_NULLABLE RzSearchHit *hit);
+
+RZ_API RZ_OWN RzList /*<RzSearchHit *>*/ *rz_search_run(RZ_NONNULL RzSearchOpt *opt, RZ_NONNULL RzSearchCollection *col, RZ_NONNULL RzIO *io, RZ_NONNULL RzList /*<RzIOMap *>*/ *search_in);
```
Rot127, line 58: Maybe changing the `search_in` parameter to `void *` and infer the type based on some flag in `opt`. This way we can use this function later to search other things then IO.

librz/search/search_internal.h
```diff
@@ -0,0 +1,43 @@
...
+#define RZ_SEARCH_PRIVATE_KEY_LENGTH 11
+#define RZ_SEARCH_MIN_BUFFER_SIZE    512u
+#define RZ_SEARCH_MAX_HEX_PATTERN    UT16_MAX
+
+typedef void (*RzSearchFiniCallback)(void *user);
+typedef bool (*RzSearchOverCallback)(RzPVector /*<void *>*/ *collection, ut64 address, const ut8 *buffer, size_t size, RzThreadQueue *hits);
+
+struct rz_search_collection_t {
+	RzPVector /*<void *>*/ *collection; ///< Collection of elements to search in a buffer
+	RzSearchOverCallback search_over; ///< Collection search over the collection callback
```
Rot127: Docs are a little confusing. Is the `search_over` callback called for every element in the collection above?

librz/include/rz_search.h
```diff
@@ -12,117 +12,46 @@ extern "C" {
...
-	ut64 last; // last hit hint
-} RzSearchKeyword;
+typedef struct rz_search_opt_t RzSearchOpt;
+
+typedef struct rz_search_collection_t RzSearchCollection;
 
 typedef struct rz_search_hit_t {
-	RzSearchKeyword *kw;
-	ut64 addr;
+	char *metadata; ///< Metadata for extra details (can be NULL)
```
Rot127, line 23: This seems to save mostly the type of the search hit or the things to search for (I saw `aes` and `regex`). Maybe changing it to an enum (`RzSearchMetaDataType type`) and a add a `void *meta_data` to save more complex stuff. This way we can save allocation of simple strings.

librz/core/cmd_descs/cmd_search.yaml
```diff
@@ -3,8 +3,174 @@
 ---
 name: cmd_search
 commands:
+  - name: "//"
+    summary: Repeat the last seach
```
Rot127:
```suggestion
    summary: Repeat the last search
```

librz/core/cconfig.c
```diff
@@ -3745,16 +3792,18 @@ RZ_API int rz_core_config_init(RzCore *core) {
 	SETOPTIONS(n, "auto", "rosections", "raw", NULL);
 
 	/* search */
-	SETCB("search.contiguous", "true", &cb_contiguous, "Accept contiguous/adjacent search hits");
-	SETICB("search.align", 0, &cb_searchalign, "Only catch aligned search hits");
+	SETB("search.progress", false, "Shows search progress (true: enable, false: disable)");
```
Rot127, line 3770: :heart:

librz/core/cmd/cmd_print.c
```diff
@@ -4371,6 +4371,36 @@ RZ_IPI RzCmdStatus rz_print_pattern8_handler(RzCore *core, int argc, const char
...
+		buf[i]++;
+		if (buf[i] && isalpha(buf[i])) {
+			break;
+		}
+		if (!buf[i]) {
+			i++;
+			continue;
+		}
+	}
+	// may overflow/hang/end/stop/whatever here
```
Rot127, line 4386: When running multi-threaded?

librz/include/rz_search.h
wargio: Not so sure about this, but maybe.

librz/search/search_internal.h
wargio: No, the function takes the vector and uses it

librz/include/rz_search.h
wargio: I added it for the flags so we know it's hit.regex or hit.rsa etc...

librz/search/collection.c
```diff
@@ -0,0 +1,101 @@
...
+
+/**
+ * \brief      Initialize a new RzSearchCollection
+ *
+ * \param[in]  search_over  The search over callback to set
+ * \param[in]  free         The free function to set
+ *
+ * \return     On success returns a valid pointer, otherwise NULL.
+ */
+RZ_IPI RZ_OWN RzSearchCollection *rz_search_collection_new(RZ_NONNULL RzSearchOverCallback search_over, RZ_NULLABLE RzPVectorFree free) {
```
Rot127:
```suggestion
RZ_IPI RZ_OWN RzSearchCollection *rz_search_collection_new(RZ_NONNULL RzSearchOverCallback search_over, RZ_NULLABLE RzPVectorFree element_free) {
```

Or something similar.

librz/core/cmd/cmd_print.c
wargio: I'm sure this is old code

librz/include/rz_search.h
Rot127: Ah I see. What about `hit_desc` then? Meta data seems kinda broad IMHO.

wargio: sure for me.

librz/core/cmd/cmd_print.c
wargio: this is a copy and paste of the functions. i didn't even check what they do.

librz/core/cmd/cmd_print.c
```diff
@@ -4371,6 +4371,36 @@ RZ_IPI RzCmdStatus rz_print_pattern8_handler(RzCore *core, int argc, const char
 	return RZ_CMD_STATUS_OK;
 }
 
+static void incAlphaBuffer(ut8 *buf, int bufsz) {
```
XVilka, line 4374: These are very weird functions

librz/core/cmd/cmd_search.c
```diff
@@ -10,203 +10,12 @@
...
-
-static int preludecnt = 0;
-static int searchflags = 0;
-static int searchshow = 0;
-static const char *searchprefix = NULL;
-
-struct search_parameters {
-	RzCore *core;
+typedef struct search_parameters {
+	RzCore *core; ///< RzCore instance to use
```
XVilka: I think the search parameters should not store `RzCore` and `RzCmdStateOutput`, or maybe rename the structure in that case.

librz/core/cmd/cmd_search.c
```diff
@@ -255,77 +64,52 @@ RZ_IPI RzCmdStatus rz_cmd_detail_gadget_handler(RzCore *core, int argc, const ch
...
-		rz_buf_free(ref);
-		from++;
-	}
-	rz_buf_free(b);
-	rz_cons_break_pop();
-}
+typedef struct search_prelude_cb_ctx {
+	int counter; ///< Number of preludes found
+	RzCore *core; ///< RzCore to use
+	int depth; ///< analysis.depth value
```
XVilka: Should it be `size_t` or `ut32` for both `depth` and `counter`?

librz/core/cmd_descs/cmd_search.yaml
```diff
@@ -3,8 +3,174 @@
...
+    type: RZ_CMD_DESC_TYPE_ARGV_STATE
+    default_mode: RZ_OUTPUT_MODE_STANDARD
+    modes:
+      - RZ_OUTPUT_MODE_STANDARD
+      - RZ_OUTPUT_MODE_JSON
+      - RZ_OUTPUT_MODE_QUIET
+      - RZ_OUTPUT_MODE_TABLE
+    args: []
+
+  - name: "/p"
```
XVilka: For every search subcommand we need more detailed description and few examples.

librz/main/rz-find.c
```diff
@@ -15,6 +15,18 @@
 #include <rz_io.h>
 #include <rz_bin.h>
 
+typedef enum {
+	SEARCH_MODE_BYTES = 0,
```
XVilka: I think these should be in `librz/include/rz_search.h` and also documented.

librz/search/aes_search.c
```diff
@@ -0,0 +1,91 @@
+// SPDX-FileCopyrightText: 2024 RizinOrg <info@rizin.re>
+// SPDX-FileCopyrightText: 2024 deroad <wargio@libero.it>
+// SPDX-FileCopyrightText: 2007 Victor Muñoz
+// SPDX-License-Identifier: CC0-1.0
+
+/*
```
XVilka, line 6: Convert into `\file` Doxygen

librz/search/collection.c
```diff
@@ -0,0 +1,102 @@
...
+ * \brief      Initialize a new RzSearchCollection
+ *
+ * \param[in]  find      The find callback to set
+ * \param[in]  is_empty  The callback to use to check if collection is empty
+ * \param[in]  free      The callback to use to free the context
+ * \param      user      The additional context needed.
+ *
+ * \return     On success returns a valid pointer, otherwise NULL.
+ */
+RZ_IPI RZ_OWN RzSearchCollection *rz_search_collection_new(RZ_NONNULL RzSearchFindCallback find, RZ_NONNULL RzSearchIsEmptyCallback is_empty, RZ_NULLABLE RzSearchFreeCallback free, RZ_NULLABLE void *user) {
```
XVilka, line 18: Interesting concept. Probably could be generalized and put into RzUtil. But that could be done after this PR

librz/search/magic_search.c
```diff
@@ -0,0 +1,55 @@
...
+}
+
+/**
+ * \brief      Allocates and initialize a magic RzSearchCollection
+ *
+ * \param[in]  magic_dir  The magic directory to use for loading the signatures
+ *
+ * \return     On success returns a valid pointer, otherwise NULL
+ */
+RZ_API RZ_OWN RzSearchCollection *rz_search_collection_magic(RZ_NONNULL const char *magic_dir) {
```
XVilka, line 41: If it's an API and RZ_NONNULL please add also assert, here and in other places.

librz/search/pkey_search.c
```diff
@@ -0,0 +1,151 @@
+// SPDX-FileCopyrightText: 2024 RizinOrg <info@rizin.re>
+// SPDX-FileCopyrightText: 2024 deroad <wargio@libero.it>
+// SPDX-FileCopyrightText: 2008 Nadia Heninger and J. Alex Halderman
+// SPDX-FileCopyrightText: 2008 J. Alex Halderman
+// SPDX-License-Identifier: BSD-3-Clause
+
+// Originally from RSAKeyFinder 1.0 (2008-07-18) By Nadia Heninger and J. Alex Halderman
```
XVilka, line 7: Same, please convert into `\file`

librz/search/pkey_search.c
```diff
@@ -0,0 +1,151 @@
...
+	{ 0x02, 0x01, 0x01, 0x04 },
+};
+
+static PKeyMarker safecurves_version = {
+	"safecurves",
+	/// RFC 8410 for Edwards-curve ASN1-bytes representing version field
+	{ 0x02, 0x01, 0x00, 0x30 },
+};
+
+/*Baby BER parser, just good enough for private keys.
```
XVilka, line 44: Since there is comment already for this function , maybe convert to Doxygen?

librz/search/pkey_search.c
```diff
@@ -0,0 +1,151 @@
...
+	}
+
+	const size_t lensize = start[1] & 0x7f;
+	for (size_t i = 0; i < lensize; i++) {
+		*len = (*len << 8) | start[2 + i];
+	}
+	return start + 2 + lensize;
+}
+
+/* Check if `start` points to an ensemble of BER fields
```
XVilka, line 67: Same

librz/search/pkey_search.c
```diff
@@ -0,0 +1,151 @@
...
+	parse_next_field(ptr, &field_len);
+
+	if (!field_len || field_len > KEY_MAX_LEN) {
+		return false;
+	}
+
+	return true;
+}
+
+// Finds and return index of a private key based on the asn1 bytes
```
XVilka, line 93: Same

librz/search/string_search.c
```diff
@@ -0,0 +1,149 @@
+// SPDX-FileCopyrightText: 2024 RizinOrg <info@rizin.re>
+// SPDX-FileCopyrightText: 2024 deroad <wargio@libero.it>
+// SPDX-License-Identifier: LGPL-3.0-only
+
+#include <rz_search.h>
+#include <rz_util.h>
+
```
XVilka, line 7: Please add `\file` to note that it searches not only ANSI/ASCII but also Unicode and some other encodings

XVilka (changes requested):
This is great work!

librz/core/cmd/cmd_print.c
wargio: these were moved from the `cmd_search.c` since they are used by some commands here, i just moved to this file.

librz/core/cmd/cmd_search.c
wargio: this is just a structure that will be removed. this was used when doing the search for each hit. now we will still have a function with a `void *` pointer which can be set to whatever, but we will use this only to cancel the search (since it will be handled within rz_search itself).

librz/core/cmd_descs/cmd_search.yaml
wargio: i think `/p` will be removed since it was linked to a function which nobody knew what was doing nor the original help explained.

librz/main/rz-find.c
wargio: No, it won't. this will be used only to select what to do when calling `rz-find [option]`. it's the same for `rz-diff`

librz/search/collection.c
wargio: yes, it could be.

librz/search/pkey_search.c
wargio: this comment is actually wrong. the code is looking for DER encoding, but i will fix this once it will be "ready"

PR #4745, opened 2024-11-26
Port `v` and `V` to RzShell

Rot127:

**Detailed description**

This brings some changes:

- Visual commands passed to `V`, must be given as normal argument (space between command and string).
- Changes "load layout" `v [layout-name]` to `vl [layout-name]`, to have a hard distinction between `v` and `vl`.
- Change "store layout" `v= [layout-name]` tp `vs [layout-name]`. Because it is more descriptive and matches `vl` a little better.
- Drops the vi command for two reasons:
  - vi (open file in editor) is not at all related to the visual mode. Hence shouldn't be in the same group.
  - (Opinionated) Users can just open the file in another terminal window. Why go the extra way over Rizin?

**Test plan**

All green

**Closing issues**

...


wargio (approved):

librz/core/cmd_descs/cmd_interactive.yaml
```diff
@@ -0,0 +1,58 @@
...
+          - text: ""
+            arg_str: "Press: ?"
+            comment: Opens the help.
+          - text: ""
+            arg_str: "Press: p"
+            comment: Switches between different printing modes (disassembly, hex view, etc.).
+          - text: ""
+            arg_str: "Press: <TAB>"
+            comment: Toggles different configurations of the current printing mode.
+      - name: Examples
```
XVilka: Add more information what the argument means - a sequence of the pressed keys in the visual mode. Also, it's better to have a few examples for those cases.

XVilka (changes requested):
In my practice, commands like `Vp`, `Vpp`, and `VP` are common, maybe add some aliases for those particular cases?

librz/core/cmd_descs/cmd_interactive.yaml
XVilka: Also, this is not full help. Either implement a full help or make `V?` corresponding to the one press of `?` in the `V` mode, and `V??` corresponding to two presses of `?` in the `V` mode.

XVilka (changes requested):

librz/core/cmd_descs/cmd_interactive.yaml
Rot127: Then I go with `VH`/`VHH` for showing the help of the `V` mode. `V?` should be really just for the command.
But I'd like to have more details available to the user before they enter the V mode.
I remember I got confused my first time. It was the "How do I exit vim moment".

XVilka: Ok to have `VH` for the full help, but still examples section and explanation what the argument exactly is needed in the `V?` and `V??` output.

XVilka (approved):

wargio (approved):
i'm ok with these.

XVilka:
@Rot127 please rebase one more time to resolve the conflict

Issue #4231, opened 2024-02-15
Extend graph implementations


Rot127:
**Is your feature request related to a problem? Please describe.**

Currently we have only a single graph implementation which is list based.
The main problem, besides high lookup times, is that we cannot (easily) define our own indices for nodes.
Instead we have to trace them in a hash map on our own.

This is annoying, because:
- Many graphs nodes we have in Rizin, could be easily identified by an address.
- Lookup times could be reduced, if we would use an adjacent matrix implementationaa with a hash map.
- ~~`rz_graph_add_node` doesn't check, if the node was already added (because it needs to compare the user supplied data, since the index says nothing about the node). Really annoying.~~ This is "fixed" (very inefficiently) in https://github.com/rizinorg/rizin/pull/4212 

**Describe the solution you'd like**

1. Extending the list base graph, so it supports arbitrary node identifiers (and checks for edge and node duplicates).
2. Add a adjacency matrix based graph.
3. Add the ability to pre-allocate memory for X nodes/edges (see: https://github.com/rizinorg/rizin/issues/4231#issuecomment-3368401723)
4. The graph should not just associate data with nodes, but also with edges. So different kind of edges can be added to the graph. Or weights can be assigned.

**Describe alternatives you've considered**

Point 1. from above.

**Additional context**

We probably should resolve https://github.com/rizinorg/rizin/issues/397 before.

Also relevant is https://github.com/rizinorg/rizin/issues/992

See files:
- `librz/util/graph.c`
- `librz/util/graph_drawable.c`
- `librz/core/agraph.c`

Rot127:
By any means necessary. The new graph should hash its nodes and edges. And document how exactly it identifies the nodes.

Currently, it compares only pointers of nodes to determine if a node is in the graph or not.
So it is not possible to make a copy of a node and check if it is in the graph. Pointers will mismatch, although the content is the same, and it says no. Which is of course nonsense.

Heersin:
checking code. may I ask the meaning of " it supports arbitrary node identifiers"  @Rot127 ?

Rot127:
Check `RzGraphNode`. It has an index `idx` (I think into the node list) and the `data`. The `idx` is the only unique identifier for the node. It can't have a name or something.

E.g. if you have a graph with unique names, we cannot give a node a string identifier. It is always identified by the index into the node list.

It would be good to allow for different identifiers like strings or any hashable object.

Rot127:
[...]

PR #4753, opened 2024-11-29
Port debug commands to RzShell

Rot127:

**Detailed description**

- Remove `dmiv` command, since it is an alias for `dmi`
- Removes `dta` command:
  Originally it limited the addresses to trace.
  But it was not used in any test, and the implementation was not optimal.
  It allowed only specific addresses (no range).
  Also worked on strings (not ut64 values).
  Also, the usefulness is limited, considering the alternative
  to just not tracing a certain address.
- Removes undocumented `dmias` command, because no idea what it does.

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/librz/core/cdebug.c
+++ b/librz/core/cdebug.c
@@ -103,15 +103,17 @@ RZ_IPI void rz_core_debug_continue(RzCore *core) {
  * \param addr The address to execute to.
  *
  * \return true On success.
- * \return false On otherwise.
+ * \return false Otherwise.
  */
 RZ_API bool rz_core_debug_continue_until(RzCore *core, ut64 addr) {
+#if RZ_BUILD_DEBUG
+	long level = 0;
+	ut64 prev_pc = UT64_MAX;
+#endif
 	ut64 pc;
 	if (!strcmp(core->dbg->btalgo, "trace") && core->dbg->arch && !strcmp(core->dbg->arch, "x86") && core->dbg->bits == 4) {
 		unsigned long steps = 0;
-		long level = 0;
 		const char *pc_name = core->dbg->reg->name[RZ_REG_NAME_PC];
-		ut64 prev_pc = UT64_MAX;
 		bool prev_call = false;
 		bool prev_ret = false;
 		const char *sp_name = core->dbg->reg->name[RZ_REG_NAME_SP];
@@ -137,8 +139,7 @@ RZ_API bool rz_core_debug_continue_until(RzCore *core, ut64 addr) {
 				frame->bp = old_sp;
 				rz_list_prepend(core->dbg->call_frames, frame);
 				RZ_LOG_DEBUG("%ld Call from 0x%08" PFMT64x " to 0x%08" PFMT64x " ret 0x%08" PFMT32x "\n",
-					level, prev_pc, pc, ret_addr);
-				level++;
+					level++, prev_pc, pc, ret_addr);
 				old_sp = cur_sp;
 				prev_call = false;
 			} else if (prev_ret) {
@@ -147,8 +148,7 @@ RZ_API bool rz_core_debug_continue_until(RzCore *core, ut64 addr) {
 					RZ_LOG_DEBUG("*");
 				} else {
 					rz_list_pop_head(core->dbg->call_frames);
-					RZ_LOG_DEBUG("%ld", level);
-					level--;
+					RZ_LOG_DEBUG("%ld", level--);
 				}
 				RZ_LOG_DEBUG(" Ret from 0x%08" PFMT64x " to 0x%08" PFMT64x "\n",
 					prev_pc, pc);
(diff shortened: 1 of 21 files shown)


librz/core/cdebug.c
```diff
@@ -96,13 +96,25 @@ RZ_IPI void rz_core_debug_continue(RzCore *core) {
...
 
-RZ_API bool rz_core_debug_continue_until(RzCore *core, ut64 addr, ut64 to) {
+/**
+ * \brief Continue the execution of the debugged binary until \p addr.
+ *
+ * \param core The current core.
+ * \param addr The address to execute to.
+ *
+ * \return true On success.
+ * \return false On otherwise.
```
XVilka: Just "otherwise".

librz/include/rz_cmd.h
```diff
@@ -91,6 +91,7 @@ typedef enum {
 	RZ_OUTPUT_MODE_LONG_JSON = 1 << 6,
 	RZ_OUTPUT_MODE_TABLE = 1 << 7,
 	RZ_OUTPUT_MODE_QUIETEST = 1 << 8,
+	RZ_OUTPUT_MODE_GRAPH = 1 << 9,
```
XVilka, line 94: Nice addition but there could be multiple types of graph - e.g. native, dot one, interactive, and so on. While it's less relevant in this PR, we should think this over nevertheless, possibly at some later stage. But preferably before the release.

XVilka (changes requested):

librz/include/rz_cmd.h
Rot127: Yeah, would be nice. I can attempt to add it here. If it is too much struggle we can still do it later.

Rot127: Ok, wouldn't do it here. Because it requires some refactor. Not necessarily. We can just pass core in this case, but we shouldn't complicate our code more, if we need to modularize the graph implementation nonetheless.

See: https://github.com/rizinorg/rizin/issues/4231#issuecomment-2512159813

XVilka: Sounds good to me.

XVilka (approved):
@wargio take a look please

XVilka:
@Rot127 could you please resend this PR from inside rizinorg with a branch starting from `dist-` to trigger all CI jobs?

librz/core/cdebug.c
```diff
@@ -127,35 +138,37 @@ RZ_API bool rz_core_debug_continue_until(RzCore *core, ut64 addr, ut64 to) {
...
 				}
-				eprintf(" Ret from 0x%08" PFMT64x " to 0x%08" PFMT64x "\n",
+				RZ_LOG_DEBUG(" Ret from 0x%08" PFMT64x " to 0x%08" PFMT64x "\n",
 					prev_pc, pc);
 				prev_ret = false;
 			}
 			if (steps % 500 == 0 || pc == addr) {
-				eprintf("At 0x%08" PFMT64x " after %lu steps\n", pc, steps);
+				RZ_LOG_DEBUG("At 0x%08" PFMT64x " after %lu steps\n", pc, steps);
 			}
```
wargio, line 160: optimize this since RZ_LOG_DEBUG is not compiled and this branch becomes dead code, by adding `#if RZ_BUILD_DEBUG`

wargio: you can define a new logger within cdebug.c and have this clause which is used to print when x times has been printed.

librz/core/linux_heap_glibc.c
```diff
@@ -1307,30 +1307,25 @@ static void GH(print_heap_bin)(RzCore *core, GHT m_arena, MallocState *main_aren
 		offset = 12 * SZ + sizeof(int) * 2;
 	}
 
-	switch (input[0]) {
-	case '\0': // dmhb
+	if (!input[0]) {
 		PRINT_YA("Bins {\n");
-		for (i = 0; i < NBINS - 1; i++) {
+		for (int i = 0; i < NBINS - 1; i++) {
```
wargio: i hope `NBINS` is a macro, otherwise is a terrible idea to have `signed int` here.

wargio (approved):
LGTM, beside that small comment

librz/core/cdebug.c
Rot127: Doesn't the compiler remove it in this case?

librz/core/linux_heap_glibc.c
Rot127: It is `NBINS = 128`. Changed it to `size_t` nonetheless.

librz/core/cdebug.c
wargio: depends on the optimizations, i would still add the if/endif

Rot127: Done

wargio (approved):

wargio:
```
../librz/core/cdebug.c:115:17: error: variable 'steps' set but not used [-Werror,-Wunused-but-set-variable]
                unsigned long steps = 0;
```

PR #4759, opened 2024-12-07
Revert "rz-test: Set default test dir to '.' (#4741)"

kazarmy:
**Your checklist for this pull request**
- [X] I've read the [guidelines for contributing](https://github.com/rizinorg/rizin/blob/master/DEVELOPERS.md) to this repository
- [X] I made sure to follow the project's [coding style](https://github.com/rizinorg/rizin/blob/master/DEVELOPERS.md#code-style)
- [ ] I've documented or updated the documentation of every function and struct this PR changes. If not so I've explained why.
- [ ] I've added tests that prove my fix is effective or that my feature works (if possible)
- [ ] I've updated the [rizin book](https://github.com/rizinorg/book) with the relevant information (if needed)

**Detailed description**

This pr reverts #4741 since apparently it has code that allows tests in the official Rizin test hierarchy to find their test binaries (in `test/bins`) no matter what the current working directory is.

**Test plan**

Given the `test/db/cmd/cmd_pd` testfile pathname, all of the following should work:
1. `rz-test test/db/cmd/cmd_pd` in the root folder
2. `rz-test db/cmd/cmd_pd` in the `test/` folder
3. `rz-test cmd/cmd_pd` in the `test/db` folder
4. `rz-test cmd_pd` in the `test/db/cmd` folder

One cannot prepend `./` to the `rz-test` arguments above though (except for number 2 for some reason -- `rz-test ./db/cmd/cmd_pd` in the `test/` folder works), but that's an issue for another pr.

Also, all builds should be green.

**Closing issues**

...

PR #4766, opened 2024-12-14
Remove rz_test_chdir()

kazarmy:

**Detailed description**

I've finally realized that the likely reason for the `binrz/rz-test/` requirement in `rz_test_chdir()` is that it assumes an in-source (or in-tree) build, where binaries are placed next to their source when building. [Meson only allows out-of-source (or out-of-tree) builds](https://mesonbuild.com/[Overview](https://mesonbuild.com/Overview.html#terminology).html#terminology), and additionally we have a `meson install` step, so the final location of Rizin binaries can't really be used to find the Rizin test dirs.

Thus, this pr removes the not-relevant-to-us `rz_test_chdir()` code, with the extra benefit that adding `./` to a test filename doesn't result in surprising behavior (as noted in #4759). I've added extra details to the `rz_test_chdir_fromtest()` Doxygen comment since the code can fall back to the current working directory if a test filename isn't provided i.e. `rz-test` is invoked without an argument.

**Test plan**

The above makes sense and there are no significant regressions like what is described in #4759. All builds are green.

**Closing issues**

...

### DIFF
--- a/binrz/rz-test/rz-test.c
+++ b/binrz/rz-test/rz-test.c
@@ -78,7 +78,7 @@ static int help(bool verbose) {
 			"-r",           "[rizin]",        "Path to rizin executable (default is " RIZIN_CMD_DEFAULT ")",
 			"-m",           "[rz-asm]",       "Path to rz-asm executable (default is " RZ_ASM_CMD_DEFAULT ")",
 			"-f",           "[file]",         "File to use for JSON tests (default is " JSON_TEST_FILE_DEFAULT ")",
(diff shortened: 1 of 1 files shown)


kazarmy:
Or maybe this should be kept? One can possibly link to the binaries in the Meson build dir and so this could be useful. Perhaps I should just fix rz_test_chdir() and update the Doxygen comment.

kazarmy:
Perhaps best to simplify matters for now and add search from binary location later

kazarmy:
... _maybe_ add search from binary location later

wargio (approved):

Issue #129, opened 2020-12-08
Add a ruler to horizontal histogram output (e.g. `p==`)


XVilka:
Show also the value ranges on the right, like here:
![image](https://user-images.githubusercontent.com/203261/62100844-ce3e4680-b2c5-11e9-9e6e-fa324786c6fc.png)

Note, if `scr.utf8=true` is enabled, the ruler line should use vertical Unicode line character to be continuous. Also there is no need to name the axes. And no need for horizontal ruler, just vertical one.

ret2libc:
I'd move this to idea repo. @XVilka ok?

DimitrisVita:
Hello. I've added y-axis numbering, which is computed as 255 - threshold, where the threshold is calculated for each row. Please let me know if this approach is acceptable or if any adjustments are needed.

XVilka:
> Hello. I've added y-axis numbering, which is computed as 255 - threshold, where the threshold is calculated for each row. Please let me know if this approach is acceptable or if any adjustments are needed.

@DimitrisVita please send a PR. It's hard to judge without the code.

Maijin:
This is done now and tested

PR #4776, opened 2024-12-17
Add ruler to histogram output

DimitrisVita:

**Detailed description**

This pull request adds a vertical ruler to the horizontal histogram output in `librz/cons/histogram.c`. The ruler includes value ranges and uses a vertical Unicode line character if `scr.utf8=true`.

I've added y-axis numbering, which is computed as 255 - threshold, where the threshold is calculated for each row. Please let me know if this approach is acceptable or if any adjustments are needed. Should it be added to the interactive histogram too?

Here are two images showing the expected output:
* Image 1: `p== 50` with `scr.utf8=true`

peqeq50_true

* Image 2: `p== 50` with `scr.utf8=false`

peqeq50_false

**Test plan**

1. Enable `scr.utf8=true` in the configuration.
2. Generate a horizontal histogram using the `p==` command.
3. Verify that the vertical ruler is displayed correctly with value ranges.
4. Disable `scr.utf8=true` and repeat the test to ensure the ruler is displayed correctly without Unicode characters.

**Closing issues**

Addresses #129 but might need improvement

### DIFF
--- a/librz/cons/histogram.c
+++ b/librz/cons/histogram.c
@@ -41,6 +41,7 @@ RZ_API RZ_OWN RzStrBuf *rz_histogram_horizontal(RZ_NONNULL RzHistogramOptions *o
 	ut32 rows = height > 0 ? height : 10;
 	const char *vline = opts->unicode ? RUNE_LINE_VERT : "|";
 	const char *block = opts->unicode ? UTF_BLOCK : "#";
+	const char *ruler = opts->unicode ? "│" : "|";
 	const char *kol[5];
 	kol[0] = opts->pal->call;
 	kol[1] = opts->pal->jmp;
@@ -63,6 +64,7 @@ RZ_API RZ_OWN RzStrBuf *rz_histogram_horizontal(RZ_NONNULL RzHistogramOptions *o
 					rz_strbuf_append(buf, " ");
 				}
 			}
+			rz_strbuf_appendf(buf, " %s%3zu", ruler, (255 - threshold));
 			rz_strbuf_append(buf, "\n");
 		}
 		return buf;


XVilka:
In such PRs (related to UI) please attach screenshots.

XVilka:
@DimitrisVita I think such ruler should be moved to the left, not the right. Or this can be one more histogram option

librz/cons/histogram.c
```diff
@@ -63,6 +64,7 @@ RZ_API RZ_OWN RzStrBuf *rz_histogram_horizontal(RZ_NONNULL RzHistogramOptions *o
 					rz_strbuf_append(buf, " ");
 				}
 			}
+			rz_strbuf_appendf(buf, " %s%3zu", ruler, (255 - threshold));
```
XVilka, line 67: One more thing - please make ruler optional, controlled by `RzHistogramOptions` structure.

XVilka (changes requested):
I also think, that while vertical ruler should be added to horizontal histogram, horitzontal ruler should be added to vertical histogram.

librz/cons/histogram.c
XVilka: And these options should be created from a new option `scr.hist.ruler`, just like `scr.hist.block` right now.

Rot127: Is it really necessary to make it optional?
It adds extra code, but I cannot think of a situation when it is useful to not have the ruler.

Rot127 (dismissed):

librz/cons/histogram.c
XVilka: Yes, ruler would be unnecessary when the space on the screen is not enough. But it should be enabled by default.

wargio: i agree. it should be optional

DimitrisVita: Thank you guys for your feedback! I will fix it as soon as possible.

XVilka:
@DimitrisVita hi, have you had some time to continue working on this?

notxvilka:
@DimitrisVita if you have any questions, don't hesitate to ask them on the Rizin Dev channel on our Mattermost.

notxvilka:
Closing this because no activity and reaction

Issue #992, opened 2021-04-11
An API to get the function BB graph, callgraph, etc


XVilka:
Currently the commands in `librz/core/cmd_analysis.c` and called functions in `librz/core/canalysis.c` draw the graph directly.
It would be very useful to introduce the API to return such graphs only once in `RzGraph` data structure, then use such an API in these commands as well.

Also split `librz/core/agraph.c` into two parts - purely graph things (probably move some of these into `RzUtil` (`librz/util/`)), printing parts, and TUI parts (these should go into `librz/core/tui/`).

Also add unit/integration C tests

ret2libc:
See as well https://github.com/rizinorg/rizin/pull/637#issuecomment-779973443 . RzGraphDrawable was one first initial attempt by @karliss to improve the situation. I agree we should have some better way to deal with all those graph-drawing commands... How to do that exactly is another story :)

ret2libc:
@XVilka can we remove this from 0.3.0? With you being loaded with types issues and me on commands, I think we don't have time in the next weeks/month to work on this. WDYT?

XVilka:
Yes, sure, it's a long term issue.

PR #4788, opened 2024-12-22
Refactor rizin graph in rz_util

Heersin:

**Detailed description**

working on implement functions
Add more graph representation for rizin, see https://github.com/rizinorg/rizin/issues/4231

1. support identifier for nodes
2. matrix based graph

**Test plan**

1. unit test
2. replace old graph api and test

**Closing issues**

...


librz/include/rz_util/rz_graph_new.h
```diff
@@ -0,0 +1,82 @@
+#ifndef BUILD_RZ_GRAPH_NEW_H
+#define BUILD_RZ_GRAPH_NEW_H
+
+#include <rz_list.h>
+
+/**
+ * @brief Represents a node in a graph, node should be hashable.
```
XVilka: Rizin uses `\brief` syntax, here and everywhere please use this one.

XVilka (dismissed):

librz/include/rz_util/rz_graph_new.h
```diff
@@ -0,0 +1,82 @@
...
+**/
+
+// impl-related functions
+typedef struct rz_graph_edge_ops_t {
+	bool (*add_edge)(RzGraphNew *graph, const RzGraphNodeNew *from, const RzGraphNodeNew *to);
+	bool (*remove_edge)(RzGraphNew *graph, const RzGraphNodeNew *from, const RzGraphNodeNew *to);
+	RzList *(*get_in_neighbours)(RzGraphNew *graph, const RzGraphNodeNew *node);
+	RzList *(*get_out_neighbours)(RzGraphNew *graph, const RzGraphNodeNew *node);
+	bool (*is_directed_connect)(RzGraphNew *graph, const RzGraphNodeNew *from, const RzGraphNodeNew *to);
+	bool (*del_node)(RzGraphNew *graph, RzGraphNodeNew *node);
```
Rot127: I wonder if it is useful to have the same callbacks, but taking the node ids, instead of the node pointers.
This way users can decide they don't want to mess with the pointers at all.

librz/include/rz_util/rz_graph_new.h
```diff
@@ -0,0 +1,82 @@
...
+} RzGraphEdgeOps;
+
+typedef struct rz_graph_edge_set_t_new {
+	RzGraphEdgeOps *ops;
+	void *impl; // different implementations for different edge
+} RzGraphEdgeSet;
+
+struct rz_graph_t_new {
+	unsigned int n_nodes;
+	unsigned int n_edges;
```
Rot127:
```suggestion
	ut64 n_nodes;
	ut64 n_edges;
```

Just in case :)

librz/include/rz_util/rz_graph_new.h
```diff
@@ -0,0 +1,82 @@
...
+
+	HtUP *nodes; /* <hash_id, RzGraphNode> */
```
Rot127: I think you can add a `typedef RzGraphHashId;`

librz/util/graph.c
```diff
@@ -316,3 +316,893 @@ RZ_API void rz_graph_dfs(RzGraph *g, RzGraphVisitor *vis) {
...
+
+#include <rz_util/rz_graph_new.h>
+
+typedef struct rz_graph_list_edge_impl_t {
+	HtUP *adjacent_table; // <ut64, RzList *>
+} RzGraphListEdgeImpl;
+
+typedef struct rz_graph_matrix_edge_impl_t {
+	st64 **adj_matrix;
+	unsigned int n_nodes;
```
Rot127:
```suggestion
	ut64 n_nodes;
```

librz/util/graph.c
```diff
@@ -316,3 +316,893 @@ RZ_API void rz_graph_dfs(RzGraph *g, RzGraphVisitor *vis) {
...
+ * @param graph Listed Implementation of the graph.l
+ * @param from source node.
+ * @param to destination node.
+ * @return true if the edge is added successfully, false otherwise.
+ */
+static bool list_impl_add_edge(RzGraphNew *graph, const RzGraphNodeNew *from, const RzGraphNodeNew *to) {
+	if (!graph || !from || !to) {
+		rz_return_val_if_fail(graph && from && to, false);
+		return false;
+	}
```
Rot127: Why not a simple rz_return_val_if`?

librz/util/graph.c
```diff
@@ -316,3 +316,893 @@ RZ_API void rz_graph_dfs(RzGraph *g, RzGraphVisitor *vis) {
...
+	if (!impl) {
+		RZ_LOG_WARN("No implementation found for the graph")
+		return false;
+	}
+
+	bool from_found = false, to_found = false;
+	ht_up_find(graph->nodes, from->hash_id, &from_found);
+	ht_up_find(graph->nodes, to->hash_id, &to_found);
+	if (!from_found || !to_found) {
+		RZ_LOG_WARN("Node not found in the graph")
```
Rot127:
```suggestion
		RZ_LOG_WARN("Node not found in the graph.\n");
```

The new lines are missing.

librz/util/graph.c
```diff
@@ -316,3 +316,893 @@ RZ_API void rz_graph_dfs(RzGraph *g, RzGraphVisitor *vis) {
...
+		return NULL;
+	}
+	// clone every node in the list
+	RzListIter *it;
+	RzGraphNodeNew *n;
+	rz_list_foreach (neighbours, it, n) {
+		rz_list_append(new_neighbours, n);
+	}
+
+	return new_neighbours;
```
Rot127: I think we can just return the actual list as constant pointer. No need to clone the list internally. These "get_neighbors" functions get called a lot in algorithms and it would cost us a lot of runtime.

Instead, we could add a function to the API like `clone_neighbor_list` or similar. Once we need it. Not here.

librz/util/graph.c
```diff
@@ -316,3 +316,893 @@ RZ_API void rz_graph_dfs(RzGraph *g, RzGraphVisitor *vis) {
...
+		RzList *neighbours = ht_up_find(impl->adjacent_table, hash_id, NULL);
+		if (!neighbours) {
+			continue;
+		}
+		if (rz_list_contains(neighbours, node)) {
+			RzGraphNodeNew *in_node = ht_up_find(graph->nodes, hash_id, NULL);
+			rz_list_append(in_neighbours, in_node);
+		}
+	}
+
```
Rot127: Same as above. In this case I would vote to have two lists for each node. One for `in` and one for `outgoing` edges. If the graph is undirected, just one is used.
Not sure if it is the best solution. A list of edges instead of nodes could be nice as well.

But cloning the list on every `get_neighbor()` doesn't scale well. Exchanging runtime complexity against memory complexity is better IMHO.

librz/util/graph.c
```diff
@@ -316,3 +316,893 @@ RZ_API void rz_graph_dfs(RzGraph *g, RzGraphVisitor *vis) {
...
+	}
+	ht_up_free(list_impl->adjacent_table);
+	free(list_impl);
+	return true;
+}
+
+/**
+ * @brief Initializes the list implementation of the graph.
+ * @param graph The graph to initialize.
+ * @return true if the graph is initialized successfully, false otherwise.
```
Rot127:
```suggestion
 * @return true if the edge is initialized successfully, false otherwise.
```

librz/util/graph.c
```diff
@@ -316,3 +316,893 @@ RZ_API void rz_graph_dfs(RzGraph *g, RzGraphVisitor *vis) {
...
 }
+
+#include <rz_util/rz_graph_new.h>
+
+typedef struct rz_graph_list_edge_impl_t {
+	HtUP *adjacent_table; // <ut64, RzList *>
+} RzGraphListEdgeImpl;
+
+typedef struct rz_graph_matrix_edge_impl_t {
+	st64 **adj_matrix;
```
Rot127:
```suggestion
	ut8 **adj_matrix;
```

Any reason to make the cells so big? With `ut8` we could already encode 255 edge types.

librz/util/graph.c
```diff
@@ -316,3 +316,893 @@ RZ_API void rz_graph_dfs(RzGraph *g, RzGraphVisitor *vis) {
...
+		RZ_LOG_WARN("Node not found in the graph")
+		return false;
+	}
+
+	if (impl->adj_matrix[from->hash_id][to->hash_id] == 1) {
+		RZ_LOG_INFO("An edge already exists between the nodes")
+		return false;
+	}
+
+	impl->adj_matrix[from->hash_id][to->hash_id] = 1;
```
Rot127: Move the magic values like `1` and `0` into macros. `EDGE_SET`, `EDGE_UNSET` or similar.
It might be even useful to add some small `inline` functions to test this property.

E.g.:

```c
static inline bool edge_set(impl, from, to) { return impl->adj_matrix[from->hash_id][to->hash_id] & 1; }
```

This way we can extend the graph easier with special edges. By using the upper bits.

librz/util/graph.c
Rot127: Or `bit0` is set or unset. And the upper bits are some other properties. Edge color for example.

librz/util/graph.c
```diff
@@ -316,3 +316,893 @@ RZ_API void rz_graph_dfs(RzGraph *g, RzGraphVisitor *vis) {
...
+		return NULL;
+	}
+
+	RzList *neighbours = rz_list_new();
+	if (!neighbours) {
+		RZ_LOG_WARN("Failed to allocate memory for the neighbours list")
+		return NULL;
+	}
+
+	for (unsigned int i = 0; i < impl->n_nodes; i++) {
```
Rot127:
```suggestion
	for (ut64 i = 0; i < impl->n_nodes; i++) {
```

librz/util/graph.c
```diff
@@ -316,3 +316,893 @@ RZ_API void rz_graph_dfs(RzGraph *g, RzGraphVisitor *vis) {
...
+		if (impl->adj_matrix[i][node->hash_id] == 1) {
+			RzGraphNodeNew *new_node = ht_up_find(graph->nodes, i, NULL);
+			if (!new_node) {
+				RZ_LOG_WARN("Failed to clone the node")
+				rz_list_free(neighbours);
+				return NULL;
+			}
+			rz_list_append(neighbours, new_node);
+		}
+	}
```
Rot127: Same as above. Let's not clone all the nodes just because the neighbors are requested. The pointers to the neighbors are enough.

It could instead return a `RzIterator` for example.

Rot127: This would largely prevent the graph to be accessed by multiple threads. Because the pointers returned by the iterator might live longer then the nodes in the graph.

But we would need to add another implementation for this use case anyways.

librz/util/graph.c
```diff
@@ -316,3 +316,893 @@ RZ_API void rz_graph_dfs(RzGraph *g, RzGraphVisitor *vis) {
...
+/**
+ * @brief Creates a new graph with the given user hash function and implementation.
+ * Note that any Node should not be free explicitly, it will be freed when the graph is freed.
+ * every Node is RZBORROW, use rz_graph_clone_node to clone the node if needed.
+ * @param user_hash The user hash function to hash the user defined identifier.
+ * @param user_free The user free function to free the user defined data stored in node.
+ * @param impl The implementation of the graph. IMPLEMENTATION_LIST or IMPLEMENTATION_MATRIX etc.
+ * @return The new graph, NULL if the graph is not created successfully.
+ */
+RZ_API RzGraphNew *rz_graph_new_new(RzGraphIdentifierHash user_hash, RzGraphUserdataFree user_free, RzGraphImpl impl) {
```
Rot127:
```suggestion
RZ_API RZ_OWN RzGraphNew *rz_graph_new_new(RzGraphIdentifierHash user_hash, RzGraphUserdataFree user_free, RzGraphImpl impl) {
```

librz/util/graph.c
```diff
@@ -316,3 +316,893 @@ RZ_API void rz_graph_dfs(RzGraph *g, RzGraphVisitor *vis) {
...
+}
+
+/**
+ * get all nodes connected from the given node, nodes -> neighbours
+ * @param g graph
+ * @param node node as center
+ * @return RzList contains all out neighbour nodes (borrowed instance)
+ */
+RZ_API RzList *rz_graph_get_out_nodes_new(RzGraphNew *g, RzGraphNodeNew *node) {
+	return g->edge_set->ops->get_out_neighbours(g, node);
```
Rot127: These callbacks might are a performance problem (see: https://github.com/rizinorg/rizin/issues/4769). We would need to test this and maybe replace them with a switch.

Rot127 (commented):
Looks already way way better then before! Thanks a lot.

Some questions you might have already considered.

1. Should the graph's API mostly work on Node pointers or the node ids? Having both would be nice. If the node content doesn't matter to the user. But working on pointers is fine for me as well for now.
2. Runtime complexity for getting neighbors. Please let's not clone the nodes every time the neighbor list is requested. This doesn't scale well. Instead a read only `RzIterator` over the lists/matrix is a better fit I think.
3. Could the elements in the List based graph implementation contain edges? It would maybe reduce internal complexity getting the neighbors. Especially for directed graphs.

librz/util/graph.c
Heersin: i guess just auto-complete and choose one of them

librz/util/graph.c
Heersin: yeah, it's the heaviest operation for get neighbours, at that time, i didn't know if a better way rizin util could support for User get a reference of Node set

librz/util/graph.c
Heersin: nice, I would fix this dirty and quick implementation

librz/include/rz_util/rz_graph_new.h
Heersin: in the design, these pointers should all be internal to the graph API, with external access only through a unified set of functions.

Heersin:
> Looks already way way better then before! Thanks a lot.
> 
> Some questions you might have already considered.
> 
>     1. Should the graph's API mostly work on Node pointers or the node ids? Having both would be nice. If the node content doesn't matter to the user. But working on pointers is fine for me as well for now.
> 
>     2. Runtime complexity for getting neighbors. Please let's not clone the nodes every time the neighbor list is requested. This doesn't scale well. Instead a read only `RzIterator` over the lists/matrix is a better fit I think.
> 
>     3. Could the elements in the List based graph implementation contain edges? It would maybe reduce internal complexity getting the neighbors. Especially for directed graphs.

1. yes, this is also something I have been contemplating. don't know if it would be confused to provide both way to access. so I defined them mostly work on Node pointer, and they could turn `id` into pointer through hash_find
2. exactly, it did bring crazy workload for runtime. I'm looking for a better way to return neighbors, do u know which file provide a best practice sample for returning `RzIterator`?
3. no they don't contain explicit edges, initially I implemented it as edge and nodes, but later remove edges to abstract the RzGraph and implement PoC. You r right It's good to have edge in impl for list-based graph, and to reduce the complexity, thanks for point out that

Rot127:
>  do u know which file provide a best practice sample for returning RzIterator

You can checkout `ht_inc.c`. The functions `Ht_(as_iter_keys)`, `Ht_(as_iter_mut)` and the corresponding (`iter_next` etc.) return an iterator over a hash map. Your usecase is even simpler I think.

> no they don't contain explicit edges, initially I implemented it as edge and nodes, but later remove edges to abstract the RzGraph and implement PoC. You r right It's good to have edge in impl for list-based graph, and to reduce the complexity, thanks for point out that

You make a good point there. Considering code complexity (meaning: many lines of code) vs. low runtime complexity.

If you don't want to add an edge list for now (because too much work/ too many things to consider), it is fine.
We can still extend it.

cc @wargio

wargio:
agreed with @Rot127

I think it is ok to have first an implementation that works, then we can improve the execution time.

notxvilka:
@Heersin have you had any updates since?

notxvilka:
@Heersin ping

Heersin:
> @Heersin ping

Waiting to implement DFS. A question is I reserved edge_data in RzGraphEdgeNew, not sure if we need such a data for analysis inside rizin @XVilka  @Rot127

Rot127:
@Heersin 
> A question is I reserved edge_data in RzGraphEdgeNew, not sure if we need such a data for analysis inside rizin

Please, yes!
I am currently missing exactly this feature:

My use case is a control flow graph. Each node is a basic block (a sequence of instructions with the last instruction being a branch (jump or call)).

Now, the edges of the CFG are the jumps. Obvious.
But there is no edge between the call and its next instruction (the instruction where the callee would return to). But it would be nice to have some different kind of edge between call and return point (can eases analysis).

Then I would have edges of type I for normal branches, and edges of type II for "follows call".

notxvilka:
i agree, it makes sense to support generic graphs with whatever data provided.

notxvilka (changes requested):
It probably makes sense to split graph into two files: implementation and algorithms.

librz/util/graph.c
```diff
@@ -316,3 +316,1162 @@ RZ_API void rz_graph_dfs(RzGraph *g, RzGraphVisitor *vis) {
...
+    impl->out_edges = ht_up_new(NULL, edge_vec_free_cb);
+    impl->in_edges = ht_up_new(NULL, edge_vec_free_cb);
+    if (!impl->out_edges || !impl->in_edges) {
+        rz_graph_list_impl_fini(impl);
+        return NULL;
+    }
+    return impl;
+}
+
+/* Matrix Impl */
```
notxvilka: I think it makes sense to write "adjacency matrix" instead of just "matrix" so it's clear exactly what does it mean.

notxvilka (changes requested):

librz/util/graph.c
Heersin: ok, will split to two files after unit test

librz/util/graph.c
```diff
@@ -316,3 +316,1162 @@ RZ_API void rz_graph_dfs(RzGraph *g, RzGraphVisitor *vis) {
...
+            goto fail_clean;
+    }
+fail_clean:
+    rz_pvector_free(g->node_vec);
+    ht_up_free(g->nodes);
+    free(g);
+    return NULL;
+}
+
+RZ_API void rz_graph_free_new(RZ_NULLABLE RZ_OWN RzGraphNew *g) {
```
Rot127:
```suggestion
RZ_API void rz_graph_free(RZ_NULLABLE RZ_OWN RzGraphNew *g) {
```

Rot127: Just so we don't forget about it :)

Rot127 (changes requested):

Heersin:
> It probably makes sense to split graph into two files: implementation and algorithms.

DONE
should we replace the old API in another PR? or just replace them in the same one ? @notxvilka @Rot127

Rot127:
@Heersin Please replace it here, so the test coverage is higher.

wargio:
Replace it

notxvilka (changes requested):
Don't forget to rebase it as well, please.

See also this: https://github.com/rizinorg/rizin/issues/992 (could be done separately but gives you an idea of the possible APIs needed)

notxvilka (changes requested):
Many compilation errors

Rot127 (changes requested):
There is a use case I still miss in the tests and API.
It is also very very common, so I think it should be implemented and be an example in the README or similar.

The case is:
How to I check for nodes and edges in a graph, if I don't know the `RzGraphNode` pointers?
I think this is the default scenario for Graph APIs. Because the internal node pointers shouldn't be managed by the API users. Hence, we cannot expect users to know them.

I would rename `rz_graph_has_edge()` to something like `rz_graph_has_edge_by_nptr` or similar.
And change `rz_graph_has_edge(RzGraph *g, RzGraphNode *from, RzGraphNode *to)` to `rz_graph_has_edge(RzGraph *g, void *id_node_a, void *id_node_b)`.

This makes it way more intuitive to use IMO.
Because tbh, I have yet to run into a situation where I identify a node by its node pointer, instead by its identifier.
And the requirement by the old graph to always use the node pointer was a huge annoyance.

Same logic applies to the `has_node` and `del_node`, `del_edge` etc.

I think this should also ease the refactoring in all the other code.

cc @notxvilka Because it is the Graph API topic.

Edit:

To be clear though, I think we should keep the `_by_nptr` variants in the API. It maybe is helpful for some use cases.

notxvilka:
> I would rename `rz_graph_has_edge()` to something like `rz_graph_has_edge_by_nptr` or similar. And change `rz_graph_has_edge(RzGraph *g, RzGraphNode *from, RzGraphNode *to)` to `rz_graph_has_edge(RzGraph *g, void *id_node_a, void *id_node_b)`.
> 
> This makes it way more intuitive to use IMO. Because tbh, I have yet to run into a situation where I identify a node by its node pointer, instead by its identifier. And the requirement by the old graph to always use the node pointer was a huge annoyance.
> 
> Same logic applies to the `has_node` and `del_node`, `del_edge` etc.

Good call, yes, I agree it's a good suggestion.

Heersin:
> There is a use case I still miss in the tests and API. It is also very very common, so I think it should be implemented and be an example in the README or similar.
> 
> The case is: How to I check for nodes and edges in a graph, if I don't know the `RzGraphNode` pointers? I think this is the default scenario for Graph APIs. Because the internal node pointers shouldn't be managed by the API users. Hence, we cannot expect users to know them.
> 
> I would rename `rz_graph_has_edge()` to something like `rz_graph_has_edge_by_nptr` or similar. And change `rz_graph_has_edge(RzGraph *g, RzGraphNode *from, RzGraphNode *to)` to `rz_graph_has_edge(RzGraph *g, void *id_node_a, void *id_node_b)`.
> 
> This makes it way more intuitive to use IMO. Because tbh, I have yet to run into a situation where I identify a node by its node pointer, instead by its identifier. And the requirement by the old graph to always use the node pointer was a huge annoyance.
> 
> Same logic applies to the `has_node` and `del_node`, `del_edge` etc.
> 
> I think this should also ease the refactoring in all the other code.
> 
> cc @notxvilka Because it is the Graph API topic.
> 
> Edit:
> 
> To be clear though, I think we should keep the `_by_nptr` variants in the API. It maybe is helpful for some use cases.

OK, would check them this week. Yes, node pointer is expected to be fetched by calculating hash(identifier), I would add wrapper for this in this phase. @Rot127 is it work in your case ?

Heersin:
> Many compilation errors

Yep, found it's a huge change to replace those graph API (many old graph API relying on accessing inside data). haven't replace them in rz_test/* yet 😭 it keeps reporting error @notxvilka

Rot127:
> add wrapper for this in this phase.

Yes, this would be totally fine. Keep it simple. We optimize later. It is a convenience thing.

> Yep, found it's a huge change to replace those graph API (many old graph API relying on accessing inside data). haven't replace them in rz_test/* yet 😭 it keeps reporting error

I feel you :)
It shows how much needed this refactor was, and how much it will make everything better.

notxvilka:
I also sent a few recommendations on some easy optimizations that can be done in the Mattermost private message. Take a look when you have time.

Heersin:
Still Some DB fixing WIP

librz/util/graph_algorithm.c
```diff
@@ -0,0 +1,289 @@
+// SPDX-FileCopyrightText: 2025-2026 heersin <teablearcher@gmail.com>
+// SPDX-FileCopyrightText: 2007-2020 pancake <pancake@nopcode.org>
+// SPDX-FileCopyrightText: 2007-2020 ret2libc <sirmy15@gmail.com>
+// SPDX-License-Identifier: LGPL-3.0-only
+
+#include <rz_util/rz_graph.h>
+#include <rz_util/rz_iterator.h>
+#include <rz_util/rz_stack.h>
+
+enum {
```
notxvilka: Add a note/doxygen about meaning of these colors, otherwise it might mislead people who read it.

librz/util/graph_algorithm.c
```diff
@@ -0,0 +1,289 @@
...
+/**
+ * DFS edge policy, decide which visitor callback to call based on edge color
+ * \param g graph
+ * \param from
+ * \param to
+ * \param edge_color color array for all nodes, used to determine edge is visited or not, and end point or not
+ * \param visitor callbacks
+ */
+static void dfs_edge_policy(RzGraph *g, RzGraphNode *from, RzGraphNode *to, ut8 *edge_color, RzGraphVisitor *visitor) {
+	// assert g, from, to, edge_color, visitor NON NULL
```
notxvilka, line 55: You could just use `rz_return_if_fail()`

librz/util/graph_algorithm.c
```diff
@@ -0,0 +1,289 @@
...
+
+/**
+ * Helper function to push neighbours of current node into stack for DFS
+ * @param g
+ * @param node
+ * @param stack
+ * @param forward
+ */
+static void dfs_push_neighbours(RzGraph *g, RzGraphNode *node, RzStack *stack, bool forward) {
+	// assert g, node, stack NON NULL
```
notxvilka: Same, `return_if_fail()`

librz/util/graph_algorithm.c
```diff
@@ -0,0 +1,289 @@
...
+			}
+			continue;
+		}
+
+		// normal node (has in and out)
+		if (from && to) {
+			dfs_edge_policy(g, from, to, color, visitor);
+		}
+
+		// CORLOR changes
```
notxvilka, line 154: A typo in COLOR

librz/util/graph_algorithm.c
```diff
@@ -0,0 +1,289 @@
...
+
+/**
+ * Wrapper function for DFS, use visitor pattern to provide callbacks for DFS events
+ * User can implement the visitor callbacks to perform operations
+ * discover_node: called when a node is first discovered (marked GRAY)
+ * finish_node: called when a node is finished (marked BLACK)
+ * tree_edge: called when a tree edge is found (to a WHITE node)
+ * back_edge: called when a back edge is found (to a GRAY node)
+ * fcross_edge: called when a forward or cross edge is found (to a BLACK node)
+ * @param g graph
```
notxvilka: Please use `\param` syntax

librz/util/graph_algorithm.c
```diff
@@ -0,0 +1,289 @@
...
+/**
+ * Wrapper function for DFS, use visitor pattern to provide callbacks for DFS events
+ * Would perform DFS from specified start node, if start is NULL, DFS will be performed for each node without parent
+ * User can implement the visitor callbacks to perform operations
+ * discover_node: called when a node is first discovered (marked GRAY)
+ * finish_node: called when a node is finished (marked BLACK)
+ * tree_edge: called when a tree edge is found (to a WHITE node)
+ * back_edge: called when a back edge is found (to a GRAY node)
+ * fcross_edge: called when a forward or cross edge is found (to a BLACK node)
+ * @param g graph
```
notxvilka: Same

librz/util/graph_algorithm.c
```diff
@@ -0,0 +1,289 @@
...
```
notxvilka: Same

librz/util/graph_impl.c
```diff
@@ -0,0 +1,1650 @@
...
+}
+
+/**
+ * \brief helper func,
+ * Free an edge struct only, user data is freed by graph.
+ * \param e edge to free
+ */
+// user data free by graph not edge_free
+static void edge_free(RzGraphEdge *e) {
+	if (!e) {
```
notxvilka: It could be actually omitted and free() called directly since `free()` handles NULL well. Mark it also as inline then, or as a macro

librz/util/graph_impl.c
```diff
@@ -0,0 +1,1650 @@
...
+	free(e);
+}
+
+/**
+ * \brief helper func,
+ * Create a new RzPVector to hold edges.
+ * \param edge_data_free optional free callback for edge user data
+ * \return A new RzPVector or NULL on failure
+ */
+static RZ_OWN RzPVector *edge_vec_new(RzGraphEdgeDataFree edge_data_free) {
```
notxvilka: Add `inline` here

librz/util/graph_impl.c
```diff
@@ -0,0 +1,1650 @@
...
+ * \brief Add a node to the adjacency list implementation (no-op).
+ *
+ * In the list-based implementation, nodes are managed by the graph itself.
+ * An orphan node simply has no edges in the edge table.
+ *
+ * \param g graph
+ * \param node node to add
+ * \return always true
+ */
+static RZ_OWN bool rz_graph_list_impl_add_node(RzGraph *g, RzGraphNode *node) {
```
notxvilka: Add `inline`

librz/util/graph_impl.c
```diff
@@ -0,0 +1,1650 @@
...
+ * \brief HtUP value free callback to free an edge vector.
+ *
+ * Registered as ht_up free callback, called when a hash table entry
+ * holding an edge RzPVector is removed.
+ *
+ * \param value the RzPVector to free
+ */
+static void edge_vec_free_cb(void *value) {
+	RzPVector *vec = (RzPVector *)value;
+	if (vec) {
```
notxvilka: You can omit the check, `rz_pvector_free()` handles the NULL check well

librz/util/graph_impl.c
```diff
@@ -0,0 +1,1650 @@
...
+/**
+ * \brief Free callback for neighbour iterator state.
+ *
+ * Frees the inner edge iterator and the state struct.
+ *
+ * \param user_data the RzNeighbourIterState to free
+ */
+static void neighbour_iter_free(void *user_data) {
+	RzNeighbourIterState *state = (RzNeighbourIterState *)user_data;
+	if (state) {
```
notxvilka: You could flip this condition and return early instead

notxvilka (changes requested):
A few small nitpicks

notxvilka (changes requested):
```
Missing type comment at <rizin/librz/util/graph_impl.c:1547:59> (token is not a comment)
Missing type comment at <rizin/librz/util/graph_impl.c:1564:46> (token is not a comment)
Missing type comment at <rizin/librz/util/graph_impl.c:1573:46> (token is not a comment)
Missing type comment at <rizin/librz/util/graph_impl.c:1583:46> (token is not a comment)
Missing type comment at <rizin/librz/util/graph_impl.c:1593:46> (token is not a comment)
Missing type comment at <rizin/librz/util/graph_impl.c:1603:77> (token is not a comment)
Missing type comment at <rizin/librz/util/graph_impl.c:1613:61> (token is not a comment)
Missing type comment at <rizin/librz/util/graph_impl.c:1622:60> (token is not a comment)
Missing type comment at <rizin/librz/util/graph_impl.c:1631:65> (token is not a comment)
Missing type comment at <rizin/librz/util/graph_impl.c:1640:64> (token is not a comment)
Missing type comment at <rizin/librz/util/graph_impl.c:1649:54> (token is not a comment)
Missing type comment at <rizin/librz/util/graph_impl.c:1658:53> (token is not a comment)
Missing type comment at <rizin/librz/util/graph_impl.c:1667:87> (token is not a comment)
Missing type comment at <rizin/librz/util/graph_algorithm.c:54:38> (token is not a comment)
Missing type comment at <rizin/librz/util/graph_algorithm.c:89:42> (token is not a comment)
Missing type comment at <rizin/librz/util/graph_algorithm.c:120:37> (token is not a comment)
Missing type comment at <rizin/librz/util/graph_algorithm.c:184:31> (token is not a comment)
Missing type comment at <rizin/librz/util/graph_algorithm.c:239:35> (token is not a comment)
Missing type comment at <rizin/librz/util/graph_algorithm.c:254:43> (token is not a comment)
Missing type comment at <rizin/librz/util/graph_algorithm.c:270:45> (token is not a comment)
Missing type comment at <rizin/librz/util/graph_algorithm.c:286:53> (token is not a comment)
Missing type comment at <rizin/librz/util/graph_algorithm.c:296:43> (token is not a comment)
Missing type comment at <rizin/librz/util/graph_algorithm.c:348:77> (token is not a comment)
Missing type comment at <rizin/librz/util/graph_algorithm.c:462:87> (token is not a comment)
...
```

librz/util/graph_algorithm.c
notxvilka: @Heersin you forgot also this one

notxvilka:
Also compilation on Windows should be fixed:
```
aph_new.c.obj "/c" ../test/unit/test_graph_new.c
../test/unit/test_graph_new.c(40): error C2036: 'void *': unknown size
../test/unit/test_graph_new.c(45): error C2036: 'void *': unknown size
../test/unit/test_graph_new.c(49): error C2036: 'void *': unknown size
../test/unit/test_graph_new.c(53): error C2036: 'void *': unknown size
../test/unit/test_graph_new.c(56): error C2036: 'void *': unknown size
../test/unit/test_graph_new.c(59): error C2036: 'void *': unknown size
../test/unit/test_graph_new.c(70): error C2036: 'void *': unknown size
../test/unit/test_graph_new.c(71): error C2036: 'void *': unknown size
../test/unit/test_graph_new.c(72): error C2036: 'void *': unknown size
../test/unit/test_graph_new.c(105): error C2036: 'void *': unknown size
../test/unit/test_graph_new.c(106): error C2036: 'void *': unknown size
../test/unit/test_graph_new.c(107): error C2036: 'void *': unknown size
../test/unit/test_graph_new.c(133): error C2036: 'void *': unknown size
../test/unit/test_graph_new.c(134): error C2036: 'void *': unknown size
../test/unit/test_graph_new.c(135): error C2036: 'void *': unknown size
../test/unit/test_graph_new.c(136): error C2036: 'void *': unknown size
../test/unit/test_graph_new.c(181): error C2036: 'void *': unknown size
../test/unit/test_graph_new.c(182): error C2036: 'void *': unknown size
../test/unit/test_graph_new.c(183): error C2036: 'void *': unknown size
../test/unit/test_graph_new.c(227): error C2036: 'void *': unknown size
../test/unit/test_graph_new.c(228): error C2036: 'void *': unknown size
../test/unit/test_graph_new.c(229): error C2036: 'void *': unknown size
../test/unit/test_graph_new.c(261): error C2036: 'void *': unknown size
../test/unit/test_graph_new.c(262): error C2036: 'void *': unknown size
../test/unit/test_graph_new.c(263): error C2036: 'void *': unknown size
../test/unit/test_graph_new.c(281): error C2036: 'void *': unknown size
../test/unit/test_graph_new.c(282): error C2036: 'void *': unknown size
../test/unit/test_graph_new.c(283): error C2036: 'void *': unknown size
../test/unit/test_graph_new.c(284): error C2036: 'void *': unknown size
../test/unit/test_graph_new.c(312): error C2036: 'void *': unknown size
../test/unit/test_graph_new.c(313): error C2036: 'void *': unknown size
../test/unit/test_graph_new.c(314): error C2036: 'void *': unknown size
../test/unit/test_graph_new.c(315): error C2036: 'void *': unknown size
../test/unit/test_graph_new.c(316): error C2036: 'void *': unknown size
../test/unit/test_graph_new.c(349): error C2036: 'void *': unknown size
../test/unit/test_graph_new.c(350): error C2036: 'void *': unknown size
../test/unit/test_graph_new.c(351): error C2036: 'void *': unknown size
../test/unit/test_graph_new.c(373): error C2036: 'void *': unknown size
../test/unit/test_graph_new.c(374): error C2036: 'void *': unknown size
../test/unit/test_graph_new.c(375): error C2036: 'void *': unknown size
../test/unit/test_graph_new.c(448): error C2036: 'void *': unknown size
../test/unit/test_graph_new.c(453): error C2036: 'void *': unknown size
../test/unit/test_graph_new.c(457): error C2036: 'void *': unknown size
../test/unit/test_graph_new.c(461): error C2036: 'void *': unknown size
../test/unit/test_graph_new.c(464): error C2036: 'void *': unknown size
../test/unit/test_graph_new.c(467): error C2036: 'void *': unknown size
../test/unit/test_graph_new.c(478): error C2036: 'void *': unknown size
../test/unit/test_graph_new.c(479): error C2036: 'void *': unknown size
../test/unit/test_graph_new.c(480): error C2036: 'void *': unknown size
../test/unit/test_graph_new.c(513): error C2036: 'void *': unknown size
../test/unit/test_graph_new.c(514): error C2036: 'void *': unknown size
../test/unit/test_graph_new.c(515): error C2036: 'void *': unknown size
../test/unit/test_graph_new.c(541): error C2036: 'void *': unknown size
../test/unit/test_graph_new.c(542): error C2036: 'void *': unknown size
../test/unit/test_graph_new.c(543): error C2036: 'void *': unknown size
../test/unit/test_graph_new.c(544): error C2036: 'void *': unknown size
../test/unit/test_graph_new.c(589): error C2036: 'void *': unknown size
../test/unit/test_graph_new.c(590): error C2036: 'void *': unknown size
../test/unit/test_graph_new.c(591): error C2036: 'void *': unknown size
../test/unit/test_graph_new.c(632): error C2036: 'void *': unknown size
../test/unit/test_graph_new.c(633): error C2036: 'void *': unknown size
../test/unit/test_graph_new.c(634): error C2036: 'void *': unknown size
../test/unit/test_graph_new.c(666): error C2036: 'void *': unknown size
../test/unit/test_graph_new.c(667): error C2036: 'void *': unknown size
../test/unit/test_graph_new.c(668): error C2036: 'void *': unknown size
../test/unit/test_graph_new.c(686): error C2036: 'void *': unknown size
../test/unit/test_graph_new.c(687): error C2036: 'void *': unknown size
../test/unit/test_graph_new.c(688): error C2036: 'void *': unknown size
../test/unit/test_graph_new.c(689): error C2036: 'void *': unknown size
../test/unit/test_graph_new.c(716): error C2036: 'void *': unknown size
../test/unit/test_graph_new.c(717): error C2036: 'void *': unknown size
../test/unit/test_graph_new.c(718): error C2036: 'void *': unknown size
../test/unit/test_graph_new.c(719): error C2036: 'void *': unknown size
../test/unit/test_graph_new.c(720): error C2036: 'void *': unknown size
../test/unit/test_graph_new.c(753): error C2036: 'void *': unknown size
../test/unit/test_graph_new.c(754): error C2036: 'void *': unknown size
../test/unit/test_graph_new.c(755): error C2036: 'void *': unknown size
../test/unit/test_graph_new.c(777): error C2036: 'void *': unknown size
../test/unit/test_graph_new.c(778): error C2036: 'void *': unknown size
../test/unit/test_graph_new.c(779): error C2036: 'void *': unknown size
```

Rot127:
@Heersin I continue working on it here: https://github.com/rizinorg/rizin/pull/6152
Can you push to the repo branches as well? Otherwise I can cherry-pick your changes you do here into the `dist-` branch

Heersin:
> @Heersin I continue working on it here: #6152 Can you push to the repo branches as well? Otherwise I can cherry-pick your changes you do here into the `dist-` branch

ok

notxvilka:
Superseded by https://github.com/rizinorg/rizin/pull/6152

PR #4851, opened 2025-01-20
Fuzz missing null check

### LINKED ISSUE #4842: aaaa segmentation fault
### Work environment

| OS/arch/bits (mandatory)                             | Gentoo x86-64, Gentoo x64
| Architecture/bits of the file (mandatory)            | x86-64
| `rizin -v` full output, **not truncated** (mandatory)  | rizin 0.8.0 @ linux-x86-64
commit: afab991fa6f80d6be547682598f28d8388378bed

### Expected behavior

Rizin should not crash during analysis

### Actual behavior

Analysis should not crash

```
rizin bigboy2_nosymb 
 -- Use the '[' and ']' keys in visual mode to adjust the screen width
[0x0000c550]> aaaa
0x0000c550]> e analysis.hasnext=true
[0x0000c550]> aaaa
[x] Analyze function calls
[x] Analyze len bytes of instructions for references
[x] Check for classes
[x] Analyze local variables and arguments
[Segmentation faultnalysis for all functions
```

[tesbin.zip](https://github.com/user-attachments/files/18429622/tesbin.zip)

Rot127:

**Detailed description**

Add missing NULL check.

**Test plan**

Added.

**Closing issues**

closes https://github.com/rizinorg/rizin/issues/4842

### DIFF
--- a/librz/core/analysis_tp.c
+++ b/librz/core/analysis_tp.c
@@ -733,6 +733,9 @@ void propagate_types_among_used_variables(RzCore *core, HtUP *op_cache, RzAnalys
 				RzAnalysisOp *jmp_op = { 0 };
 				ut64 jmp_addr = next_op->jump;
 				RzAnalysisBlock *jmpbb = rz_analysis_fcn_bbget_in(core->analysis, fcn, jmp_addr);
+				if (!jmpbb) {
+					continue;
+				}
 
 				// Check exit status of jmp branch
 				int i;
--- a/test/db/cmd/cmd_af
+++ b/test/db/cmd/cmd_af
@@ -462,3 +462,16 @@ EXPECT=<<EOF
 main + 12
 EOF
 RUN
+
+NAME=Issue 4842
+FILE=bins/fuzzed/issue-4842
+CMDS=<<EOF
+# It only triggers if aaaa is run twice. Very obscure.
+# Can't pinpoint it.
+# But it is related to the `aft` code (type propagation).
+aaaa
+aaaa
+EOF
+REGEXP_FILTER_ERR=Doesn'tMatterJustShouldCatchSegfaults
+EXPECT_ERR=
+RUN


wargio (approved):

wargio:
would be nice to check on coverity if there is a codepath to this (just go to the same line and see if there are CIDs)

kazarmy (commented):
`aa*` commands should be idempotent if nothing has changed (meaning subsequent runs after the first should clear the analysis away and start over so that they follow the exact same code path as the first run), but one shouldn't run the `aa*` commands twice anyway

wargio:
> `aa*` commands should be idempotent if nothing has changed (meaning subsequent runs after the first should clear the analysis away and start over so that they follow the exact same code path as the first run), but one shouldn't run the `aa*` commands twice anyway

we could lock `aa`, `aaa` and `aaaa` and just forbid re-run.

kazarmy:
> we could lock `aa`, `aaa` and `aaaa` and just forbid re-run.

one might want to modify the binary and re-run the analysis

wargio:
> > we could lock `aa`, `aaa` and `aaaa` and just forbid re-run.
> 
> one might want to modify the binary and re-run the analysis

in that case should we drop all the info and rerun from clean state?

kazarmy:
> one might want to modify the binary and re-run the analysis

or modify config and re-run analysis like in #4842

> in that case should we drop all the info and rerun from clean state?

yep ... can probably ignore my previous comment regarding not running the `aa*` commands twice 🤪

Rot127:
Most of this stuff is replaced anyways. I think it is better then to mark certain functions as analyzed. And only analyze them again, if the user made some edits to it.

If we currently reset to a clean state, I wouldn't even know what needs to be reset tbh. Do you know?

kazarmy:
> Most of this stuff is replaced anyways. I think it is better then to mark certain functions as analyzed. And only analyze them again, if the user made some edits to it.

Technically I think this is the best (most efficient) but most complicated option ... might be hard to show that the PR for this works as intended

> If we currently reset to a clean state, I wouldn't even know what needs to be reset tbh. Do you know?

Um no ... probably every analysis step needs to be looked at

notxvilka:
> > Most of this stuff is replaced anyways. I think it is better then to mark certain functions as analyzed. And only analyze them again, if the user made some edits to it.
> 
> Technically I think this is the best (most efficient) but most complicated option ... might be hard to show that the PR for this works as intended
> 
> > If we currently reset to a clean state, I wouldn't even know what needs to be reset tbh. Do you know?
> 
> Um no ... probably every analysis step needs to be looked at

You are right, thus I think the proper fix is better to be done during the reimplementation of the analysis on top of RzArch and RZIL. We just need to make sure to add tests for this.

kazarmy:
> You are right, thus I think the proper fix is better to be done during the reimplementation of the analysis on top of RzArch and RZIL. We just need to make sure to add tests for this.

Ok but before the reimplementation happens, probably best to do what @wargio suggests and forbid re-run of `aa`, `aaa` and `aaaa`

wargio:
single commands (`af` `aac` etc..) are ok, just the automated analysis could be forbidden.

notxvilka:
> single commands (`af` `aac` etc..) are ok, just the automated analysis could be forbidden.

I don't think we should forbid anything. Lets just wait until a proper rewrite/refactor.

notxvilka:
@Rot127 test is failing:
```
 [XX] 184 secs db/cmd/cmd_af Issue 4842
RZ_NOPLUGINS=1 /usr/bin/rizin -escr.utf8=0 -escr.color=0 -escr.interactive=0 -eflirt.sigdb.load.system=false -eflirt.sigdb.load.home=false -N -qc '# It only triggers if aaaa is run twice. Very obscure.
# Can't pinpoint it.
# But it is related to the `aft` code (type propagation).
aaaa
aaaa
' bins/fuzzed/issue-4842
-- stdout
--- expected
+++ actual
@@ -1,1 +1,0 @@
-

-- stderr
--- expected
+++ actual
@@ -1,0 +1,1 @@
+

```
Could you please also rebase?

notxvilka (approved):

PR #4862, opened 2025-01-24
`/e`: Set NUL (`\0`) as newline

kazarmy:

**Detailed description**

This pr sets NUL (`\0`) as the regex newline for `/e` (regex search). This is to prevent the C string NUL terminator from being part of a match, which is not only unexpected but also does not work well with greedy regex operators.

**Test plan**

All builds are green.

**Closing issues**

...

### DIFF
--- a/librz/util/regex.c
+++ b/librz/util/regex.c
@@ -13,7 +13,7 @@
 #include <rz_util.h>
 
 typedef pcre2_general_context RzRegexGeneralContext; ///< General context.
-typedef pcre2_compile_context RzRegexCompContext; ///< The context for compiling.
+// typedef pcre2_compile_context RzRegexCompContext; ///< The context for compiling.
 typedef pcre2_match_context RzRegexMatchContext; ///< The context for matching.
 
 typedef struct {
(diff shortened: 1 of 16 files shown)


notxvilka (approved):
Wait for @Rot127 who rewrites the search code first please

librz/util/regex.c
```diff
@@ -13,7 +13,7 @@
 #include <rz_util.h>
 
 typedef pcre2_general_context RzRegexGeneralContext; ///< General context.
-typedef pcre2_compile_context RzRegexCompContext; ///< The context for compiling.
+// typedef pcre2_compile_context RzRegexCompContext; ///< The context for compiling.
```
wargio, line 16: Why this is removed?

kazarmy: Because of the `typedef` in [`librz/include/rz_util/rz_regex.h`](https://github.com/rizinorg/rizin/pull/4862/files#diff-00facbeb71a3f5571b1525f4ecdecd84138e59e1c9fc30bcaf7e7e2477c397cc):

```c
typedef void RzRegexCompContext; ///< A PCRE2 compile context.
```

wargio (approved):
LGTM @Rot127 Give a look

Rot127 (approved):
Since it is not the default it is fine.

For the regex search it might be unexpected behavior if combined with `RZ_REGEX_MUTILINE`. But I won't extend it with this now. Because the focus is more on the `RzShell` refactor. Not on improving/extending of the search.

kazarmy:
> For the regex search it might be unexpected behavior if combined with `RZ_REGEX_MULTILINE`.

Some testing is probably needed to figure out what `^` and `$` mean here, with and without `RZ_REGEX_MULTILINE`. Thanks for the approval!

PR #4886, opened 2025-02-08
Add bit counting and address alignment calculation helper.

Rot127:

**Detailed description**

Adds bit counting helper and a function to calculate the padding to align an address. 

**Test plan**

Added.

**Closing issues**

...

### DIFF
--- a/librz/include/rz_util/rz_bits.h
+++ b/librz/include/rz_util/rz_bits.h
@@ -12,7 +12,12 @@ extern "C" {
 #include <rz_types_base.h>
 
 /**
- * \brief Reference: https://graphics.stanford.edu/~seander/bithacks.html#CountBitsSetParallel
+ * \brief Count number of 1s in the given value.
+ * Reference: https://graphics.stanford.edu/~seander/bithacks.html#CountBitsSetParallel
+ *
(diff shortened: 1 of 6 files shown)


wargio (commented):
What do you mean with count bits? Count how many ones? Or how many bits the number uses? Like 3 uses 2 bits, and 2 uses 2 bits too?

Rot127:
Yes, how many bits are set in the `ut64`. Will add some doxygen then :D

wargio:
> Yes, how many bits are set in the `ut64`. Will add some doxygen then :D

Be careful on what you write, because "how many bits are set" is counting how many ones.

I think you meant is "how many bits are required to represent a number" or better "the size of a number in bits".

Rot127:
> Be careful on what you write, because "how many bits are set" is counting how many ones.

This is the correct interpretation. It is counting ones in the value. Added it as doxygen as well. I'll rename it to `count_ones`

wargio (approved):
Ok, now it makes a lot of sense.

notxvilka (approved):
Where do you plan using it?

wargio:
It was used to ensure the memory alignment is defined by just one bit, which makes sense

PR #4897, opened 2025-02-13
string search json: Remove `data` property

kazarmy:

**Detailed description**

For fixed-length string search (not regex search), it is sufficient to have only the hit offset / address to determine the boundaries of a match. Thus, this pr proposes removing the `data` property from string search json, putting the onus on the scripter to decide what to do next with the hit offsets. The main advantage of removal is to reduce the space needed for the json string, but this also sidesteps the issues of `data` encoding and length, and cleans up the code a bit. 

**Test plan**

**Closing issues**

...

### DIFF
--- a/test/db/cmd/cmd_json
+++ b/test/db/cmd/cmd_json
@@ -37,8 +37,7 @@ EXPECT=<<EOF
 [
   {
     "offset": 4203952,
-    "type": "string",
-    "data": "cation1\\Release\\ConsoleApplication1.pdb..... ... ......"
+    "type": "string"
   }
 ]
 EOF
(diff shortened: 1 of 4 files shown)


test/db/cmd/cmd_json
```diff
@@ -37,8 +37,7 @@ EXPECT=<<EOF
 [
   {
     "offset": 4203952,
-    "type": "string",
-    "data": "cation1\\Release\\ConsoleApplication1.pdb..... ... ......"
+    "type": "string"
```
wargio, line 40: this looks wrong.

wargio:
hmmm. it does make sense that we should return only where the match happens. let's see what the others says. i'm ok with this

@Rot127 @notxvilka @thestr4ng3r

Rot127:
This is already the case in the refactored search
```

	char *hit_desc; ///< Hit description (can be NULL)
	ut64 address; ///< Address/offset of the matched data.
	size_t size; ///< Size of the matched data (can be 0), in bytes.
} RzSearchHit;

```

notxvilka:
No need for this since will be fixed in a few days via merging the RzSearch PR

PR #4901, opened 2025-02-13
Allow an thread list iterator to cancel the iteration, if it failed.

Rot127:

**Detailed description**

From https://github.com/rizinorg/rizin/pull/4762

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/librz/util/thread_iterators.c
+++ b/librz/util/thread_iterators.c
@@ -52,8 +52,8 @@ static void *thread_iterate_list_cb(th_list_ctx_t *context) {
 		context->head = rz_list_iter_get_next(context->head);
 		rz_th_lock_leave(lock);
 
-		if (element) {
-			iterator(element, user);
+		if (element && !iterator(element, user)) {
+			break;
 		}
 	} while (true);
 	return NULL;
--- a/test/unit/test_threads.c
+++ b/test/unit/test_threads.c
@@ -138,9 +138,10 @@ bool test_thread_ht(void) {
 	mu_end;
 }
 
-void thread_set_bool_arg(bool *value, bool *user) {
+bool thread_set_bool_arg(bool *value, bool *user) {
 	*value = true;
 	*user = true;
+	return true;
 }
 
 bool test_thread_iterator_list(void) {
--- a/librz/include/rz_th.h
+++ b/librz/include/rz_th.h
@@ -34,7 +34,7 @@ typedef struct rz_th_t RzThread;
 typedef struct rz_th_pool_t RzThreadPool;
 typedef struct rz_th_queue_t RzThreadQueue;
 typedef void *(*RzThreadFunction)(void *user);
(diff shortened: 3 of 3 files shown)


librz/util/thread_iterators.c
```diff
@@ -52,8 +52,8 @@ static void *thread_iterate_list_cb(th_list_ctx_t *context) {
 		context->head = rz_list_iter_get_next(context->head);
 		rz_th_lock_leave(lock);
 
-		if (element) {
-			iterator(element, user);
+		if (element && !iterator(element, user)) {
+			break;
```
wargio, line 56: you cannot do this. it will kill only one thread, not all of them.

test/unit/test_threads.c
```diff
@@ -138,9 +138,10 @@ bool test_thread_ht(void) {
 	mu_end;
 }
 
-void thread_set_bool_arg(bool *value, bool *user) {
+bool thread_set_bool_arg(bool *value, bool *user) {
 	*value = true;
 	*user = true;
+	return true;
```
wargio, line 144: Why?

librz/util/thread_iterators.c
wargio: This also breaks compatibility between rz_th_iterate_list and rz_th_iterate_pvector

Rot127: Mh, this was a change copied from your search branch. What was the intention back then?

Rot127: We can revert it, wouldn't be a big problem.

wargio (changes requested):
This is totally wrong.

librz/util/thread_iterators.c
wargio: If i wrote that, i must have been under heavy drugs. We can close this, and i will submit the correct patch.

wargio: image
Just checked. this is the correct patch. i do not know where that comes from

Rot127: Not this specific line. Just the return value. Just added it here because otherwise the change of the return value would have been ignored? Or was it meant as a general way to signal if the iterator finished properly?

notxvilka:
> This is totally wrong.

Total recall.

librz/util/thread_iterators.c
wargio: When you write thread stuff you need to consider that the stack duplicated (thus not shared) for each thread context.
Threads needs syncronization mechanisms when you need to stop all of them, and this is usually done via notification system for this or a shared value.

rz_atomic_XXX does this, via shared value. we have a thread safe shared value (a boolean) which we use to ensure a thread can be stopped by any thread that fails. why we need a thread safe value? because depending on the CPU cache, a thread might see a non updated value.

With the original patch, if you had 10 threads, and you fail in one, the other 9 will keep running and the failed thread will just halt.

PR #4832, opened 2025-01-10
Migrate from Capstone to Zydis (x86 architecture)

tushar3q34:
**Your checklist for this pull request**
- [X] I've read the [guidelines for contributing](https://github.com/rizinorg/rizin/blob/master/DEVELOPERS.md) to this repository
- [X] I made sure to follow the project's [coding style](https://github.com/rizinorg/rizin/blob/master/DEVELOPERS.md#code-style)
- [ ] I've documented or updated the documentation of every function and struct this PR changes. If not so I've explained why.
- [ ] I've added tests that prove my fix is effective or that my feature works (if possible)
- [ ] I've updated the [rizin book](https://github.com/rizinorg/book) with the relevant information (if needed)

**Detailed description**

- [X] Build zydis as a dependency
- [X] Use zydis structs / functions etc for disassembly and analysis
- [X] Use zydis for `asm_x86` file
- [X] Use zydis for `analysis_x86` file and update zydis to x86 ESIL generation
- [X] Migrate to zydis for x86 RzIL
- [X] Change Tests/code to account for differences in outputs presented by Capstone and Zydis
- [ ] Extend big endian host system support in Zydis (in particular, S390/System Z architecture)
- [ ] Document functions of `analysis_x86` and `asm_x86` files

**Test plan**

...

**Closing issues**

closes #4709

...

PR #4918, opened 2025-02-20
Prefix Analysis memmove fix

tushar3q34:

**Detailed description**

As discussed in #4832 , this change prevents ASAN build fails.

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/librz/core/canalysis.c
+++ b/librz/core/canalysis.c
@@ -5208,8 +5208,12 @@ static void *AnalysisBytesContext_next(RzIterator *it) {
 		*sp = 0;
 		if (op->prefix) {
 			char *p = strchr(sp + 1, ' ');
-			*p = 0;
-			memmove(ctx->mnemonic, sp + 1, p - sp);
+			if (!p) {
+				rz_str_ncpy(ctx->mnemonic, sp + 1, strlen(sp + 1) + 1);
+			} else {
+				*p = 0;
+				rz_str_ncpy(ctx->mnemonic, sp + 1, p - sp);
+			}
 		}
 	}
 	op->mnemonic = mnem;


librz/core/canalysis.c
```diff
@@ -5208,8 +5208,12 @@ static void *AnalysisBytesContext_next(RzIterator *it) {
 		*sp = 0;
 		if (op->prefix) {
 			char *p = strchr(sp + 1, ' ');
-			*p = 0;
-			memmove(ctx->mnemonic, sp + 1, p - sp);
+			if (!p) {
+				memmove(ctx->mnemonic, sp + 1, strlen(sp + 1) + 1);
```
Rot127: Is it guaranteed that `sp` is `\0` terminated? Otherwise using `rz_str_ncpy` here is better, because we get the '\0' byte for sure.

Rot127 (changes requested):

librz/core/canalysis.c
tushar3q34: `ctx->mnemonic` is character array filled with `'\0'` initially, so non null terminated strings should not affect anything.
Edit : `sp` is null terminated as we set `*sp = 0` in line 5208. The original string is split into two different strings. After operation `mnem` will have the prefix and `sp+1` will have actual mnemonic(which will be null terminated as the original one was)

Rot127: Mh, idk. Better use `rz_str_ndup` here. You never know how this stuff changes.

Rot127 (approved):

notxvilka:
Not sure if related, but it broke this test on FreeBSD:
```
[XX]  24 ms   db/analysis/x86_32 mnemonic - repe/cmpsb
RZ_NOPLUGINS=1 /home/build/bin/rizin -escr.utf8=0 -escr.color=0 -escr.interactive=0 -eflirt.sigdb.load.system=false -eflirt.sigdb.load.home=false -N -qc 'e asm.arch=x86
e asm.bits=64
a8 f3a6~mnemonic[1]
' =
-- stdout
--- expected
+++ actual
@@ -1,1 +1,0 @@
-cmpsb
```
https://builds.sr.ht/~thestr4ng3r/job/1435225#task-test

tushar3q34:
> Not sure if related, but it broke this test on FreeBSD:

Yes it is related.
So I searched about it a bit, not entirely sure but seems like `strlcpy` or `strncpy` do not handle overlapping memory so that might be causing the issue. Probably that is the reason why `memove` was used in the first place.

We can use something like this to ensure it is null terminated and handles overlapping memory as well:
```
memmove(ctx->mnemonic, sp + 1, strlen(sp + 1));
ctx->mnemonic[strlen(sp + 1)] = '\0';
```

notxvilka:
> > Not sure if related, but it broke this test on FreeBSD:
> 
> Yes it is related. So I searched about it a bit, not entirely sure but seems like `strlcpy` or `strncpy` do not handle overlapping memory so that might be causing the issue. Probably that is the reason why `memove` was used in the first place.
> 
> We can use something like this to ensure it is null terminated and handles overlapping memory as well:
> 
> ```
> memmove(ctx->mnemonic, sp + 1, strlen(sp + 1));
> ctx->mnemonic[strlen(sp + 1)] = '\0';
> ```

Yes, that should work.

PR #4943, opened 2025-02-27
search: Clear progress before results

kazarmy:

**Detailed description**

The search progress notification is of significant length, so if the first result line is short, then there will be an unwanted tail, e.g.

slash-z-first-has-tail

due to use of just `\r`. This pr clears the last progress notification before printing the results so the above will not happen. It also sleeps first before printing search progress, rather than the other way around, to prevent unnecessary flicker.

**Test plan**

All builds are green. Test may be racy, but hopefully CI computers are fast enough.

**Closing issues**

...

### DIFF
--- a/librz/search/search.c
+++ b/librz/search/search.c
@@ -565,12 +565,15 @@ static void *search_cancel_th(void *user) {
 	RzSearchOpt *opt = ctx->opt;
 
 	do {
+		rz_sys_usleep(RZ_SEARCH_CANCEL_CHECK_INTERVAL_USEC);
+		if (!rz_atomic_bool_get(ctx->loop)) {
+			break;
+		}
 		size_t n_hits = rz_th_queue_size(ctx->hits);
 		if (opt->cancel_cb(opt->cancel_usr, n_hits, RZ_SEARCH_CANCEL_REGULAR_CHECK)) {
 			rz_atomic_bool_set(ctx->loop, false);
 			break;
 		}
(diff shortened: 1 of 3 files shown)


Rot127 (approved):

librz/search/search.c
```diff
@@ -565,12 +565,15 @@ static void *search_cancel_th(void *user) {
...
+		if (!rz_atomic_bool_get(ctx->loop)) {
+			break;
+		}
 		size_t n_hits = rz_th_queue_size(ctx->hits);
 		if (opt->cancel_cb(opt->cancel_usr, n_hits, RZ_SEARCH_CANCEL_REGULAR_CHECK)) {
 			rz_atomic_bool_set(ctx->loop, false);
 			break;
 		}
-		rz_sys_usleep(RZ_SEARCH_CANCEL_CHECK_INTERVAL_USEC);
 	} while (rz_atomic_bool_get(ctx->loop));
```
wargio: i don't think these changes are really needed.

kazarmy: no but they make the ux less annoying and theoretically one doesn't need to show progress immediately

kazarmy: no but they make the ux less annoying and theoretically one doesn't need to show progress immediately after getting a task

wargio: then just invert the do-while into a while(loop) { sleep(); ... }

kazarmy: ok b6f320dd8e70a20c0758bc7062298c8b92861c44

wargio (approved):

PR #4950, opened 2025-03-01
search.show_progress: Show search intervals

kazarmy:

**Detailed description**

This pr makes `search.show_progress` show search intervals as can be seen in [this test output](https://github.com/rizinorg/rizin/compare/dev...kazarmy:rizin:search.show_progress-show-search-intervals?expand=1#diff-e85e3180c7387bbc70d11122adca6d61576a3a1ddce417ce1cd3e4a94b72275cR146) except that they will be unsorted. This is so one can confirm whether the search intervals are correct.

**Test plan**

The output makes sense. All builds are green.

**Closing issues**

...


librz/search/bytes_search.c
```diff
@@ -215,6 +216,7 @@ static bool bytes_find(RzSearchFindOpt *fopts, void *user, ut64 address, const R
 	const ut8 *raw_buf = rz_buf_get_whole_hot_paths((RzBuffer *)buffer, &size);
 	void **it = NULL;
 	RzPVector /*<BytesPattern *>*/ *patterns = (RzPVector *)user;
+	*num_hits = 0;
```
wargio, line 219: why? `RzThreadQueue *hits` contains the hits, just invoke `rz_th_queue_size()` for the current size of the queue, i.e. the number of hits.

librz/search/search.c
```diff
@@ -604,10 +604,13 @@ static bool search_iterator_io_map_cb(void *element, void *user) {
...
 	rz_th_lock_leave(ctx->io_lock);
 
+	size_t num_hits = 0;
 	RzSearchFindBytesCallback find = col->find;
-	if (!find(ctx->opt->find_opts, col->user, at, buffer, ctx->hits)) {
+	if (!find(ctx->opt->find_opts, col->user, at, buffer, ctx->hits, &num_hits)) {
 		RZ_LOG_ERROR("search: failed search at 0x%08" PFMT64x "\n", at);
 		goto failure;
+	} else if (ctx->opt->show_progress) {
+		eprintf("Searching in [0x%" PFMT64x ", 0x%" PFMT64x ") ... hits: %zu\n", at, at + size, num_hits);
```
wargio: This is wrong for me, since each thread has a different search range. if you want to do this, you should follow what Basefind does.

librz/search/bytes_search.c
kazarmy: `RzThreadQueue *hits` contains all of the hits that have been found so far in the search, not the number of hits in the current search interval

librz/search/search.c
kazarmy: > This is wrong for me, since each thread has a different search range.

Well if 2 threads have the same search range, they are doing duplicate work.

> if you want to do this, you should follow what Basefind does.

Why?

kazarmy: I looked at https://github.com/LindLL/basefind/tree/main?tab=readme-ov-file#1sony and https://github.com/sgayou/rbasefind?tab=readme-ov-file#example, they have more compact output (I think) but they are definitely not sorted

wargio: no, basefind output is the one that you shown, and we do the same, let me explain better what i meant.

`rz_th_iterate_*` iterates elements over threads, each thread will get **only one** element of the array/list.
In search.c we fill the list with RzInterval, so each thread will get only one interval, so if you want to print the number of hits for each RzInterval, then we can just add a mutex and print the nHits at the end of each `find` callback.

The issue i see here is that search and TUI prints should go in different threads, so when we bring this code to cutter, we wont have prints over terminal when one searches but on the GUI. so if you want to have nhits for each interval, then we should define a new structure and store the metadata there, like

```c
typedef struct rz_search_interval_t {
    RzInterval interval;
    size_t n_hits;
} RzSearchInterval;
```

This is what we do for basefind: https://github.com/rizinorg/rizin/blob/dev/librz/include/rz_basefind.h#L22

wargio: So for printing these results, for me, we have 2 ways:

- Print them live via the `search_cancel_th` thread (we need to define a new RzThreadQueue to inform the `search_cancel_th` callback to print stuff when the queue is not empty).
- Print them at the end of the execution.

kazarmy: > - Print them at the end of the execution.

Deferring printing until the end means not printing during the search (about search progress) so I think only the first way is viable.

wargio: If we do live printing, then we can even have a percentage of progress.

kazarmy:
> ...  print the nHits at the end of each `find` callback.

Now I see I can (probably) get the search interval size from the buffer 🤦

librz/search/search.c
kazarmy: > ...  print the nHits at the end of each `find` callback.

Now I see I can (probably) get the search interval size from the buffer 🤦

librz/search/search.c
```diff
@@ -604,10 +604,13 @@ static bool search_iterator_io_map_cb(void *element, void *user) {
...
 	rz_th_lock_leave(ctx->io_lock);
 
+	size_t num_hits = 0;
 	RzSearchFindBytesCallback find = col->find;
-	if (!find(ctx->opt->find_opts, col->user, at, buffer, ctx->hits)) {
+	if (!find(ctx->opt->find_opts, col->user, at, buffer, ctx->hits, &num_hits)) {
 		RZ_LOG_ERROR("search: failed search at 0x%08" PFMT64x "\n", at);
 		goto failure;
+	} else if (ctx->opt->show_progress) {
+		eprintf("[0x%" PFMT64x ", 0x%" PFMT64x "): %zu\n", at, at + size, num_hits);
```
Rot127:
```suggestion
		eprintf("[0x%" PFMT64x ", 0x%" PFMT64x "): %" PFMTSZu "\n", at, at + size, num_hits);
```

~~Also I think it is an end inclusive range `[x, y]`. Could be wrong though.~~ Can't read. all good as it is.

librz/search/options.c
```diff
@@ -66,6 +66,12 @@ RZ_API bool rz_search_opt_set_max_threads(RZ_NONNULL RzSearchOpt *opt, RzThreadN
 	return true;
 }
 
+RZ_API bool rz_search_opt_set_show_progress(RZ_NONNULL RzSearchOpt *opt, bool show_progress) {
+	rz_return_val_if_fail(opt, false);
+	opt->show_progress = show_progress;
```
Rot127: I think though this should be an extra option. Something like `show_progress_verbose`. Because if the file is big, there will be many intervals, which clutters the output. Maybe it could be also a choice:

`number_of_hits`, `percent`, `hits_per_interval/sub-graph/knowledge domain` etc. Might be also useful for later, when we have searches over graphs or the KB.

Rot127 (changes requested):
> Well if 2 threads have the same search range, they are doing duplicate work.

This should not happen btw. Did you experience it? If so, it is a bug.

librz/search/search.c
Rot127: Agree with the design of the latest commit (https://github.com/rizinorg/rizin/pull/4950/commits/ac9f5ce5631981325437d2469ccb1efc829a75a2).

Passing `*num_hits` to the `find()` function is a little meh, because it clutters it, but can't come up with a simpler idea as well. Please document the parameter though!

wargio:
> This should not happen btw. Did you experience it? If so, it is a bug.

i think it was just a misunderstanding of the API

kazarmy:
> This should not happen btw. Did you experience it? If so, it is a bug.

No I didn't experience it

librz/search/search.c
kazarmy: Documented the parameter in 3b78cb6aa8987a9f3f2e279666929a5aaf2ab0eb. I did think of discarding the parameter and doing final nHits processing at the end of each `find` callback, but I think it's better to reduce code duplication as much as possible.

wargio:
for me you could just return `st64` and say that if negative, we consider this as "an error occurred", so we can avoid passing too many arguments, or just add the `RzSearchInterval` struct and instead of passing the address + size, we pass the whole structure which does the same.

librz/search/search.c
kazarmy: > - Print them live via the `search_cancel_th` thread ...

See 204643c809051ae7042f8a5be10f3ec708c2020c

librz/search/options.c
kazarmy: I think that a hybrid approach with `show_progress` i.e. it accepts `false`, `true` and the special value `interval_hits` (shorter than `hits_per_interval`) is currently the best way to handle this, but it will take some time.

kazarmy: See 97ae8557d1c660f11744f8b81f1558b94fb9479f ... I'm using `intervals` since it's shorter than `interval_hits` yet basically has the same meaning (i think)

librz/core/cconfig.c
```diff
@@ -3747,7 +3747,7 @@ RZ_API int rz_core_config_init(RzCore *core) {
 		NULL);
 	SETPREF("search.prefix", "hit", "Prefix name in search hits label");
 	SETI("search.maxhits", 0, "Maximum number of hits ('0' means no limit)");
-	SETBPREF("search.show_progress", "true", "Show the search process.");
+	SETPREF("search.show_progress", "1", "Show the search process (true, false, intervals)");
```
wargio: make this as a option see `str.encoding` for the example..

kazarmy: ok b9355bdcf7767ae0ddc442b7dea8fa47a87c742b ... the reason `"1"` is used instead of `"true"` as the initial value is because of what appears to be a config design flaw that limits config options to only boolean values if the initial value is either `"true"` or `"false"`

librz/core/cconfig.c
```diff
@@ -2651,6 +2651,21 @@ static bool cb_search_in(void *user, void *data) {
...
 
+static bool cb_search_show_progress(void *user, void *data) {
+	RzConfigNode *node = (RzConfigNode *)data;
+	if (rz_str_is_true(node->value)) {
+		free(node->value);
+		node->value = rz_str_dup("true");
+	} else if (rz_str_is_false(node->value)) {
+		free(node->value);
+		node->value = rz_str_dup("false");
+	} else if (strcmp(node->value, "intervals")) {
```
Rot127:
```suggestion
	} else if (!RZ_STR_EQ(node->value, "intervals")) {
```

Rot127: Also, the the `intervals` case is missing?

librz/search/options.c
```diff
@@ -66,6 +66,18 @@ RZ_API bool rz_search_opt_set_max_threads(RZ_NONNULL RzSearchOpt *opt, RzThreadN
 	return true;
 }
 
+RZ_API bool rz_search_opt_set_show_progress_from_str(RZ_NONNULL RzSearchOpt *opt, const char *show_progress) {
+	rz_return_val_if_fail(opt, false);
+	if (rz_str_is_false(show_progress)) {
+		opt->show_progress = RZ_SEARCH_PROGRESS_DISABLED;
+	} else if (!strcmp(show_progress, "intervals")) {
```
Rot127:
```suggestion
	} else if (RZ_STR_EQ(show_progress, "intervals")) {
```

librz/search/search.c
```diff
@@ -556,8 +556,20 @@ typedef struct search_ctx {
 	RzSearchOpt *opt; ///< User options
 	RzThreadQueue /* RzSearchHits */ *hits; ///< Hits list
 	RzAtomicBool *loop; ///< If set, the execution will continue until it terminates. If unset, the execution cancels.
+	RzThreadQueue /* RzSearchInterval */ *intervals; ///< Interval list
```
Rot127:
```suggestion
	RzThreadQueue /* RzSearchInterval */ *finished_intervals; ///< Interval queue
```

This is the correct interpretation, right?

librz/search/search.c
```diff
@@ -556,8 +556,20 @@ typedef struct search_ctx {
...
 } search_ctx_t;
 
+static void print_intervals(RZ_NONNULL RzThreadQueue *intervals) {
+	rz_return_if_fail(intervals);
+
+	RzSearchInterval *search_interval = NULL;
+	while ((search_interval = rz_th_queue_pop(intervals, false))) {
+		RzInterval *itv = &search_interval->interval;
+		eprintf("[0x%" PFMT64x ", 0x%" PFMT64x "): %zu\n", itv->addr, itv->addr + itv->size,
```
Rot127:
```suggestion
		eprintf("[0x%" PFMT64x ", 0x%" PFMT64x "): %" PFMTSZu "\n", itv->addr, itv->addr + itv->size,
```

Rot127 (changes requested):
Just those nitpicks. Thanks, love this UX stuff!

librz/search/search.c
kazarmy: Yes. A complete commit is in 33b30d29a65e036bc9719cd455fc773bbb1a44c8.

librz/core/cconfig.c
kazarmy: Nothing needs to be done for the `intervals` case and the code can just return true.

librz/core/cconfig.c
```diff
@@ -3747,7 +3762,9 @@ RZ_API int rz_core_config_init(RzCore *core) {
 		NULL);
 	SETPREF("search.prefix", "hit", "Prefix name in search hits label");
 	SETI("search.maxhits", 0, "Maximum number of hits ('0' means no limit)");
-	SETBPREF("search.show_progress", "true", "Show the search process.");
+	n = NODECB("search.show_progress", "1", &cb_search_show_progress);
+	SETDESC(n, "Show the search process (true, false, intervals)");
+	SETOPTIONS(n, "true", "false", "intervals", NULL);
```
wargio: tbh, i don't like this. is not really clear...

kazarmy: how about a12dccf5d976ad7ac852e9ee567ef1250dfe71e7?

Rot127: https://github.com/rizinorg/rizin/commit/a12dccf5d976ad7ac852e9ee567ef1250dfe71e7 is nice IMHO

Rot127 (approved):

PR #4995, opened 2025-03-12
Port `/h` and `/s` to RzSearch.

Rot127:

**Detailed description**

Ports the `/s` (calculate blocks entropies and select those above a threshold) and `/h` to RzSearch and RzShell command `/ce`.

- Add `/ch` command for hash search and an `/ce` alias for `entropy_fract` search.
- Adds the block size as command to `/ch`. So the command no loner depends on `core->blocksize`.
- Extends `RzSearchHit` with `details`. They can be set arbitrary by the search and printed as well. For the entropy search this is the fractured entropy value.

**Test plan**

Added

**Closing issues**

...

### DIFF
--- a/librz/include/rz_search.h
+++ b/librz/include/rz_search.h
@@ -2,6 +2,7 @@
 #define RZ_SEARCH_H
 
 #include <rz_types.h>
+#include <rz_hash.h>
 #include <rz_util.h>
 #include <rz_list.h>
 #include <rz_io.h>
@@ -146,6 +147,18 @@ typedef struct rz_search_interval_t RzSearchInterval;
 
 typedef struct rz_search_collection_t RzSearchCollection;
 
+/**
+ * \brief Data passed to hash search find() workers.
+ */
+typedef struct rz_search_hash_find_data_t RzSearchHashFindData;
+
+typedef struct {
+	char *algo; ///< The hash algorithm used for the.
+	ut8 *hash; ///< Calculated hash.
+	size_t hash_size; ///< Hash size in bytes.
+	char *hash_str; ///< The hash result as printable string.
+} RzSearchiHitDetailHash;
+
 typedef enum {
 	RZ_SEARCH_HIT_DETAIL_NONE = 0,
 	RZ_SEARCH_HIT_DETAIL_HASH,
@@ -232,6 +245,12 @@ typedef enum {
 RZ_API RZ_OWN RzSearchCollection *rz_search_collection_cryptographic();
 RZ_API bool rz_search_collection_cryptographic_add(RZ_NONNULL RzSearchCollection *col, RzSearchCollectionCryptographicType type);
 RZ_API bool rz_search_collection_cryptographic_name_to_type(RZ_NONNULL const char *name, RzSearchCollectionCryptographicType *type);
+
+RZ_API RZ_OWN RzSearchCollection *rz_search_collection_hash();
+RZ_API RZ_OWN RzSearchHashFindData *rz_search_hash_get_find_data(const RzHash *rz_hash, RZ_NONNULL const char *algo_name, RZ_NONNULL const char *expected_digits, const char *block_size_arg);
+RZ_API bool rz_search_collection_hash_add(RZ_NONNULL RzSearchCollection *col, RZ_OWN RZ_NONNULL RzSearchHashFindData *data);
+RZ_API bool rz_search_collection_hash_name_to_type(RZ_NONNULL const char *name);
+RZ_API ut64 rz_search_hash_get_element_size(RZ_NONNULL RzSearchCollection *collection);
 
 /**
  * \brief Maximum value width to search for is currently 64bits/8bytes.
--- a/librz/include/rz_search.h
+++ b/librz/include/rz_search.h
@@ -146,9 +146,16 @@ typedef struct rz_search_interval_t RzSearchInterval;
 
 typedef struct rz_search_collection_t RzSearchCollection;
 
+typedef enum {
+	RZ_SEARCH_HIT_DETAIL_NONE = 0,
+	RZ_SEARCH_HIT_DETAIL_HASH,
+} RzSearchHitDetailType;
+
 typedef struct rz_search_hit_t {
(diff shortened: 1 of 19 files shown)


notxvilka (approved):

librz/include/rz_search.h
```diff
@@ -213,13 +231,31 @@ typedef enum {
...
 	RZ_SEARCH_COLLECTION_CRYPTOGRAPHIC_SAFECURVES,
 	RZ_SEARCH_COLLECTION_CRYPTOGRAPHIC_X509,
+	RZ_SEARCH_COLLECTION_CRYPTOGRAPHIC_END_CRYPTO,
+	RZ_SEARCH_COLLECTION_CRYPTOGRAPHIC_ALL_KEY_BASED, ///< All key based algos from above.
+
+	RZ_SEARCH_COLLECTION_CRYPTOGRAPHIC_BEGIN_HASH, ///< Hash related algorithms
+	RZ_SEARCH_COLLECTION_CRYPTOGRAPHIC_ENTROPY, ///< Block entropies.
+	RZ_SEARCH_COLLECTION_CRYPTOGRAPHIC_HASH, ///< Block hashes.
+	RZ_SEARCH_COLLECTION_CRYPTOGRAPHIC_END_HASH,
+
```
wargio: TBH, i would have a different collection for hash and entropy.

Rot127: I thought about it as well. But decided entropy alone is is too simple to get its own collection. Especially, since there are not additions. Also it fits neatly into the same design of the hash search. The enum is weird though, waited what you would say.
Maybe we can just extract the hash + entropy into a separated collection, so they share one?

But I would like to keep the passing of data to the `find()` workers. Because we might add crypto algos which require it.

Rot127: Never mind, I remove the find() data from the crypto search. There is no need to have dead code for "maybe usage".

librz/include/rz_search.h
```diff
@@ -146,9 +147,48 @@ typedef struct rz_search_interval_t RzSearchInterval;
...
+typedef enum {
+	RZ_SEARCH_COLLECTION_HASH_ADLER32,
+	RZ_SEARCH_COLLECTION_HASH_ENTROPY_FRACT,
+	RZ_SEARCH_COLLECTION_HASH_ENTROPY,
+	RZ_SEARCH_COLLECTION_HASH_MD2,
+	RZ_SEARCH_COLLECTION_HASH_MD5,
+	RZ_SEARCH_COLLECTION_HASH_SHA1,
+	RZ_SEARCH_COLLECTION_HASH_SM3,
+	RZ_SEARCH_COLLECTION_HASH_XOR,
+	RZ_SEARCH_COLLECTION_HASH_CRC,
```
wargio: There are many CRC algos. I would just push the name to RzHash instead of enums.

librz/include/rz_search.h
```diff
@@ -213,14 +255,22 @@ typedef enum {
 	RZ_SEARCH_COLLECTION_CRYPTOGRAPHIC_ECC,
 	RZ_SEARCH_COLLECTION_CRYPTOGRAPHIC_SAFECURVES,
 	RZ_SEARCH_COLLECTION_CRYPTOGRAPHIC_X509,
+	RZ_SEARCH_COLLECTION_CRYPTOGRAPHIC_ALL, ///< All algos from above.
```
wargio: I would use a #define for this setting it to enum_size

librz/include/rz_search.h
```diff
@@ -213,14 +255,22 @@ typedef enum {
...
+
 	// Always the last element to define enum size
 	RZ_SEARCH_COLLECTION_CRYPTOGRAPHIC_ENUM_SIZE,
 } RzSearchCollectionCryptographicType;
 
-RZ_API RZ_OWN RzSearchCollection *rz_search_collection_cryptographic(bool add_all_methods);
+RZ_API RZ_OWN RzSearchCollection *rz_search_collection_cryptographic();
```
wargio, line 245: Revert this?

librz/search/hash_search.c
```diff
@@ -0,0 +1,389 @@
...
+	}
+
+	return rz_mem_eq(a->digits, b->digits, a->digits_len);
+}
+
+static ut8 *parse_digits(RzSearchCollectionHashType algo, const char *expected_digits, ut32 digit_size) {
+	rz_return_val_if_fail(expected_digits, NULL);
+	ut8 *out = RZ_NEWS0(ut8, digit_size);
+	switch (algo) {
+	case RZ_SEARCH_COLLECTION_HASH_ENUM_SIZE:
```
wargio: This should be default

wargio (changes requested):
For the entropy I would forbid it in the hash command and have 3 new commands to which checks if entropy is less than/equal/greater than

librz/include/rz_search.h
Rot127: I'd like to keep it this way. We have `rz_search_collection_cryptographic_add` for adding methods. Adding search method in the initialization function would delude the consistency of the design.
Meaning: After initialization of the any collection, it is expected to be empty. Search methods have to be added in a separated step.

But I extract this change into a separated PR in a few minutes.

Rot127: https://github.com/rizinorg/rizin/pull/5002

Rot127:
> For the entropy I would forbid it in the hash command and have 3 new commands to which checks if entropy is less than/equal/greater than

Equal wouldn't work too well due to rounding errors. The less than case is nice. But can I open an issue about this please? I have too many other tasks to take care of currently.

wargio:
Yes. If you want we can merge it as is and I can just fix this with a pr.
Potentially this can be just a search between x and y values.

wargio:
Ah this probably should be implemented for entropy and fractional entropy

Rot127:
> Yes. If you want we can merge it as is and I can just fix this with a pr.
> Potentially this can be just a search between x and y values.
> Ah this probably should be implemented for entropy and fractional entropy

Sounds good to me.

notxvilka (approved):

wargio (approved):

wargio:
Please rebase

PR #5007, opened 2025-03-17
Add entropy cmds and detail struct for value extraction

wargio:
# SQUASH ME
 

**Detailed description**

- Implements entropy commands `/ce` and `/cef` as its own search collection.
- Adds `RzSearchHitDetail` struct to hold different types of info, such as `bytes`, `string`, `signed integer`, `unsigned integer`, `double`.
- Removes the `comment` field in favour of `RzSearchHitDetail`.
- SM4 keys are now extracted as flag comment when found.
- Magic search now shows the details and sets them as flag comment


librz/include/rz_search.h
```diff
@@ -147,28 +147,9 @@ typedef struct rz_search_interval_t RzSearchInterval;
...
-typedef enum {
-	RZ_SEARCH_HIT_DETAIL_NONE = 0,
-	RZ_SEARCH_HIT_DETAIL_HASH,
-} RzSearchHitDetailType;
-
 typedef struct rz_search_hit_t {
 	char *hit_desc; ///< Hit one word description. If set, it is added to the flag name of the hit. Optional, can be NULL.
 	char *comment; ///< A detailed comment about the hit. Set as flag comment. Optional, can be NULL.
-	void *details; ///< Search dependent details of the hit.
-	RzSearchHitDetailType detail_type; ///< The type of the detail, if any.
```
Rot127, line 171: I would have preferred the more rich details tbh. Our current search only gives very simple results. But this might change with other search spaces like graphs or the knowledge base.

But more importantly API side programming is difficult if the results are returned as string. The tool getting the search hits has to parse the string or calculate the entropy again. Preventing this was the main reason I added these richer details. This would be the case for all searches which process the byte data before making a hit/no-hit decision.

wdyt?

librz/search/hash_search.c
```diff
@@ -68,127 +71,98 @@ static ut8 *parse_digits(const char *algo, const char *expected_digits, ut32 dig
...
+		RZ_LOG_ERROR("search: invalid hash algorithm '%s'.\n", algo_name);
+		return NULL;
+	}
+	// required to reset the status.
+	rz_hash_cfg_final(md);
+
 	ut32 digit_size = rz_hash_cfg_size(md, algo_name);
-	if (digit_size == 0) {
-		RZ_LOG_ERROR("The hash algorithm '%s' is not supported for the search. List supported with the command 'Lh'.\n", algo_name);
+	if (digit_size < 1) {
```
Rot127: `digest` Sorry, this typo sits deep in my head :D

librz/include/rz_search.h
wargio: I think a formatted string is enough and it will appear as a comment

notxvilka: I tend to agree with @Rot127, using rizin as an API makes it more important. I know this by heart because used Rizin as a library from Rust code.

librz/search/hash_search.c
```diff
@@ -68,127 +71,98 @@ static ut8 *parse_digits(const char *algo, const char *expected_digits, ut32 dig
...
-	return hit;
+
+	return rz_search_hit_new(data->algo, address, buf_size, NULL);
 }
 
 static bool hash_find(RzSearchFindOpt *fopts, void *user, ut64 address, const RzBuffer *buffer,
 	RZ_OUT RzThreadQueue *hits, RZ_OUT size_t *n_hits) {
-	RzPVector /*<RzSearchHashFindData *>*/ *pvec = (RzPVector *)user;
-
-	RzHashCfg *hcfg = init_hash_config(pvec);
```
Rot127, line 188: Just for my understanding. The `RzHashCfg` is updated with each `rz_hash_cfg_init()`, `rz_hash_cfg_update()` and `rz_hash_cfg_final()`, correct?

But here it is a shared resource and the different threads add their data to the same `RzHashCfg` instance. So don't we get race conditions?

Rot127 (changes requested):

librz/include/rz_search.h
Rot127: We can keep it like this. I just wanted to consider the API use case.
We can add a better version with the API examples in the book.
This way we can discuss it together again, what is the most ergonomic way.

wargio: Should we have a multi type detail?

librz/search/hash_search.c
wargio: Hmm you might be right and this could be a race condition

librz/include/rz_search.h
wargio: merged into a special structure capable of holding different types of values.

librz/search/search.c
```diff
@@ -790,31 +790,25 @@ RZ_IPI int rz_search_hit_cmp(RZ_NULLABLE RzSearchHit *a, RZ_NULLABLE RzSearchHit
...
- * \param[in]  size      The size of the hit data (can be 0)
+ * \param[in]  hit_desc    The hit description linked to the hit (can be NULL)
+ * \param[in]  address     The address where the hit happened
+ * \param[in]  size        The size of the hit data (can be 0)
+ * \param[in]  hit_detail  Contains additional information about the hit (can be NULL)
  *
  * \return     On success returns a valid pointer, otherwise NULL
  */
-RZ_IPI RZ_OWN RzSearchHit *rz_search_hit_new(const char *hit_desc, ut64 address, size_t size, const char *hit_comment) {
+RZ_IPI RZ_OWN RzSearchHit *rz_search_hit_new(const char *hit_desc, ut64 address, size_t size, RzSearchHitDetail *hit_detail) {
```
Rot127:
```suggestion
RZ_IPI RZ_OWN RzSearchHit *rz_search_hit_new(const char *hit_desc, ut64 address, size_t size, RZ_OWN RzSearchHitDetail *hit_detail) {
```

librz/search/search_internal.h
```diff
@@ -150,14 +162,10 @@ struct rz_search_hash_find_data_t {
 		continue; \
 	}
 
-RZ_IPI RZ_OWN RzSearchHit *rz_search_hit_new(const char *hit_desc, ut64 address, size_t size, const char *hit_comment);
-RZ_IPI RZ_OWN bool rz_search_hit_add_details(RZ_NONNULL RzSearchHit *hit, RzSearchHitDetailType type, RZ_OWN void *details);
+RZ_IPI RZ_OWN RzSearchHit *rz_search_hit_new(const char *hit_desc, ut64 address, size_t size, RzSearchHitDetail *hit_detail);
```
Rot127:
```suggestion
RZ_IPI RZ_OWN RzSearchHit *rz_search_hit_new(const char *hit_desc, ut64 address, size_t size, RZ_OWN RzSearchHitDetail *hit_detail);
```

Rot127 (approved):
Just this one nitpick. Good call with moving the entropy in its own command btw. Search details are also way better than mine. Thanks!

librz/search/search.c
wargio: Rzown should not go there, but nullable should be there.

Rot127: Both then? Because it assigns the pointer directly and not cloning the content. And frees it later. Or was I looking incorrectly?

wargio: ok

notxvilka (approved):
Nice output enhancement!

PR #5013, opened 2025-03-18
Allow virtual files to represent real files.

Rot127:

**Detailed description**

This comes in handy if a bin-plugin has to load real, and more importantly, distinct files from the main one.
The bin plugin API only deals with virtual files. But the buffer of the virtual files, canstill point to a different file then the main one.

If this happened, the read offset calculation was wrong. Because when IOMaps are built the IO module just assumed `virtual file == some segment in the main file`. And to read properly, it increased the read offset by `map->delta`.
This `map->delta` should be 0 though, if the actual buffer read, belongs to a different file than the main file. 

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/librz/bin/bin.c
+++ b/librz/bin/bin.c
@@ -1089,6 +1089,15 @@ RZ_API void rz_bin_map_free(RZ_NULLABLE RzBinMap *map) {
 	free(map);
 }
 
+RZ_API RZ_OWN RzBinMap *rz_bin_map_clone(RZ_NONNULL RzBinMap *map) {
+	rz_return_val_if_fail(map, NULL);
+	RzBinMap *clone = RZ_NEW0(RzBinMap);
+	rz_mem_copy(clone, sizeof(RzBinMap), map, sizeof(RzBinMap));
+	clone->name = rz_str_dup(map->name);
+	clone->vfile_name = map->vfile_name ? rz_str_dup(map->vfile_name) : NULL;
+	return clone;
+}
+
 /**
  * \brief Create a pvector of RzBinMap from RzBinSections queried from the given file
  *
--- a/librz/bin/bin.c
+++ b/librz/bin/bin.c
@@ -1080,6 +1080,36 @@ RZ_API void rz_bin_virtual_file_free(RZ_NULLABLE RzBinVirtualFile *vfile) {
 	free(vfile);
 }
 
+/**
+ * \brief Clones a virtual file. Note, the buffer will never be cloned.
+ * This function will assign the buffer to of \p vfile but always sets buf_owned = false;
+ *
+ * \p vfile The virtual to clone.
+ *
+ * \return A clone of the vfile but with clone->buf_owned = false.
+ */
+RZ_API RZ_OWN RzBinVirtualFile *rz_bin_virtual_file_copy(RZ_BORROW RZ_NONNULL RzBinVirtualFile *vfile) {
+	rz_return_val_if_fail(vfile, NULL);
+	RzBinVirtualFile *clone = RZ_NEW0(RzBinVirtualFile);
+	if (!clone) {
+		return NULL;
+	}
+	clone->represents_real_file = vfile->represents_real_file;
+	clone->buf_owned = false;
+	clone->buf = vfile->buf;
+	if (!clone->buf) {
+		return NULL;
+	}
+	clone->name = rz_str_dup(vfile->name);
+	if (!clone->name) {
+		if (clone->buf_owned) {
+			rz_buf_free(clone->buf);
+		}
+		return NULL;
+	}
+	return clone;
+}
+
 RZ_API void rz_bin_map_free(RZ_NULLABLE RzBinMap *map) {
 	if (!map) {
 		return;
--- a/librz/core/cbin.c
+++ b/librz/core/cbin.c
@@ -847,6 +847,25 @@ static bool io_create_mem_map(RzIO *io, RZ_NULLABLE RzCoreFile *cf, RzBinMap *ma
 	return true;
 }
 
+// If the map's virtual file is actually a real and distinct file, the delta offset is 0.
+// Because the IO layer will read from the file buffer of the virtual file.
+// Not of the file initially opened with Rizin.
+static bool bin_map_references_real_file(const RzBinFile *bf, const RzBinMap *map) {
+	if (!map->vfile_name) {
+		return false;
+	}
+	const RzBinVirtualFile *map_vf = NULL;
+	void **it;
+	rz_pvector_foreach(bf->o->vfiles, it) {
+		RzBinVirtualFile *vf = *it;
+		if (RZ_STR_EQ(vf->name, map->vfile_name)) {
+			map_vf = vf;
+			break;
+		}
+	}
+	return map_vf && map_vf->represents_real_file;
+}
+
 static void add_map(RzCore *core, RZ_NULLABLE RzCoreFile *cf, RzBinFile *bf, RzBinMap *map, ut64 addr, int fd) {
 	RzIODesc *io_desc = rz_io_desc_get(core->io, fd);
 	if (!io_desc || UT64_ADD_OVFCHK(map->psize, map->paddr) ||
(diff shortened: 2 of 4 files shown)


librz/bin/bin.c
```diff
@@ -1080,6 +1080,36 @@ RZ_API void rz_bin_virtual_file_free(RZ_NULLABLE RzBinVirtualFile *vfile) {
...
 
+/**
+ * \brief Clones a virtual file. Note, the buffer will never be cloned.
+ * This function will assign the buffer to of \p vfile but always sets buf_owned = false;
+ *
+ * \p vfile The virtual to clone.
+ *
+ * \return A clone of the vfile but with clone->buf_owned = false.
+ */
+RZ_API RZ_OWN RzBinVirtualFile *rz_bin_virtual_file_copy(RZ_BORROW RZ_NONNULL RzBinVirtualFile *vfile) {
```
thestr4ng3r, line 1091:
```suggestion
RZ_API RZ_OWN RzBinVirtualFile *rz_bin_virtual_file_clone(RZ_BORROW RZ_NONNULL RzBinVirtualFile *vfile) {
```
copy suggests something like memcpy where you provide the destination rather than getting a new instance returned.

librz/core/cbin.c
```diff
@@ -906,8 +925,9 @@ static void add_map(RzCore *core, RZ_NULLABLE RzCoreFile *cf, RzBinFile *bf, RzB
 		perm |= RZ_PERM_X;
 	}
 
+	bool map_references_real_file = bin_map_references_real_file(bf, map);
 	if (size) {
-		RzIOMap *iomap = rz_io_map_add_batch(core->io, fd, perm, map->paddr, addr, size);
+		RzIOMap *iomap = rz_io_map_add_batch(core->io, fd, perm, map_references_real_file ? 0 : map->paddr, addr, size);
```
thestr4ng3r, line 930: I don't quite understand why the `represents_real_file` flag is necessary at all. Can't you just set `map->paddr` to 0 in your bin plugin? This is what the property is meant for after all, to specify the base address in the raw file (or vfile) to map from.

thestr4ng3r (changes requested):

librz/bin/bin.c
thestr4ng3r: I would also like to know the use-case for this. Normally you don't need to clone a virtual file since it can be used in multiple maps as-is.

Rot127: I used it in the `mdt` loader. The `mdt` loader sets up it's virtual file instances (for each firmware file once). And then clones it when they are requested by the core. The core expects to own the virtual files retuned, but the plugin also needs an instance of them. So I clone them with the same file buffer, but the cloned version doesn't have the ownership of the buffer. But I would also revert this here. Actually, I forgot why I added them in this PR anyways.

librz/core/cbin.c
Rot127: Because then every single firmware file would require to be mapped to `paddr == 0` then. Which certainly would break some logic further down.

Quick explanation:

The Qualcomm peripheral firmware format splits up it's components in separated files:

```
ll cdsp.*
-rwxr-xr-x. 1 user user  468 Mar  1 07:41 cdsp.b00
-rwxr-xr-x. 1 user user 7.1K Mar  1 07:41 cdsp.b01
-rwxr-xr-x. 1 user user 230K Mar  1 07:41 cdsp.b02
-rwxr-xr-x. 1 user user 104K Mar  1 07:41 cdsp.b03
-rwxr-xr-x. 1 user user  880 Mar  1 07:41 cdsp.b04
-rwxr-xr-x. 1 user user 2.3M Mar  1 07:41 cdsp.b05
-rwxr-xr-x. 1 user user  87K Mar  1 07:41 cdsp.b06
-rwxr-xr-x. 1 user user 426K Mar  1 07:41 cdsp.b08
-rwxr-xr-x. 1 user user 2.7K Mar  1 07:41 cdsp.b09
-rwxr-xr-x. 1 user user  54K Mar  1 07:41 cdsp.b11
-rwxr-xr-x. 1 user user 7.5K Mar  1 07:41 cdsp.mdt
```

The `mdt` file is an ELF which just describes in the program headers how each file should be mapped in the memory of the peripheral device:

```
> readelf -l cdsp.mdt 

Elf file type is EXEC (Executable file)
Entry point 0x87400000
There are 13 program headers, starting at offset 52

Program Headers:
  Type           Offset   VirtAddr   PhysAddr   FileSiz MemSiz  Flg Align
  NULL           0x000000 0x00000000 0x00000000 0x001d4 0x00000     0
  NULL           0x001000 0x89200000 0x89200000 0x01c18 0x02000     0x1000
  LOAD           0x003000 0xfe000000 0x87400000 0x39798 0xaf000 R E 0x100000
  LOAD           0x03d000 0xfe0af000 0x874af000 0x19dfc 0x1a000 R E 0x1000
  LOAD           0x057000 0xc00c9000 0x874c9000 0x00370 0x01000 R E 0x1000
  LOAD           0x058000 0xc00ca000 0x874ca000 0x239de0 0x45f000 RWE 0x1000
  LOAD           0x292000 0xc0529000 0x87929000 0x15936 0x16000 RW  0x1000
  LOAD           0x2a8000 0xc053f000 0x8793f000 0x00000 0xc1000 R   0x1000
  LOAD           0x2a8000 0x01000000 0x87a00000 0x6a420 0xd0000 RWE 0x1000
  LOAD           0x313000 0x010d0000 0x87ad0000 0x00a88 0x01000 RW  0x1000
  LOAD           0x314000 0xc0600000 0x87ad1000 0x00000 0x2f000 R   0x1000
  LOAD           0x314000 0x74000000 0x87b00000 0x0d694 0x42000 RWE 0x1000
  LOAD           0x322000 0x74042000 0x87b42000 0x00000 0x16be000 RWE 0x1000
```

Note they have all different physical and virtual addresses. Because those are actually describe the peripheral device's memory.

But in our case we still need to keep a _distinct_ file buffer open for each of those segments (`cdsp.b00` - `cdsp.b11`).

This means the `paddr` does not describe the offset into the `cdsp.mdt` file when reading actual file data. It describes the offset in the peripheral device's memory.

It is of course possible to concatenate all the `.bNN` files together and load it. And there is this one script everyone passes around since years for this. But this is not convenient IMHO. I don't want to tell users:

"Please first run this Python script to concatenate all the files to one, so Rizin can deal with it."

I want to have the `rizin cdsp.mdt => done` experience.

So what the RIzin end state should be: the firmware components are mapped as in the memory of the peripheral device, but the actual firmware files are opened separately.

librz/bin/bin.c
thestr4ng3r: Ok, I think it shouldn't be necessary to clone the vfile if the plugin still needs to point to it, but it will be better to late take a closer look again in the pr with the plugin.

librz/core/cbin.c
thestr4ng3r: > This means the paddr does not describe the offset into the cdsp.mdt file when reading actual file data. It describes the offset in the peripheral device's memory.

I see. But this is the significant difference because in rizin "paddr" usually just means an offset in a file rather than an actual physical memory address of a device, especially in io and bin maps. So with that definition, putting the paddr from the elf into the maps is simply wrong.

> Because then every single firmware file would require to be mapped to paddr == 0 then. Which certainly would break some logic further down.

That would be the solution here indeed. Which logic specifically do you expect to be broken by this? Remember you can still have the paddrs from the program headers in the segments (`iSS`) if you want to build something on top of this information.

librz/bin/bin.c
Rot127: The main reason for me to keep an instance of the vfile in the plugin, is for future extension.
I'd like to add commands with the bin plugin to do common tasks with the firmware files.

E.g. they have compressed segments. But the destination address for the uncompressed data is not in the ELF or somewhere else. It has to be reverse engineered. I would like to add a command which just takes the offset and then unpacks the segment to a destination.

Another idea would be to patch a segment, compute the hash and patch the hash in the authentication segment (`.b01`). Stuff like this.

But all this requires to keep the instances of the vfiles, sections, segments etc. in the plugin. So they don't need to be parsed again and again.

librz/core/cbin.c
Rot127: > But this is the significant difference because in rizin "paddr" usually just means

You are absolutely right, it just snapped. Took just a few hours to recognize it :D

PR #5018, opened 2025-03-20
Rename `RzBinMap->paddr` and `RzBinSection->paddr` to `offset`.

Rot127:
# Do not squash

 

**Detailed description**

This name change is an attempt to be more precise what the member actually does.

The paddr member was effectively used as an offset into the binary file.
It holds the offset where the map starts.

The name "paddr" is commonly used for "physical address" though.
But a file offset is not the same as a physical address in most contexts.

A physical address commonly refers to a memory space which has some kind of address
translation unit, which manages the memory space by giving out virtual addresses to
memory users.
The physical address refers then to the actual offset into the memory space
(while the virtual addresses don't need fixed to the same physical memory location).

In the context of RzBinMap the paddr is an offset into the binary file.
Not into a memory space. One could argue that the binary file itself is a "physical memory space",
but this is a rather uncommon interpretation.

More so, if RzBinMap is used in context of binary formats which have the concept of mapping (ELF, PE etc)
this naming can be a source of confusion. Because physical address in those binary formats refers to
actual physical addresses as described above. And offsets into files are referred as offset.
For example ELF has for program headers: p_paddr, p_vaddr and p_offset.
For sections only: sh_offset, sh_addr (sh_addr is virtual or physical depending what the machine supports).

The name change from paddr to offset attempts to fix this potential source of confusion.

**Test plan**

All green

**Closing issues**

...

### DIFF
--- a/librz/arch/p/analysis/analysis_java.c
+++ b/librz/arch/p/analysis/analysis_java.c
@@ -41,7 +41,7 @@ static ut64 java_analysis_find_method(RzAnalysis *a, ut64 addr) {
 		if (!(sec->perm & RZ_PERM_X) || addr < from || addr > to) {
 			continue;
 		}
-		return sec->paddr;
+		return sec->offset;
 	}
 
 	return addr;
--- a/librz/arch/p/asm/asm_java.c
+++ b/librz/arch/p/asm/asm_java.c
@@ -46,7 +46,7 @@ static ut64 java_asm_find_method(RzAsm *a) {
 		if (!(sec->perm & RZ_PERM_X) || addr < from || addr > to) {
 			continue;
 		}
-		return sec->paddr;
+		return sec->offset;
 	}
 
 	return addr;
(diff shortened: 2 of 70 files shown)


notxvilka:
Not sure about it. `paddr` has a nice symmetry to `vaddr`. I think, it's just better to address with documenting instead.
But if you go with renaming, `offset` might become ambiguous too, something like `source_offset` or `fileoffset` might be more on point probably.

Rot127:
> Not sure about it. paddr has a nice symmetry to vaddr. I think, it's just better to address with documenting instead.

Would wait for the comment from @thestr4ng3r and then probably close this. As @wargio already said in Mattermost, it is a general design flaw of the IO not being specific on what address abstraction level it works (OS, MMU, file offsets).

Rot127:
Superseded by https://github.com/rizinorg/rizin/pull/5024

PR #5019, opened 2025-03-20
Fix asm tokens leaking horrendously because they used Vector and PVector

Rot127:

**Detailed description**

I must have missed some places back then, when converting to pvector. And it leaked like there is no tomorrow.
This should fix it now.

https://github.com/rizinorg/rizin/pull/5015 is still valid, but I probably ran the wrong build (without LSAN). Hence thought it fixed it.

**Test plan**

All green

**Closing issues**

...


wargio (approved):

notxvilka (approved):

Rot127:
Still missed some somewhere.

notxvilka (changes requested):
1. Please rebase
2. There is a double free somewhere?
```
[XX] 828 ms   db/esil/ppc_32 long ESIL ppc-32
RZ_NOPLUGINS=1 /usr/bin/rizin -escr.utf8=0 -escr.color=0 -escr.interactive=0 -eflirt.sigdb.load.system=false -esearch.show_progress=false -eflirt.sigdb.load.home=false -N -qc 'e asm.arch=ppc
e asm.bits=32
e cfg.bigendian=true
(pi bytes;wx ${bytes};pi 1;pie 1)
.(pi 4200fff0)
' =
-- stdout
--- expected
+++ actual
@@ --1,4 +-1,3 @@
 bc 0x10, lt, 0xfffffff0
-0x00000000 1,ctr,-=,$z,!,?{,0xfffffff0,pc,=,},

-- stderr
free(): invalid pointer

-- exit status: -6

```

librz/arch/aop.c
```diff
@@ -16,6 +16,7 @@ RZ_API void rz_asm_op_init(RZ_NULLABLE RzAsmOp *op) {
 	if (!op) {
 		return;
 	}
+	rz_asm_token_string_free(op->asm_toks);
```
notxvilka: Interesting solution.

notxvilka (approved):
LGTM. The only doubt I have is that `_free()` function in the `_init()`

librz/arch/aop.c
Rot127: Yes, this is bad. But the to cite the commit messaage:

> In practice rz_asm_op_init() is the best place to free the asm_toks reliably.
Simply because this is always called in the old code, before an instruction is disassembled.
> Only in fini() is not enough, because there are too many place it is not called, leaking the tokens.

wargio: This looks incorrect. We expect this structure to possibly contain garbage.

wargio (changes requested):
This looks wrong

notxvilka (approved):
This approach is certainly better!

notxvilka:
@wargio take a look again please

PR #5022, opened 2025-03-21
Reduce false postives for xrefs based on register+offset.

### LINKED ISSUE #1828: double-clicking address or name in PIE binary seeks to the wrong address
**Environment information**
* Operating System: ALL
* Cutter version: 2.0.3
* Obtained from:
  - Downloaded release from Cutter website or GitHub 
* File format: ELF

**Describe the bug**

**To Reproduce**

1. Open Binary
2. Double-Click on the name or address

**Expected behavior**

Open PIE binary and click on address analyzed as `main`, in this example it's `0x433c`, Cutter seeks to `0x3214` instead.

**Screenshots**

**Additional context**
Example binary is in the attachments.
[ziy.zip](https://github.com/rizinorg/cutter/files/7293778/ziy.zip)

karliss:

**Detailed description**

Make the code detecting xrefs based on instruction `disp` more strict.  Skip it if any other xref check found something.

See #1828 for more details.

**Test plan**

* "db/analysis/x86_32 Fix 15315 - Honor RzAnalysisOp.disp in aar and ao" test not broken
* one new test, to make sure there are no duplicate xrefs for rip relative instruction
* expected output for one test adjusted because it included the incorrectly detected xref

**Closing issues**

closes #1828


karliss:
Confirmed that unit test failure was caused by xref changes.

librz/core/canalysis.c
```diff
@@ -2310,6 +2305,16 @@ RZ_API int rz_core_analysis_search_xrefs(RZ_NONNULL RzCore *core, ut64 from, ut6
 			default:
 				break;
 			}
+
+			// find references
+			if (count - count_before == 0 && op.addr > 512 && op.disp > 512 && op.disp && op.disp != UT64_MAX) {
```
notxvilka: Could you please extract this condition (the `op` part) into a simple inline function?

notxvilka (changes requested):
As for the analysis graph test - high chances that the new output/graph might be better. Did you check the exact difference locally?

karliss:
> chances



> As for the analysis graph test - high chances that the new output/graph might be better. Did you check the exact difference locally?

I am already in the process of doing that.  The 5 which disappeared are probably same type of false positives this PR tries to remove.   (so far checked 2/5)

But there is also a 6th entry which changed content need to further investigate that.

I can observe the 5 removed when comparing output of `aga dot @ main`, unfortunately it only lists reference target but not which disassembly line references it.

I am also comparing output of `afx @ main`. Considering the amount of changes there (1180->743) it's surprising that only 5 got removed from `afx` it just happened that most of false xrefs where hitting something that the fuction referenced properly in some other part of code.

karliss:
The difference:
```diff
-8 [URL="0x00017792", color="lightgray", label="0x00017792"]
-10 [URL="0x00017771", color="lightgray", label="0x00017771"]
-26 [URL="0x000174d0", color="lightgray", label="0x000174d0"]
-29 [URL="0x000174ae", color="lightgray", label="0x000174ae"]
-125 [URL="0x0001b8d7", color="lightgray", label="0x0001b8d7"]

-69 [URL="str.", color="lightgray", label="str."]
+65 [URL="data.00018ae0", color="lightgray", label="data.00018ae0"]
```

Verified that first 5 are something like this:
```
d 0x00005a62 -> 0x0001b8d7 lea rcx, [data.00021340]  # <-- incorrect xref
d 0x00005a62 -> 0x00021340 lea rcx, [data.00021340]
```

Still working on the 6th one
```diff
-d 0x00004b99 -> 0x00013f40 lea rdx, [str.]
-d 0x00004b99 -> 0x00018ae0 lea rdx, [str.]
+d 0x00004b99 -> 0x00018ae0 lea rdx, [data.00018ae0]
```

karliss:
"str." -> "[data.00018ae0]" change is also fine. It still refers to the same thing. 

Change of label is caused by fact the way rizin marks strings (should create a separate PR for fixing that #5023). When there are two identical strings or at least two for which rizin would assign identical labels, only the first one gets labeled as "str.simplified_string".  

The one at 0x00018ae0 didn't have a duplicate, but it was consisting purely from special symbols "*=>@|" so all that remains was "str." and due to removal of  false xrefs order of string processing probably changed and different string consisting purely of special characters got the label "str."

That executable for some reason has quite a few duplicate strings, I am bit surprised the compiler didn't  deduplicate them. Would have been a lot harder to figure out what's happening with 0x18ae0 if there weren't a bunch of real duplicate strings.

wargio (approved):
LGTM once the other comment is addressed

librz/core/canalysis.c
notxvilka: Please address this and we can merge.

karliss: Ok will do it if you really want. Although I am a bit skeptical about it improving readability of code.

notxvilka (approved):

PR #5027, opened 2025-03-22
Removes limit of just printing program entrypoints.

Rot127:

**Detailed description**

entries_initfini_print() with initfini=false is only called by the 'ie' command handler. 'ie' prints entrypoints, so users should also be able to see all entrypoints (of any type) the binary has. This was also the plan as it looks like, because further down the entry point type is also emitted. But since initfini=false for 'ie' only entries of type RZ_BIN_ENTRY_TYPE_PROGRAM were printed.

**Test plan**

All green

**Closing issues**

...


Rot127:
Please check the changes to the tests. Tbh, I didn't get what `initfini` does in the first place. So I might have fixed stuff incorrectly.

Rot127:
Ah ok. If the `.init` and `.fini` section should be included. But still I think the change is valid. Otherwise users don't have a way to print the .fini and .init entries. Alternatively we could add a flag to the command.

notxvilka (approved):
I think the new output is better. Please wait for more opinions though.

wargio (approved):
nice find

kazarmy (approved):

PR #5068, opened 2025-04-04
Add hint about the command name and to settings.

Rot127:
In mattermost we always referred to this command as 'basefind', but it can't be found under this name when searching. Hence, the addition.

 

**Detailed description**

In mattermost we always referred to this command as 'basefind',
but it can't be found under this name when searching (I went to the PR introducing it to find the command).
Hence, the addition.

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/librz/core/cmd_descs/cmd_descs.yaml
+++ b/librz/core/cmd_descs/cmd_descs.yaml
@@ -186,7 +186,7 @@ commands:
     summary: Display or change the block size
     subcommands: cmd_block
   - name: B
-    summary: Computes the possibles firmware locations in memory (CPU intensive)
+    summary: Computes the possibles firmware locations in memory a.k.a basefind (CPU intensive)
     cname: basefind_compute
     type: RZ_CMD_DESC_TYPE_ARGV_STATE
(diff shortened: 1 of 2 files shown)


librz/core/cmd_descs/cmd_descs.yaml
```diff
@@ -199,6 +199,11 @@ commands:
         type: RZ_CMD_ARG_TYPE_CHOICES
         optional: true
         choices: ["32", "64"]
+    details:
+      - name: "Settings of this command"
+        entries:
+          - text: "Check out the settings for this command with 'el basefind.'"
+            comment: ""
```
Rot127, line 206: @wargio You and me tend to put these messages, which only uses one of them, into `text` or `comment`. I prefer `text`, because it has a brighter color with the default theme for me. But with the search I remember you preferred `comment`. We should probably agree on one.

notxvilka (approved):

librz/core/cmd_descs/cmd_descs.yaml
wargio: As you prefer.

wargio (approved):

PR #5076, opened 2025-04-06
/z: Warn if |search literal| < search.str.min_length

kazarmy:

**Detailed description**

For a manual search initiated by `/z`, this pr sets the minimum string length to 1. This is because something like `w ELF; /z ELF` on `=` should just work -- it is not obvious that the previous search failed because `search.str.min_length` was not set properly and it definitely appeared at first glance that Rizin was broken.

~~If a minimum string length is required (e.g. for a regex search), one can use a regex constraint e.g. "`/z .{3,} r`", or a `/zt` with a size constraint e.g. "`/zt:size/ge/3 .+ r`".~~

**Test plan**

**Closing issues**

...


kazarmy:
Added note to [pr description](https://github.com/rizinorg/rizin/pull/5076#issue-2974762406) regarding alternatives for setting a minimum string search length.

wargio (commented):
I'm against this, it should be configurable.
Maybe makes more sense to set the config variable to 1

Rot127:
Alternatively we could also add a log message like "Skipped X detected strings with string length < min_length". But this number will always be enormously high. Especially with `overlap` and `str.encoding=guess` enabled.

kazarmy:
Tbh I currently don't have a good use case ... maybe I'll revisit this later.

wargio:
@kazarmy maybe let's open an issue for 0.9

kazarmy:
If the search string is a literal (not regex), `min_length` should be set to the literal's length, but not yet sure whether `min_length` is in bytes or characters or whether they are equivalent

Rot127:
`min_length` is in characters. also documented:
```
[0x00000000]> el search.str.min_length
search.str.min_length: Smallest string length (in number of characters).
```

But you are right. It should be.

wargio:
Yes you are right. It should be

kazarmy:
ok this is somewhat more problematic than i thought

Rot127:
> ok this is somewhat more problematic than i thought

The string detection part?

kazarmy:
yep, only 1 out of 3 was detected

Rot127:
> yep, only 1 out of 3 was detected

Would suggest to print a warning if a search with less then X characters + `encoding = guess` is done, that the user should consider defining the string encoding with `e str.encoding`. And highlight that they can get false negatives otherwise.
The string detection has more general problems (see: https://github.com/rizinorg/rizin/pull/4762#issuecomment-2629118850) and there is an implementation to fix it. But it is a too large edition to add before v0.8 IMHO.

kazarmy:
> Would suggest to print a warning if a search with less then X characters + `encoding = guess` is done ...

I just want to do d9221b0b9511db55f1daa4821a960ca37c813e28 ... it doesn't appear that `encoding` is relevant here

Rot127:
@kazarmy I am fine with this. Just please: `size_t utf8_len` (no `int`)

kazarmy:
Ok 03c2c2143a308a1c31d585da02979badbef74ae4

Rot127 (approved):

PR #5087, opened 2025-04-14
Split very large basic blocks

### LINKED ISSUE #5086: Unexpected scrolling behavior in file containing mostly 0
**Environment information**
* Operating System: Linux
* Cutter version: 2.3.4
* Obtained from:
  - [ ] Built from source
  - [ ] Distribution repository
* File format: raw binary

**Describe the bug**

Trying to scroll up while view is positioned near the end of 200kB file, causes it to jump up to ~0xfa00 .

Scrubbing through the file using navbar also result in occasional stuttering. Sometimes up to few second UI freezing.

This only happens for the attached file which contains mostly 0x00, and little bit of 0x90  at the start. Repeating the same using file that contains only 0x90 doesn't produce such problems. 

**To Reproduce**

1. Open attached file
2. Drag the mouse back and forth across navbar. Observe occasional freeze.
3. Jump to roughly the end of file using navbar
4. scroll up using scrollwheel
5. observe that the view jumped up from ~0x2f000 to 0xfa00 . 

Tested on x86_64. If you are testing on an arm computer might be necessary to manually switch architecture in advanced settings while opening the file.

**Expected behavior**

View scrolls up by ~3 lines for each wheel tick. All navigation is smooth just like with file containing only 0x90.

**Screenshots**

**Additional context**

I am guessing some kind of edge case in disassembler or rizin interpreting disassembler output.  Or maybe some kind of logic repeatedly attempting to skip all the 0x00 for some reason.

karliss:
Ensure instruction offset fits in uint16.

 

- [x] I've updated the [Rizin book](https://github.com/rizinorg/book) with the relevant information (if needed). rizinorg/book#140

**Detailed description**

Basic block instruction offsets are stored in 16bit integers. Artificially split basic blocks during analysis to ensure the technical limit isn't broken.

Slightly changes meaning of `analysis.bb.maxsize`. Previously setting it to:
* <64k split blocks into disconnected chunks
* 64k-256k split into chunks but incorrectly stored instruction positions causing all kind of bugs
* \>256k does nothing since MAX_FCN_SIZE check would return error first

Previous default was 512k making it useless. After the changes split blocks are connected using jump field.

Introduce `analysis.fnc_max_size` as a replacement limit in case when executable contains something weird or the function analysis fails to stop for some other reason. Currently using `ninstr` since it was the beast easily available size metric. Linear size could potentially be much bigger (especially if function does something like hot/cold code split), and the other size function iterates through all the basic blocks to add up size.

Rest of the changes

read_ahead limit in fcn.c `bytes_read = sizeof(buf);` . The old limit was purely a sanity check so no need to do exact limit when reading bytes for instruction. Reading as much max potential instruction size ensures that it reads whole intstruction and doesn't try to parse partial instruction bytes near the boundary of block split.

canalysis.c next refactoring,  the code adding new item was inside

```
do {
    if (fcnlen == fcnlen == RZ_ANALYSIS_RET_END) {
       ...
        addNext(...);
    }
} while (fcnlen != RZ_ANALYSIS_RET_END);
```
So it would never have more than single entry. Simplify by replacing custom dynamic array code with single variable for single address. That and rest of changes in `__core_analysis_fcn` is from when I attempted to get it working the way `analysis.bb.maxsize` resumed analysis previously.

**Test plan**

Beside the the added test.

* open ~300K file containing 0 (as raw x64 binary)
* `aaa`
* `s 0x23000`
* visual mode
* try to move up

**Closing issues**

Closes #5086

### DIFF
--- a/librz/arch/block.c
+++ b/librz/arch/block.c
@@ -336,6 +336,10 @@ RZ_API bool rz_analysis_block_merge(RzAnalysisBlock *a, RzAnalysisBlock *b) {
 		}
 	}
 
+	if (a->size + b->size > a->analysis->opt.bb_max_size) {
+		return false;
+	}
+
 	// Keep a ref to b, but remove all references of b from its functions
 	rz_analysis_block_ref(b);
 	while (!rz_list_empty(b->fcns)) {
--- a/librz/arch/fcn.c
+++ b/librz/arch/fcn.c
@@ -560,7 +560,7 @@ static RzAnalysisBBEndCause run_basic_block_analysis(RzAnalysisTaskItem *item, R
 	RzAnalysisFunction *fcn = item->fcn;
 	RzStackAddr sp = item->sp;
 	ut64 addr = item->start_address;
-	ut64 len = RZ_MIN(analysis->opt.bb_max_size, UT16_MAX);
+	ut64 len = RZ_MIN(analysis->opt.bb_max_size, RZ_ANALYSIS_BLOCK_MAX_SIZE);
 	ReadAhead read_ahead_cache = { 0 };
 	const int continue_after_jump = analysis->opt.afterjmp;
 	const int addrbytes = analysis->iob.io ? analysis->iob.io->addrbytes : 1;
@@ -789,7 +789,7 @@ static RzAnalysisBBEndCause run_basic_block_analysis(RzAnalysisTaskItem *item, R
 			if (newbbsize >= len) {
 				// Instruction offsets are stored in u16,
 				// artificially introduce bb split to keep the offsets within limits.
-				RzAnalysisBlock* next = fcn_append_basic_block(analysis, fcn, at);
+				RzAnalysisBlock *next = fcn_append_basic_block(analysis, fcn, at);
 				if (!next) {
 					gotoBeach(RZ_ANALYSIS_RET_ERROR);
 				}
--- a/librz/arch/fcn.c
+++ b/librz/arch/fcn.c
@@ -18,10 +18,6 @@
 // 16 KB is the maximum size for a basic block
 #define MAX_FLG_NAME_SIZE 64
 
-// 64KB max size
-// 256KB max function size
-#define MAX_FCN_SIZE (1024 * 256)
-
 #define DB             a->sdb_fcns
 #define EXISTS(x, ...) snprintf(key, sizeof(key) - 1, x, ##__VA_ARGS__), sdb_exists(DB, key)
 #define SETKEY(x, ...) snprintf(key, sizeof(key) - 1, x, ##__VA_ARGS__);
@@ -564,7 +560,7 @@ static RzAnalysisBBEndCause run_basic_block_analysis(RzAnalysisTaskItem *item, R
 	RzAnalysisFunction *fcn = item->fcn;
 	RzStackAddr sp = item->sp;
 	ut64 addr = item->start_address;
-	ut64 len = analysis->opt.bb_max_size;
+	ut64 len = RZ_MIN(analysis->opt.bb_max_size, UT16_MAX);
 	ReadAhead read_ahead_cache = { 0 };
 	const int continue_after_jump = analysis->opt.afterjmp;
 	const int addrbytes = analysis->iob.io ? analysis->iob.io->addrbytes : 1;
@@ -698,11 +694,12 @@ static RzAnalysisBBEndCause run_basic_block_analysis(RzAnalysisTaskItem *item, R
 		}
 	}
 	if ((maxlen - (addrbytes * idx)) > MAX_SCAN_SIZE) {
+		// XXX idx is always 0 here, and maxlen comes from amalysis.bb.maxsize. This makes no sense.
 		RZ_LOG_DEBUG("Skipping large memory region during basic block analysis.\n");
 		maxlen = 0;
 	}
 
-	while (addrbytes * idx < maxlen) {
+	while (true) {
 		ut32 at_delta;
 		ut64 at;
 		if (!last_is_reg_mov_lea) {
@@ -719,7 +716,7 @@ static RzAnalysisBBEndCause run_basic_block_analysis(RzAnalysisTaskItem *item, R
 			rz_analysis_task_item_new(analysis, tasks, fcn, bb, at, sp);
 			break;
 		}
-		ut64 bytes_read = RZ_MIN(len - at_delta, sizeof(buf));
+		ut64 bytes_read = sizeof(buf);
 		ret = read_ahead(&read_ahead_cache, analysis, at, buf, bytes_read);
 
 		if (ret < 0) {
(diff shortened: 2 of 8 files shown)


karliss:
Need to rethink this. As this makes other existing limits ineffective.

karliss:
Summary of situation:

`analysis.bb.maxsize` by default is set to 512k way above 2^16 .
MAX_FCN_SIZE is hardcoded to 256*1024 making `analysis.bb.maxsize` useless, it causes functiona analysis to stop with error when single basic block size exceeds it. 

Clamping `analysis.bb.maxsize` to UINT16_MAX or lower causes it to split larger basic blocks, so it wasn't necessary for me to introduce additional mechanism for it.  But it makes `MAX_FCN_SIZE` ineffective and there doesn't seem to be any other limiter (beside binary size) preventing  function growing infinitely when analysis goes wrong or attempting to analyze memory which isn't actually a proper function.

My proposal is:
* clamp analysis.bb.maxsize to ~UINT16_MAX - 256. Due to implementation details it needs to be at least single instruction under the real limit.
* Introduce a new config variable `analysis.max_fcn_size` and replace existing MAX_FCN_SIZE check with it. Use it to check the total function size not basic block size. Use whatever function size metric is conveniently available. Seems like it might need to be `RzAnalysisFunction::ninstr` since it's the only one directly tracked and stored, size in bytes is dynamically calculated from basic blocks that's not something you want to repeatedly recalculate during analysis of each instruction.
* In the basic block merge functions add check for bb.maxsize

karliss:
Small correction existing analysis.bb.max_size logic splits the block, but doesn't chain them.

librz/arch/block.c
```diff
@@ -336,6 +336,10 @@ RZ_API bool rz_analysis_block_merge(RzAnalysisBlock *a, RzAnalysisBlock *b) {
 		}
 	}
 
+	if (a->size + b->size > a->analysis->opt.bb_max_size) {
+		return false;
```
Rot127, line 340: Print an error message.

librz/core/canalysis.c
```diff
@@ -1005,25 +984,19 @@ static int __core_analysis_fcn(RzCore *core, ut64 at, ut64 from, int reftype, in
...
+			if (has_next) {
+				ut64 newaddr = rz_analysis_function_max_addr(fcn);
+				RzIOMap *map = rz_io_map_get(core->io, newaddr);
+				if (!map || (map && (map->perm & RZ_PERM_X))) {
+					next = newaddr;
+					rz_core_analysis_fcn(core, next, next, 0, depth - 1);
 				}
-				free(next);
+			}
+			if (core->analysis->cur && core->analysis->cur->arch && !strcmp(core->analysis->cur->arch, "x86")) {
```
Rot127, line 995:
```suggestion
			if (core->analysis->cur && core->analysis->cur->arch && RZ_STR_EQ(core->analysis->cur->arch, "x86")) {
```

test/integration/test_analysis_fcn.c
```diff
@@ -0,0 +1,54 @@
...
+		if (block->jump != UT64_MAX) {
+			mu_assert_eq(block->jump, block->addr + block->size, "blocks aren't chained");
+			block = rz_analysis_fcn_bbget_at(core->analysis, f, block->jump);
+		} else {
+			block = NULL;
+		}
+	}
+	mu_assert_eq(end, 240000, "Blocks don't cover whole function");
+
+	rz_core_free(core);
```
Rot127, line 45: `f` and `cf` need to be freed?

Rot127 (approved):
I just have those nitpicks. Although, I struggle to understand the changes in the context of the functions they are in. Because they are so much spaghetti.
But your argument makes sense and since it is tested it lgtm.

Please be aware that we will re-implement the whole analysis loop in the next few months (based on RzIL). So don't spent too much brain power on it in the future.

librz/core/canalysis.c
karliss: I Haven't touched that line of code except moving it few lines up inside the big `if (fcn)`

test/integration/test_analysis_fcn.c
karliss: Are the functions not owned by analysis and transitively the core?

 And I would hope that the file is by the io system although I am less sure about that one.    From a quick glance none of the io unit tests do any manual freeing of individual files beside the io subsystem itself.

Rot127: Didn't check, but it's just optimal. If a dev builds and tests with ASAN, running these integration tests becomes annoying if they leak, because they spam the terminal output.

librz/arch/block.c
karliss: There are 3 other error conditions already none of which print any error messages.  I am hesitant to adding error message in place like this unless it's an optional debug log printed at highest verbosity level, that's what the error return code is for. Deep low level functions are not the place for any user output. Especially in situation like this where failure is normal, requirements for merging two blocks weren't met nothing bad happens function remains in a valid state. As long as caller doesn't ignore return code and doesn't act with assumption that merging always succeeds it should be fine. Looking at how it's used by automerge, it just tries merging more or less everything which could potentially be merged leaving verification of finer requirements to `rz_analysis_block_merge`,  printing an error here would likely result in large amount of useless noise.

wargio: no need to put an error here. this is called recursively.

librz/arch/fcn.c
```diff
@@ -698,11 +694,12 @@ static RzAnalysisBBEndCause run_basic_block_analysis(RzAnalysisTaskItem *item, R
 		}
 	}
 	if ((maxlen - (addrbytes * idx)) > MAX_SCAN_SIZE) {
+		// XXX idx is always 0 here, and maxlen comes from amalysis.bb.maxsize. This makes no sense.
 		RZ_LOG_DEBUG("Skipping large memory region during basic block analysis.\n");
 		maxlen = 0;
 	}
```
wargio, line 700: this looks like remnants of old code. maybe it could be deleted?

wargio (approved):

librz/arch/fcn.c
karliss: As pointed out by Rot127 whole thing is going to be rewritten soon. And there are definitely  a bunch of weird stuff. Can't fix it all in this PR. 

The focus this time was on parts that have direct negative effect by producing bad instruction positions which breaks navigation.

karliss:
What's the state of this? Anything left to do? Can be merged to dev?

Rot127 (approved):
Go ahead I'd say. @wargio @notxvilka

test/integration/test_analysis_fcn.c
```diff
@@ -0,0 +1,54 @@
+// SPDX-FileCopyrightText: 2024 KarlisS <karlis3p70l1ij@gmail.com>
```
notxvilka: 2025

notxvilka (approved):
LGTM with one small nitpick

notxvilka (approved):

wargio (approved):

PR #5094, opened 2025-04-22
rz_core_warn_after_output: Print warnings after command output

kazarmy:

**Detailed description**

This pr adds a facility to rz_core that allows warnings to be printed after command output. This is specifically done for `/z` since there is the possibility that there are so many hits, one has to scroll up to see the warnings.

**Test plan**

The addition makes sense. All builds are green.

**Closing issues**

...

### DIFF
--- a/librz/core/cmd/cmd.c
+++ b/librz/core/cmd/cmd.c
@@ -2792,6 +2792,7 @@ DEFINE_HANDLE_TS_FCN(statements) {
 		RzCmdStatus cmd_res = handle_ts_stmt(state, command);
 		if (state->split_lines) {
 			rz_cons_flush();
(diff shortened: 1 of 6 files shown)


Rot127 (approved):
The idea is good. And for now I think it is ok.

But it feels weird to have a separated function IMHO. As user I would just like to pass a flag to `RZ_WARN`. But I guess this is not possible without major refactoring.

kazarmy:
> As user I would just like to pass a flag to `RZ_WARN`.

C doesn't have default arguments so it would have to be `RZ_LOG_WARN_AFTER` I guess, and that does seem to be a better API.

Thanks for the review.

Issue #4787, opened 2024-12-22
Proper Renesas H8 family support


XVilka:
Currently, Rizin supports only H8/300, also (somewhat) incompletely. But H8 family resembles PIC family in how it contains multiple, sometimes quite different, instruction sets and different bitness. Thus, the code should be:
- Refactored to support multiple ISAs from this family
- Add missing H8 ISAs in addition to H8/300:
	- [x] H8/300L
	- [x] H8/300H
	- [x] H8/500
	- [ ] H8/SX

- Any ESIL remnants removed, and RzIL added for all ISAs: https://github.com/rizinorg/rizin/issues/2080

b1llow:
@notxvilka 

**H8/SX** does not provide opcode binary encoding information. Can we use some of the code related to H8/SX in binutils-gdb as part of the H8/SX disassembler?

https://www.renesas.com/en/document/mah/h8sx-family-software-manual?queryID=3d19fe3eb70ec2c318136e4abcdd7810

https://github.com/bminor/binutils-gdb/blob/2df2e47c5ae84fce2a51cd3d05abbd09f8a752c3/include/opcode/h8300.h

Or do we need to rewrite it based on the opcode table in the code? But that seems like it would require more work, because the code is harder to understand than the documentation.

PR #5119, opened 2025-05-04
Add RZIL support for h8300/h8300h

b1llow:
# SQUASH ME

**Detailed description**

...

**Test plan**

...

**Closing issues**

...


librz/arch/isa/h8300/h8300_il.c
```diff
@@ -0,0 +1,195 @@
+#include "h8300_disas.h"
```
notxvilka, line 4: Dont forget SPDX

notxvilka (changes requested):
Please add "emulateme" test

librz/arch/isa/h8300/h8300_esil.c
```diff
@@ -0,0 +1,492 @@
+// SPDX-FileCopyrightText: 2025 Billow <billow.fun@gmail.com>
```
notxvilka, line 4: Since its the old code, please preserve copyright from the original file

notxvilka (changes requested):

Rot127:
@imbillow Is it fine for you if we mark this PR as "Draft"? I get confused in my notifications.

wargio (commented):
where can i find the h8300 isa manual? is [this](https://www.renesas.com/en/document/gde/h8300-programmig-manual?srsltid=AfmBOorF0AhfRfgb0OWX5I_vvs1bxH4NRtcTSsudKN4uuIPcalDbLtTs) ?

b1llow:
> where can i find the h8300 isa manual? is [this](https://www.renesas.com/en/document/gde/h8300-programmig-manual?srsltid=AfmBOorF0AhfRfgb0OWX5I_vvs1bxH4NRtcTSsudKN4uuIPcalDbLtTs) ?

yes

b1llow:
The h8300 is mostly done, but the elf file I compiled with `h8300-elf-gcc -mn` has h8300h-specific instructions, and considering that h8300h seems to be just an extension of h8300, I'm going to do the two together?

notxvilka:
> The h8300 is mostly done, but the elf file I compiled with `h8300-elf-gcc -mn` has h8300h-specific instructions, and considering that h8300h seems to be just an extension of h8300, I'm going to do the two together?

Yes, please do it in the same PR.

See https://github.com/rizinorg/rizin/issues/4787

Though, I recommend adding only the minimum necessary in this PR (just H8/300H but not the rest), and other variants in separate PRs.

notxvilka (changes requested):
I recommend putting H8/300 extension instructions in a separate test file. It will help later, when more ISAs of the H8 family are added.

notxvilka (changes requested):
Could you please also add analysis information for opcodes? Value, pointer information, etc. Also, please add a few analysis tests for H8300 that show the result of the analysis - function, arguments detection, string reference, etc. To make the support of H8/300 and H8/300H more holistic rather than very basic like it was before. Like you did for Tricore before. It's easier to do in the same PR while you have context of all these instructions, etc.

Rot127 (changes requested):
Just checked the emulateme test so far. The binary seems to use something around 10-20 instructions.

Could you compile one more test binary which uses more instructions?

Just so we can increase the emulation test coverage. Especially for the more complex ones (where the bugs are more likely in my experience).

librz/arch/isa/h8300/h8300_disas.c
```diff
@@ -5,984 +5,1426 @@
...
+	if (id >= RZ_ARRAY_SIZE(register32_names)) {
+		return NULL;
 	}
+	return register32_names[id];
+}
 
-	strncpy(cmd->instr, commands[bytes[0]], H8300_INSTR_MAXLEN - 1);
-	cmd->instr[H8300_INSTR_MAXLEN - 1] = '\0';
+#define SIGN_EXT(value, bits) \
+	((int)((unsigned int)(value) << (32 - (bits))) >> (32 - (bits)))
```
Rot127: Can you move this into a `static inline` function in `rz_bits.h` please?
Feel like we have a few copies of it around already

b1llow:
the `main.elf` is built using this: https://github.com/celerizer/h8300h-test-rom.git

librz/arch/p/analysis/analysis_h8300.c
```diff
@@ -12,666 +13,371 @@
...
+		case H8300_INSN_FORMAT_ABS:
+			op->type = RZ_ANALYSIS_OP_TYPE_CALL;
+			op->jump = INS_OP(0).imm;
+			break;
+		case H8300_INSN_FORMAT_MI8:
+			op->type = RZ_ANALYSIS_OP_TYPE_ICALL;
+			op->ptr = INS_OP(0).imm;
+			op->disp = INS_OP(0).imm;
+			break;
+		default: break;
```
wargio: Maybe set this as op->type = RZ_ANALYSIS_OP_TYPE_ICALL; ?

wargio (approved):
LGTM besides that comment

Rot127 (changes requested):
Generally looks really good!

But I am a little paranoid about the little test coverage the IL tests still give.

This is the coverage I get with `rz-test test/db/rzil/h83000` (see https://github.com/rizinorg/rizin/pull/5289 how to generate it if you don't already have a better way):

[cov.zip](https://github.com/user-attachments/files/21529695/cov.zip)

Do you see a way to maybe add another test which covers some more of the essential instructions (some more shifts, load/store, all branches/jumps).

The thing I am concerned about is, that if some semantics there are wrong, the debugging effort of the bugs emerging from it will be enormous. These bugs are potentially very subtle.

notxvilka (approved):
Great work! LGTM. The next thing would be [adding other ISA from the H8 family](https://github.com/rizinorg/rizin/issues/4787).

notxvilka:
> The thing I am concerned about is, that if some semantics there are wrong, the debugging effort of the bugs emerging from it will be enormous. These bugs are potentially very subtle.

There is H8 emulator but I don't know how good it is: https://github.com/celerizer/libh8300h
Maybe it's suitable for the tracetesting, maybe it's not.

b1llow:
However, there doesn't seem to be a good solution. The ISA description is not as strict as the code in some places, and there doesn't seem to be a reliable simulator for h8300 to perform trace tests.

Rot127 (approved):
Looks good.

Regarding the trace test discussion:

Do you have several binaries? You can just add a `printf(instruction->id)` somewhere in the disassembler. Then run `aaa` and pipe the ids into a `file`.

`cat "file" | sort | uniq -c` can give you the number of most common IDs disassembled. If you have manually tested these most common instructions we should be fine I think.

notxvilka:
Please send all further improvements as separate PRs. Merged this one. Thank you for this one last blocker (apart from nearly finished SPARC) for ESIL deprecation in analysis!

PR #5102, opened 2025-04-26
Make code compilable under `checks_level` 0 and 3

kazarmy:
**Your checklist for this pull request**
- [X] I've read the [guidelines for contributing](https://github.com/rizinorg/rizin/blob/master/DEVELOPERS.md) to this repository.
- [ ] I made sure to follow the project's [coding style](https://github.com/rizinorg/rizin/blob/master/DEVELOPERS.md#code-style).
- [ ] I've documented every `RZ_API` function and struct this PR changes.
- [ ] I've added tests that prove my changes are effective (required for changes to `RZ_API`).
- [ ] I've updated the [Rizin book](https://github.com/rizinorg/book) with the relevant information (if needed).

**Detailed description**

~~This pr attempts to align the functions in `rz_assert.h` to be more in the spirit of the C `assert()` function -- either fail politely or fail spectacularly.~~ This pr makes the code compilable under `checks_level` 0 and 3.

**Test plan**

The changes makes sense. All builds are green.

**Closing issues**

...

PR #5144, opened 2025-05-11
Set `checks_level`=0 on `linux-meson-clang-tests` build

kazarmy:

**Detailed description**

This pr sets `checks_level` to 0 on the `linux-meson-clang-tests` build. This is just to guard against side-effects in code being asserted by `rz_assert.h` macros. This kind of problem should be easy to fix, so it should be ok that the `linux-meson-clang-tests` build is not run on every pr.

**Test plan**

The change makes sense. The `linux-meson-clang-tests` build is green.

**Closing issues**

...

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -72,7 +72,7 @@ jobs:
             build_system: meson
             compiler: clang
             run_tests: true
-            meson_options: -Dbuildtype=release --werror
+            meson_options: -Dbuildtype=release -Dchecks_level=0 --werror
             enabled: ${{ (github.event_name != 'pull_request' || contains(github.head_ref, 'clang')) && needs.changes.outputs.edited == 'true' }}
             timeout: 45
             allow_failure: false


Rot127:
What were the side effects you encountered?
Because we would miss ASAN problems in the assert statements itself this way, right?

kazarmy:
> What were the side effects you encountered?

An example of problematic side-effect code is the original code in [this diff](https://github.com/rizinorg/rizin/pull/5102/files#diff-1516d60b54e295ef4d1c093ce199abefa39ba88b6986b37f7476acb365c4a089) ... `checks_level`=0 will simply remove the first argument of `rz_return_val_if_fail()` as if it didn't exist, and thus the conversion of RL78 operand to string does not happen, resulting in test failures like this:

RL78-test-failure
(https://github.com/rizinorg/rizin/actions/runs/14774758066/job/41480902215#step:18:15463)

This is not an assertion failure and thus `checks_level`=3 will not catch this.

> Because we would miss ASAN problems in the assert statements itself this way, right?

The ASAN builds are still built with `checks_level`=2 so no removal of code will be done (and they probably will never be built with `checks_level`=0 since they are `debugoptimized` builds). This pr doesn't touch the ASAN builds.

Rot127 (approved):
> This pr doesn't touch the ASAN builds.

Was on the go and misread it.
Thanks. All good.

PR #5156, opened 2025-05-19
Add confirmation to large hexdump reads.

vykt:

**Detailed description**

If you mistakenly provide a very large length parameter to `px` (in my case, `px rsp`), rizin will continue the read until it is complete. This can take several minutes, and can't be cancelled with SIGINT.

This PR adds a check via `rz_cons_yesno()`, requesting the user to confirm hexdumps above some threshhold. For now, this threshhold is 0x1000.

If a `rz_core_print_hexdump_or_hexdiff_str()` returns NULL, an additional error message is printed. To avoid this message in case the user answers `n` to the confirmation, the length is now normalised and confirmed before calling `rz_core_print_hexdump_or_hexdiff_str()`.

I imagine the real solution here is to register a temporary SIGINT handler that would cancel the read. I'm inclined to say that an even better solution is probably to register a generic SIGINT handler, and have it call a callback for whichever operation is currently ongoing. I'm not sure how feasible that is given the current architecture. I'm also not familiar with the codebase enough yet to implement something like this.

...

**Test plan**

```
$ rizin -d a.out
> s main
> px 0x40000000
[attempt both y & n paths]
```
I'd be happy to define some tests for this if you'd like.

...

**Closing issues**

N/A
...

### DIFF
--- a/librz/core/cprint.c
+++ b/librz/core/cprint.c
@@ -261,7 +261,7 @@ static inline void len_fixup(RzCore *core, ut64 *addr, int *len) {
 	bool is_positive = *len > 0;
 	if (RZ_ABS(*len) > core->blocksize_max) {
 		RZ_LOG_ERROR("this <len> is too big (0x%" PFMT32x
-			     " < 0x%" PFMT32x ").",
+			     " < 0x%" PFMT32x ").\n",
 			*len, core->blocksize_max);
 		*len = (int)core->blocksize_max;
 	}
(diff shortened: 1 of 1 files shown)


vykt:
I should note that other calls to `rz_core_print_hexdump_or_hexdiff_str()` pass constants below `core->blocksize_max` as the length (`core->blocksize` & `128`), so normalisation can safely be moved up the callstack here.

notxvilka:
```
 Formatting [84/1854] librz/core/cprint.c
librz/core/cprint.c:390:18: error: code should be clang-formatted [-Wclang-format-violations]
        if (len > 0x1000
                        ^
```

PR #5191, opened 2025-06-08
Class Analysis in C++

tushar3q34:

**Detailed description**

- [X] Identify allocated objects
- [ ] Mark allocated objects with class name
- [ ] Devirtualize calls

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/librz/core/cmd/cmd_analysis.c
+++ b/librz/core/cmd/cmd_analysis.c
@@ -1605,13 +1605,21 @@ RZ_IPI RzCmdStatus rz_analysis_recover_rtti_all_handler(RzCore *core, int argc, 
 	return RZ_CMD_STATUS_OK;
 }
 
-RZ_IPI RzCmdStatus rz_analysis_virtual_xrefs_handler(RzCore *core, int argc, const char **argv) {
-	if (argc == 2) {
-		const char *vfunc_name = (const char *)argv[1];
+RZ_IPI RzCmdStatus rz_analysis_virtual_xrefs_handler(RzCore *core, int argc, const char **argv, RzCmdStateOutput *state) {
+	const char *vfunc_name = argv[1];
+	switch (state->mode) {
+	case RZ_OUTPUT_MODE_STANDARD:
 		rz_analysis_virtual_xrefs_print(core->analysis, vfunc_name);
-		return RZ_CMD_STATUS_OK;
+		break;
+	case RZ_OUTPUT_MODE_TABLE: {
+		rz_cmd_state_output_set_columnsf(state, "X", "from");
+		rz_analysis_virtual_xrefs_print_table(core->analysis, vfunc_name, state->d.t);
+		break;
 	}
-	return RZ_CMD_STATUS_ERROR;
+	default:
+		return RZ_CMD_STATUS_ERROR;
+	}
+	return RZ_CMD_STATUS_OK;
 }
 
 RZ_IPI RzCmdStatus rz_analysis_rtti_demangle_class_name_handler(RzCore *core, int argc, const char **argv) {
--- a/librz/core/cmd/cmd_analysis.c
+++ b/librz/core/cmd/cmd_analysis.c
@@ -6099,6 +6099,11 @@ RZ_IPI RzCmdStatus rz_delete_global_imports_handler(RzCore *core, int argc, cons
 	return RZ_CMD_STATUS_OK;
 }
 
+RZ_IPI RzCmdStatus rz_analysis_devirtualize_handler(RzCore *core, int argc, const char **argv) {
+	rz_analysis_devirtualize_methods(core->analysis);
+	return RZ_CMD_STATUS_OK;
+}
+
 RZ_IPI RzCmdStatus rz_analysis_data_handler(RzCore *core, int argc, const char **argv) {
 	int count = argc > 1 ? rz_num_math(core->num, argv[1]) : 2 + (core->blocksize / 4);
 	if (count < 1) {
--- a/librz/core/cmd/cmd_analysis.c
+++ b/librz/core/cmd/cmd_analysis.c
@@ -1605,6 +1605,15 @@ RZ_IPI RzCmdStatus rz_analysis_recover_rtti_all_handler(RzCore *core, int argc, 
 	return RZ_CMD_STATUS_OK;
 }
 
+RZ_IPI RzCmdStatus rz_analysis_virtual_xrefs_handler(RzCore *core, int argc, const char **argv) {
+	if (argc == 2) {
+		const char *vfunc_name = (const char *)argv[1];
+		rz_analysis_virtual_xrefs_print(core->analysis, vfunc_name);
+		return RZ_CMD_STATUS_OK;
+	}
+	return RZ_CMD_STATUS_ERROR;
+}
+
 RZ_IPI RzCmdStatus rz_analysis_rtti_demangle_class_name_handler(RzCore *core, int argc, const char **argv) {
 	if (argc == 2) {
 		char *classname = (char *)argv[1];
--- a/librz/core/cmd_descs/cmd_analysis.yaml
+++ b/librz/core/cmd_descs/cmd_analysis.yaml
@@ -1541,6 +1541,11 @@ commands:
           - RZ_OUTPUT_MODE_RIZIN
           - RZ_OUTPUT_MODE_JSON
         args: []
+      - name: avD
+        summary: devirtualize virtual calls
+        type: RZ_CMD_DESC_TYPE_ARGV
+        cname: analysis_devirtualize
+        args: []
       - name: avg
         summary: Global variables
         subcommands:
--- a/librz/core/cmd_descs/cmd_analysis.yaml
+++ b/librz/core/cmd_descs/cmd_analysis.yaml
@@ -1650,6 +1650,13 @@ commands:
         args:
           - name: classname
             type: RZ_CMD_ARG_TYPE_STRING
+      - name: avx
+        summary: show addresses of virtual calls of virtual function
+        type: RZ_CMD_DESC_TYPE_ARGV
+        cname: analysis_virtual_xrefs
+        args:
+          - name: vfunction name
+            type: RZ_CMD_ARG_TYPE_STRING
   - name: ax
     summary: Cross references (xrefs)
     subcommands:
--- a/librz/core/cmd_descs/cmd_analysis.yaml
+++ b/librz/core/cmd_descs/cmd_analysis.yaml
@@ -1542,7 +1542,7 @@ commands:
           - RZ_OUTPUT_MODE_JSON
         args: []
       - name: avD
-        summary: devirtualize virtual calls
+        summary: Mark objects and devirtualize calls to virtual functions for function at the current offset
         type: RZ_CMD_DESC_TYPE_ARGV
         cname: analysis_devirtualize
         args: []
@@ -1651,11 +1651,14 @@ commands:
           - name: classname
             type: RZ_CMD_ARG_TYPE_STRING
       - name: avx
-        summary: show addresses of virtual calls of virtual function
-        type: RZ_CMD_DESC_TYPE_ARGV
+        summary: Show the addresses of calls to the virtual function.
+        type: RZ_CMD_DESC_TYPE_ARGV_STATE
         cname: analysis_virtual_xrefs
+        modes:
+          - RZ_OUTPUT_MODE_STANDARD
+          - RZ_OUTPUT_MODE_TABLE
         args:
-          - name: vfunction name
+          - name: virtual function name
             type: RZ_CMD_ARG_TYPE_STRING
   - name: ax
     summary: Cross references (xrefs)
--- a/librz/core/devirtualize_cxx.c
+++ b/librz/core/devirtualize_cxx.c
@@ -7,9 +7,14 @@
 #include "core_private.h"
 
 /**
+ * \file
+ * Partially uses ESIL uplifting but will be ported to RzIL
+ */
+
+/**
  * \brief Used to describe current state of tainting
  */
-typedef enum {
+typedef enum rz_taint_devirt_t {
 	RZ_TAINT_MODE_OFF, ///< No tainting in this mode
 	RZ_TAINT_MODE_RED, ///< Mode when `new` is identified but constructor is yet to be identified
 	RZ_TAINT_MODE_BLUE, ///< Mode after constructor is also identified, write to variable is yet to be identified
@@ -21,22 +26,25 @@ typedef struct var_t {
 	ut64 instr_addr; ///< address of instruction which writes to variable
 	RzList /*<char *>*/ *class_names; ///< Single variable might store multiple classes based on conditionals
 	RzList /*<ut64 *>*/ *object_addr; ///< Address of allocated object which stores vtable ptr
-} Variable;
+} CppVariable;
 
 typedef struct rz_taint_state_t {
-	RzTaintDevirt taint;
-	RzAnalysisOp *op;
-	ut64 addr;
-	char *class_name;
-	RzVariableBook *var_book;
+	RzTaintDevirt taint; ///< current taint mode
+	RzAnalysisOp *op; ///< analysis instruction op
+	ut64 addr; ///< current instruction address
+	char *class_name; ///< class name to be stored in variable
+	RzCppVariableBook *var_book; ///< collection of all stack variables
 } RzTaintState;
 
 typedef struct rz_virtual_calls_t {
-	HtUP /*<ut64,RzSetS*>*/ *virt_calls;
+	HtUP /*<ut64,RzSetS *>*/ *virt_calls; ///< hash map of instruction address and class method name
 } RzVirtualCalls;
 
 #define RZ_TAINT_VALUE 0x1A2B3C
 
+/**
+ * \return list of `RzAnalysisFunction *` of all allocator i.e. `new` functions
+ */
 RzList /*<RzAnalysisFunction *>*/ *list_new_fcns(RzCore *core) {
 	RzList *list = rz_analysis_function_list(core->analysis);
 	if (!list) {
@@ -54,13 +62,13 @@ RzList /*<RzAnalysisFunction *>*/ *list_new_fcns(RzCore *core) {
 	return ret_list;
 }
 
-static bool rz_taint_vm_step(RzCore *core, RzTaintDevirt *taint) {
+static void rz_taint_vm_step(RzCore *core) {
 	if (!rz_core_il_step(core, 1)) {
-		return false;
+		return;
 	}
 
 	if (!core->analysis || !core->analysis->il_vm) {
-		return false;
+		return;
 	}
 
 	RzILVM *vm = core->analysis->il_vm->vm;
@@ -77,7 +85,7 @@ static bool rz_taint_vm_step(RzCore *core, RzTaintDevirt *taint) {
 		case RZ_IL_EVENT_MEM_WRITE: {
 			ut64 new_val = rz_bv_to_ut64(evt->data.mem_write.new_value);
 			if (new_val == RZ_TAINT_VALUE) {
-				return true;
+				return;
 			}
 		}
 		default: {
@@ -85,10 +93,12 @@ static bool rz_taint_vm_step(RzCore *core, RzTaintDevirt *taint) {
 		}
 		}
 	}
-	return false;
 }
 
-static void xrefs_of_new(RzList /*<RzAnalysisFunction *>*/ *list, RzAnalysis *analysis, ut64 addr) {
+/**
+ * \brief adds xrefs to an addr to `list`
+ */
+static void xrefs_of_new(RZ_OUT RzList /*<RzAnalysisFunction *>*/ *list, RzAnalysis *analysis, ut64 addr) {
 	RzList *curr_list = rz_analysis_xrefs_get_to(analysis, addr);
 	RzListIter *it;
 	RzAnalysisXRef *xref;
@@ -121,8 +131,12 @@ static int compare(const void *void_value, const void *void_list_data, void *use
 	}
 }
 
+/**
+ * \brief initialises stack and registers for tainting
+ */
 void rz_taint_init(RzAnalysis *analysis, RzCore *core) {
-	rz_core_analysis_esil_init_mem(core, NULL, 0x1000, 0x1050); // Memory allocation
+	// Random Memory allocation representing function stack
+	rz_core_analysis_esil_init_mem(core, NULL, 0x1000, 0x1050);
 	rz_core_analysis_il_reinit(core); // initializing VM
 	// TODO : Make general for other archs
 	if (!rz_str_cmp(analysis->arch_target->arch, "x86", -1)) {
@@ -134,7 +148,10 @@ void rz_taint_init(RzAnalysis *analysis, RzCore *core) {
 	}
 }
 
-RzList /*<Variable *>*/ *get_variable_writes(RzAnalysisFunction *fcn) {
+/**
+ * \return list of all stack variables written in the function call
+ */
+static RzList /*<CppVariable *>*/ *get_variable_writes(RzAnalysisFunction *fcn) {
 	void **it;
 	RzList *var_list = rz_list_new();
 	rz_pvector_foreach (&fcn->vars, it) {
@@ -145,7 +162,7 @@ RzList /*<Variable *>*/ *get_variable_writes(RzAnalysisFunction *fcn) {
 				continue;
 			}
 			ut64 addr = fcn->addr + acc->offset;
-			Variable *write_var = RZ_NEW0(Variable);
+			CppVariable *write_var = RZ_NEW0(CppVariable);
 			write_var->instr_addr = addr;
 			write_var->name = rz_str_dup(var->name);
 			write_var->class_names = rz_list_new();
@@ -156,15 +173,21 @@ RzList /*<Variable *>*/ *get_variable_writes(RzAnalysisFunction *fcn) {
 	return var_list;
 }
 
-static Variable *var_at_write(RzTaintState *state, RzList /*<Variable *>*/ *var_write_list, ut64 stack_addr) {
+/**
+ * \param state current tainting state
+ * \param var_write_list list of all stack variables\
+ * \param stack_addr stack address accessed during write operation
+ * \return returns variable with stack address \p stack_addr
+ */
+static CppVariable *var_at_write(RzTaintState *state, RzList /*<CppVariable *>*/ *var_write_list, ut64 stack_addr) {
 	HtUP *stack_vars = state->var_book->class_variables;
 	bool found = false;
-	Variable *ht_var = ht_up_find(stack_vars, stack_addr, &found);
+	CppVariable *ht_var = ht_up_find(stack_vars, stack_addr, &found);
 	if (found) {
 		return ht_var;
 	}
 	RzListIter *it;
-	Variable *var;
+	CppVariable *var;
 	rz_list_foreach (var_write_list, it, var) {
 		if (var->instr_addr == state->addr) {
 			var->stack_addr = stack_addr;
@@ -176,6 +199,10 @@ static Variable *var_at_write(RzTaintState *state, RzList /*<Variable *>*/ *var_
 	return NULL;
 }
 
+// TODO : Convert this into a hashmap of func name v/s class name
+/**
+ * \brief returns class name from function name
+ */
 static char *get_class_name_from_func(const char *func_name) {
 	// Get class name from method.class_name.init
 	const char *first = strchr(func_name, '.');
@@ -199,6 +226,10 @@ static char *get_class_name_from_func(const char *func_name) {
 	return class_name;
 }
 
+/**
+ * \param addr instruction address of constructor call
+ * \return class name if a valid constructor call else `NULL`
+ */
 static char *get_class_name(RzAnalysis *analysis, ut64 addr) {
 	RzList *list = rz_analysis_xrefs_get_from(analysis, addr);
 	RzListIter *it;
@@ -232,7 +263,10 @@ static bool is_call_instruction(RzAnalysisOp *op) {
 	}
 }
 
-static void add_taint_value(RzAnalysis *analysis) {
+/**
+ * \brief changes the value in register to `RZ_TAINT_VALUE` during tainting
+ */
+static void taint_register(RzAnalysis *analysis) {
 	RzCore *core = analysis->core;
 	if (!rz_str_cmp(analysis->arch_target->arch, "x86", -1)) {
 		rz_core_analysis_il_vm_set(core, "rax", RZ_TAINT_VALUE);
@@ -241,6 +275,9 @@ static void add_taint_value(RzAnalysis *analysis) {
 	}
 }
 
+/**
+ * \return stack address of memory write operation else 0
+ */
 static ut64 rz_var_stack_address_track(RzAnalysis *analysis) {
 	RzCore *core = analysis->core;
 	if (!rz_core_il_step(core, 1)) {
@@ -273,7 +310,11 @@ static ut64 rz_var_stack_address_track(RzAnalysis *analysis) {
 	return 0;
 }
 
-static void add_class_name_to_var(Variable *var, char *name) {
+/**
+ * \param var variable containing object(s)
+ * \brief adds class name to \p var (can have multiple class names)
+ */
+static void add_class_name_to_var(CppVariable *var, char *name) {
 	RzList *class_names = var->class_names;
 	RzListIter *it;
 	char *elem_name;
@@ -285,7 +326,12 @@ static void add_class_name_to_var(Variable *var, char *name) {
 	rz_list_insert(class_names, 0, rz_str_dup(name));
 }
 
-static void define_and_mark_variable(RzAnalysis *analysis, RzTaintState *state, const Variable *var) {
+/**
+ * \param state current tainting state
+ * \param var variable to be marked
+ * \brief labels the type of the variable with appropriate class name
+ */
+static void define_and_mark_variable(RzAnalysis *analysis, RzTaintState *state, const CppVariable *var) {
 	RzList *class_names = var->class_names;
 	ut64 len = rz_list_length(class_names);
 	char *type = NULL;
@@ -323,6 +369,11 @@ static void define_and_mark_variable(RzAnalysis *analysis, RzTaintState *state, 
 	free(var_type);
 }
 
+/**
+ * \param state current taint state
+ * \brief tainting step when state if `RZ_TAINT_MODE_OFF`
+ * \return `true` if skipping an instruction is required else `false`
+ */
 static bool rz_taint_off_step(RzAnalysis *analysis, RzTaintState *state) {
 	RzList *list = list_new_fcns(analysis->core);
 	RzList *list_xrefs = rz_list_new();
@@ -336,7 +387,7 @@ static bool rz_taint_off_step(RzAnalysis *analysis, RzTaintState *state) {
 	RzCore *core = analysis->core;
 	if (rz_list_find(list_xrefs, p_addr, compare, NULL) != NULL) {
 		state->taint = RZ_TAINT_MODE_RED;
-		add_taint_value(analysis);
+		taint_register(analysis);
 		free_xrefs_of_new(list_xrefs);
 		rz_list_free(list);
 		return true;
@@ -346,12 +397,17 @@ static bool rz_taint_off_step(RzAnalysis *analysis, RzTaintState *state) {
 		rz_list_free(list);
 		return true;
 	}
-	rz_taint_vm_step(core, &(state->taint));
+	rz_taint_vm_step(core);
 	free_xrefs_of_new(list_xrefs);
 	rz_list_free(list);
 	return false;
 }
 
+/**
+ * \param state current taint state
+ * \brief tainting step when state if `RZ_TAINT_MODE_RED`
+ * \return `true` if skipping an instruction is required else `false`
+ */
 static bool rz_taint_red_step(RzAnalysis *analysis, RzTaintState *state) {
 	RzAnalysisOp *op = state->op;
 	RzCore *core = analysis->core;
@@ -367,10 +423,15 @@ static bool rz_taint_red_step(RzAnalysis *analysis, RzTaintState *state) {
 		state->taint = RZ_TAINT_MODE_BLUE;
 		return true;
 	}
-	rz_taint_vm_step(core, &(state->taint));
+	rz_taint_vm_step(core);
 	return false;
 }
 
+/**
+ * \param state current taint state
+ * \brief tainting step when state if `RZ_TAINT_MODE_BLUE`
+ * \return `true` if skipping an instruction is required else `false`
+ */
 static bool rz_taint_blue_step(RzAnalysis *analysis, RzTaintState *state) {
 	RzAnalysisOp *op = state->op;
 
@@ -383,7 +444,7 @@ static bool rz_taint_blue_step(RzAnalysis *analysis, RzTaintState *state) {
 	if (stack_addr == 0) {
 		return false;
 	}
-	Variable *var = var_at_write(state, state->var_book->stack_variables, stack_addr);
+	CppVariable *var = var_at_write(state, state->var_book->stack_variables, stack_addr);
 	add_class_name_to_var(var, state->class_name);
 	define_and_mark_variable(analysis, state, var);
 	state->taint = RZ_TAINT_MODE_OFF;
@@ -398,7 +459,7 @@ static void free_taint_state(RzTaintState *state) {
 	free(state);
 }
 
-static void free_variable(Variable *var) {
+static void free_variable(CppVariable *var) {
 	RzListIter *it;
 
 	// Free class names
@@ -421,7 +482,12 @@ static void free_variable(Variable *var) {
 	free(var);
 }
 
-RZ_API RzVariableBook *rz_analysis_mark_classes(RzAnalysis *analysis) {
+/**
+ * \param analysis rizin analysis object
+ * \brief marks all variables that store object(s) with respective class name(s)
+ * \return pointer to `RzCppVariableBook` book containing marked variable information
+ */
+RZ_API RzCppVariableBook *rz_analysis_mark_classes(RzAnalysis *analysis) {
 
 	RzList *list = list_new_fcns(analysis->core);
 
@@ -452,7 +518,7 @@ RZ_API RzVariableBook *rz_analysis_mark_classes(RzAnalysis *analysis) {
 
 	rz_taint_init(analysis, core);
 
-	RzVariableBook *var_book = RZ_NEW0(RzVariableBook);
+	RzCppVariableBook *var_book = RZ_NEW0(RzCppVariableBook);
 	var_book->class_variables = ht_up_new(NULL, NULL);
 	var_book->function = function;
 	var_book->stack_variables = var_write_list;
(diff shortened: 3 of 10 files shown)


librz/core/devirtualize_cxx.c
```diff
@@ -0,0 +1,234 @@
...
+
+/**
+ * `RZ_TAINT_MODE_OFF`   : No tainting
+ * `RZ_TAINT_MODE_RED`   : `new` identified, taint till constructor
+ * `RZ_TAINT_MODE_BLUE`  : constructor identified, taint till var write
+ */
+typedef enum {
+	RZ_TAINT_MODE_OFF,
+	RZ_TAINT_MODE_RED,
+	RZ_TAINT_MODE_BLUE,
```
Rot127:
```suggestion
	RZ_TAINT_MODE_OFF, ///< No tainting
	RZ_TAINT_MODE_RED, ///< `new` identified, taint till constructor - Be a little more descriptive here. Or describe the algorithm in more detail somewhere else and refer to it.
	RZ_TAINT_MODE_BLUE, ///< Constructor identified, taint until variable write - Be more descriptive as well, as above.
```

Better to be consistent in the whole code base.

Also, please write out the sentences with proper grammar and all. Using slang or abbreviations can be hard for people with English as their second language.

For the details, you can also describe the meaning of the colors in more detail by describing the enum with `\brief `

librz/core/devirtualize_cxx.c
```diff
@@ -0,0 +1,234 @@
...
+RzList *list_of_operator_new_functions(RzCore *core) {
+	RzList *list = rz_analysis_function_list(core->analysis);
+	if (!list) {
+		return NULL;
+	}
+	RzListIter *it;
+	RzAnalysisFunction *fcn;
+	RzList *ret_list = rz_list_new();
+	rz_list_foreach (list, it, fcn) {
+		char *res = strstr(fcn->name, "sym.imp.operator_new");
```
Rot127, line 87: Don't hardcode the names please. Users must be able to define them via an options for example.

librz/core/devirtualize_cxx.c
```diff
@@ -0,0 +1,234 @@
...
+		}
+		default: {
+			break;
+		}
+		}
+	}
+	return false;
+}
+
+void xrefs_of_new(RzList *list, RzAnalysis *analysis, ut64 addr) {
```
Rot127:
```suggestion
static void xrefs_of_new(RzList *list, RzAnalysis *analysis, ut64 addr) {
```

librz/core/devirtualize_cxx.c
```diff
@@ -0,0 +1,234 @@
...
+	RzList *curr_list = rz_analysis_xrefs_get_to(analysis, addr);
+	RzListIter *it;
+	RzAnalysisXRef *xref;
+	rz_list_foreach (curr_list, it, xref) {
+		rz_list_append(list, &(xref->from));
+	}
+}
+
+int compare(const void *void_value, const void *void_list_data, void *user) {
```
Rot127:
```suggestion
static int compare(const void *void_value, const void *void_list_data, void *user) {
```

Rot127: Make the arguments `ut64 *` if you know they are one. You ca cast the function pointer down below..

librz/core/devirtualize_cxx.c
```diff
@@ -0,0 +1,234 @@
...
+	// TODO : Make general for other archs
+	rz_core_analysis_il_vm_set(core, "rbp", 0x1fff);
+	rz_core_analysis_il_vm_set(core, "rsp", 0x1fff);
+	rz_core_analysis_il_vm_set(core, "rax", RZ_TAINT_VALUE);
+}
+
+RzList *get_variable_writes(RzAnalysisFunction *fcn) {
+	void **it;
+	RzList *var_list = rz_list_new();
+	rz_pvector_foreach (&fcn->vars, it) {
```
Rot127, line 179: These variables are discovered by ESIL. You can't depend on  them I am afraid.

librz/core/devirtualize_cxx.c
```diff
@@ -0,0 +1,234 @@
...
+				refresh_vm = true;
+				taint = RZ_TAINT_MODE_BLUE;
+			} else {
+				rz_taint_step(core, &taint);
+			}
+			break;
+		}
+		case RZ_TAINT_MODE_BLUE: {
+			bool ret = rz_taint_step(core, &taint);
+			if (ret) {
```
Rot127: Invert conditions and break early

librz/core/devirtualize_cxx.c
```diff
@@ -0,0 +1,561 @@
...
+
+	bool refresh_vm = false;
+
+	while (start < end) {
+		rz_analysis_op(analysis, op, start, bytes + offset, end - start, RZ_ANALYSIS_OP_MASK_ALL);
+		if (refresh_vm) {
+			rz_core_analysis_il_reinit(core);
+			refresh_vm = false;
+		}
+		refresh_vm = is_call_instruction(op);
```
notxvilka: This function could be probably split to avoid so many depth levels

librz/core/devirtualize_cxx.c
```diff
@@ -0,0 +1,561 @@
...
+		old_type = rz_str_newf("RZ_%s_HYBRID", var->name);
+		var_type = old_type;
+	}
+
+	char *exists = rz_type_format(analysis->typedb, old_type);
+	if (exists) {
+		rz_type_db_del(analysis->typedb, old_type);
+	}
+	rz_types_define(analysis->core, type);
+	RzType *v_type = rz_type_parse_string_single(analysis->typedb->parser, var_type, NULL);
```
notxvilka: I think this is very round-about way to define a type. Just use the helper APIs to create new type directly. If it doesn't exist - just add new function, but please avoid creating a string then calling a parser - that would add so many unnecessary computations...

librz/core/devirtualize_cxx.c
```diff
@@ -0,0 +1,561 @@
...
+	case RZ_ANALYSIS_OP_TYPE_RCALL:
+		return true;
+	case RZ_ANALYSIS_OP_TYPE_JMP:
+		return true;
+	default:
+		return false;
+	}
+}
+
+static void add_taint_value(RzAnalysis *analysis) {
```
notxvilka: It doesn't taint any value, just a particular register, the name probably should reflect that

librz/core/devirtualize_cxx.c
```diff
@@ -0,0 +1,561 @@
...
+			var->stack_addr = stack_addr;
+			ht_up_insert(stack_vars, stack_addr, var);
+			rz_list_insert(state->var_book->class_var_list, 0, var);
+			return var;
+		}
+	}
+	return NULL;
+}
+
+static char *get_class_name_from_func(const char *func_name) {
```
notxvilka: It feels like a hack, tbh

notxvilka (changes requested):

librz/core/devirtualize_cxx.c
tushar3q34: Yes it is, and I plan to change it later. Eventually, this will be replaced by a hash table of key as constructor methods and class name as value. I didn't add it now because I was adding too many things at a time so I avoided it to get things tested at least.

notxvilka (changes requested):
The biggest part is needed to be done first, even if incomplete - tests

tushar3q34:
> The biggest part is needed to be done first, even if incomplete - tests

Okay, I will add them asap. Was resolving memory leaks but that is done partially so CI will be red in a few cases.

librz/core/devirtualize_cxx.c
```diff
@@ -0,0 +1,736 @@
...
+
+#include <rz_util.h>
+#include <rz_analysis.h>
+#include <rz_core.h>
+#include "core_private.h"
+
+/**
+ * \brief Used to describe current state of tainting
+ */
+typedef enum {
```
notxvilka: We proably will need to generalize tainting once @Rot127 ESIL to RzIL migration is done. But at this point it's fine as is.

librz/core/devirtualize_cxx.c
```diff
@@ -0,0 +1,736 @@
...
+	RZ_TAINT_MODE_BLUE, ///< Mode after constructor is also identified, write to variable is yet to be identified
+} RzTaintDevirt;
+
+typedef struct var_t {
+	char *name; ///< for now, checks varibles discovered by ESIL
+	ut64 stack_addr; ///< address of the stack where variable is stored according to RzIL VM
+	ut64 instr_addr; ///< address of instruction which writes to variable
+	RzList /*<char *>*/ *class_names; ///< Single variable might store multiple classes based on conditionals
+	RzList /*<ut64 *>*/ *object_addr; ///< Address of allocated object which stores vtable ptr
+} Variable;
```
notxvilka: Please avoid such a generic name, especially in case we will need to put it in a header file some time after

librz/core/devirtualize_cxx.c
```diff
@@ -0,0 +1,736 @@
+// SPDX-FileCopyrightText: 2025 tushar3q34 <tushar3q34@gmail.com>
+// SPDX-License-Identifier: LGPL-3.0-only
+
+#include <rz_util.h>
+#include <rz_analysis.h>
+#include <rz_core.h>
+#include "core_private.h"
+
```
notxvilka, line 8: Add a `\file` note that currently it uses ESIL but later will be ported to RzIL. cc @Rot127

librz/core/devirtualize_cxx.c
```diff
@@ -0,0 +1,736 @@
...
+typedef struct var_t {
+	char *name; ///< for now, checks varibles discovered by ESIL
+	ut64 stack_addr; ///< address of the stack where variable is stored according to RzIL VM
+	ut64 instr_addr; ///< address of instruction which writes to variable
+	RzList /*<char *>*/ *class_names; ///< Single variable might store multiple classes based on conditionals
+	RzList /*<ut64 *>*/ *object_addr; ///< Address of allocated object which stores vtable ptr
+} Variable;
+
+typedef struct rz_taint_state_t {
```
notxvilka, line 60: Doxygen pls

librz/core/devirtualize_cxx.c
notxvilka: In this particular case the "operator_new" part is fine. C++ has no way to rename such symbol, IIRC

librz/core/devirtualize_cxx.c
```diff
@@ -0,0 +1,736 @@
...
+		return 0;
+	} else if (*value > *list_data) {
+		return 1;
+	} else {
+		return -1;
+	}
+}
+
+void rz_taint_init(RzAnalysis *analysis, RzCore *core) {
+	rz_core_analysis_esil_init_mem(core, NULL, 0x1000, 0x1050); // Memory allocation
```
notxvilka: Please add a comment why these particular values

librz/core/devirtualize_cxx.c
notxvilka: As we discussed earlier on Mattermost this is okay to depend on them. Once RzIL migration is complete, variables should work as before, just they will be created by the RzIL-based analysis instead.

librz/core/devirtualize_cxx.c
```diff
@@ -0,0 +1,736 @@
...
+
+	// TODO : Check for leaks
+	// TODO : Implement free for lists
+	free_taint_state(state);
+	rz_list_free(list);
+	free(bytes);
+	return var_book;
+}
+
+static ut64 allocate_and_store(RzCore *core, RzVector /*<RzAnalysisVtable>*/ *vtables, Variable *var, ut64 addr) {
```
notxvilka: Despite being internal, I think these functions deserve a one line comment explaining what they are for

librz/core/devirtualize_cxx.c
notxvilka: Don't forget this one

notxvilka (changes requested):
You can also expand `ac` commands to show some of the information without relying on the disassembly every time. Easier for testing and more succinct output.

librz/core/devirtualize_cxx.c
wargio: This can be CppClassVariable

tushar3q34: I named it as `CppVariable` because it is used for all stack variables, not just the ones that store class objects.

librz/core/devirtualize_cxx.c
tushar3q34: Please check this and resolve if it looks fine.

librz/core/devirtualize_cxx.c
tushar3q34: These values were randomly taken by me. Ideally, there should be a user option added both for the stack of function and memory allocation for objects. Any suggestions on how to include this as an option for user?

librz/core/cmd_descs/cmd_analysis.yaml
```diff
@@ -1541,6 +1541,11 @@ commands:
           - RZ_OUTPUT_MODE_RIZIN
           - RZ_OUTPUT_MODE_JSON
         args: []
+      - name: avD
+        summary: devirtualize virtual calls
```
notxvilka: Probably worth adding detailed description as well

librz/core/cmd_descs/cmd_analysis.yaml
```diff
@@ -1645,6 +1650,13 @@ commands:
         args:
           - name: classname
             type: RZ_CMD_ARG_TYPE_STRING
+      - name: avx
+        summary: show addresses of virtual calls of virtual function
+        type: RZ_CMD_DESC_TYPE_ARGV
+        cname: analysis_virtual_xrefs
```
notxvilka, line 1656: Add multiple modes, default one and table

notxvilka (changes requested):
1. Rebase and fix cmd linter (regenerate C files)
2. Please address the previous feedback, including one about types creation

Then we can merge, I think.

notxvilka:
Please also rename the PR title and description to reflect the contents better.

librz/core/cmd/cmd_analysis.c
```diff
@@ -1605,6 +1605,15 @@ RZ_IPI RzCmdStatus rz_analysis_recover_rtti_all_handler(RzCore *core, int argc,
 	return RZ_CMD_STATUS_OK;
 }
 
+RZ_IPI RzCmdStatus rz_analysis_virtual_xrefs_handler(RzCore *core, int argc, const char **argv) {
+	if (argc == 2) {
```
Rot127: This check is unecessary if no argument is optional. The cmd parser will handle it before.

librz/core/devirtualize_cxx.c
```diff
@@ -200,37 +201,60 @@ static CppVariable *var_at_write(RzTaintState *state, RzList /*<CppVariable *>*/
...
+		SdbKv *kv = *iter;
+		const char *class_name = sdbkv_key(kv);
+
+		RzVector *methods = rz_analysis_class_method_get_all(analysis, class_name);
+		RzAnalysisMethod *meth;
+		rz_vector_foreach (methods, meth) {
+			ut64 addr = meth->addr;
+			RzAnalysisFunction *function = rz_analysis_get_fcn_in(core->analysis, addr, RZ_ANALYSIS_FCN_TYPE_ROOT);
+			if (!function) {
+				function = rz_analysis_get_fcn_in(core->analysis, addr, 0);
```
Rot127: No magic numbers please. What does `0` means here?

librz/core/devirtualize_cxx.c
```diff
@@ -645,6 +675,37 @@ static void rz_preserve_stack(RzAnalysis *analysis, RzCppVariableBook *var_book,
 	fill_value_at_stack_address(core, m_var->stack_addr, obj_addr);
 }
 
+/**
+ * \brief devirtualizing step to mark virtual calls
+ */
+static void devirtualize_step(RzCore *core, RzVirtualCalls *virt_calls, RzAnalysisOp *op, ut64 start) {
+	RzAnalysisILVM *vm = core->analysis->il_vm;
+	RzILVal *reg = rz_il_vm_get_var_value(vm->vm, RZ_IL_VAR_KIND_GLOBAL, op->reg);
+	if (reg) {
```
Rot127: Invert condition return early.

librz/core/devirtualize_cxx.c
```diff
@@ -645,6 +675,37 @@ static void rz_preserve_stack(RzAnalysis *analysis, RzCppVariableBook *var_book,
...
+
+		RzBitVector *bv = rz_il_value_to_bv(reg);
+		ut64 vf_addr = rz_bv_to_ut64(bv);
+		rz_core_analysis_il_vm_set(core, op->reg, 0);
+		rz_bv_free(bv);
+
+		RzAnalysisFunction *vfunc = rz_analysis_get_fcn_in(core->analysis, vf_addr, RZ_ANALYSIS_FCN_TYPE_ROOT);
+		if (!vfunc) {
+			vfunc = rz_analysis_get_fcn_in(core->analysis, vf_addr, 0);
```
Rot127: Magic `0`

librz/core/devirtualize_cxx.c
```diff
@@ -645,6 +675,37 @@ static void rz_preserve_stack(RzAnalysis *analysis, RzCppVariableBook *var_book,
...
+		}
+		if (vfunc) {
```
Rot127: Invert return early.

librz/core/devirtualize_cxx.c
```diff
@@ -769,6 +806,23 @@ static bool free_virt_calls(void *user, const ut64 key, const void *v) {
 	return true;
 }
 
+static bool add_virtual_xref(RzAnalysis *analysis, const ut64 key, RzSetS *vfunc_set) {
+	HtSP *virtual_xref = analysis->ht_cpp_virtual_xrefs;
+	RzPVector *pvect = rz_set_s_to_vector(vfunc_set);
+	void **it;
+	rz_pvector_foreach (pvect, it) {
+		const char *vfunc = (const char *)(*it);
```
Rot127: Cast isn't necessary

Rot127 (changes requested):

librz/core/cmd/cmd_analysis.c
tushar3q34: No argument is not optional for this command. Should I make it as optional?

librz/core/devirtualize_cxx.c
notxvilka: This one still left

notxvilka (changes requested):
Ok, thanks for addressing most of the feedback, just a type one left. Don't forget to rebase on top of the latest `dev` as well.

tushar3q34:
Yes working on the type one.

librz/core/devirtualize_cxx.c
tushar3q34: Please check and resolve this if it's okay

notxvilka (approved):
LGTM from my side. Lets wait for others.

librz/core/devirtualize_cxx.c
```diff
@@ -0,0 +1,956 @@
...
+		case RZ_IL_EVENT_MEM_WRITE: {
+			ut64 new_val = rz_bv_to_ut64(evt->data.mem_write.new_value);
+			if (new_val == RZ_TAINT_VALUE) {
+				return;
+			}
+		}
+		default: {
+			break;
+		}
+		}
```
wargio: maybe just make an `if (evt->type != RZ_IL_EVENT_MEM_WRITE) continue;` otherwise
```suggestion
		switch (evt->type) {
		case RZ_IL_EVENT_MEM_WRITE: {
			ut64 new_val = rz_bv_to_ut64(evt->data.mem_write.new_value);
			if (new_val == RZ_TAINT_VALUE) {
				return;
			}
			break;
		}
		default:
			break;
		}
```

librz/core/devirtualize_cxx.c
```diff
@@ -0,0 +1,956 @@
...
+static int compare(const void *void_value, const void *void_list_data, void *user) {
+	const ut64 *value = void_value;
+	const ut64 *list_data = void_list_data;
+	if (*value == *list_data) {
+		return 0;
+	} else if (*value > *list_data) {
+		return 1;
+	} else {
+		return -1;
+	}
```
wargio:
```suggestion
	}
	return -1;
```

librz/core/devirtualize_cxx.c
```diff
@@ -0,0 +1,956 @@
...
+
+	rz_pvector_foreach (vm->events, it) {
+		evt = *it;
+		switch (evt->type) {
+		case RZ_IL_EVENT_MEM_WRITE: {
+			return rz_bv_to_ut64(evt->data.mem_write.address);
+		}
+		default: {
+			break;
+		}
```
wargio:
```suggestion
		default:
			break;
```

librz/core/devirtualize_cxx.c
```diff
@@ -0,0 +1,956 @@
...
+
+	RzCppVariableBook *var_book = RZ_NEW0(RzCppVariableBook);
+	var_book->class_variables = ht_up_new(NULL, NULL);
+	var_book->function = function;
+	var_book->stack_variables = var_write_list;
+	var_book->class_var_list = rz_list_new();
+	state->var_book = var_book;
+
+	state->method_class_map = ht_sp_new(HT_STR_DUP, NULL, free);
+	get_method_class_map(core, state->method_class_map);
```
wargio: Check each pointer you allocated for null & check if `rz_io_read_at` fails.

librz/core/devirtualize_cxx.c
```diff
@@ -0,0 +1,956 @@
...
+static ut64 store_objects(RzCore *core, RzVector /*<RzAnalysisVtable>*/ *vtables, CppVariable *var, ut64 addr) {
+	RzAnalysisVTable *vtable;
+	rz_vector_foreach (vtables, vtable) {
+		ut8 buf[8];
+		rz_write_le64(buf, vtable->addr);
+		char str[17]; // 2 chars per byte + null terminator
+		for (int i = 0; i < 8; i++) {
+			sprintf(&str[i * 2], "%02x", buf[i]);
+		}
+		str[16] = '\0'; // null-terminate
```
wargio: `rz_hex_bin2str(buf, sizeof(buf), str);`

wargio: maybe call `str` as `hexpair` ?

librz/core/devirtualize_cxx.c
```diff
@@ -0,0 +1,956 @@
...
+
+		rz_core_write_hexpair(core, addr, str);
+		ut64 *ptr = RZ_NEW0(ut64);
+		*ptr = addr;
+		rz_list_insert(var->object_addr, 0, ptr);
+		addr += 0x8;
```
wargio: check for NULL pointer

librz/core/devirtualize_cxx.c
```diff
@@ -0,0 +1,956 @@
...
+static void fill_value_at_stack_address(RzCore *core, ut64 stack_addr, ut64 value) {
+	ut8 buf[8];
+	rz_write_le64(buf, value);
+	char str[17];
+	for (int i = 0; i < 8; i++) {
+		sprintf(&str[i * 2], "%02x", buf[i]);
+	}
+	str[16] = '\0';
+	rz_core_write_hexpair(core, stack_addr, str);
+}
```
wargio: Maybe move this up and use it also in `store_objects`. also use `rz_hex_bin2str(buf, sizeof(buf), str);`

librz/core/devirtualize_cxx.c
```diff
@@ -0,0 +1,956 @@
...
+
+static bool add_comment(void *user, const ut64 key, const void *v) {
+	RzCore *core = (RzCore *)user;
+	RzSetS *set = (RzSetS *)v;
+
+	RzPVector *vect = rz_set_s_to_vector(set);
+	void **it;
+
+	char *comment = NULL;
```
wargio: Use `RzStrBuf` apis.

wargio (changes requested):
mostly ok, many small issues.

librz/core/cmd/cmd_analysis.c
Rot127: No, since you need it. But in this case the `if (argc == 2)` condition can be removed

librz/core/cmd/cmd_analysis.c
```diff
@@ -1605,6 +1605,27 @@ RZ_IPI RzCmdStatus rz_analysis_recover_rtti_all_handler(RzCore *core, int argc,
 	return RZ_CMD_STATUS_OK;
 }
 
+RZ_IPI RzCmdStatus rz_analysis_virtual_xrefs_handler(RzCore *core, int argc, const char **argv, RzCmdStateOutput *state) {
+	if (argc == 2) {
+		const char *vfunc_name = (const char *)argv[1];
```
Rot127:
```suggestion
		const char *vfunc_name = argv[1];
```

librz/core/cmd_descs/cmd_analysis.yaml
```diff
@@ -1541,6 +1541,11 @@ commands:
           - RZ_OUTPUT_MODE_RIZIN
           - RZ_OUTPUT_MODE_JSON
         args: []
+      - name: avD
+        summary: mark objects and devirtualize calls to virtual functions
```
Rot127:
```suggestion
        summary: Mark objects and devirtualize calls to virtual functions.
```

For these user facing docs please take are of grammar and spelling.

librz/core/cmd_descs/cmd_analysis.yaml
```diff
@@ -1645,6 +1650,16 @@ commands:
         args:
           - name: classname
             type: RZ_CMD_ARG_TYPE_STRING
+      - name: avx
+        summary: show addresses of virtual calls of virtual function
```
Rot127:
```suggestion
        summary: Show the addresses of virtual calls to the virtual function.
```

Rot127: Or shorten it to `Show the addresses of calls to the virtual function.` because the "virtual" in "virtual call" is implied when we call about virtual functions.

librz/core/cmd_descs/cmd_analysis.yaml
```diff
@@ -1645,6 +1650,16 @@ commands:
...
+        type: RZ_CMD_DESC_TYPE_ARGV_STATE
+        cname: analysis_virtual_xrefs
+        modes:
+          - RZ_OUTPUT_MODE_STANDARD
+          - RZ_OUTPUT_MODE_TABLE
+        args:
+          - name: vfunction name
```
Rot127:
```suggestion
          - name: virtual function name
```

librz/core/devirtualize_cxx.c
```diff
@@ -0,0 +1,956 @@
...
+ * \brief Used to describe current state of tainting
+ */
+typedef enum rz_taint_devirt_t {
+	RZ_TAINT_MODE_OFF, ///< No tainting in this mode
+	RZ_TAINT_MODE_RED, ///< Mode when `new` is identified but constructor is yet to be identified
+	RZ_TAINT_MODE_BLUE, ///< Mode after constructor is also identified, write to variable is yet to be identified
+} RzTaintDevirt;
+
+typedef struct var_t {
+	char *name; ///< for now, checks varibles discovered by ESIL
```
Rot127:
```suggestion
	char *name; ///< ESIL variable name.
```

librz/core/devirtualize_cxx.c
```diff
@@ -0,0 +1,956 @@
...
+	RzTaintDevirt taint; ///< current taint mode
+	RzAnalysisOp *op; ///< analysis instruction op
+	ut64 addr; ///< current instruction address
+	char *class_name; ///< class name to be stored in variable
+	RzCppVariableBook *var_book; ///< collection of all stack variables
+	HtSP /*<char *,char *>*/ *method_class_map; ///< maps methods to class name
+} RzTaintState;
+
+typedef struct rz_virtual_calls_t {
+	HtUP /*<ut64,RzSetS *>*/ *virt_calls; ///< hash map of instruction address and class method name
```
Rot127, line 70:
```suggestion
	HtUP /*<ut64,RzSetS *>*/ *virt_calls; ///< Maps instruction addresses to the class method names they call.
```

librz/core/devirtualize_cxx.c
```diff
@@ -0,0 +1,956 @@
...
+} RzVirtualCalls;
+
+#define RZ_TAINT_VALUE 0x1A2B3C
```
Rot127: Is this a commonly know good value?
Because I think it might be better to be odd, so it doesn't randomly conflict with real addresses (which are commonly aligned to 2 or 4). If this is just some random value, better replace it with one which has bit0 set.

librz/core/devirtualize_cxx.c
```diff
@@ -0,0 +1,956 @@
...
+
+/**
+ * \return list of `RzAnalysisFunction *` of all allocator i.e. `new` functions
+ */
+RzList /*<RzAnalysisFunction *>*/ *list_new_fcns(RzCore *core) {
```
Rot127:
```suggestion
static RzList /*<RzAnalysisFunction *>*/ *list_new_fcns(RzCore *core) {
```

librz/core/devirtualize_cxx.c
```diff
@@ -0,0 +1,956 @@
...
+		return;
+	}
+
+	RzILVM *vm = core->analysis->il_vm->vm;
+
+	void **it;
+	RzILEvent *evt;
+
+	rz_config_set_cb(core->config, "rzil.step.events.read", "true", NULL);
+	rz_config_set_cb(core->config, "rzil.step.events.write", "true", NULL);
```
Rot127: Why do you disable them here? If necessary please add a comment explaining it.
Otherwise remove them

librz/core/devirtualize_cxx.c
Rot127: Second the `if` check over `switch`

librz/core/devirtualize_cxx.c
```diff
@@ -0,0 +1,956 @@
...
+			break;
+		}
+		}
+	}
+}
+
+/**
+ * \brief adds xrefs to an addr to `list`
+ */
+static void xrefs_of_new(RZ_OUT RzList /*<RzAnalysisFunction *>*/ *list, RzAnalysis *analysis, ut64 addr) {
```
Rot127:
```suggestion
static void xrefs_of_new(RZ_OUT RzList /*<ut64 *>*/ *list, RzAnalysis *analysis, ut64 addr) {
```

But the list could also just store primitive ut64 instead of an allocated one, right?

librz/core/devirtualize_cxx.c
```diff
@@ -0,0 +1,956 @@
...
+		return 1;
+	} else {
+		return -1;
+	}
+}
+
+/**
+ * \brief initialises stack and registers for tainting
+ */
+void rz_taint_init(RzAnalysis *analysis, RzCore *core) {
```
Rot127:
```suggestion
static void rz_taint_init(RzAnalysis *analysis, RzCore *core) {
```

librz/core/devirtualize_cxx.c
```diff
@@ -0,0 +1,956 @@
...
+			write_var->class_names = rz_list_new();
+			write_var->object_addr = rz_list_new();
+			rz_list_push(var_list, write_var);
+		}
+	}
+	return var_list;
+}
+
+/**
+ * \param state current tainting state
```
Rot127, line 199: Add a `\brief`

librz/core/devirtualize_cxx.c
```diff
@@ -0,0 +1,956 @@
...
+ * \param var_write_list list of all stack variables\
```
Rot127:
```suggestion
 * \param var_write_list List of all stack variables.
```

librz/core/devirtualize_cxx.c
```diff
@@ -0,0 +1,956 @@
...
+static bool is_call_instruction(RzAnalysisOp *op) {
+	switch (op->type) {
+	case RZ_ANALYSIS_OP_TYPE_CALL:
+		return true;
+	case RZ_ANALYSIS_OP_TYPE_RCALL:
+		return true;
+	case RZ_ANALYSIS_OP_TYPE_JMP:
+		return true;
+	default:
+		return false;
```
Rot127:
```suggestion
	case RZ_ANALYSIS_OP_TYPE_CALL:
	case RZ_ANALYSIS_OP_TYPE_RCALL:
	case RZ_ANALYSIS_OP_TYPE_JMP:
		return true;
	default:
		return false;
```

librz/core/devirtualize_cxx.c
```diff
@@ -0,0 +1,956 @@
...
+		if (class_name) {
+			rz_list_free(list);
+			return class_name;
+		}
+	}
+	rz_list_free(list);
+	return NULL;
+}
+
+static bool is_call_instruction(RzAnalysisOp *op) {
```
Rot127: There are many call types missing and includes jumps. So better rename it to something more descriptive. `is_branch_type_to_method` maybe? To signal these specific op types are common to jump/call methods?

librz/core/devirtualize_cxx.c
```diff
@@ -0,0 +1,956 @@
...
+		return true;
+	case RZ_ANALYSIS_OP_TYPE_JMP:
+		return true;
+	default:
+		return false;
+	}
+}
+
+/**
+ * \brief changes the value in register to `RZ_TAINT_VALUE` during tainting
```
Rot127:
```suggestion
 * \brief Changes the value in the `ARG0` register to `RZ_TAINT_VALUE`.
```

librz/core/devirtualize_cxx.c
```diff
@@ -0,0 +1,956 @@
...
+		return 0;
+	}
+
+	RzILVM *vm = core->analysis->il_vm->vm;
+
+	void **it;
+	RzILEvent *evt;
+
+	rz_config_set_cb(core->config, "rzil.step.events.read", "true", NULL);
+	rz_config_set_cb(core->config, "rzil.step.events.write", "true", NULL);
```
Rot127: Same as above, why is this necessary?

librz/core/devirtualize_cxx.c
```diff
@@ -0,0 +1,956 @@
...
+			break;
+		}
+		}
+	}
+	return 0;
+}
+
+/**
+ * \param var variable containing object(s)
+ * \brief adds class name to \p var (can have multiple class names)
```
Rot127:
```suggestion
 * \brief Adds a class name to \p var (can have multiple class names).
 * \param var Variable containing object(s)
```

`name` is missing

librz/core/devirtualize_cxx.c
```diff
@@ -0,0 +1,956 @@
...
+ */
+static void add_class_name_to_var(CppVariable *var, char *name) {
```
Rot127:
```suggestion
static void add_class_name_to_var(CppVariable *var, const char *name) {
```

librz/core/devirtualize_cxx.c
```diff
@@ -0,0 +1,956 @@
...
+	rz_analysis_devirtualize(analysis, var_book);
+	free_variable_book(var_book);
+}
+
+static bool print_virtual_xrefs(RzCore *core, ut64 key, void *val) {
+	rz_cons_printf("C 0x%08" PFMT64x " ", key);
+	rz_core_seek(core, key, true);
+	rz_core_print_disasm_instructions(core, 0, 1);
+	return true;
+}
```
Rot127: Please don't call commands this way. Use the API instead.

librz/core/devirtualize_cxx.c
```diff
@@ -0,0 +1,956 @@
...
+	RzCore *core = analysis->core;
+	HtSP *ht_virtual_xrefs = analysis->ht_cpp_virtual_xrefs;
+	bool found = false;
+	RzSetU *set = ht_sp_find(ht_virtual_xrefs, vfunc, &found);
+	if (!found) {
+		RZ_LOG_ERROR("Cannot find virtual xrefs to function %s", vfunc);
+		return;
+	}
+	rz_cons_printf("Virtual xrefs to %s\n", vfunc);
+	ht_up_foreach(set, (HtUPForeachCallback)print_virtual_xrefs, core);
```
Rot127, line 964: You can also iterate over the hash map with an `RzIterator`. This might be more natural.

Rot127: Checkout `ht_up_as_iter` or `ht_up_as_iter_keys`

librz/core/devirtualize_cxx.c
```diff
@@ -0,0 +1,956 @@
...
+
+	free(var);
+}
+
+/**
+ * \param analysis rizin analysis object
+ * \brief marks all variables that store object(s) with respective class name(s)
+ * \return pointer to `RzCppVariableBook` book containing marked variable information
+ */
+RZ_API RzCppVariableBook *rz_analysis_mark_classes(RzAnalysis *analysis) {
```
Rot127, line 558: Please fix the spelling in doxygen and add the type annotations.

librz/core/devirtualize_cxx.c
```diff
@@ -0,0 +1,956 @@
...
+	RzCore *core = analysis->core;
+
+	RzAnalysisFunction *function = rz_analysis_get_fcn_in(core->analysis, core->offset, RZ_ANALYSIS_FCN_TYPE_ROOT);
+	if (!function) {
+		function = rz_analysis_get_fcn_in(core->analysis, core->offset, RZ_ANALYSIS_FCN_TYPE_NULL);
+	}
+	if (!function) {
+		RZ_LOG_ERROR("Cannot find function at 0x%08" PFMT64x "\n", core->offset);
+		return;
+	}
```
Rot127, line 782: Please correct me if I missed something. But the devirtualization only works for the function at the current offset. It doesn't do it for all functions. Right?

But the `avD` command description suggests it does it for all functions.
I think this is also how it should be tbh.

librz/core/devirtualize_cxx.c
```diff
@@ -0,0 +1,956 @@
...
+		return rz_str_dup(class_name);
+	}
+	return NULL;
+}
+
+/**
+ * \param addr instruction address of constructor call
+ * \return class name if a valid constructor call else `NULL`
+ */
+static char *get_class_name(RzAnalysis *analysis, RzTaintState *state) {
```
Rot127, line 279: Doxygen is off. Please add `\brief` description what class name it returns

librz/core/devirtualize_cxx.c
```diff
@@ -0,0 +1,956 @@
...
+		rz_core_analysis_il_vm_set(core, "rax", RZ_TAINT_VALUE);
+	} else if (!rz_str_cmp(analysis->arch_target->arch, "arm", -1)) {
+		rz_core_analysis_il_vm_set(core, "x0", RZ_TAINT_VALUE);
+	}
+}
+
+/**
+ * \return stack address of memory write operation else 0
+ */
+static ut64 rz_var_stack_address_track(RzAnalysis *analysis) {
```
Rot127: Please correct me if this is wrong, but the function returns any value which was written to a register. If it is a stack address or something else is not known.

So the function name seems misleading?

librz/core/devirtualize_cxx.c
```diff
@@ -0,0 +1,956 @@
...
+		state->addr += op->size;
+		offset += op->size;
+		core->offset = state->addr;
+		rz_analysis_op_fini(op);
+	}
+	core->offset = function->addr;
+	rz_core_analysis_il_reinit(core);
+
+	// TODO : Check for leaks
+	// TODO : Implement free for lists
```
Rot127:
```suggestion
```

librz/core/devirtualize_cxx.c
```diff
@@ -0,0 +1,956 @@
...
+
+/**
+ * \param var_book collection of stack variables
+ * \brief allocates simulated objects of all class variables
+ */
+static void allocate_objects(RzAnalysis *analysis, RzCppVariableBook *var_book) {
+	RzList *var_list = var_book->class_var_list;
+	RzListIter *it;
+	CppVariable *var;
+	ut64 addr = 0x3000;
```
Rot127, line 689: What is this magic address?

librz/core/devirtualize_cxx.c
Rot127: Or add an argument to the command with the function address. This is probably better due to time constraints cc @notxvilka @wargio

Rot127 (changes requested):
I have quite a lot of nitpicks. Most of them are about documentation.

I might not be necessary for this PR. But for the next one please ensure that the functions' doxygen comments are:
- Complete with all brief, return, and all parameters.
- Have proper spelling and grammar (Sentences start with capital letters and such).
- Are in the correct order (first `brief`, then `parameters`, then `return`.

For this one please fix at least:
- the order of comments (put `\brief` first) and correct mistakes in there.
- fix/add the doxygen to all API functions. The description should be more extensive in this case.

Lastly, please add an overview of your algorithm to the `\file` description.
It is a little hard to understand just by reading the code what the analysis cycle is.

librz/core/devirtualize_cxx.c
```diff
@@ -0,0 +1,945 @@
...
+// SPDX-License-Identifier: LGPL-3.0-only
+
+#include <rz_util.h>
+#include <rz_analysis.h>
+#include <rz_core.h>
+#include "core_private.h"
+
+/**
+ * \file
+ * Partially uses ESIL uplifting but will be ported to RzIL
```
wargio: Maybe describe how the detection works etc..

Examples:
- https://github.com/rizinorg/rizin/blob/dev/librz/diff/diff.c#L5
- https://github.com/rizinorg/rizin/blob/dev/librz/bin/p/bin_nin64.c#L5
- https://github.com/rizinorg/rizin/blob/dev/librz/sign/pat.c#L5

librz/core/devirtualize_cxx.c
wargio: we could use as tainted value one that is unaligned

librz/core/devirtualize_cxx.c
wargio: @tushar3q34 this was not fixed.

librz/core/devirtualize_cxx.c
wargio: @tushar3q34 same here.

librz/core/devirtualize_cxx.c
tushar3q34: I think that's true for vector but not for list. I'll check it once again though.

wargio: why not just use RzVector?

librz/core/devirtualize_cxx.c
tushar3q34: Please check the value, now it is odd.

librz/core/devirtualize_cxx.c
tushar3q34: This is a random address for simulating objects. Should we take this as an optional argument to the command?

librz/core/devirtualize_cxx.c
tushar3q34: I changed description of the command `avD`. So the user can call it as `avD @ <func_name>`. Should I add it as an argument?

librz/core/devirtualize_cxx.c
tushar3q34: Will keep in mind to use this for later PRs.

librz/core/devirtualize_cxx.c
notxvilka: I think this way is good enough.

notxvilka (approved):
LGTM from my side.

librz/core/devirtualize_cxx.c
notxvilka: I think just adding a comment about that is enough at this stage.

librz/core/devirtualize_cxx.c
wargio: no, i agree, you can do then `avD @@F` for all the functions

wargio (approved):

PR #5212, opened 2025-06-19
[RzEgg] Remove Static variables in `emit_arm.c`

ahmed-kamal2004:

**Detailed description**

Created a non-public struct `EmitArmArg`, eliminating global/static variables in file `emit_arm.c`
> `typedef struct emit_arm_arg {`
> `	int lastarg;`
>`	char lastargs[16][32];`
>`} EmitArmArg;`

...

**Test plan**

Testing with RzEgg based on `emit_x86`, `emit_x64`, `emit_arm` and `emit_trace`
...

**Closing issues**

...

### DIFF
--- a/librz/egg/egg.c
+++ b/librz/egg/egg.c
@@ -115,9 +115,11 @@ RZ_API void rz_egg_free(RzEgg *egg) {
 	if (!egg) {
 		return;
 	}
+
 	if (egg->remit && egg->remit->fini) {
-		egg->remit->fini(egg);
+		egg->remit->fini(egg->remit);
 	}
+
 	rz_path_free(egg->sys_path);
 	rz_buf_free(egg->src);
 	rz_buf_free(egg->buf);
@@ -144,6 +146,9 @@ RZ_API void rz_egg_reset(RzEgg *egg) {
 
 RZ_API bool rz_egg_setup(RzEgg *egg, const char *arch, int bits, int endian, const char *os) {
 	const char *asmcpu = NULL; // TODO
+	if (egg->remit && egg->remit->fini) {
+		egg->remit->fini(egg->remit);
+	}
 	egg->remit = NULL;
 
 	egg->os = os ? rz_str_djb2_hash(os) : RZ_EGG_OS_DEFAULT;
@@ -180,6 +185,9 @@ RZ_API bool rz_egg_setup(RzEgg *egg, const char *arch, int bits, int endian, con
 		egg->endian = endian;
 	} else {
 		return false;
+	}
+	if (egg->remit && egg->remit->init) {
+		egg->remit->init(egg);
 	}
 	return true;
 }
--- a/librz/egg/egg.c
+++ b/librz/egg/egg.c
@@ -61,6 +61,7 @@ RZ_API RzEgg *rz_egg_new(void) {
 		goto beach;
 	}
 	egg->remit = &emit_x86;
+	egg->remit->init(egg);
 	egg->syscall = rz_syscall_new();
 	if (!egg->syscall) {
 		goto beach;
@@ -112,6 +113,7 @@ RZ_API void rz_egg_free(RzEgg *egg) {
 	if (!egg) {
 		return;
 	}
+	egg->remit->fini(egg);
 	rz_path_free(egg->sys_path);
 	rz_buf_free(egg->src);
 	rz_buf_free(egg->buf);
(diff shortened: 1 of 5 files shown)


librz/egg/egg.c
```diff
@@ -61,6 +61,9 @@ RZ_API RzEgg *rz_egg_new(void) {
 		goto beach;
 	}
 	egg->remit = &emit_x86;
+	if (egg->remit->init) {
+		egg->remit->init(egg);
+	}
```
wargio, line 66: This should be also done in `rz_egg_setup`

librz/include/rz_egg.h
```diff
@@ -164,6 +165,7 @@ typedef struct rz_egg_emit_t {
 	void (*branch)(RzEgg *egg, char *b, char *g, char *e, char *n, int sz, const char *dst);
 	void (*mathop)(RzEgg *egg, int ch, int sz, int type, const char *eq, const char *p);
 	void (*get_while_end)(RzEgg *egg, char *out, const char *ctxpush, const char *label);
+	void *emit_data; ///< Holds the data/args of each emit type architecture.
```
wargio:
```suggestion
	void *emit_context; ///< May holds the context data of each emit type architecture.
```

librz/egg/emit_arm.c
```diff
@@ -17,11 +17,29 @@
...
 
 static void emit_init(RzEgg *egg) {
 	/* TODO */
+	egg->remit->emit_data = create_new_emit_arm();
+}
+
+static void emit_fini(RzEgg *egg) {
+	if (!egg->remit->emit_data) {
+		return;
+	}
```
wargio: just free, no need to do this check here. `free()` handles also `NULL`

wargio (changes requested):
^

librz/egg/egg.c
wargio: also call fini in `rz_egg_setup` to ensure when we change arch, the context is also freed.

ahmed-kamal2004: yes, I handled it wrongly bec of misunderstanding of 
`extern RzEggEmit emit_x86;` meaning, but done.

ahmed-kamal2004: well, I think we need this too, what if `rz_egg_new` is called, but `rz_egg_setup` is not called after ?
@wargio WDYT ?

wargio: it's ok in `rz_egg_new`. just add it in `rz_egg_setup`.

ahmed-kamal2004: well, I did that already, any idea why it failling ?

librz/egg/emit_arm.c
```diff
@@ -17,11 +17,29 @@
...
+
+static EmitArmArg *create_new_emit_arm() {
+	EmitArmArg *emit_arm = RZ_NEW0(EmitArmArg);
+	emit_arm->lastarg = 0;
+	return emit_arm;
+}
 
 static void emit_init(RzEgg *egg) {
 	/* TODO */
+	egg->remit->emit_context = create_new_emit_arm();
```
wargio:
```suggestion
static void emit_init(RzEgg *egg) {
	/* TODO */
	egg->remit->emit_context = RZ_NEW0(EmitArmArg);
```

wargio: `RZ_NEW0` already zeros the `EmitArmArg` struct

librz/egg/emit_x86.c
```diff
@@ -521,6 +521,7 @@ RzEggEmit EMIT_NAME = {
 	.arch = RZ_ARCH,
 	.size = RZ_SZ,
 	.init = emit_init,
+	.fini = NULL,
```
wargio, line 525:
```suggestion
```

wargio: this is not needed.

librz/egg/emit_arm.c
wargio: also remove the `/* TODO */`

librz/egg/egg.c
```diff
@@ -175,6 +186,9 @@ RZ_API bool rz_egg_setup(RzEgg *egg, const char *arch, int bits, int endian, con
 	} else {
 		return false;
 	}
+	if (egg->remit && egg->remit->begin) {
+		egg->remit->begin(egg);
+	}
```
wargio: I think you misunderstood me.
We use `init` and `fini` as constructors and destructor functions. but here `init` is used to start the creation of the assembly file. i wanted you to convert the original `init` function into `begin` so you would free up the `init` method to be the real constructor function which we did not have before.

wargio (changes requested):
.

wargio (approved):

PR #5217, opened 2025-06-20
Remove relocation of offsets which were relative.

### LINKED ISSUE #5215: Undefined Behaviour when running tests
When working on development of this PR [5212](https://github.com/rizinorg/rizin/pull/5212)

I faced a strange behaviour when running `meson test -C build`, where it sometimes fail without a clear reason
after talking with @wargio , we think it is something like uninitialized variable.

Image

At the moment, everything seems to be working fine. we can probably close it. but, I think @wargio have a clearer idea about the underlying cause if it's worth keeping track of.

Image

Rot127:

**Detailed description**

Ok, so this is a weird one. The PR fixes the UB. But it looks like there was a bigger problem.

The variables have an offset parameter which is relative to
the owning function's entry point.

Because it is relative to the entry point, it is not affected
if the function is relocated.
Since the offset relative to the entry point doesn't change.

This removes the unnecessary and incorrect relocation efforts
on the offset value.

Additionally, undefined behavior was fixed.
Whenever the offset calculation happened the addresses were
cast to st64. This underflows if we have large enough addresses (something like `0x8000000000000030 - 0x8000000000000000`). 

Simply using ut64 is good enough, because we are only interested in
the difference, not the sign.

@thestr4ng3r Please take a look as well and check my assumptions. You seem to have introduced these changes here and maybe I jsut misunderstood something completely: https://github.com/rizinorg/rizin/commit/4560120be25da483047aac7ecb36f8ae4581144f

**Test plan**

Changed and all green.

**Closing issues**

closes https://github.com/rizinorg/rizin/issues/5215

### DIFF
--- a/librz/arch/function.c
+++ b/librz/arch/function.c
@@ -232,7 +232,7 @@ RZ_API bool rz_analysis_function_relocate(RzAnalysisFunction *fcn, ut64 addr) {
 	ht_up_delete(fcn->analysis->ht_addr_fun, fcn->addr);
 
 	// relocate the var accesses (their addrs are relative to the function addr)
-	st64 delta = (st64)addr - (st64)fcn->addr;
+	st64 delta = addr - fcn->addr;
 	void **it;
 	rz_pvector_foreach (&fcn->vars, it) {
 		RzAnalysisVar *var = *it;
--- a/librz/arch/var.c
+++ b/librz/arch/var.c
@@ -802,7 +802,7 @@ RZ_API int rz_analysis_var_get_argnum(RzAnalysisVar *var) {
 
 RZ_API RZ_BORROW RzPVector /*<RzAnalysisVar *>*/ *rz_analysis_function_get_vars_used_at(RzAnalysisFunction *fcn, ut64 op_addr) {
 	rz_return_val_if_fail(fcn, NULL);
-	return ht_up_find(fcn->inst_vars, (st64)op_addr - (st64)fcn->addr, NULL);
+	return ht_up_find(fcn->inst_vars, op_addr - fcn->addr, NULL);
 }
 
 RZ_DEPRECATE RZ_API RzAnalysisVar *rz_analysis_get_used_function_var(RzAnalysis *analysis, ut64 addr) {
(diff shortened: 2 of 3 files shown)


wargio (commented):
@thestr4ng3r can you give a look since you wrote this?

notxvilka (approved):
Sounds reasonable, also it doesn't break anything, which is a good sign.

wargio (approved):

thestr4ng3r (changes requested):
Ok this is quite old considering the original changes still uses `r_` namings.
According to the short doc comment at https://github.com/rizinorg/rizin/blob/362bac16e2178fb1636dfeb31dc9e27b5c5a1acd/librz/include/rz_analysis.h#L1556-L1558 and the implementation of `rz_analysis_function_relocate` not changing any of the basic blocks or other info, its purpose is really only to change the entrypoint, but leave all other info as-is.
Since the variable access addresses are relative to that entrypoint, but their absolute addresses should not change by the relocation, this means they **do** have to be relocated, so this code must not be removed. The fact that the unit tests had to be changed also points to this.

So I suggest the correct fix here would be to just use `ut64` instead of `st64` to fix the UB and improve the documentation of `rz_analysis_function_relocate` to explicitly state that it changes the entrypoint, but none of the contents of the function.

Rot127:
> So I suggest the correct fix here would be to just use ut64 instead of st64 to fix the UB and improve the documentation of rz_analysis_function_relocate to explicitly state that it changes the entrypoint, but none of the contents of the function.

The whole `change_entrypoint` function seems dubious to me then.
"Changing only the entry point" of a function in your description means we create a new function.
Because if we need to change the variable offsets, it implies the instructions of the function no longer match the old one.
But then we would need to remove the old and add a new function.
Otherwise it is inconsistent anyways.

On the other hand, if we change the entry point to an address within the basic blocks of the function, then we don't need to change the variable accesses.

Or, the whole idea was to have something like "rename function" originally?

In any case, we would need to have multiple entry points for functions in the future.

thestr4ng3r:
> Because if we need to change the variable offsets, it implies the instructions of the function no longer match the old one.

No. What is changed here is only an internal implementation detail. The absolute addresses and thus the instructions of variable accesses before and after the "relocation" will stay identical and only the internal representation is adjusted to preserve this.

> On the other hand, if we change the entry point to an address within the basic blocks of the function, then we don't need to change the variable accesses.

Exactly, this is what the "relocation" of the variable accesses does, it makes sure the semantics of the variable accesses stay the same.

> In any case, we would need to have multiple entry points for functions in the future.

This is technically already possible because basic blocks are not owned by a single function, but refcounted and can be part of multiple functions. But while this can theoretically support all relevant cases, it is primarily useful when metadata other than the basic blocks alone (variables, arguments, etc.) do not match between the different "instances" (entry points).
For cases like a single nop before a function and calls to both the nop and directly after the nop a single function with multiple entrypoints will be more convenient.

Rot127:
@thestr4ng3r I reverted the removals. I still find this use variables confusing and would change it. But we have to talk about this soon again anyways, when we switch to RzIL analysis. So probably not worth the effort currently.

thestr4ng3r (approved):

PR #5305, opened 2025-08-20
Enable RzIL VM to halt on exceptions.

Rot127:

**Detailed description**

There was no message to the user that a program threw an exception.
An event is generated, but no message is printed.
Because we often ignore these with RzIL and don't terminate, this would give the false
impression the program is in a valid state.

On a real machine at least the div by zero exception would halt the program.
This enables it by default for div0 exceptions. But allows to disable the behavior.

Additionally, the user can halt on any other exception we have implemented.

**Test plan**

...

**Closing issues**

...


wargio:
how about doing that if on the event code and with a flag to enable/disable the warning log?

notxvilka (approved):

Rot127:
Will add an option to halt on user defined exceptions.

notxvilka (approved):
@wargio @thestr4ng3r take a look please

librz/core/cconfig.c
```diff
@@ -3920,6 +3930,9 @@ RZ_API int rz_core_config_init(RzCore *core) {
 	/* RzIL config */
 	SETB("rzil.step.events.read", false, "enables/disables printing aezse read event");
 	SETB("rzil.step.events.write", true, "enables/disables printing aezse write event");
+	n = NODECB("rzil.step.events.halt_on_exc", "div0,fp_invalid_op", &rzil_halt_on_exec);
+	SETDESC(n, "Enables/disable exceptions the VM should halt if reached.");
+	SETOPTIONS(n, "div0", "fp_div0", "fp_inexact", "fp_underflow", "fp_overflow", "fp_invalid_op", "none", "all", NULL);
```
wargio, line 3935: tbh, i would have just placed a boolean.

librz/il/theory_bitv.c
```diff
@@ -230,7 +230,7 @@ void *rz_il_handler_div(RzILVM *vm, RzILOpBitVector *op, RzILTypePure *type) {
 		if (rz_bv_is_zero_vector(y)) {
 			result = rz_bv_new(y->len);
 			rz_bv_set_all(result, true);
-			rz_il_vm_event_add(vm, rz_il_event_exception_new("division by zero"));
+			rz_il_vm_event_add(vm, rz_il_event_exception_new(RZ_IL_EVENT_EXC_DIV_ZERO, "division by zero"));
```
wargio: i'm not sure if this makes a lot of sense. for me `rz_il_event_exception_new` was supposed to hold any human readable exception, and not something linked to a flag.

librz/core/cconfig.c
Rot127: I disagree. The other floating point exceptions might not halt the machine. Also it depends on the architecture. For example ARM has a flag to just round division by zero to 0 if set (also implemented like this currently).
Giving some flexibility is better here I think. Especially the more exceptions we get.

librz/il/theory_bitv.c
Rot127: In general agree. But they simply missed a common unique identifier. A string is problematic I think. The moment we have a typo somewhere it is another exception.

wargio: if you want to go with enums, then we do not need a string there.
we can convert the enum to string later in the log/event

wargio: just remove the string parameter and create an enum `RZ_IL_EVENT_EXC_*` and a function to convert the enum to `const char*`

notxvilka (approved):

wargio (approved):

thestr4ng3r (approved):
Some of it could perhaps be abstracted into hooks. But it can all be easily refactored so lgtm.

PR #5312, opened 2025-08-23
librz/util/subprocess: fix inability run .com exes without extension

nir9:
…ables

 

**Detailed description**

When piping Rizin command output on Windows, using `more` is convenient, however Rizin only tries appending `.exe` automatically and Windows uses the `.com` extension for the `more` program. Hence the following error is encountered when running for example `? | more`:
```
[0x00000000]> ? | more
ERROR: Cannot start subprocess.
```

The way to solve it currently is to run explicitly `more.com`:
```
[0x00000000]> ? | more.com
Usage: [.][times][cmd][~grep][@[@iter]addr][|>pipe] ; ...
[...]
```

This PR solves this issue by also trying to append the `.com` extension, and not giving up after trying `.exe`.

Also, I did a minor refactor, passing NULL in the last argument of `SearchPathW` instead of the variable, since it was not used in the code, and according to the Win32 docs, it is an optional parameter, ref: https://learn.microsoft.com/en-us/windows/win32/api/processenv/nf-processenv-searchpathw

**Test plan**

Explained above

**Closing issues**

None

### DIFF
--- a/librz/util/subprocess.c
+++ b/librz/util/subprocess.c
@@ -262,7 +262,6 @@ RZ_API RZ_OWN RzSubprocess *rz_subprocess_start_opt(RZ_NONNULL const RzSubproces
 	HANDLE stdin_read = curr_stdin_handle;
 	HANDLE stdout_write = curr_stdout_handle;
 	HANDLE stderr_write = curr_stderr_handle;
-	LPWSTR lpFilePart;
 	PWCHAR cmd_exe = RZ_NEWS0(WCHAR, MAX_PATH);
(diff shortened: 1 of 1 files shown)


nir9:
Are tests required for this change?

notxvilka (approved):
LGTM but you can also use embedded `less` - `~..` and `~..`

wargio:
i think this is ok as is.

wargio (approved):

nir9:
@notxvilka cool! didn't know about the embedded less feature

PR #5313, opened 2025-08-23
Add: Mark API

PremadeS:

**Detailed description**

* This PR adds `bookmark` support, which are just `flags` that cover a range instead of single address
    - Currently bookmarks are only shown in `Hexdump` in the `comment` section *(excluding the rizin shell commands of course)* where `[s]` is used to indicate the start of bookmark followed by its `name` and `[e]` is used to specify the end, again followed the bookmarks `name`
```bash
[0x00000000]> m marked 0x41
[0x00000000]> pxc
- offset -   0 1  2 3  4 5  6 7  8 9  A B  C D  E F  0123456789ABCDEF  comment
0x00000000  ffff ffff ffff ffff ffff ffff ffff ffff  ................  ; [s] marked (0x0-0x41)
0x00000010  ffff ffff ffff ffff ffff ffff ffff ffff  ................
0x00000020  ffff ffff ffff ffff ffff ffff ffff ffff  ................
0x00000030  ffff ffff ffff ffff ffff ffff ffff ffff  ................
0x00000040  ffff ffff ffff ffff ffff ffff ffff ffff  ................  ; [e] marked (0x0-0x41)
0x00000050  ffff ffff ffff ffff ffff ffff ffff ffff  ................
0x00000060  ffff ffff ffff ffff ffff ffff ffff ffff  ................
0x00000070  ffff ffff ffff ffff ffff ffff ffff ffff  ................
0x00000080  ffff ffff ffff ffff ffff ffff ffff ffff  ................
0x00000090  ffff ffff ffff ffff ffff ffff ffff ffff  ................
0x000000a0  ffff ffff ffff ffff ffff ffff ffff ffff  ................
0x000000b0  ffff ffff ffff ffff ffff ffff ffff ffff  ................
0x000000c0  ffff ffff ffff ffff ffff ffff ffff ffff  ................
0x000000d0  ffff ffff ffff ffff ffff ffff ffff ffff  ................
0x000000e0  ffff ffff ffff ffff ffff ffff ffff ffff  ................
0x000000f0  ffff ffff ffff ffff ffff ffff ffff ffff  ................
``` 

[Hexdump Bookmark Highlighting PR for Cutter](https://github.com/rizinorg/cutter/pull/3510) is dependent on **this** PR, meaning this PR needs to be merged before **Hexdump highlighting** for **Cutter** can be implemented
...

**Test plan**

`m?` to see the list of possible commands. `unit` and `integration` tests are also provided 
...

**Closing issues**

...


wargio:
i like the work you have done, but i'm quite sure flags has a size. wouldn't that be enough for the bookmark feature?

wargio:
@notxvilka

notxvilka (commented):
I think having a separate bookmark API is great, but underneath it could use flags, indeed. No strong opinion about that because of https://github.com/rizinorg/rizin/pull/4418 also

PremadeS:
I get the idea but wouldn't that make the API difficult to maintain in the long run? 
More features might be added to `bookmarks` in the future which would make the API use existing `flags` logic for some features and new implementations for others @wargio 
Also #4418 (as @notxvilka mentioned)

wargio:
then maybe this for now is ok, but it would have been nice to have this unified with flags.

librz/include/rz_bookmark.h
```diff
@@ -0,0 +1,83 @@
...
+
+#define RZ_BOOKMARK_NAME_SIZE 512
+
+typedef struct rz_bookmark_item_t {
+	ut64 from; /* starting address of bookmark */
+	ut64 to; /* ending address of bookmark */
+	char *name; /* unique name for each bookmark, escaped to avoid issues with rizin shell */
+	char *realname; /* real name, without any escaping */
+	char *comment; /* item comment */
+	char *color; /* item color */
```
Rot127: Replace all the duplicate flags with an `RzFlagItem`.
The `RzFlagItem` should be in the flag space `bookmarks`/`notes`.

librz/include/rz_bookmark.h
```diff
@@ -0,0 +1,83 @@
...
+
+typedef struct rz_bookmarks_at_offset_t {
+	ut64 off;
+	RzList /*<RzBookmarkItem *>*/ *bookmarks;
+} RzBookmarksAtOffset;
+
+typedef struct rz_bookmark_t {
+	HtSP *ht_name; // name -> RzBookmarkItem*
+	RzList /*<RzBookmarkItem *>*/ *items; // All bookmarks
+	RzSkipList *by_off; // offset -> RzBookmarksAtOffset*
```
Rot127:
```suggestion
	HtSP *ht_name; ///< name -> RzBookmarkItem*
	RzList /*<RzBookmarkItem *>*/ *items; ///< All bookmarks
	RzSkipList *by_off; ///< offset -> RzBookmarksAtOffset*
```

Rot127 (changes requested):
About the more general design discussion.

`RzBookmarkItem` really seems duplicate most of `RzFlagItem` members.
Except the range property which is defined in `RzFlagZoneItem`.

I think we should make the task of flags and "bookmarks" more distinct. Otherwise we might end up with two systems doing slightly similar things and we get confused which one to use for what.

For that we should defined the difference between flags and "bookmarks" properly.

I would suggest:

**Flags**
- Add a small amount of context to a single address. Like a name/identifier of a function or an object.
- Cannot refer to a range of addresses.

**Bookmarks**

- Adds detailed information and more context to an address range or **multiple ranges**!

I would suggest to rename "bookmark" to "note". Bookmark has too many connotations in common with flag IMHO. And hence we get the confusion. Because they are a similar concept. A note on the other hand can be distinct, and we don't need to have `note` depend on `flags`.

Additionally, `RzFlagZoneItem` should be removed here and in `rz_flags.h` we need a `\file` comment at the top to explain that `flags` only annotate single addresses. And refer to `notes` if more information needs to be attached.

librz/include/rz_bookmark.h
Rot127: Depending on our decision above I mean.
Sorry, didn't add this

PremadeS: > Depending on our decision above I mean. Sorry, didn't add this

Thanks for the detailed suggestion above, I'm guessing I should wait before the decision is final right?

notxvilka:
> I would suggest to rename "bookmark" to "note". Bookmark has too many connotations in common with flag IMHO. And hence we get the confusion. Because they are a similar concept. A note on the other hand can be distinct, and we don't need to have note depend on flags.

Not sure about that, "note" is closer to the "comment" for which we already have `CC` commands. Bookmark (or just mark) is more appropriate, I think.

> Additionally, RzFlagZoneItem should be removed here and in rz_flags.h we need a \file comment at the top to explain that flags only annotate single addresses. And refer to notes if more information needs to be attached.

This suggestion makes sense, but the flag zone could have range by definition, so it doesn't overlap with notes/marks.

Rot127:
> Bookmark (or just mark) is more appropriate, I think.

I would vote for `mark` then. In my head bookmark and flag refer largely to the same concept. Giving a name to a single place.
And I honestly think the issue is mostly our definition of the feature. Otherwise I like the addition a lot.

PremadeS:
Summary of the changes:
* renamed to `Mark`
* every new `Mark` now adds a `flag` at the starting offset
* showing "brackets" for overlapping `Marks` is not as simple as I initially expected. Hence why `Marks` are still shown as comments in hexdump for now (see the hexdump output in PR description)

notxvilka (changes requested):
Since you also added a new command, please add test for it too, in `test/db/cmd/cmd_mark`

librz/core/cmark.c
```diff
@@ -0,0 +1,135 @@
...
+	PJ *pj;
+	RzTable *tbl;
+	bool in_range;
+	ut64 range_from;
+	ut64 range_to;
+};
+
+static bool print_mark_name(RzMarkItem *bm, void *user) {
+	struct print_mark_t *u = (struct print_mark_t *)user;
+	if (u->in_range && (bm->to < u->range_from || bm->from > u->range_to)) {
```
Rot127: Please use RZ_BETWEEN macro

librz/core/cmd/cmd_mark.c
```diff
@@ -0,0 +1,331 @@
...
+static bool mark_to_mark_foreach(RzMarkItem *bi, void *user) {
+	struct mark_to_mark_t *u = (struct mark_to_mark_t *)user;
+	if (bi->from < u->next && bi->from > u->offset) {
+		u->next = bi->from;
+	}
+	return true;
+}
+
+static int mark_to_mark(RzCore *core, const char *glob) {
+	rz_return_val_if_fail(glob, 0);
```
Rot127, line 44: If you check for glob, also check for core please.

librz/core/cmd/cmd_mark.c
```diff
@@ -0,0 +1,331 @@
...
+	if (prev_space) {
+		rz_flag_space_set(f, prev_space->name);
+	}
+}
+
+RZ_IPI RzCmdStatus rz_mark_add_handler(RzCore *core, int argc, const char **argv) {
+	RzMarkItem *item;
+	bool addMark = true;
+	ut64 end = rz_num_math(core->num, argv[2]);
+	if ((item = rz_mark_get_at(core->marks, core->offset))) {
```
Rot127, line 76: This is fine for now. But please open an issue about this that there cannot be two marks with the same starting offset.

librz/core/cmd_descs/cmd_mark.yaml
```diff
@@ -0,0 +1,150 @@
...
+# SPDX-License-Identifier: LGPL-3.0-only
+---
+name: cmd_mark
+commands:
+  - name: m
+    cname: mark_add
+    summary: Add the mark if there are no existing marks
+    description: >
+      Adds the mark to the current offset only if
+      no mark exists at this range already
```
Rot127: at this offset already...

librz/core/cmd_descs/cmd_mark.yaml
```diff
@@ -0,0 +1,150 @@
...
+      - name: end
+        type: RZ_CMD_ARG_TYPE_NUM
+      - name: comment
+        type: RZ_CMD_ARG_TYPE_STRING
+        optional: true
+  - name: m+
+    cname: mark_append
+    summary: Add mark
+    description: >
+      Adds the mark to the current offset
```
Rot127: Ignoring/overwriting any existing ones, right?

librz/core/cmd_descs/cmd_mark.yaml
```diff
@@ -0,0 +1,150 @@
...
+      - name: end
+        type: RZ_CMD_ARG_TYPE_NUM
+      - name: comment
+        type: RZ_CMD_ARG_TYPE_STRING
+        optional: true
+  - name: m-
+    cname: mark_remove
+    summary: Remove mark
+    description: >
+      If the glob is supplied it removes just mark items matching the pattern.
```
Rot127: glob matches the name? Not the comment?
Please specify here.
Also consider to rename glob to regex_pattern since I assume it can be any extended regex?

librz/core/cmd_descs/cmd_mark.yaml
```diff
@@ -0,0 +1,150 @@
...
+    summary: Set a color for the given mark / Show the color for the given mark
+    args:
+      - name: name
+        type: RZ_CMD_ARG_TYPE_FLAG
+      - name: color
+        type: RZ_CMD_ARG_TYPE_STRING
+        optional: true
+  - name: mC
+    cname: mark_comment
+    summary: Set a comment for the given mark / Show the comment for the given mark
```
Rot127:
```suggestion
    summary: Set a comment for the given mark. If no comment is given, it shows the current one for the mark.
```

librz/core/cmd_descs/cmd_mark.yaml
```diff
@@ -0,0 +1,150 @@
...
+    cname: mark_rename
+    summary: Rename mark
+    args:
+      - name: old
+        type: RZ_CMD_ARG_TYPE_FLAG
+      - name: new
+        type: RZ_CMD_ARG_TYPE_STRING
+  - name: mN
+    cname: mark_realname
+    summary: Show the realname of the mark / Set the realname of the mark
```
Rot127: Explain what is the real name vs. the normal name.

librz/core/cmd_descs/cmd_mark.yaml
```diff
@@ -0,0 +1,150 @@
...
+    args:
+      - name: name
+        type: RZ_CMD_ARG_TYPE_FLAG
+      - name: realname
+        type: RZ_CMD_ARG_TYPE_STRING
+        optional: true 
+  - name: mm
+    cname: mark_move
+    summary: Move a mark to new location
```
Rot127:
```suggestion
    summary: Move a mark to a new location.
```

librz/core/cmd_descs/cmd_mark.yaml
```diff
@@ -0,0 +1,150 @@
...
+    args:
+      - name: name
+        type: RZ_CMD_ARG_TYPE_FLAG
+      - name: new_start
+        type: RZ_CMD_ARG_TYPE_RZNUM
+      - name: new_end
+        type: RZ_CMD_ARG_TYPE_RZNUM
+  - name: mf
+    cname: mark_distance
+    summary: Distance in bytes to reach the next mark
```
Rot127: Explain in more detail what it means.
From where to where is the distance measured?
Start, end?
What happens with overlaps?
Also consider renaming glob to name_regex_pattern or similar again.

librz/core/cmd_descs/cmd_mark.yaml
```diff
@@ -0,0 +1,150 @@
...
+    args:
+      - name: glob
+        type: RZ_CMD_ARG_TYPE_STRING
+        optional: true
+  - name: md
+    summary: Describe mark
+    subcommands:
+      - name: md
+        cname: mark_describe
+        summary: Describe mark containing current offset or using name + delta from current offset
```
Rot127: Please describe in more detail what you mean.
Does it show the comment? Or name + range of mark?
Does it show all marks overlapping with the offset?

You can also consider using examples.
Check out the search commands how they add examples to the commands.

librz/core/cmd_descs/cmd_mark.yaml
```diff
@@ -0,0 +1,150 @@
...
+          - RZ_OUTPUT_MODE_STANDARD
+          - RZ_OUTPUT_MODE_TABLE
+          - RZ_OUTPUT_MODE_JSON
+        args:
+          - name: name
+            type: RZ_CMD_ARG_TYPE_FLAG
+            optional: true
+  - name: mi
+    cname: mark_range
+    summary: Show marks in the block or custom range
```
Rot127:
```suggestion
    summary: Show marks in the current block or in the next <size> bytes.
```

librz/core/core.c
```diff
@@ -1,6 +1,8 @@
 // SPDX-FileCopyrightText: 2009-2020 pancake <pancake@nopcode.org>
 // SPDX-License-Identifier: LGPL-3.0-only
 
+#include "rz_mark.h"
+#include "rz_util/rz_str.h"
```
Rot127:
```suggestion
#include <rz_util/rz_str.h>
```

Also in the other files.

librz/core/core.c
```diff
@@ -955,10 +957,39 @@ RZ_API int rz_core_fgets(char *buf, int len, void *user) {
...
+static void append_marks(RzStrBuf *sb, RzList /*<RzMarkItem *>*/ *list, ut64 addr, bool is_start) {
+	RzListIter *it;
+	RzMarkItem *bi;
+	bool first = (rz_strbuf_length(sb) == 0);
+
+	rz_list_foreach (list, it, bi) {
+		if ((is_start && addr == bi->from) || (!is_start && addr == bi->to)) {
+			if (!first)
+				rz_strbuf_append(sb, "  ; ");
+			rz_strbuf_appendf(sb, is_start ? "[s] %s (0x%" PFMT64x "-0x%" PFMT64x ")" : "[e] %s (0x%" PFMT64x "-0x%" PFMT64x ")",
```
Rot127:
```suggestion
			rz_strbuf_appendf(sb, is_start ? "[s] %s [0x%" PFMT64x "-0x%" PFMT64x "]" : "[e] %s [0x%" PFMT64x "-0x%" PFMT64x "]",
```
Or is the range exclusive?

librz/include/rz_mark.h
```diff
@@ -0,0 +1,87 @@
...
+// NOTE: Marks are currently used for the Hex Widget only
+// but the API is kept generic in case of future reuse
+
+RZ_LIB_VERSION_HEADER(rz_mark);
+
+#define RZ_MARK_NAME_SIZE 512
+
+typedef struct rz_mark_item_t {
+	ut64 from; /* starting address of mark */
+	ut64 to; /* ending address of mark */
```
Rot127: Inclusive or exclusive?

librz/mark/mark.c
```diff
@@ -0,0 +1,479 @@
...
+		return true;
+	}
+	return false;
+}
+
+static bool update_mark_item_name(RzMark *b, RzMarkItem *item, const char *newname, bool force) {
+	if (!b || !item || !newname) {
+		return false;
+	}
+	if (!force && (item->name == newname || (item->name && !strcmp(item->name, newname)))) {
```
Rot127:
```suggestion
	if (!force && (item->name == newname || (RZ_STR_EQ(item->name, newname)))) {
```

librz/mark/mark.c
```diff
@@ -0,0 +1,479 @@
...
+		return true;
+	}
+	free(fname);
+	return false;
+}
+
+static bool mark_item_matches(RzMarkItem *bi, ut64 off, RzMarkMatchMode mode) {
+	switch (mode) {
+	case RZ_MARK_MATCH_CONTAINS:
+		return (off >= bi->from && off <= bi->to);
```
Rot127: RZ_BETWEEN

librz/mark/mark.c
```diff
@@ -0,0 +1,479 @@
...
+
+	RzList *all = rz_mark_all_list(b);
+	if (!all) {
+		return false;
+	}
+
+	RzListIter *it;
+	RzMarkItem *bi;
+	rz_list_foreach (all, it, bi) {
+		if ((bi->from >= from && bi->from <= to) ||
```
Rot127: RZ_BETWEEN

librz/mark/mark.c
```diff
@@ -0,0 +1,479 @@
...
+	RzList *ret = rz_list_new();
+	if (!b || !ret) {
+		return ret;
+	}
+
+	rz_mark_foreach(b, append_bi, ret);
+	return ret;
+}
+
+/* checks whether any mark starts or ends in the given range.
```
Rot127:
```suggestion
/*
 * \brief Checks whether any mark starts or ends in the given range.
```

librz/mark/mark.c
```diff
@@ -0,0 +1,479 @@
...
+ * false is returned if such a item is not found. */
```
Rot127:
```suggestion
 * \return false is returned if such a item is not found. True if it is found. */
```

librz/mark/mark.c
```diff
@@ -0,0 +1,479 @@
...
+
+	update_mark_item_range(b, item, from, to, is_new, true);
+	update_mark_item_name(b, item, name, true);
+	return item;
+err:
+	rz_mark_item_free(item);
+	return NULL;
+}
+
+RZ_API RzMark *rz_mark_free(RzMark *b) {
```
Rot127:
```suggestion
RZ_API void rz_mark_free(RZ_OWN RzMark *b) {
```

Rot127 (changes requested):
German train internet prevents me to do more. But it looks good!

Please fix the API functions doxygen by using `\brief`, `\return` and documenting the parameters as well.
Also add the type annotations RZ_BORROW, RZ_OWN.
Please also use proper grammar/punctuation etc. Good documentation makes it comfortable and this will lead to more usage.

librz/include/rz_mark.h
```diff
@@ -48,35 +48,35 @@ typedef enum {
...
-RZ_API int rz_mark_count(RzMark *b, const char *glob);
-RZ_API void rz_mark_foreach(RzMark *b, RzMarkItemCb cb, void *user);
-RZ_API void rz_mark_foreach_glob(RzMark *b, const char *glob, RzMarkItemCb cb, void *user);
+RZ_API RZ_OWN RzMark *rz_mark_new(void);
+RZ_API void rz_mark_free(RZ_OWN RzMark *b);
+RZ_API RZ_OWN RzMarkItem *rz_mark_set(RZ_BORROW RzMark *b, RZ_BORROW const char *name, ut64 from, ut64 to);
+RZ_API RZ_BORROW RzMarkItem *rz_mark_get(RZ_BORROW RzMark *b, RZ_BORROW const char *name);
+RZ_API RZ_BORROW RzMarkItem *rz_mark_get_start(RZ_BORROW RzMark *b, ut64 off);
+RZ_API RZ_BORROW RzMarkItem *rz_mark_get_end(RZ_BORROW RzMark *b, ut64 off);
+RZ_API RZ_BORROW RzMarkItem *rz_mark_get_at(RZ_BORROW RzMark *b, ut64 off);
```
wargio: more than borrow in the args, i would have placed the NONNULL/NULLABLE ones, because i would expect args to be always borrow unless RZ_OWN is specified

wargio (changes requested):
Seems you forgot to update the `librz/include/meson.build` to install `rz_mark.h`

librz/core/cmd_descs/cmd_mark.yaml
```diff
@@ -0,0 +1,168 @@
+# SPDX-FileCopyrightText: 2022 RizinOrg <info@rizin.re>
```
notxvilka: 2025

notxvilka (approved):
Three quarks for Muster Mark!

librz/core/cmd/cmd_mark.c
PremadeS: Please elaborate.
There can be two or more marks with the same starting offset. This can be done using the "append" command `m+`
However the "add" command `m` does not allow the user to add another mark at the current offset if any of the previously defined marks "contain" the current offset in their range, meaning a mark using `m` can only be added if the current offset is not covered by any of the other marks (similar to flags)

wargio (approved):

librz/core/cmd/cmd_mark.c
Rot127: Never mind. With the new documentation of the commands it is clear now.

librz/include/rz_mark.h
```diff
@@ -0,0 +1,86 @@
...
+
+#define RZ_MARK_NAME_SIZE 512
+
+typedef struct rz_mark_item_t {
+	ut64 from; /* inclusive starting address of mark */
+	ut64 to; /* inclusive ending address of mark */
+	char *name; /* unique name for each mark, escaped to avoid issues with rizin shell */
+	char *realname; /* real name, without any escaping */
+	char *comment; /* item comment */
+	char *color; /* item color */
```
Rot127:
```suggestion
	ut64 from; ///< inclusive starting address of mark
	ut64 to; ///< inclusive ending address of mark
	char *name; ///< unique name for each mark, escaped to avoid issues with rizin shell
	char *realname; ///< real name, without any escaping
	char *comment; ///< item comment
	char *color; ///< item color
```

librz/mark/mark.c
```diff
@@ -0,0 +1,601 @@
...
+		RzListIter *it;
+		RzMarkItem *bi;
+
+		rz_list_foreach (cur->marks, it, bi) {
+			if (mark_item_matches(bi, off, mode)) {
+				rz_list_append(out, bi);
+				if (stop_on_first) {
+					return;
+				}
+			}
```
Rot127:
```suggestion
			if (!mark_item_matches(bi, off, mode)) {
				continue;
			}
			rz_list_append(out, bi);
			if (stop_on_first) {
				return;
			}

```

librz/mark/serialize_mark.c
```diff
@@ -0,0 +1,183 @@
...
+/**
+ * \brief Save marks to a database.
+ *
+ * Serializes all marks from the given container \p bm into the namespace
+ * "marks" within the database \p db.
+ *
+ * \param db Database to save into.
+ * \param bm Mark container to serialize.
+ */
+RZ_API void rz_serialize_mark_save(RZ_NONNULL Sdb *db, RZ_NONNULL RzMark *bm) {
```
Rot127, line 49: Missing `rz_return_if_fail(db && bm)` because it is an API function.

librz/mark/serialize_mark.c
```diff
@@ -0,0 +1,183 @@
...
+ * \param db Database to load from.
+ * \param bm Mark container to populate.
+ * \param res Optional result info structure for reporting errors, may be NULL.
+ * \return True if loading succeeded. False if an error occurred.
+ *
+ * \note If no "marks" namespace exists, this is treated as an empty set
+ *       of marks (not an error).
+ */
+RZ_API bool rz_serialize_mark_load(RZ_NONNULL Sdb *db, RZ_NONNULL RzMark *bm, RZ_NULLABLE RzSerializeResultInfo *res) {
+	rz_mark_unset_all(bm);
```
Rot127: Missing NULL checks.

Rot127 (changes requested):
Excellent job! I wanted to have something like this for assembly annotation since the very beginning I started out with Rizin.

Should be easy to add now, thanks to you!

Please just address these small nitpicks. Then we are ready to merge IMO.

Rot127 (approved):

PR #5319, opened 2025-08-26
librz/main/rz-ax: fix bug of stdin mode unworking

nir9:

**Detailed description**

This PR fixes a bug in which when launching `rz-ax` with no arguments returns immediately while the expected behavior is that it enters CLI mode as described in the book, ref: https://book.rizin.re/src/introduction/overview.html#rz-ax

The reason this happens is that 0 is passed for the flags pointer which causes the function responsible to return immediately to avoid dereferencing zero. This seems to be a mistake, while the intention was to pass a pointer to a `flags = 0` variable.

**Test plan**

Run `rz-ax` with no args.

**Closing issues**

None

### DIFF
--- a/librz/main/rz-ax.c
+++ b/librz/main/rz-ax.c
@@ -747,10 +747,10 @@ dotherax:
 RZ_API int rz_main_rz_ax(int argc, const char **argv) {
 	int i, fm = 0;
 	RzNum *num = rz_num_new(NULL, NULL, NULL);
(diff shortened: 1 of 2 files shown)


wargio (approved):

wargio (changes requested):
Add a test under test/db/tools

wargio (approved):

PR #5355, opened 2025-09-10
Add corefile support for linux mips32/64

wargio:

**Detailed description**

Tested against gdb


librz/bin/format/elf/elf_notes.c
```diff
@@ -104,9 +121,15 @@ static RzBinElfPrStatusLayout prstatus_layouts[ARCH_LEN] = {
 	[SPARC_V9] = { SPARC64_REGS_SIZE, SPARC64_PR_STATUS_REG_OFFSET, 64, SPARC64_PR_STATUS_REG_OFFSET_SP },
 	[OPENBSD_SPARC_V9] = { SPARC64_REGS_SIZE, SPARC64_OPENBSD_REG_OFFSET, 64, SPARC64_PR_STATUS_REG_OFFSET_SP },
 
+	[MIPS_32] = { MIPS32_REGS_SIZE, MIPS_GPR32_STATUS_OFFSET, 0, 0 },
```
Rot127, line 124: SP offset missing?
It is used actually, so skipping it might not be good.

Rot127 (changes requested):
Lgtm, but what about the SP?

librz/bin/format/elf/elf_notes.c
wargio: i cannot get a good idea from the GDB sources nor LLVM, so i don't actually know.

Rot127: Ah I see. The SP offset is from the base of the register values.
So in this case the registers `zero, v0, v1, a0-a3..., k0, k1, gp` come before (32bit MIPS I mean). I think it is not defined outside of Rizin.
It is just used in `Elf_(rz_bin_elf_get_sp_val)`.

notxvilka (approved):

librz/bin/format/elf/elf_notes.c
notxvilka: Step by step. It's better to figure out by playing with Rizin and GDB dynamically on the MIPS machine(s).

PR #5372, opened 2025-09-12
Add several memory helpers (required for speed up of string search).

Rot127:

**Detailed description**

Adds a few more memory and bit related helpers.
I'll use them for the speed up of the string search of unaligned memory and foreign endianness encodings.

Functions added:

- Count trailing zeros
- Check pointer alignment
- Memcpy with offset into src.
- Swap byte blocks of 2 and 4 bytes.

**Test plan**

Added

**Closing issues**

...


Rot127:
Apparently do most compilers produce code with unaligned data.
Will fix it tomorrow.

librz/include/rz_util/rz_bits.h
```diff
@@ -33,6 +33,45 @@ DEFINE_COUNT_ONES(ut32);
 DEFINE_COUNT_ONES(ut16);
 DEFINE_COUNT_ONES(ut8);
 
+static inline size_t rz_bits_trailing_zeros(ut64 v) {
```
notxvilka, line 44: Add Doxygen

librz/util/mem.c
```diff
@@ -331,3 +331,251 @@ RZ_API void rz_mem_memzero(void *dst, size_t l) {
...
+	}
+	if (!rz_mem_copy(dst, buf_size, buf, buf_size)) {
+		free(dst);
+		return NULL;
+	}
+	return rz_mem_swap_bytes_2_inplace(dst, buf_size);
+}
+
+/**
+ * \brief Swaps the bytes in 2 byte blocks from the given buffer and returns
```
notxvilka, line 417: Are you aware of `rz_swap_*()` functions in the `librz/include/rz_endian.h`?

notxvilka (changes requested):

librz/util/mem.c
Rot127: Ah, damn. I actually wasn't. Must have overlooked them. Will use them here.

librz/include/rz_util/rz_bits.h
notxvilka: If I remember correctly, compilers should have `__builtin_ctz()` for that. Please check if could be used instead.

notxvilka: See https://gcc.gnu.org/onlinedocs/gcc/Bit-Operation-Builtins.html
Please detect it via Meson like other builtins

notxvilka (changes requested):

notxvilka (approved):

meson.build
```diff
@@ -439,6 +439,7 @@ foreach it : ccs
   it_userconf.set10('HAVE_ZLIB', get_option('use_zlib'))
   it_userconf.set10('SUPPORTS_PCRE2_JIT', pcre2_jit_supported)
   it_userconf.set('N_THREAD_LIMIT', get_option('n_thread_limit'))
+  it_userconf.set10('HAVE___BUILTIN_CTZLL', cc.has_function('__builtin_ctzll'))
```
notxvilka: Why not put together with BUILTIN_CLZLL?

Rot127: I don't know why. But the `HAVE___BUILTIN_CTZLL` value is simply not handled by Muon. Meson gives a warning as well, but doesn't fail. It just generates
`#define HAVE___BUILTIN_CTZLL ` (without `1` or `0`).

notxvilka (approved):

Rot127:
Removed the merge-when-green` label. Please let us merge the CI fixes first and then I can rebase. Just in case...

notxvilka (changes requested):
It breaks MSVC builds:
```
"link"  /MACHINE:x64 /OUT:librz/util/sdb_gen.exe librz/util/sdb_gen.exe.p/sdb_src_main.c.obj "/release" "/nologo" "/OPT:REF" "librz/util/rz_util.lib" "ws2_32.lib" "wininet.lib" "psapi.lib" "/SUBSYSTEM:CONSOLE" "kernel32.lib" "user32.lib" "gdi32.lib" "winspool.lib" "shell32.lib" "ole32.lib" "oleaut32.lib" "uuid.lib" "comdlg32.lib" "advapi32.lib"
   Creating library librz\util\sdb_gen.lib and object librz\util\sdb_gen.exp
rz_util.lib(mem.c.obj) : error LNK2019: unresolved external symbol __builtin_ctzll referenced in function rz_bits_trailing_zeros
librz\util\sdb_gen.exe : fatal error LNK1120: 1 unresolved externals
[1770/2404] Linking static target librz/hash/rz_hash.lib
[1771/2404] Compiling C object librz/main/rz_main.lib.p/rz-ax.c.obj
```

librz/include/rz_types_base.h
```diff
@@ -27,6 +27,7 @@
 #define st16  short
 #define ut8   unsigned char
 #define st8   signed char
+#define utptr uintptr_t
```
wargio, line 30: not super sure this is standard, because the `_t` suffix is usually a typedef.

wargio: Oh, it's part of in C99

wargio: maybe we might need to patch this by also checking that `uintptr_t` exists otherwise use `size_t`

notxvilka: It is: https://en.cppreference.com/w/c/types/integer.html

notxvilka: No need, it's a strong C99 requirement and our minimally supported version.

Rot127: `uintptr_t` is the only type which is guaranteed to have a conversion defined between pointer <-> integer. `size_t` isn't to my knowledge.
We could just copy the `#ifdef` from the std headers, but I think `uintptr_t` pretty common since it was already in C99. So if this fails, other parts of our code probably fail as well.

wargio: since c99 :dango:

notxvilka: No need. If we ever meet compiler that doesn't know it - we will think about it. Now, everything is at least C99. Not an issue.

Issue #4896, opened 2025-02-12
Implement more sophisticated string encoding detection.


Rot127:
**Is your feature request related to a problem? Please describe.**

String detection should be reliable transparent. If the decoding is not known, it should give the use a numbre possible strings at a given offset. The user can then choose which makes the most sense.
This is not done yet. It simply chooses the encoding most similar to ASCII, which is a not very reliable metric.

**Describe the solution you'd like**

The following is already implemented here: https://github.com/rizinorg/rizin/tree/string-detection

<hr>

**Problems of the previous heuristics**

- UTF-16 and UTF-32 checks only tested for bytes with the upper bytes 0. This effectively excludes all code points > 0xff.
  But those are the ones UTF-16/32 is used in the first place. So the heuristics is off.
- The testing happened mostly sequentially and returned early. If UTF-32 matched before IBM037, it just assumed it is UTF-32, not even performing the IBM037 test. Leaving the string invisible.
- Due to this it is also hard to extend the testing. Because with every new check of a new encoding, the previous checks must somehow fail before. Forbidding the possibility of a byte sequence being valid in two encodings. (And judging from the code, @imbillow ran into exactly this problem).
- There was no easy way to test for multiple encodings at the same offset, except running the scan again after switching encodings.

**New encoding detection**

The new solution is somewhat less performant (I think, but I didn't profile it yet and I don't think it is much worse).
But, I think it is so much more worth it, because:

- We can actually control how much of a performance hit we want.
  - Check only X bytes of the buffer? Not a big deal to implement.
  - Only check for a subset of encodings? Just some `if` conditions.
- At the same time we get way more flexibility how the detection works (via the biases).
  - If we have a binary with big endian, we can give big endian strings a strong bias.
  - We want to limit the string detection to strings of a certain length, just change the bias.
  - We need the detected strings to align with even addresses, again just change the bias.
  - In general, we can give the user a selection of choises. If they see strings of encoding X make the most sense, she can assume this is the string encoding in the file, set it and no longer get any performance penalty.
  - Most generally, we reduce the number of false negatives to 0 (within the range of the detection parameters and biases).

I think this trade is very good. Because we should always be able to provide all statically knowable information to the user.
What they do with it is their choice, but not ours.

**The algorithm**

It is rather straight forward. The detection function gets a buffer it iterates twice over the buffer (`O(2n)`) and
[...]

PR #5378, opened 2025-09-14
Enable string search collection to search for multiple encodings.

Rot127:

**Detailed description**

This moves the encoding and alignment properties out of the collection
object into the RzDetectedStr object.
It essentially removes duplicates of this info.

Because of this move, rz_search_collection_string_add() will now
check for invalid collection configurations.

**Test plan**

Added

**Closing issues**

...

### DIFF
--- a/librz/search/string_search.c
+++ b/librz/search/string_search.c
@@ -13,7 +13,6 @@
 
 typedef struct string_search {
 	RzUtilStrScanOptions options; ///< String scan options
-	RzStrEnc encoding; ///< Expected encoding
 	RzPVector /*<RzDetectedString *>*/ *strings; ///< Strings to search
 } StringSearch;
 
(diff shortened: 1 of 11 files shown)


librz/search/string_search.c
```diff
@@ -13,7 +13,6 @@
 
 typedef struct string_search {
 	RzUtilStrScanOptions options; ///< String scan options
-	RzStrEnc encoding; ///< Expected encoding
```
notxvilka, line 16: We still should have "preferred" encoding, maybe in the RzUtilStrScanOptions themselves, I think. In the cases that some binaries, e.g. Windows PE could have higher probability of having UTF-16 than any other encoding. It should help with the detection speed and accuracy.

Rot127: Yes, this would be nice. But we would need to refactor the string scanning. Currently it is all linear. First encoding which decodes a valid char wins in the order of: utf-32-le, utf-16-le, utf-16-be, utf-32-be, ebcdic, ascii/utf-8.

The specific case of Windows indicates UTF-16 encoding is the idea of "bias" in https://github.com/rizinorg/rizin/issues/4896

notxvilka (approved):

librz/search/string_search.c
```diff
@@ -346,21 +348,26 @@ RZ_API bool rz_search_collection_strings_check_config_improvements(
 		return true;
 	}
 	StringSearch *ss = col->user;
-	if (ss->encoding == RZ_STRING_ENC_GUESS) {
-		if (log_suggestions) {
-			RZ_LOG_WARN("The string encoding for the search is set to \"guess\".\n"
-				    "The search will consume vastly more resources and the guessing is unreliable.\n"
-				    "You can set a specific encoding with 'e str.encoding=<encoding>'.\n");
+	void **it;
+	;
```
notxvilka: What is this?

PR #5392, opened 2025-09-17
Thread safety of regex

Rot127:
…afety.

 

**Detailed description**

Todo

- [ ] Give all regex patterns their own JIT stack. No compromise.

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/librz/core/csearch.c
+++ b/librz/core/csearch.c
@@ -144,11 +144,10 @@ RZ_API RZ_OWN RzList /*<RzSearchHit *>*/ *rz_core_search_bytes(RZ_NONNULL RzCore
 	RzList *boundaries = NULL;
 	RzSearchOpt *search_opts = NULL;
 
-	RzSearchCollection *collection = rz_search_collection_bytes();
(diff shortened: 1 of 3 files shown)


Rot127:
This PCRE2 crash during the refactor of the string search is weird.
The old implementation should have been thread safe. But crashed with UTF-32/16 and a large files.
Even confirmed that the old one is safe with an PCRE2/sjit dev (https://github.com/PCRE2Project/pcre2/issues/794).

I would finish this one nonetheless, just to be on the safe side and be sure we don't have unsafe JIT code running.

Rot127:
I tried to replicate the ASAN errors/segfaults again but failed for UTF-8.
The segfaults seem to only occur if the input buffer was moved before it is passed to match().

Added the broken example here: https://github.com/Rot127/rizin/commit/5ef62eadab4b2752dd9584c458f0b9561319c033

But it generally doesn't segfault for UTF-8 and ASCII matching, which is done for the byte search.

Because it seems unnecessary to refactor it, I close this.

PR #5444, opened 2025-10-09
h8300: enhance compatibility for h8300h/h8300l

b1llow:
h8500: add h8500h/l support 

**Detailed description**

...

**Test plan**

...

**Closing issues**

...


librz/arch/isa/h8300/h8300_disas.c
```diff
@@ -1399,7 +1417,20 @@ static int h8300_decode_2(const ut8 *instr, struct h8300_cmd *cmd) {
 	return ret;
 }
 
-int h8300_decode_command(const ut8 *instr, ut64 len, struct h8300_cmd *cmd, ut64 pc) {
+H8300CpuType h8300_cpu_type(const char *cpu) {
+	if (RZ_STR_EQ(cpu, "h8300") || RZ_STR_EQ(cpu, "H8300")) {
```
Rot127, line 1421: Could you add an `RZ_STR_EQI` for case insensitive comparison while you are at it please?

librz/arch/isa/h8300/h8300_il.c
```diff
@@ -239,6 +258,36 @@ static RzILOpEffect *rx_op_set(H8300Cmd *cmd, ut8 N, ut8 I, RzILOpPure *x) {
 #define RX_OP(N, I)   rx_op(cmd, N, I)
 #define RX_X(N, I, X) rx_op_set(cmd, N, I, X)
 
+static ut8 sp_index_from_cpu_type(int cpu_type) {
+	return cpu_type == CPU_H8300H ? H8300H_SP : H8300_SP;
+}
+
+static ut8 sp_width_from_cpu_type(int cpu_type) {
```
Rot127: Make them inline please. Also the others.

Rot127 (changes requested):

librz/arch/isa/h8300/h8300_il.c
b1llow: I thought that for modern compilers, the inline directive doesn't really do anything in .c files?

notxvilka: There are differences, see for yourself: https://godbolt.org/z/reW6jhTqY

b1llow: I've seen that as long as optimization O1 or above is turned on, all `static inline` or `static`  will be inlined, so I think the inline keyword is irrelevant in .c files, but may have other uses in .h files.

image

librz/arch/isa/h8300/h8300_disas.c
b1llow:
```
RZ_API int rz_str_casecmp(const char *s1, const char *s2) {
#ifdef _MSC_VER
	return stricmp(s1, s2);
#else
	return strcasecmp(s1, s2);
#endif
}
```

The string case-insensitive comparison in Rizin uses a non-standard function and behaves inconsistently with the standard `rz_str_cmp`. I feel this is not appropriate to implement in this PR?

librz/arch/isa/h8300/h8300_il.c
Rot127: Yeah, probably there is no difference for optimized builds.
But we don't know what kind of compilers people use. So I would still vote for being explicit.

librz/arch/isa/h8300/h8300_disas.c
notxvilka: Yes, can be done separately.

notxvilka (approved):

Rot127 (approved):

PR #5456, opened 2025-10-13
Fix(fC): show correct comment when flags overlap

### LINKED ISSUE #5316: `fC` not working properly
### Work environment

| OS/arch/bits (mandatory)                             | Kali x86_64
| File format of the file you reverse (mandatory)      | NONE
| Architecture/bits of the file (mandatory)            | NONE
| `rizin -v` full output, **not truncated** (mandatory)  | rizin 0.9.0 @ linux-x86-64 commit: 17d10b91e043bdc40f9ba0a77ceda52cdc9c6829

`fC` shows the comment for the wrong flag if two flags are defined by the user on the same offset

### Expected behavior

`fC` should show the comment of first flag

### Actual behavior

`fC` shows the comment of second flag

* `f one 0x10 first` (`f+` also leads to the same result) 
* `f+ two 0x10 second`
* `fC one`

running `fC one` will show the comment of the second flag, i.e: `second`, instead of `first`

PremadeS:

**Detailed description**

`fC` was showing comment of wrong flag if two flags existed at the same offset. See issue #5316  
...

**Test plan**

run test the provided test
OR, start rizin:
```bash
f foo 0x10 first
f+ bar 0x20 second
fC foo
``` 
the output will now be `first` instead of `second`, (the latter being incorrect)
...

**Closing issues**

closes #5316 
...

### DIFF
--- a/librz/core/cmd/cmd_flag.c
+++ b/librz/core/cmd/cmd_flag.c
@@ -881,12 +881,12 @@ RZ_IPI RzCmdStatus rz_flag_comment_handler(RzCore *core, int argc, const char **
 		}
 		return bool2status(flag_set_comment(item, argv[2]));
 	} else {
-		item = rz_flag_get_i(core->flags, rz_num_math(core->num, argv[1]));
-		if (item && item->comment) {
-			rz_cons_println(item->comment);
-		} else {
+		item = rz_flag_get(core->flags, argv[1]);
+		if (!item) {
 			RZ_LOG_ERROR("Cannot find the flag\n");
 			return RZ_CMD_STATUS_ERROR;
+		} else if (item->comment) {
+			rz_cons_println(item->comment);
 		}
 	}
 	return RZ_CMD_STATUS_OK;
--- a/test/db/cmd/cmd_flags
+++ b/test/db/cmd/cmd_flags
@@ -740,3 +740,15 @@ ERROR: Wrong number of arguments passed to `fO`, see its help with `fO?`
 ERROR: Usage: fO <glob>   # Flag as ordinals (sym.* func.* method.*)
 EOF
 RUN
+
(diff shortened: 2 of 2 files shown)


librz/core/cmd/cmd_flag.c
```diff
@@ -881,12 +881,12 @@ RZ_IPI RzCmdStatus rz_flag_comment_handler(RzCore *core, int argc, const char **
...
-		if (item && item->comment) {
-			rz_cons_println(item->comment);
-		} else {
+		item = rz_flag_get(core->flags, argv[1]);
+		if (!item) {
 			RZ_LOG_ERROR("Cannot find the flag\n");
 			return RZ_CMD_STATUS_ERROR;
+		} else if (item->comment) {
+			rz_cons_println(item->comment);
 		}
```
Rot127, line 890: Wouldn't we want to print the flag even if it doesn't have a comment?

Rot127 (changes requested):

librz/core/cmd/cmd_flag.c
PremadeS: > print the flag if the doesn't have a comment

Could you please elaborate on this?

The previous `fC` command always printed “Cannot find the flag” even when the flag actually existed but just didn’t have a comment. That was misleading since the flag was there, just missing a comment

The updated code now only shows the error if the flag truly isn’t found, and only prints the comment if one exists for that flag.
Otherwise nothing is printed
@Rot127

Rot127: Never mind. It makes sense this way.

Rot127 (approved):

wargio (approved):

notxvilka (approved):

PR #5505, opened 2025-11-05
Prototype for an RzIL based analysis chain

Rot127:
# Performance tests

- [ ] General runtime comparisons to old analysis. Slower is fine if the yield is bigger, but not slow for easily avoidable reasons.
- [ ] Analysing huge (3GB+) core dumps should never cancel due to out of memory or similar.

# TODOs

- [x] The statically known xrefs should not be an xref, but a successor address. One which shows under which condition the path was taken
- [x] Rename the BB_CFG to block CFG to make the difference between basic and non basic block clear.
- [ ] Update docs (and diagrams).
- [ ] Extend xrefs to contain the access size in bits.
- [ ] Currently it only scans sections, not segments. But this fails for firmware binaries which might only have segments.
- [x] Fix tests
- [x] Basic blocks still nedd the instruction offsets added.

# To merge

- https://github.com/rizinorg/rizin/pull/6382/
- https://github.com/rizinorg/rizin/pull/6269

# Performance TODOs

...

# Design changes

- [ ] On interpreter should be able to produce more than one abstract state.
  E.g. if the condition of a direct branch evaluates to a top value, it should produce a abstract state for each branch possibility.
  - [ ] This requires a central point which collects the state hashes. So we know when to terminate.
  - [ ] It also requires the interpreter to get a register/memory state to start out with.
  - [ ] Joining states is also not yet implemented (states are just hashed to check when an interpreter has reached a known state).
- [ ] Init all values of the VM with values according to the archs ABI. Should reduce number of false positive constants.
  - Can be implemented later with the memory model supporting stack, heap vars.
- [ ] Hashing the state is tricky. Only hashing the globals (registers) is not enough. If a program stores stuff on the stack. So if registers stay untouched the algorithm terminates too early.
  - [ ] There needs to be some configuration, what exactly belongs to the machine state and what not.
  - [ ] `hash_state` is super slow. It needs some faster way. Maybe hash the memory of the global hash table directly with sha (or other hw implemented hash functions?)
- [ ] We need an efficient way for the interpreter to store facts it observes during the execution of a basic block. E.g. if a certain register was read/written **within a given scope**.

## For review start with

- `rz_inquiry.h` and `rz_analysis.h` for the API.
- Main interpretation loop is in `rz_interpreter_run`.
- `interpreter_prototype.c` is the plugin.
- `rz_inquiry_interpreter` gives an idea how the whole setup and API usage of an interpreter looks like.
- `revng_function_detection.c` to see the detection algo. But this one is only temporary.
- Run the analysis with `aaaaIp <address>` (non is current offset). The address is the first entry point executed by the interpreter.
  - It still prints a lot of not yet disabled warnings. So better run it with `2> /dev/null`

**tldr: interpreter**

The interpreter simply collects all statically known branch targets.
Then starts interpreting at those addresses. It collects all yield the interpreter reports (one of them are xrefs).

After the interpreter finished (no new state reached) or failed. It then removes all addresses from the branch target vector which were covered by the interpreter (checks the xrefs it collected to do this).

Then picks the next branch target and starts interpreting from it again.

It continues until no branch targets are left.

**tldr: function detection**

The interpreter reports call candidates.
Those are instructions which might be calls (push address in the current basic block and branch).

The function detection iterates over all this call candidates and collects the called addresses.
These addresses are called "candidate function entry points (cfep)".

It iterates then over all cfep, adds a new function, and recurses into the basic block cfg, adding each basic block to the function.
Edges of call candidates are ignored.

Lastly, it converts the results to RzAnalysis objects.

## Notes

**Current design**

This PR is only there to discuss design decisions for now.

The RzInquiry module combines for this prototype several roles. In the final version these roles are supposed to be in separated modules. So we can improve the performance independently.

**Architecture implemented in this PR**

```
  Modules                 Queues           Interpreter                       
 RzInquiry              Ring Buffers
┌───────────────────┐
│ ┌─────────────┐   │                      ┌─────────────────┐ 
│ │             │◄────────────────────────►│                 │
│ │             │   │       ┌─────────────►│ Interpreter 0   │
│ │ IO Handler  ├◄─────┐  ┌─│─────────────►│                 │
│ │             │   │  │  │ │              └─────────────────┘
│ │             ├◄─────┼──┼─┼─┐            
│ └─────────────┘   │  │  │ │ │                               
│                   │  │  │ │ │                               
│                   │  │  │ │ │                               
│ ┌─────────────┐   │  │  │ │ │                               
│ │             ├◄─────┼──┼─┘ │            ┌─────────────────┐
│ │             │   │  └──┼───┼───────────►│                 │
│ │ Yield       ├◄────────┼───┼───────────►│ Interpreter 1   │
│ │ Consumer    │   │  ┌──┼───┼───────────►│                 │
│ │             ├◄─────┼──┼─┐ │            └─────────────────┘
│ └─────────────┘   │  │  │ │ │                               
└───────────────────┘  │  │ │ │                               
                       │  │ │ │                   ...         
  ┌─────────────┐      │  │ │ │                               
  │             ├◄─────┼──┘ │ │
  │ RzIL Cache  │      │    │ │            ┌─────────────────┐
  │             ├◄─────┘    │ └───────────►│                 │
  │             │           └─────────────►│ Interpreter n   │
  │             │◄────────────────────────►┤                 │                                           
  └─────────────┘                          └─────────────────┘
```

**Questions to discuss**

- ~~Added members `RzILOpArgsVar->hash` for performance.~~
- ~~Use abbreviation `intrpr` or `interp` instead of `interpreter` to shorten name prefixes?~~
  - I went with `interp`.
- Where to implement the "get_il_bb"? In the Arch plugins? Or should it be a helper function in `RzInquiry`? Or the IL cache?
- How to design the filter of yield queues?
  - Pass function + const filter data to interpreters?
  - Filter on queue at receiving end? (Probably super bad for performance).
- Configuration of interpreter.
  - Via plugin specific config? Like the hexagon module.
  - Passed during initialization?
- Should all the interpreters be their own module? `RzInquiry` is then only the binding. But can they be a sub-module of `RzInquiry`?
  - Then what kind of plugins does RzInquiry expose?
  - This PR has the interpereters as sub-modules of `RzInquiry`.
- Should the interpreter implement `RzILVM`? I would argue no for some reasons:
  - The interpreter should simply implement `RzAnalysisILVM`. It is way simpler.
- ~~The `eval` could only produce the delta between the old and new state.~~
  ~~This would save a lot of unnecessary copies of registers which are never touched.~~
  This would mean we can never trow away any of the earlier states.

**Performance**

- The register file in an interpreter should have an access of O(1), cannot use allocations and doesn't run any hash functions. If we aim for performance our hash map implementation is too slow IMO, just by number of instructions executed (it is still worst case O(n) I think). Instead it would be nice to have all registers concatenated as one flat array. Indexed by a reg number.
- Add Queue implementation which works with shared memory instead of lists. This would save allocations as well.
- The basic block cfg should handle bb splitting and such on the go. Not in an `O(|i2i| * |BBs|)` function.

## Function detection - Lessons learned

- Two CFGs can be generated. One with only statically know edges, one with edges discovered by the interpreters.
  - The interpreter CFG currently has BBs as nodes. The static CFG has instructions.
    Having BBs as nodes in the statuc graph is difficult to get. Because
    `RzAnalysisOp.type` are not as reliable as RzIL is. Maybe we can ignore this and
    just eat the inaccuracy?
  - This difference in node type means very expansive conversion.
    We really should get an r-tree implementation for that.
- CFG should be accessible via the API.
- CFG needs different kind of edges.
  - Normal branches.
  - Branch to unmapped address (invalid branches)
  - branch to "ignored" addresses (imported symbols and such).
  - Edges between calls an their return point (possibly not? Depends on our tail call detection).
- Different Basic Block node types
  - Normal basic block candidate
  - Unmapped basic block (imported, exported symbols and such).
- How to properly define the abstraction layers of basic blocks? Basic block (one entry, one exit), basic block candidate (one exit, possibly multiple entries, overlaps with other candidates), etc.
- Imported, exported, PLT stuff, ... symbols should be treated somehow special.
- Function Prelude's is very useful. We should have it available if possible.
- All discovered basic block candidates should be given to the interpreter.
- What ever the implementation does, it must be based on instruction packets for Sparc, Mips and Hexagon. Otherwise we never stop handling edge cases.
- The `RzAnalysisOp.type` is NOT a reliable source to distinguish jumps from calls or branches in general. `op` members should only be a source of static knowledge for the interpreter. The CFG detection should be based only on the interpretation results to generalize well.
- It's stupid to detect possible function entry points by assuming all outgoing edges from call candidates are going to an entry point. Its also not intuitive. The call candidate simply needs to store its target address it would call to.


librz/inquiry/interpreter/p/interpreter_prototype.c
```diff
@@ -14,6 +14,22 @@
 
 #define MAX_INVOCATIONS_PER_BLOCK 3
 
+bool state_as_str(RZ_NONNULL const RzInterpAbstrState *state,
+	RZ_NONNULL RZ_OUT RzStrBuf *sb,
+	void *plugin_data);
+
+RZ_OWN RzInterpAbstrVal *clone_val(const RzInterpAbstrVal *val, void *plugin_data) {
+	RzInterpAbstrVal *r = RZ_NEW0(RzInterpAbstrVal);
```
Rot127: Depending how you intend to use it, that one could also get an output pointer and copy the new data into it.

So if we actually move to the "register files are a flat vector instead of a hash map", it can simply pass a pointer into the vector here to copy the data into. Safes at least this one allocation and allows to allocate the output state once, then copy the abstr. vals into it.

librz/inquiry/interpreter/p/interpreter_prototype.c
```diff
@@ -222,6 +243,67 @@ static ut64 hash_state(RZ_NONNULL const RzInterpAbstrState *state, void *plugin_
...
+ * \brief Join (least upper bound) on var sets
+ * \return True if a was changed
+ */
+static bool join_vars(RZ_BORROW RZ_INOUT HtUP *a, RZ_BORROW RZ_IN HtUP *b) {
+	RzIterator *it = ht_up_as_iter_keys(a);
+	ut64 *k;
+	bool changed = false;
+	rz_iterator_foreach(it, k) {
+		RzInterpAbstrVal *av = ht_up_find(a, *k, NULL);
+		RzInterpAbstrVal *bv = ht_up_find(b, *k, NULL);
```
Rot127: We really should switch to vector storage at some point. Curious how much that is going to show up in the benchmarks later.

librz/inquiry/interpreter/p/interpreter_prototype.c
thestr4ng3r: Good point indeed, I also just had a case where copy would have been more appropriate. Sooner or later I will change it to that, only for now clone is a bit simpler because with copy operations you always have to consider the existing contents and whether you have to free them first.

librz/include/rz_inquiry/rz_interpreter.h
```diff
@@ -102,13 +102,22 @@ typedef struct {
...
+	RZ_INTERP_PC_CONST, ///< Single known value
+	RZ_INTERP_PC_UNREACHABLE, ///< Bottom/unreachable state, if this is set, pc field is unused and undefined
+	RZ_INTERP_PC_ANY ///< Top state, if this is set, pc field is unused and undefined
+} RzInterpPCState;
+
 typedef struct {
+	ut64 pc; ///< Interpreter location in the code. This is not necessarily identical to the ISA's program counter register, but simply points to the instruction to execute.
+	RzInterpPCState pc_state;
+	bool uninterpreted; ///< True if this state has not yet been started to interpret, i.e. is part of RzInterpFunctionState.queue
+
```
Rot127: Makes way more sense.

librz/include/rz_inquiry/rz_interpreter.h
```diff
@@ -266,14 +275,23 @@ typedef struct {
 	bool req_ok; ///< Set to true if IO request succeeded.
 } RzInterpIOResult;
 
+typedef struct {
+	RzList /*<RzInterpAbstrState>*/ *queue; ///< States that have to be interpreted still. If this is empty, a fixpoint has been reached.
```
Rot127: Could be a vector with push and pop?
Especially because vectors keep their capacity (I think lists don't?).

thestr4ng3r: Yes, definitely. I just ignore these details for now.

librz/inquiry/interpreter/prototype/eval_effect.c
```diff
@@ -76,14 +76,15 @@ RZ_IPI bool interpreter_prototype_eval_effect(RzInterpSet *iset,
 			RZ_LOG_DEBUG("PC is going to be set to an abstract value! Current PC = 0x%" PFMT64x "\n", pc);
 		}
 		ut64 target = rz_bv_to_ut64(eval_out.bv);
+		bool is_call = plugin_data->call_cand.store_addr;
```
Rot127:
```suggestion
		bool maybe_is_call = plugin_data->call_cand.store_addr;
```

I know. Super nitpicky. Feel free to ignore it.
Just highlight it here, so we don't forget to make the uncertainty of our assumption explicit.

librz/inquiry/README.md
```diff
@@ -0,0 +1,30 @@
...
+Ideas for optimization (do not implement any of these without profiling first):
+* evaluation (`eval_pure` in `interpreter.c` and possibly also `eval_effect`)
+  often temporarily allocates temporary abstract values if there is more than
+  one operand in an op.
+  E.g. for `(add x y)`, the output RzAbstractVal can be reused to evaluate `x`
+  into first, but `y` is allocated dynamically on the heap.
+  Because the size of dynamic values depends on the plugin, we can't use
+  regular stack memory. However we could build our own stack to the side and
+  re-use allocations that way.
+  For example an `RzVector` where the element size is given by the plugin.
```
Rot127: Should be enough. Since even archs with vector or matrix extentions won't handle values larger than 8KiB vectors or 16KiB matrices (absolute max 256*256 = 64KiB I guess) the L1 cache should still be big enough. Especially if we have one interpreter per core.

librz/inquiry/README.md
```diff
@@ -0,0 +1,30 @@
...
+  already known to be top.
+  E.g. for `(add x y)` with constant/top abstraction, if `x` evaluates to top,
+  we don't have to evaluate `y` anymore because the result will be top anyway.
+  However, this may not always be the case, e.g. for `(logand x y)`, if `x` is
+  top, `y` may evaluate to 0 and the result will be a constant again. And if
+  the plugin defines a more fine-grained abstraction then even more cases will
+  result in non-top results.
+  So the plugin must decide whether to short-circuit or not. It has to be
+  profiled if the cost of asking the plugin all the time is really compensated
+  by avoided evalutations in practical code.
```
Rot127: This doesn't need to be implemented with a callback to the plugin. If the flag is a simple bool flag, set by the plugin on interpreter init(), the branch predictor should easily take care of that.
Especially for stuff in logand, which is used in 90% of all instructions.

thestr4ng3r: Good point, though whether we can short-circuit not only depends on the plugin, but also the values and op, like in the example given above. At the same time, it is questionable whether these edge cases are worth handling, so this may be good to benchmark first.

librz/inquiry/interp/interpreter.c
```diff
@@ -287,6 +287,42 @@ static RzInterpBlock *interp_block_at(RzInterpRunContext *ctx, ut64 addr) {
...
+
+static int interp_block_with_op_at_cmp(const void *a, const void *b, void *user) {
+	const ut16 *av = a;
+	const ut16 *bv = b;
+	if (*av > *bv) {
+		return 1;
+	} else if (*av < *bv) {
+		return -1;
+	}
+	return 0;
```
Rot127: Couldn't they be cast to `int32_t` and simply subtract?

thestr4ng3r: Indeed, it was only not possible for ut64 iirc.

Rot127:
Can this be closed?

thestr4ng3r:
@Rot127 I would like to keep it open until the other one is merged, then rebase this one again so we do not lose any of the changes still here that might be worth looking at in the future and then close it if that makes sense.

Rot127:
Rebase is going to be very very annoying though.
And we risk loosing code when resolving conflicts.
What do you think about renaming this module here to RzStudy again, squash all commits and then rebase?

Issue #4939, opened 2025-02-26
Update shellcodes from Exploit-DB


notxvilka:
Currently, we rely solely on `sflib`, which is quite narrow. We should consider basing our database on top of the Exloit-DB one: https://gitlab.com/exploit-database/exploitdb/-/tree/main/shellcodes?ref_type=heads

PR #5530, opened 2025-11-15
Fix incorrect SFLIB include path for Windows x86-32 in rz-gg

### LINKED ISSUE #5519: rz-gg: sflib files does not exist for windows-x86-32
### Work environment

| OS/arch/bits (mandatory)                             | Windows 11 25H2 Build 26200.6899 x64
| File format of the file you reverse (mandatory)      | Compiling a C file
| Architecture/bits of the file (mandatory)            | x86/32
| `rizin -v` full output, **not truncated** (mandatory)  | ```rizin 0.9.0 @ windows-x86-64 commit: f8fa5b45b079cc722c7a978e354cc4fd9e627f16```

### Expected behavior
The file `include\librz/sflib/windows-x86-32/sflib.h` is supposed to exist since it is used by gcc to compile the c file.
### Actual behavior
File does not exist
- This is related to issue #5518 
- Command executed: `rz-gg -a x86 -b32 test.c`
- Contents of `test.c`
```c
int main() {
  write(1, "Hello World\n", 13);
  return 0;
}
```
Output of the command is
```
'gcc' -fPIC -fPIE -pie -fpic -m32 -fno-stack-protector -nostdinc -include 'C:\Users\abinr\Desktop\repos\rz\include\librz/sflib'/'windows-x86-32'/sflib.h -z execstack -fomit-frame-pointer -finline-functions -fno-zero-initialized-in-bss -o 'test.c.tmp' -S 'test.c'

''gcc'' is not recognized as an internal or external command,
operable program or batch file.
ERROR: egg: failure while parsing 'test.c'
ERROR: rz-gg: cannot load file "test.c"
```
You can see it tries to include a file `'C:\Users\abinr\Desktop\repos\rz\include\librz/sflib'/'windows-x86-32'/sflib.h` which does not exist.

Rohankaf:

### Summary
`rz-gg` attempts to include a non-existent SFLIB directory when targeting
Windows x86-32:

    include/librz/sflib/windows-x86-32/sflib.h

This directory does not exist in the Rizin include tree, causing rz-gg to fail
with a missing-header error when building 32-bit Windows payloads.

### Fix
A small fallback is added in `egg_Cfile.c` so that the nonexistent
`windows-x86-32` directory is mapped to the existing `windows-x86` directory:

```c
if (!strcmp(cEnv->TRIPLET, "windows-x86-32")) {
    free(cEnv->TRIPLET);
    cEnv->TRIPLET = rz_str_dup("windows-x86");
}
```

This ensures that rz-gg includes the correct file:

    include/librz/sflib/windows-x86/sflib.h

and avoids failing due to the missing `windows-x86-32` directory.

### Test Plan

Tested by running the following command:

./build/binrz/rz-gg/rz-gg -a x86 -b32 -k windows test.c

The generated compiler command now includes:

    .../sflib/windows-x86/sflib.h

instead of the incorrect:

    .../sflib/windows-x86-32/sflib.h

Warnings from clang on macOS are expected because Windows 32-bit binaries cannot
be compiled without MinGW. These warnings are unrelated to the fix. The key
part—the SFLIB include path—is now correct.

### Closing issues

closes #5519

### DIFF
--- a/librz/egg/egg_Cfile.c
+++ b/librz/egg/egg_Cfile.c
@@ -132,8 +132,7 @@ static struct cEnv_t *rz_egg_Cfile_set_cEnv(RZ_BORROW RZ_NONNULL RzPath *sys_pat
 		cEnv->TEXT = ".text";
 		cEnv->FMT = "pe";
 	} else if (isXNU(os)) {
-		// cEnv->TEXT = "0.__TEXT.__text";
-		cEnv->TEXT = "0..__text";
+		cEnv->TEXT = "0.__TEXT.__text";
 	} else {
 		cEnv->TEXT = ".text";
 	}
@@ -150,14 +149,16 @@ static struct cEnv_t *rz_egg_Cfile_set_cEnv(RZ_BORROW RZ_NONNULL RzPath *sys_pat
 		use_clang = true;
 		cEnv->TEXT = "0.__TEXT.__text";
 	}
-        if (!strcmp(cEnv->TRIPLET, "windows-x86-32")) {
-                free(cEnv->TRIPLET);
-                cEnv->TRIPLET = rz_str_dup("windows-x86");
-        }
+	if (!strcmp(cEnv->TRIPLET, "windows-x86-32")) {
+		free(cEnv->TRIPLET);
+		cEnv->TRIPLET = rz_str_dup("windows-x86");
+	}
 
-
-	buffer = rz_str_newf("%s -fno-stack-protector -nostdinc -include '%s'/'%s'/sflib.h",
-		cEnv->CFLAGS, cEnv->SFLIBPATH, cEnv->TRIPLET);
+	buffer = rz_str_newf(
+		"%s -fno-stack-protector -nostdinc -include '%s/%s/sflib.h'",
+		cEnv->CFLAGS,
+		cEnv->SFLIBPATH,
+		cEnv->TRIPLET);
 	if (!buffer) {
 		goto fail;
 	}
--- a/librz/egg/egg_Cfile.c
+++ b/librz/egg/egg_Cfile.c
@@ -150,6 +150,11 @@ static struct cEnv_t *rz_egg_Cfile_set_cEnv(RZ_BORROW RZ_NONNULL RzPath *sys_pat
 		use_clang = true;
 		cEnv->TEXT = "0.__TEXT.__text";
 	}
+        if (!strcmp(cEnv->TRIPLET, "windows-x86-32")) {
+                free(cEnv->TRIPLET);
+                cEnv->TRIPLET = rz_str_dup("windows-x86");
+        }
+
 
 	buffer = rz_str_newf("%s -fno-stack-protector -nostdinc -include '%s'/'%s'/sflib.h",
 		cEnv->CFLAGS, cEnv->SFLIBPATH, cEnv->TRIPLET);


notxvilka (changes requested):
Please fix the formatting, see contribution and developers guides

wargio (commented):
how about actually fixing the path in the triplet?

librz/egg/egg_Cfile.c
```diff
@@ -150,6 +150,11 @@ static struct cEnv_t *rz_egg_Cfile_set_cEnv(RZ_BORROW RZ_NONNULL RzPath *sys_pat
 		use_clang = true;
 		cEnv->TEXT = "0.__TEXT.__text";
 	}
+        if (!strcmp(cEnv->TRIPLET, "windows-x86-32")) {
+                free(cEnv->TRIPLET);
+                cEnv->TRIPLET = rz_str_dup("windows-x86");
+        }
+
 
 	buffer = rz_str_newf("%s -fno-stack-protector -nostdinc -include '%s'/'%s'/sflib.h",
```
wargio: this command buffer is weird. 

The output is like `'/usr/include/librz/sflib'/'linux-x86-32'/sflib.h` instead of `'/usr/include/librz/sflib/linux-x86-32/sflib.h'`

dotslashinit-sh:
I can't find `include/librz/sflib/windows-x86/sflib.h` either

Rohankaf:
@dotslashinit-sh 

> I can't find `include/librz/sflib/windows-x86/sflib.h` either

windows-x86 and windows-x86-32 SFLIB directories don’t exist on non-Windows builds only inside the Windows SDK.
The PR only fixes the incorrect TRIPLET mapping so that rz-gg stops referencing the nonexistent windows-x86-32 directory.
The missing files on macOS/Linux are expected.

dotslashinit-sh:
@Rohankaf 
> windows-x86 and windows-x86-32 SFLIB directories don’t exist on non-Windows builds only inside the Windows SDK.

sflib directories don't exist in the windows sdk, or maybe I can't find them. They don't seem to exist anywhere in the source either.

Rohankaf:
@dotslashinit-sh 

> sflib directories don't exist in the windows sdk, or maybe I can't find them. They don't seem to exist anywhere in the source either.

You're right  SFLIB doesn’t exist for Windows at all, not in the SDK and not in the source.  
Thanks for pointing that out.
The purpose of this PR is only to fix the wrong `windows-x86-32` triplet that caused rz-gg to generate an invalid include path.  
It doesn't attempt to add missing Windows SFLIB support that's a separate issue.
So the fix here is just to prevent rz-gg from building a path to a directory that never existed in the first place. 😃

wargio:
i don't think we ever supported sflib for windows stuff.. check `librz/include/sflib`

dotslashinit-sh:
Seems like radare2 does not support windows either. Wouldn't it be better to not include rz-gg with the releases for windows until we get proper support?

wargio:
It's not a problem of radare2 nor rizin. it's shellforge that never supported it. i would still build rz-gg for windows.

wargio:
@notxvilka & @Rot127 what do you think?

librz/egg/egg_Cfile.c
```diff
@@ -150,9 +149,16 @@ static struct cEnv_t *rz_egg_Cfile_set_cEnv(RZ_BORROW RZ_NONNULL RzPath *sys_pat
...
 	}
+	if (!strcmp(cEnv->TRIPLET, "windows-x86-32")) {
+		free(cEnv->TRIPLET);
+		cEnv->TRIPLET = rz_str_dup("windows-x86");
+	}
 
-	buffer = rz_str_newf("%s -fno-stack-protector -nostdinc -include '%s'/'%s'/sflib.h",
-		cEnv->CFLAGS, cEnv->SFLIBPATH, cEnv->TRIPLET);
+	buffer = rz_str_newf(
+		"%s -fno-stack-protector -nostdinc -include '%s/%s/sflib.h'",
```
wargio, line 158:
```suggestion
		"%s -fno-stack-protector -nostdinc -include '" RZ_JOIN_3_PATHS("%s", "%s, "sflib.h") "'",
```

In the original issue, the string of text generated was like: `C:\Users\USER\Desktop\repos\rz\include\librz/sflib'/'windows-x86-32'/sflib.h`

This will fix also the path output on windows to be like: `C:\Users\abinr\Desktop\repos\rz\include\librz\sflib\windows-x86-32\sflib.h`

Rot127:
`sflib.h` is needed to allow for syscalls in shell code?
Then we need new `sflib.h` for Windows, right? I don't think this is done quickly.
So simply removing `sflib.h` when rz-gg is build on Windows seems the simplest way.

notxvilka:
> `sflib.h` is needed to allow for syscalls in shell code? Then we need new `sflib.h` for Windows, right? I don't think this is done quickly. So simply removing `sflib.h` when rz-gg is build on Windows seems the simplest way.

There is an issue to track better option than sflib but it's way harder: https://github.com/rizinorg/rizin/issues/4939
 For now just making sure that it works for Linux binaries when on Windows platform is good enough.

Rohankaf:
@wargio Thanks for the feedback!  
Just to confirm what I should implement in this PR:

– Remove the sflib.h include when rz-gg is built on Windows  
– Keep the Linux behavior unchanged  
– Apply the suggested path join fix from @wargio

Please let me know if this is correct, and I will push the update.

wargio:
I think the best is not to remove windows sflib but instead fail if the file does not exists. @Rot127 @notxvilka 
Maybe an error like `sflib.h is not available for 'windows-x86-32'` so this behavior also applies to all the archs & os that also do not support it.

wargio:
@Rohankaf please check the checklist in the first post (i have edited it since you removed it) and tick everything that you did till now.

Rot127:
> but instead fail if the file does not exists.

I would not fail but just print a warning. Otherwise it won't be able to work on Windows by default, even if the shell code doesn't use any syscalls. Right?

Rot127:
> but then, does rz-gg work on an unsupported os?

I never used `rz-gg` so far. But according to the docs it is meant to compile code for injection later.
So it should also compile code which uses non of the OS API. E.g. if you need to exploit just a certain vulnerability in a user land process. So I think making rz-gg fully Windows compatible is more of a separated issue.

wargio:
yes indeed. so i would add a check if the file exists, if not, then throw a warning and do not include it

notxvilka:
@Rohankaf ping

PR #5562, opened 2025-11-30
Improve performance of `rz_bv_set_range()`

### LINKED ISSUE #4716: Enhance performance of `rz_bv_copy_nbits` and `rz_bv_set_range`

antonangeloff:

**Detailed description**

Changes in the `rz_bv_set_range()` implementation to process bits in bulk rather than one by one.

Execution times before changes:
```
Benchmark                                            Iterations Total time [ms] Average time [us/op] Throughput [ops/sec] 
--------------------------------------------------------------------------------------------------------------------------
rz_bv_set_range (60 bit)                                1000000      195.705000             0.195705       5109731.483611
rz_bv_set_range (100 bit)                               1000000      478.038000             0.478038       2091883.908811
```

After:
```
Benchmark                                            Iterations Total time [ms] Average time [us/op] Throughput [ops/sec] 
--------------------------------------------------------------------------------------------------------------------------
rz_bv_set_range (60 bit)                                1000000        6.586000             0.006586     151837230.488916
rz_bv_set_range (100 bit)                               1000000       19.396000             0.019396      51557022.066405
```
**Test plan**

Unit tests

**Closing issues**

Closes https://github.com/rizinorg/rizin/issues/4716

### DIFF
--- a/librz/util/bitvector.c
+++ b/librz/util/bitvector.c
@@ -203,7 +203,7 @@ RZ_API ut32 rz_bv_copy(RZ_NONNULL const RzBitVector *src, RZ_NONNULL RzBitVector
 /**
  * \brief Optimized version of rz_bv_copy_nbits() for large bitvectors (more than 64 bits) with bit positions aligned to BV_ELEM_SIZE
  */
-static ut32 rz_bv_copy_nbits_large_aligned(const RzBitVector *src, ut32 src_start_pos, RzBitVector *dst, ut32 dst_start_pos, ut32 nbit) {
+static ut32 bv_copy_nbits_large_aligned(const RzBitVector *src, ut32 src_start_pos, RzBitVector *dst, ut32 dst_start_pos, ut32 nbit) {
 	// Sanity check performed by caller
 	ut8 start_bits = RZ_MIN((BV_ELEM_SIZE - dst_start_pos) % BV_ELEM_SIZE, nbit);
 	ut8 trailing_bits = RZ_MIN((src_start_pos + nbit) % BV_ELEM_SIZE, nbit - start_bits);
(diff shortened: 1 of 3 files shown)


Rot127:
@antonangeloff Could you please add a test to see if your functions also work if the `src` and `dst` bit vector point to the same `RzBitvector` object?
We'd like to allow vector functions to work in place (without allocating a new one), and if your functions can work on the same pointer, it would be of great help in the future.

No need to do it for the previous PRs, because I add tests for this already. Just for this one it would be nice to see.

antonangeloff:
Hi @Rot127 there are test cases for large <-> large copies [here](https://github.com/rizinorg/rizin/blob/950428b0909d1db038120e085bb7246f4809d55c/test/unit/test_bitvector.c#L1290-L1294) and [here](https://github.com/rizinorg/rizin/blob/950428b0909d1db038120e085bb7246f4809d55c/test/unit/test_bitvector.c#L1330-L1333). There aren't tests for large <-> small because they are generally expected to be different vectors. 

The PR doesn't change any `rz_bv_copy_nbits()` code (apart from function renaming) btw, only `rz_bv_set_range()`. Let me know if you have any specific tests in mind, and I can add.

wargio (approved):

Rot127 (approved):
> There aren't tests for large <-> small because they are generally expected to be different vectors.

I currently change this. Because in RzIL we can have often the case that small vectors change to larger ones (e.g. when a 64bit vector is cast to a 128 bit vector or the other way around).
But these changes lgtm. `copy_nbits` work already in place. Will open a PR later and let you take a look.

notxvilka (approved):

PR #5569, opened 2025-12-04
Bitvector in-place refactor

Rot127:

**Detailed description**

Refactors multiple bit vector operations to work in place.

This is helpful for hot paths which don't want to allocate a new bit vector with each operation on them.

**Test plan**

Added

**Closing issues**

...

### DIFF
--- a/librz/util/bitvector.c
+++ b/librz/util/bitvector.c
@@ -879,11 +879,10 @@ RZ_API bool rz_bv_complement_1_inplace(RZ_INOUT RZ_NONNULL RzBitVector *bv) {
 RZ_API RZ_OWN RzBitVector *rz_bv_complement_1(RZ_NONNULL RzBitVector *bv) {
 	rz_return_val_if_fail(bv, NULL);
 
-	RzBitVector *ret = rz_bv_new(bv->len);
+	RzBitVector *ret = rz_bv_dup(bv);
 	if (!ret) {
 		return NULL;
 	}
-	rz_bv_copy(bv, ret);
 	if (!rz_bv_complement_1_inplace(ret)) {
 		rz_bv_free(ret);
 	}
--- a/librz/util/bitvector.c
+++ b/librz/util/bitvector.c
@@ -30,6 +30,21 @@ ut8 reverse_lt_8bits(ut8 x, ut8 w) {
 }
 
 /**
+ * \brief Resize or allocate bv->large_a to \p new_size bytes.
+ */
+static void resize_large_a(RzBitVector *bv, size_t n_bytes) {
+	if (bv->stack_alloc) {
+		bv->bits.large_a = RZ_NEWS0(ut8, n_bytes);
+		bv->stack_alloc = false;
+	} else if (!bv->bits.large_a) {
+		bv->bits.large_a = RZ_NEWS0(ut8, n_bytes);
+	} else {
+		bv->bits.large_a = realloc(bv->bits.large_a, n_bytes);
+	}
+	bv->_elem_len = n_bytes;
+}
+
+/**
  * \brief Initialize a RzBitVector structure
  * \param bv Pointer to a uninitialized RzBitVector instance
  * \param length int, the length of bitvector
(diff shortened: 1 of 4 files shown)


Rot127:
fyi @antonangeloff

antonangeloff:
Looks good! BTW doesn't `rz_bv_copy_nbits()` already work in place? Or `rz_bv_copy_nbits_inplace()` is added for API/naming consistency?

Rot127:
> Looks good! BTW doesn't rz_bv_copy_nbits() already work in place? Or rz_bv_copy_nbits_inplace() is added for API/naming consistency?

Yes, I wanted API consistency. As you see it uses the "not-inplace" function. So both are tested nonetheless.

notxvilka (changes requested):
While you are on it, could you please extend the bench to test also inplace functions? Would be interesting to compare the actual results.

librz/util/bitvector.c
```diff
@@ -141,15 +156,16 @@ RZ_API RZ_OWN char *rz_bv_as_hex_string(RZ_NONNULL const RzBitVector *bv, bool p
 	str[0] = '0';
 	str[1] = 'x';
 	ut32 j = 2;
-	for (ut32 i = 0; i < bv->_elem_len; i++) {
-		ut8 b8 = bv->bits.large_a[bv->_elem_len - i - 1];
+	ut32 n_elem = NELEM(bv->len, BV_ELEM_SIZE);
```
wargio, line 159: make it const if doesn't change.

librz/util/bitvector.c
```diff
@@ -383,6 +400,21 @@ static ut32 bv_copy_nbits_large_unaligned(const RzBitVector *src, ut32 src_start
...
 
+/**
+ * Copy n bits from start position of source to start position of dest, return num of copied bits
+ * \param bv Bitvector to copy bits inside.
+ * \param src_start_pos ut32, start position in source bitvector of copy
+ * \param dst_start_pos ut32, start position in destination bitvector
+ * \param nbit ut32, control the size of copy (in bits)
+ * \return copied_size ut32, Actual copied size
+ */
+RZ_API ut32 rz_bv_copy_nbits_inplace(
```
wargio: why not ut64 or a new bitv?

wargio (changes requested):
overall seems a nice improvement.

antonangeloff:
> Yes, I wanted API consistency. As you see it uses the "not-inplace" function. So both are tested nonetheless.

For what it's worth - from API perspective there seems to be 2 functions doing the same thing, and also the function `rz_bv_copy_nbits_inplace()` implies that it's counterpart (`rz_bv_copy_nbits`) doesn't do an in-place copy

If you think it makes sense `rz_bv_copy_nbits()` can be changed to do an actual out-of-place copy by allocating and returning a new bit vector

Rot127:
@notxvilka Here are the benchmark results of the `add` and `sub` (with debug build):

```
rz_bv_add(a, b, NULL)                                   1000000     1336.091000             1.336091        748452.014122
rz_bv_add_inplace(a, b, NULL)                           1000000     1011.513000             1.011513        988618.040500
rz_bv_sub(a, b, NULL)                                   1000000     1730.198000             1.730198        577968.533081
rz_bv_sub_inplace(a, b, NULL)                           1000000     1475.617000             1.475617        677682.623608
```

So something around 20-25%.
Which easily justifies it, if we assume the analysis is for some automated scanning service.

Generally I'd like to see the analysis of a BB happen mostly in cache. Accessing RAM only for IO reads/writes or lifting of rarely used instructions.

I did run some `perf mem record` already, and IO access had most of the cache misses.
But it is too early to say. Because we need the interpreter be in a state to analyze a whole binary.

librz/util/bitvector.c
notxvilka: @Rot127

notxvilka (approved):

librz/util/bitvector.c
Rot127: Sorry, overlooked it yesterday.
Removed it as @antonangeloff suggested.

wargio (approved):

PR #5590, opened 2025-12-08
Add getter for queue condition to allow unblocking waiting consumers.

Rot127:

**Detailed description**

Adds a getter for the thread condition of the queue. This allows to unblock a thread if it waits in `rz_th_queue_wait_pop`.
The main thread simply has to call `rz_th_cond_signal_all(rz_th_queue_get_cond(queue))`

**Test plan**

Added

**Closing issues**

...

### DIFF
--- a/librz/util/thread_queue.c
+++ b/librz/util/thread_queue.c
@@ -249,6 +249,35 @@ RZ_API size_t rz_th_queue_size(RZ_NONNULL RzThreadQueue *queue) {
 }
 
 /**
+ * \brief Returns the thread condition of the \p queue.
+ * In combination with rz_th_cond_signal() this can be used to unblock
+ * a waiting consumer.
+ *
+ * Example:
+ *   Consumer:
+ *   ```
+ *   // Waits to get element.
+ *   void *v = rz_th_queue_wait_pop(queue);
+ *   // After signal was sent and thread returned from rz_th_queue_wait_pop()
+ *   assert(v == NULL);
+ *   ```
+ *
+ *   Supplier:
+ *   ```
+ *   // Signals all threads waiting in the queue to return.
+ *   rz_th_cond_signal_all(rz_th_queue_get_cond(queue));
+ *   ```
+ *
+ * \param queue The queue to get the condition from.
+ *
+ * \return The condition or NULL in case of failure.
+ */
+RZ_API RZ_BORROW RzThreadCond *rz_th_queue_get_cond(RZ_NONNULL RzThreadQueue *queue) {
+	rz_return_val_if_fail(queue, NULL);
+	return queue->cond;
+}
+
+/**
  * \brief  Removes all elements from the queue, but does not awaits when empty.
  *
  * \param  queue The RzThreadQueue to pop from
--- a/librz/include/rz_th.h
+++ b/librz/include/rz_th.h
@@ -91,6 +91,7 @@ RZ_API RZ_OWN void *rz_th_queue_wait_pop(RZ_NONNULL RzThreadQueue *queue, bool t
 RZ_API RZ_OWN RzList /*<void *>*/ *rz_th_queue_pop_all(RZ_NONNULL RzThreadQueue *queue);
 RZ_API bool rz_th_queue_is_empty(RZ_NONNULL RzThreadQueue *queue);
 RZ_API bool rz_th_queue_is_full(RZ_NONNULL RzThreadQueue *queue);
+RZ_API RZ_BORROW RzThreadCond *rz_th_queue_get_cond(RZ_NONNULL RzThreadQueue *queue);
 RZ_API size_t rz_th_queue_size(RZ_NONNULL RzThreadQueue *queue);
 
 RZ_API RZ_OWN RzAtomicBool *rz_atomic_bool_new(bool value);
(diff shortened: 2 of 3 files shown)


librz/util/thread_queue.c
```diff
@@ -248,6 +248,20 @@ RZ_API size_t rz_th_queue_size(RZ_NONNULL RzThreadQueue *queue) {
 	return size;
 }
 
+/**
+ * \brief Returns the thread condition of the \p queue.
+ * In combination with rz_th_cond_signal() this can be used to unblock
+ * a waiting consumer.
```
notxvilka, line 254: Add an example please

notxvilka (changes requested):
LGTM with one small request

wargio:
why you need this? i don't think you should expose the internals of the queue.

Rot127:
@wargio As I said in the PR description. If a thread waits in the queue it can only be unblocked by using `rz_th_cond_signal_all`. Otherwise it waits there forever if the parent process has to exit.

wargio:
> @wargio As I said in the PR description. If a thread waits in the queue it can only be unblocked by using `rz_th_cond_signal_all`. Otherwise it waits there forever if the parent process has to exit.

Depends on how you use the queue. i would just add a call to `rz_th_cond_signal_all` into the free method to ensure the queue is not used.

What is the bug that you have found which cause a deadlock?

Rot127:
> i would just add a call to rz_th_cond_signal_all into the free method to ensure the queue is not used.

That is way better!

> What is the bug that you have found which cause a deadlock?

No bug, I just use the queue excessively in the interpreter prototype and ran into this deadlock.

Rot127:
Actually, I think moving this into the free function can introduce race conditions.

Because if the parent process has `n` users of the queue and it frees it. Some of those users are waiting, get a `NULL` result and know they have to exit.
But others might not waiting but use the queue nonetheless without wait. Leading to a use after free.

If the parent process cannot join all users before it frees the queue, this is a problem.
Right?

wargio:
can i see the code?

Rot127:
> can i see the code?

In my use case your proposal would work.
Here the child popping+wait from the queue: https://github.com/rizinorg/rizin/blob/35f09d39bbd56391d9d1de1060f73712d8bf07b5/librz/inquiry/interpreter/interpreter.c#L348-L353

Here the parent pushing into it: https://github.com/rizinorg/rizin/blob/35f09d39bbd56391d9d1de1060f73712d8bf07b5/librz/inquiry/inquiry.c#L287-L298

But my point from above is meant generally. That it is too easy to implement race conditions like this.

E.g.

```c
void parent() {
  RzThreadQueue q = ...;
  // dispatch multiple childs with the queue
  while (true) {
    SomeObject o = gen_object()
    if (!o) {
       // Parent needs to abort. How to signal this the children? Without an additional atomic bool guard?
       break;
    }
  }
  // Would signal all children to exit the queue.
  // And free the queue pointer directly afterwards
  rz_th_queue_free(q);
}

void child(RzThreadQueue q) {
  // A: Do something
  SomeObject *o = rz_tht_queue_pop(q);
  // B: Do something else
  o = rz_tht_queue_wait_pop(q);
  if (!o) {
    // parent signaled to leave the queue, return
    return;
  }
}
```

If the parent frees the queue while the child is in waiting in `rz_tht_queue_wait_pop()` everything is fine.
But if the child is executing `// A: Do something`, it will use the `q` pointer.

On the contrary, if we apply something like this PR here, there are two steps to 1. signal all childs to leave the queue, 2. free the queue.
So something like this is possible (parent only):

```c

void parent() {
  RzThreadQueue q = ...;
  // dispatch multiple childs with the queue
  while (true) {
    SomeObject o = gen_object()
    if (!o) {
       // Signal all children to leave the queue.
       break;
    }
  }
  // Wait for all children to terminate.
  rz_th_join(all_children);\
  // Only after all chilren aere joined it is guaranteed no child can use the q pointer.
  // free it.
  rz_th_queue_free(q);
}
```

wargio:
yeah, indeed is missing something. probably a close queue event, where the wait always fails.
Maybe we should implement `rz_th_queue_close/open()` to signal all and change the state to close.

Rot127:
Are you ok with moving `rz_th_cond_signal_all(queue->cond);` into `rz_th_queue_close()` or `rz_th_queue_close_call()`?. I think `rz_th_queue_open` is not needed. Because a queue is techincally open as soon as it is created.

Rot127:
Ok, this thing doesn't work as well. Just ran into the case that the child can of course dead lock if the signal is send just before it runs into `wait_pop`. Then it stays there indefinitely. Will send a PR when I have a proper solution.

wargio:
i think you should add a `bool` in the queue to check its status when it enters the `wait_pop`

wargio:
i think you want it to behave like the channels on go, right?

Rot127:
> i think you want it to behave like the channels on go, right?

I never used Go, but I think yes.

> i think you should add a bool in the queue to check its status when it enters the wait_pop

Will do something like this and open a new PR.

wargio:
let me propose the apis and fix the current impl

wargio:
> let me propose the apis and fix the current impl

https://github.com/rizinorg/rizin/pull/5627

PR #5637, opened 2025-12-18
doc(reg): improve README and add examples

Maijin:

Gemini 3 Flash / Antigravity

**Detailed description**
This PR improves the documentation and provides a practical example for the `librz/reg` module.

**Key Changes:**

- **Enhanced `librz/reg/README.md`**:
- **New Example

### DIFF
--- a/examples/api/reg/reg_example.c
+++ b/examples/api/reg/reg_example.c
@@ -1,0 +1,85 @@
+// SPDX-FileCopyrightText: 2025 Maijin <Maijin21@gmail.com>
+// SPDX-License-Identifier: LGPL-3.0-only
+
+/**
+ * Example: Using the RzReg module for register management
+ *
+ * This example demonstrates:
+ * - Creating an RzReg object
+ * - Loading a custom register profile
+ * - Setting and getting register values by name and role
+ * - Using arena snapshots (push/pop)
+ */
+
+#include <rz_reg.h>
+#include <rz_util.h>
+#include <stdio.h>
+#include <inttypes.h>
+
+int main(int argc, char **argv) {
+	// 1. Create a new register module instance
+	RzReg *reg = rz_reg_new();
+	if (!reg) {
+		fprintf(stderr, "Failed to create RzReg instance\n");
+		return 1;
+	}
+
+	// 2. Define and load a register profile
+	// Definitions: type name .bits offset packed
+	// Aliases: =role name
+	const char *profile =
+		"gpr  rax  .64  0  0\n"
+		"gpr  rbx  .64  8  0\n"
+		"gpr  rcx  .64  16 0\n"
+		"gpr  rip  .64  24 0\n"
+		"gpr  rsp  .64  32 0\n"
+		"flg  zf   .1   40.0 0\n"
+		"flg  cf   .1   40.1 0\n"
+		"=PC  rip\n" // program counter
+		"=SP  rsp\n"; // stack pointer
+
+	printf("Loading register profile...\n");
+	if (!rz_reg_set_profile_string(reg, profile)) {
+		fprintf(stderr, "Failed to load register profile\n");
+		rz_reg_free(reg);
+		return 1;
+	}
+
+	// 3. Set and get register values by name
+	printf("Setting rip to 0x400000\n");
+	rz_reg_setv(reg, "rip", 0x400000);
+
+	ut64 rip_val = rz_reg_getv(reg, "rip");
+	printf("rip = 0x%" PRIx64 "\n", rip_val);
+
+	// 4. Access values by role
+	ut64 pc_val = rz_reg_get_value_by_role(reg, RZ_REG_NAME_PC);
+	printf("PC (rip) = 0x%" PRIx64 "\n", pc_val);
+
+	// 5. Working with flags (bit-level access)
+	printf("Setting Zero Flag (zf) and Carry Flag (cf)\n");
+	rz_reg_setv(reg, "zf", 1);
+	rz_reg_setv(reg, "cf", 0);
+
+	printf("zf = %" PRIu64 "\n", rz_reg_getv(reg, "zf"));
+	printf("cf = %" PRIu64 "\n", rz_reg_getv(reg, "cf"));
+
+	// 6. Arena snapshots (Push/Pop)
+	printf("\n--- Arena Snapshot Example ---\n");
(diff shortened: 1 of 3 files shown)


librz/reg/README.md
```diff
@@ -0,0 +1,91 @@
+# RzReg
+
+`RzReg` is the Rizin module for managing and manipulating register sets for various architectures. It provides a portable way to define register layouts, store their values, and access them by name or role.
+
+## Components
+
```
Rot127:
```suggestion
```

librz/reg/README.md
```diff
@@ -0,0 +1,91 @@
...
+The module is divided into several key components:
+
+### RzReg (Core)
```
Rot127:
```suggestion
### RzReg
```

librz/reg/README.md
```diff
@@ -0,0 +1,91 @@
...
+- **Aliases**: Mapping common roles (like PC, SP) to specific registers.
+
+### RzRegArena
+A byte buffer that stores the actual values of registers. `RzReg` maintains a pool of arenas for each register type, supporting features like snapshots/undo.
+
+### RzRegItem
+Represents an individual register with its metadata (name, type, size, offset).
+
+## Register Profile Format
+
```
Rot127:
```suggestion
```

librz/reg/README.md
```diff
@@ -0,0 +1,91 @@
...
+Rizin uses a simple text format to define register profiles:
+
+- **Alias**: `=<role> <name>`
+- **Definition**: `(<type>@)<arena> <name> .<size> <offset> <packed> (# <comment>)`
```
Rot127, line 26: Explain what the `()` mean

librz/reg/README.md
```diff
@@ -0,0 +1,91 @@
...
+
+Example for a fictional 8-bit architecture:
+```
+gpr a .8 0 0
+gpr b .8 1 0
+=PC a
+```
```
Rot127: Duplicate I think. Above there was already one.

Rot127 (changes requested):
I struggle to see how the example covers a common use case.
Because the reg profile on its own is not that important I think?
Could you elaborate on this?

librz/reg/README.md
Maijin: Yeah I was thinking for flow had it again here, but I'll remove it yes

Maijin:
> I struggle to see how the example covers a common use case.
Because the reg profile on its own is not that important I think?
Could you elaborate on this?

@Rot127 Thanks for the review -  Yes I thought of this flow mostly because that way I can then show the whole chain from ingestion of the profile to the manipulation. Most Reg Profiles are actually hardcoded in different part of Rizin code instead of being defined here - perhaps that's something that could be improved in future. 

Open for ideas though if you think of a better chain to showcase this library.

Rot127:
> Open for ideas though if you think of a better chain to showcase this library.

For RzReg idk tbh. Don't have an immediate good use case.
But what would be great IMHO are examples which show basic analysis and how to get the results of it.

Something like:

- Open file
- Set some configs
- Analyze it
- Get all xrefs or similar.

Although this should wait until we have the new analysis ready in a few months.

But what would be useful is an example how to get the RzIL for an instruction.
Because it can be used easily from outside libraries.

Maijin:
@Rot127 - Yes I'm tackling each librz/* libs one by one 👍  I'll keep that in mind when working on librz/analysis and friends

examples/api/reg/reg_example.c
```diff
@@ -0,0 +1,83 @@
...
+	// Definitions: type name .bits offset packed
+	// Aliases: =role name
+	const char *profile =
+		"gpr  rax  .64  0  0\n"
+		"gpr  rbx  .64  8  0\n"
+		"gpr  rcx  .64  16 0\n"
+		"flg  zf   .1   24.0 0\n"
+		"flg  cf   .1   24.1 0\n"
+		"=PC  rax\n"
+		"=SP  rbx\n";
```
wargio:
```suggestion
		"=PC  rax\n" // program counter
		"=SP  rbx\n"; // stack pointer
```

Also why PC is rax?

librz/reg/README.md
```diff
@@ -0,0 +1,88 @@
...
+
+To add a new register profile for a supported or new architecture, follow these steps:
+
+1. **Create the Profile File**: Create a new `.sdb.txt` file in `librz/reg/d/`. Use the format described in the "Register Profile Format" section.
+2. **Update the Build System**: Open `librz/reg/d/meson.build` and add the base filename (without `.sdb.txt`) to the `sdb_files` list.
+
+```meson
+sdb_files = [
+    'avr-ATmega8-8',
+    'my-new-arch'  # Add your profile here
```
wargio, line 85: maybe i remember this wrong, but i think you can also load custom files, without changing the meson stuff.

Maijin: The README is to describe how to contribute so I think for now this is fine.

notxvilka (approved):

wargio (approved):

librz/reg/README.md
```diff
@@ -0,0 +1,91 @@
...
+]
```
Rot127, line 86: This is not true as far as I can see. The files in `reg/d/` map register names to memory addresses.
The register profile is only defined in `analysis_ARCH.c` as far as I know.

Rot127 (changes requested):

librz/reg/README.md
wargio: this is related to mmio registers, where for example in AVR the registers are also memory mapped, maybe it should be specified.

Rot127: Yes, and the docs here say
> Use the format described in the "Register Profile Format" section.

Which is not the correct syntax for those files.

Maijin: Yes actually the LLM had it right in the first place and then I got carried away when I was reviewing it and I changed this by mistake. I'll update this section and separate both.

When I was looking at these I was even thinking that registry profiles should all be moved away from the analysis_ARCH files and moved over here instead of being hardcoded in but that's a different discussion.

Rot127: > Yes actually the LLM had it right in the first place and then I got carried away when I was reviewing it and I changed this by mistake. 

This is why I think documentation is written in the best case, while you are actively doing something with the module. So you not just read the code, but interact with it.

PR #5650, opened 2025-12-19
Fix double-free bug in byte_pattern_search (CID 898818)

Ayushd785:
- [x] I've documented every `RZ_API` function and struct this PR changes. *(no public API changed)*
- [ ] I've added tests that prove my changes are effective (required for changes to `RZ_API`). *(not applicable)*
- [ ] I've updated the [Rizin book](https://github.com/rizinorg/book) with the relevant information (if needed). *(not required)*
- [x] I've used AI tools to generate fully or partially these code changes and I'm sure the changes are not copyrighted by somebody else. *(adjust as needed)*

**Detailed description**

This PR fixes a double-free / use-after-free in `librz/core/cmd/cmd_search.c` inside `byte_pattern_search()`:

- **Root cause:** `pattern` was passed to `rz_core_search_bytes()` which takes ownership and may free it. The `error` label in `byte_pattern_search()` unconditionally freed `pattern` again, causing use-after-free when the callee freed it and returned `NULL`.

- **Change:** after calling `rz_core_search_bytes(core, search_opts, pattern)` the local pointer is cleared (`pattern = NULL`) to mark the ownership transfer. The error cleanup now frees `pattern` only if the local pointer is non-NULL.
- **Scope:** single-file, minimal fix. No API or behavioral changes other than preventing memory corruption.

**Why this approach**

Nulling the pointer after ownership transfer is a small, common, and reviewer-friendly defensive pattern in C to avoid accidental double-free. The alternative (boolean flag) is logically equivalent; nulling the pointer is slightly cleaner and idiomatic.

**Test plan**

Test is not required here the change is very minimal


librz/core/cmd/cmd_search.c
```diff
@@ -1920,6 +1920,7 @@ static RzCmdStatus byte_pattern_search(RzCore *core, RZ_OWN RzSearchBytesPattern
 		goto error;
```
Rot127, line 1920: Free pattern here.

Rot127 (changes requested):
This is not the right approach I think. Just free `pattern` earlier and not at the `error` label.

librz/core/cmd/cmd_search.c
Ayushd785: Okay, got it so you are trying to say just free it after the use, no need for error validation.

Ayushd785: > Free pattern here.

yes it makes more sense i think because if we free it immediately after the use we don't have to validate later

Issue #3568, opened 2023-06-09
Reflines glitch in the presence of multiple CODE xrefs


XVilka:
### Work environment

| Questions                                            | Answers
|------------------------------------------------------|--------------------
| OS/arch/bits (mandatory)                             | *
| File format of the file you reverse (mandatory)      | ELF
| Architecture/bits of the file (mandatory)            | ARM
| `rizin -v` full output, **not truncated** (mandatory)  |  https://github.com/rizinorg/rizin/commit/6f3969f8c08ad8bb211b4fde4c6477b6513e6ee3

### Expected behavior

No additional reflines in lines with `CODE XREF...`

### Actual behavior

```
[0x00005fa8]> pdf @ sym._Unwind_VRS_Pop
            ; CALL XREFS from sym.__gnu_unwind_execute @ 0x81a8, 0x837a
│           ; UNKNOWN XREF from aav.0x00000100 @ 0x760
│           ;-- (0x00007c8f) aav.0x00007c8f:
╭ sym._Unwind_VRS_Pop(int16_t arg1, uint16_t arg2, int16_t arg3, int16_t arg4);
│           ; arg int16_t arg1 @ r0
│           ; arg uint16_t arg2 @ r1
│           ; arg int16_t arg3 @ r2
│           ; arg int16_t arg4 @ r3
│           0x00007c8e  ~   f0b5           push  {r4, r5, r6, r7, lr}
│           0x00007c90      041c           adds  r4, arg1              ; arg1
│           0x00007c92      c7b0           sub   sp, aav.0x0000011c    ; "43\x01"
│           0x00007c94      171c           adds  r7, arg3              ; arg3
│           0x00007c96      1d1c           adds  r5, arg4              ; arg4
│           0x00007c98      0429           cmp   r1, 4                 ; aav.0x00000004 ; arg2
│       ╭─< 0x00007c9a      00d9           bls   0x7c9e
│      ╭──< 0x00007c9c      83e0           b     0x7da6
│      │╰─> 0x00007c9e      081c           adds  r0, r1, 0
│      │    0x00007ca0      00f09cfb       bl    sym.__gnu_thumb1_case_uqi ; sym.__gnu_thumb1_case_uqi
│      │    0x00007ca4      031d           adds  r3, r0, 4             ; aav.0x00000004
│      │    0x00007ca6      812b           cmp   r3, 0x81              ; aav.0x00000081
│      │    0x00007ca8      5700           lsls  r7, r2, 1
[...]

PR #5659, opened 2025-12-21
Analysis support for GCC's `__gnu_thumb1_case_uqi` jump tables on ARM / Thumb-1

antonangeloff:

**Detailed description**

When targeting ARM (Thumb-1 mode) and optimizing for size, GCC uses a runtime helper function `__gnu_thumb1_case_uqi` for jumping to switch case blocks.

Currently the analysis code considers `bl __gnu_thumb1_case_uqi` as a standard function call and therefore skips analysis/indexing of basic blocks corresponding to the referenced switch cases. This PR adds analysis support for that kind of jump table which allows analyzing the case-related basic blocks.

Regarding https://github.com/rizinorg/rizin/issues/3568 - there are no refline glitches after indexing all of the function's basic blocks although the refline glitch itself is probably a separate issue, not fixed by this PR. The glitch can still be reproduced if `e analysis.jmp.tbl=false` is set before analysis.

Before:
image

After:
image

This is how it looks around `0x00007da6`:
image

**Test plan**

Added integration test

**Closing issues**

None

### DIFF
--- a/librz/arch/fcn.c
+++ b/librz/arch/fcn.c
@@ -1275,6 +1275,19 @@ static RzAnalysisBBEndCause run_basic_block_analysis(RzAnalysisTaskItem *item, R
 				}
 				gotoBeach(RZ_ANALYSIS_RET_END);
 			}
+
+			if (analysis->gnu_thumb1_case_uqi_addr && op.jump == analysis->gnu_thumb1_case_uqi_addr && analysis->opt.jmptbl) {
+				RzAnalysisJmpTableParams params = {
+					.jmp_address = op.addr,
+					.entry_size = 1,
+					.jmptbl_loc = op.addr + op.size,
+					.jmptbl_off = op.addr + op.size,
+					.sp = sp,
+					.tasks = tasks
+				};
+				ret = rz_analysis_walkthrough_arm_thumb1_case_uqi_table(analysis, fcn, bb, &params);
+				gotoBeach(RZ_ANALYSIS_RET_BRANCH);
+			}
 			break;
 		case RZ_ANALYSIS_OP_TYPE_UJMP:
 		case RZ_ANALYSIS_OP_TYPE_RJMP:
--- a/librz/arch/jmptbl.c
+++ b/librz/arch/jmptbl.c
@@ -433,6 +433,54 @@ RZ_API bool rz_analysis_walkthrough_arm_jmptbl_style(RZ_NONNULL RzAnalysis *anal
 }
 
 /**
+ * \brief Marks for analysis an ARM/Thumb-1 specific jump table (GCC's `__gnu_thumb1_case_uqi` format)
+ *
+ * This function works similarly to `rz_analysis_walkthrough_jmptbl`, but handles the compact
+ * byte-offset format used by the GCC's `__gnu_thumb1_case_uqi` helper function for size-optimized
+ * switch statements
+ *
+ * \param analysis Pointer to RzAnalysis instance
+ * \param fcn Pointer to RzAnalysisFunction to add the new cases
+ * \param block Pointer to RzAnalysisBlock that originates the switch table
+ * \param params Pointer to RzAnalysisJmpTableParams necessary to analyze the jump table
+ */
+RZ_API bool rz_analysis_walkthrough_arm_thumb1_case_uqi_table(RZ_NONNULL RzAnalysis *analysis, RZ_NONNULL RzAnalysisFunction *fcn, RZ_NONNULL RzAnalysisBlock *block, RZ_NONNULL RzAnalysisJmpTableParams *params) {
+	rz_return_val_if_fail(analysis && analysis->read_at && fcn && block && params, false);
+	ut8 table_size = 0;
+
+	if (params->table_count == 0) {
+		params->table_count = analysis->opt.jmptbl_maxcount;
+	}
+
+	for (table_size = 0; table_size < params->table_count; table_size++) {
+		ut8 offs;
+
+		if (!analysis->read_at(analysis, params->jmptbl_loc + table_size, &offs, 1)) {
+			return false;
+		}
+
+		if (offs == 0) {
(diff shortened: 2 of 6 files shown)


librz/arch/fcn.c
```diff
@@ -1275,6 +1275,19 @@ static RzAnalysisBBEndCause run_basic_block_analysis(RzAnalysisTaskItem *item, R
 				}
 				gotoBeach(RZ_ANALYSIS_RET_END);
 			}
+
+			if (analysis->gnu_thumb1_case_uqi_addr && op.jump == analysis->gnu_thumb1_case_uqi_addr && analysis->opt.jmptbl) {
```
wargio, line 1279: maybe we should test also if `is_arm`.

librz/arch/jmptbl.c
```diff
@@ -432,6 +432,54 @@ RZ_API bool rz_analysis_walkthrough_arm_jmptbl_style(RZ_NONNULL RzAnalysis *anal
...
+ * \param block Pointer to RzAnalysisBlock that originates the switch table
+ * \param params Pointer to RzAnalysisJmpTableParams necessary to analyze the jump table
+ */
+RZ_API bool rz_analysis_walkthrough_arm_thumb1_case_uqi_table(RZ_NONNULL RzAnalysis *analysis, RZ_NONNULL RzAnalysisFunction *fcn, RZ_NONNULL RzAnalysisBlock *block, RZ_NONNULL RzAnalysisJmpTableParams *params) {
+	rz_return_val_if_fail(analysis && analysis->read_at && fcn && block && params, false);
+	ut8 table_size = 0;
+
+	if (params->table_count == 0) {
+		params->table_count = analysis->opt.jmptbl_maxcount;
+	}
```
wargio:
```suggestion
	if (!params->table_count) {
		params->table_count = analysis->opt.jmptbl_maxcount;
	}
```

librz/arch/jmptbl.c
```diff
@@ -432,6 +432,54 @@ RZ_API bool rz_analysis_walkthrough_arm_jmptbl_style(RZ_NONNULL RzAnalysis *anal
...
+
+	for (table_size = 0; table_size < params->table_count; table_size++) {
+		ut8 offs;
+
+		if (!analysis->read_at(analysis, params->jmptbl_loc + table_size, &offs, 1)) {
+			return false;
+		}
+
+		if (offs == 0) {
```
wargio:
```suggestion
		if (!offs) {
```

librz/arch/jmptbl.c
```diff
@@ -432,6 +432,54 @@ RZ_API bool rz_analysis_walkthrough_arm_jmptbl_style(RZ_NONNULL RzAnalysis *anal
...
+RZ_API bool rz_analysis_walkthrough_arm_thumb1_case_uqi_table(RZ_NONNULL RzAnalysis *analysis, RZ_NONNULL RzAnalysisFunction *fcn, RZ_NONNULL RzAnalysisBlock *block, RZ_NONNULL RzAnalysisJmpTableParams *params) {
+	rz_return_val_if_fail(analysis && analysis->read_at && fcn && block && params, false);
+	ut8 table_size = 0;
+
+	if (params->table_count == 0) {
+		params->table_count = analysis->opt.jmptbl_maxcount;
+	}
+
+	for (table_size = 0; table_size < params->table_count; table_size++) {
+		ut8 offs;
```
wargio:
```suggestion
		ut8 offs = 0;
```
Always initialize

librz/arch/jmptbl.c
```diff
@@ -432,6 +432,54 @@ RZ_API bool rz_analysis_walkthrough_arm_jmptbl_style(RZ_NONNULL RzAnalysis *anal
...
+		ut64 jmp_ptr = params->jmptbl_loc + offs * 2;
+		apply_case(analysis, block, params->jmp_address, params->entry_size, jmp_ptr, table_size, params->jmptbl_loc + table_size);
+		rz_analysis_task_item_new(analysis, params->tasks, fcn, NULL, jmp_ptr, params->sp);
+	}
+
+	if (table_size == 0) {
+		return false;
+	}
+
+	if (params->default_case == 0 || params->default_case == UT32_MAX) {
```
wargio:
```suggestion
	if (!table_size) {
		return false;
	}

	if (!params->default_case || params->default_case == UT32_MAX) {
```

librz/arch/jmptbl.c
antonangeloff: Changing, although this part of the code was copied from another/existing function - `rz_analysis_walkthrough_casetbl()`

librz/arch/fcn.c
antonangeloff: Good point. There is an arch check when setting `analysis->gnu_thumb1_case_uqi_addr` [here ](https://github.com/rizinorg/rizin/pull/5659/changes#diff-e4de03bacabad8d61bf51daee587cbf69970dae65a272486119cdaebfbf9e87aR2315). So non-zero implies it's arm, but I can do additional check here if you think it helps. Let me know

librz/arch/jmptbl.c
antonangeloff: `offs` get [initialized by read_at()](https://github.com/rizinorg/rizin/pull/5659/changes#diff-59367e20bd29702216565a27df5c5e62f5acb0c7b7ebc94bdc8e898e935cad08R458) so even if we initialize to `0` here, the value will get discarded anyway. But I can still add that if you think it's better

notxvilka (approved):

librz/arch/jmptbl.c
wargio: always initialize.

librz/arch/fcn.c
wargio: ok, makes sense. ignore my previous comment

wargio (approved):

PR #5663, opened 2025-12-22
New `/Rgl` cmd

### LINKED ISSUE #5389: Improve `/Rg` output
Currently, we have two commands to print gadget information - `/R` and `/Rg`. First one contains bytes and disassembly along with addresses and the second - detailed information about the gadget. But there is no reasonable way to show both at the same time. I propose extending `/Rg` or adding more modes, e.g. `/Rgl`:
```
[0x00014348]> /R bx lr
...
  0x00022584           b4490100  strheq r4, [r1], -r4
  0x00022588           1eff2fe1  bx lr
Gadget size: 8

[0x00014348]> /Rg 0x22584
Gadget 0x22584
Stack change: 0x0
Changed registers: r1
Register dependencies:
Memory Write: r4 Initial Value: 0xffff New Value: 0xffff
Memory Write: r1 Initial Value: 0xffff New Value: 0xffff
Var Read: r4
Var Read: r1
````
I think adding `/Rgl` where `l` stands for the `RZ_OUTPUT_MODE_LONG` makes sense that will print both kind of information.

### PR DESCRIPTION
**Pull request by Arya-1-HR**

**Detailed description**

The current output of `/R` and `/Rg` which were basically printing ROP gadgets info , the output of these are combined in this functionality, This PR introduces `/Rgl` a longer/verbose version of `/Rg`.

**Test plan**
Screenshot 2026-01-03 at 11 31 50 PM

**Closing issues**
closes #5389

### CODE REVIEW

File: librz/core/rop.c [hunk 1]
```diff
@@ -943,6 +943,99 @@ static void rz_rop_gadget_print_json_mode(const RzCore *core, const RzRopGadgetI
...
+		return;
+	}
+	if (rz_io_read_at(core->io, addr, buf, size) <= 0) {
+		free(buf);
+		return;
+	}
+	int req_width = 50;
+	const bool colorize = rz_config_get_i(core->config, "scr.color");
+	rz_cons_printf("Gadget 0x%" PFMT64x " (size %d bytes)\n", addr, size);
+	rz_cons_printf("%s--%s\n", rz_str_repeat("-", req_width), rz_str_repeat("-", req_width));
```
wargio: `rz_str_repeat` returns chars that needs to be freed.

librz/core/rop.c
```diff
@@ -943,6 +943,99 @@ static void rz_rop_gadget_print_json_mode(const RzCore *core, const RzRopGadgetI
...
+		char *hex = rz_hex_bin2strdup(buf + idx, len);
+		const char *asm_str = rz_asm_op_get_asm(&asmop);
+		RzStrBuf *colored_asm = NULL;
+		if (colorize) {
+			colored_asm = get_colored_asm_str(core, &asmop, &aop);
+			asm_str = colored_asm ? rz_strbuf_get(colored_asm) : asm_str;
+		}
+		const char *reset_color = colorize ? Color_RESET : "";
+		int asm_len_clean = rz_str_ansi_len(asm_str);
+		int visible_len = 22 + asm_len_clean;
```
wargio: don't use `int`. use `size_t` for string lengths/sizes.

librz/core/rop.c
```diff
@@ -943,6 +943,99 @@ static void rz_rop_gadget_print_json_mode(const RzCore *core, const RzRopGadgetI
...
+		rz_cons_printf("  0x%08" PFMT64x "  %-8s  %s%s", addr + idx, hex ? hex : "", asm_str, reset_color);
+		int padding = req_width - visible_len;
+		if (padding > 0) {
+			rz_cons_printf("%*s", padding, "");
+		}
+		rz_cons_print(" | ");
+		if (instr_count == 0) {
+			rz_cons_printf("Stack change: 0x%" PFMT64x "\n", gadget_info->stack_change);
+		} else if (instr_count == 1) {
```
wargio:
```suggestion
		if (instr_count < 1) {
			rz_cons_printf("Stack change: 0x%" PFMT64x "\n", gadget_info->stack_change);
		} else if (instr_count == 1) {
```

librz/core/rop.c
```diff
@@ -943,6 +943,99 @@ static void rz_rop_gadget_print_json_mode(const RzCore *core, const RzRopGadgetI
...
+	}
+	int req_width = 50;
+	const bool colorize = rz_config_get_i(core->config, "scr.color");
+	rz_cons_printf("Gadget 0x%" PFMT64x " (size %d bytes)\n", addr, size);
+	rz_cons_printf("%s--%s\n", rz_str_repeat("-", req_width), rz_str_repeat("-", req_width));
+	RzAsmOp asmop = { 0 };
+	RzAnalysisOp aop = RZ_EMPTY;
+	int idx = 0;
+	int instr_count = 0;
+	while (idx < size) {
```
wargio: you can probably make this into a for loop

librz/core/rop.c
```diff
@@ -943,6 +943,99 @@ static void rz_rop_gadget_print_json_mode(const RzCore *core, const RzRopGadgetI
...
+		rz_asm_set_pc(core->rasm, addr + idx);
+		int len = rz_asm_disassemble(core->rasm, &asmop, buf + idx, size - idx);
+		if (len < 1)
+			break;
```
wargio: brackets.

librz/core/rop.c
```diff
@@ -943,6 +943,99 @@ static void rz_rop_gadget_print_json_mode(const RzCore *core, const RzRopGadgetI
...
+					}
+				}
+			}
+			rz_cons_newline();
+		} else {
+			rz_cons_newline();
+		}
+		if (colored_asm) {
+			rz_strbuf_free(colored_asm);
+		}
```
wargio:
```suggestion
		rz_strbuf_free(colored_asm);
```
No need to check for NULL when you call free.

librz/core/rop.c
```diff
@@ -943,6 +943,99 @@ static void rz_rop_gadget_print_json_mode(const RzCore *core, const RzRopGadgetI
...
+		} else if (instr_count == 1) {
+			rz_cons_printf("Modified regs: ");
+			if (gadget_info->modified_registers) {
+				void **it;
+				bool first = true;
+				rz_pvector_foreach (gadget_info->modified_registers, it) {
+					RzRopRegInfo *reg_info = (RzRopRegInfo *)*it;
+					if (reg_info && reg_info->name) {
+						if (!first)
+							rz_cons_printf(" ");
```
wargio: brackets

librz/core/rop.c
```diff
@@ -943,6 +943,99 @@ static void rz_rop_gadget_print_json_mode(const RzCore *core, const RzRopGadgetI
...
+	if (rz_io_read_at(core->io, addr, buf, size) <= 0) {
+		free(buf);
+		return;
+	}
+	int req_width = 50;
+	const bool colorize = rz_config_get_i(core->config, "scr.color");
+	rz_cons_printf("Gadget 0x%" PFMT64x " (size %d bytes)\n", addr, size);
+	rz_cons_printf("%s--%s\n", rz_str_repeat("-", req_width), rz_str_repeat("-", req_width));
+	RzAsmOp asmop = { 0 };
+	RzAnalysisOp aop = RZ_EMPTY;
```
wargio: decide what you want to use. it's either `{ 0 }` or `RZ_EMPTY`

librz/core/rop.c
```diff
@@ -943,6 +943,99 @@ static void rz_rop_gadget_print_json_mode(const RzCore *core, const RzRopGadgetI
...
+static void rz_rop_gadget_print_long_mode(const RzCore *core, const RzRopGadgetInfo *gadget_info, const RzRopSearchContext *context) {
+	rz_return_if_fail(gadget_info && context);
+
+	ut64 addr = gadget_info->address;
+	int size = gadget_info->size;
+	ut8 *buf = RZ_NEWS0(ut8, size);
+	if (!buf) {
+		return;
+	}
+	if (rz_io_read_at(core->io, addr, buf, size) <= 0) {
```
wargio:
```suggestion
	if (!buf || rz_io_read_at(core->io, addr, buf, size) < 1) {
```

librz/core/rop.c
```diff
@@ -943,6 +943,99 @@ static void rz_rop_gadget_print_json_mode(const RzCore *core, const RzRopGadgetI
 	pj_end(pj);
 }
 
+static void rz_rop_gadget_print_long_mode(const RzCore *core, const RzRopGadgetInfo *gadget_info, const RzRopSearchContext *context) {
+	rz_return_if_fail(gadget_info && context);
```
wargio: you should not add an assert here. this should be checked in the main function.

librz/core/rop.c
wargio: also you can allocate once and reuse the variable within the printf twice.

librz/core/rop.c
```diff
@@ -943,6 +943,99 @@ static void rz_rop_gadget_print_json_mode(const RzCore *core, const RzRopGadgetI
...
+		} else if (instr_count == 2) {
+			rz_cons_printf("Dependencies:  ");
+			if (gadget_info->dependencies) {
+				RzListIter *iter;
+				RzRopRegInfo *dep_info;
+				bool first = true;
+				rz_list_foreach (gadget_info->dependencies, iter, dep_info) {
+					if (dep_info && dep_info->name) {
+						if (!first)
+							rz_cons_printf(" ");
```
wargio: brackets

librz/core/rop.c
```diff
@@ -943,6 +943,99 @@ static void rz_rop_gadget_print_json_mode(const RzCore *core, const RzRopGadgetI
...
+			}
+			rz_cons_newline();
+		} else if (instr_count == 2) {
+			rz_cons_printf("Dependencies:  ");
+			if (gadget_info->dependencies) {
+				RzListIter *iter;
+				RzRopRegInfo *dep_info;
+				bool first = true;
+				rz_list_foreach (gadget_info->dependencies, iter, dep_info) {
+					if (dep_info && dep_info->name) {
```
wargio: invert the if and call `continue` in it.

librz/core/rop.c
```diff
@@ -943,6 +943,99 @@ static void rz_rop_gadget_print_json_mode(const RzCore *core, const RzRopGadgetI
...
+		if (instr_count == 0) {
+			rz_cons_printf("Stack change: 0x%" PFMT64x "\n", gadget_info->stack_change);
+		} else if (instr_count == 1) {
+			rz_cons_printf("Modified regs: ");
+			if (gadget_info->modified_registers) {
+				void **it;
+				bool first = true;
+				rz_pvector_foreach (gadget_info->modified_registers, it) {
+					RzRopRegInfo *reg_info = (RzRopRegInfo *)*it;
+					if (reg_info && reg_info->name) {
```
wargio: invert the `if` and call `continue` in it

librz/core/rop.c
```diff
@@ -943,6 +943,99 @@ static void rz_rop_gadget_print_json_mode(const RzCore *core, const RzRopGadgetI
...
+		return;
+	}
+	int req_width = 50;
+	const bool colorize = rz_config_get_i(core->config, "scr.color");
+	rz_cons_printf("Gadget 0x%" PFMT64x " (size %d bytes)\n", addr, size);
+	rz_cons_printf("%s--%s\n", rz_str_repeat("-", req_width), rz_str_repeat("-", req_width));
+	RzAsmOp asmop = { 0 };
+	RzAnalysisOp aop = RZ_EMPTY;
+	int idx = 0;
+	int instr_count = 0;
```
wargio: use `size_t`, not `int`

librz/core/rop.c
```diff
@@ -943,6 +943,99 @@ static void rz_rop_gadget_print_json_mode(const RzCore *core, const RzRopGadgetI
...
+	int size = gadget_info->size;
+	ut8 *buf = RZ_NEWS0(ut8, size);
+	if (!buf) {
+		return;
+	}
+	if (rz_io_read_at(core->io, addr, buf, size) <= 0) {
+		free(buf);
+		return;
+	}
+	int req_width = 50;
```
wargio: this is hardcoded to `50`, make it const.

wargio (changes requested):
see comments

librz/core/rop.c
Arya-1-HR: the rest printing modes also have added the assert , shall I remove them too?

librz/core/rop.c
Arya-1-HR: some API's like `rz_asm_disassemble` return negative values , so I will try to keep the returned length in `int`

librz/core/rop.c
wargio: only if you have modified/added them.
the assert should be triggered at the first check. if the same UNMODIFIED value is used over and over in other sub functions, these are supposed to skip the check since it was already performed.

Arya-1-HR: The check is already present in them before calling them all

notxvilka (changes requested):
Please add a test for this command

test/db/cmd/cmd_rop
```diff
@@ -2867,3 +2867,572 @@ Gadget size: 8
...
+  0x08048474  8db600000000  lea esi, dword [esi]         | Stack change: 0xffffffff00000008
+  0x0804847a  8dbf00000000  lea edi, dword [edi]         | Modified regs: esi edi esp ebp
+  0x08048480  c9        leave                        | Dependencies:  esi edi ebp esp esp esp esp
+  0x08048481  c3        ret                          | 
+
+Gadget 0x8048475 (size 13 bytes)
+------------------------------------------------------------------------------------------------------
+  0x08048475  b600      mov dh, 0x00                 | Stack change: 0x4
+  0x08048477  0000      add byte [eax], al           | Modified regs: edx ecx esp
+  0x08048479  008dbf000000  add byte [ebp+0xbf], cl      | Dependencies:  edx eax eax eax ebp ebp ecx ecx esp esp
```
notxvilka: Looks like padding algorithm should be improved. In theory we have RzTable structure and API, but it can't handle multiline cells, so for now it's better to keep it manual. Nevertheless, it should calculate maximum string length (and maybe cap it to some value?) before printing the table, to provide better looks.

notxvilka (changes requested):

test/db/cmd/cmd_rop
Arya-1-HR: It was fixed length though, sometimes the assembly will be too long to push the info to new line

notxvilka: @Arya-1-HR yeah, so the idea is to:
- calculate the proper value before printing instead of relying on the fixed length
- still limit the length to some maximum value and do not print disassembly further (just add `...` at the end of the disassembly line)

notxvilka:
Please also rebase to resolve the conflict in the `cmd_rop` test

notxvilka:
Please resolve the alignment thing and it's in a good shape

Arya-1-HR:
> Please resolve the alignment thing and it's in a good shape

I am working on the issue [https://github.com/rizinorg/rizin/issues/5682](url) , once its done I will implement it here

notxvilka:
> > Please resolve the alignment thing and it's in a good shape
> 
> I am working on the issue [https://github.com/rizinorg/rizin/issues/5682](url) , once its done I will implement it here

Emm, I recommend finishing this PR before - just manually building the output like you did already. That RzTable issue is likely quite tricky and can take way more time.

notxvilka (changes requested):
Looks good now, only couple nitpicks left

notxvilka:
Strange, for some reason I don't see my comments. Thus doubling it here:
- Some code can be shared with the "standard" ROP printing, just extract these shared pieces into small static functions, like it's done for `get_colored_asm_str()`, for example and reuse them in both functions
- For horizontal and vertical line use longer characters when `scr.utf8` is enabled, e.g. in `rz_cons_printf("%s--%s\n", rep_str, rep_str);`

librz/core/rop.c
```diff
@@ -943,6 +948,97 @@ static void rz_rop_gadget_print_json_mode(const RzCore *core, const RzRopGadgetI
...
+	ut8 *buf = RZ_NEWS0(ut8, size);
+	if ((!buf || rz_io_read_at(core->io, addr, buf, size) < 1)) {
+		free(buf);
+		return;
+	}
+	const int req_width = 50;
+	char *rep_str = rz_str_repeat("-", req_width);
+	const bool colorize = rz_config_get_i(core->config, "scr.color");
+	rz_cons_printf("Gadget 0x%" PFMT64x " (size %d bytes)\n", addr, size);
+	rz_cons_printf("%s--%s\n", rep_str, rep_str);
```
notxvilka: Dont forget about unicode symbols

librz/core/rop.c
```diff
@@ -943,6 +948,97 @@ static void rz_rop_gadget_print_json_mode(const RzCore *core, const RzRopGadgetI
...
+		rz_cons_printf("  0x%08" PFMT64x "  %.16s", addr + idx, hex ? hex : "");
+		int hex_len = hex ? RZ_MIN((int)strlen(hex), 16) : 0;
+		rz_cons_printf("%*s%s%s", 2 + (16 - hex_len), "", asm_str, reset_color);
+		size_t asm_len_clean = rz_str_ansi_len(asm_str);
+		size_t left_col_len = 10 + asm_len_clean;
+		int padding = req_width - left_col_len;
+		if (padding > 0) {
+			rz_cons_printf("%*s", padding, "");
+		}
+		rz_cons_print(" | ");
```
notxvilka: same

notxvilka (changes requested):

librz/core/rop.c
Arya-1-HR: Screenshot 2025-12-30 at 10 26 10 PM

notxvilka: Put this into the PR description

test/db/cmd/cmd_rop
```diff
@@ -2948,3 +2948,572 @@ Testing MOV_OP_CONST
 Testing MOV_OP_REG
 EOF
 RUN
+
+NAME=/Rgl test
+FILE=bins/elf/crackme0x05
+CMDS=<<EOF
+aaa
+/Rgl @ 0x804868f
```
notxvilka: `@ 0x804868f` here is actually unnecessary since `/Rgl` just lists *all* gadgets

notxvilka (changes requested):

librz/core/rop.c
```diff
@@ -943,6 +948,104 @@ static void rz_rop_gadget_print_json_mode(const RzCore *core, const RzRopGadgetI
...
+	rz_return_if_fail(core && core->analysis);
+
+	ut64 addr = gadget_info->address;
+	ut32 size = gadget_info->size;
+	ut8 *buf = RZ_NEWS0(ut8, size);
+	if ((!buf || rz_io_read_at(core->io, addr, buf, size) < 1)) {
+		free(buf);
+		return;
+	}
+	const int req_width = 50;
```
notxvilka, line 1045: Please don't use predefined constant but calculate it on the fly before printing.

notxvilka (changes requested):

librz/core/rop.c
Arya-1-HR: It doesn't make the padding 50 but subtracts after calculating the `len` it , Its like `max_len`

notxvilka: Then I wonder why the line in tests is too long. It should correspond to the actual disasm + gadget information length

librz/core/rop.c
```diff
@@ -943,6 +948,104 @@ static void rz_rop_gadget_print_json_mode(const RzCore *core, const RzRopGadgetI
...
+		int len = rz_asm_disassemble(core->rasm, &asmop, buf + idx, size - idx);
+		if (len < 1) {
+			break;
+		}
+		rz_analysis_op(core->analysis, &aop, addr + idx, buf + idx, size - idx, RZ_ANALYSIS_OP_MASK_BASIC);
+		char *hex = rz_hex_bin2strdup(buf + idx, len);
+		char *asm_str = get_colored_asm(core, &asmop, &aop);
+		const char *reset_color = colorize ? Color_RESET : "";
+		rz_cons_printf("  0x%08" PFMT64x "  %.16s", addr + idx, hex ? hex : "");
+		int hex_len = hex ? RZ_MIN((int)strlen(hex), 16) : 0;
```
notxvilka: If you use `size_t` here, no need for cast to `int`

librz/core/rop.c
```diff
@@ -943,6 +948,104 @@ static void rz_rop_gadget_print_json_mode(const RzCore *core, const RzRopGadgetI
...
+		size_t asm_len_clean = rz_str_ansi_len(asm_str);
+		size_t left_col_len = 10 + asm_len_clean;
+		int padding = req_width - left_col_len;
+		if (padding > 0) {
+			rz_cons_printf("%*s", padding, "");
+		}
+		rz_cons_print(utf8 ? " │ " : " | ");
+		if (instr_count < 1) {
+			rz_cons_printf("Stack change: 0x%" PFMT64x "\n", gadget_info->stack_change);
+		} else if (instr_count == 1) {
```
notxvilka: Maybe split this big block into 2 different small functions that handle each instruction count case separately

notxvilka (changes requested):

librz/core/rop.c
Arya-1-HR: Actually the logic prints like line by line so for one `addr`if you are printing the gadget info before you calculate the next `|` position the previous `|` is already printed , so keeping this manual is the only option

librz/core/rop.c
Arya-1-HR: in next line you will get a compiler warning while printing , even if you try to change the format specifier

librz/core/rop.c
notxvilka: Make it shorter:

Image

notxvilka (changes requested):

notxvilka (changes requested):
```
Missing type comment at <rizin/librz/core/rop.c:989:79> (token is not a comment)
Missing type comment at <rizin/librz/core/rop.c:989:95> (token is not a comment)
Missing type comment at <rizin/librz/core/rop.c:989:111> (token is not a comment)
Missing type comment at <rizin/librz/core/rop.c:989:132> (token is not a comment)
```

librz/core/rop.c
```diff
@@ -943,6 +948,150 @@ static void rz_rop_gadget_print_json_mode(const RzCore *core, const RzRopGadgetI
...
+	const int req_width = 50;
+	char *rep_str = NULL;
+	bool utf8 = rz_config_get_b(core->config, "scr.utf8");
+	const bool colorize = rz_config_get_i(core->config, "scr.color");
+	rz_cons_printf("Gadget 0x%" PFMT64x " (size %d bytes)\n", addr, size);
+	if (utf8) {
+		rep_str = rz_str_repeat("–", req_width);
+		rz_cons_printf("%s––%s\n", rep_str, rep_str);
+	} else {
+		rep_str = rz_str_repeat("--", req_width);
```
notxvilka: Here it should be only one `-`, that's why lines are too long. FIx this and I will merge

notxvilka (changes requested):

notxvilka (approved):

librz/core/rop.c
Arya-1-HR: > Here it should be only one `-`, that's why lines are too long. FIx this and I will merge

ok cool👍🏻

PR #5674, opened 2025-12-25
Added tests for wai command

### LINKED ISSUE #2714: Add test for wai command
The command `wai <instructions1> <instructions2> ... ` which should assemble instruction(s) and write bytes inside the current instruction boundaries is not tested within `test/db/`

This instruction also is not mapped on the book.

What the command does is tring to fit a new instruction within the boundaries of the old one.

for example in x86 changing a long instruction like `lea` which may use **8** bytes with a `mov` which uses **4**, will be automatically handled like `lea ....` to `mov ... ; nop ; nop ; nop ; nop`

so the command usage is the same as `wa` with the exception that you do not have to care about the boundaries of the instruction that you are overwriting

Ayushd785:
- [ ] I've used AI tools to generate fully or partially these code changes and I'm sure the changes are not copyrighted by somebody else. (AI is not used to write the tests)

**Detailed description**

This PR adds tests for the `wai` (write assembly inside) command in `test/db/cmd/cmd_w`, addressing the issue where the command was not tested within `test/db/`.

**What was added:**
- Test for single instruction: `wai nop` - replaces a 7-byte `lea` instruction with a 1-byte `nop` + 6 NOP padding bytes
- Test for multiple instructions: `wai nop nop` - replaces a 7-byte `lea` instruction with 2-byte `nop nop` + 5 NOP padding bytes

Both tests verify:
- The command correctly fits new instructions within old instruction boundaries
- Automatic NOP padding works correctly
- The syntax `wai <instructions1> <instructions2> ...` works as specified

**Test plan**
image
image

**Closing issues**

closes #2714


Ayushd785:
hii @wargio,I’ve added two tests for wai: one for replacing with a single instruction and one with multiple instructions. Both correctly fit within the original instruction size and pad the remaining bytes with NOPs automatically. I am open to your feedback and implement wisely. Thank you!

notxvilka (changes requested):
Try using some different instruction instead of "nop" to trigger more code coverage

Ayushd785:
> Try using some different instruction instead of "nop" to trigger more code coverage

Okay, I'm on it

Ayushd785:
Hii @notxvilka, now it tests for `mov rax, rbx` (3 Bytes) which results into 3 bytes MOV + 4 bytes NOP padding and for multiple instructions case it tests for instructions `mov rax, rbx` and `mov rcx, rdx` which is (6 bytes in total) it results 6 bytes + 1 byte NOP padding.

notxvilka (approved):

test/db/cmd/cmd_w
```diff
@@ -233,6 +233,51 @@ nop
...
+wai "mov rax, rbx"
+s 0
+pi 1
+p8 7
+EOF
+EXPECT=<<EOF
+lea rax, qword [rbp+0x34567806]
+488d8506785634
+mov rax, rbx
+4889d890909090
```
wargio: this should output `mov;nop;nop;..`
instead of using `pi` we should use `pD <n-bytes>`

test/db/cmd/cmd_w
```diff
@@ -233,6 +233,51 @@ nop
 EOF
 RUN
 
+NAME=wai
```
wargio, line 236: merge together all these tests.

wargio (changes requested):
merge together the tests

Ayushd785:
hii @wargio, I have replaced `pi 1` with `pD 7`, which now prints all the 7 Bytes for case 1 `mov rax, rbx` (3 Bytes) + 4 Nop and for case 2 `mov rax, rbx` and `mov rcx, rdx` (3 Bytes each,) meaning 6 Bytes in total + 1 Nop padding and merged both the tests into a single test as per your suggestion. I am open for any feedback and happy to implement further. Thank you!

wargio (approved):
All the `s 0` could be removed, but l think is ok as is

Ayushd785:
> All the `s 0` could be removed, but l think is ok as is

You're right `s 0` isn’t strictly necessary. I only kept it to ensure everything starts from `offset 0` for consistency. If you prefer, I can remove it, no problem at all.

notxvilka (changes requested):
1. Remove `s 0` then I will merge
2. For the future, we have `@` modifier that allows specifying address, e.g. `wx 9090 @ 0x10` will write two bytes at the `0x10` offset.

Ayushd785:
> 1. Remove `s 0` then I will merge
> 2. For the future, we have `@` modifier that allows specifying address, e.g. `wx 9090 @ 0x10` will write two bytes at the `0x10` offset.

Sure, I have removed all the occurrences of `s 0 `from the test. Thank you!

notxvilka (approved):

PR #5700, opened 2025-12-29
New function rz_pvector_find_index

SSharshunov:

**Detailed description**

...

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/librz/include/rz_vector.h
+++ b/librz/include/rz_vector.h
@@ -315,7 +315,7 @@ RZ_API void **rz_pvector_contains(RzPVector *vec, const void *x);
 RZ_API RZ_BORROW void **rz_pvector_find(RZ_NONNULL const RzPVector *vec, RZ_NONNULL const void *element, RZ_NONNULL RzPVectorComparator cmp, void *user);
 
 // find the element index in the vec based on cmparator
-RZ_API RZ_BORROW size_t rz_pvector_find_index(RZ_NONNULL const RzPVector *vec, RZ_NONNULL const void *element, RZ_NONNULL RzPVectorComparator cmp, void *user);
+RZ_API size_t rz_pvector_find_index(RZ_NONNULL const RzPVector *vec, RZ_NONNULL const void *element, RZ_NONNULL RzPVectorComparator cmp, void *user);
 
 // join two pvector into one, pvec1 should free the joined element in pvec2
 RZ_API bool rz_pvector_join(RZ_NONNULL RzPVector *pvec1, RZ_NONNULL RzPVector *pvec2);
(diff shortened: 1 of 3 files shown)


librz/util/vector.c
```diff
@@ -462,6 +462,26 @@ RZ_API RZ_BORROW void **rz_pvector_find(RZ_NONNULL const RzPVector *vec, RZ_NONN
...
+
+	void **iter = NULL;
+	size_t i = 0;
+	rz_pvector_enumerate (vec, iter, i) {
+		if (!cmp(value, *iter, user)) {
+			return i;
+		}
+	}
+	return SZT_MAX;
+}
```
wargio, line 483: how about returning bool and the index as arg?

wargio (changes requested):
These changes are welcome, but why do you need this?

librz/util/vector.c
notxvilka: Ideally - yes. But it's better to be consistent - e.g. `rz_pvector_find()` also returns the element, not `bool`. I think this one could be merged as is, but *both* functions changed at the same time in a different PR.

SSharshunov:
> These changes are welcome, but why do you need this?

for another PR, there is a replacement of list with vectors
Now still prepared, one more function (unique values) will be added in a separate PR

librz/include/rz_vector.h
```diff
@@ -314,6 +314,9 @@ RZ_API void **rz_pvector_contains(RzPVector *vec, const void *x);
 // find the element in the vec based on cmparator
 RZ_API RZ_BORROW void **rz_pvector_find(RZ_NONNULL const RzPVector *vec, RZ_NONNULL const void *element, RZ_NONNULL RzPVectorComparator cmp, void *user);
 
+// find the element index in the vec based on cmparator
+RZ_API RZ_BORROW size_t rz_pvector_find_index(RZ_NONNULL const RzPVector *vec, RZ_NONNULL const void *element, RZ_NONNULL RzPVectorComparator cmp, void *user);
```
Rot127:
```suggestion
RZ_API size_t rz_pvector_find_index(RZ_NONNULL const RzPVector *vec, RZ_NONNULL const void *element, RZ_NONNULL RzPVectorComparator cmp, void *user);
```

librz/util/vector.c
```diff
@@ -462,6 +462,26 @@ RZ_API RZ_BORROW void **rz_pvector_find(RZ_NONNULL const RzPVector *vec, RZ_NONN
 	return NULL;
 }
 
+/**
+ * \brief Find the \p element in the \p vec
+ * \param vec the RzPVector to search in
+ * \param value the value that elements in pvector compare against by \p cmp
+ * \param cmp the comparator function
+ * \return the index of the element if found, SZT_MAX otherwise
```
Rot127: Also document what happens with duplicates.

Rot127:
```suggestion
/**
 * \brief Find the \p element in the \p vec.
 * \param vec the RzPVector to search in.
 * \param value the value that elements in pvector compare against by \p cmp.
 * \param cmp the comparator function.
 * \return the index of the element if found, SZT_MAX otherwise.
```

Rot127 (changes requested):

librz/util/vector.c
SSharshunov: I didn't understand your question

notxvilka: He probably meant that you need to write something like "Returns the index of the first matching element"

notxvilka (approved):

PR #5739, opened 2026-01-06
Fix leaks of RzILMem buffers and clarify ownership of buffer.

Rot127:

**Detailed description**

The previous assumption of `RzILMem` was that the buffer is not owned by it. This made sense, because in the beginning it only held an `RzIO` buffer.

But the Xtensa and Sparc `RzAnalysisILInitCallback` definition initializes an additional `RzILMem` object. This additional `RzILMem` has a sparse buffer in both cases. The Xtensa and Sparc plugin pass the sparse buffer to `RzILMem` with the assumption that it is owned. They have to, because there is no `fini()` version of `RzAnalysisILInitCallback` in which they could free the buffer. There are also no API functions to free the buffer in `RzILMem`. So the plugins don't have a nice way to clean it up in a theoretical `fini()` callback.

This commit fixes the leak by adding a flag to `rz_il_mem_new()` to mark the ownership of the given buffer.
This seems the most natural solution, because buffers can abstract from all kind of backends (`RzIO`, a file, some memory etc.). Each time the ownership of the buffer is different. So the creator of `rz_il_mem_new()` can decide what buffer with what ownership they pass.

**Test plan**

All green no crashes.

**Closing issues**

...

### DIFF
--- a/librz/il/definitions/mem.c
+++ b/librz/il/definitions/mem.c
@@ -7,15 +7,7 @@
 
 #define KEY_LEN_MAX 64 // because RzBuffer uses ut64 addresses
 
-/**
- * Create a memory for accessing the given buffer.
- *
- * \param buf The buffer of the memory.
- * \param key_len The number of bits a memory key requires.
- *
- * \return The new RzILMem object, or NULL in case of failure.
- */
-RZ_API RZ_OWN RzILMem *rz_il_mem_new(RZ_NONNULL RZ_BORROW RzBuffer *buf, ut32 key_len) {
+static RzILMem *mem_new(RZ_NONNULL RzBuffer *buf, ut32 key_len, bool take_buf_ownerhip) {
 	rz_return_val_if_fail(buf && key_len, NULL);
 	if (key_len > KEY_LEN_MAX) {
 		// no assertion because it's not stricly a programming error to call this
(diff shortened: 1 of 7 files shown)


librz/il/definitions/mem.c
```diff
@@ -12,10 +12,11 @@
  *
  * \param buf The buffer of the memory.
  * \param key_len The number of bits a memory key requires.
+ * \param take_buf_ownerhip If set, RzILMem takes ownership of the buffer.
  *
  * \return The new RzILMem object, or NULL in case of failure.
  */
-RZ_API RZ_OWN RzILMem *rz_il_mem_new(RZ_NONNULL RZ_BORROW RzBuffer *buf, ut32 key_len) {
+RZ_API RZ_OWN RzILMem *rz_il_mem_new(RZ_NONNULL RzBuffer *buf, ut32 key_len, bool take_buf_ownerhip) {
```
notxvilka: I wonder if it's better to split on the API level instead:
```c
RZ_API RZ_OWN RzILMem *rz_il_mem_new_from_owned(RZ_NONNULL RzBuffer *buf, ut32 key_len)
RZ_API RZ_OWN RzILMem *rz_il_mem_new_from_borrowed(RZ_NONNULL RzBuffer *buf, ut32 key_len)
```
or something like that.

@thestr4ng3r @wargio what do you think?

notxvilka (commented):
LGTM with one unclear thing

librz/il/definitions/mem.c
wargio: yes, split the API.

notxvilka (approved):
LGTM

wargio (approved):

notxvilka (approved):

Issue #4823, opened 2025-01-04
No-name symbols should have a different color


wargio:
Sometimes elf symbols do not have names (as per spec), but rizin (since https://github.com/rizinorg/rizin/pull/4821) will autoname these.

These should have a different color to ensure the user knows that the name has been autogenerated by rizin itself.

prathmesh703:
@wargio I would like to work on this issue . Can you assign this issue to me?

wargio:
No need to get it assigned to you. Just work on it (nobody is) and open a PR once is "ready" to be reviewed

Sl4y3r-07:
@wargio can i colorise the full output of the `is` command?

wargio:
this is not linked to the `is` command but to the flags (RzFlag) that are used in the code.
This is not a simple issue to fix since these flags are linked to the RzCore printer

Sl4y3r-07:
Ohh, I thought to do it like this..
<img width="1108" height="310" alt="Image" src="https://github.com/user-attachments/assets/e20e4962-c01d-4009-b5ea-011e3d9cb2e6" />

wargio:
given the comment from @notxvilka i just read on the `oml` PR from another user, i would say you can colorize those in a similar way. 

https://github.com/rizinorg/rizin/pull/5704

Sl4y3r-07:
yeah, i read his comment. Can I make the PR for your review?

wargio:
sure, open it

PR #5741, opened 2026-01-06
Add color support to no-name symbols #4823

Sl4y3r-07:

**Detailed description**
This minor change is done to resolve the issue.
```
if (value && rz_str_startswith(value, "unknown_")) {
			return ctx->pal.diff_unknown;
		}
```
image

...

...

### DIFF
--- a/librz/bin/bobj_process_symbol.c
+++ b/librz/bin/bobj_process_symbol.c
@@ -3,7 +3,6 @@
 // SPDX-License-Identifier: LGPL-3.0-only
 #include <rz_bin.h>
 #include "i/private.h"
-#include <rz_core.h>
 
 static void process_objc_symbol(RzBinObject *o, RzBinSymbol *symbol) {
 	if (!symbol->classname) {
@@ -86,27 +85,15 @@ static void process_handle_symbol(RzBinSymbol *symbol, RzBinObject *o, const RzD
 	// methods and fields.
 	language_cb(o, symbol);
 }
-RZ_API RzBinSymbol *rz_core_bin_get_symbol_by_name(RZ_NONNULL const RzCore *core, const char *name) {
-	rz_return_val_if_fail(core && name, NULL);
-	RzBinObject *obj = rz_bin_cur_object(core->bin);
-	if (!obj || !obj->symbols) {
-		return NULL;
-	}
-	void **it;
-	RzBinSymbol *element;
-	rz_pvector_foreach (obj->symbols, it) {
-		element = *it;
-		if (RZ_STR_EQ(element->name, name)) {
-			return element;
-		}
-	}
-	return NULL;
-}
 
 RZ_IPI void rz_bin_normalize_symbol(RzBinSymbol *symbol) {
 	rz_return_if_fail(symbol);
 	if (RZ_STR_ISEMPTY(symbol->name)) {
-		symbol->name = (symbol->vaddr && symbol->vaddr != UT64_MAX) ? rz_str_newf("unknown_0x%" PFMT64x, symbol->vaddr) : rz_str_newf("unknown_%d", symbol->ordinal);
+		if (symbol->vaddr && symbol->vaddr != UT64_MAX) {
+			symbol->name = rz_str_newf("unknown_0x%" PFMT64x, symbol->vaddr);
+		} else {
+			symbol->name = rz_str_newf("unknown_%d", symbol->ordinal);
+		}
 		symbol->is_auto_generated = true;
 	} else {
 		symbol->is_auto_generated = false;
@@ -122,6 +109,9 @@ RZ_IPI void rz_bin_process_symbols(RzBinFile *bf, RzBinObject *o, const RzDemang
 	ht_sp_free(o->import_name_symbols);
 	o->import_name_symbols = ht_sp_new(HT_STR_DUP, NULL, NULL);
 
+	ht_sp_free(o->name_to_symbol);
+	o->name_to_symbol = ht_sp_new(HT_STR_DUP, NULL, NULL);
+
 	RzBinProcessLanguage language_cb = rz_bin_process_language_symbol(o);
 
 	void **it;
@@ -129,6 +119,9 @@ RZ_IPI void rz_bin_process_symbols(RzBinFile *bf, RzBinObject *o, const RzDemang
 	rz_pvector_foreach (o->symbols, it) {
 		element = *it;
 		rz_bin_normalize_symbol(element);
+		if (RZ_STR_ISNOTEMPTY(element->name) && !ht_sp_find(o->name_to_symbol, element->name, NULL)) {
+			ht_sp_insert(o->name_to_symbol, element->name, element);
+		}
 		process_handle_symbol(element, o, demangler, flags, language_cb);
 	}
 }
--- a/librz/bin/bobj_process_symbol.c
+++ b/librz/bin/bobj_process_symbol.c
@@ -3,6 +3,7 @@
 // SPDX-License-Identifier: LGPL-3.0-only
 #include <rz_bin.h>
 #include "i/private.h"
+#include <rz_core.h>
 
 static void process_objc_symbol(RzBinObject *o, RzBinSymbol *symbol) {
 	if (!symbol->classname) {
@@ -85,6 +86,33 @@ static void process_handle_symbol(RzBinSymbol *symbol, RzBinObject *o, const RzD
 	// methods and fields.
 	language_cb(o, symbol);
 }
+RZ_API RzBinSymbol *rz_core_bin_get_symbol_by_name(RZ_NONNULL const RzCore *core, const char *name) {
+	rz_return_val_if_fail(core && name, NULL);
+	RzBinObject *obj = rz_bin_cur_object(core->bin);
+	if (!obj || !obj->symbols) {
+		return NULL;
+	}
+	void **it;
+	RzBinSymbol *element;
+	rz_pvector_foreach (obj->symbols, it) {
+		element = *it;
+		if (RZ_STR_EQ(element->name, name)) {
+			return element;
+		}
+	}
+	return NULL;
+}
+
+RZ_IPI void rz_bin_normalize_symbol(RzBinSymbol *symbol) {
+	rz_return_if_fail(symbol);
+	if (RZ_STR_ISEMPTY(symbol->name)) {
+		symbol->name = (symbol->vaddr && symbol->vaddr != UT64_MAX) ? rz_str_newf("unknown_0x%" PFMT64x, symbol->vaddr) : rz_str_newf("unknown_%d", symbol->ordinal);
+		symbol->is_auto_generated = true;
+	} else {
+		symbol->is_auto_generated = false;
+	}
+	return;
+}
 
 RZ_IPI void rz_bin_process_symbols(RzBinFile *bf, RzBinObject *o, const RzDemanglerPlugin *demangler, RzDemanglerFlag flags) {
 	if (rz_pvector_len(o->symbols) < 1) {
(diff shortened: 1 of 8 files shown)


librz/core/cmd/cmd_api.c
```diff
@@ -2565,6 +2565,9 @@ static const char *core_cmd_default_table_color(const char *value, const char *c
 	} else if (table_value_is_flags(column)) {
 		return ctx->pal.flag;
 	} else if (table_value_is_name(column)) {
+		if (value && rz_str_startswith(value, "unknown_")) {
```
wargio: probably is worth checking what the bin plugins do when these symbols are "empty" and create a function which normalizes this.

wargio (changes requested):
As mentioned, maybe is worth checking all the bin plugins for these usecases in symbols and maybe is worth adding a configurable "prefix" variable to RzBin which can then be reused also here

@notxvilka what do you think?

wargio:
(ah, the changes are ok)

notxvilka:
> As mentioned, maybe is worth checking all the bin plugins for these usecases in symbols and maybe is worth adding a configurable "prefix" variable to RzBin which can then be reused also here
> 
> @notxvilka what do you think?

What if there is a symbol called `unknown_17`?
Configurable prefix sounds right, I think. Ideally though, we should have a better method to distinguish those rather than a simple string comparison.

Sl4y3r-07:
> What if there is a symbol called `unknown_17`?
> Configurable prefix sounds right, I think. Ideally though, we should have a better method to distinguish those rather than a simple string comparison.

Just a suggestion, as in some debugger tools like IDA the symbols naming is done like `sub_<address>` `loc_<address>`
So, may be like this way we can have these symbols naming

And yes we can come up with a better method instead of just string comparison.

wargio:
> > What if there is a symbol called `unknown_17`?
> > Configurable prefix sounds right, I think. Ideally though, we should have a better method to distinguish those rather than a simple string comparison.
> 
> Just a suggestion, as in some debugger tools like IDA the symbols naming is done like `sub_<address>` `loc_<address>` So, may be like this way we can have these symbols naming
> 
> And yes we can come up with a better method instead of just string comparison.

true, but if you check the address where that symbol is located, you will notice it might clash with others. the logic should be like: is it a valid vaddr? if not maybe just use an index value

Sl4y3r-07:
> true, but if you check the address where that symbol is located, you will notice it might clash with others. the logic should be like: is it a valid vaddr? if not maybe just use an index value

yeah, I will go through other bin plugins as well for the mentioned usecase and add a function.

Sl4y3r-07:
Sorry for my inactivity due to some unavoidable circumstances. 
I went through all the bin plugins code and checked how they are handling the symbol names when they are "empty".
Some plugins are not handling the "empty" cases as handled in the elf one. For e.g, in `smd plugin`
```
static void addsym(RzPVector /*<RzBinSymbol *>*/ *ret, const char *name, ut64 addr) {
	RzBinSymbol *ptr = RZ_NEW0(RzBinSymbol);
	if (!ptr) {
		return;
	}
	ptr->name = rz_str_dup(name ? name : "");
	ptr->paddr = ptr->vaddr = addr;
	ptr->size = 0;
	ptr->ordinal = 0;
	rz_pvector_push(ret, ptr);
}
```

For `mach0` plugin, 
`ptr->name = rz_str_dup((char *)syms[i].name);`

Similar is the case with other plugins too. Shall I proceed with handling the `empty` cases in these plugins as in `elf` plugin and then creating a function normalizing it?

wargio:
create an RzBin function to normalize this and we could allow to have `NULL` names and since we iterate over them in the RzBin code, we can fix them there.

https://github.com/rizinorg/rizin/blob/dev/librz/bin/bobj_process_symbol.c#L101

Sl4y3r-07:
@wargio I have made the changes, kindly review them. Also, in `elf_relocs_conversion.c` 
I have removed that `unknown_%d` string formation from there, as normalising function is handling that, but because of this leak related some test has failed. 

And, names are getting generated (as you suggested)
```
symbol->name = (symbol->vaddr && symbol->vaddr != UT64_MAX) ? rz_str_newf("unknown_0x%" PFMT64x, symbol->vaddr) : rz_str_newf("unknown_%d", symbol->ordinal);
		symbol->is_auto_generated = true;
```

Sl4y3r-07:
@wargio according to your convenience, please review the changes.
If the changes are fine, I'll fix the broken testcases

librz/bin/bobj_process_symbol.c
```diff
@@ -3,6 +3,7 @@
 // SPDX-License-Identifier: LGPL-3.0-only
 #include <rz_bin.h>
 #include "i/private.h"
+#include <rz_core.h>
```
wargio: this is not ok

librz/bin/bobj_process_symbol.c
```diff
@@ -85,6 +86,33 @@ static void process_handle_symbol(RzBinSymbol *symbol, RzBinObject *o, const RzD
 	// methods and fields.
 	language_cb(o, symbol);
 }
+RZ_API RzBinSymbol *rz_core_bin_get_symbol_by_name(RZ_NONNULL const RzCore *core, const char *name) {
```
wargio: core bin stuff cannot go in RzBin

wargio (changes requested):

librz/bin/bobj_process_symbol.c
```diff
@@ -85,6 +86,33 @@ static void process_handle_symbol(RzBinSymbol *symbol, RzBinObject *o, const RzD
...
+			return element;
+		}
+	}
+	return NULL;
+}
+
+RZ_IPI void rz_bin_normalize_symbol(RzBinSymbol *symbol) {
+	rz_return_if_fail(symbol);
+	if (RZ_STR_ISEMPTY(symbol->name)) {
+		symbol->name = (symbol->vaddr && symbol->vaddr != UT64_MAX) ? rz_str_newf("unknown_0x%" PFMT64x, symbol->vaddr) : rz_str_newf("unknown_%d", symbol->ordinal);
```
wargio: use an if else structure here, this is too messy

librz/core/cmd/cmd_api.c
```diff
@@ -2565,6 +2565,11 @@ static const char *core_cmd_default_table_color(const char *value, const char *c
 	} else if (table_value_is_flags(column)) {
 		return ctx->pal.flag;
 	} else if (table_value_is_name(column)) {
+		RzBinSymbol *sym = rz_core_bin_get_symbol_by_name(core, value);
```
wargio, line 2568: i would prefer to use rz_str_startswith

Sl4y3r-07: I used this since we discussed for a better approach instead of direct string comparison. So it fetches the symbol and check whether is_auto_generated true. If rz_str_startswith is to be used, then initial change is good enough ig.

librz/bin/bobj_process_symbol.c
Sl4y3r-07: Should I move it to `librz/core/cbin.c`?

wargio: yes

librz/core/cmd/cmd_api.c
Sl4y3r-07: Should I continue with this change?

wargio:
@notxvilka what do you think?

notxvilka (commented):
Yes, the idea is good, but I don't like that it searches symbol by name for every entry. Probably it's better to rethink logic of the flow a bit to reduce costly operations.

librz/bin/bobj_process_symbol.c
notxvilka: @Sl4y3r-07 RzBin should not have dependency on RzCore

notxvilka (changes requested):

librz/bin/bobj_process_symbol.c
Sl4y3r-07: Yeah, I have fixed that. Thanks.

Sl4y3r-07:
> Yes, the idea is good, but I don't like that it searches symbol by name for every entry. Probably it's better to rethink logic of the flow a bit to reduce costly operations.

Okay, I'll go through it and try to optimise it. Thanks.

librz/bin/bobj_process_symbol.c
notxvilka: But the line is still here...

Sl4y3r-07: I meant I have fixed it locally, will push it with other changes.

Sl4y3r-07:
@wargio @notxvilka can you please review the updated changes?

librz/bin/bobj_process_symbol.c
```diff
@@ -86,6 +86,21 @@ static void process_handle_symbol(RzBinSymbol *symbol, RzBinObject *o, const RzD
...
 }
 
+RZ_IPI void rz_bin_normalize_symbol(RzBinSymbol *symbol) {
+	rz_return_if_fail(symbol);
+	if (RZ_STR_ISEMPTY(symbol->name)) {
+		if (symbol->vaddr && symbol->vaddr != UT64_MAX) {
+			symbol->name = rz_str_newf("unknown_0x%" PFMT64x, symbol->vaddr);
+		} else {
+			symbol->name = rz_str_newf("unknown_%d", symbol->ordinal);
+		}
```
wargio, line 96:
```suggestion
		if (symbol->vaddr && symbol->vaddr != UT64_MAX) {
			symbol->name = rz_str_newf("unknown_va_0x%" PFMT64x, symbol->vaddr);
		} else if (symbol->paddr && symbol->paddr != UT64_MAX) {
			symbol->name = rz_str_newf("unknown_pa_0x%" PFMT64x, symbol->paddr);
		} else {
			symbol->name = rz_str_newf("unknown_ord_%u", symbol->ordinal);
		}
```

wargio (commented):
tbh, i'm not sure if this is the right way. i feel like we could just have a RzBin variable that defines the prefix to use for null string functions, apply it globally when a function doesn't have a valid name and do rz_str_startswith.

The current implementation (which ok/correct from logical PoV) is wasting a lot of cycles & memory.

@notxvilka @Rot127 can you give your thought about this?

notxvilka:
> tbh, i'm not sure if this is the right way. i feel like we could just have a RzBin variable that defines the prefix to use for null string functions, apply it globally when a function doesn't have a valid name and do rz_str_startswith.
> 
> The current implementation (which ok/correct from logical PoV) is wasting a lot of cycles & memory.
> 
> @notxvilka @Rot127 can you give your thought about this?

Not sure what do you mean, it should just set if the name autogenerated or not. Then, instead of doing string comparison each time it should just use this flag to set the proper color.

wargio:
> Not sure what do you mean, it should just set if the name autogenerated or not. Then, instead of doing string comparison each time it should just use this flag to set the proper color

I think the whole issue that we are trying to fix is kinda weird. it's like outputting a different color for functions that does not match a symbols,  but in this case for a symbol without a name.

The implementation uses a hash table to find if a given string is a symbol, which for me is over engineered and the memory is never released (unless you close the file).

I would have preferred a more simple approach similar to what we do for the function prefix, which we use as identifier for functions which do not have symbols.

Sl4y3r-07:
> > Not sure what do you mean, it should just set if the name autogenerated or not. Then, instead of doing string comparison each time it should just use this flag to set the proper color
> 
> I think the whole issue that we are trying to fix is kinda weird. it's like outputting a different color for functions that does not match a symbols, but in this case for a symbol without a name.
> 
> The implementation uses a hash table to find if a given string is a symbol, which for me is over engineered and the memory is never released (unless you close the file).
> 
> I would have preferred a more simple approach similar to what we do for the function prefix, which we use as identifier for functions which do not have symbols.

If that's the case, then string comparison seems to be the simpler approach i think

PR #5764, opened 2026-01-09
Sparc: Fix semantical IL bug of unconditional jumps executing NOP in delayed slot.

Rot127:

**Detailed description**

Unconditional jumps have often a nop instruction in the delayed slot.
But we should not execute it. Because it will set the PC temporarily (before the JUMP is executed) to the instruction *after* the delayed slot.
This is semantically not correct and can lead to bugs if the IL op is used in analysis.
So replace it with an EMPTY effect.

NOTE: The jump of an unconditional branch is still executed on the
PC of the delayed slot! This is semantically also not correct,
because the delayed slot instruction should never be executed for unconditional jumps.
But we need to wait for RzArch implementation and support for instruction packets
to do this properly. So I safe the time spent on a work around.

**Test plan**

Fixed tests

**Closing issues**

...


notxvilka (approved):
How did you find it?

Rot127:
The xrefs found by the prototype had the wrong `from` addresses.

Rot127:
~~`rett` seems affected as well?~~
Never mind.

librz/arch/p/analysis/analysis_sparc_cs.c
```diff
@@ -157,6 +157,21 @@ static int analyze_op(RzAnalysis *a, RzAnalysisOp *op, ut64 addr, const ut8 *buf
 
 		op->il_op = rz_sparc_cs_get_il_op(sparc->handle, insn, mode, sparc);
 		if (has_delayed_branch) {
```
wargio, line 159: i think this should not be a problem as the VM should re-arrange the code in case of delayed slots.

Rot127: The VM doesn't have a concept of delayed slots. Because it only sees IL effects, the globals and the PC.

Rot127: But with RzArch and the Instruction packet as abstraction layer we have this problem no longer anyways. Because the VM just asks for "atomically executable units" and gets the IL effect of the branch + the dalyed slot combined.

wargio: yes i know. this is an issue also on mips

Rot127: Can we still merge this? Or do you have an alternative in mind?

PR #5769, opened 2026-01-10
rz_assert: Add rz_break_if_reached/fail()

kazarmy:

**Detailed description**

This pr adds the `rz_break_if_reached()` and `rz_break_if_fail()` macros to rz_assert, and uses `rz_break_if_fail()` instead of `rz_goto_if_fail()` for exiting loops. The advantage of the former over the latter is that one does not need to figure out where to place the goto label.

Unfortunately, the `do ... while (0)` construct can't be used with `break` because it is a loop. I see 2 reasons why the `do ... while (0)`  construct is used. Both have to do with the `if ... else ...` statement as shown in https://github.com/radareorg/radare2/pull/11286#issuecomment-421272260:

1. The macro body has to be a single statement that is not an `if` statement. This is important for `if ... else ...` statements that do not use braces, since otherwise the semantics of the `if ... else ...` statement will not be as expected. Fortunately, a compound statement i.e. multiple statements enclosed with curly braces `{` `}` is considered a single statement, and that is what is used here.
2. The end-of-statement semicolon `;` has to be eaten. Unfortunately, apparently this cannot be done by macros here that have a `break` statement in their body, but the implications are only that there will be a compilation error if the macro is used in `if ... else ...` statements without curly braces as noted in https://github.com/radareorg/radare2/pull/11286#issuecomment-421272260.

Since it's unlikely in Rizin that the macros here will be used in `if ... else ...` statements without curly braces, the possible compilation error above is unlikely to be an issue, and even if there is an issue, the problematic code will not enter dev.

**Test plan**

The addition makes sense. All builds are green.

**Closing issues**

...

### DIFF
--- a/librz/bin/format/pe/pe_clr.c
+++ b/librz/bin/format/pe/pe_clr.c
@@ -51,7 +51,7 @@ RZ_OWN RzList /*<RzBinSymbol *>*/ *PE_(rz_bin_pe_get_clr_symbols)(RzBinPEObj *bi
 		Pe_image_metadata_methoddef *methoddef = *it;
 
 		if ((type_name || type_namespace) && i >= type_methods_start && i >= type_methods_end) {
-			rz_goto_if_fail(type_it, loop_end);
+			rz_break_if_fail(type_it);
 
 			// Update class and namespace
 			free(type_name);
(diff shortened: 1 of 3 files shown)


kazarmy:
Redness is unrelated

wargio (approved):

Rot127 (approved):

PR #5782, opened 2026-01-11
Fix memory leak in RzMagicLine

Ayushd785:

**Detailed description**

Fixes a memory leak where dynamically allocated strings for `ml->name` and `ml->test_string` in `magic_parse_value() ` were never freed. These strings are allocated via `malloc()` and assigned to the `RzMagicLine` structure, but `rz_magic_line_free() `was missing the corresponding `free()` calls.

**Test plan**
The change is very minimal just added a free() to fix the memory leak.Though i have Verified the fix by building locally and runing the test suits.

**Closing issues**

### DIFF
--- a/librz/magic/magic-test.c
+++ b/librz/magic/magic-test.c
@@ -92,7 +92,7 @@ int file_printf(char **s, const char *fmt, ...) {
 
 static int magic_test_line(RzMagicLine *, RzMagicState *);
 
-static RzMagicLine *magic_get_named(RzMagic *m, const char *name) {
+static RzMagicLine *magic_get_named(RzMagic *m, char *name) {
 	RzMagicLine ml;
 
 	ml.name = name;
--- a/librz/magic/magic.c
+++ b/librz/magic/magic.c
@@ -143,5 +143,7 @@ void rz_magic_line_free(RZ_OWN RZ_NULLABLE RzMagicLine *ml) {
 	free(ml->type_string);
 	free(ml->result);
 	free(ml->mimetype);
+	free(ml->name);
+	free(ml->test_string);
 	free(ml);
 }
--- a/librz/include/rz_magic.h
+++ b/librz/include/rz_magic.h
@@ -343,7 +343,7 @@ struct rz_magic_line_t {
 	char indirect_operator;
 	int64_t indirect_operand;
 
-	const char *name;
+	char *name;
 
 	enum magic_type type;
 	char *type_string;
@@ -352,7 +352,7 @@ struct rz_magic_line_t {
 
 	char test_operator;
 	int test_not;
-	const char *test_string;
+	char *test_string;
 	size_t test_string_size;
 	ut64 test_unsigned;
 	int64_t test_signed;


librz/magic/magic.c
```diff
@@ -143,5 +143,7 @@ void rz_magic_line_free(RZ_OWN RZ_NULLABLE RzMagicLine *ml) {
 	free(ml->type_string);
 	free(ml->result);
 	free(ml->mimetype);
+	free((void *)ml->name);
```
notxvilka: `(void *)` is completely unnecessary

notxvilka (changes requested):

librz/magic/magic.c
Ayushd785: Got it! I thought the cast was needed for `const char *`  but you're right, it's implicit. Will remove.

wargio: revert this. it was correct since you cannot call `free` when a pointer is `const`

Ayushd785: Okay!

notxvilka: Why did you return it?

Ayushd785: because free() expects a `non-const void *` since, `ml->name` and `ml->test_string` are declared as `const char *` in the struct, we would need the void cast to remove the const qualifier

Ayushd785: @notxvilka Sir, if you see this struct
```
struct rz_magic_line_t {
	RBNode rb;
	RzMagic *root;
	ut32 line;
	ut32 strength;
	RzMagicLine *parent;

	char strength_operator;
	ut32 strength_value;

	int text;

	int64_t offset;
	int offset_relative;

	char indirect_type;
	int indirect_relative;
	int64_t indirect_offset;
	char indirect_operator;
	int64_t indirect_operand;

	const char *name;   // declared as const char *

	enum magic_type type;
	char *type_string;
	char type_operator;
	int64_t type_operand;

	char test_operator;
	int test_not;
	const char *test_string;  // declared as const char *
	size_t test_string_size;
	ut64 test_unsigned;
	int64_t test_signed;
	double test_double;

	int stringify;
	char *result;
	char *mimetype;

	RzList *children;
};
```

Rot127: Then you need to check if it somewhere assigns a constant string or not.
Otherwise this segfaults if it tries to free the constant string.
Then remove the `const` IF it only holds heap pointers.

Ayushd785: Yes @Rot127, I think that would be a better approach as they are not truly const, they hold heap-allocated memory that we own and free.

Ayushd785:
hii @notxvilka @wargio,
 The CI is faiing becuse `magic_get_named()` in `magic-test.c:98` assigns a const string to `ml.name`, while `magic_parse_value()` assigns heap-allocated strings. I think its better to use `void *` casting at freeing. 
 I am open to your feedback/ Approach. Thank you!

Ayushd785:
I think this approach is better. The cast in `magic_get_named()` is safe because the temporary `struct ml` is only used for searching. It's on the stack and is automatically destroyed when the function returns. The `name` parameter is never freed through this temporary struct, so it's safe to cast there.

Thanks, @Rot127, for the help!

wargio:
> becuse `magic_get_named()` in `magic-test.c:98` assigns a const string to `ml.name`

you can change that function to return `char *` and if the function tries to return any actual `const char*`, you can call `rz_str_dup()` on them to always allocate on the heap.

Ayushd785:
> > becuse `magic_get_named()` in `magic-test.c:98` assigns a const string to `ml.name`
> 
> you can change that function to return `char *` and if the function tries to return any actual `const char*`, you can call `rz_str_dup()` on them to always allocate on the heap.

got your point but isn't the most recent change is a valid approach?

wargio:
> got your point but isn't the most recent change is a valid approach?

Ok, i checked the code and i think is partially correct. i would change the function magic_get_named to use `char *` as argument.

So according to the code, `name` is always allocated via `malloc()` (ignore `copy` which is used in case of fail to free the allocated memory)

https://github.com/rizinorg/rizin/blob/c594b413ffc43a42e0193b93c3fb8dd9a7b7bae4/librz/magic/magic-load.c#L840-L853

but (always according to the code) then the value is used by `magic_get_named()` to search inside the rbtree.

https://github.com/rizinorg/rizin/blob/c594b413ffc43a42e0193b93c3fb8dd9a7b7bae4/librz/magic/magic-test.c#L95-L109

Since the `RzMagicLine` inside the `magic_get_named` is never really used anywhere, then the value is safe to never be freed in this function.

**So, TLDR: change the definition of `magic_get_named` to use `char *`, remove the comment you added and we can probably merge it.**

```c
static RzMagicLine *magic_get_named(RzMagic *m, char *name) {
``

**Comment by Ayushd785** [comment]
> > got your point but isn't the most recent change is a valid approach?
> 
> Ok, i checked the code and i think is partially correct. i would change the function magic_get_named to use `char *` as argument.
> 
> So according to the code, `name` is always allocated via `malloc()` (ignore `copy` which is used in case of fail to free the allocated memory)
> 
> https://github.com/rizinorg/rizin/blob/c594b413ffc43a42e0193b93c3fb8dd9a7b7bae4/librz/magic/magic-load.c#L840-L853
> 
> but (always according to the code) then the value is used by `magic_get_named()` to search inside the rbtree.
> 
> https://github.com/rizinorg/rizin/blob/c594b413ffc43a42e0193b93c3fb8dd9a7b7bae4/librz/magic/magic-test.c#L95-L109
> 
> Since the `RzMagicLine` inside the `magic_get_named` is never really used anywhere, then the value is safe to never be freed in this function.
> 
> **So, TLDR: change the definition of `magic_get_named` to use `char *`, remove the comment you added and we can probably merge it.**
> 
> ```c
> static RzMagicLine *magic_get_named(RzMagic *m, char *name) {
> ``
> ```

Alright, I got your point and learned something new. That was the most detailed explanation one could ever get. Thank you for the explanation. I’ll make the changes soon.

File: librz/magic/magic-test.c [hunk 2]
```diff
@@ -95,7 +95,11 @@ static int magic_test_line(RzMagicLine *, RzMagicState *);
 static RzMagicLine *magic_get_named(RzMagic *m, const char *name) {
```
wargio:
```suggestion
static RzMagicLine *magic_get_named(RzMagic *m, char *name) {
```

librz/magic/magic-test.c
```diff
@@ -95,7 +95,11 @@ static int magic_test_line(RzMagicLine *, RzMagicState *);
...
 	RzMagicLine ml;
 
-	ml.name = name;
+	/*
+	 * Casting away const is safe here: ml is a temporary struct used
+	 * only for searching the tree and is never freed.
+	 */
+	ml.name = (char *)name;
```
wargio:
```suggestion
	ml.name = name;
```

librz/magic/magic-test.c
Ayushd785: Implemented

librz/magic/magic-test.c
Ayushd785: Implemented

wargio (approved):

Rot127 (approved):

PR #5786, opened 2026-01-12
rz_assert: Order macros consistently

kazarmy:

**Detailed description**

This pr just reorders the rz_assert macros in a consistent manner to reduce confusion.

**Test plan**

**Closing issues**

...

### DIFF
--- a/librz/include/rz_util/rz_assert.h
+++ b/librz/include/rz_util/rz_assert.h
@@ -26,18 +26,18 @@ RZ_API void rz_assert_log(RzLogLevel level, const char *fmt, ...) RZ_PRINTF_CHEC
 #define RZ_FUNCTION ((const char *)("???"))
 #endif
 
-#define rz_warn_if_reached() \
-	do { \
-		rz_assert_log(RZ_LOGLVL_WARN, "(%s:%d):%s%s code should not be reached\n", \
-			__FILE__, __LINE__, RZ_FUNCTION, RZ_FUNCTION[0] ? ":" : ""); \
-	} while (0)
-
 #define rz_warn_if_fail(expr) \
 	do { \
 		if (!(expr)) { \
 			rz_assert_log(RZ_LOGLVL_WARN, "(%s:%d):%s%s runtime check failed: (%s)\n", \
 				__FILE__, __LINE__, RZ_FUNCTION, RZ_FUNCTION[0] ? ":" : "", #expr); \
 		} \
+	} while (0)
+
+#define rz_warn_if_reached() \
+	do { \
+		rz_assert_log(RZ_LOGLVL_WARN, "(%s:%d):%s%s code should not be reached\n", \
+			__FILE__, __LINE__, RZ_FUNCTION, RZ_FUNCTION[0] ? ":" : ""); \
 	} while (0)
 
 /*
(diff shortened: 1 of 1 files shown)


librz/include/rz_util/rz_assert.h
```diff
@@ -71,23 +71,23 @@ RZ_API void rz_assert_log(RzLogLevel level, const char *fmt, ...) RZ_PRINTF_CHEC
 		return (val); \
 	} while (0)
 
+#define rz_goto_if_fail(expr, where) \
+	do { \
+		; \
+	} while (0)
```
wargio, line 77: if these produce empty statements, why not just declare them as `#define foo(a, b)`

kazarmy: it's so that the end-of-statement semicolon is not forgotten

Rot127: More fundamental question  (not just this PR).

I don't think it is good if the control flow changes just because `RZ_CHECKS_LEVEL == 0`. This macro should still do the `goto`.

Also above with `rz_return_if_fail`.

In Capstone I just had a bug due to this (assertion macros changed control flow in release build).

kazarmy: `RZ_CHECKS_LEVEL == 0` is not the default ... it's there if anyone wants to super-optimize the code ... there is a build (`linux-meson-clang-tests`) that runs tests with `RZ_CHECKS_LEVEL == 0`

kazarmy: > In Capstone I just had a bug due to this (assertion macros changed control flow in release build).

As per #5102, `RZ_CHECKS_LEVEL == 0` won't be the default for release builds in the foreseeable future.

kazarmy: Disregarding `expr` entirely is also useful for detecting code that has side effects in `expr`.

Rot127: So if we follow the logic, shouldn't all those macros evaluate to `;`?

kazarmy: Yes to all of the `_if_fail` macros, since `RZ_CHECKS_LEVEL == 0` assumes the check will not fail, ever (it's basically an assertion and checks in assertions should not fail, ever). All of the `_if_reached` macros can evaluate to `;` but then control flow will change and that's a headache that's not needed.

Rot127: Just to make sure we don't talk past each other.
`RZ_CHECKS_LEVEL == 0` is not part of the release, which is of course good.

But should or shouldn't `RZ_CHECKS_LEVEL == 0` change the control flow?
If not, then this line here needs a `goto`. If it should, then all the others need their `goto`s and `return`s removed. Right?

kazarmy: `RZ_CHECKS_LEVEL == 0` is about not doing any checking at all (it's at level 0). For example, the basic form of `rz_goto_if_fail()` is:

```c
do { \
	if (!(expr)) { \
		goto where; \
	} \
} while (0)
```

At `RZ_CHECKS_LEVEL == 0`, the `expr` check doesn't happen, so the above reduces to:

```c
do { \
	; \
} while (0)
```

The `_if_reached` macros don't do any checking, so their basic form stays the same.

wargio: > it's so that the end-of-statement semicolon is not forgotten

ok, ignore me then.

notxvilka (approved):

PR #5805, opened 2026-01-16
RzMemory for mapped IO interactions.

wargio:

**Detailed description**

Should replace the whole RzIO layer and most of the RzIO handlers should be translated in RzBuffer equivalents


librz/include/rz_memory/rz_memory.h
```diff
@@ -0,0 +1,44 @@
...
+#include <rz_util.h>
+#include <rz_memory/rz_memory_map.h>
+#include <rz_memory/rz_memory_region.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+typedef enum {
+	RZ_MEMORY_VISIBILITY_HIDDEN = 0,
```
Rot127, line 18: hidden means: region is completely covered by another memory region?

wargio: yes

Rot127:
One requirement I would have is for the behavior of reading. It should be configurable. By passing a flag or something.

```c
enum {
  FIXED_BLOCK, ///< Reads the requested number of bytes. If it reads less than the request number of bytes, the read fails.
  UP_TO, ///< Reads up to the number of requested bytes. If it reads less than that, the read still succeeds.
  AT_LEAST, ///< The number of requested bytes is the minimum bytes which must be read. Otherwise the read fails.
} RzMemReadBehvior;
```

I am so so annoyed about hour inconsistent and complicated reading behavior.
Having control over it would be a blessing.

Rot127:
Also, keep in mind that RzBuffer [can't properly handle](https://github.com/rizinorg/rizin/issues/5806) addresses above `ST64_MAX`.

wargio:
> Also, keep in mind that RzBuffer [can't properly handle](https://github.com/rizinorg/rizin/issues/5806) addresses above `ST64_MAX`.

i forgot about that!

wargio:
> One requirement I would have is for the behavior of reading. It should be configurable

i can replace the `bool sparse` arg with this enum.

wargio:
isn't `FIXED_BLOCK` and `AT_LEAST` the same? because otherwise i would need to define an extra variable to say `min x bytes` and `max x bytes`

Rot127:
> isn't FIXED_BLOCK and AT_LEAST the same? because otherwise i would need to define an extra variable to say min x bytes and max x bytes

Ah yes, sorry. It was late while writing this.

PR #5817, opened 2026-01-18
Make 6502 magic constant configurable

### LINKED ISSUE #5798: Make 6502 magic constant configurable
In the 6502 architecture, certain illegal opcodes rely on a "magic constant" that is highly unpredictable. It often depends on factors like time and temperature.

Some common values are `0xee` `0xff` `0x00`
Currently, this value is hardcoded to `0xee` in `librz/arch/isa/6502/6502_il.inc`

Would like to make this a config variable that the user can set.

Keeping it hardcoded restricts the accuracy for specific edge cases.

**Additional context**
This feature was discussed during the implementation of [PR #5748](https://github.com/rizinorg/rizin/pull/5748) [here](https://github.com/rizinorg/rizin/pull/5748#discussion_r2679104151)
Documentation is available at https://csdb.dk/release/?id=248511

historicattle:

**Detailed description**
- Fixed incorrect logic for the opcode `ane #$ff`
- Added a new config to set the magic number used by 2 opcodes, `ane #$ff` and `lax #$ff`. Default value is set to `0xee`
- Added and corrected appropriate tests

**Test plan**
- Corrected the test for `ane #$ff` in `test/db/asm/6502`
- Added a test in `test/db/rzil/6502`

**Closing issues**
Closes #5798

### DIFF
--- a/librz/arch/p/analysis/analysis_6502.c
+++ b/librz/arch/p/analysis/analysis_6502.c
@@ -19,7 +19,7 @@
 #include "snes/snes_op_table.h"
 #include "6502/6502_il.inc"
 #include <6502/6502dis.h>
-#include <rz_core.h>
+#include <rz_config.h>
 
 enum {
 	_6502_FLAGS_C = (1 << 0),
--- a/librz/arch/p/analysis/analysis_6502.c
+++ b/librz/arch/p/analysis/analysis_6502.c
@@ -843,9 +843,13 @@ static int _6502_op(RzAnalysis *analysis, RzAnalysisOp *op, ut64 addr, const ut8
 		op->cycles = 2;
 		op->size = 2;
 		if (mask & RZ_ANALYSIS_OP_MASK_IL) {
+			ut8 magic = 0xee;
 			_6502ILAddr addr;
 			_6502_il_immediate(&addr, data[1]);
-			op->il_op = _6502_il_op_laximm(&addr, (ut8)rz_config_get_i(cfg_state->cfg, "plugins.6502.magic"));
+			if (cfg_state && cfg_state->cfg) {
+				magic = (ut8)rz_config_get_i(cfg_state->cfg, "plugins.6502.magic");
+			}
+			op->il_op = _6502_il_op_laximm(&addr, magic);
 		}
 		break;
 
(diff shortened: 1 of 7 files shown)


historicattle:
I need to look into these errors further. <del>They pass `rz_return_val_if_fail(analysis, NULL)` but fail on `assert(rz_analysis && "This function can only be used if RzAnalysis and RzAsm were set up before.")`, which should be logically equivalent. And, these tests seem to pass locally for me. </del>

### Edit: 
They pass `rz_return_val_if_fail(analysis, NULL)` but fail on `assert(rasm && "This function can only be used if RzAnalysis and RzAsm were set up before.")`.

But according to [this](https://github.com/rizinorg/rizin/pull/4667#issue-2583916960) "The assumption made here is that RzAsm is always initialized. Also without any RzAnalysis. But RzAnalysis is never initialized without a companion RzAsm.
Hence a plugin requiring a common state, always accesses RzAsm->plugin_data, no matter if it was called from RzAnalysis or RzAsm."

Since they pass `rz_return_val_if_fail(analysis, NULL)` , doesnt this break that assumption? And because of this I wont be able to use the pointer sharing hacks. (The hexagon plugin seems to work fine with these hacks since its asm tests dont require any IL validation)

@Rot127 How do i proceed?

librz/arch/p/asm/asm_6502.c
```diff
@@ -9,17 +9,70 @@
...
+	SETICB("plugins.6502.magic", 0xee, &_6502_cfg_set, "Determines the magic number certain illegal opcodes use.");
+	*user = state;
+	return true;
+}
+
+void _6502State_fini(_6502State *state) {
+	if (!state) {
+		return;
+	}
+	rz_config_free(state->cfg);
```
Rot127, line 57: state itself is not freed. Leaking here.

Rot127 (changes requested):
The asm tests don't set up an RzAnalysis instance. You need to handle these exceptions.
Simply pass a default value if `cfg` is not set up.

librz/arch/p/asm/asm_6502.c
historicattle: This would cause a double free, since  `_6502_fini(void *user)` frees `user` after calling `_6502State_fini(user)`.

Rot127: Ah sorry, I missed it in the lines below

historicattle:
All tests are green now. Sorry for tagging you in the edited comment, i got really confused about why it was going wrong

Rot127 (approved):

historicattle:
@wargio @kazarmy Could you also please review this?

librz/arch/p/analysis/analysis_6502.c
```diff
@@ -18,6 +18,8 @@
 #include <rz_analysis.h>
 #include "snes/snes_op_table.h"
 #include "6502/6502_il.inc"
+#include <6502/6502dis.h>
+#include <rz_core.h>
```
wargio: this is not ok. if you want to make it configurable just do like the hexagon code does.
and you will need to only include `rz_config.h`

historicattle: Alright, i will make the changes. But why is including `rz_core.h` not appropriate?

Rot127: Because dependencies should be kept minimal. And you don't use anything from rz_core.h here. Only from rz_config.h

historicattle: Okay makes sense

wargio (approved):

Issue #5806, opened 2026-01-16
`RzBuffer` can't handle offsets bigger `ST64_MAX`


Rot127:
### Work environment

| Questions                                            | Answers
|------------------------------------------------------|--------------------
| OS/arch/bits (mandatory)                             | ...
| File format of the file you reverse (mandatory)      | ...
| Architecture/bits of the file (mandatory)            | ...
| `rizin -v` full output, **not truncated** (mandatory)  | all

### Expected behavior

`RzBuffer` especially the one having `RzIO` as a backend should be able to read beyond
the `ST64_MAX` offset.

### Actual behavior

It can't. The `RzBuffer` `IPI` uses `st64` as offset type (likely to be compatible with unix) which makes these accesses impossible.

Instead the error conditions should be returned over a bool, and the number of bytes read returned over a parameter.

### Steps to reproduce the behavior

https://github.com/rizinorg/rizin/blob/3c44884a9bc750ca55d96faf5f5bc75f1b000de6/librz/include/rz_util/rz_buf.h#L20-L29

### Additional Logs, screenshots, source code,  configuration dump, ...

...

Issue #4386, opened 2024-03-25
Make `ht` and buffer reading faster


Rot127:
**Is your feature request related to a problem? Please describe.**

In the `callgrind` of https://github.com/rizinorg/rizin/issues/4385 one can see that 30% execution time are due to `ht_insert` and 50% due to `rz_buf_read_at`.
Since these functions are potentially used very often in algorithms (in this example debug symbol processing), they should be more efficient if possible.

**Describe the solution you'd like**

Check `ht_insert` and the `read` component of the IO layer for inefficiencies.
At least for the hash table it might be even worth considering to replace it with another HT model.
[This talk](https://www.youtube.com/watch?v=ncHmEUmJZf4) is for C++, but maybe worth a watch. 

**Describe alternatives you've considered**

Doing nothing and just try to minimize the buffer reading and ht usage. But this is not really practical and hard to code.

**Additional context**

None

Rot127:
@antonangeloff I you are interested in more performance stuff, you can check this one out.

PR #5819, opened 2026-01-19
Improve performance of `mem.c::read_n_bits()`

### LINKED ISSUE #5796: Optimize `mem.c::read_n_bits_into()`

RzIL instructions read a lot of memory.
For that, `LOAD` and `LOADW` usually query the IO layer over an `RzBuf` object.
Then some buffer is allocated into which the IO layer copies from the binary.
Then the buffer is moved/copied to an `RzBitVector` and lastly used by the IL VM (see `read_n_bits_into`).

So we have in total one allocation (the temporary buffer) and two copies (binary -> temp buffer, temp buffer -> bit vector).

These random access moves and copies are a significant part of the runtime.
They could be reduced to the absolute minimum of one copy.

Somehow the data should be copied directly into the `RzBitVector` buffer. Without the need for a temporary buffer. The bitvector then just needs to swap endianness if required.

Only more complex stuff, like refactoring RzBitVector to hold a transparent pointer into the binary. But this is over-engineering at its finest.

**Additional context**

antonangeloff:

**Detailed description**

This PR adds alternative implementation of `read_n_bits()` (main part of `rz_il_mem_loadw()`) which avoids double copy through a temporary buffer, and also avoids the heap allocation of that buffer.

Note: the PR adds a dependency between `rz_bitvector.h` and `rz_buffer.h` because the new setter functions `rz_bv_set_from_buffer_xx()` take RzBuffer as parameter. If the dependency needs to be avoided the new functions could be moved to a separate file, e.g. `rz_bitvector_utils.h` or `rz_bitvector_helpers.h`.

Reading a 64-bit little-endian value seems to be improved by `~6.5` times and `~19` times for big-endian (x64 host environment). Larger reads of 256-bit improved above `30` times. 

`read_n_bits_into()` before:
*benchmark to be removed before merge because the function is not yet part of the `dev` branch*
```
Benchmark                                       Iterations Total time [ms] Average time [us/op] Throughput [ops/sec] 
---------------------------------------------------------------------------------------------------------------------
bench_read_n_bits_into: big endian (64 bit)        1000000      674.402000             0.674402       1482795.128128
bench_read_n_bits_into: little endian (64 bit)     1000000      206.953000             0.206953       4832014.998575
bench_read_n_bits_into: big endian (256 bit)       1000000     2474.733000             2.474733        404083.996132
bench_read_n_bits_into: little endian (256 bit)    1000000     1960.925000             1.960925        509963.410125
```

`read_n_bits_into()` after:
```
Benchmark                                       Iterations Total time [ms] Average time [us/op] Throughput [ops/sec] 
---------------------------------------------------------------------------------------------------------------------
bench_read_n_bits_into: big endian (64 bit)        1000000       35.094000             0.035094      28494899.413005
bench_read_n_bits_into: little endian (64 bit)     1000000       31.781000             0.031781      31465340.926969
bench_read_n_bits_into: big endian (256 bit)       1000000       82.193000             0.082193      12166486.197121
bench_read_n_bits_into: little endian (256 bit)    1000000       41.800000             0.041800      23923444.976077
```

`bench_rz_il_mem_loadw()` (which uses `read_n_bits()` instead of `read_n_bits_into()` currently) before:
```
Benchmark                                       Iterations Total time [ms] Average time [us/op] Throughput [ops/sec] 
---------------------------------------------------------------------------------------------------------------------
bench_rz_il_mem_loadw: big endian (64 bit)        1000000      875.947000             0.875947       1141621.582128
bench_rz_il_mem_loadw: little endian (64 bit)     1000000      404.605000             0.404605       2471546.322957
bench_rz_il_mem_loadw: big endian (256 bit)       1000000     2714.149000             2.714149        368439.610353
bench_rz_il_mem_loadw: little endian (256 bit)    1000000     2268.531000             2.268531        440813.901155
```

`bench_rz_il_mem_loadw()` after:
```
Benchmark                                       Iterations Total time [ms] Average time [us/op] Throughput [ops/sec] 
---------------------------------------------------------------------------------------------------------------------
bench_rz_il_mem_loadw: big endian (64 bit)         1000000      229.170000             0.229170       4363572.893485
bench_rz_il_mem_loadw: little endian (64 bit)      1000000      230.407000             0.230407       4340145.915706
bench_rz_il_mem_loadw: big endian (256 bit)        1000000      428.938000             0.428938       2331339.261152
bench_rz_il_mem_loadw: little endian (256 bit)     1000000      371.948000             0.371948       2688547.861529
```

**Test plan**

Added unit tests for new functions and micro benchmarks.

**Closing issues**

Closes https://github.com/rizinorg/rizin/issues/5796


Rot127:
What is the result of the benchmark btw?

antonangeloff:
> What is the result of the benchmark btw?

Seems to be around `3.8` times for LE and `12` times for BE reads. I've added the results to the PR description.

BTW are the other functions in `mem.c` assuming the RzBuffer is going be at a specific position before accessing it? If not we could drop calls to rz_buf_tell() and second call to rz_buf_seek() [here](https://github.com/rizinorg/rizin/pull/5819/changes#diff-d3f96a8d4e90bd7b96d432934da85f814335acd86508e377c6a060ce1284bb66R139) which should speed up the function more

Also, do you have any suggestions for [this](https://github.com/rizinorg/rizin/pull/5819/changes#diff-d3f96a8d4e90bd7b96d432934da85f814335acd86508e377c6a060ce1284bb66R128-R131)?

librz/il/definitions/mem.c
```diff
@@ -113,28 +113,34 @@ RZ_API bool rz_il_mem_store(RzILMem *mem, RzBitVector *key, RzBitVector *value)
...
 
-	ut8 *data = calloc(n_bytes, 1);
-	if (!data) {
+		// TODO:
+		//   this seem to fail for addresses above UT64_MAX/2; possible options:
+		//      - should we return NULL (which breaks other tests)?
+		//      - call rz_warn_if_reached() and return zero bitvector?
+		//      - otherwise fail silently and return a zero bitvector (current case)
+		//
 		return value;
```
Rot127: This is essentially https://github.com/rizinorg/rizin/issues/5806.

I would vote for an `RZ_LOG_WARN("Attempted read beyone ST64_MAX")` or similar.
But let's wait what the others say.
cc @notxvilka @wargio

Rot127:
> BTW are the other functions in mem.c assuming the RzBuffer is going be at a specific position before accessing it? 

I don't think so, but would need to check.
But generally we can't assume this. The performance improvement here is particularity important for emulations and reading large files. While file reading might be linear, emulation is totally random access. So we sadly can't assume that the `seek` is at a certain position.

librz/il/definitions/mem.c
notxvilka: Yep, sounds reasonable.

wargio: ok for me.

antonangeloff:
> I don't think so, but would need to check. But generally we can't assume this. The performance improvement here is particularity important for emulations and reading large files. While file reading might be linear, emulation is totally random access. So we sadly can't assume that the `seek` is at a certain position.

It seems `rz_il_mem_load*()`/`rz_il_mem_store*()` functions always seek to the address they want to read/write, in this perspective it may not be necessary to seek back to the original address after read in `read_n_bits()` perhaps. But on the other hand the RzILMem doesn't own the buffer - so if the same buffer is also used elsewhere, not restoring the buffer position in `read_n_bits()` might mess up other users of the buffer.

librz/il/definitions/mem.c
```diff
@@ -123,22 +123,15 @@ static RzBitVector *read_n_bits(RzBuffer *buf, ut32 n_bits, RzBitVector *key, bo
...
 	if ((current_address = rz_buf_tell(buf)) < 0 || rz_buf_seek(buf, rz_bv_to_ut64(key), RZ_BUF_SET) < 0) {
 		// Seek failed
-
-		// TODO:
-		//   this seem to fail for addresses above UT64_MAX/2; possible options:
-		//      - should we return NULL (which breaks other tests)?
-		//      - call rz_warn_if_reached() and return zero bitvector?
-		//      - otherwise fail silently and return a zero bitvector (current case)
-		//
+		RZ_LOG_WARN("Attempted read beyond ST64_MAX");
```
Rot127:
```
		RZ_LOG_WARN("Attempted read beyond ST64_MAX. See issue #...\n");
```

Rot127:
@antonangeloff I miss-understood you. Yes, I think seeking back is not necessary.

antonangeloff:
> @antonangeloff I miss-understood you. Yes, I think seeking back is not necessary.

Removed it and this seems to have noticable effect. I've updated the PR description with then new results

librz/include/rz_endian.h
```diff
@@ -1862,6 +1862,20 @@ static inline ut64 rz_swap_ut64(ut64 val) {
 }
 #endif
 
+/**
+ * \brief Reverses the byte order of an arbitrary sized byte array.
+ * \param bytes The byte array.
+ * \param n Length of the byte array.
+ */
+static inline void rz_swap_n_bytes(ut8 *bytes, ut32 n) {
```
Rot127: You can use the `rz_mem_swap_` inplace functions for this. It is generally more efficient and should work because the pointer of the buffer should be aligned.

librz/util/bitvector.c
```diff
@@ -1835,6 +1835,126 @@ RZ_API void rz_bv_set_from_bytes_ble(RZ_NONNULL RzBitVector *bv, RZ_IN RZ_NONNUL
...
+ *
+ * The buffer position remains changed, so the caller is expected to seek the buffer cursor back if necessary.
+ *
+ * Similar to `rz_bv_set_from_bytes_le()`:
+ *  - The bitvector's size is unchanged.
+ *  - If `bv->len` < `bit_size`, additional bits are cut off, if `bv->len` > `bit_size`, the rest is filled up with 0.
+ */
+RZ_API void rz_bv_set_from_buffer_le(RZ_NONNULL RzBitVector *bv, RZ_NONNULL RzBuffer *buf, ut32 bit_size) {
+	rz_return_if_fail(bv && buf && bit_size);
+	bit_size = RZ_MIN(bit_size, bv->len);
```
Rot127:
```suggestion
	bit_size = RZ_MIN(bit_size, rz_bv_len(bv));
```

librz/include/rz_endian.h
Rot127: Please move the function to `rz_mem.h` though. And maybe leave a comment here about its existence in in `rz_mem.h`

librz/util/bitvector.c
```diff
@@ -1835,6 +1835,126 @@ RZ_API void rz_bv_set_from_bytes_ble(RZ_NONNULL RzBitVector *bv, RZ_IN RZ_NONNUL
...
+ * \param buf RzBuffer containing at least `(bit_size + 7) / 8` bytes at\after the current position.
+ * \param bit_size number of bits to read from buf.
+ *
+ * The buffer position remains changed, so the caller is expected to seek the buffer cursor back if necessary.
+ *
+ * Similar to `rz_bv_set_from_bytes_le()`:
+ *  - The bitvector's size is unchanged.
+ *  - If `bv->len` < `bit_size`, additional bits are cut off, if `bv->len` > `bit_size`, the rest is filled up with 0.
+ */
+RZ_API void rz_bv_set_from_buffer_le(RZ_NONNULL RzBitVector *bv, RZ_NONNULL RzBuffer *buf, ut32 bit_size) {
```
Rot127:
```suggestion
RZ_API void rz_bv_set_from_buffer_le(RZ_NONNULL RZ_OUT RzBitVector *bv, RZ_NONNULL RzBuffer *buf, ut32 bit_size) {
```

Also I would suggest to return `bool` success/failure.

librz/util/bitvector.c
```diff
@@ -1835,6 +1835,126 @@ RZ_API void rz_bv_set_from_bytes_ble(RZ_NONNULL RzBitVector *bv, RZ_IN RZ_NONNUL
...
+ * \param buf RzBuffer containing at least `(bit_size + 7) / 8` bytes at\after the current position.
+ * \param bit_size number of bits to read from \p buf.
+ *
+ * The buffer position remains changed, so the caller is expected to seek the buffer cursor back if necessary.
+ *
+ * Similar to `rz_bv_set_from_bytes_be()`:
+ *  - The bitvector's size is unchanged.
+ *  - If `bv->len` < `bit_size`, additional bits are cut off, if `bv->len` > `bit_size`, the rest is filled up with 0.
+ */
+RZ_API void rz_bv_set_from_buffer_be(RZ_NONNULL RzBitVector *bv, RZ_NONNULL RzBuffer *buf, ut32 bit_size) {
```
Rot127:
```suggestion
RZ_API void rz_bv_set_from_buffer_be(RZ_NONNULL RZ_OUT RzBitVector *bv, RZ_NONNULL RzBuffer *buf, ut32 bit_size) {
```

Also here, return bool.

librz/util/bitvector.c
```diff
@@ -1835,6 +1835,126 @@ RZ_API void rz_bv_set_from_bytes_ble(RZ_NONNULL RzBitVector *bv, RZ_IN RZ_NONNUL
...
+	rz_return_if_fail(bv && buf && bit_size);
+	bit_size = RZ_MIN(bit_size, bv->len);
```
Rot127: use API

librz/util/bitvector.c
```diff
@@ -1835,6 +1835,126 @@ RZ_API void rz_bv_set_from_bytes_ble(RZ_NONNULL RzBitVector *bv, RZ_IN RZ_NONNUL
...
+	// Handle sub-byte copies
+	if (bit_size < 8) {
+		ut8 data = 0;
+		rz_buf_read(buf, &data, 1);
+		rz_bv_set_from_bytes_be(bv, &data, 0, bit_size);
+		return;
+	}
+
+	// Specialized handling for small bitvectors (<= 64 bit)
+	if (bv->len <= 64) {
```
Rot127: API

librz/util/bitvector.c
```diff
@@ -1835,6 +1835,126 @@ RZ_API void rz_bv_set_from_bytes_ble(RZ_NONNULL RzBitVector *bv, RZ_IN RZ_NONNUL
...
+		rz_buf_read(buf, ((ut8 *)&bv->bits.small_u), byte_size);
```
Rot127: `((ut8 *)&bv->bits.small_u)`

Not sure if we should have a `get_buf_hot_path` function for this access. So the warning, it is only for exceptions, can be written somewhere.

cc @notxvilka @wargio

librz/util/bitvector.c
```diff
@@ -1835,6 +1835,126 @@ RZ_API void rz_bv_set_from_bytes_ble(RZ_NONNULL RzBitVector *bv, RZ_IN RZ_NONNUL
...
+}
+
+/**
+ * \brief Helper function for calling `rz_bv_set_from_buffer_be()` or `rz_bv_set_from_buffer_le()` based on a flag
+ * \param bv bitvector to assign the new value to.
+ * \param buf RzBuffer containing at least `(bit_size + 7) / 8` bytes at\after the current position.
+ * \param bit_size number of bits to read from \p buf
+ * \param big_endian control flag for specifying endian type
+ */
+RZ_API void rz_bv_set_from_buffer_ble(RZ_NONNULL RzBitVector *bv, RZ_NONNULL RzBuffer *buf, ut32 bit_size, bool big_endian) {
```
Rot127:
```suggestion
RZ_API bool rz_bv_set_from_buffer_ble(RZ_NONNULL RZ_OUT RzBitVector *bv, RZ_NONNULL RzBuffer *buf, ut32 bit_size, bool big_endian) {
```

test/unit/test_bitvector.c
```diff
@@ -980,6 +980,53 @@ bool test_rz_bv_set_from_bytes_le(void) {
...
+		RzBitVector bv;
+		rz_bv_init(&bv, len[i]);
+
+		for (ut32 bits_to_copy = 1; bits_to_copy < bv.len + 10; bits_to_copy++) {
+			rz_buf_seek(buf, 0, RZ_BUF_SET);
+			rz_bv_set_from_buffer_ble(&bv, buf, bits_to_copy, big_endian);
+			char *result = rz_bv_as_hex_string(&bv, true);
+
+			rz_bv_set_from_bytes_ble(&bv, data, 0, bits_to_copy, big_endian);
+			char *ref = rz_bv_as_hex_string(&bv, true);
```
Rot127, line 1006: I understand your urge to not repeat code, but here the result and the reference depend both on the function which is supposed to be tested.
This is dangerous.
Also the loop is a little bit hard to follow. And IMHO a test should be understandable on the first look.
If @notxvilka or @wargio are fine with it, it is ok. But I would prefer if you could rewrite the test the "straight forward stupid" way as the tests above. This is also easier to debug.

Rot127 (changes requested):
Big fan of the changes!
I have a few points pedantic points though.

librz/include/rz_endian.h
antonangeloff: I've moved the function to `rz_mem.h`.

> You can use the rz_mem_swap_ inplace functions for this. It is generally more efficient and should work because the pointer of the buffer should be aligned.

The current `rz_mem_swap_` seem to do individual swap on each ut32/ut64 value in an array; Though for [this use case](https://github.com/rizinorg/rizin/pull/5819/changes#diff-6d51abcfd5287e29f93781fe3fdcbc9ca8b5d2dcb8ceb2068b64c1ceb319c7b1R1925) we need to swap all bytes of a single large int, e.g. in a 80-bit bitvector, byte 0 should become 9, rather than 3

test/unit/test_bitvector.c
antonangeloff: Actually it was more about making sure the new functions `rz_bv_set_from_buffer_*` are functionally identical to `rz_bv_set_from_bytes_*`. The assumption is `rz_bv_set_from_bytes_*` are already correct and tested and can be used as reference. 

> Also the loop is a little bit hard to follow. And IMHO a test should be understandable on the first look.

It just tests all possible variations the `bit_size` param to reduce risk of uncovered handling scenarios; and I guess large part of the variations are redundant. I could hard code different `bit_sizes` instead of doing [1..len+10] to avoid redundancy and hopefully make the test more clear.

I can also just do a series of concrete tests, similar to [test_rz_bv_set_from_buffer_ble()](https://github.com/antonangeloff/rizin/blob/e8cf5ca3cf35c3f9be856c322f988109784db92a/test/unit/test_bitvector.c#L789).

Let me know which you think is better

Rot127: >  The assumption is rz_bv_set_from_bytes_* are already correct and tested and can be used as reference.

Sorry, it was late yesterday. I misread `rz_bv_set_from_bytes_ble` for `rz_bv_set_from_buffer_ble`.

Its good then. Keep it as it is.

Rot127 (approved):
Good job. Let me trigger the travis CI first though, to see if it works there.

Rot127:
@antonangeloff You can click "resolve conversation" once you addressed a comment. Keeps the PR clean and easier to follow.

librz/include/rz_endian.h
Rot127: All good. I was thinking about: if the bv is <= 32 bits wise, use the ut32_swaps etc. But this is over-engineered I guess. Yours is fine.

notxvilka (approved):

notxvilka:
@antonangeloff you did a great job on improving performance of various aspects, would you also be able to check this one? https://github.com/rizinorg/Rizin/issues/4386

antonangeloff:
Yes sure, I'll take a look later this week

PR #5820, opened 2026-01-19
Add new implementation of RzConfig

wargio:

**Detailed description**

Add new implementation of RzConfig to coexist with the old one while we convert all the configs variables to use the new APIs.

### DIFF
--- a/librz/arch/isa/6502/6502dis.c
+++ b/librz/arch/isa/6502/6502dis.c
@@ -149,7 +149,7 @@ int disass_6502(ut64 pc, RzAsmOp *op, const ut8 *buf, ut64 len) {
 		int len = ops[i].len;
 		switch (len) {
 		case 1:
-			rz_asm_op_setf_asm(op, "%s", ops[i].name);
+			rz_asm_op_set_asm(op, ops[i].name);
 			break;
 		case 2:
 			if (len > 1) {
--- a/librz/arch/isa/6502/6502dis.c
+++ b/librz/arch/isa/6502/6502dis.c
@@ -141,54 +141,46 @@ static struct {
 };
 
 int disass_6502(ut64 pc, RzAsmOp *op, const ut8 *buf, ut64 len) {
-	int i;
-	for (i = 0; ops[i].name != NULL; i++) {
-		if (ops[i].op == buf[0]) {
-			int len = ops[i].len;
-			switch (ops[i].len) {
-			case 1:
-				rz_asm_op_setf_asm(op, "%s", ops[i].name);
-				break;
-			case 2:
-				if (len > 1) {
-					rz_asm_op_setf_asm(op, ops[i].name, buf[1]);
-				} else {
-					rz_asm_op_set_asm(op, "truncated");
-					len = -1;
-				}
-				break;
-			case 3:
-				if (len > 2) {
-					rz_asm_op_setf_asm(op, ops[i].name, buf[1] + 0x100 * buf[2]);
-				} else {
-					rz_asm_op_set_asm(op, "truncated");
-					len = -1;
-				}
-				break;
-			case 4:
-				if (len > 3) {
-					rz_asm_op_setf_asm(op, ops[i].name, buf[1] + 0x100 * buf[2] + 0x10000 * buf[3]);
-				} else {
-					rz_asm_op_set_asm(op, "truncated");
-					len = -1;
-				}
-				break;
-			default:
-				rz_asm_op_set_asm(op, "invalid");
-				goto beach;
-			}
-			return len;
+	for (size_t i = 0; ops[i].name != NULL; i++) {
+		if (ops[i].op != buf[0]) {
+			continue;
 		}
+
+		int len = ops[i].len;
+		switch (len) {
+		case 1:
+			rz_asm_op_setf_asm(op, "%s", ops[i].name);
+			break;
+		case 2:
+			if (len > 1) {
+				rz_asm_op_setf_asm(op, ops[i].name, buf[1]);
+			} else {
+				rz_asm_op_set_asm(op, "truncated");
+				len = -1;
+			}
+			break;
+		case 3:
+			if (len > 2) {
+				rz_asm_op_setf_asm(op, ops[i].name, buf[1] + 0x100 * buf[2]);
+			} else {
+				rz_asm_op_set_asm(op, "truncated");
+				len = -1;
+			}
+			break;
+		case 4:
+			if (len > 3) {
+				rz_asm_op_setf_asm(op, ops[i].name, buf[1] + 0x100 * buf[2] + 0x10000 * buf[3]);
+			} else {
+				rz_asm_op_set_asm(op, "truncated");
+				len = -1;
+			}
+			break;
+		default:
+			rz_asm_op_set_asm(op, "invalid");
+			goto beach;
+		}
+		return len;
 	}
 beach:
 	return snesDisass(1, 1, pc, op, buf, len);
-}
-
-_6502State *_6502_state_new() {
-	_6502State *state = RZ_NEW0(_6502State);
-	if (!state) {
-		RZ_LOG_ERROR("Could not allocate memory for _6502State!");
-		return NULL;
-	}
-	return state;
 }
--- a/librz/arch/p/asm/asm_6502.c
+++ b/librz/arch/p/asm/asm_6502.c
@@ -15,52 +15,52 @@ static int _6520_disassemble(const RzAsm *a, RzAsmOp *op, const ut8 *buf, int le
 	return op->size = RZ_MAX(dlen, 0);
 }
 
-RzConfig *_6502_get_config(void *plugin_data) {
-	rz_return_val_if_fail(plugin_data, NULL);
-	_6502State *state = (_6502State *)plugin_data;
-	return rz_config_clone(state->cfg);
+static bool _6520_config_magic_get(void *user, void *data) {
+	ut64 *value = data;
+	_6502State *state = user;
+	if (!state) {
+		return false;
+	}
+	*value = state->magic;
+	return true;
 }
 
-static bool _6502_cfg_set(void *user, void *data) {
-	rz_return_val_if_fail(user && data, false);
+static bool _6520_config_magic_set(void *user, const void *data) {
+	const ut64 *value = data;
 	_6502State *state = user;
-	RzConfig *pcfg = state->cfg;
+	if (!state) {
+		return false;
+	}
+	state->magic = *value;
+	return true;
+}
 
-	RzConfigNode *cnode = (RzConfigNode *)data; // Config node from core.
-	RzConfigNode *pnode = rz_config_node_get(pcfg, cnode->name); // Config node of plugin.
-	if (pnode == cnode) {
-		return true;
+static RzConfig *_6502_get_config(void *plugin_data) {
+	rz_return_val_if_fail(plugin_data, NULL);
+	_6502State *state = (_6502State *)plugin_data;
+
+	RzConfig *cfg = rz_config_new(NULL);
+	if (!cfg) {
+		return NULL;
 	}
-	if (cnode) {
-		pnode->i_value = cnode->i_value;
-		free(pnode->value);
-		pnode->value = rz_str_dup(cnode->value);
-		return true;
-	}
-	return false;
+
+	// Add nodes
+	rz_config_add_integer_bind(cfg, "plugins.6502.magic", "Determines the magic number certain illegal opcodes use.", _6520_config_magic_get, _6520_config_magic_set, NULL, state);
+
+	return cfg;
 }
 
 static bool _6502_init(void **user) {
-	_6502State *state = _6502_state_new();
-	rz_return_val_if_fail(state, false);
-	RzConfig *cfg = state->cfg = rz_config_new(state);
-	rz_return_val_if_fail(state->cfg, false);
-
-	SETICB("plugins.6502.magic", 0xee, &_6502_cfg_set, "Determines the magic number certain illegal opcodes use.");
+	_6502State *state = RZ_NEW0(_6502State);
+	if (!state) {
+		return false;
+	}
+	state->magic = 0xee;
 	*user = state;
 	return true;
 }
 
-void _6502State_fini(_6502State *state) {
-	if (!state) {
-		return;
-	}
-	rz_config_free(state->cfg);
-	return;
-}
-
 static bool _6502_fini(void *user) {
-	_6502State_fini(user);
 	free(user);
 	return true;
 }
--- a/librz/arch/p/asm/asm_6502.c
+++ b/librz/arch/p/asm/asm_6502.c
@@ -45,7 +45,12 @@ static RzConfig *_6502_get_config(void *plugin_data) {
 	}
 
 	// Add nodes
-	rz_config_add_integer_bind(cfg, "plugins.6502.magic", "Determines the magic number certain illegal opcodes use.", _6520_config_magic_get, _6520_config_magic_set, NULL, state);
+	rz_config_add_integer_bind(cfg,
+		"plugins.6502.magic",
+		"Determines the magic number certain illegal opcodes use.",
+		_6520_config_magic_get,
+		_6520_config_magic_set,
+		NULL, state);
 
 	return cfg;
 }
--- a/librz/arch/p/asm/asm_hexagon.c
+++ b/librz/arch/p/asm/asm_hexagon.c
@@ -150,10 +150,26 @@ RZ_API RZ_OWN RzConfig *hexagon_get_config(void *plugin_data) {
 	}
 
 	// Add nodes
-	rz_config_add_bool_bind(cfg, "plugins.hexagon.imm.hash", "Display ## before 32bit immediates and # before immidiates with other width.", hexagon_config_imm_hash_get, hexagon_config_imm_hash_set, NULL, state);
-	rz_config_add_bool_bind(cfg, "plugins.hexagon.imm.sign", "True: Print them with sign. False: Print signed immediates in unsigned representation.", hexagon_config_imm_sign_get, hexagon_config_imm_sign_set, NULL, state);
-	rz_config_add_bool_bind(cfg, "plugins.hexagon.sdk", "Print packet syntax in objdump style.", hexagon_config_sdk_get, hexagon_config_sdk_set, NULL, state);
-	rz_config_add_bool_bind(cfg, "plugins.hexagon.reg.alias", "Print the alias of registers (Alias from C0 = SA0).", hexagon_config_reg_alias_get, hexagon_config_reg_alias_set, NULL, state);
+	rz_config_add_bool_bind(cfg, "plugins.hexagon.imm.hash",
+		"Display ## before 32bit immediates and # before immidiates with other width.",
+		hexagon_config_imm_hash_get,
+		hexagon_config_imm_hash_set,
+		NULL, state);
+	rz_config_add_bool_bind(cfg, "plugins.hexagon.imm.sign",
+		"True: Print them with sign. False: Print signed immediates in unsigned representation.",
+		hexagon_config_imm_sign_get,
+		hexagon_config_imm_sign_set,
+		NULL, state);
+	rz_config_add_bool_bind(cfg, "plugins.hexagon.sdk",
+		"Print packet syntax in objdump style.",
+		hexagon_config_sdk_get,
+		hexagon_config_sdk_set,
+		NULL, state);
+	rz_config_add_bool_bind(cfg, "plugins.hexagon.reg.alias",
+		"Print the alias of registers (Alias from C0 = SA0).",
+		hexagon_config_reg_alias_get,
+		hexagon_config_reg_alias_set,
+		NULL, state);
 
 	return cfg;
 }
--- a/librz/arch/p/asm/asm_hexagon.c
+++ b/librz/arch/p/asm/asm_hexagon.c
@@ -85,53 +85,51 @@ static RZ_OWN RzPVector /*<RzAsmTokenPattern *>*/ *get_token_patterns() {
 	return pvec;
 }
 
-/**
- * \brief Setter for the plugins RzConfig nodes.
- *
- * \param user The user of the RzConfig node. If this callback is called by Core \p user = RzCore.
- * If it is called by the plugins config setup \p user = HexState.
- * \param data The node to set. Again, if called by RzCore \p date = Node from RzCore config.
- * If it is called by the plugins config setup \p data = a plugins config node.
- * \return bool True if the config was set. False otherwise.
- */
-static bool hex_cfg_set(void *user, void *data) {
-	rz_return_val_if_fail(user && data, false);
-	HexState *state = user;
-	RzConfig *pcfg = state->cfg;
-
-	RzConfigNode *cnode = (RzConfigNode *)data; // Config node from core.
-	RzConfigNode *pnode = rz_config_node_get(pcfg, cnode->name); // Config node of plugin.
-	if (pnode == cnode) {
-		return true;
-	}
-	if (cnode) {
-		pnode->i_value = cnode->i_value;
-		free(pnode->value);
-		pnode->value = rz_str_dup(cnode->value);
-		return true;
-	}
-	return false;
-}
-
 static bool hexagon_fini(void *user) {
 	hexagon_state_fini(user);
 	free(user);
 	return true;
 }
 
+#define hexagon_getter_config(name, variable) \
+	static bool hexagon_##name(void *user, void *data) { \
+		bool *value = data; \
+		HexState *state = user; \
+		if (!state) { \
+			return false; \
+		} \
+		*value = state->variable; \
+		return true; \
+	}
+
+#define hexagon_setter_config(name, variable) \
+	static bool hexagon_##name(void *user, const void *data) { \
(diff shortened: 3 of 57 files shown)


librz/core/tui/panels.c
```diff
@@ -3285,7 +3285,7 @@ int __config_toggle_cb(void *user) {
 	RzPanelsMenuItem *child = panels_menu_item_get_selected_sub(parent);
 	RzStrBuf *tmp = rz_strbuf_new(child->name);
 	(void)rz_str_split(rz_strbuf_get(tmp), ':');
-	rz_config_toggle(core->config, rz_strbuf_get(tmp));
+	rz_config_toggle_boolean(core->config, rz_strbuf_get(tmp));
```
notxvilka: Too verbose, `toggle_bool` is more than enough

librz/core/tui/visual.c
```diff
@@ -105,7 +105,7 @@ RZ_IPI void rz_core_visual_toggle_decompiler_disasm(RzCore *core, bool for_graph
 		return;
 	}
 	hold = rz_config_hold_new(core->config);
-	rz_config_hold_s(hold, "asm.hint.pos", "asm.cmt.col", "asm.offset", "asm.lines",
+	rz_config_hold_variable(hold, "asm.hint.pos", "asm.cmt.col", "asm.offset", "asm.lines",
```
notxvilka: Could be `hold_var`, for example

notxvilka (changes requested):

librz/config/config.c
```diff
@@ -1,586 +1,998 @@
...
-			node->getter(cfg->user, node);
+RZ_IPI RZ_OWN RzList /*<char *>*/ *rz_config_string_list_dup(RZ_NULLABLE const RzList /*<char *>*/ *list) {
+	const char* elem;
+	const RzListIter *it;
+	RzList *dup = rz_list_newf(free);
+	if (!dup) {
+		return NULL;
+	}
+
+	rz_list_foreach(list, it, elem) {
```
Rot127: `list` allowed to be nullable but not checked for being NULL up until here.

librz/config/config.c
```diff
@@ -1,586 +1,998 @@
...
- * Reads the value of the config variable of \p name only and only if
- * the variable is boolean. Returns false in case of the failure.
+ * \brief      Enable or disable the lock of a given RzConfig which allows or disallows to add new variables
+ *
+ * \param      cfg          The RzConfig to lock/unlock
+ * \param[in]  enable_lock  When true locks the RzConfig structure.
+ */
+RZ_API void rz_config_lock(RZ_NONNULL RzConfig *cfg, bool enable_lock) {
+	rz_return_if_fail(cfg);
+	cfg->lock = enable_lock;
```
Rot127: Doesn't need to be in this PR, but I would use an atomic lock here.
RzConfig is one of those things we might want to share between threads.

librz/config/config_internal.h
```diff
@@ -0,0 +1,67 @@
...
+
+#include <rz_config.h>
+#include <rz_util.h>
+#include <rz_vector.h>
+
+typedef struct rz_config_node_t {
+	char *name; ///< Variable name
+	char *desc; ///< Description of the variable
+	RzList /*<char *>*/ *options; ///< Holds the string options
+	ut32 flags; ///< define the type of the data via RzConfigNodeFlags
```
Rot127:
```suggestion
	ut32 /*RzConfigNodeFlags*/ flags; ///< define the type of the data.
```

librz/config/config_internal.h
```diff
@@ -0,0 +1,67 @@
...
+		char *string;
+		ut64 integer;
+		bool boolean;
+		RzList /*<char *>*/ *list;
+	} value;
+} ConfigValue;
+
+struct rz_config_hold_t {
+	RzConfig *cfg;
+	RzVector /* ConfigValue */ variables;
```
Rot127:
```suggestion
	RzVector /*<ConfigValue>*/ variables;
```

librz/config/config_internal.h
```diff
@@ -0,0 +1,67 @@
...
+	RzConfigCallbackGet getter; ///< Gets the value via the callback (`data` can be `const char **` or `ut64 *` or `bool *`)
+	RzConfigCallbackSet setter; ///< Sets the value via the callback (`data` can be `const char **` or `ut64 *` or `bool *`)
+	RzConfigCallbackOpt setopt; ///< Sets the options via the callback (`data` is always RzList <char *> *` which can be modified)
+} RzConfigNode;
+
+struct rz_config_t {
+	bool lock; ///< When true forbids the add/remove of variables
+	void *user; ///< Pointer passed to the callbacks (getter, setter, setopt)
+	RzList /*<RzConfigNode *>*/ *nodes; ///< List of stored variables.
+	HtSP *ht; ///< HashMap for fast access to the nodes.
```
Rot127:
```suggestion
	HtSP *node_index; ///< HashMap for fast access to the nodes.
```

Why not use only a hash map?

librz/config/config_internal.h
```diff
@@ -0,0 +1,67 @@
...
+}
+
+RZ_IPI RZ_OWN RzList /*<char *>*/ *rz_config_string_list_dup(RZ_NULLABLE const RzList /*<char *>*/ *list);
+
+RZ_IPI bool rz_config_node_set_bool(RzConfig* cfg, RzConfigNode *node, bool value);
+RZ_IPI bool rz_config_node_set_integer(RzConfig* cfg, RzConfigNode *node, ut64 value);
+RZ_IPI bool rz_config_node_set_string(RzConfig* cfg, RzConfigNode *node, const char *value);
+RZ_IPI bool rz_config_node_set_list(RzConfig *cfg, RzConfigNode *node, RZ_NULLABLE const RzList /*<char *>*/ *value);
+
+RZ_IPI RZ_BORROW const char *rz_config_node_get_string(RzConfig *cfg, const RzConfigNode *node);
```
Rot127:
```suggestion
RZ_IPI const char *rz_config_node_get_string(RzConfig *cfg, const RzConfigNode *node);
```

Borrow is implied by `const`

librz/config/config_internal.h
```diff
@@ -0,0 +1,67 @@
...
+RZ_IPI ut64 rz_config_node_get_integer(RzConfig *cfg, const RzConfigNode *node);
+RZ_IPI bool rz_config_node_get_bool(RzConfig *cfg, const RzConfigNode *node);
+RZ_IPI RZ_BORROW const RzList /*<char *>*/ *rz_config_node_get_list(RzConfig *cfg, const RzConfigNode *node);
```
Rot127:
```suggestion
RZ_IPI const RzList /*<char *>*/ *rz_config_node_get_list(RzConfig *cfg, const RzConfigNode *node);
```

librz/config/config_internal.h
```diff
@@ -0,0 +1,67 @@
...
+
+struct rz_config_t {
+	bool lock; ///< When true forbids the add/remove of variables
+	void *user; ///< Pointer passed to the callbacks (getter, setter, setopt)
+	RzList /*<RzConfigNode *>*/ *nodes; ///< List of stored variables.
+	HtSP *ht; ///< HashMap for fast access to the nodes.
+};
+
+typedef struct config_value_t {
+	ut32 flags;
```
Rot127: Add comment what enum is used here.

librz/config/config.c
```diff
@@ -1,586 +1,998 @@
...
+	} else if (!(node->flags & RZ_CONFIG_VAR_WRITABLE)) {
+		RZ_LOG_ERROR("config: variable %s is read only\n", node->name);
+		return false;
+	}
+
+	return rz_config_node_set_bool(cfg, node, value);
+}
+
+/**
+ * \brief      Sets the value (type string list) of a given variable of the config
```
Rot127:
```suggestion
 * \brief      Sets the value (type string list) of a given variable of the config.
```

Rot127 (changes requested):
Only checked the main files and hexagon. Looks like a way better architecture now.

Will do a full review later (currently pretty busy).

General request: Please document the high level design somewhere while you still remember the details.

librz/config/config.c
wargio: rz_list_foreach does the check.

librz/config/config.c
wargio: this lock is a fake one. tbh we could even remove it completely as it does not add any real world value.

librz/config/config_internal.h
wargio: i have remove this (and i'm trying to figure it out where to place this) but essentially is to store the nodes in a sorted way, so when you iterate over them you always have the variables sorted.

Rot127: Wouldn't `rz_list_add_sorted` work?

I mean, does it need to be fast? The configs are no hot path.

wargio: added sorted method.

notxvilka (changes requested):
Some of these changes could be extracted in separate PRs to reduce the line count. For example, changes in the `rz_util` and in `librz/arch/*`

ret2libc:
Sorry, noob question, but what are the advantages of this new impl vs the old one?

wargio:
> Sorry, noob question, but what are the advantages of this new impl vs the old one?

We get proper bindings and not the hot mess the current one is. example:

`asm.arch` is a value that is hold by RzConfig, but has a `setter` method which does some stuff, but the setter is also used to query the `=?` but also to update the options.

With the new implementation, let's assume `asm.bits` is `32`, if i do `RzAsm->bits = 64` via the C apis and then i call `rz_config_get_integer(cfg, "asm.bits")` i will get `64`

So, with the new implementation you can just bind to the value (using setter/getter) and the value is always owned by the original structure and live and dies with that structure, so no more recursive sets (yes this happens and is horrendous), desync between the value in the structure and the value held by RzConfig.

The other perks are, less memory footprint, new special types (list of strings and intervals), automatic update of options, proper apis for management and new RzConfigHold implementation which is "type" agnostic (as the type is automatically handled internally)

wargio:
other cool things you could do is have dynamic options, for example:

If you load a binary and is raw, then you can have as `search.in` just the raw mapped memory, but if is an elf, you could have all the current mapped options we have + the segments name, sections etc..

The new APIs are very strict and checks always the types and what you is defined, for example if an option is a bool, you cannot call `rz_config_set_integer` for it, but as well you cannot get a string from a bool type, unless you call the `as_string` method which will return an owned string that is guaranteed to be a copy of the stored value.

librz/arch/isa/6502/6502dis.c
```diff
@@ -141,54 +141,46 @@ static struct {
...
-				goto beach;
+	for (size_t i = 0; ops[i].name != NULL; i++) {
+		if (ops[i].op != buf[0]) {
+			continue;
+		}
+
+		int len = ops[i].len;
+		switch (len) {
+		case 1:
+			rz_asm_op_setf_asm(op, "%s", ops[i].name);
```
notxvilka: `rz_asm_op_set_asm(op, ops[i].name)`

librz/arch/p/asm/asm_6502.c
```diff
@@ -15,52 +15,52 @@ static int _6520_disassemble(const RzAsm *a, RzAsmOp *op, const ut8 *buf, int le
...
-	SETICB("plugins.6502.magic", 0xee, &_6502_cfg_set, "Determines the magic number certain illegal opcodes use.");
-	*user = state;
-	return true;
+	RzConfig *cfg = rz_config_new(NULL);
+	if (!cfg) {
+		return NULL;
+	}
+
+	// Add nodes
+	rz_config_add_integer_bind(cfg, "plugins.6502.magic", "Determines the magic number certain illegal opcodes use.", _6520_config_magic_get, _6520_config_magic_set, NULL, state);
```
notxvilka: Maybe split this long line?

librz/arch/p/asm/asm_hexagon.c
```diff
@@ -145,7 +143,19 @@ static bool hexagon_init(void **plugin_data) {
...
 	HexState *state = plugin_data;
-	return rz_config_clone(state->cfg);
+
+	RzConfig *cfg = rz_config_new(NULL);
+	if (!cfg) {
+		return NULL;
+	}
+
+	// Add nodes
+	rz_config_add_bool_bind(cfg, "plugins.hexagon.imm.hash", "Display ## before 32bit immediates and # before immidiates with other width.", hexagon_config_imm_hash_get, hexagon_config_imm_hash_set, NULL, state);
```
notxvilka: Same, maybe split those?

librz/config/config.c
```diff
@@ -1,586 +1,1278 @@
...
-		RZ_LOG_DEBUG("(error: '%s' is not a boolean variable)\n", name);
+
+	if (!entry->is_variable) {
+		return rz_config_node_set_readonly(&entry->node, read_only);
+	}
+
+	return config_var_set_readonly(&entry->var, read_only);
+}
+
+RZ_API bool rz_config_is_readonly(RZ_NONNULL RzConfig *cfg, RZ_NONNULL const char *name) {
```
notxvilka, line 155: Doxygen

librz/config/config.c
```diff
@@ -1,586 +1,1278 @@
...
-		RZ_LOG_DEBUG("(error: '%s' config key is read only)\n", name);
+
+	if (!entry->is_variable) {
+		return rz_config_node_is_ro(&entry->node);
+	}
+
+	return rz_config_var_is_readonly(&entry->var);
+}
+
+RZ_API void rz_config_iterate_over(RZ_NONNULL const RzConfig *cfg, RZ_NONNULL RzConfigIterator iterator, RZ_NULLABLE void *user) {
```
notxvilka, line 177: Here Doxygen is even more important

librz/config/config.c
```diff
@@ -1,586 +1,1278 @@
...
+	if (!var->value.string && value) {
 		return false;
 	}
-	return rz_str_is_true(node->value);
+	return true;
 }
 
-RZ_API const char *rz_config_node_type(RzConfigNode *node) {
-	rz_return_val_if_fail(node, "");
+static inline bool config_init_var_list(RzConfigVar *var, const char *name, const char *desc, RZ_OWN RzList /*<char *>*/ *value) {
```
notxvilka, line 224: Maybe we should use sets instead of lists here?

librz/config/config.c
```diff
@@ -1,586 +1,1278 @@
...
 }
 
-RZ_API RZ_BORROW RzConfigNode *rz_config_set_cb(RZ_BORROW RzConfig *cfg, const char *name, const char *value, RzConfigCallback cb) {
-	RzConfigNode *node = rz_config_set(cfg, name, value);
-	if (node && (node->setter = cb)) {
-		if (!node->setter(cfg->user, node)) {
-			return node;
-		}
+static inline bool config_init_var_bind(RzConfigVar *var, const char *name, const char *desc, ut32 flags, RzConfigBindGet get, RzConfigBindSet set, RzConfigBindOpts opts, void *user) {
+	var->name = rz_str_dup(name);
```
notxvilka, line 266: Shouldn't this be moved after the NULL check?

librz/config/config.c
```diff
@@ -1,586 +1,1278 @@
...
 	}
-	return node;
+
+	config_add_entry(cfg, name, &new_entry, true);
+	return true;
 }
 
-static bool __is_true_or_false(const char *s) {
-	return s && (!rz_str_casecmp(s, "true") || !rz_str_casecmp(s, "false"));
+RZ_API bool rz_config_add_integer(RZ_NONNULL RzConfig *cfg, RZ_NONNULL const char *name, RZ_NULLABLE const char *desc, ut64 value) {
```
notxvilka, line 315: Out of the scope of this PR but maybe we should have both signed and unsigned integer types in the config...

librz/config/config.c
```diff
@@ -1,586 +1,1278 @@
...
-				rz_list_append(cfg->nodes, node);
-			}
-		} else {
-			RZ_LOG_ERROR("(locked: no new keys can be created (%s))\n", name);
-		}
+	config_add_entry(cfg, name, &new_entry, true);
+	return true;
+}
+
+RZ_API bool rz_config_add_options(RZ_NONNULL RzConfig *cfg, RZ_NONNULL const char *name, RZ_NULLABLE const char *desc, ...) {
```
notxvilka, line 367: This one definitely deserves Doxygen since it's unclear what exactly it does just from the name

librz/config/config.c
```diff
@@ -1,586 +1,1278 @@
...
+		RZ_LOG_ERROR("config: failed to initialize '%s'.\n", name);
+		config_var_fini(&new_entry.var);
+		return false;
+	}
+
+	config_add_entry(cfg, name, &new_entry, true);
+	return true;
+}
+
+RZ_API bool rz_config_add_string_bind(RZ_NONNULL RzConfig *cfg, RZ_NONNULL const char *name, RZ_NULLABLE const char *desc, RZ_NONNULL RzConfigBindGet get, RZ_NULLABLE RzConfigBindSet set, RZ_NULLABLE RzConfigBindOpts opts, RZ_NULLABLE void *user) {
```
notxvilka: These functions look like exact copy of each other with just different type. Not sure, but maybe worth to make a macro for this?

librz/config/config.c
```diff
@@ -1,586 +1,1278 @@
...
+	}
+
+	ut64 value = 0;
+	if (!config_var_bind_get_value(var, &value)) {
+		return 0;
+	}
+	return value;
+}
+
+RZ_API const char *rz_config_var_get_string(RZ_NONNULL const RzConfigVar *var) {
```
notxvilka, line 646: Same comment here about the macro

librz/config/config.c
```diff
@@ -1,586 +1,1278 @@
...
-		if (!node->value) {
-			node = NULL;
-			goto beach;
+		rz_strbuf_append(&sb, "readonly");
+	}
+
+	return rz_strbuf_drain_nofree(&sb);
+}
+
+RZ_IPI bool rz_config_var_set_bool(RzConfigVar *var, bool value) {
```
notxvilka, line 828: And for these - could be deduplicated with the macro

librz/config/config.c
```diff
@@ -1,586 +1,1278 @@
...
+	}
+
+	if (!entry->is_variable) {
+		return rz_config_set(cfg, name, value) != NULL;
+	}
+
+	return rz_config_var_set_any(&entry->var, value);
+}
+
+RZ_API bool rz_config_get_bool(RZ_NONNULL const RzConfig *cfg, RZ_NONNULL const char *name) {
```
notxvilka, line 1297: Same here, could be deduplicated with the macro (possibly)

librz/config/config.c
```diff
@@ -1,586 +1,1278 @@
...
+
+	if (entry->is_variable) {
+		return rz_config_var_get_flags(&entry->var);
+	}
+	return rz_config_node_get_var_flags(&entry->node);
 }
 
-RZ_API bool rz_config_unserialize(RZ_NONNULL RzConfig *config, RZ_NONNULL Sdb *db, RZ_NULLABLE char **err) {
-	sdb_foreach(db, load_config_cb, config);
+RZ_API bool rz_config_set_options(RZ_NONNULL RzConfig *cfg, RZ_NONNULL const char *name, RZ_NULLABLE RZ_OWN RzList /*<char *>*/ *options) {
```
notxvilka: Like I mentioned before, we could use set instead of a list here

librz/config/config_internal.h
```diff
@@ -0,0 +1,76 @@
...
+#ifndef CONFIG_INTERNAL_H
+#define CONFIG_INTERNAL_H
+
+#include <rz_config.h>
+#include <rz_util.h>
+#include <rz_vector.h>
+
+typedef struct config_value_t {
+	const char *name; ///< Variable name
+	ut32 /*RzConfigVarFlags*/ flags; ///< Copy of the variable flags
```
notxvilka: What is this comment? It will likely cause a problem for rz-bindgen

librz/config/old_config.c
notxvilka, line 1: What for you need an old config code?

librz/core/tui/config.c
```diff
@@ -12,156 +12,188 @@
 #include <rz_util/rz_strbuf.h>
 #include <cmd_descs.h>
 
+typedef struct config_visual_ctx {
+	const char *fs;
+	const char *fs2;
+	const char *desc;
+	int i;
```
notxvilka, line 19: Could you please add comments here, especially for these short-named fields?

librz/core/tui/config.c
```diff
@@ -12,156 +12,188 @@
...
+		char *buf = rz_core_editor(core, NULL, value);
+		rz_config_set_any(core->config, name, buf);
+		free(buf);
+		free(value);
 		return;
 	}
-	if (rz_config_node_is_bool(node)) {
-		rz_config_set_i(core->config, name, node->i_value ? 0 : 1);
+
+	// FGETS AND SO
```
notxvilka: What?

librz/core/tui/config.c
```diff
@@ -12,156 +12,188 @@
...
+	free(value);
+}
+
+static void show_config_options(RzCore *core, const char *name) {
+	const RzList *options = rz_config_get_options(core->config, name);
+	if (rz_list_empty(options)) {
+		return;
+	}
+
+	int h, w = rz_cons_get_size(&h);
```
notxvilka: `h` seems unused here, could be `_h`

librz/core/tui/config.c
```diff
@@ -12,156 +12,188 @@
...
-			rz_cons_show_cursor(false);
-			rz_config_set(core->config, name, buf);
-			// node->value = rz_str_dup (node->value, buf);
+		ctx->show = 0;
+	}
+	if (ctx->show) {
+		if (ctx->option == ctx->i) {
+			ctx->hit = 1;
 		}
+		if ((ctx->i >= ctx->option - ctx->delta) && ((ctx->i < ctx->option + ctx->delta) || ((ctx->option < ctx->delta) && (ctx->i < (ctx->delta << 1))))) {
```
notxvilka: Split this into two lines pls

librz/core/tui/config.c
```diff
@@ -12,156 +12,188 @@
...
+	const char *name = rz_config_entry_get_name(entry);
+	const char *desc = rz_config_entry_get_desc(entry);
+
+	if (!rz_str_ccmp(name, ctx->fs, '.')) {
+		if (ctx->option == ctx->i) {
+			ctx->fs2 = name;
+			ctx->desc = desc;
+			ctx->hit = 1;
+		}
+		if ((ctx->i >= ctx->option - ctx->delta) && ((ctx->i < ctx->option + ctx->delta) || ((ctx->option < ctx->delta) && (ctx->i < (ctx->delta << 1))))) {
```
notxvilka: Same, split

librz/core/cautocmpl.c
```diff
@@ -549,22 +549,33 @@ static void autocmplt_cmd_arg_choices(RzCore *core, RzLineNSCompletionResult *re
...
-static void autocmplt_cmd_arg_eval_key(RzCore *core, RzLineNSCompletionResult *res, const char *s, size_t len) {
-	RzListIter *iter;
-	RzConfigNode *bt;
-	rz_list_foreach (core->config->nodes, iter, bt) {
-		if (!strncmp(bt->name, s, len)) {
-			rz_line_ns_completion_result_add(res, bt->name);
-		}
+typedef struct autocmplt_context_s {
+	RzLineNSCompletionResult *res;
+	const char *s;
```
notxvilka: Add a comment please

librz/include/rz_config.h
```diff
@@ -41,97 +41,192 @@ typedef struct rz_config_node_t {
...
+	RzConfigBindGet get_value;
+	RzConfigBindSet set_value;
+	RzConfigBindOpts get_options;
+} RzConfigBind;
+
+typedef struct rz_config_var_t {
+	char *name; ///< Variable name
+	char *desc; ///< Description of the variable
+	RzList /*<char *>*/ *options; ///< Variable possible values
+	ut32 /*RzConfigVarFlags*/ flags; ///< Define the type of the data via RzConfigVar
```
notxvilka: Same here -  a question about types

notxvilka (changes requested):
Cool cleanup overall! Some minor things here and there though. But the main question is - wouldn't this affect projects? If yes, then the version should be upped and migration is implemented.

wargio:
> Cool cleanup overall! Some minor things here and there though. But the main question is - wouldn't this affect projects? If yes, then the version should be upped and migration is implemented.

it should not. i tried my best to keep the code as compatible as possible. as the new types re not used anywhere yet (besides the unit test) there is no need to bump anything.

the only change is on a specific variable that was duplicated for some reasons in cconfig.c.

librz/config/config.c
wargio: i actually was thinking the same when i was introducing the changes, but i could not find a place where we needed this.

we use everywhere int ot ut64, but the int types are never negative.

librz/config/old_config.c
wargio: we need to keep the old code till we move all the evars in cconfig.c to the new code.

librz/core/tui/config.c
wargio: tbh i kept the same names as this is not a refacoring of the tui code. i have no idea what they do, just extracted them into a struct to keep the same behaviour

notxvilka (approved):

PR #5825, opened 2026-01-21
Removed RZ_PACKED

### LINKED ISSUE #5813: Refactor bin_dol.c to remove RZ_PACKED
Currently `DolHeader` in `librz/bin/p/bin_dol.c` uses `RZ_PACKED`.

The idea is to rewrite the `load_buffer` function to not call `rz_buf_fread_at` but instead use the `rz_buf_offset_*` APIs.

Also this plugins is missing tests, which should be added to the test suite.

### how to use the RzBuffer api.

Define a structure in a normal way (so no RZ_PACKED, etc...):

```c
typedef struct foo_s {
   ut32 num32;
   char name[10];
   ....
   ut16 val16;
} foo_t;
```

Then you use the offset API that i mention:

```c
static bool parse_foo(RzBuffer *b, foo_t *foo, bool big_endian) {
    ut64 offset = 0;
    return rz_buf_read_ble32_offset(b, &offset, &foo->num32, big_endian) &&
               rz_buf_read_offset(b, &offset, foo->name, sizepf(foo->name)) &&
               ...
              rz_buf_read_ble16_offset(b, &offset, &foo->val16, big_endian) ;
}
```

This parses the same structure as the packed one via `memcpy(foo, buffer, sizeof(foo));` but is portable.

Real world example:

https://github.com/rizinorg/rizin/blob/602781b2dc171659c0d47659525492fb562cf5f2/librz/bin/p/bin_ninds.c#L83-L137

Farhan-25:

**Detailed description**

Refactored `librz/bin/p/bin_dol.c` to remove the use of `RZ_PACKED` structs for dol files. Using a function dol_parse_header to read through all the values.
`check_buffer `updated to remove memcpy usage.

...

**Test plan**

I have also included the test file for dol which is using the file https://github.com/rizinorg/rizin-testbins/blob/master/dol/sqrxz4-gc.dol. Added a list of cmds and their outputs to test.

I also ran a local test using rz-test and got the following result
image

...

**Closing issues**

...


wargio (changes requested):
merge the tests into one, no need to fragment them

librz/bin/p/bin_dol.c
```diff
@@ -23,39 +23,62 @@
...
+	 } 
+	ut32 text0_off, entry;
+    if (!rz_buf_read_be32_at(buf, 0x00, &text0_off)) { 
+		return false; 
+	} 
+	if (!rz_buf_read_be32_at(buf, 0xE0, &entry)) { 
+		return false; 
+	}  
+	bool text0_valid = (text0_off >= 0x100) && (text0_off % 4 == 0);
+	bool entry_valid = (entry & 0xF0000000) == 0x80000000;
```
wargio, line 76: these tests looks different from what was the old code. can you explain why these changes?

Farhan-25: The old file was using checking the first 6 bytes of the memory file with hardcoded bytes `"\x00\x00\x01\x00\x00\x00"`  `"\x00\x00\x00\x00\x00\x00"` to check if the dol file is accurate. 
In my version i am using the the valid dol header size for c which is 0x100 and just checking if the file header is of 0x100 if yes then its a valid dol file. 
Then the `text_valid ` is checking that if the text_0 starts after 0x100 and then check every byte is of 4 bytes or not.
`entry_valid `checks that if the entrypoint is valid which is 0x80000000 inside the memory range for dol file.

Farhan-25:
> merge the tests into one, no need to fragment them

Sorry but i didnt get it. I referred to other test files and in all that it was using like this , every command then its output. So can you please explain it a bit? 
I think you want the table commands together like separated by a column(:). Am i right?

wargio:
> > merge the tests into one, no need to fragment them
> 
> Sorry but i didnt get it. I referred to other test files and in all that it was using like this , every command then its output. So can you please explain it a bit? I think you want the table commands together like separated by a column(:). Am i right?

You added several tests using the same file (which is ok), since these tests are very simple, you can squash them together by calling all the commands in one go. check the `test/README.md`

librz/bin/p/bin_dol.c
wargio: here is the thing i do not understand.

Where is the `0x80000000` coming from? i cannot even find any mention from the Dolphin sources.

Farhan-25: https://wiibrew.org/wiki/Memory_map
I got that number from AI but i verified it and then did it
Also see this 
image
https://wiki.tockdom.com/wiki/DOL_(File_Format)
Also on the official GameCube code check this comment 
image

https://github.com/Cuyler36/Ghidra-GameCube-Loader/blob/master/src/main/java/gamecubeloader/dol/DOLHeader.java

wargio: tbh, i would not use this ghidra parser as reference, but the reference from the wiki is acceptable.

Do this: add a good comment above the function `check_buffer` explaining what those values are from and why that specific check. you can add a link to the wiki.

If you want to go beyond that (which is very appreciated), you can do like i did here:

https://github.com/rizinorg/rizin/blob/d9f41e80746fe951c6c73d825227ce2537200093/librz/bin/p/bin_sms.c#L7-L50

Farhan-25: ok sure i will do that. Can you help me with the error i got in this? 
Would you please point out where i am wrong , it failed in the table commands like aft and aftl, so i can fix it

wargio: error on the ci?

Farhan-25: yes

wargio: because the command aftl does not exists

Issue #3669, opened 2023-07-19
New command to print all RzIL instructions in function


XVilka:
Currently there are `pdf` and `pif` that print disassembly (enriched) and instructions for the current function. Ideally, having similar command but printing RzIL instead would be a good addition.

```
[0x100006948]> pdf
            ; CALL XREFS from sym.func.100005ddc @ 0x100005f34, 0x100006014
╭ sym.func.100006948(uint64_t arg1);
│           ; arg uint64_t arg1 @ x0
│           0x100006948      1f300071       cmp   w0, 0xc              ; arg1
│       ╭─< 0x10000694c      c2000054       b.hs  0x100006964
│       │   0x100006950      00840011       add   w0, w0, 0x21         ; arg1
│       │   0x100006954      d0071eca       eor   x16, x30, x30, lsl 1
│      ╭──< 0x100006958      5000f0b6       tbz   x16, 0x3e, 0x100006960
│      ││   0x10000695c      208e38d4       brk   0xc471
│     ╭╰──> 0x100006960      76030014       b     sym.imp.nl_langinfo  ; sym.imp.nl_langinfo
│     │ ╰─> 0x100006964      7f2303d5       pacibsp
│     │     0x100006968      fd7bbfa9       stp   x29, x30, [sp, -0x10]!
│     │     0x10000696c      fd030091       mov   x29, sp
╰     │     0x100006970      be020094       bl    sym.imp.abort        ; sym.imp.abort ; void abort(void)
[0x100006948]> pif
cmp w0, 0xc
b.hs 0x100006964
add w0, w0, 0x21
eor x16, x30, x30, lsl 1
tbz x16, 0x3e, 0x100006960
brk 0xc471
b sym.imp.nl_langinfo
pacibsp
stp x29, x30, [sp, -0x10]!
mov x29, sp
bl sym.imp.abort
[0x100006948]>
```
Currently it can be done with `aoi @@i` and `aoi @@i @@b` but not very obvious, also probably we should support "enriched" RzIL output, when addresses substituted with labels, etc, just like in `pdf` output:

```
[0x100006948]> aoi @@i
0x100006948 (seq (set a (cast 32 false (var x0))) (set b (bv 32 0xc)) (set r (- (var a) (var b))) (set cf (ule (var b) (var a))) (set vf (&& (^^ (msb (var a)) (msb (var b))) (^^ (msb (var a)) (msb (var r))))) (set zf (is_zero (var r))) (set nf (msb (var r))))
[...]

PR #5936, opened 2026-02-17
RzIL: add unicode print command

Ehab-24:

**Detailed description**

This PR partially implements the `pLf` command. It currently only prints RzIL in unicode and is NOT "enriched". This is meant to be a first iteration since it would require some experimentation to find the best representation of RzIL in unicode. Therefore, I have not updated the Rizin book or added a test case yet. 

...

**Test plan**

Start rizin shell, run `aaa` then `pLf`.
Sample output:
image

...

**Closing issues**

refers to #3669 

...

*Edit: Update screenshot*


notxvilka (changes requested):
It should work only if `scr.utf8=true` is set.

librz/il/il_export_string_unicode.c
```diff
@@ -0,0 +1,803 @@
+#include "rz_types.h"
```
notxvilka: Missing SPDX license header. Please also remove unnecessary header `<sys/types.h>`

notxvilka (changes requested):

notxvilka (changes requested):
Please fix formatting:
```
librz/core/cmd/cmd_print.c:3129:104: error: code should be clang-formatted [-Wclang-format-violations]
RZ_IPI RzCmdStatus rz_print_function_rzil_enriched_handler(RzCore *core, int argc, const char **argv) {
                                                                                                       ^
librz/core/cmd/cmd_print.c:3130:54: error: code should be clang-formatted [-Wclang-format-violations]
    if (!rz_config_get_b(core->config, "scr.utf8")) {
                                                     ^
librz/core/cmd/cmd_print.c:3131:65: error: code should be clang-formatted [-Wclang-format-violations]
        RZ_LOG_ERROR("Config variable 'scr.utf8' is not set\n");
                                                                ^
librz/core/cmd/cmd_print.c:3132:19: error: code should be clang-formatted [-Wclang-format-violations]
        goto exit;
                  ^
```

Rot127 (changes requested):
Please add tests,

librz/il/il_export_string_unicode.c
```diff
@@ -0,0 +1,799 @@
...
+	il_op_pure_string_resolve(opx->body, sb);
+	rz_strbuf_append(sb, ")");
+}
+
+static void il_opdmp_bool_false(const RzILOpPure *op, RzStrBuf *sb) {
+	rz_strbuf_append(sb, "𝔽");
+}
+
+static void il_opdmp_bool_true(const RzILOpPure *op, RzStrBuf *sb) {
+	rz_strbuf_append(sb, "𝕋");
```
wargio: i wouldn't use these for true or false, as i was confused on what `𝔽` meant since i was expecting it to be related to a mathematical set or anti-derivative of a function.

librz/il/il_export_string_unicode.c
```diff
@@ -0,0 +1,799 @@
...
+	free(num);
+}
+
+static void il_opdmp_msb(const RzILOpPure *op, RzStrBuf *sb) {
+	il_op_param_1("⊤", op->op.msb, pure, bv);
+}
+
+static void il_opdmp_lsb(const RzILOpPure *op, RzStrBuf *sb) {
+	il_op_param_1("⊥", op->op.lsb, pure, bv);
+}
```
wargio: i expect these to mean always true or false, but they are used for msb & lsb

librz/il/il_export_string_unicode.c
```diff
@@ -0,0 +1,799 @@
...
+
+static void il_opdmp_is_zero(const RzILOpPure *op, RzStrBuf *sb) {
+	il_op_param_1("Φ", op->op.lsb, pure, bv);
+}
```
wargio: this should be something else, and not `phi`

librz/il/il_export_string_unicode.c
```diff
@@ -0,0 +1,799 @@
...
+	rz_strbuf_append(sb, "(ŝ");
+	append_superscript(sb, opx->mem);
+	il_op_pure_string_resolve(opx->key, sb);
+	rz_strbuf_append(sb, " ");
+	il_op_pure_string_resolve(opx->value, sb);
+	rz_strbuf_append(sb, ")");
+}
+
+static void il_opdmp_nop(const RzILOpEffect *op, RzStrBuf *sb) {
+	il_op_param_0("Ǿ");
```
wargio: too similar to `∅`

librz/il/il_export_string_unicode.c
```diff
@@ -0,0 +1,799 @@
...
+}
+
+static void il_opdmp_empty(const RzILOpEffect *op, RzStrBuf *sb) {
+	il_op_param_0("Ø");
```
wargio:
```suggestion
	il_op_param_0("∅");
```

librz/il/il_export_string_unicode.c
```diff
@@ -0,0 +1,799 @@
...
+
+static void il_opdmp_set(const RzILOpEffect *op, RzStrBuf *sb) {
+	const RzILOpArgsSet *opx = &op->op.set;
+	rz_strbuf_appendf(sb, "(%s←", opx->v);
+	il_op_pure_string_resolve(opx->x, sb);
+	rz_strbuf_append(sb, ")");
+}
+
+static void il_opdmp_jmp(const RzILOpEffect *op, RzStrBuf *sb) {
+	il_op_param_1("⇡", op->op.jmp, pure, dst);
```
wargio: why not `↷`

librz/il/il_export_string_unicode.c
```diff
@@ -0,0 +1,799 @@
...
+}
+
+static void il_opdmp_goto(const RzILOpEffect *op, RzStrBuf *sb) {
+	const RzILOpArgsGoto *opx = &op->op.goto_;
+	rz_strbuf_appendf(sb, "(⇣%s)", opx->lbl);
```
wargio: not sure about this

librz/il/il_export_string_unicode.c
```diff
@@ -0,0 +1,799 @@
...
+	il_op_param_2("⁒", op->op.sdiv, pure, x, pure, y);
+}
+
+static void il_opdmp_mod(const RzILOpPure *op, RzStrBuf *sb) {
+	il_op_param_2("⊢", op->op.mod, pure, x, pure, y);
+}
+
+static void il_opdmp_smod(const RzILOpPure *op, RzStrBuf *sb) {
+	il_op_param_2("⊨", op->op.smod, pure, x, pure, y);
+}
```
wargio: i expect modulo to be `%`

librz/il/il_export_string_unicode.c
```diff
@@ -0,0 +1,799 @@
...
+static void il_opdmp_logand(const RzILOpPure *op, RzStrBuf *sb) {
+	il_op_param_2("∧", op->op.logand, pure, x, pure, y);
+}
+
+static void il_opdmp_logor(const RzILOpPure *op, RzStrBuf *sb) {
+	il_op_param_2("∨", op->op.logor, pure, x, pure, y);
+}
+
+static void il_opdmp_logxor(const RzILOpPure *op, RzStrBuf *sb) {
+	il_op_param_2("⩟", op->op.logxor, pure, x, pure, y);
```
wargio:
```suggestion
	il_op_param_2("⊕", op->op.logxor, pure, x, pure, y);
```

librz/il/il_export_string_unicode.c
```diff
@@ -0,0 +1,799 @@
...
+static void il_opdmp_shiftl(const RzILOpPure *op, RzStrBuf *sb) {
+	il_op_param_3("≪", op->op.shiftl, pure, x, pure, y, pure, fill_bit);
+}
+
+static void il_opdmp_eq(const RzILOpPure *op, RzStrBuf *sb) {
+	il_op_param_2("≡", op->op.ule, pure, x, pure, y);
+}
+
+static void il_opdmp_sle(const RzILOpPure *op, RzStrBuf *sb) {
+	il_op_param_2("≦", op->op.sle, pure, x, pure, y);
```
wargio:
```suggestion
	il_op_param_2("≦⁺", op->op.sle, pure, x, pure, y);
```

librz/il/il_export_string_unicode.c
```diff
@@ -0,0 +1,799 @@
...
+}
+
+static void il_opdmp_ule(const RzILOpPure *op, RzStrBuf *sb) {
+	il_op_param_2("≼", op->op.ule, pure, x, pure, y);
```
wargio:
```suggestion
	il_op_param_2("≦", op->op.ule, pure, x, pure, y);
```

librz/il/il_export_string_unicode.c
```diff
@@ -0,0 +1,799 @@
...
+}
+
+static void il_opdmp_cast(const RzILOpPure *op, RzStrBuf *sb) {
+	const RzILOpArgsCast *opx = &op->op.cast;
+	rz_strbuf_append(sb, "(");
+	il_op_pure_string_resolve(opx->val, sb);
+	rz_strbuf_append(sb, "⇒");
```
wargio:
```suggestion
	rz_strbuf_append(sb, "≈");
```

librz/il/il_export_string_unicode.c
```diff
@@ -0,0 +1,799 @@
...
+static void il_opdmp_fneg(const RzILOpPure *op, RzStrBuf *sb) {
+	il_op_param_1("−", op->op.fneg, pure, f);
+}
+
+static void il_opdmp_fabs(const RzILOpPure *op, RzStrBuf *sb) {
+	il_op_param_1("⌈", op->op.fabs, pure, f);
+}
+
+static void il_opdmp_fcast_int(const RzILOpPure *op, RzStrBuf *sb) {
+	il_op_param_1_with_mode_length("ꜰᴜ⇒", op->op.fcast_int, f, pure, mode, length);
```
wargio:
```suggestion
	il_op_param_1_with_mode_length("ꜰᴜ≈", op->op.fcast_int, f, pure, mode, length);
```

librz/il/il_export_string_unicode.c
```diff
@@ -0,0 +1,799 @@
...
+}
+
+static void il_opdmp_fcast_sint(const RzILOpPure *op, RzStrBuf *sb) {
+	il_op_param_1_with_mode_length("ꜰɪ⇒", op->op.fcast_sint, f, pure, mode, length);
```
wargio:
```suggestion
	il_op_param_1_with_mode_length("ꜰɪ≈", op->op.fcast_sint, f, pure, mode, length);
```

librz/il/il_export_string_unicode.c
```diff
@@ -0,0 +1,799 @@
...
+}
+
+static void il_opdmp_fcast_float(const RzILOpPure *op, RzStrBuf *sb) {
+	il_op_param_1_with_mode_format("ꜰᴜꜰ⇒", op->op.fcast_float, bv, pure, mode, format);
```
wargio:
```suggestion
	il_op_param_1_with_mode_format("ꜰᴜꜰ≈", op->op.fcast_float, bv, pure, mode, format);
```

librz/il/il_export_string_unicode.c
```diff
@@ -0,0 +1,799 @@
...
+}
+
+static void il_opdmp_fcast_sfloat(const RzILOpPure *op, RzStrBuf *sb) {
+	il_op_param_1_with_mode_format("ꜰꜰ⇒", op->op.fcast_sfloat, bv, pure, mode, format);
```
wargio:
```suggestion
	il_op_param_1_with_mode_format("ꜰꜰ≈", op->op.fcast_sfloat, bv, pure, mode, format);
```

librz/il/il_export_string_unicode.c
```diff
@@ -0,0 +1,799 @@
...
+}
+
+static void il_opdmp_fconvert(const RzILOpPure *op, RzStrBuf *sb) {
+	il_op_param_1_with_mode_format("⇒", op->op.fconvert, f, pure, mode, format);
```
wargio:
```suggestion
	il_op_param_1_with_mode_format("⩭", op->op.fconvert, f, pure, mode, format);
```

librz/il/il_export_string_unicode.c
```diff
@@ -0,0 +1,799 @@
...
+}
+
+static void il_opdmp_fsqrt(const RzILOpPure *op, RzStrBuf *sb) {
+	il_op_param_1_with_rmode("ʀ", op->op.fsqrt, f, pure, rmode);
```
wargio:
```suggestion
	il_op_param_1_with_rmode("²√", op->op.fsqrt, f, pure, rmode);
```

librz/il/il_export_string_unicode.c
```diff
@@ -0,0 +1,799 @@
...
+}
+
+static void il_opdmp_frsqrt(const RzILOpPure *op, RzStrBuf *sb) {
+	il_op_param_1_with_rmode("Ȓ", op->op.frsqrt, f, pure, rmode);
```
wargio:
```suggestion
	il_op_param_1_with_rmode("¹/√", op->op.frsqrt, f, pure, rmode);
```

librz/il/il_export_string_unicode.c
```diff
@@ -0,0 +1,799 @@
...
+}
+
+static void il_opdmp_fround(const RzILOpPure *op, RzStrBuf *sb) {
+	il_op_param_1_with_rmode("[", op->op.frsqrt, f, pure, rmode);
```
wargio:
```suggestion
static void il_opdmp_fround(const RzILOpPure *op, RzStrBuf *sb) {
	il_op_param_1_with_rmode("⭂", op->op.fround, f, pure, rmode);
```

librz/il/il_export_string_unicode.c
```diff
@@ -0,0 +1,799 @@
...
+}
+
+static void il_opdmp_frequal(const RzILOpPure *op, RzStrBuf *sb) {
+	const RzILOpArgsFrequal *opx = &op->op.frequal;
+	rz_strbuf_append(sb, "(");
+	rz_strbuf_append(sb, rz_il_float_stringify_rmode(opx->x));
+	rz_strbuf_append(sb, "≡");
```
wargio:
```suggestion
	rz_strbuf_append(sb, "≡f");
```

librz/il/il_export_string_unicode.c
```diff
@@ -0,0 +1,799 @@
...
+static void il_opdmp_fmad(const RzILOpPure *op, RzStrBuf *sb) {
+	il_op_param_3_with_rmode("ᴍᴀ", op->op.fmad, x, pure, y, pure, z, pure, rmode);
+}
+
+static void il_opdmp_fpown(const RzILOpPure *op, RzStrBuf *sb) {
+	il_op_param_2_with_rmode("⋏", op->op.fpown, f, pure, n, pure, rmode);
+}
+
+static void il_opdmp_frootn(const RzILOpPure *op, RzStrBuf *sb) {
+	il_op_param_2_with_rmode("√", op->op.frootn, f, pure, n, pure, rmode);
```
wargio:
```suggestion
	il_op_param_2_with_rmode("ⁿ√", op->op.frootn, f, pure, n, pure, rmode);
```

wargio:
i think there should be spaces between symbols

librz/il/il_export_string_unicode.c
```diff
@@ -0,0 +1,799 @@
...
+	const RzILOpArgsLet *opx = &op->op.let;
+	rz_strbuf_appendf(sb, "(%s=", opx->name);
+	il_op_pure_string_resolve(opx->exp, sb);
+	rz_strbuf_append(sb, " ");
+	il_op_pure_string_resolve(opx->body, sb);
+	rz_strbuf_append(sb, ")");
+}
+
+static void il_opdmp_bool_false(const RzILOpPure *op, RzStrBuf *sb) {
+	rz_strbuf_append(sb, "𝔽");
```
Rot127: Please move all those UTF characters into macros
`#define BOOL_FALSE "𝔽"` etc.
Makes it way way easier to change them.

Rot127 (changes requested):

librz/il/il_export_string_unicode.c
wargio: yes, this is a good idea

librz/il/il_export_string_unicode.c
Ehab-24: perhaps we could use ＠

Ehab-24:
@wargio I believe this looks better now
Also, wouldn't it be better to drop the `pLf` command and use unicode rzil in `plf` and `aoi[p]` if `scr.utf8` is set?

test/db/rzil/x86
```diff
@@ -104,4 +116,5 @@ psi @ obj.seckrit
 EOF
 EXPECT=Hello from RzIL!
 EXPECT_ERR=
+BROKEN=1
```
notxvilka: Why mark this as broken?

notxvilka (changes requested):
Overall, nice work! We could merge with these symbols and refine later, after some more use.

ehab-hazen:
> Overall, nice work! We could merge with these symbols and refine later, after some more use. 

Thanks! But this PR still feels incomplete though.

1. There is currently no pretty print for unicode rzil
2. I believe `aoi[p]` should output unicode rzil if scr.utf8 is set
3. The current `pLf` handler should probably just be a special case of `plf` if scr.utf8 is set

notxvilka:
> * I believe `aoi[p]` should output unicode rzil if scr.utf8 is set

No, at least for now. We can do that once we polish the implementation and representation.

> * The current `pLf` handler should probably just be a special case of `plf` if scr.utf8 is set

Definitely no. It's like a difference between `pi` and `pd` outputs. `pi` is simple instructions printer, while `pd` is the "enriched" output.

Idea that `pLf` might have even more information just like `pd` instead of just pure unicode representation. Reflines as the one example.

Ehab-24:
> Definitely no. It's like a difference between `pi` and `pd` outputs. `pi` is simple instructions printer, while `pd` is the "enriched" output.
> 
> Idea that `pLf` might have even more information just like `pd` instead of just pure unicode representation. Reflines as the one example.

Right! But the `pLf` implementation in this branch does not produce enriched output, it is the same as `plf` but uses unicode

test/db/rzil/x86
Ehab-24: unintentional

Rot127:
@Ehab-24 Please mark it as ready for review once the tests pass all.

Also mark the suggestions from above as resolved, if you finished them.

notxvilka:
It should not increase the time "emulateme" tests running - it should only affect printing. Something is wrong if it comes to that.

Ehab-24:
@notxvilka the increased time is due to the analysis phase (`aaa`) before the actual `pLf` command in regression tests

librz/core/cmd_descs/cmd_print.yaml
```diff
@@ -840,10 +840,18 @@ commands:
...
-    args: []
+  - name: pl
+    summary: Print RzIL
+    cname: cmd_print_rzil
+    subcommands:
+      - name: plf
+        summary: Print RzIL of the function
+        cname: print_function_rzil
+        args: []
+      - name: pLf
```
Rot127: It should be `plF`. `pl` is the command stub, so `l` cannot be changed to `pL`

librz/il/il_export_string_unicode.c
```diff
@@ -0,0 +1,885 @@
...
+		rz_strbuf_append(sb, " "); \
+		il_op_##sort2##_string_resolve(opx.v2, sb); \
+		rz_strbuf_append(sb, ")"); \
+	} while (0)
+
+#define append_superscript(sb, num) \
+	do { \
+		char *num_str = subscript_ut32(num); \
+		if (RZ_STR_ISEMPTY(num_str)) { \
+			rz_strbuf_append(sb, "⁰"); \
```
Rot127: If it fails it shouldn't just print 0.
At the minimum this needs an error.

librz/il/il_export_string_unicode.c
```diff
@@ -0,0 +1,885 @@
...
+	}
+	return result;
+}
+
+static void il_opdmp_var(const RzILOpPure *op, RzStrBuf *sb) {
+	const RzILOpArgsVar *opx = &op->op.var;
+	if (!rz_str_cmp(opx->v, "_val", 4)) {
+		rz_strbuf_appendf(sb, "%s", opx->v);
+	} else {
+		rz_strbuf_appendf(sb, "%s", opx->v);
```
Rot127: What is that about? The branches are the same

librz/il/il_export_string_unicode.c
```diff
@@ -0,0 +1,885 @@
...
+			rz_strbuf_append(sb, "⁰"); \
+		} else { \
+			rz_strbuf_append(sb, num_str); \
+			RZ_FREE(num_str); \
+		} \
+	} while (0)
+
+static RZ_OWN RZ_NULLABLE char *subscript_ut32(ut32 n) {
+	char buffer[32];
+	ssize_t len = snprintf(buffer, RZ_ARRAY_SIZE(buffer), "%u", n);
```
Rot127: use rz_strf

Rot127: Also init the buffer with 0

librz/il/il_export_string_unicode.c
```diff
@@ -0,0 +1,885 @@
...
+static void il_opdmp_msb(const RzILOpPure *op, RzStrBuf *sb) {
+	il_op_param_1(UCD_MSB, op->op.msb, pure, bv);
+}
+
+static void il_opdmp_lsb(const RzILOpPure *op, RzStrBuf *sb) {
+	il_op_param_1(UCD_LSB, op->op.lsb, pure, bv);
+}
+
+static void il_opdmp_is_zero(const RzILOpPure *op, RzStrBuf *sb) {
+	il_op_param_1_rtl(UCD_IS_ZERO, op->op.lsb, pure, bv);
```
Rot127:
```suggestion
	il_op_param_1_rtl(UCD_IS_ZERO, op->op.is_zero, pure, bv);
```

librz/il/il_export_string_unicode.c
```diff
@@ -0,0 +1,885 @@
...
+static void il_opdmp_shiftr(const RzILOpPure *op, RzStrBuf *sb) {
+	il_op_param_3(UCD_SHIFTR, op->op.shiftr, pure, x, pure, y, pure, fill_bit);
+}
+
+static void il_opdmp_shiftl(const RzILOpPure *op, RzStrBuf *sb) {
+	il_op_param_3(UCD_SHIFTL, op->op.shiftl, pure, x, pure, y, pure, fill_bit);
+}
+
+static void il_opdmp_eq(const RzILOpPure *op, RzStrBuf *sb) {
+	il_op_param_2(UCD_EQ, op->op.ule, pure, x, pure, y);
```
Rot127:
```suggestion
	il_op_param_2(UCD_EQ, op->op.eq, pure, x, pure, y);
```

librz/il/il_export_string_unicode.c
```diff
@@ -0,0 +1,885 @@
...
+static void il_opdmp_goto(const RzILOpEffect *op, RzStrBuf *sb) {
+	const RzILOpArgsGoto *opx = &op->op.goto_;
+	rz_strbuf_appendf(sb, "(" UCD_GOTO "%s)", opx->lbl);
+}
+
+static void il_opdmp_seq_inner(const RzILOpEffect *op, RzStrBuf *sb) {
+	const RzILOpArgsSeq *seq = &op->op.seq;
+	if (seq->x->code == RZ_IL_OP_SEQ) {
+		il_opdmp_seq_inner(seq->x, sb);
+		rz_strbuf_append(sb, "\n        ");
```
Rot127: move those spaces for indent also into a macro.

librz/il/il_export_string_unicode.c
```diff
@@ -0,0 +1,885 @@
...
+	} else {
+		il_op_effect_string_resolve(seq->x, sb);
+	}
+	if (seq->y->code != RZ_IL_OP_SEQ && seq->x->code != RZ_IL_OP_SEQ) {
+		rz_strbuf_append(sb, " ");
+	}
+	if (seq->y->code == RZ_IL_OP_SEQ) {
+		if (seq->x->code != RZ_IL_OP_SEQ) {
+			rz_strbuf_append(sb, "\n        ");
```
Rot127: also those

librz/il/il_export_string_unicode.c
```diff
@@ -0,0 +1,885 @@
...
+static void il_opdmp_repeat(const RzILOpEffect *op, RzStrBuf *sb) {
+	il_op_param_2(UCD_REPEAT, op->op.repeat, pure, condition, effect, data_eff);
+}
+
+static void il_opdmp_branch(const RzILOpEffect *op, RzStrBuf *sb) {
+	il_op_param_3(UCD_BRANCH, op->op.branch, pure, condition, effect, true_eff, effect, false_eff);
+}
+
+static void il_op_pure_string_resolve(const RzILOpPure *op, RzStrBuf *sb) {
+	if (!op) {
```
Rot127, line 659: rz_warn_if_reached

librz/il/il_export_string_unicode.c
```diff
@@ -0,0 +1,885 @@
...
+	default:
+		rz_warn_if_reached();
+		rz_strbuf_appendf(sb, UCD_UNK "%u", op->code);
+		return;
+	}
+}
+
+static void il_op_effect_string_resolve(const RzILOpEffect *op, RzStrBuf *sb) {
+	if (!op) {
+		rz_strbuf_append(sb, UCD_NOP);
```
Rot127: warn if reached

librz/il/il_export_string_unicode.c
```diff
@@ -0,0 +1,885 @@
...
+		il_opdmp_branch(op, sb);
+		return;
+	default:
+		rz_warn_if_reached();
+		rz_strbuf_appendf(sb, UCD_UNK "%u", op->code);
+		return;
+	}
+}
+
+RZ_API void rz_il_op_pure_stringify_unicode(RZ_NONNULL RzILOpPure *op, RZ_NONNULL RzStrBuf *sb) {
```
Rot127: Add doxygen

librz/il/il_export_string_unicode.c
```diff
@@ -0,0 +1,885 @@
...
+	rz_return_if_fail(op && sb);
+	il_op_pure_string_resolve(op, sb);
+}
+
+RZ_API void rz_il_op_effect_stringify_unicode(RZ_NONNULL RzILOpEffect *op, RZ_NONNULL RzStrBuf *sb) {
```
Rot127: Add doxygen

librz/il/il_export_string_unicode.c
Rot127: Also I think these functions should return bool for success and failure.
Returning false if any of the invali states are reached.

test/db/cmd/cmd_plF
Rot127, line 1: This still needs a test for the floating point operations. You can simply check for binaries or write instruction bytes from the asm tests with `wx`. Then create a function with `af+` and `afb+`, then print it.

librz/il/il_export_string_unicode.c
```diff
@@ -0,0 +1,885 @@
...
+}
+
+static void il_opdmp_bitv(const RzILOpPure *op, RzStrBuf *sb) {
+	const RzILOpArgsBv *opx = &op->op.bitv;
+	char *num = rz_bv_as_hex_string(opx->value, false);
+	rz_strbuf_append(sb, "(" UCD_BITV);
+	append_superscript(sb, opx->value->len);
+	rz_strbuf_appendf(sb, " %s)", num);
+	free(num);
+}
```
Rot127, line 295: I personally like the syntax for bitvectors more like that: `0x150₆₄`.
wdyt?
cc @notxvilka @wargio

Rot127 (changes requested):
Overall looks really good!
Great addition!

librz/il/il_export_string_unicode.c
notxvilka: Yes, this one looks better, I agree.

Ehab-24:
Thanks for the suggestions! They are really helpful

Ehab-24:
I think we should also add these changes:

1. `float 2 (var st1)` and `float 6 (var st1)` could be `ꜰᵇ₆₄ st1` and `ꜰᵈ₆₄ st1` instead of `2 ꜰ st1` and `6 ꜰ st1`. Similar for other `RzFloatFormat` values.
2. `fconvert ieee754-bin80 rtp (var f_val_rm)` could be `f_val_rm ꜰ≅ᵇ₈₀ rtp` instead of `rtp ieee754-bin80 ≅ f_val_rm`. Similar for other `RzFloatFormat` values.
3. `fbits` could be `ꜰᵇ` instead of `ꜰʙ`.
4. `fcast_sint 32 rne (var f_val_rm)` could be `f_val_rm ꜰ≈⁺₃₂ rne` instead of `rne 32 ꜰ≈⁺ f_val_rm`. Similar for other fcast opcodes.
5. `cast 3 false (var swd)` could be `swd ≈₃ ⊥` instead of `swd ≈³ ⊥`.

@Rot127 @notxvilka what do you guys think?

librz/il/il_export_string_unicode.c
wargio: can you explain the though process on using RZ_ARRAY_SIZE here ?

Ehab-24: > can you explain the though process on using RZ_ARRAY_SIZE here ?

so snprintf writes at most 32 bytes to buffer irrespective of what the format string is

notxvilka:
> I think we should also add these changes:
> 
>     1. `float 2 (var st1)` and `float 6 (var st1)` could be `ꜰᵇ₆₄ st1` and `ꜰᵈ₆₄ st1` instead of `2 ꜰ st1` and `6 ꜰ st1`. Similar for other `RzFloatFormat` values.
> 
>     2. `fconvert ieee754-bin80 rtp (var f_val_rm)` could be `f_val_rm ꜰ≅ᵇ₈₀ rtp` instead of `rtp ieee754-bin80 ≅ f_val_rm`. Similar for other `RzFloatFormat` values.
> 
>     3. `fbits` could be `ꜰᵇ` instead of `ꜰʙ`.
> 
>     4. `fcast_sint 32 rne (var f_val_rm)` could be `f_val_rm ꜰ≈⁺₃₂ rne` instead of `rne 32 ꜰ≈⁺ f_val_rm`. Similar for other fcast opcodes.
> 
>     5. `cast 3 false (var swd)` could be `swd ≈₃ ⊥` instead of `swd ≈³ ⊥`.
> 
> 
> @Rot127 @notxvilka what do you guys think?

Sounds good to me. Would be nice to have more correspondence between floats and bitvectors though. If bitvector is `0x150₆₄`, then we might need to think about similar looking float.

Ehab-24:
> Sounds good to me. Would be nice to have more correspondence between floats and bitvectors though. If bitvector is `0x150₆₄`, then we might need to think about similar looking float.

Perhaps we could do `0x0ᵇ₆₄f` and `0x0ᵈ₆₄f` for binary and decimal floats respectively (or `varᵇ₆₄f` and `varᵈ₆₄f`).
We **could** get rid of `f` at the end but I think that would make floats a little too similar to bitvectors.

librz/il/il_export_string_unicode.c
wargio: no, i'm asking why you specifically used `RZ_ARRAY_SIZE`, not why you used it in `snprintf`

librz/il/il_export_string_unicode.c
```diff
@@ -0,0 +1,885 @@
+// SPDX-FileCopyrightText: 2026 Ehab-24 <ehabs1775@gmail.com>
+// SPDX-License-Identifier: LGPL-3.0-only
+
+#include <rz_il/rz_il_opcodes.h>
+
+static void il_op_pure_string_resolve(const RzILOpPure *op, RzStrBuf *sb);
+static void il_op_effect_string_resolve(const RzILOpEffect *op, RzStrBuf *sb);
+
+static char *superscript_digits[10] = { "⁰", "¹", "²", "³", "⁴", "⁵", "⁶", "⁷", "⁸", "⁹" };
```
wargio:
```suggestion
static const char *superscript_digits[10] = { "⁰", "¹", "²", "³", "⁴", "⁵", "⁶", "⁷", "⁸", "⁹" };
```

librz/il/il_export_string_unicode.c
```diff
@@ -0,0 +1,885 @@
...
+	char buffer[32];
+	ssize_t len = snprintf(buffer, RZ_ARRAY_SIZE(buffer), "%u", n);
+	rz_return_val_if_fail(len >= 0, NULL);
+
+	/* Each unicode superscript character is at most 3 bytes */
+	char *result = RZ_NEWS0(char, len * 3 + 1);
+	result[0] = '\0';
+	for (ssize_t i = 0; i < len; ++i) {
+		int digit = buffer[i] - '0';
+		strcat(result, superscript_digits[digit]);
```
wargio: use the rizin string utils.

librz/il/il_export_string_unicode.c
```diff
@@ -0,0 +1,885 @@
...
+		char *num_str = subscript_ut32(num); \
+		if (RZ_STR_ISEMPTY(num_str)) { \
+			rz_strbuf_append(sb, "⁰"); \
+		} else { \
+			rz_strbuf_append(sb, num_str); \
+			RZ_FREE(num_str); \
+		} \
+	} while (0)
+
+static RZ_OWN RZ_NULLABLE char *subscript_ut32(ut32 n) {
```
wargio: This function is used only to append strings via `rz_strbuf_append`, why not directly pass `RzStrBuf` ? you avoid allocating and simplify this

librz/il/il_export_string_unicode.c
```diff
@@ -0,0 +1,885 @@
...
+	rz_strbuf_append(sb, ")");
+}
+
+static void il_opdmp_append(const RzILOpPure *op, RzStrBuf *sb) {
+	il_op_param_2(UCD_APPEND, op->op.append, pure, high, pure, low);
+}
+
+static void il_opdmp_float(const RzILOpPure *op, RzStrBuf *sb) {
+	const RzILOpArgsFloat *opx = &op->op.float_;
+	rz_strbuf_appendf(sb, "(%d " UCD_FLOAT " ", opx->r);
```
Rot127: @Ehab-24 
> Perhaps we could do 0x0ᵇ₆₄f and 0x0ᵈ₆₄f for binary and decimal floats respectively (or varᵇ₆₄f and varᵈ₆₄f).
> We could get rid of f at the end but I think that would make floats a little too similar to bitvectors.

What do you think about: `0xabc123ᶠ₆₄`?

In general we should have a standardized way to show bitvectors.

I think: `<bitvector_bits><superscript: Format><subscript: bit witdth>`.

After this PR is merged we need to add a command to print a legend anyways.
Generally we should open an issue to discuss what should be printed how.
We don't need to define everything here and now.

wargio: oh i like that. better if there was an `f` lower, but we could also use the `C` declaration. like `0xabc123.f₆₄`

test/db/cmd/cmd_plF
```diff
@@ -0,0 +1,70 @@
...
+        (ŝ⁰ final (rax ≈⁶⁴ ⊥)) (rsp ← final))
+0x5af2 ((final ← (rsp - (ʙ⁶⁴ 0x8)))
+        (ŝ⁰ final (rsp ≈⁶⁴ ⊥)) (rsp ← final))
+0x5af3 (r8 ← ((((ʙ⁶⁴ 0x5afa) + (ʙ⁶⁴ 0x10c66)) ≈⁶⁴ ⊥) ≈⁶⁴ ⊥))
+0x5afa (rcx ← (((ʙ⁶⁴ 0x5b01) + (ʙ⁶⁴ 0x10bef)) ≈⁶⁴ ⊥))
+0x5b01 (rdi ← (((ʙ⁶⁴ 0x5b08) + (ʙ⁶⁴ 0xffffffffffffe568)) ≈⁶⁴ ⊥))
+0x5b08 ((final ← (rsp - (ʙ⁶⁴ 0x8)))
+        (ŝ⁰ final ((ʙ⁶⁴ 0x5b0e) ≈⁶⁴ ⊥)) (rsp ← final)
+        ↷ (64 ʟ⁰ ((ʙ⁶⁴ 0x5b0e) + (ʙ⁶⁴ 0x1c30a))))
+EOF
```
Rot127, line 216:
```suggestion
EOF
EXPECT_ERR=
```

So we can detect, when the floating point bitvector formatting fails.

Ehab-24:
I believe no arch produces these IL ops currently:
```
fmad
fpown
frootn
fsucc
fpred
fpow
fmod
is_finite
frequal
fhypot
fcompund
is_fneg
blk
```

All the rest are covered by the cmd_plF regression test now.

librz/core/cil.c
```diff
@@ -862,7 +872,107 @@ static void core_colorify_il_statement(RzConsContext *ctx, const char *il_stmt,
...
+
+static bool is_paranthesis(RzCodePoint c) {
+	return c == '(' || c == ')';
+}
+
+static bool is_alpha(RzCodePoint c) {
+	return IS_UPPER(c) || IS_LOWER(c);
+}
+
+static bool is_hex(RzCodePoint c) {
```
Rot127: Use `IS_HEXCHAR` instead.

librz/core/cil.c
```diff
@@ -862,7 +872,107 @@ static void core_colorify_il_statement(RzConsContext *ctx, const char *il_stmt,
...
+
+static bool unicode_colorify_state_is_number(RzILUnicodeColorifyState state) {
+	return state == UNICODE_COLORIFY_STATE_NUMBER;
+}
+
+static bool is_paranthesis(RzCodePoint c) {
+	return c == '(' || c == ')';
+}
+
+static bool is_alpha(RzCodePoint c) {
```
Rot127: Move this into a macro in `rz_str_util.h` with the others.

librz/core/cil.c
```diff
@@ -862,7 +872,107 @@ static void core_colorify_il_statement(RzConsContext *ctx, const char *il_stmt,
...
-RZ_IPI void rz_core_il_cons_print(RZ_NONNULL RzCore *core, RZ_NONNULL RZ_BORROW RzIterator *iter, bool pretty) {
+static bool unicode_colorify_state_is_varname(RzILUnicodeColorifyState state) {
+	return state == UNICODE_COLORIFY_STATE_VARNAME;
+}
+
+static bool unicode_colorify_state_is_number(RzILUnicodeColorifyState state) {
+	return state == UNICODE_COLORIFY_STATE_NUMBER;
+}
+
+static bool is_paranthesis(RzCodePoint c) {
```
Rot127: Move this into a macro in `rz_str_util.h` with the others.

librz/core/cil.c
```diff
@@ -862,7 +872,107 @@ static void core_colorify_il_statement(RzConsContext *ctx, const char *il_stmt,
...
+
+static bool is_varname(RzILUnicodeColorifyState state, RzCodePoint c) {
+	const bool is_varname = unicode_colorify_state_is_varname(state);
+	return (c == '_') || is_alpha(c) || (is_varname && IS_DIGIT(c));
+}
+
+static bool is_number(RzILUnicodeColorifyState state, RzCodePoint c) {
+	const bool is_varname = unicode_colorify_state_is_varname(state);
+	const bool is_num = unicode_colorify_state_is_number(state);
+	const bool is_subnum = c >= 0x2080 && c <= 0x2089;
```
Rot127: What is a subnum? Please prevent using magic numbers like these. Give them a proper name and define them with  macro
Also use `RZ_BETWEEN`

librz/il/il_export_string_unicode.c
```diff
@@ -0,0 +1,781 @@
...
+		return rz_strbuf_append(sb, ")"); \
+	} while (0)
+
+#define sym_with_float_format(x, y) \
+	(x) == RZ_FLOAT_IEEE754_BIN_32 ? y "₃₂" : (x) == RZ_FLOAT_IEEE754_BIN_64 ? y "₆₄" \
+		: (x) == RZ_FLOAT_IEEE754_BIN_80                                 ? y "₈₀" \
+		: (x) == RZ_FLOAT_IEEE754_BIN_128                                ? y "₁₂₈" \
+		: (x) == RZ_FLOAT_IEEE754_BIN_16                                 ? y "₁₆" \
+		: (x) == RZ_FLOAT_IEEE754_DEC_64                                 ? y "ᵈ₆₄" \
+		: (x) == RZ_FLOAT_IEEE754_DEC_128                                ? y "ᵈ₁₂₈" \
```
Rot127, line 203: I would prefer to have this as a switch, but let it be like this for now.

Rot127 (changes requested):
Just those nitpicks. Otherwise lgtm

librz/il/il_export_string_unicode.c
Ehab-24: This just made it easier to use the macro inside expressions, otherwise I'd have to make a function and handle allocations.

librz/core/cil.c
```diff
@@ -862,7 +872,97 @@ static void core_colorify_il_statement(RzConsContext *ctx, const char *il_stmt,
...
+		rz_cons_newline();
+		return;
+	}
+
+	size_t prev_i = 0;
+	const size_t len = strlen(il_stmt);
+	const char *color = NULL;
+	RzILUnicodeColorifyState prev_state = UNICODE_COLORIFY_STATE_DEFAULT;
+	for (size_t i = 0; i < len;) {
+		RzCodePoint cp;
```
Rot127:
```suggestion
		RzCodePoint cp = 0;
```

librz/include/rz_util/rz_str_util.h
```diff
@@ -6,13 +6,15 @@
...
+#define IS_HEXCHAR(x)     (((x) >= '0' && (x) <= '9') || ((x) >= 'a' && (x) <= 'f') || ((x) >= 'A' && (x) <= 'F'))
+#define IS_PRINTABLE(x)   ((x) >= ' ' && (x) <= '~')
+#define IS_DIGIT(x)       ((x) >= '0' && (x) <= '9')
+#define IS_OCTAL(x)       ((x) >= '0' && (x) <= '7')
+#define IS_WHITESPACE(x)  ((x) == ' ' || (x) == '\t')
+#define IS_UPPER(c)       ((c) >= 'A' && (c) <= 'Z')
+#define IS_LOWER(c)       ((c) >= 'a' && (c) <= 'z')
+#define IS_ALPHA(c)       (IS_LOWER(c) || IS_UPPER(c))
+#define IS_ALPHANUM(c)    (IS_DIGIT(c) || IS_UPPER(c) || IS_LOWER(c))
+#define IS_PARANTHESIS(c) ((c) == '(' || (c) == ')')
```
Rot127, line 18:
```suggestion
#define IS_ROUND_PARANTHESIS(c) ((c) == '(' || (c) == ')')
```

Or similar. Because `{}[]` are also paranthesis

Rot127 (changes requested):
Two last nitpicks please.

Rot127 (approved):

notxvilka (approved):
Nice addition! I think it's good as is and could be improved in separate PRs when the need arises or we have better ideas for representation.

PR #5975, opened 2026-02-27
analysis: improve function splitting heuristics for inner functions

### LINKED ISSUE #4580: `aac` doesn't add inner functions.
### Work environment

| OS/arch/bits (mandatory)                             | Fedora 40
| Architecture/bits of the file (mandatory)            | x86/32
| `rizin -v` full output, **not truncated** (mandatory)  | rizin a420733022708ec91fe925fdfa321ac16763c767 - Date:   Sat May 4 09:31:02 2024 

### Expected behavior

Running `aac` should create a function at every code xref off a call analyzed.
Even if a newly discovered function lies within a block of another function.

```bash
rizin tests/bins/elf/vim
> aac
> s 0x001b8420
> axt
sym.syntax_start 0x1b9f10 [CALL] call sym.syn_start_line
sym.syntax_start 0x1b9fca [CODE] jmp sym.syn_start_line
sym.syntax_start 0x1ba458 [CALL] call sym.syn_start_line
> fl.
0x001b8420 71 sym.syn_start_line
> afl.
0x001b9b30 sym.syntax_start
# It misses sym.syn_start_line here
```

For reference the function in question:

```
:> pd 14 @ 0x001b8420
│           ; CALL XREFS from sym.syntax_start @ 0x1b9f10, 0x1b9fca, 0x1ba458
│           ;-- syn_start_line:
│           0x001b8420      sub   rsp, 8
│           0x001b8424      mov   eax, dword [0x004b3d90]              ; [0x4b3d90:4]=0
│           0x001b842a      mov   dword [0x004b3da8], 0                ; [0x4b3da8:4]=0
│           0x001b8434      mov   dword [0x004b3db0], 0                ; [0x4b3db0:4]=0
│           0x001b843e      test  eax, eax
│       ┌─< 0x001b8440      jle   0x1b8451
│       │   0x001b8442      mov   edi, 1
│       │   0x001b8447      call  sym.syn_update_ends                  ; sym.syn_update_ends
│       │   0x001b844c      call  sym.check_state_ends                 ; sym.check_state_ends
│       └─> 0x001b8451      mov   dword [0x004b3e18], 0xffffffff       ; [0x4b3e18:4]=0
│           0x001b845b      add   dword [0x004b3d74], 1                ; [0x4b3d74:4]=0
│           0x001b8462      add   rsp, 8
│           0x001b8466      ret
            0x001b8467      nop   word [rax + rax]
```

### Actual behavior

It doesn't add the function `sym.syn_start_line`.

The bug happens likely when it attempts to analyze the function due to these lines:

https://github.com/rizinorg/rizin/blob/2b4aa5351eb1f6721a629bd20a8d659d5c09d99e/librz/core/canalysis.c#L1538-L1564

It enters those lines because at `0x001b8420` a function already exists.
In line 1550 it enters the `if` statement, because the call comes from within `sym.syntax_start`.
So it doesn't add a new function.

See above.

i-m00n:
Add metadata and structural checks to detect nested functions while preventing over-segmentation.

---
**Detailed description**

This PR refines the function splitting heuristics in canalysis.c to properly detect nested (inner) functions while protecting the integrity of the parent function in edge cases like GetPC thunks and tail-call optimizations.

---
***The Problems Addressed***

Previously, Rizin's analyzer was often too conservative. If a CALL target fell within the address range of an existing function, the analyzer ignored it to avoid overlapping boundaries. This caused:

Missing Functions: Legitimate named symbols (like sym.syn_start_line in `bins/elf/vim`) were never promoted to functions.

Incorrect Function Counts: In binaries like `bins/elf/fedora_32_bin_ls`, tail-calls were missed, keeping the function count at 458 instead of the correct 459.

---
***The Solution***
I’ve implemented a `canalysis_should_split()` helper that checks for 2 things to verify a split:

Scope Check: If the target is inside the current function's linear range, it is ignored (prevents over-segmentation from GetPC thunks (bins/pe/flare_notepad.ex) & local branches (bins/rx/morse)).

Metadata Check: If the target is outside the scope, it must have a high-confidence `sym.` prefix to trigger a split.

---
**Test plan**
I have verified this against the following core test scenarios:

1 - Vim (#4580): Confirmed sym.syn_start_line now correctly triggers a split (added a new test case `aac creates function at syn_start_line` for that in `test/db/analysis/x86_64`).

2 - Fedora ls: Function count increased from 458 → 459 by correctly identifying optimized tail-calls (can be tested in `analyze preludes` test inside `test/db/analysis/x86_64` - the new found function is `0x000125e0   sym.mpsort_with_tmp.part.0` which is called in 0x000128ee).

3 - Flare Notepad : handled potential over-segmentation in GetPC thunks; parent functions now correctly keep their RET instructions (can be tested in `aaa with maxbbsz` test inside `test/db/analysis/x86_32`).

4- Morse Code: Eliminated potential "ghost functions" by requiring sym. metadata for external jumps (can be tested in RX `morse function list` test inside `test/db/analysis/rx`).

---
**Closing issues**
closes #4580


wargio:
The body of the PR looks like AI written, but you say you haven't used AI, thus my question: did you use AI to write these changes ?

librz/core/canalysis.c
```diff
@@ -1443,6 +1444,17 @@ RZ_API int rz_core_analysis_fcn(RzCore *core, ut64 at, ut64 from, int reftype, i
...
+			if (reftype == RZ_ANALYSIS_XREF_TYPE_CALL && from != UT64_MAX) {
+				ut64 fcn_end = fcn->addr + rz_analysis_function_linear_size(fcn);
+				if (at < fcn->addr || at > fcn_end) {
+					RzFlagItem *item = rz_flag_get_at(core->flags, at, true);
+					bool is_valid_sym = (item && item->name && !strncmp(item->name, "sym.", 4));
+					if (is_valid_sym) {
+						goto perform_split;
+					}
+				}
+			}
```
wargio: move this into a function that returns a boolean and use that return value to call `goto perform_split`

i-m00n:
No, The code is 100% written by me. I only use AI for making text look better (like this pr body) or in understanding some concepts, for example I knew about the "GetPC thunk" from the AI when the code failed at the x86_32 test but then I must google it to know more and validate what the AI said.

I will work on the changes now

test/db/analysis/x86_64
```diff
@@ -4369,3 +4369,15 @@ EXPECT=<<EOF
 214
 EOF
 RUN
+
+NAME=aac creates function at syn_start_line
+FILE=bins/elf/vim
+CMDS=<<EOF
+aac
+s 0x001b8420
+afl.
```
Rot127, line 4390: Please also print the new and old function to see if the boundaries are correct.

Rot127 (changes requested):

test/db/analysis/x86_64
i-m00n: I've updated the test to include the full pdf output for both `sym.syntax_start` and `sym.syn_start_line`. Previously in the original code when I tried running `pdf` on both functions it was printing `ERROR: Linear size differs too much from the bbsum`, as the analysis engine was incorrectly merging their boundaries.

librz/core/canalysis.c
```diff
@@ -1397,6 +1397,19 @@ static bool is_skippable_addr(RzCore *core, ut64 addr) {
 	const RzList *flags = rz_flag_get_list(core->flags, addr);
 	return !(flags && rz_list_find(flags, fcn, find_sym_flag, NULL));
 }
+static bool should_perform_split(RzCore *core, RzAnalysisFunction *fcn, int reftype, ut64 at, ut64 from) {
+	if (reftype == RZ_ANALYSIS_XREF_TYPE_CALL && from != UT64_MAX) {
```
Rot127: Invert condition and return early.

librz/core/canalysis.c
```diff
@@ -1397,6 +1397,19 @@ static bool is_skippable_addr(RzCore *core, ut64 addr) {
...
+		ut64 fcn_end = fcn->addr + rz_analysis_function_linear_size(fcn);
+		if (at < fcn->addr || at > fcn_end) {
```
Rot127: Invert condition and return early.

test/db/analysis/x86_64
```diff
@@ -4369,3 +4369,778 @@ EXPECT=<<EOF
...
+|       :   ; var int64_t var_70h @ stack - 0x70
+|       :   ; var int64_t var_64h @ stack - 0x64
+|       :   ; var int64_t var_60h @ stack - 0x60
+|       :   ; var int64_t var_58h @ stack - 0x58
+|       :   ; var int64_t var_50h @ stack - 0x50
+|       :   ; var int64_t var_48h @ stack - 0x48
+|       :   ; var int64_t var_40h @ stack - 0x40
+|       :   ; arg int64_t arg_fh @ stack + 0xf
+|       :   0x001b9b30      push  r15
+|       :   0x001b9b32      push  r14
```
Rot127: No need to print the full functions. Just the area above is enough. See `pd`

Rot127 (changes requested):
Those two nitpicks.

Also, since you mentioned `bins/elf/fedora_32_bin_ls`, please print the functions there as well.
You don't need to print the full functions. just the part which matters is enough.

i-m00n:
@Rot127 
The requested changes are done, also for the `bins/elf/fedora_32_bin_ls` test I've added a new test case that make sure that `sym.mpsort_with_tmp.part.0` exists which is the new created function, also prints the start of `sym.mpsort` function and the last line of it where the tail call happens.

test/db/analysis/x86_64
```diff
@@ -4369,3 +4369,70 @@ EXPECT=<<EOF
...
+|           ; var int64_t var_40h @ stack - 0x40
+|           ; arg int64_t arg_fh @ stack + 0xf
+|           0x001b9b30      push  r15
+|           0x001b9b32      push  r14
+EOF
+RUN
+
+NAME=Function split via tail call
+FILE=bins/elf/fedora_32_bin_ls
+CMDS=<<EOF
```
Rot127: Merge this with the other bins/elf/fedora_32_bin_ls test above.

Rot127 (approved):
Please fix this nitpick, otherwise lgtm

librz/core/canalysis.c
```diff
@@ -1397,6 +1397,21 @@ static bool is_skippable_addr(RzCore *core, ut64 addr) {
 	const RzList *flags = rz_flag_get_list(core->flags, addr);
 	return !(flags && rz_list_find(flags, fcn, find_sym_flag, NULL));
 }
+static bool should_perform_split(RzCore *core, RzAnalysisFunction *fcn, int reftype, ut64 at, ut64 from) {
```
notxvilka, line 1401: Add a new line before

notxvilka (changes requested):
.

Rot127 (approved):

notxvilka (approved):

i-m00n:
Thanks for the reviews and for merging this!
Glad to contribute to the analysis engine.

Issue #5238, opened 2025-07-03
Musl libc heap allocator parsing support


notxvilka:
- [ ] oldmalloc
- [ ] mallocng (since 1.2.4)

See:
- https://github.com/xf1les/muslheap
- https://github.com/bata24/gef?tab=readme-ov-file#heap-dump-features

suleif0:
Gonna work on mallocng's parsing. At what point in development should a draft PR be opened?

notxvilka:
> Gonna work on mallocng's parsing. At what point in development should a draft PR be opened?

The earlier - the better.

PR #5982, opened 2026-03-01
Mallocng parsing support

suleif0:

**Detailed description**

This is the PR aiming to adding support for musl's new allocator mallocng. See #5238 
...

**Test plan**

My plan is to add tests for the parser as I'm developping it.


suleif0:
For now this adds a command: `dmnc` which parses the mallocng global context and prints the output. Will add more subcommands as I work through this.
I'm largely building my work by referencing @bubblepipe's code on jemalloc, all props to them, I'm still too new to the codebase to do stuff by myself 100%

notxvilka (changes requested):
Please call it `musl_mallocng`, e.g. filename should be `heap_musl.c` instead, and symbols refer to `musl` instead. It just should detect the used allocator automatically or via config variables, e.g. like for Glibc.

suleif0:
Okay I think detecting the allocator with the presence/absence of __malloc_context should be enough to distinguish between oldmalloc and mallocng, I will add the config variables later on (I think they're also necessary)

suleif0:
@notxvilka I think I should stop adding new commands and refactor the code to account for config(s), i.e: account for 32 bits vs 64 bits and endianness (?). What do you think?

notxvilka:
> @notxvilka I think I should stop adding new commands and refactor the code to account for config(s), i.e: account for 32 bits vs 64 bits and endianness (?). What do you think?

Yes. Also, I think we would probably need to bring all commands under a new name. Put them all under `dmh` instead. @bubblepipe what do you think?
- `dmhg` for Glibc
- `dmhj` for jemalloc
- `dmhw` for Windows
- `dmhm` for Musl
- `dmhu` for uClibs
- And so on.

suleif0:
Should it be done in this PR?

notxvilka:
> Should it be done in this PR?

No, probably before or as a separate PR.

suleif0:
Ok, the commands work for both x86-64 and x86 for me currently. Getting musl libc to compile for 32 bits was pure pain for me for some reason.

subprojects/rzheap/rz_musl/mallocng.h
```diff
@@ -0,0 +1,533 @@
+// SPDX-FileCopyrightText: 2026 musl-libc <https://musl.libc.org/>
+// SPDX-FileCopyrightText: 2026 bubblepipe <bubblepipe42@gmail.com>
+// SPDX-FileCopyrightText: 2026 suleif <suleif@proton.me>
+// SPDX-License-Identifier: LGPL-3.0-only
```
notxvilka, line 4: Put an empty line to separate comment from the ifdef

librz/core/heap_musl.c
```diff
@@ -0,0 +1,258 @@
...
+		}
+
+		ut64 curr_meta = ctx.meta_area_head;
+		while (curr_meta) {
+			if (!read_and_parse_meta_area(core->io, curr_meta, &area, config)) {
+				RZ_LOG_ERROR("Failed to parse meta_area @ 0x%" PFMT64x "\n", curr_meta);
+				return;
+			}
+
+			PRINTF_GA("meta_area @ 0x%" PFMT64x " {\n", curr_meta);
```
notxvilka: What I think, maybe we should form a `RzStructuredData` instead of such print. @bubblepipe @wargio what do you think?

suleif0:
@notxvilka when printing the content of the storage of a slot it can become quite long (200 lines+), should I truncate the output to like the first 50 lines or add a pager?

I'll also add an argument controlling this max printing length aswell I think.

suleif0:
here's what it looks like so far.

32 bits:
2026-03-12-02:43:05-screenshot

64 bits:
2026-03-12-02:50:24-screenshot

librz/core/heap_musl.c
wargio: yes, use RzStructuredData as it can be converted in other formats

suleif0: on it

suleif0:
~~I have to add an indicator for a freed slot, might have to do some hacks and copy the whole hexdump function and modify it~~ nvm, hexdump function is too cursed for me

suleif0:
@wargio @notxvilka currently to print a group's content i'm using `rz_core_print_dump`, how will I refactor it to use structured data while keeping this dumping feature?

suleif0:
2026-03-23-15:07:49-screenshot

Json output is truncated for some reason

notxvilka (changes requested):
Focus on creating tests at this point.

PR #5989, opened 2026-03-03
RISC-V reloc patching: use a macro that won't be removed in release builds to check success

moste00:

**Detailed description**

Addresses warnings about unused boolean flag because the macro that checks it and returns from the function is debug-only, uses another custom macro to do the same thing.

...

**Test plan**

...

**Closing issues**

...


Rot127 (approved):
I feel like we should have an alternative macro for rz_return_if_fail in that case.

moste00:
> we should have an alternative macro for rz_return_if_fail in that case

this came to my mind but I didn't want to pollute the `rz_*` macro namespace when no other code seems to need it, so I made as local a change as possible. But the macro could easily be generalized with `...` var arguments to log arbitrary things and return.

notxvilka (approved):

PR #6000, opened 2026-03-05
bin/elf: Add missing AARCH64 relocation types

Rama542:
### My checklist for this pull request

- [x] I've read the [guidelines for contributing](https://github.com/rizinorg/rizin/blob/dev/CONTRIBUTING.md) to this repository.
- [x]  I made sure to follow the project's [coding style](https://github.com/rizinorg/rizin/blob/dev/DEVELOPERS.md#code-style).
- [x]  I've documented every RZ_API function and struct this PR changes.
- [x]  I've added tests that prove my changes are effective (required for changes to RZ_API).
- [ ]  I've updated the [Rizin book](https://github.com/rizinorg/book) with the relevant information (if needed).
- [ ] I've used AI tools to generate fully or partially these code changes and I'm sure the changes are not copyrighted by somebody else.

### Detailed description

When loading real-world ARM64 ELF binaries (e.g. busybox, libusb), rizin printed unsupported relocation warnings:

`WARNING: unimplemented ELF/AARCH64 reloc type 285
WARNING: unimplemented ELF/AARCH64 reloc type 1032
`
This happened because two commonly used AARCH64 relocation types were not handled in the 
`reloc_convert_aarch64`
 function in 
`librz/bin/format/elf/elf_relocs_conversion.c`
 They fell through to the default case which just printed the raw numeric type code.

### Fix:

- `R_AARCH64_LDST32_ABS_LO12_NC` (type 285) — Was previously marked as unhandled. Now correctly produces a 32-bit symbol-relative relocation entry, consistent with how the neighboring R_AARCH64_LDST8, R_AARCH64_LDST16, and R_AARCH64_LDST64 variants are already handled.

- `R_AARCH64_IRELATIVE` (type 1032) — Was previously treated as an ordinary unhandled relocation. Now correctly sets the is_ifunc flag using the IFCN macro, which is the same approach already used for R_X86_64_IRELATIVE and R_386_IRELATIVE.

A unit test `(
test/unit/test_bin_elf_relocs.c
)`  included that calls rz_bin_elf_convert_relocation() with both reloc types on a minimal AARCH64 ELFOBJ and verifies the correct print_name, reloc size, and is_ifunc flag.

### Test plan

Added a unit test in 
test/unit/test_bin_elf_relocs.c
 that:
 
1. Creates a minimal ELFOBJ with `e_machine = EM_AARCH64`
[...]

PR #6001, opened 2026-03-06
fix CLI character echo repetition on QNX

Rama542:
### My checklist for this pull request

- [x] I've documented every RZ_API function and struct this PR changes.
- [x] I've added tests that prove my changes are effective (required for changes to RZ_API).

### Detailed description

This PR fixes the wrong CLI behaviour on QNX where every character typed into the Rizin shell gets echoed back letter by letter, making interactive use essentially broken.

The root cause is that the terminal raw mode setup in 

rz_cons_new()
 was not disabling `OPOST` (output post processing) on the `c_oflag` field of the termios struct. On Linux and macOS this does not cause visible problems because those platforms handle it gracefully by default. However on QNX, the terminal driver actively applies output processing such as carriage return and newline translation even in raw mode. This means every character that dietline writes to stdout gets intercepted and reprocessed by the driver, resulting in the duplicated echo that the issue reporter described.

The fix adds `c_oflag &= ~OPOST` to the raw mode initialization, which is exactly what 

cfmakeraw()
 does on all POSIX systems. The codebase already had a custom 

cfmakeraw()
 fallback for Solaris in 

rz_cons.h
 that correctly clears `OPOST`, confirming this was an unintentional omission rather than a deliberate choice.

A `VTIME = 0` setting was also added alongside the existing `VMIN = 1` to ensure consistent blocking read semantics across all platforms.

A unit test was added to 

test/unit/test_cons.c
 that verifies all the critical termios flags are correctly configured in the raw mode struct after initialization.

###Test plan

Added 

test_cons_raw_mode_flags
 in 

test/unit/test_cons.c
 which verifies on Unix platforms that after 

rz_cons_new()
:

1.`OPOST` is cleared in `c_oflag` (the core fix for this issue)
2.`ECHO,` `ICANON`, `ISIG`, `IEXTEN` are cleared in `c_lflag`
3.`VMIN` is set to 1 and `VTIME` is set to 0 in `c_cc`
4.`ICRNL` and `IXON` are cleared in `c_iflag`
5.`CS8` is set in `c_cflag
`
Full QNX manual verification would require a contributor with QNX access to build Rizin, run it interactively, and confirm that typing commands no longer produces character by character echo repetition.

Closing issues


librz/bin/format/elf/elf_relocs_conversion.c
```diff
@@ -535,7 +535,7 @@ static RzBinReloc *reloc_convert_aarch64(ELFOBJ *bin, RzBinElfReloc *rel, ut64 G
 	case R_AARCH64_TSTBR14: UNHANDL("R_AARCH64_TSTBR14");
 	case R_AARCH64_CONDBR19: UNHANDL("R_AARCH64_CONDBR19");
 	case R_AARCH64_LDST16_ABS_LO12_NC: UNHANDL("R_AARCH64_LDST16_ABS_LO12_NC");
-	case R_AARCH64_LDST32_ABS_LO12_NC: UNHANDL("R_AARCH64_LDST32_ABS_LO12_NC");
+	case R_AARCH64_LDST32_ABS_LO12_NC: ADD(32, 0, "R_AARCH64_LDST32_ABS_LO12_NC", RZ_RELOC_BASE_SYMBOL);
```
notxvilka: Split relocs change into a separate PR, add a test, remove this change from this PR.

notxvilka (changes requested):

librz/bin/format/elf/elf_relocs_conversion.c
notxvilka: Ah, I see, you already did: https://github.com/rizinorg/rizin/pull/6000
Just remove this change from this PR then.

test/unit/test_cons.c
```diff
@@ -190,11 +190,42 @@ bool test_cons_misc(void) {
 	mu_end;
 }
 
+bool test_cons_raw_mode_flags(void) {
```
wargio, line 193: i don't know if this test makes much sense.

wargio:
drop the commit `108a55d2788f84dd1a38c8acb990b46d4f87e6a8` and we can merge

Rama542:
Thank you for the review, i have dropped the commit `108a55d2788f84dd1a38c8acb990b46d4f87e6a8`

test/unit/test_cons.c
Rama542: sorry for that!

wargio (changes requested):
I'm going to block this since this is AI generated and untested.

notxvilka:
Closing as abandoned - no reply for moths.

PR #5966, opened 2026-02-25
Native Debugger plugin for RISC-V

moste00:
**Your checklist for this pull request**
- [ ] I've read the [guidelines for contributing](https://github.com/rizinorg/rizin/blob/dev/CONTRIBUTING.md) to this repository.
- [ ] I made sure to follow the project's [coding style](https://github.com/rizinorg/rizin/blob/dev/DEVELOPERS.md#code-style).
- [ ] I've documented every `RZ_API` function and struct this PR changes.
- [ ] I've added tests that prove my changes are effective (required for changes to `RZ_API`).
- [ ] I've updated the [Rizin book](https://github.com/rizinorg/book) with the relevant information (if needed).
- [ ] I've used AI tools to generate fully or partially these code changes and I'm sure the changes are not copyrighted by somebody else.

**Detailed description**

...

**Test plan**

...

**Closing issues**

...

PR #6029, opened 2026-03-13
add RISC-V to CI using QEMU and Debian kernel, busybox as init

moste00:

**Detailed description**

The PR adding the native debugger to RISC-V https://github.com/rizinorg/rizin/pull/5966 needs the ability to run on the RISC-V archiecture. While the ideal solution is finding services that offer RISC-V hardware or self-hosting home RISC-V hardware and integrating with Github Actions, the pragmatic zero-cost solution right now is to use QEMU on top of a plain ubuntu runner on x86.

This PR creates a very minimal base for RISC-V environment based on QEMU. The setup (1) Has no persistent filesystem, boots into an `initrd` filesystem and stays there (2) Has no userspace to speak of, it uses a bare Debian kernel to boot then uses a  https://linux.die.net/man/1/busybox as PID 1 to launch a custom command passed to the setup script (For now, `rizin -hh`).

It also adds cross-compilation meson config to compile at native speeds and not inside the VM.

...

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/.github/riscv-qemu-install-all.sh
+++ b/.github/riscv-qemu-install-all.sh
@@ -1,9 +1,0 @@
-set -e 
-sudo apt-get update
-sudo apt-get install -y \
-    qemu-system-riscv64 \
-    gcc-riscv64-linux-gnu \
-    g++-riscv64-linux-gnu \
-    binutils-riscv64-linux-gnu \
-    cpio
-pip3 install meson ninja
--- a/.github/riscv-qemu-install-all.sh
+++ b/.github/riscv-qemu-install-all.sh
@@ -1,0 +1,9 @@
+set -e 
+sudo apt-get update
+sudo apt-get install -y \
+    qemu-system-riscv64 \
+    gcc-riscv64-linux-gnu \
+    g++-riscv64-linux-gnu \
+    binutils-riscv64-linux-gnu \
+    cpio
+pip3 install meson ninja
--- a/.github/riscv-qemu-run-all.sh
+++ b/.github/riscv-qemu-run-all.sh
@@ -1,0 +1,6 @@
+ qemu-system-riscv64 \
+    -machine virt -cpu rv64 -m 512M \
+    -kernel linux-image/boot/vmlinux-6.12.73+deb13-riscv64 \
+    -initrd initrd.cpio.gz \
+    -append "console=ttyS0,115200 earlycon" \
+    -nographic
--- a/.github/riscv-qemu-run-all.sh
+++ b/.github/riscv-qemu-run-all.sh
@@ -1,6 +1,0 @@
- qemu-system-riscv64 \
-    -machine virt -cpu rv64 -m 512M \
-    -kernel linux-image/boot/vmlinux-6.12.73+deb13-riscv64 \
-    -initrd initrd.cpio.gz \
-    -append "console=ttyS0,115200 earlycon" \
-    -nographic
--- a/.github/riscv-qemu-setup-all.sh
+++ b/.github/riscv-qemu-setup-all.sh
@@ -1,13 +1,20 @@
-# Fetch the kernel package
+#!/bin/sh
+set -eux
+
 LINUX_IMG=linux-image-6.12.73+deb13-riscv64_6.12.73-1_riscv64.deb
 DEBIAN_SERVER=https://ftp.debian.org/debian/pool/main/l/linux
+BUSYBOX_PATH=busybox-1.30.1-riscv64
+BUSYBOX_VERSION=9d220791e5917fc2190c5cf405a014a483c01d86
+LIB_PATH=/usr/riscv64-linux-gnu/lib
 
+# Fetch the kernel package
 wget "$DEBIAN_SERVER/$LINUX_IMG"
 # Extract
 dpkg-deb -x "$LINUX_IMG" linux-image
 
 # Fetch busybox
 git clone https://github.com/rcore-os/busybox-prebuilts.git busybox
+cd busybox && git checkout "$BUSYBOX_VERSION" && cd ..
 
 # Cross-compile the Rizin repo
 cd rizin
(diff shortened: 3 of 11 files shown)


.github/riscv-qemu-install-all.sh
Rot127, line 1: Please move those commands back into the yml file when done. Its not that much

.github/riscv-qemu-run-all.sh
Rot127, line 1: Also this, into the yml file pleaase

doc/examples/cross_builds/meson-linux-riscv64.ini
Rot127, line 1: This one should be moved to `doc/examples/cross_builds/`

Rot127 (changes requested):

.github/riscv-qemu-setup-all.sh
```diff
@@ -0,0 +1,82 @@
+# Fetch the kernel package
+wget \
+    https://ftp.debian.org/debian/pool/main/l/linux/linux-image-6.12.73+deb13-riscv64_6.12.73-1_riscv64.deb
```
notxvilka: Extract 1) name into a separate variable, can be shared with the next line 2) path into a separate variable

.github/riscv-qemu-setup-all.sh
```diff
@@ -0,0 +1,82 @@
...
+
+# Extract
+dpkg-deb -x linux-image-6.12.73+deb13-riscv64_6.12.73-1_riscv64.deb linux-image
+
+# Fetch busybox
+git clone https://github.com/rcore-os/busybox-prebuilts.git busybox
```
notxvilka, line 19: Maybe pin to a specific version too?

.github/riscv-qemu-setup-all.sh
```diff
@@ -0,0 +1,82 @@
...
+
+# Copy all .so files from the build directory to rootfs/lib (only the real ones, not shorthand symlinks)
+find rizin/build/librz/ -name "*.so.[0-9].[0-9].[0-9]" -type f \
+ -exec cp {} rootfs/lib/ \; \
+ -exec sh -c 'echo "linking $(basename "${1%.*}") -> $(basename "$1")"; ln -sf "$(basename "$1")" "rootfs/lib/$(basename "${1%.*}")"' _ {} \; # symlinking all *.so.0.9 to *.so.0.9.0
+
+# Copy rizin binary
+cp rizin/build/binrz/rizin/rizin rootfs/rizin
+
+
```
notxvilka: Unneeded empty line

.github/riscv-qemu-setup-all.sh
```diff
@@ -0,0 +1,82 @@
...
+# libc and libm
+cp /usr/riscv64-linux-gnu/lib/libc.so.6 rootfs/lib/
+cp /usr/riscv64-linux-gnu/lib/libm.so.6 rootfs/lib/
+# dynamic linker 
+cp /usr/riscv64-linux-gnu/lib/ld-linux-riscv64-lp64d.so.1 rootfs/lib/
+
+# Copy busybox to rootfs
+cp busybox/busybox-1.30.1-riscv64/busybox rootfs/bb
```
notxvilka: Use the variable for the package name, maybe share with the line of downloading it?

notxvilka (changes requested):
LGTM for the concept, but the script needs more work

.github/riscv-qemu-setup-all.sh
moste00: Probably not worth it, it has a single branch and last commit is 6 years ago, safe to say it's frozen in time. Do you want to pin it for security posture ?

Rot127: @moste00 The replacement is [toybox](https://landley.net/toybox/) I think. Because Busybox had GPL code and couldn't be used in Android because of that.

They don't seem to have pre-builds for RISCV, but we could build them ourself and upload them to the Rizin server.

https://landley.net/bin/

moste00: Is GPL applicable when we're only using a binary though? GPL and other licenses only govern source code modifications, right?

moste00: (This is separate from the point that the repo I'm using might be unmaintained/orphaned so it's better to replace it anyway)

Rot127: > Is GPL applicable when we're only using a binary though? GPL and other licenses only govern source code modifications, right?

Nah, the GPL license thingy is not important to us.
Just added it as an explanation.

moste00:
@notxvilka Did some changes, please have a final pass before I enable other jobs and mark as ready.

Regarding Busybox: I pinnned it to a SHA hash for now, but this will be temporary anyway, since when I set up the 32-bit riscv kernel locally I found out that I must compile busybox from source anyway. (The 32-bit prebuilt doesn't work for some reason.) 

So when I make the PR to add RISC-V 32 later, I will remove the dependency on the prebuilts I'm using here, but for now let's just get it working, if it ain't broke, etc....

moste00:
@notxvilka ping

notxvilka (approved):
LGTM

moste00:
@Rot127 I think this is ready for merging, could you please approve ?

.github/riscv-qemu-setup-all.sh
```diff
@@ -0,0 +1,89 @@
...
+LIB_M=libm.so.6
+DYN_LD=ld-linux-riscv64-lp64d.so.1
+
+# Fetch the kernel package
+wget "$DEBIAN_SERVER/$LINUX_IMG"
+# Extract
+dpkg-deb -x "$LINUX_IMG" linux-image
+
+# Fetch busybox
+git clone https://github.com/rcore-os/busybox-prebuilts.git busybox
```
wargio, line 19: we could toss a release in our [fallback repository](https://github.com/rizinorg/fallback-repo)
also `--depth 1`

wargio (approved):

moste00:
@Rot127 ping

wargio (approved):

PR #6030, opened 2026-03-13
RISC-V: Implement various cpus profiles & features

wargio:

**Detailed description**

...

**Test plan**

...

**Closing issues**

...


test/db/formats/zimg
```diff
@@ -238,35 +238,36 @@ zImage:
...
+        |   0x0000002e      unimp
         |   ;-- str.RISCV:
         |   0x00000030     .string "RISCV" ; len=6
-        |   0x00000036      illegal
-        |   0x00000038      lw    t1, 52(sp)
-        |   0x0000003a      fmadd.s fa0, ft0, ft0, ft0, rne
-        |   0x0000003e      illegal
+        |   0x00000036      unimp
+        |   0x00000038      lw    t1, 0x34(sp)
+        |   0x0000003a      invalid
```
moste00, line 268: Flagging this as a regression because the original output at the same address is `fmadd.s` which is likely `f` or `d` extension and they should  be enabled so this shouldn't happen, if the bug happens despite `f` or `d` being enabled then it's a bug that requires deeper investigation beyond this PR.

librz/arch/p/asm/asm_riscv_cs.c
```diff
@@ -16,6 +16,7 @@ static int riscv_disassemble(RzAsm *a, RzAsmOp *op, const ut8 *buf, int len) {
 	cs_insn *insn;
 	cs_mode mode = (a->bits == 64) ? CS_MODE_RISCV64 : CS_MODE_RISCV32;
 	mode |= mode_from_arch_string(a->cpu);
+	mode |= resolve_features_from_list(a->features);
```
moste00, line 19: This would end up unconditionally enabling every single mode that Capstone exposes right ? If so then it's contradictory, some modes conflict with each other and result in test failures for their fellow modes.

Also, even if contradiction isn't a problem, I can imagine users being surprised that invalid instructions that don't make sense in their context suddenly make sense because we're forcibly enabling the mode that makes them make sense (e.g. a binary for a CPU without `c` and having invalid instructions, enabling `c` changes the meaning of those invalid bytes to meaningful instructions).

But the contradiction is the biggest issue.

moste00: On second thought, maybe the `features` in `a` is a user-chosen subset from the complete list of `features` in the plugin definition below, in which case ignore both of my points. This is also how `cpus` work, probably.

wargio: yes, cpus is the base and features is the addition. i would expect `rv64` as cpu `a,f,d,c,v` for example as features

test/db/formats/zimg
Rot127: Then we need an issue though

Rot127 (approved):

PR #6035, opened 2026-03-14
Improve performance of utf string processing and `io_memory_read`

### LINKED ISSUE #4386: Make `ht` and buffer reading faster

antonangeloff:

**Detailed description**

This PR introduces a few performance improvement (based on profiling, these parts of the code were taking noticable time):

 - Add a bitmap to speed up [undefined_ranges](https://github.com/antonangeloff/rizin/blob/85620596c728e79902693000c09ca4a5033a5b85/librz/util/unicode.c#L89) lookup when checking [whether an UTF code page is defined](https://github.com/antonangeloff/rizin/blob/85620596c728e79902693000c09ca4a5033a5b85/librz/util/unicode.c#L983)
 - Try to reduce heap allocations (when possible) when processing UTF strings
 - Avoid setting output buffer to `0xff` in the success path of [io_memory_read](https://github.com/rizinorg/rizin/pull/6035/changes#diff-965d7435f1166906d2dd57500cb154497d25683447982b3e695ed26051689815R97).

I've run parts of the regression test suite to measure execution time changes:
| Tests | Before (s) | After (s) | Change ratio (`after / before`) |
|---|---:|---:|---:|
| `rz-test db/format` | 38.682 | 27.497 | 141% |
| `rz-test db/analysis` | 67.495 | 46.924 | 143% |
| `o bins/dmp/bmp.dmp` | 2.942 | 2.482 | 118% |

Some other parts of the test suite, that I've check, don't seem to be affected much.

**Test plan**

Existing tests should pass.

**Closing issues**

Closes https://github.com/rizinorg/rizin/issues/4386

### DIFF
--- a/librz/io/io_memory.c
+++ b/librz/io/io_memory.c
@@ -95,12 +95,13 @@ bool io_memory_resize(RzIO *io, RzIODesc *fd, ut64 count) {
 }
 
 int io_memory_read(RzIO *io, RzIODesc *fd, ut8 *buf, size_t count) {
-	memset(buf, 0xff, count);
 	if (!fd || !fd->data) {
+		memset(buf, 0xff, count);
 		return -1;
 	}
 	ut32 mallocsz = _io_malloc_sz(fd);
 	if (_io_malloc_off(fd) > mallocsz) {
+		memset(buf, 0xff, count);
 		return -1;
 	}
 	if (_io_malloc_off(fd) + count >= mallocsz) {
--- a/librz/io/io_memory.c
+++ b/librz/io/io_memory.c
@@ -105,7 +105,9 @@ int io_memory_read(RzIO *io, RzIODesc *fd, ut8 *buf, size_t count) {
 		return -1;
 	}
 	if (_io_malloc_off(fd) + count >= mallocsz) {
-		count = mallocsz - _io_malloc_off(fd);
+		size_t new_count = mallocsz - _io_malloc_off(fd);
+		memset(buf + new_count, 0xff, count - new_count);
+		count = new_count;
 	}
 	memcpy(buf, _io_malloc_buf(fd) + _io_malloc_off(fd), count);
 	_io_malloc_set_off(fd, _io_malloc_off(fd) + count);
--- a/librz/util/unicode.c
+++ b/librz/util/unicode.c
@@ -903,6 +903,50 @@ static const RzUnicodeCaseMap uppercase_mapping = {
 };
 
 // clang-format on
+static ut64 *undefined_ranges_bitmap = NULL;
+static ut32 undefined_ranges_bitmap_size = 0;
+static ut32 undefined_ranges_skip_entries = 0;
+
+#ifdef RZ_DEFINE_CONSTRUCTOR_NEEDS_PRAGMA
+#pragma RZ_DEFINE_CONSTRUCTOR_PRAGMA_ARGS(undefined_ranges_bitmap_bake)
+#endif
+RZ_DEFINE_CONSTRUCTOR(undefined_ranges_bitmap_bake)
+static void undefined_ranges_bitmap_bake(void) {
+	const ut32 bits_per_element = sizeof(undefined_ranges_bitmap[0]) * 8;
+	undefined_ranges_bitmap_size = 0x110000;
+	undefined_ranges_bitmap = RZ_NEWS0(ut64, (undefined_ranges_bitmap_size + (bits_per_element - 1)) / bits_per_element);
+
+	for (ut32 table_index = 0; table_index < RZ_ARRAY_SIZE(undefined_ranges); table_index++) {
+		ut32 from = undefined_ranges[table_index].from;
+		ut32 to = RZ_MIN(undefined_ranges[table_index].to + 1, undefined_ranges_bitmap_size);
+
+		for (RzCodePoint code = from; code < to; code++) { // todo: confirm range inclusion/exclusion
+			undefined_ranges_bitmap[code / bits_per_element] |= 1ull << (code % bits_per_element);
+		}
+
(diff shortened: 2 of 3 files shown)


antonangeloff:
Is it possible the coverage for [this line](https://app.codecov.io/gh/rizinorg/rizin/pull/6035?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=rizinorg#39561ed20757f7273c8e146988d794ce-R925) is reported incorrectly? Except for TCC (where this particular optimization is disabled), that line should be executed on all other platforms, otherwise it would lead to incorrect results and would have broken regression tests.

Rot127:
Could be. Depends what the system is where the coverage was generated on.

librz/io/io_memory.c
```diff
@@ -95,12 +95,13 @@ bool io_memory_resize(RzIO *io, RzIODesc *fd, ut64 count) {
...
 
 int io_memory_read(RzIO *io, RzIODesc *fd, ut8 *buf, size_t count) {
-	memset(buf, 0xff, count);
 	if (!fd || !fd->data) {
+		memset(buf, 0xff, count);
 		return -1;
 	}
 	ut32 mallocsz = _io_malloc_sz(fd);
 	if (_io_malloc_off(fd) > mallocsz) {
+		memset(buf, 0xff, count);
```
wargio, line 104: probably you should also memset the difference in the `if` later?

librz/util/unicode.c
```diff
@@ -903,6 +903,52 @@ static const RzUnicodeCaseMap uppercase_mapping = {
...
 // clang-format on
+static ut64 *undefined_ranges_bitmap = NULL;
+static ut32 undefined_ranges_bitmap_size = 0;
+static ut32 undefined_ranges_skip_entries = 0;
+
+#ifdef RZ_HAS_CONSTRUCTORS
+#ifdef RZ_DEFINE_CONSTRUCTOR_NEEDS_PRAGMA
+#pragma RZ_DEFINE_CONSTRUCTOR_PRAGMA_ARGS(undefined_ranges_bitmap_bake)
+#endif // RZ_DEFINE_CONSTRUCTOR_NEEDS_PRAGMA
+RZ_DEFINE_CONSTRUCTOR(undefined_ranges_bitmap_bake)
```
wargio, line 914: i really don't think this is a good idea. @notxvilka

librz/io/io_memory.c
antonangeloff: Ah yes, missed that

librz/util/unicode.c
antonangeloff: I've saw this used [here](https://github.com/antonangeloff/rizin/blob/dev/librz/util/sys.c#L1463). Did you had problems with constructors before? Alternative could be delayed initialization on first lookup, with atomic pointer.

wargio: yes, i'm aware, but i prefer if we have a singleton that is never freed.

notxvilka: I don't have a strong opinion on that. Both approaches have their merit. @Rot127 WDYT?

antonangeloff: To summarize the approaches I've thought of:

1. Hardcode static bitmap in the `unicode.c` as array - this will require adding a script for regenerating the array in case changes are needed
2. Generate the bitmap dynamically on startup (we are here) - it's based on `undefined_ranges` so we have only 1 source of truth, it's initialized once at startup so no locking/synchronization concerns
3. Generate the bitmap on first read - similar to 2. but we'll need an atomic pointer so the delayed initialization works correctly in multi-threaded setup

Note: actually another upside of point 3. is that it will make the (bitmap lookup) optimization possible on compilers which don't support constructors consistently (e.g. TCC)

Rot127: I prefer 1).
Mostly because the `undefined_ranges` table is generated manually as well. So if it is updated we must run scripts anyways. Also option 2)/3) are complicated. Personally I prefer the stupid simple solutions if they give the same result (or do I miss something).

Running one more script whenever we update unicode is not that much of an effort.
We would need a comment at `undefined_ranges` as a reminder of course.

The more fancy scripting solution would be to write a single script to update the unicode tables (and the lookup table in this case). It could generate `.inc` files we include there.

Then the whole process is running only one script once for an unicode update.

antonangeloff: Okay, another consideration I forgot to mention: point 1) will increase binary size with around `138kb` (size of the bitmap), while point 2)/3) will use only heap for storing the bitmap. 

If it's good to proceed with 1), let me know.

Also @Rot127 any suggestion where to place the new script?

> Running one more script whenever we update unicode is not that much of an effort.

Yes, not much effort. I was thinking that having it runtime-generated will (by design) not allow any mistakes during manual re-generations, and avoid keeping duplicate data (the table itself and the bitmap).

Rot127: 138kb is actually quite a lot.

@notxvilka @wargio @thestr4ng3r wdyt?

PR #6052, opened 2026-03-19
analysis: add enum immediate hints with ahie

### LINKED ISSUE #697: Allow to set the enum value for the operand.

IndAlok:

**Detailed description**

This PR adds enum hints for immediate operands through the new `ahie` / `ahie-` commands.

The main goal is to let an immediate operand be rendered as an enum member instead of a raw number when the value matches a loaded enum. For example, after defining an enum and applying `ahie BLA`, an instruction like `cmp r12b, 7` is displayed as `cmp r12b, BLA.BAR`.

`ahi` is handled as an immediate-hint command group now, so `ahie` is exposed alongside the existing immediate-base hint commands.

On the analysis side, enum hints are now:
- stored in address hints,
- merged and freed correctly,
- serialized/deserialized in analysis projects,
- shown in hint listings,
- used by the disassembly filter to replace matching immediates with `Enum.Member`.

`ahie` also accepts an optional operand index, matching the existing `ahi` behavior for selecting the N-th immediate operand.

**Test plan**

Tested in the meson dev env (win).

Passing:
```powershell
meson compile -C builddir -j 1
meson devenv -C builddir powershell -Command "& 'C:\Users\Admin\Desktop\rizin\builddir\test\unit\test_analysis_hints.exe'"
meson devenv -C builddir powershell -Command "& 'C:\Users\Admin\Desktop\rizin\builddir\test\unit\test_serialize_analysis.exe'"
meson devenv -C builddir powershell -Command "& 'C:\Users\Admin\Desktop\rizin\builddir\binrz\rz-test\rz-test.exe' 'C:\Users\Admin\Desktop\rizin\test\db\cmd\cmd_ahi'"
meson devenv -C builddir powershell -Command "& 'C:\Users\Admin\Desktop\rizin\builddir\binrz\rizin\rizin.exe' -q -n -c 'ah?; q'"
```

Verified:
- `test_analysis_hints.exe` passed.
- `test_serialize_analysis.exe` passed.
- `cmd_ahi` regression tests passed, including `ahie` cases.
- `ah?` now shows `ahi[-e?]`

Coverage added:
- command regression tests for enum rendering, removal, and non-matching values,
- unit coverage for enum hint storage,
- unit coverage for enum hint serialization/deserialization.

**Closing issues**

closes #697

### DIFF
--- a/librz/arch/filter.c
+++ b/librz/arch/filter.c
@@ -9,6 +9,7 @@
 #include <rz_util/rz_regex.h>
 #include <rz_types.h>
 #include <rz_arch.h>
+#include <rz_type.h>
 
 #define isx86separator(x) ( \
 	(x) == ' ' || (x) == '\t' || (x) == '\n' || (x) == '\r' || (x) == ' ' || \
@@ -417,6 +418,17 @@ static bool filter(RzParse *p, ut64 addr, RzFlag *f, RzAnalysisHint *hint, char 
 				*ptr = 0;
 				snprintf(str, len, "%s%s%s", data, hint->offset, (ptr != ptr2) ? ptr2 : "");
 				return true;
+			}
+			if (hint->enum_name && *hint->enum_name && p && p->analb.analysis) {
+				const char *member = rz_type_db_enum_member_by_val(
+					p->analb.analysis->typedb, hint->enum_name, off);
+				if (member) {
+					char ename[256];
+					snprintf(ename, sizeof(ename), "%s.%s", hint->enum_name, member);
+					*ptr = 0;
+					snprintf(str, len, "%s%s%s", data, ename, (ptr != ptr2) ? ptr2 : "");
+					return true;
+				}
 			}
 			strncpy(num, ptr, sizeof(num) - 2);
 			pnum = num + parse_number(num);
--- a/librz/arch/filter.c
+++ b/librz/arch/filter.c
@@ -43,6 +43,41 @@ static void insert(char *dst, const char *src) {
 	free(endNum);
 }
 
+static bool replace_number_token(char *out, int out_len, char *data, char *num_start, char *num_end, const char *value) {
+	rz_return_val_if_fail(out && data && num_start && num_end && value, false);
+	*num_start = 0;
+	return snprintf(out, out_len, "%s%s%s", data, value, (num_start != num_end) ? num_end : "") >= 0;
+}
+
+static bool replace_enum_hint(RzParse *p, RzAnalysisHint *hint, ut64 off, char *data, char *out, int out_len, char *num_start, char *num_end) {
+	rz_return_val_if_fail(p && hint && data && out && num_start && num_end, false);
+
+	if (RZ_STR_ISEMPTY(hint->enum_name) || !p->analb.analysis || !p->analb.analysis->typedb) {
+		return false;
+	}
+
+	const char *member = rz_type_db_enum_member_by_val(
+		p->analb.analysis->typedb, hint->enum_name, off);
+	if (!member) {
+		return false;
+	}
+
+	char ename[512];
+	size_t ename_len = strlen(hint->enum_name) + strlen(member) + 2;
+	if (ename_len <= sizeof(ename)) {
+		return replace_number_token(out, out_len, data, num_start, num_end,
+			rz_strf(ename, "%s.%s", hint->enum_name, member));
+	}
+
+	char *ename_dyn = rz_str_newf("%s.%s", hint->enum_name, member);
+	if (!ename_dyn) {
+		return false;
+	}
+	bool ok = replace_number_token(out, out_len, data, num_start, num_end, ename_dyn);
+	free(ename_dyn);
+	return ok;
+}
+
 static int parse_number(const char *str) {
 	const char *p = str;
 	// Parse as hexadecmial (0x notation)
@@ -406,6 +441,8 @@ static bool filter(RzParse *p, ut64 addr, RzFlag *f, RzAnalysisHint *hint, char 
 			}
 		}
 		if (hint) {
+			char *num_start = ptr;
+			char *num_end = ptr2;
 			const int nw = hint->nword;
 			if (count != nw) {
 				ptr = ptr2;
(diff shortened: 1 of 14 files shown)


test/db/cmd/cmd_ahi
```diff
@@ -109,6 +109,16 @@ EXPECT=<<EOF
 EOF
 RUN
 
+NAME=ah? shows ahie
```
notxvilka: No need for this particular test

test/db/cmd/cmd_ahi
```diff
@@ -380,3 +390,70 @@ EXPECT=<<EOF
             0x00000000      cmp   rax, 18446744073709551615
 EOF
 RUN
+
+NAME=ahie enum operand
```
notxvilka: Merge them into one, but add more complex enum handling and more instructions. Moreover, not only use x86 but also some other architectures for better coverage: ARM, ARM64, PPC, MIPS, 8051, Tricore, Hexagon.

notxvilka (changes requested):

test/db/cmd/cmd_ahi
IndAlok: Yep was just testing locally, forgot to remove

librz/arch/filter.c
```diff
@@ -418,6 +419,17 @@ static bool filter(RzParse *p, ut64 addr, RzFlag *f, RzAnalysisHint *hint, char
 				snprintf(str, len, "%s%s%s", data, hint->offset, (ptr != ptr2) ? ptr2 : "");
 				return true;
 			}
+			if (hint->enum_name && *hint->enum_name && p && p->analb.analysis) {
+				const char *member = rz_type_db_enum_member_by_val(
+					p->analb.analysis->typedb, hint->enum_name, off);
+				if (member) {
+					char ename[256];
+					snprintf(ename, sizeof(ename), "%s.%s", hint->enum_name, member);
```
notxvilka: Use `rz_strf()`

librz/type/parser/types_parser.c
```diff
@@ -114,6 +114,12 @@ int parse_primitive_type(CParserState *state, TSNode node, const char *text, Par
 		free(real_type);
 		return 0;
 	}
+	if ((*tpair = c_parser_get_typedef(state, real_type))) {
```
notxvilka, line 117: Could you please explain how did you come to the conclusion this is needed here?

librz/type/parser/types_parser.c
```diff
@@ -150,6 +156,12 @@ int parse_sized_primitive_type(CParserState *state, TSNode node, const char *tex
...
```
notxvilka, line 159: Same

notxvilka (changes requested):

librz/type/parser/types_parser.c
IndAlok: surely, the failing CI I mentioned in mattermost pm to you, before th last commit (https://github.com/rizinorg/rizin/actions/runs/23305164073/job/67777119533?pr=6052), there I saw:
Primitive type "size_t" already exists
then, Error forming RzType and RzBaseType pair out of primitive type

type desc like size_t, already existed in the typedb, but as typedef backed types, not atomics...
and c_parser_get_primitive_type() returned null cuz they weren't atomic (i saw it only returned a match when base type was RZ_BASE_TYPE_KIND_ATOMIC), and so the parser then wrongly got to c_parser_new_primitive_type(), which ofc refused cuz a base type with the same name was there...

these parser paths didn't hav the typedef resolution...

hence i did add:
```
if ((*tpair = c_parser_get_typedef(state, real_type))) {
	(*tpair)->type->identifier.is_const = is_const;
	...
}
```

if we see in line 179, 
int parse_sole_type_name(CParserState *state, TSNode node, const char *text, ParserTypePair **tpair, bool is_const) {

it is already having this similar typedef fallback.

about c_parser_get_typedef, if we see in types_storage.c, (line 739), to reuse an existing typedef-backed base type, it was used:
RZ_OWN ParserTypePair *c_parser_get_typedef(CParserState *state, RZ_NONNULL const char *name) {

also, c_parser_get_typedef built identifiers with is_const to false (default), but these parser paths were carrying qualifier state separately. so just did that so const size_t doesn't loose constness.

librz/arch/filter.c
```diff
@@ -418,6 +419,17 @@ static bool filter(RzParse *p, ut64 addr, RzFlag *f, RzAnalysisHint *hint, char
 				snprintf(str, len, "%s%s%s", data, hint->offset, (ptr != ptr2) ? ptr2 : "");
 				return true;
 			}
+			if (hint->enum_name && *hint->enum_name && p && p->analb.analysis) {
+				const char *member = rz_type_db_enum_member_by_val(
+					p->analb.analysis->typedb, hint->enum_name, off);
+				if (member) {
+					char ename[256];
```
notxvilka: Probably easy to overflow

notxvilka (changes requested):

librz/arch/filter.c
IndAlok: then can rz_str_newf be used for dynamic allocation? (prolly you mean user inp. could be too long to make (enum_name + "." + member) much longer and rz_strf would be trunc. the 256th and so on char(s)), and then freeing the ename?

IndAlok: sth like:
```
if (member) {
    char *ename = rz_str_newf("%s.%s", hint->enum_name, member);
    if (ename) {
        *ptr = 0;
        snprintf(str, len, "%s%s%s", data, ename,
                 (ptr != ptr2) ? ptr2 : "");
        free(ename);
        return true;
    }
}
```

notxvilka: It depends on how "hot" is this path. If it is - better to think about using stack but with reasonable (higher) limits, if not - dynamic allocation should do the job.

wargio: Maybe rename also those `ptr` & `ptr2` to something that makes more sense. also move out this portion into a function, because as is, it's unreadable

IndAlok: @notxvilka @wargio Please have a look, I've kept the dynamic alloc with rz_str_newf as a fallback and char ename[512]; for default...
and used num_start and num_end for ptr and ptr2

notxvilka (approved):

librz/arch/filter.c
```diff
@@ -43,6 +43,41 @@ static void insert(char *dst, const char *src) {
 	free(endNum);
 }
 
+static bool replace_number_token(char *out, int out_len, char *data, char *num_start, char *num_end, const char *value) {
```
wargio: use size_t for sizes

librz/arch/filter.c
```diff
@@ -43,6 +43,41 @@ static void insert(char *dst, const char *src) {
...
+	rz_return_val_if_fail(out && data && num_start && num_end && value, false);
+	*num_start = 0;
+	return snprintf(out, out_len, "%s%s%s", data, value, (num_start != num_end) ? num_end : "") >= 0;
+}
+
+static bool replace_enum_hint(RzParse *p, RzAnalysisHint *hint, ut64 off, char *data, char *out, int out_len, char *num_start, char *num_end) {
+	rz_return_val_if_fail(p && hint && data && out && num_start && num_end, false);
```
wargio: why you need this?

librz/arch/filter.c
```diff
@@ -43,6 +43,41 @@ static void insert(char *dst, const char *src) {
 	free(endNum);
 }
 
+static bool replace_number_token(char *out, int out_len, char *data, char *num_start, char *num_end, const char *value) {
+	rz_return_val_if_fail(out && data && num_start && num_end && value, false);
```
wargio: same, why you need this?

librz/arch/filter.c
```diff
@@ -43,6 +43,41 @@ static void insert(char *dst, const char *src) {
...
+	*num_start = 0;
+	return snprintf(out, out_len, "%s%s%s", data, value, (num_start != num_end) ? num_end : "") >= 0;
```
wargio: just convert this function to void.

wargio (changes requested):
almost perfect

librz/arch/filter.c
IndAlok: mb, were only called by single filter() path with alr controlled args, removed them.

wargio (approved):
lgtm

PR #4732, opened 2024-11-23
Port `ae` commands to RzShell

Rot127:
**Your checklist for this pull request**
- [x] I've read the [guidelines for contributing](https://github.com/rizinorg/rizin/blob/master/DEVELOPERS.md) to this repository
- [x] I made sure to follow the project's [coding style](https://github.com/rizinorg/rizin/blob/master/DEVELOPERS.md#code-style)
- [x] I've documented or updated the documentation of every function and struct this PR changes. If not so I've explained why.
- [ ] I've added tests that prove my fix is effective or that my feature works (if possible)
- [ ] I've updated the [rizin book](https://github.com/rizinorg/book) with the relevant information (if needed)

**Detailed description**

Port the `ae` commands to RzShell.

**Remove**

- [x] `ae*` - It was marked as not working and had no direct advantage. With the transition to RzIL it becomes obsolete.

**To RzShell, but remove with ESIL**
- [x] `aepc`
- [x] `ae `
- [x] `ae??`
- [x] `aex`

**To RzShell, but port to RzIL**

- [x] `aeb`
- [x] `aef`
- [x] `aefa`
- [x] `aeA`
- [x] `aea`
- [x] `aeab`
- [x] `aeafj`
- [x] `aeafj`
- [x] `aeab`
- [x] `aeabj`

**To RzShell, but reimplement**
- [x] `aek`
- [x] `aek-`
- [x] `aeli`
- [x] `aeli`
- [x] `aelir`

**Last**

- [x] Remove `a*` (doesn't work)
- [x] Remove legacy `rz_cmd_analysis` handler + fixup in yaml files.

**Test plan**

All green

**Closing issues**

closes https://github.com/rizinorg/rizin/issues/1533

Issue #834, opened 2021-03-16
Unbreak broken tests


XVilka:
Currently 938 tests marked as broken. Some are outdated and can be fixed with the new output, some are indicative of the serious bugs. We should go through all of them and fix one by one, by either fixing the output, fixing the issue, or removing the test as invalid.
```
[i] ℤ rg "BROKEN=1" test/db                                                                                                                                                                                                       16:46:55 
test/db/tools/rz_bin
256:BROKEN=1
279:BROKEN=1
302:BROKEN=1
709:BROKEN=1
891:BROKEN=1
1225:BROKEN=1
1234:BROKEN=1

test/db/tools/rz_gg
92:BROKEN=1
109:BROKEN=1

test/db/tools/rz_asm
52:BROKEN=1 # Only on Windows
377:BROKEN=1 # Only on Windows
414:BROKEN=1
531:BROKEN=1
548:BROKEN=1
565:BROKEN=1
574:BROKEN=1
583:BROKEN=1
592:BROKEN=1

test/db/tools/rz
27:BROKEN=1

test/db/tools/rz_ax
409:BROKEN=1

test/db/tools/rz_find
95:BROKEN=1
104:BROKEN=1
114:BROKEN=1
129:BROKEN=1

test/db/analysis/msp430
57:BROKEN=1

test/db/analysis/mips
521:BROKEN=1

test/db/analysis/avr
41:BROKEN=1

test/db/analysis/arm
512:BROKEN=1
742:BROKEN=1
809:BROKEN=1
852:BROKEN=1

test/db/analysis/arc
118:BROKEN=1
167:BROKEN=1
428:BROKEN=1
586:BROKEN=1
800:BROKEN=1
990:BROKEN=1

test/db/formats/xbe
37:BROKEN=1
51:BROKEN=1

test/db/formats/pe/create
3:BROKEN=1

test/db/tools/rz_diff
60:BROKEN=1

test/db/formats/mz/cdogs
16:BROKEN=1
25:BROKEN=1
34:BROKEN=1

test/db/formats/mz/chkdsk
26:BROKEN=1

test/db/formats/mz/blocks
26:BROKEN=1

test/db/formats/elf/thumb
53:BROKEN=1

test/db/formats/pe/96emptysections
18:BROKEN=1

test/db/formats/elf/core
11:BROKEN=1

test/db/formats/elf/strings
14:BROKEN=1
29:BROKEN=1
44:BROKEN=1
59:BROKEN=1
[...]

PR #6077, opened 2026-03-22
Fix expected output for aek- regression test

MrQuantum1915:
The previous expected output matched an old crash regression run, but the C handler `rz_analyze_esil_sdb_reset_handler` now already catches NULL `esil` states (preventing crash) , without outputting anything. Hence the expected output for the test is also obsolete.

 

**Detailed description**

This PR fixes the aek- - crash test in [cmd_aek](https://github.com/rizinorg/rizin/blob/dev/test/db/cmd/cmd_aek) by fixing both the Expected Output and removes its BROKEN=1 tag.

As per the one of the instructions in #834 to 'fix the output' by @XVilka, this PR corrects the outdated EXPECT output for the aek- regression test

While auditing broken tests, I made sure to investigate the root cause rather than blindly removing the BROKEN tag. In this case, the bug in the source code has already been fixed, but the test file's expected output was never updated to reflect the fix.

**Full Context and analyses for root cause**

**History:** This test was originally imported in 2019 (in radare2 repo) from radare2-regressions (Commit ae61aa8e379) by @thestr4ng3r .

The `aek-` cmd is to reset esil info sdb instance [cmd_analysis.yaml](https://github.com/rizinorg/rizin/blob/16357f4be027e97bd34a1ce88980ea80a422148b/librz/core/cmd_descs/cmd_analysis.yaml#L2460) :

```yaml
- name: aek-
            summary: Resets the ESIL info sdb instance.
            cname: analyze_esil_sdb_reset
            args: []
```

And hence the defining code is in `cmd_analysis.c`'s `rz_analyze_esil_sdb_reset_handler` function.

The name of the test `aek- - crash` implies it was written to catch a segmentation fault when executing aek- (reset ESIL stats) in an uninitialized environment (without `aei`) which is clear from the query handler function just above the reset handler function, because if `aei` is not initialised it returns the error and says to Run `aei`. 

[cmd_analysis.c | Line:6213](https://github.com/rizinorg/rizin/blob/16357f4be027e97bd34a1ce88980ea80a422148b/librz/core/cmd/cmd_analysis.c#L6213-L6233)

```c
RZ_IPI RzCmdStatus rz_analyze_esil_sdb_query_handler(RzCore *core, int argc, const char **argv) {
	RzAnalysisEsil *esil = core->analysis->esil;
	if (!esil || !esil->stats) {
		RZ_LOG_ERROR("core: esil.stats is empty. Run 'aei'\n");
		return RZ_CMD_STATUS_ERROR;
	}
	char *out = sdb_querys(esil->stats, NULL, 0, argv[1]);
	if (out) {
		rz_cons_println(out);
		free(out);
	}
	return RZ_CMD_STATUS_OK;
}

RZ_IPI RzCmdStatus rz_analyze_esil_sdb_reset_handler(RzCore *core, int argc, const char **argv) {
	RzAnalysisEsil *esil = core->analysis->esil;
	if (esil) {
		sdb_reset(esil->stats);
	}
	return RZ_CMD_STATUS_OK;
}
```

Without aei, `core->analysis->esil` is NULL. The older code may have executed `sdb_reset(esil->stats)` without checking the esil instance, resulting in a null pointer dereference and causing a crash. 

Also the 
```
EXPECT=<<EOF
1
0
EOF
RUN
``` 
block in the test appears to have been copy-pasted from the test directly above it, because crashes don't produce clean `1\n0` stdout anyway. So this expected output check was surely wrong from the Start.

The Source Code is already patched: If we look at the current reset handler in `cmd_analysis.c` file, it already first checks for an esil instance before running sdb_reset().

**Note**:  while reviewing git history using `git blame`, I recognized that the `aek-` command was recently ported to RzShell in #4732 by @Rot127 (as part of "To RzShell, but reimplement"). The C handler safely catches uninitialized ESIL states without outputting anything, but the old regression test still maintained its garbage `EXPECT` output from a historical segfault. This PR aligns the test with the current modern handler.

**Why it was marked BROKEN**: Because of the `if (esil)` check, the code correctly prevents the crash and exits gracefully with NO stdout. However, test runner still sees the old, copy-pasted EXPECT=1\n0 block and compares it to the empty output, marking it as a failure.

**Fix:**
I updated `cmd_aek` test by removing the incorrect `1\n0` expected output to match the correct, intended behavior of the C code handler. I also removed the BROKEN=1 flag.

By applying this fix, the test is no longer bypassed and now acts as an active, passing regression test against future aek- segmentation faults.

**Test plan**

As the fix is only in cmd_aek regression test, we need to run that test to check output:

image

As you can see that before the fix, the output of test shows 1 broken test from 2. And after the fix, there are 0 broken test and 2 passed test.

...

**Closing issues**

...

### DIFF
--- a/test/db/cmd/cmd_aek
+++ b/test/db/cmd/cmd_aek
@@ -22,10 +22,6 @@ RUN
 
 NAME=aek- - crash
 FILE==
-BROKEN=1
 CMDS=aek-
-EXPECT=<<EOF
-1
-0
-EOF
+EXPECT=
 RUN
--- a/test/db/cmd/cmd_aek
+++ b/test/db/cmd/cmd_aek
@@ -6,6 +6,7 @@ e asm.bits=64
 e analysis.arch=x86
 e esil.stats=true
 wx 41574156415541544989f4555389fd4881eca8030000488b3e64488b042528
+aek-
 aei
 aeim
 aeip
@@ -18,10 +19,4 @@ EXPECT=<<EOF
 1
 0
 EOF
-RUN
-
-NAME=aek- - crash
-FILE==
-CMDS=aek-
-EXPECT=
 RUN


test/db/cmd/cmd_aek
```diff
@@ -22,10 +22,6 @@ RUN
 
 NAME=aek- - crash
```
wargio: merge this test with the one before. 

place an `aek-` on the line before `aei`

wargio (changes requested):

test/db/cmd/cmd_aek
MrQuantum1915: implemented that part.
testing : 
image


also running aek- before aei currently, silently return ok without error.




does the project prefer aek- to remain silent when uninitialized, or should it report an error like the query handler? Or it will have breaking changes if done so?

```c
RZ_IPI RzCmdStatus rz_analyze_esil_sdb_query_handler(RzCore *core, int argc, const char **argv) {
	RzAnalysisEsil *esil = core->analysis->esil;
	if (!esil || !esil->stats) {
		RZ_LOG_ERROR("core: esil.stats is empty. Run 'aei'\n");
		return RZ_CMD_STATUS_ERROR;
	}
	char *out = sdb_querys(esil->stats, NULL, 0, argv[1]);
	if (out) {
		rz_cons_println(out);
		free(out);
	}
	return RZ_CMD_STATUS_OK;
}

RZ_IPI RzCmdStatus rz_analyze_esil_sdb_reset_handler(RzCore *core, int argc, const char **argv) {
	RzAnalysisEsil *esil = core->analysis->esil;
	if (esil) {
		sdb_reset(esil->stats);
	}
	return RZ_CMD_STATUS_OK;
}
```

for that then we have to expect some output when aek- is run before aei in testing

Rot127 (approved):

wargio (approved):

wargio:
your changes are correct, no need to touch the code.

PR #6081, opened 2026-03-23
Implement ring buffer

Rot127:

**Detailed description**

Adds a ring buffer implementation.
Its main purpose is to serve as more performant data structure compared to the queue, while essentially giving the same API.

It can achieve a higher performance than a queue, because it allows to pass objects without allocating them on the heap first.

A sender can simple initialize the data on its stack and copy it into the ring buffer.
The `memcpy` call is much faster than allocating the object and then sending the pointer over the queue.

**Test plan**

Added

**Closing issues**

...


Rot127:
@notxvilka @wargio @antonangeloff 

I didn't test anything yet. So there are maybe heavy logic flaws in there.
But one design decision is: Should the Ring Buffer support two modes? An overflow and "block" mode.

```c
typedef enum {
	RZ_THREAD_RING_BUF_BLOCK = 0, ///< The ring buffer blocks if it is full.
	RZ_THREAD_RING_BUF_OVERFLOW, ///< The ring buffer overwrites the oldest element when full.
} RzThreadRingBufMode;
```

Supporting both in the same structure, adds at least two branches in the `put()` function. And since this data structure is inherently a hot path we should think about, if we really need the overflow version.
I can't think of any use case tbh. It is usually used in audio/video processing where data loss doesn't matter too much.

Options are:
1. Remove the overflow version. And if we need it, add it as separated module `RzThreadRingBufOv` or something like that.
2. Keep it and check back if these branches actually eat too much runtime.

antonangeloff:
> I didn't test anything yet. So there are maybe heavy logic flaws in there. But one design decision is: Should the Ring Buffer support two modes? An overflow and "block" mode.

Apart from the put function, I assume the block mode wouldn't affect the rest of the ring buffer implementation? Perhaps having 2x write functions (e.g. rz_th_ring_buf_put(), rz_th_ring_buf_put_overwrite()) would avoid branching? Also the overwrite version could be added at any point (when there's use case), if not required right away.

wargio:
why you need this feature?

Rot127:
@wargio 
> why you need this feature?

It is for the new analysis. I use queues there currently, but they simply don't fit the problem.
The results of the interpretation could be way simpler and more efficiently passed on via a shared memory like this.

Rot127:
> Perhaps having 2x write functions (e.g. rz_th_ring_buf_put(), rz_th_ring_buf_put_overwrite()) would avoid branching? Also the overwrite version could be added at any point (when there's use case), if not required right away.

That is a good idea!
So let's scrap the mode and just implement two functions.

wargio:
> It is for the new analysis.

i don't see how this can help you for the analysis, but ok.

Rot127:
> i don't see how this can help you for the analysis, but ok.


We don't need to merge it now.
I will use it there and show you.

Rot127:
But the basic problem is that a queue only takes pointers. And with the pointers I need to manage their lifetime. Which gets annoying.
With that buffer each thread can just write and forget.

librz/util/thread_ring_buf.c
```diff
@@ -0,0 +1,361 @@
+// SPDX-FileCopyrightText: 2026 Rot127 <rot127@posteo.com>
+// SPDX-License-Identifier: LGPL-3.0-only
+
```
notxvilka, line 3: Add a Doxygen file description pls

librz/util/thread_ring_buf.c
```diff
@@ -0,0 +1,361 @@
...
+	if (!rbuf->lock || !rbuf->counter_lock || !rbuf->writer_wait_cond || !rbuf->reader_wait_cond) {
+		goto err_free;
+	}
+
+	return rbuf;
+
+err_free:
+	rz_warn_if_reached();
+	free(rbuf->buf);
+	rz_th_lock_free(rbuf->lock);
```
notxvilka: Should we follow the reverse order of how locks and conditions were created?

notxvilka: Just a style thing though.

notxvilka (approved):
Looks good!

PR #5952, opened 2026-02-21
RISCV: Make analysis function set mnemonic field to enable ROP search

moste00:
**Your checklist for this pull request**
- [ ] I've read the [guidelines for contributing](https://github.com/rizinorg/rizin/blob/dev/CONTRIBUTING.md) to this repository.
- [ ] I made sure to follow the project's [coding style](https://github.com/rizinorg/rizin/blob/dev/DEVELOPERS.md#code-style).
- [ ] I've documented every `RZ_API` function and struct this PR changes.
- [ ] I've added tests that prove my changes are effective (required for changes to `RZ_API`).
- [ ] I've updated the [Rizin book](https://github.com/rizinorg/book) with the relevant information (if needed).
- [ ] I've used AI tools to generate fully or partially these code changes and I'm sure the changes are not copyrighted by somebody else.

**Detailed description**
Fixing https://github.com/rizinorg/rizin/issues/5545, ROP search used to not work correctly with RISC-V

...

**Test plan**

...

**Closing issues**

closes https://github.com/rizinorg/rizin/issues/5545

Issue #6061, opened 2026-03-21
Crash in ROP search


VersaNexusIX:
```
u0_a245@localhost ~> cd ~/dumb/NeoDAX/
u0_a245@localhost ~/d/NeoDAX (main)> nano target.c
u0_a245@localhost ~/d/NeoDAX (main)> gcc -O3 -march=armv8-a+crypto -flto -s -fomit-frame-pointer target.c -o target_final
u0_a245@localhost ~/d/NeoDAX (main)> strip --strip-all target_final
u0_a245@localhost ~/d/NeoDAX (main)> rizin -nc -A ./target_final
ERROR: Command '-A' does not exist.
 -- Use scr.accel to browse the file faster!
[0x00000000]> /R stp x29, x30
WARNING: is_invalid_instruction: assertion 'opst' failed (line 1122)
fish: Job 1, 'rizin -nc -A ./target_final' terminated by signal SIGSEGV (Address boundary error)
u0_a245@localhost ~/d/NeoDAX (main) [SIGSEGV]> cat target.c
#include <stdio.h>
#include <stdint.h>
#include <stdlib.h>

typedef uint64_t u64;
typedef uint8_t  u8;   // 🔥 FIX: define u8

static u64 key = 0x9e3779b97f4a7c15ULL;

static int opaque(u64 x) {
    return ((x * x + x) % 2) == (x % 2);
}

u64 vm(u64 input) {
    u64 r = input;
    u64 ip = 0;

    static u8 code[] = {
        3,1,4,1,5,9,2,6,5,3,5,8,9,7,9,3
    };

    static void *tbl[] = {
        &&ADD, &&XOR, &&ROL, &&MIX, &&END
    };

    while (1) {
        u8 op = code[ip++ & 0xF] % 5;

        if (!opaque(r)) {
            r ^= 0xDEADBEEF;
        }

        goto *tbl[op];

ADD:
        r += key;
        key ^= r;
        continue;

XOR:
        r ^= (key >> (r & 7));
        continue;

ROL:
        r = (r << 3) | (r >> 61);
        continue;

MIX:
        r ^= (r >> 17);
        r *= 0xed5ad4bbU;
        continue;

END:
        break;
    }

    return r;
}

int main(int argc, char **argv) {
    if (argc < 2) return 0;
    u64 x = strtoull(argv[1], NULL, 10);

    // 🔥 FIX format (ARM64 = unsigned long)
    printf("%lu\n", vm(x));

    return 0;
}
[...]

PR #6108, opened 2026-03-27
Fix Inconsistent ROP gadget info between detail printing and listing

### LINKED ISSUE #5491: Inconsistent ROP gadget info between detail printing and listing
### Work environment

| Architecture/bits of the file (mandatory)            | PPC, x86/32, x86/64 etc.
| `rizin -v` full output, **not truncated** (mandatory)  | latest

I found inconsistent ROP gadget info between detailed printing (/Rg) and listing (/R) during the implementation of https://github.com/rizinorg/rizin/pull/5490.

I use `test/bins/arm/crackme.arm32.bin` as the target. (I didn't set `e asm.arch=arm` and `e asm.bits=32`. But the similar stuff happens even with that setup.)

First, run `/R`, we can see in the tail:
```
  0x00002d28         e8070020e8  call 0xffffffffe8202d34
Gadget size: 5

  0x00002d2a               0020  add byte [rax], ah
  0x00002d2c         e8070020f0  call 0xfffffffff0202d38
Gadget size: 7

  0x00002d2c         e8070020f0  call 0xfffffffff0202d38
Gadget size: 5
```

Then I run `/Rg`.

The output gadget near `0x00002d2c`.
```
Gadget 0x20aa
Stack change: 0x9068
Changed registers: rax rcx rsp
Register dependencies:
Memory Read: rax Value: 0x68
Memory Read: rax Value: 0x68
Var Read: rcx

Gadget 0x20ae
Stack change: 0x9068
Changed registers: rcx rsp
Register dependencies:
Var Read: rcx

Gadget 0x20b0
Stack change: 0x9068
Changed registers: rsp
Register dependencies:

Gadget 0x20e0
Stack change: 0x368
Changed registers: rsp
Register dependencies:
```

The gadget ` 0x00002d2c` is not shown. 

Is this expected? I think that maybe some failures during performing gadget analysis cause this.

MrQuantum1915:

**Detailed description**

[ UPDATED ]

This PR, closes issue [#5491](https://github.com/rizinorg/rizin/issues/5491) about inconsistent output of /R and /Rg.

After discussion below and two merged PR's
1. fix mips analysis for 64bit and nmips: https://github.com/rizinorg/rizin/pull/6122
2. refactor of gadget search : https://github.com/rizinorg/rizin/pull/6130

This PR is on top of that changes to fix the original issue which was because previously /Rg search include COP and JOP related gadgets while /R didn't.

**COMMIT 1**: Remove JOP and COP from ROP search, Unify terminal instruction checks and Fix conditional terminator filter.
- in `gadget_rop.c` i removed call and jump instructions from switch block
- instead of calling different function for checking valid terminator for /R and /Rg, i replaced them with single fxn to be called from `gadget.c` based on Gadget type.
- the original switch block was flawed (breaking allow_conditional gadgets) because `ao->type` is always a bitmask and not a simple int enum. So first AND `aop->type` with `_MASK` to get RET and CRET  (regardless of MEM write,  REG, etc bits being set). And in case of CRET return allow_conditional

- TESTs: 
	- updated tests for AVR, RISCV, cmd_rop 
	- The reason for increased gadgets can be explained by AVR's changes, the new `ret` gadgets you see are because previously engine took `call` as valid terminator and when it reach `ret` the count of terminators becomes 2 and hence it discards that gadget, but now as call is considered normal instruction it can be considered in the ret gadget, forming new gadget.
	- for RISCV the PR that added the test (https://github.com/rizinorg/rizin/pull/5952) was to fix regex search for RISCV. The first test is for search with substring, So i simply replace `jal` with `ret` for ROP. The second test was regex search [intruction + operand], so i replace  instruction found near `ret` for regex search. This leaves original regression test intact.
	- I added a new test in `cmd_rop` for my fix of conditional gadgets. 
	- Also in `cmd_rop` i replaced jump and call search with `ret`. Also addition of new gadgets in search is explained above.
---
**NOTE:** there was one test : `search rop for AVR` in `cmd_rop`. This test wrote call and jump gadgets in rizin session and then ran `/R ldi` (filter by string) for searching  and it output all gadgets having `ldi`. But now output for it becomes zero as ROP search is ROP specific now. So i REPLACED that test in `cmd_rop` by writing `ret` gadgets instead. 

And i made a copy of original test in `cmd_cop` and `cmd_jop`,  
My thought is to update this tests in my PR where I implement /J and /C for jop and cop. They will be relevant there. For now "no output" for /R is correct

***What's your suggestion? what should we do to that original test?***

---

**COMMIT 2**: Fix detail search for delay slot archs.
- originally the `/Rg` search for delay slot archs was broken because the code only checked last instruction of gadget. I fixed that by passing `delay_size` for the gadget from upstream and fetching the actual terminal instruction instead of only seeing last instruction.
- TEST: I added a test for MIPS arch, `/Rg` now correctly outputs gadgets. On dev branch it would output nothing.

---

**COMMIT 3:** Added the test that perfectly shows the fix for original issue [#5491](https://github.com/rizinorg/rizin/issues/5491)

---

**COMMIT 4:** updated Doxygen comments in `gadget.c` for delay size, and some should have been updated in refactor PR itself, i am really sorry about missing that!

---

**COMMIT 5:** Add type comment in function arg for hitlist

---

**COMMIT 6/[12-15]** Mem leaks fix , ~50 leaks fixed

- **Commit-6** :
leak in `gadget_print_long_mode`'s disassembly loop.
- not related to my changes, but the trace in CI results pointed to `gadget_print_long_mode`'s `rz_asm_disassemble`.
- the issue was that in the disassembly loop, the `asmop` object was not `fini` at the end of the iteration of loop. It was only called after the loop ended. Each call to `rz_asm_disassemble()` allocated a new internal RzStrBuf (rz_asm_op_set_asm -> rz_strbuf_set), overwriting the previous pointer. This means every iteration [except the last] leaks the previously allocated strbuf.
- i added `rz_asm_op_fini(&asmop);` at end of iteration

- **Commit-12**: RzAnalysisILVM was not cleaned
- **Commit-13**: hitlist was not freed on some fini paths
- **Commit-14**: Fixed 1 leak. OG dev transfered ownership of `context` to `rz_core_gadget_info` and i suspected that context might be leaking when early return `rz_return_val_if_fail` because then the cmd_search.c was not freeing it. Other handlers in cmd_search.c frees context on their own. So i  made it kinda consistent.

- **Commit-15**: I doubted that hitlist might be leaking when `unique_hitlist` was null and `rz_return_val_if_fail` ran due to inconsistent ownership transfers. I am not able to verify though.

-  i would like to point out that OG dev seems to have mixed ownerships at places. Like see function `handle_gadget_list()` in `gadget.c`. Hitlist is marked as RZ_OWN, but actually its freed only at failing paths and success paths just re-transfers ownership to the caller. **This is making it difficult to trace mem freeup logic**. I tried to change it here to RZ_BORROW. Please reply if this makes sense and apply same at other places.

- other leaks solved in : 
    - https://github.com/rizinorg/rizin/pull/6278 and 
    - https://github.com/rizinorg/rizin/pull/6281

---

**NOTE:** I will fix the highlighting of conditional terminal instructions as suggested by @notxvilka  in another PR.

...

**Test plan**

following original issue's binary : `rizin -A test/bins/elf/analysis/mips64r2-busybox`
image

...

**Closing issues**

closes #5491 
...


wargio:
i'm kinda sure that calls are not ret.
image
For me these changes are wrong.

I would expect for rop generator to sequences where you have `return` or `jump registers`, but not call types

@notxvilka

notxvilka:
> i'm kinda sure that calls are not ret. image For me these changes are wrong.
> 
> I would expect for rop generator to sequences where you have `return` or `jump registers`, but not call types
> 
> @notxvilka

Yes, you are right

MrQuantum1915:
> i'm kinda sure that calls are not ret. image For me these changes are wrong.
> 
> I would expect for rop generator to sequences where you have `return` or `jump registers`, but not call types
> 
> @notxvilka

@wargio @notxvilka 

yeah.
ROP sequence dont use call instructions.  Its used in COP (call oriented).

So this means current rop search engine which accepts several call types  should be more restrictive, it should not return true for COP related instructions:

```c
static bool is_end_gadget(const RzAnalysisOp *aop, const ut8 crop) {
	if (aop->family == RZ_ANALYSIS_OP_FAMILY_SECURITY) {
		return false;
	}
	switch (aop->type) {
	case RZ_ANALYSIS_OP_TYPE_TRAP:
	case RZ_ANALYSIS_OP_TYPE_RET:
	case RZ_ANALYSIS_OP_TYPE_UCALL:
	case RZ_ANALYSIS_OP_TYPE_RCALL:
	case RZ_ANALYSIS_OP_TYPE_ICALL:
	case RZ_ANALYSIS_OP_TYPE_IRCALL:
	case RZ_ANALYSIS_OP_TYPE_UJMP:
	case RZ_ANALYSIS_OP_TYPE_RJMP:
	case RZ_ANALYSIS_OP_TYPE_IJMP:
	case RZ_ANALYSIS_OP_TYPE_IRJMP:
	case RZ_ANALYSIS_OP_TYPE_JMP:
	case RZ_ANALYSIS_OP_TYPE_CALL:
		if (crop) {
			return is_cond_end_gadget(aop);
		}
		return true;
	default:
		return false;
	}
}

````

ANd hence both `/R` and `/Rg` consider call as valid terminators.

To fix this properly, I suggest we should fix the root of the problem in the search engine itself.
 I suggest to still keep the removal of the `is_ret_gadget` check in `/Rg` in my PR,  because otherwise it will also accidentally blocks  valid jump gadgets like MIPS `jr $ra`

what do you prefer ?
1.  remove direct call and direct jump types (direct address) from is_end_gadget(). The engine will now search for return and jump registers. we keep indirect ones.
2. make config:  remove them by default, but leave an option to enable them for COP?

**Comment by Rot127** [comment]
> accidentally blocks valid jump gadgets like MIPS jr $ra

Shouldn't this be fixed in `analysis_mips.c` then?

**Comment by MrQuantum1915** [comment]
> > accidentally blocks valid jump gadgets like MIPS jr $ra
> 
> Shouldn't this be fixed in `analysis_mips.c` then?

@Rot127 

thanks!!!
for pointing out that! it helped me find root cause and some other issues in that file

the gadget with `jr $ra` instruction were shown in `/R` but not in `/Rg`.
and as you pointed out, this should have been promoted to `ret` by the analysis phase, but it is NOT.

I checked the type of one of instruction:
image

the type is `rjmp` instead of `ret`

---

i checked `analysis_mips_cs.c` file, in the [analyse function](https://github.com/rizinorg/rizin/blob/1b18421d73888e8518aec480dee041d86ec58f34/librz/arch/p/analysis/analysis_mips_cs.c#L808-L812):
```c
		// register is $ra, so jmp is a return
		if (insn->detail->mips.operands[0].reg == MIPS_REG_RA) {
			op->type = RZ_ANALYSIS_OP_TYPE_RET;
			ctx->t9_pre = UT64_MAX;
		}
```

so the reason is that the **enum** is not matching.
I confirmed it , the value of `MIPS_REG_RA` is 22, while our binary has 331 (just added fprintf):
image

And 331 belongs to `MIPS_REG_RA_64` in Capstone's [`mips.h` file](https://github.com/capstone-engine/capstone/blob/08e655c2b76b02028569d29df01e6f0022734924/include/capstone/mips.h#L366)

Our test binary was `mips64r2-busybox` : 64 bit!

22 is for 32 bit

adding a check for `MIPS_REG_RA_64` solved the bug of jr $ra not being promoted to `ret`.

---

This is a mistake because for Capstone version >=6 (which split 32bit and 64 bit reg classes), we should compare it with 64 bit's enum not 32 bit.

Actually I analysed whole file to see any other such mistake, i found a few other blind spots where we are missing these modern Capstone checks. (NOTE: in somecases we correctly check for 64bit like line [391](https://github.com/rizinorg/rizin/blob/1b18421d73888e8518aec480dee041d86ec58f34/librz/arch/p/analysis/analysis_mips_cs.c#L391), but in other we are missing the check, which can actually break analyses if we test for 64bit mips binary):

List:
1. [Current $ra check](https://github.com/rizinorg/rizin/blob/1b18421d73888e8518aec480dee041d86ec58f34/librz/arch/p/analysis/analysis_mips_cs.c#L796-L812)
2. [Line 468 : ADDIU](https://github.com/rizinorg/rizin/blob/1b18421d73888e8518aec480dee041d86ec58f34/librz/arch/p/analysis/analysis_mips_cs.c#L468)
3. [Line 471 : ADDIU for stack tracking](https://github.com/rizinorg/rizin/blob/1b18421d73888e8518aec480dee041d86ec58f34/librz/arch/p/analysis/analysis_mips_cs.c#L471)
4. [Line 546: ANDI for stack alignment](https://github.com/rizinorg/rizin/blob/1b18421d73888e8518aec480dee041d86ec58f34/librz/arch/p/analysis/analysis_mips_cs.c#L546)

Also Capstone version 6 introduced nanoMIPS, so I think we should add `_NM` checks too for `CS_NEXT_VERSION>=6`, like `MIPS_REG_GP_NM=12`, etc...

---

If you agree, i would like to open a **NEW PR** to fix mips analysis file with this fixes. This will automatically fix issue of `jr $ra` being not promoted to `ret`.

Also how would you prefer I handle the Capstone v6 enums?

1.  Just append the _64 and _NM checks using explicit || logic wrapped in #if CS_NEXT_VERSION >= 6 at the places they are missing.

2. abstract it by adding centralized macros (like  `RZ_MIPS_IS_RA(reg)`) to hide the Capstone versioning checks? Though the only benefit would be avoiding future such mistakes of missing _64 _Nm checks i guess.

MrQuantum1915:
@Rot127 @wargio @notxvilka 

[Ignoring my changes in PR]

Regardless of my previous comment on `analysis_mips_cs.c` issue, the /Rg will still fail for MIPS because of this line in `perform_gadget _analysis` functin [`rop.c`](https://github.com/rizinorg/rizin/blob/1b18421d73888e8518aec480dee041d86ec58f34/librz/core/rop.c#L1408):

```c
	const RzCoreAsmHit *hit_last = (RzCoreAsmHit *)rz_list_last_val(hitlist);
	if (!is_ret_gadget(core, hit_last, crop)) {
		return rop_gadget_info;
	}
```

MIPS gadget has `jr $ra` as SECOND LAST instruciton followed by delay slot like `nop`, ` daddiu sp, sp, 0x90`, etc...

But currently we are only checking if LAST instruction is `ret` or not. We should also check whether LAST SECOND is `ret` or not like this :

```c
	const RzCoreAsmHit *hit_last = (RzCoreAsmHit *)rz_list_last_val(hitlist);
	const RzCoreAsmHit *hit_prev = (RzCoreAsmHit *)rz_list_get_n(hitlist, rz_list_length(hitlist) - 2);
	if (!is_ret_gadget(core, hit_last, crop)) {
		if (!hit_prev || !is_ret_gadget(core, hit_prev, crop)) {
			return rop_gadget_info;
		}
	}
```
**_NOTE_:** this is just temp fix, like using `rz_list_get_n` will be O(N) in linked list, we can optimise that part, and also we can only trigger second check only when there is delay slot , to avoid false positives.

After this change `/Rg` correctly identifies and analyse MIPS gadgets correctly.

---

Also we need to decide now whether `/R` showing more gadgets then `/Rg` by including COP related instructions, is EXPECTED behaviour or not.

- If we want /R to list more gadgets which also have `call` instructions, then the Original Issue of inconsistent rop info becomes Invalid. And this would mean `/R` is intended to be a broader raw search than the strict `/Rg` analysis.

- If this is NOT expected than we should remove `call` type from valid end_gadget check as discussed in my prior comment.

notxvilka:
A good point. JOP and COP are sometimes useful but quite niche, thus calls should not be a part of the default ROP chain listing. I suggest moving them into a separate command probably. Maybe part of the new `/J` command?

MrQuantum1915:
> A good point. JOP and COP are sometimes useful but quite niche, thus calls should not be a part of the default ROP chain listing. I suggest moving them into a separate command probably. Maybe part of the new `/J` command?

@notxvilka 

hmm...makes sense.

So for ROP we should remove JMP and CALL types from checks.

```c
static bool is_end_gadget(const RzAnalysisOp *aop, const ut8 crop) {
	if (aop->family == RZ_ANALYSIS_OP_FAMILY_SECURITY) {
		return false;
	}
	switch (aop->type) {
	case RZ_ANALYSIS_OP_TYPE_TRAP:
	case RZ_ANALYSIS_OP_TYPE_RET:
	case RZ_ANALYSIS_OP_TYPE_UCALL:
	case RZ_ANALYSIS_OP_TYPE_RCALL:
	case RZ_ANALYSIS_OP_TYPE_ICALL:
	case RZ_ANALYSIS_OP_TYPE_IRCALL:
	case RZ_ANALYSIS_OP_TYPE_UJMP:
	case RZ_ANALYSIS_OP_TYPE_RJMP:
	case RZ_ANALYSIS_OP_TYPE_IJMP:
	case RZ_ANALYSIS_OP_TYPE_IRJMP:
	case RZ_ANALYSIS_OP_TYPE_JMP:
	case RZ_ANALYSIS_OP_TYPE_CALL:
		if (crop) {
			return is_cond_end_gadget(aop);
		}
		return true;
	default:
		return false;
	}
}
```

For ROP we should keep:
```c
case RZ_ANALYSIS_OP_TYPE_TRAP:
case RZ_ANALYSIS_OP_TYPE_RET:
```

For COP:
```c
case RZ_ANALYSIS_OP_TYPE_UCALL:
case RZ_ANALYSIS_OP_TYPE_RCALL:
case RZ_ANALYSIS_OP_TYPE_ICALL:
case RZ_ANALYSIS_OP_TYPE_IRCALL:
case RZ_ANALYSIS_OP_TYPE_CALL:
```

For JOP:
```c
case RZ_ANALYSIS_OP_TYPE_UJMP:
case RZ_ANALYSIS_OP_TYPE_RJMP:
case RZ_ANALYSIS_OP_TYPE_IJMP:
case RZ_ANALYSIS_OP_TYPE_IRJMP:
case RZ_ANALYSIS_OP_TYPE_JMP:
```

We can then use them in their respective check

**IMP:** By doing so I am assuming that specific architecture analysis plugins correctly promote their specific return idioms (like MIPS `jr $ra` or ARM `bx lr`) to `RZ_ANALYSIS_OP_TYPE_RET`. As I discussed in earlier comment to @Rot127 ...


We can implement this in a new `jop.c`:
```bash
/J : List all indirect branch gadgets (both JOP and COP).

/Jj: List only pure JOP gadgets (ends in jmp).

/Jc: List only COP gadgets (ends in call).
````
---


- So for **THIS PR:** i will remove JOP and COP related checks from `is_end_gadget` function (fixing inconsistency between /R and /Rg), and also FIX to also check the second last instruction for delay slot arch (fixing /Rg for MIPS) as discussed above...

- And  I will open a **NEW PR**, for implementing new `/J` command for JOP and COP.

Does this make sense? Should I proceed?

**Comment by notxvilka** [comment]
> * So for **THIS PR:** i will remove JOP and COP related checks from `is_end_gadget` function (fixing inconsistency between /R and /Rg), and also FIX to also check the second last instruction for delay slot arch (fixing /Rg for MIPS) as discussed above...
> 
> * And  I will open a **NEW PR**, for implementing new `/J` command for JOP and COP.

Yes, exactly. Be sure to mimic `/J` output and subcommands to the corresponding `/R` output and subcommands. Maybe some code can be shared between two (printing functions, etc).

File: test/db/cmd/cmd_cop [hunk 1]
```diff
@@ -0,0 +1,13 @@
...
+CMDS=<<EOF
+e asm.arch=avr
+e cfg.bigendian=false
+e scr.color=false
+wx 11241fbecfefd8e0debfcdbf12e0a0e0b1e0e6e3f4e302c005900d92a431b107d9f722e0a4e1b2e001c01d92ae3cb207e1f710e0c2e6d0e004c02197fe010e94041ac136d107c9f70e944f090c9456160c940000ef92ff920f931f93cf93df93cdb7deb7c054d1090fb6f894debf0fbecdbf6ce071e08fe192e00e946b04ce0101967c018e010f5c1f4ffc011192e017f107e1f720e330e0a7016ae08fe192e00e94d8080297f8f0c801d701fd01ee19ff092d91222321f0fc012193cf01f6cfe00ff11f108260e071e0c8010e949816892b49f46de171e08fe192e00e946b0481e090e008c066e471e08fe192e00e946b0480e090e0c05cdf4f0fb6f894debf0fbecdbfdf91cf911f910f91ff90ef9008956de671e08fe192e00e946b0485b1805885b908956fe771e08fe192e00e946b048bb180588bb908956de671e08fe192e00e946b0485b1805885b985b1805885b985b1805885b908956fe771e08fe192e00e946b048bb180588bb98bb180588bb98bb180588bb908950f931f93cf93df93cdb7deb727970fb6f894debf0fbecdbf9c018e010f5f1f4fd801ad01401b510bf90191919f01992311f09d93f6cffa01e00ff11f10826fe871e0c8010e949816892b81f40e948c00892bf9f027960fb6f894debf0fbecdbfdf91cf911f910f910c940f0163e971e0c8010e949816892b61f427960fb6f894debf0fbecdbf
+/R ldi
+q
+EOF
+EXPECT=
+EXPECT_ERR=
```
wargio: no output?

MrQuantum1915: sorry i forgot to mark this PR as draft, 

i was gonna update the PR description  to ask what to do with this test.
anyways the thing is that `cmd_rop` test had a test for AVR: `search rop for AVR`

this test wrote call and jump gadgets in rizin session and then ran `/R ldi` (filter by string) for searching
and it output all gadgets having `ldi`.

but as now ROP search is restricted to ROP only, it does not output any call and jump oriented gadgets = no output

so i REPLACED that test in `cmd_rop`  by writing `ret` gadgets instead.

and i made a copy of original test in `cmd_cop` and `cmd_jop`, 
my thought is to update this tests in my PR where I implement /J and /C for jop and cop. They will be relevant there.

For now "no output" for /R is correct

What's your suggestion? what should we do to that original test?

MrQuantum1915:
from 18 new leaks detected (from my code changes); although none are due to my changes;

i fixed the one in `gadget.c` from the trace shown in CI results (details in PR description: COMMIT 6)

Other 17 leaks seems to be upstream leaks and not related to direct code of `gadget.c`
all those trace to common path:

```bash
rz_core_il_step()
|-- rz_analysis_il_vm_step_while()
    |-- analysis_il_vm_step_while()
        |-- rz_analysis_op()
        |-- rz_il_vm_step()
````

File: librz/core/gadget.c [hunk 2]
```diff
@@ -37,14 +37,25 @@ static bool gadget_is_valid_terminator(const RzGadgetType gadget_type, const RzC
...
 
+static RzCoreAsmHit *find_gadget_terminator(const RzPVector /*<RzCoreAsmHit *>*/ *hitlist, int delay_size) {
+	if (!hitlist || delay_size < 0) {
+		return NULL;
+	}
+	const size_t len = rz_pvector_len(hitlist);
+	if ((size_t)delay_size >= len) {
+		return NULL;
+	}
+	return rz_pvector_at(hitlist, len - 1 - (size_t)delay_size);
```
wargio: shouldn't the if be `delay_size + 1 >= len` ?

MrQuantum1915: hmm...
actually it should be `delay_size + 1 > len`
fixed that

test/db/cmd/cmd_cop
wargio: if this is ok, then i'm resolving this comment

test/db/cmd/cmd_jop
```diff
@@ -0,0 +1,13 @@
+NAME=search rop for AVR
+FILE=malloc://512
+CMDS=<<EOF
+e asm.arch=avr
+e cfg.bigendian=false
+e scr.color=false
+wx 11241fbecfefd8e0debfcdbf12e0a0e0b1e0e6e3f4e302c005900d92a431b107d9f722e0a4e1b2e001c01d92ae3cb207e1f710e0c2e6d0e004c02197fe010e94041ac136d107c9f70e944f090c9456160c940000ef92ff920f931f93cf93df93cdb7deb7c054d1090fb6f894debf0fbecdbf6ce071e08fe192e00e946b04ce0101967c018e010f5c1f4ffc011192e017f107e1f720e330e0a7016ae08fe192e00e94d8080297f8f0c801d701fd01ee19ff092d91222321f0fc012193cf01f6cfe00ff11f108260e071e0c8010e949816892b49f46de171e08fe192e00e946b0481e090e008c066e471e08fe192e00e946b0480e090e0c05cdf4f0fb6f894debf0fbecdbfdf91cf911f910f91ff90ef9008956de671e08fe192e00e946b0485b1805885b908956fe771e08fe192e00e946b048bb180588bb908956de671e08fe192e00e946b0485b1805885b985b1805885b985b1805885b908956fe771e08fe192e00e946b048bb180588bb98bb180588bb98bb180588bb908950f931f93cf93df93cdb7deb727970fb6f894debf0fbecdbf9c018e010f5f1f4fd801ad01401b510bf90191919f01992311f09d93f6cffa01e00ff11f10826fe871e0c8010e949816892b81f40e948c00892bf9f027960fb6f894debf0fbecdbfdf91cf911f910f910c940f0163e971e0c8010e949816892b61f427960fb6f894debf0fbecdbf
+/R ldi
+q
```
wargio: you do not need to run `q`

test/db/cmd/cmd_cop
```diff
@@ -0,0 +1,13 @@
+NAME=search rop for AVR
+FILE=malloc://512
+CMDS=<<EOF
+e asm.arch=avr
+e cfg.bigendian=false
+e scr.color=false
+wx 11241fbecfefd8e0debfcdbf12e0a0e0b1e0e6e3f4e302c005900d92a431b107d9f722e0a4e1b2e001c01d92ae3cb207e1f710e0c2e6d0e004c02197fe010e94041ac136d107c9f70e944f090c9456160c940000ef92ff920f931f93cf93df93cdb7deb7c054d1090fb6f894debf0fbecdbf6ce071e08fe192e00e946b04ce0101967c018e010f5c1f4ffc011192e017f107e1f720e330e0a7016ae08fe192e00e94d8080297f8f0c801d701fd01ee19ff092d91222321f0fc012193cf01f6cfe00ff11f108260e071e0c8010e949816892b49f46de171e08fe192e00e946b0481e090e008c066e471e08fe192e00e946b0480e090e0c05cdf4f0fb6f894debf0fbecdbfdf91cf911f910f91ff90ef9008956de671e08fe192e00e946b0485b1805885b908956fe771e08fe192e00e946b048bb180588bb908956de671e08fe192e00e946b0485b1805885b985b1805885b985b1805885b908956fe771e08fe192e00e946b048bb180588bb98bb180588bb98bb180588bb908950f931f93cf93df93cdb7deb727970fb6f894debf0fbecdbf9c018e010f5f1f4fd801ad01401b510bf90191919f01992311f09d93f6cffa01e00ff11f10826fe871e0c8010e949816892b81f40e948c00892bf9f027960fb6f894debf0fbecdbfdf91cf911f910f910c940f0163e971e0c8010e949816892b61f427960fb6f894debf0fbecdbf
+/R ldi
+q
```
wargio: same here

wargio (commented):
for me it is a bit weird that there are tests not outputting anything. maybe change the name of the test and print the disassembly

MrQuantum1915:
> for me it is a bit weird that there are tests not outputting anything. maybe change the name of the test and print the disassembly

i updated both tests, does it make more sense now?? https://github.com/rizinorg/rizin/pull/6108/changes/160146aeb799f0eab031234cef3674550b862610

test/db/cmd/cmd_rop
```diff
@@ -3472,9 +3818,23 @@ Gadget 0x8048641 (size 2 bytes)
   0x08048641  5d               pop ebp | Stack change: 0x8
   0x08048642  c3               ret     | Modified regs: ebp esp
 
+Gadget 0x8048655 (size 9 bytes)
+------------------------------------------------------------------------------------------------------
+  0x08048655  e8c6fdffff       call 0x8048420 | Stack change: 0xffffffff00000010
```
wargio, line 3823: shouldn't this just be `0x10` since the bitness is 32?

MrQuantum1915: yeah thats wrong in case of 32bit  setting

the reason is in calculation of stack change and the datatype of stackchange variable.

stack_change is `ut64` 
and see this line
```c
gadget_info->stack_change += rz_bv_to_ut64(new_val) - rz_bv_to_ut64(init_val);
```

the subtraction is done as unsigned 64bit

in 32bitness, when `new_val`< `init_val`, the delta wraps modulo 2^64 
this causes the 32bit delta to wrap into 64bit 

results in  `0xffffffff00000010` instead of a signed negative change



I am opening a new PR to fix this

also should we just show the magnitude of stack_change and 2s complement or along with sign to show stack increase or decrease?

then we will need to change `stack_change` variable to `st64` instead of `ut64`

wargio: that subtraction should be done using bitvectors.

wargio (approved):

notxvilka (approved):
LGTM but please wait for @moste00 first to check the RISC-V ROP output changes in tests first.

MrQuantum1915:
> LGTM but please wait for @moste00 first to check the RISC-V ROP output changes in tests first.

 sorry but i didnt get what you mean?
 do you mean @moste00  to check my currrent output of RISCV ROP test??
 
 or do you mean his [PR#6230](https://github.com/rizinorg/rizin/pull/6230) to be merged before this?
 (his PR is marked as archive PR though)

notxvilka:
> do you mean @moste00 to check my currrent output of RISCV ROP test??

Yes, just to verify

moste00:
@MrQuantum1915 Sorry I was not following this, looks good at first glance. 

From what I understand this is just a cleanup of different behaviours that the `/R*` family of commands had unintentionally, but as long as **some** gadgets end up appearing then the RISC-V tests are correct.

Just a couple of questions:

1- Can the original behaviour (looking for JAL gadgets) ever be recovered or is it now impossible in Rizin? If impossible, then didn't we break the workflow for the original reporter of the RISC-V bug who gave the example of searching for the JAL instruction ? I think this is a pretty breaking change right ?

2- If the answer to (1) is yes it's possible, then we can leave intact the original test that looks for JAL right ? I suppose the answer is no though.

3- What other types of terminators are valid other than `ret` ? I'm asking because I have the impression that if you eliminate COP and JOP, then isn't the only terminator left is `ret` (i.e. ROP), and in this case it wouldn't make sense to use it as a string filter on the output of /R (since all gadgets will have `ret` in them anyway so there is no real "filtering" going on). But I'm not sure, are there gadgets that don't contain "ret" in their disassembly at all?

moste00:
To be clear my long comment can be summarized as "LGTM", I'm just raising concerns that appear serious to me. Sorry if I'm rehashing things already discussed before but I wasn't following this decision chain in mattermost. My preference is just that tests remain stable (hence why I'm asking about recovering old behaviour) and users are surprised as least as possible.

MrQuantum1915:
@moste00 

thanks for your review, i will answer it pointwise

1. Temporarily yes for now , jal gadget's can't be recovered using `/R` as its restricted to ROP. **BUT** not permanently. Once /J and /C will be implemented (JOP and COP search) (which i am working on currently), they will again be recovered but using /J and /C command.

---

2. NO we cant, using jal  returns only one gadget now, which has ret as terminator and `jalr` before it. So I needed to replace the test.
image

---

3. ON HIGHER level, the bug was "REGEX search not work in riscv", 

so  you are Right for 1st test (/R/ jal); using `/R/ ret` doesnt make a difference  (i was used to old mix output  rop+jop+cop, sorry for that), i will change it to `/R/  ld` (targeting function epilogue gadgets) and it does work, it outputs 13 Gadgets 
image

while normal /R/ returns 35 gadgets
image


For The other test i replaced (`/R/ "jalr.*t1.*"`) with (`/R/ "ld.*ra.*sp"`) as they come in context of `ret`  . This test is to test REGEX search, which does get satified by my new test. So i dont think we need to change that test to something else

MrQuantum1915:
@moste00 @notxvilka 

updated 1st riscv test, check it

moste00:
@MrQuantum1915 Nice, thank you. Would it make sense to keep the old tests then and add BROKEN=1 to them? then removing the BROKEN when you JOP and COP?

Your current tests could be new tests unrelated to the JAL ones.

MrQuantum1915:
> @MrQuantum1915 Nice, thank you. Would it make sense to keep the old tests then and add BROKEN=1 to them? then removing the BROKEN when you JOP and COP?
> 
> Your current tests could be new tests unrelated to the JAL ones.

yeah makes sense
i am fine with it if u want to keep original issue's exact instructions.

i will revert

MrQuantum1915:
@moste00 @notxvilka @wargio 

i think we can merge now

moste00:
> @moste00 @notxvilka @wargio
> 
> i think we can merge now

Btw you can keep the expectations (`EXPECT=`) in the broken tests, broken tests are not run anyway so no attempt to diff output would be made for them, keep the old expectation so that when you add JOP later you can compare the output of your new implementation with the old implementation you replaced.

MrQuantum1915:
@moste00 
done

notxvilka (approved):

wargio (approved):

librz/core/gadget.c
```diff
@@ -1391,7 +1402,7 @@ static RzPVector /*<RzCoreAsmHit *>*/ *construct_gadget(RzCore *core, ut8 *buf,
 	return hitlist;
 }
 
-static RzGadgetInfo *perform_gadget_analysis(const RzGadgetType type, RzCore *core, const bool allow_conditional, const RzPVector /*<RzCoreAsmHit *>*/ *hitlist) {
+static RzGadgetInfo *perform_gadget_analysis(const RzGadgetType type, RzCore *core, const bool allow_conditional, const RzPVector /*<RzCoreAsmHit *>*/ *hitlist, int delay_size) {
```
Rot127, line 1405: `delay_size` can be negative? Or why is it not an unsigned?

librz/core/gadget.c
```diff
@@ -444,10 +444,9 @@ RZ_API RZ_OWN RzGadgetRegInfo *rz_core_gadget_reg_info_dup(RZ_BORROW RZ_NONNULL
 
 /**
  * \brief Find the Gadget Register information for the given register
- * \param gadget_info Pointer to the RzGadgetInfo
- * \param name Register to filter the dependencies
- *
- * \return Pvector of register gadget information
+ * \param gadget_info RZ_NONNULL Pointer to the RzGadgetInfo object.
```
Rot127: NONNULL is already in the declaration. Remove duplicate here.
Also below

Rot127 (changes requested):
Please also fix the reported leaks.

librz/core/gadget.c
wargio: it's `int` because is `int` in our struct. i agree. should be `ut8/16`

MrQuantum1915: no it cant be negative infact not even exceed 3 in any arch

i just followed previous definition
original dev used int in the struct definition for `Endlistpair`
see
```c
/**
 * \brief Pair representing an end gadget with instruction offset and delay size.
 */
typedef struct rz_gadget_endlist_pair_t {
	int instr_offset; ///< Instruction offset.
	int delay_size; ///< Delay size.
} RzGadgetEndListPair;
```

i again checked the code, both instruction offset and delaysize will not be ever negative

i would need to change this struct for your suggestion

for alignment would use `ut32` or `ut64` for it (using ut8 doesnt make a difference)

i am not sure whether use ut32 or ut64 here
because where the `instr_offset` is calculated itself has 1 issue

see `delta` is defined as `int` (the 64bit address diff),

```c
const int delta = context->to - context->from;
for (int i = 0; i < delta; i += context->increment) {
.
.
.

		if (gadget_is_valid_end_gadget(context->type, &end_gadget, context->allow_conditional)) {
			RzGadgetEndListPair *epair = RZ_NEW0(RzGadgetEndListPair);
			if (epair) {
				epair->instr_offset = i + (end_gadget.delay ? context->increment : 0);
				epair->delay_size = end_gadget.delay;
				rz_list_append(end_list, epair);
			}
		}
.
.
.
	}
``` 

the same delta is defined as `ut64` in other functions which seems to be fine as storing 64bit difference

according to this, instr_offset does not exceed int_max

i would suggest changin that delta to ut64 similarly to other places in code
and 
then change instr_offset to ut64

but then what to use for delay_size??
in the struct definition using ut8/ut16/ut32 will not make a difference due to alignment anyway?

MrQuantum1915:
> Please also fix the reported leaks.

although not due to my changes

i tried to solve them, 
from 18leaks i solved one [above](https://github.com/rizinorg/rizin/pull/6108#issuecomment-4248125812) 
other 17seems to upstream

but i will  take a deeper look again if i can solve :)

librz/core/gadget.c
Rot127: size_t is perfectly fine I think.

MrQuantum1915: hey 
changing them to `size_t` is requiring more changes then i thouhgt
like it is cascading to many more variables and also some `RzAnalysisOp` fields too (size, delay)

e.g.

`i` of for loop and delay field of RzAnalysisOp is int
```c
epair->instr_offset = i + (end_gadget.delay ? context->increment : 0);
epair->delay_size = end_gadget.delay;
```

`idx` and `aop.size` both are `int`
```c
if (end_gadget->instr_offset <= idx - aop.size) {
			valid = end_gadget->instr_offset == idx - aop.size;
			goto cleanup;
}
```
line 1804 - `i` is  int*
```c
	if (*i > (*end_gadget)->instr_offset) {
```

1811
```c
*i = (*end_gadget)->instr_offset - gadget_depth;
```
1852, delta is ut64 but `i` is int (i changed `next` to size_t)
```c
const size_t next = end_gadget->instr_offset;
	for (int i = 0; i < delta && context->max_count; i += context->increment) {
	.
	.
	
	if (i > next && !update_end_gadget(&i, gadget_depth, &end_gadget, context)) {
			break;
	}
```

seems like going into a rabbit hole :(

MrQuantum1915: the RzAnalysisOp fields are radare era's

the delay is itself int :(
See
https://www.rada.re/vdoc/libr/Radare.RAnal.Op.html

librz/core/gadget.c
MrQuantum1915: i changed other OG duplication too*

librz/core/gadget.c
notxvilka: I would say lets merge this as is, and address remaining issues in separate PRs

MrQuantum1915:
> > Please also fix the reported leaks.
> 
> although not due to my changes
> 
> i tried to solve them, from 18leaks i solved one [above](https://github.com/rizinorg/rizin/pull/6108#issuecomment-4248125812) other 17seems to upstream
> 
> but i will take a deeper look again if i can solve :)

looks like in last commit i solved ~8 more leaks 
Total leaks down to 40475 from ~40500

i thought that was the reason of those 17 leaks
but sadly those 17 leaks are still there, trying again...

MrQuantum1915:
@Rot127 @wargio @notxvilka 
found another mem leak of hitlist https://github.com/rizinorg/rizin/pull/6108/commits/d95e4bd0f998b316b9092a9507dcfca55999ee7e

the one more leak from that ci trace is in `theory_bitv.c`, fixed in new PR : https://github.com/rizinorg/rizin/pull/6278

hope those 17leaks dont come again :/

MrQuantum1915:
dont know why bindgen linter is failing from last 2 commit, 

doesnt relate to my changes

image

notxvilka:
> dont know why bindgen linter is failing from last 2 commit,
> 
> doesnt relate to my changes
> image

Ignore it

MrQuantum1915:
umm...
11 more leaks solved down to 40464 from 40500

but those 17 leak came again... :(

1 will solved by https://github.com/rizinorg/rizin/pull/6278

so 16 remains

librz/core/gadget.c
wargio: yes, i was going to write this too. open a new PR with these changes.

MrQuantum1915:
@notxvilka @Rot127 @wargio 

FINALLY...!

found the leak for those 16 from trace :/

leaks down to 40454 from 40500

image


they are in `arm_il64.c`

i will revert and fix in PR https://github.com/rizinorg/rizin/pull/6281

will rebase after that gets merged

MrQuantum1915:
@notxvilka @wargio @Rot127 

i have updated the PR description for all mem leak fixes, around 50 now

i would like your review on issue told in commit-15 though:
> i would like to point out that OG dev seems to have mixed ownerships at places. Like see function handle_gadget_list() in gadget.c. Hitlist is marked as RZ_OWN, but actually its freed only at failing paths and success paths just re-transfers ownership to the caller. **This is making it difficult to trace mem freeup logic**. I tried to change it here to RZ_BORROW. Please reply if this makes sense and apply same at other places.

librz/core/gadget.c
MrQuantum1915: hey all i just found out this issue : https://github.com/rizinorg/rizin/issues/6061

i highly suspect that this crash is caused by the usage of `int` instead of ut64 for `delta` above (on smaller device)...
causing sigsegv

i will try to reproduce the crash if possible
[after a week*]

test/db/analysis/riscv
```diff
@@ -3903,10 +3903,11 @@ RUN
 
 NAME=ROP Search by substring
 FILE=bins/elf/riscv_crypto_64
+BROKEN=1
```
Rot127, line 3906: Maybe missed the discussion. But those two tests are marked broken for a reason?

test/db/cmd/cmd_rop
```diff
@@ -3115,11 +3310,19 @@ Gadget 0x804840c (size 9 bytes)
 
 Gadget 0x804840e (size 7 bytes)
 ------------------------------------------------------------------------------------------------------
-  0x0804840e  02ff             add bh, bh                | Stack change: 0xffffffff00000008
+  0x0804840e  02ff             add bh, bh                | Stack change: 0x8
   0x08048410  d0585b           rcr byte [eax+0x5b], 0x01 | Modified regs: ebx esp ebp
   0x08048413  c9               leave                     | Dependencies:  ebx eax eax ebp esp esp esp esp
   0x08048414  c3               ret                       | 
 
+Gadget 0x804840f (size 6 bytes)
```
Rot127, line 3318: I count 8 bytes here, not 6. Is this a bug?

Rot127 (approved):
Lgtm. Please check the comments nonetheless. 
Thanks for fixing the leaks!

test/db/analysis/riscv
MrQuantum1915: yeah, see the discussiojn thread here: https://github.com/rizinorg/rizin/pull/6108#issuecomment-4262571417

test/db/cmd/cmd_rop
MrQuantum1915: 0x0804840f , 0x08048410
0x08048411
0x08048412
0x08048413
0x08048414

isnt it 6??

Rot127: Sorry, I am stupid and looked at the wrong line.

MrQuantum1915: happens :)
Thanks for checking...

wargio (approved):

notxvilka:
@MrQuantum1915 got broken on big endian machine:
```
[XX] 00h00m00s205ms db/cmd/cmd_rop  detail rop search  (/Rg)  on mips (1 delay slot)
RZ_COLOR=0 RZ_UTF8=0 RZ_NOPLUGINS=1 /home/travis/build/rizinorg/rizin/install/bin/rizin -escr.utf8=0 -escr.color=0 -escr.interactive=0 -eflirt.sigdb.load.system=false -esearch.show_progress=false -eflirt.sigdb.load.home=false -N -qc 'e asm.arch=mips
e asm.bits=32
wa "addiu sp, sp, 0x20; jr ra; nop"
pi 3
echo "================================================"
/R
echo "================================================"
/Rg 
' =
-- stdout
--- expected
+++ actual
@@ -2,12 +2,12 @@
 jr ra
 nop
 ================================================
-  0x00000000           2000bd27  addiu sp, sp, 0x20
-  0x00000004           0800e003  jr ra
+  0x00000000           27bd0020  addiu sp, sp, 0x20
+  0x00000004           03e00008  jr ra
   0x00000008           00000000  nop
 Gadget size: 12
 
-  0x00000004           0800e003  jr ra
+  0x00000004           03e00008  jr ra
   0x00000008           00000000  nop
 Gadget size: 8
 
[XX] 00h00m00s188ms db/cmd/cmd_rop  search rop with conditional terminator
RZ_COLOR=0 RZ_UTF8=0 RZ_NOPLUGINS=1 /home/travis/build/rizinorg/rizin/install/bin/rizin -escr.utf8=0 -escr.color=0 -escr.interactive=0 -eflirt.sigdb.load.system=false -esearch.show_progress=false -eflirt.sigdb.load.home=false -N -qc 'e asm.arch=arm
e asm.bits=32
wx 0080bd08
pi 1
/R
echo "==============================================="
e gadget.conditional=true
/R
' =
-- stdout
--- expected
+++ actual
@@ -1,5 +1,2 @@
-ldmeq sp!, {pc}
+addeq fp, r0, r8, lsl 26
 ===============================================
-  0x00000000           0080bd08  ldmeq sp!, {pc}
-Gadget size: 4
-
```
https://app.travis-ci.com/github/rizinorg/rizin/jobs/639049055#L4457

MrQuantum1915:
@notxvilka 
yeah i just saw that and was gonna fix it in https://github.com/rizinorg/rizin/pull/6242
just needs a test update
is it fine?

or rather blocking it ,  opening another fix PR

really sorry for missing that

PR #6109, opened 2026-03-27
Add RISC-V 32-bit env to CI

moste00:

**Detailed description**

This PR adds a RISC-V 32-bit build to the CI. This is more challenging than the RISC-V 64-bit build because it requires compilation of Linux with a RISC-V 32-bit toolchain, whereas the RISC-V 64-bit build can simply depend on a mainstream kernel from, e.g., Debian or Ubuntu.

In addition to the kernel, it's also easier if we compile from source both the booting firmware, OpenSBI, and the minimal userspace, busybox. All 3 are compiled here and the built artifacts are published for downloading. https://github.com/moste00/riscv-minstack.

What this CI mainly does then is downloading the Bootlin toolchain from https://toolchains.bootlin.com/ and cross-compiling Rizin using it. This toolchain is prebuilt so it doesn't require compilation like gcc's RISC-V toolchain, Bootlin is also how minstack repo compiles the kernel. 

**Test plan**

The ultimate goal of this PR is be able to run the debugger tests defined here https://github.com/rizinorg/rizin/pull/5966 inside a RISC-V 32-bit environment. But in order to reduce the coupling between PRs to avoid messy rebases, this goal is replaced with the simpler goal of simply running `rizin -hh`, which is what the booting job ultimately does.

When this PR is merged, the #5966 PR will modify the booting job to run the debugger tests.

**Closing issues**

No existing issue, but running RISC-V inside QEMU to run tests is supposed to be a temporary solution. Running on real RISC-V hardware is the ideal end-state, solutions like https://riseproject.dev/2026/03/24/announcing-the-rise-risc-v-runners-free-native-risc-v-ci-on-github/ promise to do this conveniently for free. The QEMU build after this will be mainly for documentation purposes to demonstrate the ability of Rizin to be cross-compiled for RISC-V on a non-RISC-V machine.

### DIFF
--- a/.github/riscv32-qemu-setup-all.sh
+++ b/.github/riscv32-qemu-setup-all.sh
@@ -1,0 +1,63 @@
+#!/bin/sh
+set -eux
+
+MINSTACK_VERSION=v12
+
+# Download the linux kernel image, opensbi firmware, and the busybox userspace from the minstack repo
+wget -q "https://github.com/moste00/riscv-minstack/releases/download/${MINSTACK_VERSION}/rv32-Image"
+wget -q "https://github.com/moste00/riscv-minstack/releases/download/${MINSTACK_VERSION}/rv32-fw_dynamic.bin"
+wget -q "https://github.com/moste00/riscv-minstack/releases/download/${MINSTACK_VERSION}/rv32-busybox"
+
+#---------------------------------------------------------------------------------------------------------------------
+cd rizin
+# Setup cross file
+sed -i "s|<<path-to-bootlin-installation>>|${BOOTLIN_INSTALLATION}|g" doc/examples/cross_builds/meson-linux-riscv32.ini
+export PATH="${BOOTLIN_TOOLS}:${PATH}"
+
+# Build Rizin
+meson setup build --cross-file doc/examples/cross_builds/meson-linux-riscv32.ini
+meson compile -C build || true # Expected to fail due to Zydis bug on RISC-V 32-bit
+# Quick and dirty fix for Zydis compilation error, remove when upstream merges a fix and we bump
+patch -p1 < patches/fix_zydis_amalgamated_riscv32_build
+# Build again
+meson compile -C build
+
+file build/binrz/rizin/rizin
+cd ..
+#---------------------------------------------------------------------------------------------------------------------
+
+# Create a init config for busybox
+cat << EOF > inittab
+::sysinit:/mount -t proc proc /proc
+::sysinit:/mount -t sysfs sysfs /sys
+::sysinit:/mount -t devtmpfs devtmpfs /dev
+::sysinit:/mount -t tmpfs -o size=128M,nodev,nosuid tmpfs /tmp
+::wait:$@
+::once:/poweroff -f
+EOF
+
+# Describe the boot filesystem
+cat << EOF > initramfs
+dir /bin 0755 0 0
+dir /etc 0755 0 0
+dir /lib 0755 0 0
+dir /tmp 1777 0 0
+dir /proc 0755 0 0
+dir /sys 0755 0 0 
+dir /dev 0755 0 0
+
+nod /dev/console 0600 0 0 c 5 1
+
+file /bin/busybox rv32-busybox 0755 0 0
+slink /init /bin/busybox 0777 0 0
+slink /mount /bin/busybox 0777 0 0
+slink /poweroff /bin/busybox 0777 0 0
+file /etc/inittab inittab 0644 0 0
+
+file /lib/ld-musl-riscv32.so.1 ${BOOTLIN_INSTALLATION}/riscv32-buildroot-linux-musl/sysroot/lib/ld-musl-riscv32.so.1 0755 0 0
+file /bin/rizin rizin/build/binrz/rizin/rizin 0755 0 0
+EOF
+
+# Generate the boot filesystem
+gcc rizin/sys/gen_cpio_script.c -O3 -o gen_cpio
+./gen_cpio initramfs > initrd_rv32.cpio
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -937,6 +937,7 @@ jobs:
           .\rizin_cmake_test
 
   test-riscv-inside-qemu:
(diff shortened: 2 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -388,7 +390,7 @@ jobs:
   build-centos7:
     name: Build on CentOS 7
     runs-on: ubuntu-24.04
-    if: contains(github.head_ref, 'dist') || contains(github.head_ref, 'centos') || contains(github.ref, 'release-') || github.ref == 'refs/heads/stable' || github.event_name == 'schedule'
```
notxvilka, line 391: For this to be approved please restore other workers

notxvilka (changes requested):

.github/workflows/ci.yml
moste00: Of course, but is there any other thing to change in the `setup-all` script itself or the kernel build script ? I like to make the restoring of workers the very last thing to do because it makes every test run expensive.

.github/workflows/ci.yml
```diff
@@ -388,7 +390,7 @@ jobs:
...
+    if: false
```
wargio: revert

wargio (changes requested):
while i'm ok with this, i think the whole build of the image should go in a standalone repository where we publish rv32 images which rizin pulls.

moste00:
> while i'm ok with this, i think the whole build of the image should go in a standalone repository where we publish rv32 images which rizin pulls.

Just the Linux image and the opensbi and busybox binaries right ? The Rizin binary itself should still be built here.

wargio:
yes

Rot127:
@moste00 I converted it to a draft. Please mark it ready for review when everything is done.

doc/examples/cross_builds/meson-linux-riscv32.ini
Rot127, line 1: Those are for Debian based builds? Add a comment about it to the header. Or rename the file. Since the compiler package has a different name of different OS.

moste00: It's not for any distro specifically, the Bootlin toolchain makes it almost fully statically linked, needing only `musl` libc at runtime. The actual usage in the CI is booting it under a custom-compiled very minimal Linux kernel and a busybox userspace, but I suppose you could boot it under any distro if needed.

moste00: (There are almost no mainstream distros with 32-bit RISC-V builds, that's why I had to custom-compile all the toolchain including the kernel and the firmware. That said, Buildroot can bootstrap a 32-bit build from nothing quite easily, but the default config takes 1+ hour to build.)

moste00:
@wargio @Rot127 hey, I think this is ready. There are only two things that might raise objections:

1- There is a `sed -i` hack to modify an amalgamated file in Zydis build directory called `zydis.c`, this is because there is a bug in Zydis where they declare a function conditionally depending on whether they're compiling on certain architectures, but this function is used unconditionally later in the file. RISC-V isn't on the architecture condition so the function is never declared and the build fails. 

The `sed -i` hack just adds RISC-V to the list of architectures that triggers the declaration. I searched for this file in Zydis main repo to fix it there but I didn't find it, so it will take a while to track down the problem at the root.

2- I vendored a `gen_init_cpio` script from the Linux kernel to allow creating a CPIO archive containing a `/dev/console` device file, this can't be done by the normal `cpio` command. I don't remember exactly why that is necessary but the guest shell in QEMU won't work without it, probably the build for either Linux or busybox is minimal enough that `/dev/console` isn't created by default or something.

I can create an issue for (1) but I would prefer if I didn't have to investigate (2), it's a big rabbit hole and I don't want to dive into it. The file in (2) is a GPL-2.0 code but it's not distributed with Rizin releases and only the CI ever builds it so I think there is no problem with that.

moste00:
The repo that builds the kernel, busybox, and opensbi is here https://github.com/moste00/riscv-minstack, I think we need to transfer it to Rizinorg ? Give the green light and I will invite the transfer from my end.

.github/riscv32-qemu-setup-all.sh
```diff
@@ -0,0 +1,70 @@
...
+
+# Quick and dirty fix for Zydis compilation error, remove when upstream merges a fix and we bump
+# Fix first guard (line 54440) — DecU32
+sed -i'' '54440s/$/ || defined(ZYAN_RISCV32)/' subprojects/zydis/amalgamated-dist/Zydis.c
+# Fix second guard (line 54532) — HexU32
+sed -i'' '54532s/$/ || defined(ZYAN_RISCV32)/' subprojects/zydis/amalgamated-dist/Zydis.c
+# Verify both changes
+sed -n '54440p' subprojects/zydis/amalgamated-dist/Zydis.c
+sed -n '54532p' subprojects/zydis/amalgamated-dist/Zydis.c
+# Build again
```
Rot127: Can you make those ones git patches please.
Otherwise we won't see when it fails.

Rot127 (changes requested):
Please fill out the PR template and write a proper description.
Also rebase to fix the CI

moste00:
@Rot127 @wargio hey, all green now. I assume it can be merged.

patches/fix_zydis_amalgamated_riscv32_build
wargio, line 1: i would suggest to upstream this.

wargio (approved):

patches/fix_zydis_amalgamated_riscv32_build
notxvilka: Yes, definitely.

sys/gen_cpio_script.c
notxvilka, line 1: Not exactly a script...

notxvilka (approved):
LGTM

patches/fix_zydis_amalgamated_riscv32_build
moste00: Yes I would, but it's not straightforward because the amalgamated `zydis.c` doesn't exist in the Zydis upstream, it's created during the build. Searching for the exact text of the line I want doesn't yield anything relevant either, so there is some hunting needed to figure out where to find the `#if` guard.

PR #6123, opened 2026-03-30
Move RzAnalysis as private and force usage of C API

wargio:
**Your checklist for this pull request**
- [ ] I've read the [guidelines for contributing](https://github.com/rizinorg/rizin/blob/dev/CONTRIBUTING.md) to this repository.
- [ ] I made sure to follow the project's [coding style](https://github.com/rizinorg/rizin/blob/dev/DEVELOPERS.md#code-style).
- [ ] I've documented every `RZ_API` function and struct this PR changes.
- [ ] I've added tests that prove my changes are effective (required for changes to `RZ_API`).
- [ ] I've updated the [Rizin book](https://github.com/rizinorg/book) with the relevant information (if needed).
- [ ] I've used AI tools to generate fully or partially these code changes and I'm sure the changes are not copyrighted by somebody else.

**Detailed description**

...

**Test plan**

...

**Closing issues**

...

PR #6130, opened 2026-03-31
Refactor: Generalize ROP core search to Gadget for future JOP/COP support

MrQuantum1915:

**Detailed description**

Full context:  https://github.com/rizinorg/rizin/pull/6108#issuecomment-4147411509

This PR refactors current ROP gadget search by spliting it into core search engine: `gadget.c` and type specific search `gadegt_rop.c` for future implementation of JOP and COP search seamlessly on top of current generic engine.

I have done incremental commits for better review process. So i suggest going commit by commit.

### some premise: 
I analysed current `gadget.c` code and found that most of the current code does a generic gadget search. The only deciding part happens when it checks the terminal instruction: if it is `ret` type or not. Hence this decides whether the gadget should be ROP JOP or COP. Currently as discussed in https://github.com/rizinorg/rizin/pull/6108#issuecomment-4147411509, `/R` and `/Rg` uses different functions to filter for ROP, `/R` allows all instruction types: call, jump and return, while `/Rg` does not, which is the main cause of inconsistency. 

I generalized all the functions, structs, enums etc. And then finally split the code where needed. More details below. I have added tests for new changes, but for renaming part, i think the current test works which are self sufficient to prove the changes.

### Main
1. **Commit 1**: 
	- File name changes and new file add only (NO major CODE CHANGE), so that its easier to review changes in next commit, line by line.
	- No tests needed

2. **Commit 2**: 
	- Generalised (Renamed) API functions (not internal functions), Structs and Enums across files.
	- Current tests works = Self sufficient

3. **Commit 3**: 
	 - Renamed internal functions and some variables.
	 - Current tests works = Self sufficient

4. **Commit 4**: 
	 - Renamed 2 `RzAnalysis` struct fields (`ht_gadget_semantic` and `ht_gadget` (which is for caching; it's currently not implemented, to be done in future))
	 - Renamed Config variables (ALL were General and not ROP specific). 
		 - renamed `crop` to `allow_conditional` (because crop = conditional ROP and also it doesn't mean "crop" literally , its used for allowing conditional instruction type)
	 - As a result migrated project from v21 to v22. 
	 - Also replaced a confusing comment in `gadget.c`
	 - Added migration test in `test_project_migrate.c`.  I didn't add new `.rzdb` for v21 because v20 `rzdb` file already had old config for ROP, so using it is sufficient. Hence in test i first migrate from v20 to v21 and then migrate to 22.
	 -  @wargio likely forgot to move direct access to getter functions in `test_rop.c` and `test_rop_constraint.c` in his RzAnalysis Privatisatoin [PR#6123](https://github.com/rizinorg/rizin/pull/6123). I removed `analysis_private.h` include and forced it to use getter functions. 

5. **Commit 5**
	- Change datatype from ut8 to bool for config variables and getter function too.
	- This is extra bug fix, see below for details.

6. **Commit 6**: 
	- Use `RzPVector` instead of `RzList` for hitlist and Fix gadget.subchains segfault
	- This is also extra bug fix, see below for details.

7. **Commit 7**: 
	- Move gadget.comments config to struct instead of direct access in gadget.c
	- This is also extra fix.

8. **Commit 8**: I forgot to change the datatype of `allow_conditional` to bool in function arguments in `gadget.c`. This should have been done in **Commit 5**. So this is fixup commit.

9. **Commit 9**: The Actual Split, routing based on command by setting gadget type
	- As the main distinguishing factor is terminal instruction type for ROP, JOP and COP (primarily), I defined new enum type `RzGadgetType` and declared in `RzgadgetSearchContext` struct. In `gadget.c` we are calling respective filter function based on current `gadget_type` (ROP/JOP/COP) that will be defined in respective file (e.g. `gadget_rop.c`) exposed by new internal header file `gadget_internal.h` . 
	- **NOTE:** Any future ROP specific implementation will be done in `gadget_rop.c`. `gadget.c` remains as generic engine. Same for JOP , JOP might need some extra code logic which will be done in its file.
	- i am setting the Gadget Type based on command in `cmd_search.c` handler fxns.
	- Current cmd handler for : `/R`, `/R/`, `/Rk`, `/Rg`, `/Rl`  , will be used by JOP and COP too, because they are generic. Hence I am setting Gadget Type based on `argv[0][1]` in handler functions. `/Rl` is not ROP specific so changed the handler's name.
	- `/Rs` is ROP specific (search based on stack changes), so it has special handler function that will not be used for JOP and COP.

	- **NOTE**: I have just copy pasted switch block into `gadget_rop.c`, because removing call and jump type fomr switch block would have inherently solved inconsistency issue, SO to keep this PR clean (instead of writing new tests and all here). We dont need two filter functions in it they will be unified. I will be fixing the main issue of inconsistency in this [PR: 6108](https://github.com/rizinorg/rizin/pull/6108) . Also the switch block is currently wrong and totally broken and current output is flawed, because `aop->type` is always a bitmask, but its not handled correctly in `switch()`, so currently `allow_conditional` is broken. I will be fixing that in that PR too.

10. **Commit 10**: Removed `rz_` prefix from static functions (that are in scope of this PR). 

### Extra bug fixes
- **Commit 4**:
	-  Fixed a real bug in config: the config file used `rop.subchains` while `cmd_search_gadget.c`  used `rop.subchain` (**NOTICE** the missing 's'), so this would have failed earlier. I also changed the variable name subchain in `RzGadgetSearchContext` struct.
- **Commit 5**
	-  In `RzGadgetSearchContext`, `cache` is declared as bool but when getting its value in `rz_core_gadget_search_context_new` function we use integer getter fxn :`rz_config_get_i` instead of boolean, so fixed that.
	-  I also changed the variable type from `ut8` to `bool` for `subchain` and `allow_conditional` variables in  `RzGadgetSearchContext` because they are originally used as boolean type.
- **Commit 6**
	-  while testing the bug in commit 4, i found another bug : setting `gadget.subchains=true` causes a segfault. Also this path was not tested in regression tests. 
	- The cause was **direct** manipulation of linked list's `head` pointer `hitlist` for producing subchains instead of working on its copy. 
	- After discussion, migrated from `RzList` to `RzPVector` for `hitlist` for performance.
	- using `rz_pvector_tail(hitlist)` causes warning (not tolerated by CI/CD), because tail function doesn't accept `const` vector. Hence i used `rz_pvector_at()` fxn to get last value.
	- for producing subchains i am using custom `for()` loop instead of `rz_pvector_foreach()` because it always starts from `0th` index, but we need to start from `ith` index. I also changed `print_gadget_hitlist` fxn to accept a starting index. So it prints gadget from instruction at index `start_idx` to `end`. This perfectly prints all sub-sequence of gadget.
	- **NOTE** that as per original intent, i am printing subchains from X to 2 instructions length.
	- added a test for subchains in `cmd_rop.c` 

- **Commit 7**
	-  Brought `gadget.comments` check in struct instead of direct fetching in `gadget.c` similar to how other configs are accessed. It is faster now, because we dont have to perform lookup in config hash table for EVERY gadget instruction!
	- There were no test for comments on, So i added one in `cmd_rop`.

- **Commit 9**
	-  original dev accidentally passed `argv[1]` instead of `input` in handler`rz_cmd_info_gadget_handler`. Fixed that.
### Next PRs
- [Fix /R and /Rg inconsistency](https://github.com/rizinorg/rizin/pull/6108), will remove COP and JOP types and unify both filter fxns, making `/R` `purely ROP specific. ALso fix the flawed switch case block.
- Implement /J and /C for JOP and COP search as discussed in [HERE#6108, comment](https://github.com/rizinorg/rizin/pull/6108#issuecomment-4148585182)
-  BUG: In `rz_gadget_info_add_dependency`, in case of VAR_WRITE dependency, code uses direct original `reg_info` instead of duplicate, needs more inspection though.  
...

**Test plan**

All old tests for cmd_rop passes. Which are sufficient to proove rename  and refactor changes
Running `cmd_rop` regression test passes both old and new tests:
image

All unit tests passed including new addition in project migrate test for v21 to v22
image

...

**Closing issues**

...


librz/core/gadget.c
```diff
@@ -1,26 +1,15 @@
...
-		return true;
-	default:
-		return false;
-	}
-}
+#include <rz_gadget.h>
+#include "gadget_internal.h"
 
-static bool is_ret_gadget(const RzCore *core, const RzCoreAsmHit *hit, const ut8 crop) {
+static bool rz_gadget_is_valid_terminator(const RzGadgetType gadget_type, const RzCore *core, const RzCoreAsmHit *hit, const bool allow_conditional) {
```
wargio: these are static & non-public. they cannot have the `rz_` prefix

wargio (commented):
it's a good refactoring. please rebase

librz/core/gadget.c
MrQuantum1915: i simply followed the original pattern
because in old code too, there were many instances of `rz_` prefix being used for static functions 
it was inconsistent, like somewhere `rz_` was used while somewhere it wasn't

should i change all of them then??


Eg:
image

MrQuantum1915: @wargio 

also i search for this pattern, there seems to be many instances of such usage in various files in `librz`

image

are this intentional for a reason?
if not, should I change this too in a new PR then???

wargio: these also should be changed without `rz_`.

Normally when i see them, i will rename them, but yes, you can open a PR for these too.

MrQuantum1915:
> it's a good refactoring. please rebase

- i removed `rz_` prefix from static functions
- rebase done.
- CI all green

librz/include/rz_gadget.h
```diff
@@ -0,0 +1,222 @@
...
+	RZ_GADGET_DETAIL_SEARCH_SIZE = 1 << 1, ///< Search gadgets by gadget sizes.
+	// RZ_GADGET_DETAIL_SEARCH_WRITE = 1 << 2, ///< Search gadgets by written registers.
+	// RZ_GADGET_DETAIL_SEARCH_READ = 1 << 3, ///< Search gadgets by read registers.
+} RzGadgetDetailSearchMask;
+
+/**
+ * \brief Filter conditions while searching gadgets by stack changes.
+ */
+typedef enum {
+	RZ_GADGET_DETAIL_CMP_EQ = 1, // ==
```
notxvilka: Please use Doxygen syntax for those

notxvilka (changes requested):
Looks like a great refactoring! There is a small nitpick though

MrQuantum1915:
> Looks like a great refactoring! There is a small nitpick though

fixed that

librz/core/gadget_internal.h
```diff
@@ -0,0 +1,13 @@
...
+// SPDX-License-Identifier: LGPL-3.0-only
+
+#ifndef RZ_GADGET_INTERNAL_H
+#define RZ_GADGET_INTERNAL_H
+
+#include <rz_analysis.h>
+
+// ROP (Return-Oriented Programming)
+bool rz_gadget_rop_is_end_gadget(const RzAnalysisOp *aop, const bool allow_conditional);
+bool rz_gadget_rop_is_valid_terminator(const RzAnalysisOp *aop, const bool allow_conditional);
```
Rot127: RZ_IPI

Rot127 (changes requested):
Looks like a good refactor indeed!
We should consider merging all this renaming stuff before the functional changes come into play.

librz/core/gadget_internal.h
wargio: these could be static (and remove the rz_ prefix)

MrQuantum1915: > these could be static (and remove the rz_ prefix)

i dont think we can make them static, because `gadget.c` file needs that functions as it dispatches function calls based on Gadget type that is to be searched for

i will use RZ_IPI to mark them internal...

librz/core/gadget.c
```diff
@@ -111,34 +85,34 @@ static bool rz_rop_process_asm_op(const RzCore *core, const RzCoreAsmHit *hit, R
...
 	}
 
-	if (!rz_rop_process_asm_op(core, hit, asmop, &aop, size, asmop_str, asmop_hex_str)) {
+	if (!gadget_process_asm_op(core, hit, asmop, &aop, size, asmop_str, asmop_hex_str)) {
 		rz_asm_op_free(asmop);
 		return false;
 	}
-	const ut64 addr_last = ((RzCoreAsmHit *)rz_list_last_val(hitlist))->addr;
-	if (addr_last != hit->addr) {
+	const RzCoreAsmHit *last_hit = rz_pvector_at(hitlist, rz_pvector_len(hitlist) - 1);
```
wargio: add a check if len is `0`

wargio (commented):
one last change, and we can merge

notxvilka (approved):
Looks great!

librz/core/gadget.c
MrQuantum1915: done

notxvilka:
Looks like many leaks https://github.com/rizinorg/rizin/actions/runs/24354188848/job/71116621362?pr=6130#step:15:2178

MrQuantum1915:
> Looks like many leaks https://github.com/rizinorg/rizin/actions/runs/24354188848/job/71116621362?pr=6130#step:15:2178

i am skeptical about leaks being related to my last commit, before that the CI was all green
the last commit can't cause this leaks

i rebased on dev before pushing last commit though
might be because of some changes in leak detection in previous PRs,

will check that

notxvilka:
> > Looks like many leaks https://github.com/rizinorg/rizin/actions/runs/24354188848/job/71116621362?pr=6130#step:15:2178
> 
> i am skeptical about leaks being related to my last commit, before that the CI was all green the last commit can't cause this leaks
> 
> i rebased on dev before pushing last commit though might be because of some changes in leak detection in previous PRs,
> 
> will check that

Yeah, I think we can merge this one as is, and address leaks in separate PRs. cc @Rot127

wargio:
i'm ok to merge like this.

wargio (approved):

PR #6156, opened 2026-04-04
Add detailed help for plF

### LINKED ISSUE #6146: Detailed help for the `plF` command
Commit https://github.com/rizinorg/rizin/commit/0b1905f507cf592fb0391b792d371d76c885bb65 implements the command itself, but there no "legend" of each symbol/notation meaning. 

Extend the command description/details/examples with the information about meaning of each used symbol and maybe few examples. See `pf?` and `pf??` as an example

Ehab-24:
Also updated a few symbols to improve UX.

 

**Detailed description**

This PR:
1. Adds detailed help for the `plF` command added in #5936.
2. Changes notation of the following ops:
    
    | Op | Before | Now |
    | --- | --- | --- |
    | is_finite | x < ∞ | x ≢ ∞ |
    | frequal | (r1 ꜰ≡ r2) | (r1 ≡ r2) |
    | forder | (x < y) | (x ≷ y) |
    | fmod | (r x ꜰ% y) | (r x % y) |
    | fmad | (r x ᴍᴀ y z) | (r x * y + z) |
    | frootn | (r f ⁿ√ n) | (r n ⁿ√ f) |
    | nop | ∅ | ɴᴏᴘ |
    | logand | ∧ | & |
    | logor | ∨ | \| |
    | booland | & | ∧ |
    | boolor | \| | ∨ |
    | boolxor | ^ | ⊻ |
    | load | (n ʟₘ a) | (ʟᴅₘ n a) |
    | store | (ꜱ₀ a v) | (ꜱᴛₘ v a) |
    | storew | (ŝ₀ a v) | (ꜱᴛₘ v a) |

...

**Test plan**

image

image

...

**Closing issues**

closes #6146 

...


librz/il/il_export_string_unicode.c
```diff
@@ -57,18 +57,17 @@ static const char *const subscript_digits[10] = { "₀", "₁", "₂", "₃", "
 #define UCD_FSQRT        "²√"
 #define UCD_FRSQRT       "¹/√"
 #define UCD_FROUND       "⭂"
-#define UCD_FREQUAL      "ꜰ≡"
+#define UCD_FREQUAL      "≡"
 #define UCD_FSUCC        "⌊"
 #define UCD_FPRED        "⌋"
-#define UCD_FORDER       "<"
+#define UCD_FORDER       "⋚"
```
notxvilka: I understand the intention but this sign is so unwieldy...
@Rot127 @wargio WDYT?

notxvilka (changes requested):
When you change the signs, please list such changes in the PR desription.

librz/core/cmd_descs/cmd_print.yaml
```diff
@@ -855,9 +855,252 @@ commands:
         cname: print_function_rzil
         args: []
       - name: plF
-        summary: Print enriched RzIL of the function
+        summary: Print enriched RzIL of the function in unicode characters.
```
notxvilka: I think the description from another PR fits better, actually. And it's not "enriched RzIL" yet, just a normal one.

notxvilka (changes requested):

librz/il/il_export_string_unicode.c
Ehab-24: `≷` could be another option.

notxvilka: Yeah, this one looks better.

wargio (approved):

notxvilka (approved):
When I mentioned the signs change in the PR description I meant something like "Fmod: was `ꜰ%` became `%`"

Ehab-24:
> When I mentioned the signs change in the PR description I meant something like "Fmod: was `ꜰ%` became `%`"

ahh that makes much more sense

Rot127 (changes requested):
Can I ask you to change the `nop` character to simply `nop` itself please?

You used the character which is usually used for an empty set.
And this hasn't really any association with a "no op"

librz/core/cmd_descs/cmd_print.yaml
```diff
@@ -855,9 +855,255 @@ commands:
...
+                comment: "Unsigned modulo of x and y."
+              - text: "Smod"
+                arg_str: " (x %⁺ y)"
+                comment: "Signed modulo of x and y."
+              - text: "Logand"
+                arg_str: " (x ∧ y)"
+                comment: "Bitwise AND between x and y."
+              - text: "Logor"
+                arg_str: " (x ∨ y)"
+                comment: "Bitwise OR between x and y."
```
Rot127: Since they are bit wise I think `|` and `&` are better choices here. And use the `∧` and `∨` for the non-bit versions below.

librz/core/cmd_descs/cmd_print.yaml
```diff
@@ -855,9 +855,255 @@ commands:
...
+              - text: "Logxor"
+                arg_str: " (x ⊕ y)"
+                comment: "Bitwise XOR between x and y."
```
Rot127, line 955: But for the bit wise xor we maybe should find another symbol. Using ^ is better here, for my feeling. But it is too similar to the boolean AND.
So maybe use ⊻ for bitwise XOR, and ⊕ for boolean XOR?

https://en.wikipedia.org/wiki/List_of_logic_symbols

librz/core/cmd_descs/cmd_print.yaml
```diff
@@ -855,9 +855,255 @@ commands:
...
+              - text: "Float"
+                arg_str: " n.fᵈₙ"
+                comment: "Bitvector literal n interpreted as a float of size n \
+                          with optional superscript d showing a decimal representation."
+              - text: "Fbits"
+                arg_str: " ꜰʙ x"
+                comment: "Bitvector representation of float x."
+              - text: "Fneg"
+                arg_str: " −x"
+                comment: "Floating-point negation of x."
```
Rot127, line 1002:
```suggestion
                arg_str: " −n.fᵈₙ"
                comment: "Floating-point negation of n.fᵈₙ."
```

Correct?

librz/core/cmd_descs/cmd_print.yaml
```diff
@@ -855,9 +855,255 @@ commands:
...
+            entries:
+              - text: "Load"
+                arg_str: " (n ʟₘ a)"
+                comment: "Load n bits from address a of memory index m."
+              - text: "Store"
+                arg_str: " (ꜱ₀ a v)"
+                comment: "Store v to address a of memory index m."
+              - text: "Store word"
+                arg_str: " (ŝ₀ a v)"
+                comment: "Store v to address a of memory index m."
```
Rot127, line 1103: Here I would keep it simpler like `store(a, v)`, `load(a, v)` ...

In the end it must be more readable to us. And the symbols a a little complicated, at least to me.

Rot127 (changes requested):

wargio (approved):

librz/core/cmd_descs/cmd_print.yaml
wargio: mathematically `^v` are correct. but i can agree to keep the `|&`

librz/core/cmd_descs/cmd_print.yaml
wargio: that is a good suggestion. i agree

librz/core/cmd_descs/cmd_print.yaml
Ehab-24: Here `x` may not be a float bitvector, it may be a variable name or a cast operation for example.

librz/core/cmd_descs/cmd_print.yaml
Ehab-24: Yes this is better especially since it would align with non-unicode RzIL too.

librz/core/cmd_descs/cmd_print.yaml
Ehab-24: I agree as well, this makes more w.r.t to the suggested changes above.

Ehab-24: Actually I think it is better to use ⊻ for boolxor and keep ⊕ as logxor for better cohesiveness | [mattermost thread](https://im.rizin.re/rizinorg/pl/eqbah7pcu3dyjmmsi6ggx16p8y).

notxvilka:
@Ehab-24 you need to rebase it to fix the CI

notxvilka (approved):

Rot127 (approved):

PR #6160, opened 2026-04-05
core: fix ps command truncating strings longer than 256 bytes

### LINKED ISSUE #4857: `ps` doesn't print long strings by default, because it depends on `core->block`.
### Work environment

| Architecture/bits of the file (mandatory)            | NA
| `rizin -v` full output, **not truncated** (mandatory)  | rizin 0.8.0-git @ linux-x86-64, [ee74d37](https://github.com/rizinorg/rizin/commit/ee74d37833d2995c3d91ffd6d57535340d4512b0)

### Expected behavior

`ps` (and `psW`, `psM` etc.) print a full string in a file by default.

### Actual behavior

It prints only a string of `sizeof(core->block)` by default. Which is apparently by default `256` bytes.

```
rizin -qc "psW" test/bins/cmd/search/string_encodings/Georgian-Lipsum.utf32le
ლორემ იფსუმ დოლორ სით ამეთ, ილლუმ ელაბორარეთ ად ჰას, ილლუმ იისყუ
```

But should be way longer.

(Binary from PR: https://github.com/rizinorg/rizin-testbins/pull/167)

The implementation of `rz_str_stringify_raw_buffer()` to detect the strings, should use `RzBuffer`. Not a raw `char *` buffer. This way it can get the `RzBuffer` of the current `RzIO`.

Shouldn't be too complicated to refactor. At least not for the commands.

Related: https://github.com/rizinorg/rizin/issues/2695

cheese-cakee:
- [x] I have NOT used AI tools to generate these code changes.

**Detailed description**

This PR fixes the `ps` command which was truncating strings longer than 256 bytes. The issue occurred because the URL-encode length was not properly clamped to the actual block size, causing longer strings to be cut off.

The fix properly clamps the URL-encode length to the blocksize, ensuring strings of any length within the block are correctly displayed.

**Test plan**

1. Build Rizin with the changes
2. Test `ps` command with strings longer than 256 bytes
3. Verify strings are no longer truncated
4. Run the test suite: `rizin -q -c < test/db/cmd/cmd_ps`

**Closing issues**

closes #4857


librz/core/cmd/cmd_print.c
```diff
@@ -1830,31 +1832,56 @@ static void core_print_raw_buffer(RzStrStringifyOpt *opt) {
...
+	ut64 len = 0;
+	RzBuffer *buf = rz_buf_new_with_pointers(core->block, core->blocksize, false);
+
+	if (!buf) {
+		return RZ_CMD_STATUS_ERROR;
+	}
+
+	if (offset < core->blocksize) {
+		len = core->blocksize - offset;
+	}
```
wargio, line 1759: i'm not sure this is correct. the idea is to not use `block/blocksize` and instead read from the io layer.

wargio:
i think the remaining changes are ok. @Rot127

librz/core/cmd/cmd_print.c
cheese-cakee: please check now

wargio: i don't see changes

cheese-cakee:
Hello @wargio. After your requested change, I updated the code, but CI started failing so I reverted, then re-applied the change carefully and re-verified.

Current state now:
  - The requested change is implemented.
  - From my side and current CI results, checks are passing except one:
    `linux-gcc-tests-lsan`
    
I’m not fully sure whether this remaining failure is due to my code or CI/environment. Could you please advise whether you want me to change anything else on my side? thankyou

cheese-cakee:
please review when possible @Rot127

cheese-cakee:
all checks passed pls review when possible @wargio

test/db/cmd/cmd_ps
```diff
@@ -197,3 +197,15 @@ EXPECT=<<EOF
...
+
+NAME=ps long string > 256 chars (#4857)
+FILE==
+CMDS=<<EOF
+o malloc://1024
+wx 41414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141
+ps
+EOF
+EXPECT=<<EOF
+AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA
```
Rot127: Those are exactly 256 if my editor counts correctly, not more.
Please make more of them and fix the block size to `b 256`. So we are sure the test checks what it is supposed to do.

Rot127 (changes requested):

librz/core/cmd/cmd_print.c
```diff
@@ -1744,31 +1746,66 @@ static void core_print_raw_buffer(RzStrStringifyOpt *opt) {
...
+			len = io_size - str_off;
+		}
+		len = RZ_MIN(len, (ut64)0x100000);
+	} else {
+		buf = rz_buf_new_with_pointers(core->block, core->blocksize, false);
+		if (!buf) {
+			return RZ_CMD_STATUS_ERROR;
+		}
+	}
+	len = RZ_MIN(len, (ut64)UT32_MAX);
```
Rot127, line 1782: Why this step?

librz/core/cmd/cmd_print.c
```diff
@@ -1744,31 +1746,66 @@ static void core_print_raw_buffer(RzStrStringifyOpt *opt) {
...
+		str_off = core->offset + offset;
+		buf = rz_buf_new_with_io(&core->print->iob);
+		if (!buf) {
+			return RZ_CMD_STATUS_ERROR;
+		}
+		const ut64 io_size = rz_io_size(core->io);
+		if (io_size != UT64_MAX && io_size > str_off) {
+			len = io_size - str_off;
+		}
+		len = RZ_MIN(len, (ut64)0x100000);
```
Rot127: Also here, why this specific hard coded value?

Rot127 (changes requested):

librz/core/cmd/cmd_print.c
cheese-cakee: UT32_MAX is needed because len is computed as ut64, but RzStrStringifyOpt.length is ut32; this clamp guarantees  safe narrowing before passing the value downstream

librz/core/cmd/cmd_print.c
cheese-cakee: 0x100000 is currently a guardrail only for the IO-layer path. After switching to len = io_size - str_off, scanning
 could otherwise run very far on files without a terminator. I scoped the cap to IO only to keep block-bounded
 behavior unchanged

Rot127: I see. Please define it with a macro and give it a proper name. Also warn if this is reached.

librz/core/cmd/cmd_print.c
Rot127: In what case?

cheese-cakee: len is ut64, but RzStrStringifyOpt.length is ut32. This can matter when len is larger than UT32_MAX before assigning to opt.lenght

librz/core/cmd/cmd_print.c
cheese-cakee: I replaced the hardcoded 0x100000 with CMD_PRINT_IO_STRING_SCAN_MAX and added a RZ_LOG_WARN

Rot127 (changes requested):
Sorry for the huge delays.
Can you please rebase.

PR #6166, opened 2026-04-06
table: fix shared query filtering

### LINKED ISSUE #5993: Filtering `izz` tables (possibly other tables as well) throws errors.
### Work environment

| Architecture/bits of the file (mandatory)            | All
| `rizin -v` full output, **not truncated** (mandatory)  | rizin dev

### Expected behavior

Filtering `izz` table output should not throw errors.
### Actual behavior

```
> rizin /bin/ls

[0x000037f0]> izz:string/minlen/8:length/sort/rev:*/page/0/15:csv
rz_num_calc error: ( ')' expected) in ((i9<iv}2)
rz_num_calc error: ( ')' expected) in ((iG<iTZÞ6)
rz_num_calc error: ( ')' expected) in ((º<iTZD4)
rz_num_calc error: ( ')' expected) in ((iG<iTZî3)
rz_num_calc error: ( ')' expected) in ((iG<iTZ)
rz_num_calc error: ( ')' expected) in ((iF<inY³)
rz_num_calc error: ( ')' expected) in ((E|h)
paddr,vaddr,len,size,section,type,string
0x00022f43,----------,14,15,.shstrtab,ascii,.gnu_debugdata
0x00022f34,----------,14,15,.shstrtab,ascii,.gnu_debuglink
0x00022f1e,----------,21,22,.shstrtab,ascii,.gnu.build.attributes
...
```

As above

IndAlok:

**Detailed description**

This PR fixes a table filtering issue that was reported while testing table query syntax on binaries. (as described in https://github.com/rizinorg/book/pull/159, i used `curl-example-dbg` from testbins for reproducibility and testing)

The issue is the shared table query/filter engine behind commands like `izz`, `is` and other table cmds. This caused cmds such as:

```
izz:string/minlen/8:length/sort/rev:*/page/0/15:csv
```

to emit `rz_num_calc` errors on ordinary str contents, and cmds such as:

```
is:name/uniq:addr/gt/0x1000:name/str/init:addr/sort:json
```

to behave incorrectly when the query used common names like `addr` or `length` that do not always exactly match the underlying table headers.

The changes do three things:

1. They make num parsing lazy in `rz_table_query()` / `table_filter()`.
   Before this change, row values were eagerly sent to `rz_num_math()` even for str-only operations such as `str`, `minlen`, and `maxlen`. That is what produced the `rz_num_calc` errors.
   After this change, num parsing only happens for operators that actually require num semantics.

2. They make string eq/ineq stay str based on str columns.
   This avoids accidental numeric treatment of values such as `"16"`, `"010"`, and `"0x10"` when the column itself is txt.

3. They add query time column resolution for stable aliases such as `addr -> vaddr/address/paddr` and `length -> len`, while preserving the existing `cols` behavior expected by older query chains.

Tests were also added:
- new unit coverage was added for the shared table query regressions
- db tests were added for the `izz` string/minlen case and the `is` addr-alias case
- the addr-alias db test was made deterministic by sorting by `name` after filtering by `addr`, so it doesn't depend on platform specific ordering of eq-addr rows

**Test plan**

Before:
before1
before2

After:
after1
after2

**Closing issues**

closes #5993

### DIFF
--- a/librz/util/table.c
+++ b/librz/util/table.c
@@ -125,6 +125,26 @@ static bool table_index_of_column(RzTable *t, const char *name, size_t *index) {
 	return false;
 }
 
+static bool table_column_is_numeric(const RzTableColumn *col) {
+	// query aliases like addr/length are meant for num cols only.
+	return col && (col->type == RZ_TABLE_COLUMN_TYPE_NUMBER || col->type == RZ_TABLE_COLUMN_TYPE_BOOL);
+}
+
+static bool table_query_index_of_numeric_column(RzTable *t, const char *name, size_t *index) {
+	size_t col = 0;
+	if (!table_index_of_column(t, name, &col)) {
+		return false;
+	}
+	RzTableColumn *column = rz_vector_index_ptr(t->cols, col);
+	if (!table_column_is_numeric(column)) {
+		return false;
+	}
+	if (index) {
+		*index = col;
+	}
+	return true;
+}
+
 static bool table_query_index_of_column(RzTable *t, const char *name, size_t *index) {
 	static const char *const addr_aliases[] = {
 		"vaddr",
@@ -142,14 +162,14 @@ static bool table_query_index_of_column(RzTable *t, const char *name, size_t *in
 	// Keep stable query names working across tables with slightly different headers.
 	if (RZ_STR_EQ(name, "addr")) {
 		for (size_t i = 0; addr_aliases[i]; i++) {
-			if (table_index_of_column(t, addr_aliases[i], index)) {
+			if (table_query_index_of_numeric_column(t, addr_aliases[i], index)) {
 				return true;
 			}
 		}
 	}
 	if (RZ_STR_EQ(name, "length")) {
 		for (size_t i = 0; length_aliases[i]; i++) {
-			if (table_index_of_column(t, length_aliases[i], index)) {
+			if (table_query_index_of_numeric_column(t, length_aliases[i], index)) {
 				return true;
 			}
 		}
--- a/librz/util/table.c
+++ b/librz/util/table.c
@@ -125,6 +125,174 @@ static bool table_index_of_column(RzTable *t, const char *name, size_t *index) {
 	return false;
 }
 
+static bool table_query_index_of_column(RzTable *t, const char *name, size_t *index) {
+	static const char *const addr_aliases[] = {
+		"vaddr",
+		"address",
+		"paddr",
+		NULL
+	};
+	static const char *const length_aliases[] = {
+		"len",
+		NULL
+	};
+	if (table_index_of_column(t, name, index)) {
+		return true;
+	}
+	// Keep stable query names working across tables with slightly different headers.
+	if (RZ_STR_EQ(name, "addr")) {
+		for (size_t i = 0; addr_aliases[i]; i++) {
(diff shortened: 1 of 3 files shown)


wargio (changes requested):
split this into 2 PRs. one for windows console and one for the table query

IndAlok:
> split this into 2 PRs. one for windows console and one for the table query

> Detailed description (PR)

> This PR fixes two separate problems (however testing one lead to identification of another, hence could be considered ig)

I thought in this case, could be considered. Np, will split.

wargio:
> > split this into 2 PRs. one for windows console and one for the table query
> 
> > Detailed description (PR)
> 
> > This PR fixes two separate problems (however testing one lead to identification of another, hence could be considered ig)
> 
> I thought in this case, could be considered. Np, will split.

i think you should fix first only the table, then we can fix the terminal

librz/util/table.c
```diff
@@ -125,6 +125,174 @@ static bool table_index_of_column(RzTable *t, const char *name, size_t *index) {
...
+		return true;
+	}
+	// Keep stable query names working across tables with slightly different headers.
+	if (RZ_STR_EQ(name, "addr")) {
+		for (size_t i = 0; addr_aliases[i]; i++) {
+			if (table_index_of_column(t, addr_aliases[i], index)) {
+				return true;
+			}
+		}
+	}
```
wargio, line 169: i think you should check the type. usually when you define a row you use `xxssXsu` etc.. to define if is a number or a string.

wargio (commented):
it's a quite nice improvement.

Please test the book queries: https://book.rizin.re/src/tools/rz-bin/tables.html

librz/util/table.c
IndAlok: yep, now `table_column_is_numeric()` treats `NUMBER` and `BOOL` cols as num comaptible, `table_query_index_of_numeric_column()` resolves header name first & rejects it unless the target col is num, & in `table_query_index_of_column()`, the `addr` and `length` alias fallback goes through the num only helper...
have used ex2 and 3 from book for test as well

IndAlok:
> it's a quite nice improvement.
> 
> Please test the book queries: https://book.rizin.re/src/tools/rz-bin/tables.html

ex 2 and 3 seem to be successfull (https://github.com/rizinorg/rizin/pull/6166#discussion_r3043048591), however, for ex1, this is the quote from book:
`Note: Table specifiers are applied from left to right. Output format specifiers must be specified at the end.`

and this is the ex1 provided,
`aflt:addr/cols/name/nbbs:size/gt/32:nbbs/gt/1:nbbs/lt/10:nbbs/sort/rev:fancy`

`addr/cols/name/nbbs` is handled by `table_query_select_columns()`, which turns that into a selected col list and then calls `rz_table_columns_select()`, and table has 3 cols left only: `addr`, `name` & `nbbs`. then `size/gt/32` is there, but `size` no longer exists in the table, because `cols` already removed it, and hence  we get `ERROR: table: Invalid column (size)`.

however, `aflt:size/gt/32:nbbs/gt/1:nbbs/lt/10:addr/cols/name/nbbs:nbbs/sort/rev:fancy` would work as `size` & `nbbs` is still present when it is filtered and sorted, and after that we reduce the `cols`...

wargio (approved):

notxvilka (approved):

PR #6169, opened 2026-04-07
cons: fix Windows console VT mode save and restore

IndAlok:
Title:
`cons: fix Windows console VT mode save and restore`

**Detailed description**

this pr fixes win console mode handling during rizin startup and teardown. (related to https://github.com/rizinorg/rizin/pull/6166)

while testing the table query changes in win in `meson devenv`, rizin would complete the requested cmd but still leave the terminal in a bad state and print:

```text
SetConsoleMode: (0x57) The parameter is incorrect.
```

the root problem was in the shared win console lifecycle, not in any specific cmd.

this pr keeps the fix limited to the console code paths that actually own and mutate win console state:

- `librz/cons/cons.c`
- `librz/cons/input.c`
- `librz/core/cconfig.c`

the changes are:

1. save the original console state before VT probing mutates it.
   previously, the saved console mode could already reflect side effects from early VT detection. that makes later restore unreliable because the stored state is no longer the real original host state.

2. restore only the console mode bits Rizin actually owns.
   previously, the restore path replayed the full saved mode word. in win, this is fragile, esp. when running under nested shells or `meson devenv`, because it may overwrite flags owned by the hosting console instead of only undoing rizin’s own changes.

3. fix VT input enabling in `scr_vtmode()`.
   the old code used `mode & ENABLE_VIRTUAL_TERMINAL_INPUT`, which effectively stripped other input flags instead of enabling VT input. this pr changes that to the flag preserving behavior.

4. avoid temporary input mode writes on non console stdin.
   some temporary console mode transitions assumed stdin was always a valid console handle. this pr guards those paths so batch or redirected runs do not try to call `SetConsoleMode()` on non-console input.

the idea behind the patch is to fix the shared console state contract directly.

**Test plan**

verified on win after building

**Closing issues**

N/A

### DIFF
--- a/librz/cons/cons.c
+++ b/librz/cons/cons.c
@@ -21,6 +21,17 @@ static RzConsContext rz_cons_context_default = { { { { 0 } } } };
 static RzCons rz_cons_instance = { 0 };
 #define I rz_cons_instance
 
+#if __WINDOWS__
+static HANDLE saved_input_handle;
+static HANDLE saved_output_handle;
+static bool saved_input_console;
+static bool saved_output_console;
+
+// restore only the console flags rizin owns instead of replaying the whole host mode word
+#define RZ_CONS_OUTPUT_MODE_MASK (ENABLE_PROCESSED_OUTPUT | ENABLE_WRAP_AT_EOL_OUTPUT | ENABLE_VIRTUAL_TERMINAL_PROCESSING)
+#define RZ_CONS_INPUT_MODE_MASK  (ENABLE_ECHO_INPUT | ENABLE_LINE_INPUT | ENABLE_MOUSE_INPUT | ENABLE_QUICK_EDIT_MODE | ENABLE_EXTENDED_FLAGS | ENABLE_VIRTUAL_TERMINAL_INPUT)
+#endif
+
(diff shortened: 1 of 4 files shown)


wargio (commented):
this is ok, but when we refactor the tui code, we need a way to handle all these globals.

wargio (approved):

librz/cons/cons.c
```diff
@@ -564,34 +577,46 @@ static void set_console_codepage_to_utf8(void) {
 }
 
 static void save_console_state(void) {
```
wargio, line 574: this could return a "state" structure which is required to restore the terminal

IndAlok: but in the current lifecycle, `rz_cons_new()` saves the state and `rz_cons_free()` restores it later, so the snapshot should persist across the lifetime of the console object anyway, right? returning a struct from `save_console_state()` would still require storing that struct somewhere, maybe console instance itself?

wargio: this is why i said, this needs to be rewritten when we will do a refactoring of the TUI code.
inside the cons struct like you did is perfect

Rot127 (approved):

PR #6178, opened 2026-04-07
Rewrite of existing ESIL-based analysis passes with RzIL

thestr4ng3r:

**Detailed description**

This aims to provide direct rewrites of analysis passes without changing the actual algorithms, but only using RzIL instead of ESIL. The benefits are that dependencies on ESIL can be removed and the resulting code will be much cleaner and easier to extend and adjust due to the nature of both ILs.

As long as after a certain point where instructions are disassembled, there is still ESIL-based analysis present, both RzIL and ESIL have to be lifted, which will have a performance penalty. So it can make sense to not merge this before lifting ESIL can be disabled entirely at the points where RzIL has been added.

Analysis passes included (checked means conversion finished):
- [x] arm `mov lr, pc` check
- [ ] Variable extraction

**Test plan**

Existing tests should all pass without out changes, or only with changes that have been validated to make sense, e.g. when the RzIL lifter covers more instructions than ESIL.

### DIFF
--- a/librz/arch/var.c
+++ b/librz/arch/var.c
@@ -1,4 +1,4 @@
-// SPDX-FileCopyrightText: 2022 Florian Märkl <info@florianmaerkl.de>
+// SPDX-FileCopyrightText: 2022-2026 Florian Märkl <info@florianmaerkl.de>
 // SPDX-FileCopyrightText: 2010-2020 pancake <pancake@nopcode.org>
 // SPDX-FileCopyrightText: 2010-2020 oddcoder <ahmedsoliman@oddcoder.com>
 // SPDX-License-Identifier: LGPL-3.0-only
@@ -9,6 +9,13 @@
 #include <rz_cons.h>
 
 #define ACCESS_CMP(x, y) ((st64)((ut64)(x) - (ut64)((RzAnalysisVarAccess *)y)->offset))
+
+#if 0
+#define P(...) eprintf(__VA_ARGS__)
+#define PRINTING
+#else
+#define P(...)
+#endif
 
 /**
  * \brief Compare two RzAnalysisVarAccess objects.
@@ -350,6 +357,14 @@ RZ_API RZ_BORROW RzAnalysisVar *rz_analysis_function_set_var(
 	RZ_NONNULL const char *name) {
 	rz_return_val_if_fail(fcn && name, NULL);
 	RzAnalysisVar *var = rz_analysis_function_get_var_byname(fcn, name);
+	if (stor->type == RZ_ANALYSIS_VAR_STORAGE_STACK) {
+		P("SET VAR @ stack + %"PFMT64d" with name %s\n", stor->stack_off, name);
+#ifdef PRINTING
+		rz_sys_backtrace();
+#endif
+	} else {
+		P("SET OTHER VAR\n");
+	}
 	if (var) {
 		if (!rz_analysis_var_storage_equals(&var->storage, stor)) {
 			// var name already exists at a different kind+delta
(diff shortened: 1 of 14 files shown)


librz/arch/var.c
```diff
@@ -350,6 +357,14 @@ RZ_API RZ_BORROW RzAnalysisVar *rz_analysis_function_set_var(
 	RZ_NONNULL const char *name) {
 	rz_return_val_if_fail(fcn && name, NULL);
 	RzAnalysisVar *var = rz_analysis_function_get_var_byname(fcn, name);
+	if (stor->type == RZ_ANALYSIS_VAR_STORAGE_STACK) {
+		P("SET VAR @ stack + %"PFMT64d" with name %s\n", stor->stack_off, name);
+#ifdef PRINTING
+		rz_sys_backtrace();
+#endif
```
Rot127, line 364: Can you please move this logic to the `rz_log` header or similar. Maybe even to a separated PR.
It would be more than helpful for general developing debugging.

librz/arch/var.c
```diff
@@ -1087,6 +1102,138 @@ static inline bool is_not_read_nor_write(const RzAnalysisOpDirection direction)
...
+		return false;
+	}
+	P("occurence: %s       in         %s\n", ptr_end, op_esil);
+	*ptr_end = 0;
+	char *addr = ptr_end;
+	while ((addr[0] != '0' || addr[1] != 'x') && addr >= esil_buf + 1 && *addr != ',') {
+		addr--;
+	}
+	P("   => addr = %s\n", addr);
+	if (strncmp(addr, "0x", 2)) {
```
Rot127, line 1196:
```suggestion
	if (rz_str_startswith(addr, "0x")) {
```

librz/arch/var.c
thestr4ng3r: This printing here is intended to be removed entirely once tests pass. But if I notice I have something that makes sense to be reused later, I will put it there.

librz/core/canalysis.c
```diff
@@ -1577,7 +1577,7 @@ static bool analysis_block_cb(RzAnalysisBlock *bb, BlockRecurseCtx *ctx) {
 		if (rz_cons_is_breaked()) {
 			break;
 		}
-		RzAnalysisOp *op = rz_core_analysis_op(core, pos, RZ_ANALYSIS_OP_MASK_ESIL | RZ_ANALYSIS_OP_MASK_VAL | RZ_ANALYSIS_OP_MASK_HINT);
+		RzAnalysisOp *op = rz_core_analysis_op(core, pos, RZ_ANALYSIS_OP_MASK_ESIL | RZ_ANALYSIS_OP_MASK_IL | RZ_ANALYSIS_OP_MASK_VAL | RZ_ANALYSIS_OP_MASK_HINT);
```
Rot127, line 1580: Is it possible to check if the analysis at this point even tries to deduce stack vars?
Because if so, the `RZ_ANALYSIS_OP_MASK_IL` could be made optional for that case.
So we safe the allocations whenever this is called but no stack vars are analyzed.

I think it is even worth it to add a flag in RzAnalysis or core for that.

librz/util/bitvector.c
```diff
@@ -67,6 +67,25 @@ RZ_API bool rz_bv_init(RZ_NONNULL RzBitVector *bv, ut32 length) {
 	return true;
 }
 
+/**
+ * \brief Initialize \p bv, assigning the value of \p src
+ * \param bv uninitialized RzBitVector structure
+ * \param src value to copy into the newly initialized bitvector
+ * \return true if succeeded
+ */
+RZ_API bool rz_bv_init_copy(RZ_NONNULL RzBitVector *bv, RZ_NONNULL const RzBitVector *src) {
```
Rot127, line 76: Please move to a separate PR if you find time.

librz/il/il_traverse.c
```diff
@@ -0,0 +1,200 @@
...
+
+/**
+ * \brief Recurse into \p op, pre-order, depth-first
+ *
+ * \param op the op to traverse
+ * \param effect_cb callback to be called with each effect node encountered. If NULL, RZ_IL_RECURSE_STEP_INTO is assumed.
+ * \param effect_user passed to \p effect_cb
+ * \param pure_cb callback to be called with each effect node encountered. If NULL, no recursion into pure ops is done.
+ * \param pure_user passed to \p pure_cb
+ * \return false if any of the two callbacks returned RZ_IL_RECURSE_BREAK at some point, true otherwise
```
Rot127, line 164:
```suggestion
 * \param op The op to traverse
 * \param effect_cb Callback to be called with each effect node encountered. If NULL, RZ_IL_RECURSE_STEP_INTO is assumed.
 * \param effect_user Passed to \p effect_cb
 * \param pure_cb Callback to be called with each effect node encountered. If NULL, no recursion into pure ops is done.
 * \param pure_user Passed to \p pure_cb
 * \return false if any of the two callbacks returned RZ_IL_RECURSE_BREAK at some point, true otherwise
```

librz/il/il_traverse.c
Rot127, line 1: Nice idea!

PR #6179, opened 2026-04-07
rz_list_purge speedup

Farhan-25:

**Detailed description**

This PR:
--> Optimizes rz_list_purge
--> Currently `rz_list_purge `calls `rz_list_delete `on every node, which redundantly updates head/tail and patches prev/next pointers each iteration [^1]. Since the entire list is being destroyed, none of that bookkeeping is needed.

...

**Test plan**

A bench file named `test/bench/bench_list.c` has been made for testing.

**Benchmark Output**
Average of 5 test cases ran locally,
v1 -> old `rz_list_purge `using `rz_list_delete()`
v2 -> new speeup format of `rz_list_purge`

Without `free()` function:
| Test Case (Size) | v1 Avg (µs/op) | v2 Avg (µs/op) | Improvement |
|-----------------|---------------|---------------|------------|
| 1K              | 31.27         | 28.86         | +7.7%      |
| 100K            | 4,040         | 3,895         | +3.6%      |
| 1M              | 43,358        | 42,373        | +2.3%      |

With `free()` function:
| Test Case (Size) | v1 Avg (µs/op) | v2 Avg (µs/op) | Improvement |
|-----------------|---------------|---------------|------------|
| 100K            | 7,439         | 6,138         | +17.5%     |
| 1M              | 71,874        | 68,835        | +4.2%      |

image

...

**Note** : I would be making seperate PR for each of the speedup for better understanding, also i will use the same bench file. So i would make a new PR once this one is fully closed.

### DIFF
--- a/librz/util/list.c
+++ b/librz/util/list.c
@@ -113,6 +113,26 @@ RZ_API ut32 rz_list_length(RZ_NONNULL const RzList *list) {
 	return list->length;
 }
 
+static void _list_purge_with_free(RzList *list) {
+	RzListIter *it = list->head;
+	RzListFree fn = list->free;
+	while (it) {
+		RzListIter *next = it->next;
+		fn(it->val);
+		free(it);
+		it = next;
+	}
+}
+
+static void _list_purge_no_free(RzList *list) {
+	RzListIter *it = list->head;
+	while (it) {
+		RzListIter *next = it->next;
+		free(it);
+		it = next;
+	}
+}
+
 /**
  * \brief Empties the list without freeing the list pointer
  *
@@ -120,18 +140,14 @@ RZ_API ut32 rz_list_length(RZ_NONNULL const RzList *list) {
 RZ_API void rz_list_purge(RZ_NONNULL RzList *list) {
 	rz_return_if_fail(list);
 
-	RzListIter *it = list->head;
-	RzListFree fn = list->free;
-	while (it) {
-		RzListIter *next = it->next;
-		if (fn && it->val) {
-			fn(it->val);
-		}
-		free(it);
-		it = next;
+	if (list->free) {
+		_list_purge_with_free(list);
+	} else {
+		_list_purge_no_free(list);
 	}
+	list->head = NULL;
+	list->tail = NULL;
 	list->length = 0;
-	list->head = list->tail = NULL;
 }
 
 /**
--- a/librz/util/list.c
+++ b/librz/util/list.c
@@ -121,9 +121,13 @@ RZ_API void rz_list_purge(RZ_NONNULL RzList *list) {
 	rz_return_if_fail(list);
 
 	RzListIter *it = list->head;
+	RzListFree fn = list->free;
 	while (it) {
 		RzListIter *next = it->next;
-		rz_list_delete(list, it);
+		if (fn && it->val) {
+			fn(it->val);
+		}
+		free(it);
 		it = next;
 	}
 	list->length = 0;
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,7 +3,10 @@ name: CI
 on:
   push:
     branches:
-      - "**"
+      - "master"
+      - "dev"
+      - "stable"
+      - "container-*"
     tags:
       - v*
   schedule:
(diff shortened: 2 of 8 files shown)


wargio:
@notxvilka & @Rot127 i don't think this is a major improvement

Farhan-25:
> @notxvilka & @Rot127 i don't think this is a major improvement

Yes, but this is 1 of the 11 point speedup that was there in the issue. So if we add apply all of it then if will show some significant improvement.

librz/util/list.c
```diff
@@ -111,9 +121,13 @@ RZ_API void rz_list_purge(RZ_NONNULL RzList *list) {
...
 	RzListIter *it = list->head;
+	RzListFree fn = list->free;
 	while (it) {
 		RzListIter *next = it->next;
-		rz_list_delete(list, it);
+		if (fn && it->val) {
+			fn(it->val);
+		}
+		free(it);
 		it = next;
```
Rot127: Can you try and make the if case around the loop:

```c
RzListFree fn = list->free;
if (fn) {
	while (it) {
		RzListIter *next = it->next;
		fn(it->val);
		free(it);
		it = next;
	}
} else {
	while (it) {
		RzListIter *next = it->next;
		free(it);
		it = next;
	}
}
```

Rot127 (approved):
Not major, but for a 7+% performance increase for a few lines its still good.
@wargio Keep in mind that such basic data structures are used in hot paths. So any improvement in speed has a way higher impact to the whole runtime:
https://en.wikipedia.org/wiki/Amdahl%27s_law

librz/util/list.c
Rot127: @Farhan-25 Have you build for release btw? Please do so if not. Because otherwise not all optimizations get done.

Farhan-25: This is giving msvc heap error, 
```
RZ_API void rz_list_purge(RZ_NONNULL RzList *list) {
	rz_return_if_fail(list);

	RzListIter *it = list->head;
	RzListFree fn = list->free;
	if (fn) {
		while (it) {
			RzListIter *next = it->next;
			fn(it->val);
			free(it);
			it = next;
		}
	} else {
		while (it) {
			RzListIter *next = it->next;
			free(it);
			it = next;
		}
	}
	list->head = NULL;
	list->tail = NULL;
	list->length = 0;
}
```
this is correct so you want me to try this?

wargio: maybe split the while loops it into 2 static functions (`*_with_free_cb` and `*_no_free_cb`)

wargio:
> > @notxvilka & @Rot127 i don't think this is a major improvement
> 
> Yes, but this is 1 of the 11 point speedup that was there in the issue. So if we add apply all of it then if will show some significant improvement.

this was not a comment on your implementation, but more on the algo itself. You and @Rot127 have made a good point so is ok for me.

librz/util/list.c
Rot127: @Farhan-25 I typed it here on Github. My point is: the `if` condition for `fn` could also be checked only once, then the free function is assigned to a varible, and then the loop is entered. So the condition is not checked with every single element.

Farhan-25:
@wargio @Rot127 yes i did what you guys suggested and yes this is for now the fastest. I checked both rot's and wargio's code suggestion. Please check bench_list.c and the below table for more details. Also i have updated list.c accordingly, Thanks.

Average of 5 tests was 
## Benchmark Comparison (All Versions)

| Test Case            | v1 (µs/op) | v2 (µs/op) | v3 (µs/op) | v4 (µs/op) | Best | v4 vs v1 |
|---------------------|-----------|-----------|-----------|-----------|------|----------|
| 1K (no free)        | ~30.0     | ~29.0     | ~29.1     | ~28.5     | v4   | +5%      |
| 100K (no free)      | ~4400     | ~4000     | ~3980     | ~3900     | v4   | +11%     |
| 1M (no free)        | ~43000    | ~42500    | ~42200    | ~40000    | v4   | +8%      |
| 100K (with free)    | ~5900     | ~5750     | ~5650     | ~5500     | v4   | +7%      |
| 1M (with free)      | ~71000    | ~68000    | ~66500    | ~66000    | v4   | +7%      |

Rot127 (approved):
Nice!

notxvilka (approved):

PR #6215, opened 2026-04-12
Fix some memory leaks and overflow in util, core, and rz-diff

IndAlok:
**Your checklist for this pull request**
- [ ] I've read the [guidelines for contributing](https://github.com/rizinorg/rizin/blob/dev/CONTRIBUTING.md) to this repository.
- [ ] I made sure to follow the project's [coding style](https://github.com/rizinorg/rizin/blob/dev/DEVELOPERS.md#code-style).
- [ ] I've documented every `RZ_API` function and struct this PR changes.
- [ ] I've added tests that prove my changes are effective (required for changes to `RZ_API`).
- [ ] I've updated the [Rizin book](https://github.com/rizinorg/book) with the relevant information (if needed).
- [ ] I've used AI tools to generate fully or partially these code changes and I'm sure the changes are not copyrighted by somebody else.

**Detailed description**

just fixes few memory-safety issues in util/core paths, and keep state updates fail safe on alloc failure.

## brief

- checks alloc results before dereference in `rz_strbuf_append_n` and `rz_vector_clone`
- replaces unsafe fixed-buff formatting with bounded writes in `print.c` and `disasm.c`
- keeps `rz_print_set_rowoff` metadata consistent when growth fails
- switches panel filter cmd construction to helper based dyn growth and propagates alloc failure cleanly
- makes bitvector growth report failure so in-place casts dont continue on undersized storage
- avoids leaking the original escaped string buff when the trimmed `realloc()` fails
- makes the `rz-diff` resize path allocate replacement state first and only swap after all allocs succeed
...

**Test plan**

...

**Closing issues**

...

PR #6218, opened 2026-04-12
harden diff formatting and row offset growth

IndAlok:

**Detailed description**

harden bounded formatting in `rz_print` and keep row offset growth state consistent on alloc failure.

- replaces `sprintf()` with `snprintf()` in the hexdiff byte/chr formatting helpers
- updates `rz_print_set_rowoff` to only grow `row_offsets` metadata after the backing alloc succeeds

split from https://github.com/rizinorg/rizin/pull/6215

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/librz/util/print.c
+++ b/librz/util/print.c
@@ -1090,12 +1090,13 @@ RZ_API RZ_OWN char *rz_print_hexdump_str(RZ_NONNULL RzPrint *p, ut64 addr, RZ_NO
 static const char *getbytediff(RzPrint *p, char *fmt, ut8 a, ut8 b) {
 	if (*fmt) {
 		if (a == b) {
-			sprintf(fmt, "%s%02x" Color_RESET, p->cons->context->pal.graph_true, a);
+			// fmt[64] is known at call, so just preventing overflow.
+			snprintf(fmt, 64, "%s%02x" Color_RESET, p->cons->context->pal.graph_true, a);
 		} else {
-			sprintf(fmt, "%s%02x" Color_RESET, p->cons->context->pal.graph_false, a);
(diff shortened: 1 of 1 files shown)


librz/util/print.c
```diff
@@ -1104,14 +1105,13 @@ static const char *getchardiff(RzPrint *p, char *fmt, ut8 a, ut8 b) {
 	char ch = IS_PRINTABLE(a) ? a : '.';
 	if (*fmt) {
 		if (a == b) {
-			sprintf(fmt, "%s%c" Color_RESET, p->cons->context->pal.graph_true, ch);
+			snprintf(fmt, 64, "%s%c" Color_RESET, p->cons->context->pal.graph_true, ch);
```
wargio: where is this 64 coming from?

notxvilka: See the line above: `			// fmt[64] is known at call, so just preventing overflow.`

But I agreen, it's better to add this as a function argument, to avoid any mistakes in the future, once the number has changed.

notxvilka (changes requested):

librz/util/print.c
IndAlok: yep, as @notxvilka quoted, the comment meant that, as in `rz_print_hexdiff_str`, we have `char linediff, fmt[64];` and the helper reached through:
```
#define BD(a, b) getbytediff(p, fmt, ...)
#define CD(a, b) getchardiff(p, fmt, ...)
```
getbytediff or getchardiff receive only `char *fmt`, and once the arr becomes `char *` we don't know real size, and hence buffer size hardcoded to that...

but as agreed, i'll use `sizeof(fmt)` instead

notxvilka (approved):

wargio (approved):

Rot127 (approved):

PR #6219, opened 2026-04-12
Bit-Vector: fail when cast growth cant resize storage

IndAlok:

**Detailed description**

make in place bitvector casts fail cleanly if backing storage cannot be resized
split from https://github.com/rizinorg/rizin/pull/6215

**Test plan**

...

**Closing issues**

...


wargio (commented):
@Rot127

librz/util/bitvector.c
```diff
@@ -2269,7 +2282,10 @@ RZ_API bool rz_bv_cast_inplace(RZ_INOUT RZ_NONNULL RzBitVector *bv, ut32 to_size
 	}
 	if (NELEM(to_size, BV_ELEM_SIZE) > bv->_elem_len) {
 		// The bit vector needs a larger buffer.
-		resize_large_a(bv, NELEM(to_size, BV_ELEM_SIZE));
+		// abort the cast if storage couldnt be extended, esle resize as needed.
+		if (!resize_large_a(bv, NELEM(to_size, BV_ELEM_SIZE))) {
+			return false;
```
Rot127, line 2288: `rz_warn_if_reached`

Rot127 (approved):
One nitpick. Otherwise looks good.

librz/util/bitvector.c
IndAlok: image
it says here code shouldn't be reached, if logic is correct. but cant the `resize_large_a` failure path can happen legitimately, cuz alloc or realloc could fail at runtime...?

`rz_bv_cast_inplace` uses that as a normal outcome by returning false
image
so isnt this an expected failure path instead of unreachable code, as if we use `rz_warn_if_reached` we might mean memory growth failure should never happen and reaching that branch indicates a logic bug, or i misunderstood this completely?
@Rot127

Rot127: If allocations fail on runtime there is not much else to do tbh. It is more a way to inform the user that something went wrong and further results might be bogus.

IndAlok: makes sense, thx.

notxvilka (approved):

Rot127 (approved):

PR #6234, opened 2026-04-14
Enable basic blocks to have more than two successors

Rot127:

**Detailed description**

Removes the `jump` and `fail` members of basic blocks, and replaces them with a vector of successors.
The successors now can have several different conditions under which they are taken.

This should finally allow to have basic blocks with more than two outgoing edges.
Like indirect jumps (or calls in case of procedure CFG).

One commit implements the change, one is the renaming and API usage.

**Test plan**

All green

**Closing issues**

...

### DIFF
--- a/librz/arch/block.c
+++ b/librz/arch/block.c
@@ -57,6 +57,7 @@ static RzAnalysisBlock *block_new(RzAnalysis *a, ut64 addr, ut64 size) {
 	block->op_pos_size = DFLT_NINSTR;
 	block->sp_entry = ST32_MAX;
 	rz_vector_init(&block->sp_delta, sizeof(st16), NULL, NULL);
+	rz_vector_init(&block->_succ_addrs, sizeof(RzAnalysisSuccAddr), NULL, NULL);
 	block->cmpval = UT64_MAX;
 	block->fcns = rz_list_new();
 	if (size) {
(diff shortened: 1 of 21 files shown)


notxvilka:
Agree, this approach makes sense and quite flexible.

Rot127:
I am not sure what to do with the switch logic in the basic blocks though.

They are also successor basic blocks, but currently with way more info:

```c
struct rz_analysis_bb_t {
...
	RzAnalysisSwitchOp *switch_op;
...
};

typedef struct rz_analysis_switch_obj_t {
	ut64 addr;
	ut64 min_val;
	ut64 def_val;
	ut64 max_val;
	RzList /*<RzAnalysisCaseOp *>*/ *cases;
	RzType *enum_type;
} RzAnalysisSwitchOp;
```

I don't think we should end up with two fields with successors.

Maybe just integrate them into the new `succ_addrs` vector here?

Like:
- Add a new condition `RZ_TYPE_COND_SWITCH_MATCH`
- Add a `details` member to `RzAnalysisSuccAddr`, there we place a the `RzAnalysisSwitchOp`?
 
Any idea what would be better?

cc @thestr4ng3r @wargio

wargio:
oh, this will break a lot of stuff and plugins.

Rot127:
> oh, this will break a lot of stuff and plugins.

This is not the `RzAnalysisOp.jump`. That is just the basic block.
The `RzAnalysisOp` I did not touch on purpose. Because it gets replaced anyways with RzArch.
Or do I miss your point?

PR #6235, opened 2026-04-14
Fix syscall resolution ignoring user-provided number and add non-x86 tests

### LINKED ISSUE #2583: Fix the `rz_core_syscall_*()` and `rz_syscall_*()` syscall resolution
Currently, the syscall resolution code is hacky and quite limited.

It leads to the different behavior between `rz_syscall_get()` and `rz_syscall_get_num()` functions, and failure to resolve some of the cases: `rz_core_syscall_as_string(core, 4, 0x0`) returns `read(0)`, not `stat()`.

The task would be probably to refactor syscall searching code and adding more tests, especially for non-x86 architectures.

See:
- `librz/syscall/syscall.c`
- `librz/core/csyscall.c`

FearThePLOTO:

**Detailed description**

Fixed rz_syscall_get_num() returning -1 instead of false when db is NULL. Added unit tests for non-x86 architectures (ARM, ARM64, MIPS Linux syscalls)

1. **csyscall.c (commit 1)**: Fixed `rz_core_syscall_as_string()` ignoring user-provided syscall number. The code unconditionally set `n = -1` when `defVector > 0` (always true on Linux), causing it to always read from register instead of using the user-provided syscall number.

2. **syscall.c**: Fixed `rz_syscall_get_num()` returning `-1` instead of `false` when db is NULL. The function returns `bool`, so it should return `false`.

...

**Test plan**

Run: ninja -C build test
All syscall unit tests pass (6/6 tests).

...

**Closing issues**

closes #2583

...


librz/core/csyscall.c
```diff
@@ -32,10 +33,7 @@ RZ_API RZ_OWN char *rz_core_syscall_as_string(RzCore *core, st64 n, ut64 addr) {
 	RzReg *rreg = rz_analysis_get_reg(core->analysis);
 	RzSyscall *sysc = rz_analysis_get_syscall(core->analysis);
 	int defVector = rz_syscall_get_swi(sysc);
-	if (defVector > 0) {
-		n = -1;
-	}
-	if (n == -1 || defVector > 0) {
+	if (n == -1) {
```
wargio: n < 0 ?

librz/core/csyscall.c
```diff
@@ -32,10 +33,7 @@ RZ_API RZ_OWN char *rz_core_syscall_as_string(RzCore *core, st64 n, ut64 addr) {
...
 		n = (int)rz_core_reg_getv_by_role_or_name(core, "oeax");
 		if (!n || n == -1) {
```
wargio: same here.

librz/core/csyscall.c
FearThePLOTO: I applied the changes you suggested, using n < 0 instead of n == -1 for both checks. I also verified that oeax is the syscall number register (via =SN oeax), so syscall 0 is a valid value and shouldn't trigger register fallback. The tests pass. Are these changes correct?

PR #5164, opened 2025-05-22
 Migrate from Capstone to Zydis (x86 architecture)

notxvilka:
# SQUASH ME

**Your checklist for this pull request**
- [X] I've read the [guidelines for contributing](https://github.com/rizinorg/rizin/blob/master/DEVELOPERS.md) to this repository
- [X] I made sure to follow the project's [coding style](https://github.com/rizinorg/rizin/blob/master/DEVELOPERS.md#code-style)
- [ ] I've documented or updated the documentation of every function and struct this PR changes. If not so I've explained why.
- [ ] I've added tests that prove my fix is effective or that my feature works (if possible)
- [ ] I've updated the [rizin book](https://github.com/rizinorg/book) with the relevant information (if needed)

**Detailed description**

An updated version of https://github.com/rizinorg/rizin/pull/4832

- [X] Build zydis as a dependency
- [X] Use zydis structs / functions etc for disassembly and analysis
- [X] Use zydis for `asm_x86` file
- [X] Use zydis for `analysis_x86` file and update zydis to x86 ESIL generation
- [X] Migrate to zydis for x86 RzIL
- [X] Change Tests/code to account for differences in outputs presented by Capstone and Zydis
- [X] Extend big endian host system support in Zydis (in particular, S390/System Z architecture)
- [ ] Document functions of `analysis_x86` and `asm_x86` files

**Test plan**

CI is green

**Closing issues**

closes #4709

PR #6238, opened 2026-04-14
x86_zydis: Fix mistakes in Capstone to Zydis migration

### LINKED ISSUE #6236: Missing x86 mnemonics
### Work environment

| OS/arch/bits (mandatory)                             | Arch Linux x86_64
| File format of the file you reverse (mandatory)      | N.A
| Architecture/bits of the file (mandatory)            | x86 16bit
| `rizin -v` full output, **not truncated** (mandatory)  | rizin 0.9.0 @ linux-x86-64 commit: 3bbf3b3c01525ef25ea7713e5a8272b0dc8af140

### Expected behavior
POPA and PUSHA should be properly analysed and there `type` be set properly. 

### Actual behavior
Currently the type is set to null. The reason is currently there are missing mnemonics of x86 16 bit in `x86_mnemonics.h`.
For example POPA/PUSHA which are 16bit equivalent of 32 bit POPAD/PUSHAD are absent from mnemonics header and also the `analysis_x86_zydis.c` does not have any handling of them.

You can also see there is no ESIL information set about this instructions which points to missing analysis of this instructions.

You can see the `type` field is set to null
```bash
rizin malloc://32
[0x00000000]> e asm.arch=x86
[0x00000000]> e asm.bits=16
[0000:0000]> wa "popa;pusha"
[0000:0000]> ao
address: 0x0
opcode: popa
disasm: popa
pseudo: popa 
mnemonic: popa
description: pop all general-purpose registers
mask: ff
prefix: 0
id: 610
bytes: 61
refptr: 0
size: 1
sign: false
type: null
cycles: 1
rzil: empty
opex:
  operands: []
family: cpu
[0000:0000]> s 1
[0000:0001]> ao
address: 0x1
opcode: pusha
disasm: pusha
pseudo: pusha 
mnemonic: pusha
description: push all general-purpose registers
mask: ff
prefix: 0
id: 666
bytes: 60
refptr: 0
size: 1
sign: false
type: null
cycles: 1
rzil: empty
opex:
  operands: []
family: cpu
```

### LINKED ISSUE #6237: Incorrect instruction metadata for PUSHAD/POPAD
### Work environment

| OS/arch/bits (mandatory)                             | Arch Linux x86_64
| File format of the file you reverse (mandatory)      | N.A
| Architecture/bits of the file (mandatory)            | x86 32bit
| `rizin -v` full output, **not truncated** (mandatory)  | rizin 0.9.0 @ linux-x86-64 commit: 3bbf3b3c01525ef25ea7713e5a8272b0dc8af140

### Expected behavior
Metadata for PUSHAD / POPAD instructions should show stackptr: 32. (because it affects 8 registers)
For POPAD the type should be `POP`

### Actual behavior
Metadata is wrong.
stackptr for PUSHAD/POPAD is shown as 4 (single register)
The `type` for POPAD is `RET`

```bash
❯ rizin malloc://32
 -- Seek at relative offsets with 's +<offset>' or 's -<offset>'
[0x00000000]> e asm.arch=x86
[0x00000000]> e asm.bits=32
[0x00000000]> wa "pushad;popad"
[0x00000000]> ao
address: 0x0
opcode: pushad
disasm: pushad
pseudo: pushad 
mnemonic: pushad
description: push all general-purpose registers
mask: ff
prefix: 0
id: 667
bytes: 60
refptr: 0
size: 1
sign: false
type: rpush
cycles: 1
esil: 0,esp,+,4,esp,-=,eax,esp,=[4],4,esp,-=,ecx,esp,=[4],4,esp,-=,edx,esp,=[4],4,esp,-=,ebx,esp,=[4],4,esp,-=,esp,=[4],4,esp,-=,ebp,esp,=[4],4,esp,-=,esi,esp,=[4],4,esp,-=,edi,esp,=[4]
opex:
  operands: []
family: cpu
stackop: inc
stackptr: 4
[0x00000000]> ao @ 1
address: 0x1
opcode: popad
disasm: popad
pseudo: popad 
mnemonic: popad
description: pop all general-purpose registers
mask: ff
prefix: 0
id: 611
bytes: 61
refptr: 0
size: 1
sign: false
type: ret
cycles: 3
esil: esp,[4],4,esp,+=,edi,=,esp,[4],4,esp,+=,esi,=,esp,[4],4,esp,+=,ebp,=,esp,[4],4,esp,+=,esp,[4],4,esp,+=,ebx,=,esp,[4],4,esp,+=,edx,=,esp,[4],4,esp,+=,ecx,=,esp,[4],4,esp,+=,eax,=,esp,=
rzil: empty
opex:
  operands: []
family: priv
stackop: inc
stackptr: -4
```

MrQuantum1915:

**Detailed description**

There were some mistakes in Capstone to Zydis migration of x86 analysis https://github.com/rizinorg/rizin/pull/5164

i found this when fixing ROP search tests in https://github.com/rizinorg/rizin/pull/6108

- Some capstone era macros were mapped to non existent zydis enums
- anlaysis of PUSHA/PUSHAD/POPA/POPAD was mistakenly altered in analysis file and some were entirely missing
- there was also no RzIL lifter implemented for those instructions
- along with those changes i also fixed some bugs that i noticced while fixing...

---

**Commit 1-4**: Fix PUSHA/PUSHAD/POPA/POPAD ([Ref1](https://www.felixcloutier.com/x86/pusha:pushad), [Ref2](https://www.felixcloutier.com/x86/popa:popad))

1. **Commit 1**: 
 - Addded missing macros for 4 instructions and replaced capstone era : `l`,`w` suffix with intel specification (what zydis uses)
 - added analysis for all 4 instructions (fixed mistaken fallthrough to other instructions in switch block)
      
 - added esil generators: **NOTE:** Intel docs says: - [intel: popa/popad](https://www.felixcloutier.com/x86/popa:popad)
  
>The value on the stack for the ESP or SP register is ignored. Instead, the ESP or SP register is incremented after each register is loaded.

- Hence the previous implementation in `anop_esil` for POPAD was wrong, it incorrectly emitted an explicit load into ESP from the stack. Replaced it with simple stack pointer increment (esp,+=).
- set access info: 
- for `PUSHA/PUSHAD/POPA/POPAD` I set mem info  and reg info explicitly. I explicitly set meminfo for them according to existing block of `PUSH` instruction and capstone era code also set them explicitly for PUSHA/PUSHAD [Reference](https://github.com/rizinorg/rizin/pull/5164/changes#diff-94c33a3b334a4b20fb84419833455deaa7f625ec826ab6cf1656d5c139f84820L1911). For the reg info: the loop before the switch block uses `operand_count_visible`, but PUSHA/PUSHAD/POPA/POPAD does not have explicit operands, there operands are implicit, so we have to set all 8 reg info explicitly
- So i want to ask whether setting mem block here is fine or was incorrect in that migration itself [@notxvilka @wargio @Rot127 ]
     
- while fixing the switch block in `anop` function i noticed that we use `RZ_ANALYSIS_STACK_INC` for both PUSH and POP instructions. In fact `RZ_ANALYSIS_STACK_DEC` is not implemented in `op.c` (although its declared as enum). Note that the stack frame size calc is not broken because that info is stored in "sign" of stackptr variable see:
 ```c
RZ_API RzStackAddr rz_analysis_op_apply_sp_effect(RzAnalysisOp *op, RzStackAddr sp) {
	rz_return_val_if_fail(op, sp);
	// When changing something here, also update rz_analysis_op_describe_sp_effect() to describe
	// the same effect in a textual form.
	switch (op->stackop) {
	case RZ_ANALYSIS_STACK_INC:
		return sp - op->stackptr;
	case RZ_ANALYSIS_STACK_RESET:
		return 0;
	default:
		return sp;
	}
}
```
       
- but using `RZ_ANALYSIS_STACK_DEC` , is broken , bcz it falls to default case. Actually analysis code of some archs like `arm`, `riscv`, `rx`, `h8300`,etc DOES USE `RZ_ANALYSIS_STACK_DEC` in their code; which means they are BROKEN acutally!

- [@notxvilka @wargio @Rot127 ] if you agree i would like to open reproducible and fix this issue in a seperate PR. kindly confirm

- as `POPAD` correctly set as `POP` instead of `RET` hence the output in `cmd_rop` test changed

2. **Commit 2:** RzIL lifters for 4 instructions were missing, so implemented that.
3. **Commit 3:** I noticed the generated IL for `POPAD` was applying the real mode segment translation (`ss<<4`) `ss*16 + offset`, but segment calc only work for 16bit (real mode) and not for 32bit (protected mode). [Reference](https://stackoverflow.com/questions/15335003/what-is-the-purpose-of-segment-registers-in-x86-protected-mode). And infact the [comment in `common.c`](https://github.com/rizinorg/rizin/blob/c930244ceb95460c5e16fc52ee0da739c04a464e/librz/arch/isa/x86/common.c#L691) says that only real mode is supported for now. So instead of checking `bits!=64` it should be `bits==16` to restrict it for 16bit env.

4. **Commit 4:** Added tests for changes both in x86 analysis and x86 asm (covers RzIl and other metadata)

---

**Commit 5**: Remove redundant mnemonics and non-existent zydis mapping (I ONLY focused on NON-EXISTENT ZYDIS MAPPING)
- `MOVSABS` - [MOVABS is same as MOV in intel - see first answer](https://reverseengineering.stackexchange.com/questions/2627/what-is-the-meaning-of-movabs-in-gas-x86-att-syntax)
- `SAL` -[SAL = SHL, which is already implemented](https://www.felixcloutier.com/x86/sal:sar:shl:shr#:~:text=The%20shift%20arithmetic%20left%20(SAL)%20and%20shift%20logical%20left%20(SHL)%20instructions%20perform%20the%20same%20operation%3B)
- `RETF` - [RET Far and RET Near come under same RET instruction in intel](https://www.felixcloutier.com/x86/ret#:~:text=The%20RET%20instruction%20can,executing%20program%20or%20procedure)
- `RETQ` - [Q = Quadword(64bit), intel covers all in RET, see first answer](https://stackoverflow.com/questions/42653095/what-is-the-difference-between-retq-and-ret)

---

**Commit 6**: Implement `FCOMIP`/`FUCOMIP` analysis + RzIL + Fix FCOMI helper
- capstone swaped `I` and `P`, fixed mnemonic
- analysis was missing added cases
- implemented RzIL lifter for FCOMIP/FUCOMIP
- I noticed that RzIL was having st(0) st(0) comparison for both eg `dff0`(= `fcomip st(0), st(0)`) and `dff1` (=`fcomip st(0),st(1)`), the reason was the helper for lifter of FCOMI was a bit buggy, it hardcoded `st(i)` to `op(0)` instead of selecting op(1), because visible operands for FCOMIP/FUCOMIP are two. I added if else block to dynamically fetch operand based on visible op count. (as st(0) can be implicit too [`fcomip` compares st(0) with st(i)]) .
- added tests (now rzil for `dff4` correctly compares st(0) to st(4)). On dev branch this test would fail.

**Test plan**

All test pass successfully (new+old). The analysis metadata for all instruction correct.

...

**Closing issues**

closes #6236
closes #6237
...

### CODE REVIEW

File: test/db/analysis/x86_16 [hunk 1]
```diff
@@ -507,3 +507,71 @@ EXPECT=<<EOF
 ]
 EOF
 RUN
+
+NAME=pusha analysis
```
wargio: merge the tests into one.

notxvilka (changes requested):
Ideally, we should check it with the https://github.com/rizinorg/rz-tracetest.
Also, creating a handmade program that uses these instructions for the one more "emulateme" test would work as well (compare the expected result with the calculated via the RzIL VM emulation result).

notxvilka:
cc @DMaroo please take a look too

wargio (approved):

MrQuantum1915:
> Ideally, we should check it with the https://github.com/rizinorg/rz-tracetest. Also, creating a handmade program that uses these instructions for the one more "emulateme" test would work as well (compare the expected result with the calculated via the RzIL VM emulation result).


@notxvilka i will look into this after my univ exams get over in a ~week, until then marking this as Draft PR

PR #6250, opened 2026-04-17
Make `rz_vector_assign_at()` behave consistently.

Rot127:

**Detailed description**

`rz_pvector_assign_at` and `rz_vector_assign_at` were not really documented and did not allow to assign beyond the length of the vector (even if there was still capacity left).

This is a problem, because the only other way to assign beyond the length (but within the capacity) was `rz_pvector_set`, which doesn't update the length.

**Test plan**

Added

**Closing issues**

...

### DIFF
--- a/librz/util/vector.c
+++ b/librz/util/vector.c
@@ -465,7 +465,14 @@ RZ_API void *rz_vector_shrink(RzVector *vec) {
 	return vec->a;
 }
 
-RZ_API void *rz_vector_flush(RzVector *vec) {
+/**
+ * \brief Turn the vector into a fixed-size array.
+ * This will clear the vector and return an array of its original contents whose
+ * ownership is transferred to the caller.
+ * This is useful when RzVector is used for its dynamically growing functionality as an
+ * intermediate step to generate a fixed-size array in the end.
+ */
+RZ_API void *rz_vector_flush(RZ_BORROW RzVector *vec) {
 	rz_return_val_if_fail(vec, NULL);
 	rz_vector_shrink(vec);
 	void *r = vec->a;
--- a/librz/util/vector.c
+++ b/librz/util/vector.c
@@ -479,7 +479,7 @@ RZ_API void *rz_vector_shrink(RzVector *vec) {
  * This is useful when RzVector is used for its dynamically growing functionality as an
  * intermediate step to generate a fixed-size array in the end.
  */
-RZ_API void *rz_vector_flush(RZ_BORROW RzVector *vec) {
+RZ_API RZ_OWN void *rz_vector_take_array(RZ_BORROW RzVector *vec) {
 	rz_return_val_if_fail(vec, NULL);
 	rz_vector_shrink(vec);
 	void *r = vec->a;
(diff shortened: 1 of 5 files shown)


librz/util/vector.c
```diff
@@ -83,6 +87,19 @@ RZ_API void rz_vector_free(RzVector *vec) {
...
 
+static void rz_vector_assign(RzVector *vec, void *p, const void *elem) {
+	rz_return_if_fail(vec && p && elem);
+	memcpy(p, elem, vec->elem_size);
+}
+
+/**
+ * \brief Set \p n elements, starting at element \p i to \p c.
+ */
+static void rz_vector_memset(RzVector *vec, size_t i, size_t n, ut8 c) {
```
wargio: i wonder if there are other usecases besides `zeroing`

Rot127: Yeah, good point. Let me change it.

librz/util/vector.c
```diff
@@ -83,6 +87,19 @@ RZ_API void rz_vector_free(RzVector *vec) {
...
 
+static void rz_vector_assign(RzVector *vec, void *p, const void *elem) {
+	rz_return_if_fail(vec && p && elem);
+	memcpy(p, elem, vec->elem_size);
+}
+
+/**
+ * \brief Set \p n elements, starting at element \p i to \p c.
+ */
+static void rz_vector_memset_zero(RzVector *vec, size_t i, size_t n) {
```
notxvilka: Isn't "zeroize" shorter? I know, I know, the word doesn't exist in English, but everyone would understand, I think.

librz/util/vector.c
```diff
@@ -219,6 +248,13 @@ RZ_API void rz_vector_remove_range(RzVector *vec, size_t index, size_t count, vo
 	}
 }
 
+/**
+ * \brief Deletes all elements in the vector. Capacity stays the same.
+ */
+RZ_API void rz_vector_purge(RZ_BORROW RzVector *vec) {
```
notxvilka, line 260: We have `purge` and `clear`. I think description of both should reflect more on what is the difference between two.

librz/util/vector.c
```diff
@@ -436,7 +472,14 @@ RZ_API void *rz_vector_shrink(RzVector *vec) {
...
 
-RZ_API void *rz_vector_flush(RzVector *vec) {
+/**
+ * \brief Turn the vector into a fixed-size array.
+ * This will clear the vector and return an array of its original contents whose
+ * ownership is transferred to the caller.
+ * This is useful when RzVector is used for its dynamically growing functionality as an
+ * intermediate step to generate a fixed-size array in the end.
+ */
+RZ_API void *rz_vector_flush(RZ_BORROW RzVector *vec) {
```
notxvilka: I am not sure `flush` would be a good choice here. "Flush" typically implies writing out pending data (I/O buffers, caches), not converting a container. It doesn't hint at the return type or the ownership transfer.

I think, better choices could be:
- `rz_vector_into_array`
- `rz_vector_steal_array` (like in Glib convention)

notxvilka (changes requested):
LGTM with a few naming changes

librz/util/vector.c
Rot127: What about `rz_vector_take`?
At least to me this seems the most natural way of describing ownership change

notxvilka: `rz_vector_take_array` then

notxvilka (approved):

PR #6255, opened 2026-04-18
Adding RzIL unicode/enriched lines api for cutter

NewtronReal:

**Detailed Description**
This PR introduces `rz_core_il_print_rzil` API wihch returns RzIL lines for given address and number of lines or all RzIL ops in the given chunk similar to 'rz_core_print_disasm' for printing disassembly lines. This was introduced as a helper API for accessing rzil lines from rizin to cutter [rizinorg/cutter#3196](https://github.com/rizinorg/cutter/issues/3196) but can be scaled to RzIL enriched in the future. The function returns a RzPVector or RzAnalysisDisasmText where the 
text-> field corresponds to formatted RzIL text
arrow-> corresponds to the reflines
offset-> corresponds to the vaddress of the line

Structs added

```c
typedef struct il_state_t RzILState;
typedef struct rz_core_il_print_options RzCoreILPrintOptions;
```

Functions added

```c
/* \brief Format instructions and add generated RzIL code to an analysis text vector.
 *
 * Decode instructions starting at the specified address, format their
 * RzIL equivalents into printable strings, using optional pretty printing,
 * unicode conversion and syntax highlighting, then store the strings as
 * RzAnalysisDisasmText entries in the vector provided.
 *
 * \param core      Pointer to the current RzCore instance.
 * \param addr      Starting address for instruction decoding.
 * \param buf       Pointer to memory buffer that contains instructions.
 * \param len       Length of \p buf in bytes.
 * \param n_lines   Maximum number of instructions to decode. Uses
 *                  core->blocksize if value is 0.
 * \param options   Formatting parameters and vector for storing results.
 *
 * \return Number of instructions decoded successfully.
 */
RZ_API int rz_core_il_print_rzil(RZ_NONNULL RzCore *core, ut64 addr, RZ_NONNULL ut8 *buf, size_t len, size_t n_lines, RZ_NULLABLE RzCoreILPrintOptions *options) ;

// static functions
static inline void emit_span_to_strbuf(const char *s, size_t n, const char *color, RzStrBuf *sb) ;
static void core_colorify_il_statement_to_strbuf(RzConsContext *ctx, const char *il_stmt, const char delim, ut64 addr, RzStrBuf *sb) ;
static const char *core_il_get_token_color(RzILUnicodeColorifyState state, const char *prev_color, RZ_NONNULL RzConsContext *ctx) ;
static void core_colorify_il_statement_unicode_to_strbuf(RzConsContext *ctx, const char *il_stmt, const char delim, ut64 addr, RzStrBuf *sb) ;
static ut64 core_il_get_refline_at(ut64 vat, RZ_NONNULL RzPVector /*<RzAnalysisRefline *>*/ *reflines)
static void init_asmqjmps(RZ_NONNULL RzCore *core);
static RzILState *il_state_init(RZ_NONNULL RzCore *core);
static void il_state_reflines_init(RzILState *ils);
static void core_il_stringify_single_il(RzILState *ils);
static void il_state_reflines_fini(RzILState *ils);
static bool il_state_complete(RzILState *ils);
static void il_state_free(RzILState *ils);

```

...

**Test plan**

Build and compile
Pass `test_rz_core_il_print_rzil` in tests

https://github.com/user-attachments/assets/e614a76b-691c-4e11-b343-d3b19d409288

...

**Closing issues**
None

...

### DIFF
--- a/librz/core/cil.c
+++ b/librz/core/cil.c
@@ -562,21 +562,39 @@ static void core_colorify_il_statement_unicode_to_strbuf(RzConsContext *ctx, con
  * \param unicode - whether enable unicode formatting or not
  * \param colorize - color the text
  */
-RZ_API void rz_core_il_print_rzil(RZ_NONNULL RzCore *core, RZ_NONNULL RzPVector *vec, ut64 addr, int n_lines, bool pretty, bool unicode, bool colorize) {
-	rz_return_if_fail(core && vec);
+RZ_API int rz_core_il_print_rzil(RZ_NONNULL RzCore *core, ut64 addr, RZ_NONNULL ut8 *buf, int len, int n_lines, RZ_NULLABLE RzCoreILPrintOptions *options) {
+	rz_return_val_if_fail(core && buf && (n_lines || len), 0);
+	RzPVector *vec = NULL;
+	bool pretty = false;
+	bool unicode = false;
+	bool colorize = false;
+	if (options) {
+		pretty = options->pretty;
+		colorize = options->colorize;
+		unicode = options->unicode;
+		vec = options->vec;
+	}
+	if (!vec) {
+		return 0;
+	}
+	if (!n_lines) {
+		n_lines = core->blocksize;
+	}
 	const char *il_stmt = NULL;
 	const char delim = pretty ? '\n' : ' ';
 	RzStrBuf sb;
+	RzPVector *reflines;
 	RzAnalysisOp op;
-	RzPVector *reflines = rz_analysis_reflines_get(core->analysis,
-		addr, core->block, core->blocksize, n_lines,
+	int ops_count = 0, inc = 0, idx;
+	ut64 current = addr, vat;
+	ut8 *nbuf = NULL;
+	ut8 min_op_size = rz_analysis_archinfo(core->analysis, RZ_ANALYSIS_ARCHINFO_MIN_OP_SIZE);
+complete:
+	reflines = rz_analysis_reflines_get(core->analysis,
+		addr, buf, len, n_lines,
 		rz_config_get_i(core->config, "asm.lines.out"),
 		rz_config_get_b(core->config, "asm.lines") ? rz_config_get_b(core->config, "asm.lines.call") : false);
 	rz_analysis_set_reflines(core->analysis, reflines);
-	int ops_count = 0;
-	ut64 current = addr, vat;
-	ut8 bytes[256];
-	ut8 min_op_size = rz_analysis_archinfo(core->analysis, RZ_ANALYSIS_ARCHINFO_MIN_OP_SIZE);
 	if (!core->keep_asmqjmps) { // hack
 		core->asmqjmps_count = 0;
 		ut64 *p = realloc(core->asmqjmps, RZ_CORE_ASMQJMPS_NUM * sizeof(ut64));
@@ -588,22 +606,27 @@ RZ_API void rz_core_il_print_rzil(RZ_NONNULL RzCore *core, RZ_NONNULL RzPVector 
 			}
 		}
 	}
-	while (ops_count < n_lines) {
-		rz_core_seek_arch_bits(core, current);
+	for (idx = 0; ops_count < n_lines && idx < len; ops_count++, idx += inc) {
+		current = addr + idx;
 		RzAnalysisDisasmText *dst = RZ_NEW0(RzAnalysisDisasmText);
 		dst->arrow = UT64_MAX;
 		vat = rz_core_pava(core, current);
 		rz_strbuf_init(&sb);
-		int read = rz_io_nread_at(core->io, current, bytes, sizeof(bytes));
-		if (read < 1) {
-			rz_strbuf_appendf(&sb, "0x%" PFMT64x "%creaderror", current, delim);
-			goto finish_str;
+		if (core->print->flags & RZ_PRINT_FLAGS_UNALLOC) {
+			if (!rz_io_is_valid_offset(core->io, current, 0)) {
+				rz_strbuf_appendf(&sb, "0x%" PFMT64x "%cunmapped", current, delim);
+				inc = 1;
+				goto finish_str;
+			}
 		}
+		rz_core_seek_arch_bits(core, current);
 		rz_analysis_op_init(&op);
-		if (rz_analysis_op(core->analysis, &op, current, bytes, read, RZ_ANALYSIS_OP_MASK_IL) < 1) {
+		if (rz_analysis_op(core->analysis, &op, current, buf + idx, (int)(len - idx), RZ_ANALYSIS_OP_MASK_IL) < 1) {
 			rz_strbuf_appendf(&sb, "0x%" PFMT64x "%cinvalid", current, delim);
+			inc = 1;
 			goto finalize;
 		}
+		inc = op.size;
 		if (!op.il_op) {
 			RZ_LOG_DEBUG("Empty IL at 0x%08" PFMT64x "...", op.addr);
 			rz_strbuf_appendf(&sb, "0x%" PFMT64x "%cempty il", op.addr, delim);
@@ -633,7 +656,7 @@ RZ_API void rz_core_il_print_rzil(RZ_NONNULL RzCore *core, RZ_NONNULL RzPVector 
 		} else {
 			rz_strbuf_appendf(&sb, "0x%" PFMT64x "%c%s", op.addr, delim, il_stmt);
 		}
-		free((void *)il_stmt);
+		RZ_FREE(il_stmt)
 		if (reflines) {
 			void **iter;
 			rz_pvector_foreach (reflines, iter) {
@@ -649,8 +672,7 @@ RZ_API void rz_core_il_print_rzil(RZ_NONNULL RzCore *core, RZ_NONNULL RzPVector 
 	finish_str:
 		dst->offset = rz_core_pava(core, current);
 		dst->text = rz_str_dup(rz_strbuf_get(&sb));
-		current += op.size > 0 ? op.size : min_op_size;
-		ops_count++;
+		inc = inc ? inc : RZ_MAX(1, min_op_size);
 		if (!rz_pvector_push(vec, dst)) {
 			free(dst->text);
 			free(dst);
@@ -659,6 +681,20 @@ RZ_API void rz_core_il_print_rzil(RZ_NONNULL RzCore *core, RZ_NONNULL RzPVector 
 	}
 	rz_analysis_set_reflines(core->analysis, NULL);
 	rz_pvector_free(reflines);
+	if (!options->cbytes && ops_count < n_lines) {
+		addr = current + inc;
+		if (len < 16) {
+			len = 16;
+		}
+		free(nbuf);
+		buf = nbuf = malloc(len);
+		if (nbuf) {
+			rz_io_read_at_mapped(core->io, addr, buf, len);
+			goto complete;
+		}
+	}
+	RZ_FREE(nbuf);
+	return ops_count;
 }
 
 RZ_IPI void rz_core_il_cons_print(RZ_NONNULL RzCore *core, RZ_NONNULL RZ_BORROW RzIterator *iter, bool pretty, bool unicode) {
--- a/librz/core/cil.c
+++ b/librz/core/cil.c
@@ -351,15 +351,27 @@ static inline void emit_span(const char *s, size_t n, const char *color) {
 	}
 }
 
-/**
- * \brief Colorize a stringified RzIL effect body to the cons buffer.
- *
- * Emits only the body (no address prefix, no newline) with the same palette
- * as \c plf. A NULL or empty \p il_stmt emits nothing.
- */
-RZ_IPI void rz_core_il_colorize_body(RZ_NONNULL RzConsContext *ctx, RZ_NULLABLE const char *il_stmt) {
-	rz_return_if_fail(ctx);
+static inline void emit_span_to_strbuf(const char *s, size_t n, const char *color, RzStrBuf *sb) {
+	if (n < 1) {
+		return;
+	}
+	if (color) {
+		rz_strbuf_appendf(sb, "%s%.*s" Color_RESET, color, (int)n, s);
+	} else {
+		rz_strbuf_appendf(sb, "%.*s", (int)n, s);
+	}
+}
+
+static void core_colorify_il_statement(RzConsContext *ctx, const char *il_stmt, const char delim, ut64 addr) {
+	rz_cons_printf("%s0x%" PFMT64x Color_RESET "%c", ctx->pal.label, addr, delim);
+	rz_core_il_colorize_body(ctx, il_stmt);
+	rz_cons_newline();
+}
+
+static void core_colorify_il_statement_to_strbuf(RzConsContext *ctx, const char *il_stmt, const char delim, ut64 addr, RzStrBuf *sb) {
+	rz_strbuf_appendf(sb, "%s0x%" PFMT64x Color_RESET "%c", ctx->pal.label, addr, delim);
 	if (RZ_STR_ISEMPTY(il_stmt)) {
+		rz_strbuf_appendf(sb, "\n");
 		return;
 	}
 
@@ -370,13 +382,13 @@ RZ_IPI void rz_core_il_colorize_body(RZ_NONNULL RzConsContext *ctx, RZ_NULLABLE 
 		const char ch = il_stmt[i];
 
 		if (ch == '(' || ch == ')') {
-			emit_span(il_stmt + prev, i - prev, color);
-			rz_cons_printf("%s%c" Color_RESET, ctx->pal.meta, ch);
+			emit_span_to_strbuf(il_stmt + prev, i - prev, color, sb);
+			rz_strbuf_appendf(sb, "%s%c" Color_RESET, ctx->pal.meta, ch);
 			prev = i + 1;
 			color = (ch == '(') ? ctx->pal.flow : NULL;
 		} else if (ch == ' ') {
-			emit_span(il_stmt + prev, i - prev, color);
-			rz_cons_printf(" ");
+			emit_span_to_strbuf(il_stmt + prev, i - prev, color, sb);
+			rz_strbuf_appendf(sb, " ");
 			prev = i + 1;
 			color = NULL;
 		} else if (i == prev && prev > 0 && il_stmt[prev - 1] == ' ') {
@@ -384,13 +396,8 @@ RZ_IPI void rz_core_il_colorize_body(RZ_NONNULL RzConsContext *ctx, RZ_NULLABLE 
 		}
 	}
 
-	emit_span(il_stmt + prev, len - prev, color);
-}
-
-static void core_colorify_il_statement(RzConsContext *ctx, const char *il_stmt, const char delim, ut64 addr) {
-	rz_cons_printf("%s0x%" PFMT64x Color_RESET "%c", ctx->pal.label, addr, delim);
-	rz_core_il_colorize_body(ctx, il_stmt);
-	rz_cons_newline();
+	emit_span_to_strbuf(il_stmt + prev, len - prev, color, sb);
+	rz_strbuf_appendf(sb, "\n");
 }
 
 static bool unicode_colorify_state_is_varname(RzILUnicodeColorifyState state) {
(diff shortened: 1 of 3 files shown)


librz/core/cil.c
```diff
@@ -479,6 +479,72 @@ static void core_colorify_il_statement_unicode(RzConsContext *ctx, const char *i
 	rz_cons_newline();
 }
 
+RZ_API void rz_core_il_print_rzil(RZ_NONNULL RzCore *core, RZ_NONNULL RzPVector *vec, bool pretty, bool unicode, bool colorize) {
```
notxvilka: If you use nonnull markers, please also add `rz_return_if_fail()` macro.

notxvilka: When you add an API function, please add Doxygen description

notxvilka (changes requested):
In general, we should avoid printing stuff from the API and return a string or a RzStrBuf instead for printing. Moreover, if this function becomes the main point for printing RzIL, ideally, the Rizin code for corresponding command handler also should be changed to use it.

See `plf` and `plF` commands for the reference.

librz/core/cil.c
NewtronReal: Done for boh vec and core.

NewtronReal: Done

NewtronReal:
> In general, we should avoid printing stuff from the API and return a string or a RzStrBuf instead for printing. Moreover, if this function becomes the main point for printing RzIL, ideally, the Rizin code for corresponding command handler also should be changed to use it.
> 
> See `plf` and `plF` commands for the reference.

Done instead of printing to the console then sending the buffer i've appended it into a rzstrbuf then freed it after. 

I wanted a parallel widget for rzil alongside disassembly view so i was just making an api analogous to rz_core_print_disasm. It wasn't intendend to be used as a main point for printing RzIL. It was designed cutter oriented. But if it is the convention to make it the main printing function I'm more happy to do so. As of now i have modified the function so that instead of getting the chunk of current function it is able to load all the instructions between two RVAs.

Should we create a special struct for storing rzil info instead of leveraging RzAnalysisDisasmText like RzAnalysisRzilText which will have an additional attribute report as well?
(as of now report felt irrelevant for the api so i commented it).

Here's how the cutter widget looks now

https://github.com/user-attachments/assets/55dd2e7e-5148-4806-b791-b3b3373f37d6

still working on the cutter as well as the rz_core apis. Will be updating the work.

librz/core/cil.c
```diff
@@ -479,6 +522,142 @@ static void core_colorify_il_statement_unicode(RzConsContext *ctx, const char *i
...
+
+	// if (ctx) {
+	// 	RzILTypeEffect t;
+	// 	char *report;
+	// 	rz_il_validate_effect(op->il_op, ctx, NULL, &t, &report);
+	// 	if (report) {
+	// 		rz_cons_println(report);
+	// 		free(report);
+	// 	}
+	// }
```
wargio: remove?

librz/core/cil.c
```diff
@@ -479,6 +522,142 @@ static void core_colorify_il_statement_unicode(RzConsContext *ctx, const char *i
...
+	togo:
```
wargio: `fail:`

librz/core/cil.c
NewtronReal: Should I remove it or make another structure to keep the report as well like RzILDisasmText with{addr, arrow,text,report}?

Rot127: Does the report provides any value except for debugging?
If not, I think its better to just warn if something is wrong and done.

librz/core/cil.c
```diff
@@ -479,6 +522,142 @@ static void core_colorify_il_statement_unicode(RzConsContext *ctx, const char *i
...
+		if (color) {
+			rz_strbuf_appendf(sb, "%s%.*s" Color_RESET, color, plen, il_stmt + prev_i);
+		} else {
+			rz_strbuf_appendf(sb, "%.*s", plen, il_stmt + prev_i);
+		}
+	}
+	rz_strbuf_appendf(sb, "\n");
+}
+
+/* @brief formatted rzil from a starting RVA to a RzPVector of DisasmTexts
```
notxvilka: We use the different style in Rizin - `\brief`, `\param`, etc

librz/core/cil.c
```diff
@@ -479,6 +522,142 @@ static void core_colorify_il_statement_unicode(RzConsContext *ctx, const char *i
...
+ * @param core - rz core
+ * @param vec - allocated rzpvector pointer with free function for disasm_text
+ * @param addr - starting address of instructions
+ * @param n_lines - number of lines to be represented as rzil
+ * @param pretty - whether add delimitter or not
+ * @param unicode - whether enable unicode formatting or not
+ * @param colorize - color the text
+ */
+
```
notxvilka, line 598: No empty line between Doxygen and the function - it breaks the renderer sometimes

notxvilka (changes requested):

NewtronReal:
I've been digging through the repo for a while now and from what i've discored for making something like a parallel to Disassembly View(which in my opinion would look much cooler than just viewing rzil for each functions) the current rz_op_chunk_iter won't be enough. I've analyzed how`rz_core_print_disassm` work. Although `print_disassm` function really precise and also works with formatting of individual lines, I felt like that much precision is not needed for RzIL view. So as of now I've made something like this.

> Moreover, if this function becomes the main point for printing RzIL, ideally, the Rizin code for corresponding command handler also should be changed to use it.

As far as what I've built it will be better to keep both these API's separate. One for printing function specific RzIL and the one implemented in this commit for printing rzils for n lines from a starting address.

> See `plf` and `plF` commands for the reference

the new work around adds support for reflines as well. Which in future could be used to expand this to the proposed [pLf](https://github.com/rizinorg/rizin/issues/3669#issue-1811439845)
Since this is a complete rework over the function should I create a new pull request or just update the existing one? 




@notxvilka @wargio @Rot127

NewtronReal:
https://github.com/user-attachments/assets/ce2c27f5-ebe7-4522-b8d5-d371757efcdf

NewtronReal:
Updated within the pr. Will added test cases soon.

librz/core/cil.c
NewtronReal: Done

librz/core/cil.c
NewtronReal: Used finish_str: and finalize:

librz/core/cil.c
NewtronReal: Removed VM

NewtronReal:
Marking it ready for review since added test cases.
Now the left work is to make this function the main function for printing rzil for both console and to vector just like how print_disassm does. And also i have to integrate the existing pfl and pFl to this function. As of now it only prints into the provided vector.

Also since the argument list was getting bigger, I made a RzCoreILPrintOptions struct for this in the rz_core.h.

NewtronReal:
Is there any workflow patches to be done. I have checked the error log multiple times i couldn't find any leaks in the logs just that the argument list generated by echo changes.diff is really long. Do i have to change anything in the code to fix it for green on New Leaks?

librz/core/cil.c
```diff
@@ -351,6 +351,17 @@ static inline void emit_span(const char *s, size_t n, const char *color) {
...
 }
 
+static inline void emit_span_to_strbuf(const char *s, size_t n, const char *color, RzStrBuf *sb) {
+	if (n < 1) {
+		return;
+	}
+	if (color) {
+		rz_strbuf_appendf(sb, "%s%.*s" Color_RESET, color, (int)n, s);
+	} else {
+		rz_strbuf_appendf(sb, "%.*s", (int)n, s);
```
wargio: we have an append function which accepts length parameter. use that

librz/core/cil.c
```diff
@@ -479,6 +522,207 @@ static void core_colorify_il_statement_unicode(RzConsContext *ctx, const char *i
...
+		const size_t utf_size = rz_utf8_decode((const ut8 *)il_stmt + i, len - i, &cp, false);
+		RzILUnicodeColorifyState state = unicode_colorify_state_next(prev_state, cp);
+		if (state != prev_state) {
+			const int plen = i - prev_i;
+			if (color) {
+				rz_strbuf_appendf(sb, "%s%.*s" Color_RESET, color, plen, il_stmt + prev_i);
+			} else {
+				rz_strbuf_appendf(sb, "%.*s", plen, il_stmt + prev_i);
+			}
+			switch (state) {
```
wargio: this switch could be moved into a function

librz/core/cil.c
```diff
@@ -479,6 +522,207 @@ static void core_colorify_il_statement_unicode(RzConsContext *ctx, const char *i
...
+			if (!rz_il_op_effect_stringify_unicode(&ctx, op.il_op, &sb)) {
+				RZ_LOG_ERROR("Failed to stringify IL at 0x%08" PFMT64x "\n", op.addr);
+				rz_strbuf_appendf(&sb, "0x%" PFMT64x "%cstringify failed", op.addr, delim);
+				goto finalize;
+			}
+		} else {
+			rz_il_op_effect_stringify(op.il_op, &sb, pretty);
+		}
+
+		il_stmt = rz_str_dup(rz_strbuf_get(&sb));
```
wargio: drain_nofree(sb)

librz/core/cil.c
```diff
@@ -479,6 +522,207 @@ static void core_colorify_il_statement_unicode(RzConsContext *ctx, const char *i
...
+		rz_strbuf_fini(&sb);
```
wargio: thos can be removed if you call the drain function

librz/core/cil.c
```diff
@@ -479,6 +522,207 @@ static void core_colorify_il_statement_unicode(RzConsContext *ctx, const char *i
...
+		if (reflines) {
+			void **iter;
+			rz_pvector_foreach (reflines, iter) {
+				RzAnalysisRefline *ref = *iter;
+				if (ref->from == vat) {
+					dst->arrow = ref->to;
+					break;
+				}
+			}
+		}
```
wargio: move into a function

librz/core/cil.c
```diff
@@ -479,6 +522,207 @@ static void core_colorify_il_statement_unicode(RzConsContext *ctx, const char *i
...
+	finalize:
+		rz_analysis_op_fini(&op);
+	finish_str:
+		dst->offset = rz_core_pava(core, current);
+		dst->text = rz_str_dup(rz_strbuf_get(&sb));
```
wargio: drain_nofree

librz/core/cil.c
```diff
@@ -479,6 +522,207 @@ static void core_colorify_il_statement_unicode(RzConsContext *ctx, const char *i
...
+		inc = inc ? inc : RZ_MAX(1, min_op_size);
+		if (!rz_pvector_push(vec, dst)) {
+			free(dst->text);
+			free(dst);
+		}
+		rz_strbuf_fini(&sb);
```
wargio: remove when you use drain_nofree

librz/core/cil.c
```diff
@@ -479,6 +522,207 @@ static void core_colorify_il_statement_unicode(RzConsContext *ctx, const char *i
...
+		core->asmqjmps_count = 0;
+		ut64 *p = realloc(core->asmqjmps, RZ_CORE_ASMQJMPS_NUM * sizeof(ut64));
+		if (p) {
+			core->asmqjmps_size = RZ_CORE_ASMQJMPS_NUM;
+			core->asmqjmps = p;
+			for (int i = 0; i < RZ_CORE_ASMQJMPS_NUM; i++) {
+				core->asmqjmps[i] = UT64_MAX;
+			}
+		}
+	}
```
wargio: this should be a RzVector of ut64 (there is a function to memset to ff)

wargio (changes requested):

librz/core/cil.c
NewtronReal: > this should be a RzVector of ut64

since this includes editing the core and all places it is used, should i do it in this pr or in a separate one?

notxvilka:
@NewtronReal you forgot to mark addressed review commets as "Resolved". Also please squash everything and rebase.

notxvilka:
@NewtronReal you forgot to squash and rebase

NewtronReal:
> @NewtronReal you forgot to squash and rebase

cleaning my repo now actually it got messed up a lot.

NewtronReal:
> @NewtronReal you forgot to squash and rebase

Done

NewtronReal:
Working on other issues rn. Just ping me if any other changes are needed.

librz/core/cil.c
```diff
@@ -420,6 +462,29 @@ RzILUnicodeColorifyState unicode_colorify_state_next(RzILUnicodeColorifyState st
 	}
 	return UNICODE_COLORIFY_STATE_IL_OP;
 }
+static const char *core_il_get_token_color(RzILUnicodeColorifyState state, const char *prev_color, RZ_NONNULL RzConsContext *ctx) {
```
Rot127, line 477: Add space above.
Also, shouldn't a similar functionb be used above in core_colorify_il_statement_to_strbuf ?

librz/core/cil.c
```diff
@@ -479,6 +527,191 @@ static void core_colorify_il_statement_unicode(RzConsContext *ctx, const char *i
...
+	}
+	size_t prev_i = 0;
+	const size_t len = strlen(il_stmt);
+	const char *color = NULL;
+	RzILUnicodeColorifyState prev_state = UNICODE_COLORIFY_STATE_DEFAULT;
+	for (size_t i = 0; i < len;) {
+		RzCodePoint cp = 0;
+		const size_t utf_size = rz_utf8_decode((const ut8 *)il_stmt + i, len - i, &cp, false);
+		RzILUnicodeColorifyState state = unicode_colorify_state_next(prev_state, cp);
+		if (state != prev_state) {
```
Rot127: Invert and continue early please

librz/core/cil.c
```diff
@@ -479,6 +527,191 @@ static void core_colorify_il_statement_unicode(RzConsContext *ctx, const char *i
...
+			}
+
+			color = core_il_get_token_color(state, color, ctx);
+
+			prev_state = state;
+			prev_i = i;
+		}
+		i += utf_size > 0 ? utf_size : 1;
+	}
+	if (prev_i < len) {
```
Rot127: Invert and return early

librz/core/cil.c
```diff
@@ -479,6 +527,191 @@ static void core_colorify_il_statement_unicode(RzConsContext *ctx, const char *i
...
+}
+
+/* \brief formatted rzil from a starting RVA to a RzPVector of DisasmTexts
+ * \param core - rz core
+ * \param vec - allocated rzpvector pointer with free function for disasm_text
+ * \param addr - starting address of instructions
+ * \param n_lines - number of lines to be represented as rzil
+ * \param pretty - whether add delimitter or not
+ * \param unicode - whether enable unicode formatting or not
+ * \param colorize - color the text
```
Rot127: Fix spelling. Also /return is missing

librz/core/cil.c
```diff
@@ -479,6 +527,191 @@ static void core_colorify_il_statement_unicode(RzConsContext *ctx, const char *i
...
+	const char *il_stmt = NULL;
+	const char delim = pretty ? '\n' : ' ';
+	RzStrBuf sb;
+	RzPVector /*<RzAnalysisRefline *>*/ *reflines;
+	RzAnalysisOp op;
+	int ops_count = 0, inc = 0, idx;
+	ut64 current = addr, vat;
+	ut8 *nbuf = NULL;
+	ut8 min_op_size = rz_analysis_archinfo(core->analysis, RZ_ANALYSIS_ARCHINFO_MIN_OP_SIZE);
+complete:
```
Rot127: Please don't introduce loops with labels. This becomes too complex. Use a proper loop instead

librz/core/cil.c
```diff
@@ -479,6 +527,191 @@ static void core_colorify_il_statement_unicode(RzConsContext *ctx, const char *i
...
+/* \brief formatted rzil from a starting RVA to a RzPVector of DisasmTexts
+ * \param core - rz core
+ * \param vec - allocated rzpvector pointer with free function for disasm_text
+ * \param addr - starting address of instructions
+ * \param n_lines - number of lines to be represented as rzil
+ * \param pretty - whether add delimitter or not
+ * \param unicode - whether enable unicode formatting or not
+ * \param colorize - color the text
+ */
+RZ_API int rz_core_il_print_rzil(RZ_NONNULL RzCore *core, ut64 addr, RZ_NONNULL ut8 *buf, int len, int n_lines, RZ_NULLABLE RzCoreILPrintOptions *options) {
```
Rot127: This function is too complex. Move stuff into static functions please.
E.g. the whole RzIL op decoding stuff can be its own.

librz/include/rz_core.h
```diff
@@ -915,6 +915,19 @@ RZ_API RZ_OWN char *rz_core_graph_to_sdb_str(RZ_NONNULL RzCore *core, RZ_NONNULL
 /*tp.c*/
 RZ_API void rz_core_analysis_type_match(RzCore *core, RzAnalysisFunction *fcn, HtUU *addr_loop_table);
 
+/* cil.c */
+/**
+ * \brief Options for printing rzil instructions.
+ */
+typedef struct rz_core_il_print_options {
+	int invbreak;
```
Rot127: Missing comment. Also, use bool and not int

librz/include/rz_core.h
```diff
@@ -915,6 +915,19 @@ RZ_API RZ_OWN char *rz_core_graph_to_sdb_str(RZ_NONNULL RzCore *core, RZ_NONNULL
 /*tp.c*/
 RZ_API void rz_core_analysis_type_match(RzCore *core, RzAnalysisFunction *fcn, HtUU *addr_loop_table);
 
+/* cil.c */
```
Rot127, line 928:
```suggestion
```

librz/include/rz_core.h
```diff
@@ -954,6 +967,7 @@ RZ_API RzList /*<RzCoreAsmHit *>*/ *rz_core_asm_bwdisassemble(RzCore *core, ut64
 RZ_API RzList /*<RzCoreAsmHit *>*/ *rz_core_asm_back_disassemble_instr(RzCore *core, ut64 addr, int len, ut32 hit_count, ut32 extra_padding);
 RZ_API RzList /*<RzCoreAsmHit *>*/ *rz_core_asm_back_disassemble_byte(RzCore *core, ut64 addr, int len, ut32 hit_count, ut32 extra_padding);
 RZ_API ut32 rz_core_asm_bwdis_len(RzCore *core, int *len, ut64 *start_addr, ut32 l);
+RZ_API int rz_core_il_print_rzil(RZ_NONNULL RzCore *core, ut64 addr, RZ_NONNULL ut8 *buf, int len, int n_lines, RZ_NULLABLE RzCoreILPrintOptions *options);
```
Rot127: `n_lines` should be unsigned. `len` as well. Use `size_t`

Rot127 (changes requested):

notxvilka:
@NewtronReal what is the status of it? Please rebase on top of the latest `dev`

NewtronReal:
Will be working on the conflict before rebasing. Will be done by tmrw.

librz/core/cil.c
NewtronReal: it was only for unicode which was added later. for readability we can add 

```
if (ch == '(' || ch == ')') {
			emit_span_to_strbuf(il_stmt + prev, i - prev, color, sb);
			rz_strbuf_appendf(sb, "%s%c" Color_RESET, ctx->pal.meta, ch);
			prev = i + 1;
			color = (ch == '(') ? ctx->pal.flow : NULL;
		} else if (ch == ' ') {
			emit_span_to_strbuf(il_stmt + prev, i - prev, color, sb);
			rz_strbuf_appendf(sb, " ");
			prev = i + 1;
			color = NULL;
		} else if (i == prev && prev > 0 && il_stmt[prev - 1] == ' ') {
			color = IS_DIGIT(ch) ? ctx->pal.num : ctx->pal.comment;
		}
```
this as a separate function

librz/include/rz_core.h
NewtronReal: Done

librz/include/rz_core.h
NewtronReal: Removed invbreak. It was not needed for il processing for now.

librz/core/cil.c
NewtronReal: Added an il state struct and integrated into several functions.

librz/core/cil.c
NewtronReal: Used while loop with il_state_complete condition

librz/core/cil.c
NewtronReal: Oops, missed it will be doing it asap.

librz/core/cil.c
NewtronReal: Done

librz/core/cil.c
NewtronReal: Done.

notxvilka:
Update the PR description please

wargio (commented):
while this is ok to me, wouldn't make more sense to just create a new function with the IL info and use that to populate the colored string instead of parsing the whole lisp-like output? in this way you could declare a structure with the stringify info (like params etc..) and colors. so when cutter has to print this stuff can do it directly natively and rizin can use the ANSI escape colors

NewtronReal:
> while this is ok to me, wouldn't make more sense to just create a new function with the IL info and use that to populate the colored string instead of parsing the whole lisp-like output? in this way you could declare a structure with the stringify info (like params etc..) and colors. so when cutter has to print this stuff can do it directly natively and rizin can use the ANSI escape colors

@wargio 
Since unicode and normal stringification were already there i just tried to use that in this api as well. For semantics shall i just create a struct called `RzAnalysisILText` and then pass op.il_op or some new structure like RzILInfo as one of it's params that saves the both normal and unicodification implementations in the cutter codebase. And in future this could serve as the unified api for printing all types of formattings including enriched.


EDIT

Ah now that I think again it would be much better to parse the info struct in cutter so that in future if we want some implementation like token aware tooltip.

wargio:
> Ah now that I think again it would be much better to parse the info struct in cutter so that in future if we want some implementation like token aware tooltip.

That would be the optimal solution

wargio (approved):

notxvilka (approved):

PR #6268, opened 2026-04-21
test: make darwin dpa attach target explicit

ret2libc:

**Detailed description**

...

**Test plan**

...

**Closing issues**

...


test/db/archos/darwin-arm64/dbg
```diff
@@ -41,10 +41,10 @@ RUN
 NAME=dpa issue
 FILE==
 CMDS=<<EOF
-!sh -c "./bins/mach0/sleep-macos-arm64 100 >/dev/null 2>&1 &"
-dpa $(!pgrep sleep-macos-arm64)
+!sh -c 'pidfile="${TMPDIR:-/tmp}/rz-dpa-$PPID.pid"; ./bins/mach0/sleep-macos-arm64 100 >/dev/null 2>&1 & pid=$!; echo "$pid" > "$pidfile"; while ! kill -0 "$pid" 2>/dev/null; do sleep 0.01; done'
+dpa $(!sh -c 'cat "${TMPDIR:-/tmp}/rz-dpa-$PPID.pid"')
 pi 3 @ pc
-!pkill sleep-macos-arm64
+!sh -c 'pidfile="${TMPDIR:-/tmp}/rz-dpa-$PPID.pid"; kill "$(cat "$pidfile")" 2>/dev/null; rm -f "$pidfile"'
```
wargio, line 47: instead of doing this, why not create a sleep binary which outputs its own pid to stdout and pipe that to a local file, then just read that value and attach to it.
Or just make a bash script, execute it from rizin and read the pid file.

ret2libc: Tbh I feel like it would be a waste of time . It's just a test and it would achieve the same purpose . Moreover this approach works in general and doesn't require any bin changes. Last, what's the issue with this approach? This is anyway Darwin specific.

wargio: that is true. but we will face similar issues on other platform, so having a bash script doing the heavly lifting to produce the correct pid and attach to it could be beneficial.  probably worth opening an issue later

ret2libc:
Apparently that was not really the problem. It might be a race on the darwin debugger itself.

PR #6287, opened 2026-04-25
Lifting the base (integer) extension of RISC-V, some compressed variants of integer instructions, and some multiplication/division instructions

moste00:

**Detailed description**

...

**Test plan**

...

**Closing issues**

**PR Stack Structure**
-----------------------------------------------------------------------------------------------------------------------------------------------------------

This PR is the first of a  4-PR stack that lifts the IMAFD RISC-V archiecture plus some privileged instructions. 

The following is a table for ease of navigation from any PR to any other, please filter by the commit(s) mentioned in the table when reviewing the diff in the "Files Changed" tab to avoid the noise of the base branches.

|Extension| PR | Commits To Select in "Files Changed" | State (Open, Merged, Draft, Closed)
| :---: | :----: | :---: | :---:
| I     | https://github.com/rizinorg/rizin/pull/6287 < (THIS PR) | No filtering necessary | O
| M   | https://github.com/rizinorg/rizin/pull/6364 | "implement the rest of the multiplication extension"|D
| A    | https://github.com/rizinorg/rizin/pull/6373 | "implement the atomic extension trivially by ignoring concurrency guards" | D
| F/D| https://github.com/rizinorg/rizin/pull/6374 | "add floating and double instructions, and add privlieged ISA instructions like reading CSRs to implement fflags reading and writing" | D

Because it's painful to maintain this stack manually with no Github or other tooling support, I won't grow it any further until one of those 4 PRs gets merged.

Depends-on: https://github.com/rizinorg/rz-tracetest/pull/29
Depends-on: https://github.com/BinaryAnalysisPlatform/qemu/pull/44


moste00:
@notxvilka for visiblity, not ready for merge.

Next: Adapt the code to work for 64-bit arch as well as 32-bit, and test against machine traces from BAP QEMU.

moste00:
@notxvilka @Rot127 @wargio this is ready for a look btw, it's mostly complete but I'm fighting an issue with  some compressed instruction decoding (capstone bug)

librz/arch/isa/riscv/riscv_il.c
```diff
@@ -0,0 +1,142 @@
...
+	USE_LIFTER(c_sub, C_SUB),
+	USE_LIFTER(c_and, C_AND),
+	USE_LIFTER(c_xor, C_XOR),
+	USE_LIFTER(c_addw, C_ADDW),
+	USE_LIFTER(c_addiw, C_ADDIW),
+	USE_LIFTER(c_subw, C_SUBW),
+};
+
+RZ_OWN RZ_IPI RzILOpEffect *
+rz_riscv_lift_instr(RZ_BORROW RZ_NONNULL RzAnalysis *analysis, RZ_NONNULL RzAnalysisOp *op, RZ_NONNULL cs_insn *insn, ut64 current_addr, int size) {
```
wargio:
```suggestion
RZ_OWN RZ_IPI RzILOpEffect *rz_riscv_lift_instr(RZ_BORROW RZ_NONNULL RzAnalysis *analysis,
	RZ_NONNULL RzAnalysisOp *op,
	RZ_NONNULL cs_insn *insn,
	ut64 current_addr,
	size_t size) {
```

Please avoid using `int` for any sort of size variable. always `size_t`

librz/arch/isa/riscv/riscv_il.h
```diff
@@ -0,0 +1,20 @@
...
+#include <capstone/capstone.h>
+#include <capstone/riscv.h>
+
+#include <rz_types.h>
+#include <rz_analysis.h>
+typedef RzILOpEffect *(*RiscvInstructionLifter)(RZ_BORROW RZ_NONNULL RzAnalysis *analysis, RZ_NONNULL RzAnalysisOp *op, RZ_NONNULL cs_insn *insn, ut64 current_addr, int size);
+
+RZ_OWN RZ_IPI RzILOpEffect *
+rz_riscv_lift_instr(RZ_BORROW RZ_NONNULL RzAnalysis *analysis,
+	RZ_NONNULL RzAnalysisOp *op, RZ_NONNULL cs_insn *insn, ut64 current_addr, int size);
```
wargio:
```suggestion
RZ_OWN RZ_IPI RzILOpEffect *rz_riscv_lift_instr(RZ_BORROW RZ_NONNULL RzAnalysis *analysis,
	RZ_NONNULL RzAnalysisOp *op,
	RZ_NONNULL cs_insn *insn,
	ut64 current_addr,
	size_t size);
```

librz/arch/isa/riscv/riscv_il_base.h
```diff
@@ -0,0 +1,220 @@
...
+			RZ_LOG_ERROR("operands[%d].type: %d\n", _i, insn->detail->riscv.operands[_i].type); \
+			if (insn->detail->riscv.operands[_i].type == RISCV_OP_REG) { \
+				RZ_LOG_ERROR("  REG = %d\n", insn->detail->riscv.operands[_i].reg); \
+			} else if (insn->detail->riscv.operands[_i].type == RISCV_OP_IMM) { \
+				RZ_LOG_ERROR("  IMM = 0x%" PFMT64x "\n", (ut64)insn->detail->riscv.operands[_i].imm); \
+			} else if (insn->detail->riscv.operands[_i].type == RISCV_OP_MEM) { \
+				RZ_LOG_ERROR("  MEM base = %d, disp = 0x%" PFMT64x "\n", insn->detail->riscv.operands[_i].mem.base, (ut64)insn->detail->riscv.operands[_i].mem.disp); \
+			} \
+		} \
+		exit(-1); \
```
wargio: never do this. do a graceful return.

librz/arch/isa/riscv/riscv_il_base.h
```diff
@@ -0,0 +1,220 @@
...
+	}
+
+#define REQUIRE_64_BIT(analysis) \
+	if (analysis->bits != 64) { \
+		RZ_LOG_ERROR("[%s (%d)] Expected 64-bit analysis, found %d bits\n", insn->mnemonic, insn->id, analysis->bits); \
+		exit(-1); \
```
wargio: never do this. do a graceful return.

librz/arch/p/analysis/analysis_riscv_cs.c
```diff
@@ -2148,7 +2155,7 @@ static char *get_reg_profile(RzAnalysis *analysis) {
 			"gpr	ft9	.64	488	0\n" // =f29
 			"gpr	ft10	.64	496	0\n" // =f30
 			"gpr	ft11	.64	504	0\n" // =f31
-			"gpr	fcsr	.32	512	0\n"
+			"gpr	fcsr	.64	512	0\n"
```
wargio, line 2240: According to [this](https://docs.riscv.org/reference/isa/unpriv/f-st-ext.html#21-1-2-floating-point-control-and-status-register) is 32 bit wide.

moste00: Yes, but QEMU (or the BAP fork we're using) probably models it internally as 64-bit nonethelss as a matter of policy for all CSRs, because `rz-tracetest` keeps reporting width mismatches if you leave it as 32-bit. When you model it as 64-bit the errors disappear.

librz/arch/isa/riscv/riscv_il_base.h
moste00: Acknowledged, I did it so the error will stand out from a sea of logging in `rz-tracetest`, I will change it to a return before I merge.

moste00:
NOTE-1: I initially intended this to only cover integer (rv32I and rv64I) subsets of the ISA, but in order to enable real-world testing on gcc-compiled binaries I had to support compressed subsets, so this PR is less orthogonal than I would like, there is even a couple of multiplication and division instructions in there. 

Sorry. This is annoying but trying to link against libc from non-IMAFD binaries is even more annoying. I will try to keep future extensions as self-contained as possible.

-----------------------------------------------------------------------------------------------------------------------------------------------------------

NOTE-2:
The architecture of the lifter is that each unique instruction has exactly one lifter function named after it. This enables a single-level dispatch table in the entrypoint `riscv_il.c` where every instruction is listed by its capstone ID next to its lifter function. (As opposed to (A) two-level dispatch where a first-stage table maps a family of instructions, say all branches or all I-type, to a single lifting function then a switch inside that function decides the logic for each individual instruction. (B) Or  a non-modular single-level giant switch over every RISC-V instruction in the same function, which would be bad for readability and correctness).

In order to modularize, each ISA extension is seperated into a different `.h` file. In order to avoid boilerplate, the common prelude file `riscv_il_base.h` defines several macros that allow you to define the lifter function concisely by specifying only 3 things:

1- The lifting function name
2- The function decoder, the decoder is the name of a macro that exctracts source and destination variables from the capstone operands
3- The logic of the instruction as a RzIL expression.

For example, `DEFINE_LIFTER(addi, DECODE_RD_RS_IMM, ADD(rs, imm))`

1- `addi` is the name of the lifter function, it's used later in the master dispatch table in `riscv_il.c`
2- `DECODE_RD_RS_IMM` is the name of the decoder macro which will extract rd, rs, and imm from the operand array in capstone
3- `ADD(rs, imm)` is the RzIL expression which will be inserted into an assignment to the destination register by the `DEFINE_LIFTER` macro (unless the destination register is `x0`, in which case the instruction is a NOP. This is handled by register-setter and the register-getter macros, `RISCV_GET_REG` and RISCV_SET_REG`)

This architecture allows arbitrary cross-sharing of decoding logic with any instruction instead of hardcoding it into a specific extension or instruction group. There is also `DEFINE_ALIAS_LIFTER` which will define a static const pointer that points to the lifter function of another instruction.

librz/arch/isa/riscv/riscv_il_base.h
```diff
@@ -0,0 +1,199 @@
...
+			if (insn->detail->riscv.operands[_i].type == RISCV_OP_REG) { \
+				RZ_LOG_ERROR("  REG = %d\n", insn->detail->riscv.operands[_i].reg); \
+			} else if (insn->detail->riscv.operands[_i].type == RISCV_OP_IMM) { \
+				RZ_LOG_ERROR("  IMM = 0x%" PFMT64x "\n", (ut64)insn->detail->riscv.operands[_i].imm); \
+			} else if (insn->detail->riscv.operands[_i].type == RISCV_OP_MEM) { \
+				RZ_LOG_ERROR("  MEM base = %d, disp = 0x%" PFMT64x "\n", insn->detail->riscv.operands[_i].mem.base, (ut64)insn->detail->riscv.operands[_i].mem.disp); \
+			} \
+		} \
+		return NULL; \
+	}
```
Rot127, line 109: I understand that it is at first nice to define everything in macros and then use them to generate code.
But it is horrible to debug, hides complexity and is not very maintainable.
Granted, once it is lifted it won't change much. So if the others are ok with it, I would accept it.

But I highly prefer to have more code in functions.
This block here for example is most definitely too long.

Rot127 (changes requested):
Sorry, I checked it too late.

librz/arch/isa/riscv/riscv_il_base.h
moste00: I can refactor this one into a function, it's not important for it to be a macro unlike the `DECODE_*` macros that must be macros in order to mix-and-match them with any instruction whatsoever with 0 runtime cost.

I just didn't see it as important, it's a reporting utility that only activates in debug builds to check on capstone operands and what we expect them to be.

Rot127: Please make it a function then.

moste00:
@Rot127 @wargio Ping

librz/arch/isa/riscv/riscv_il.c
```diff
@@ -0,0 +1,146 @@
...
+	rz_return_val_if_fail(analysis && op && insn, NULL);
+
+	if (insn->id == RISCV_INS_INVALID) {
+		return NULL;
+	}
+	if (insn->id >= RISCV_INS_ENDING) {
+		RZ_LOG_ERROR("Invalid RISC-V instruction id %u (0x%08x)\n", insn->id, rz_read_le32(insn->bytes));
+		return NULL;
+	}
+	if (insn->id < RZ_ARRAY_SIZE(riscv_lifters)) {
```
Rot127: invert condition, return early

librz/arch/isa/riscv/riscv_il_integer.h
Rot127, line 1: Function definitions of that size should never be in a header.

librz/arch/isa/riscv/riscv_il_base.h
```diff
@@ -0,0 +1,226 @@
...
+	def_lifter(name1, __VA_ARGS__) \
+		def_lifter(name2, __VA_ARGS__)
+
+#define THRICE_FOR(name1, name2, name3, def_lifter, ...) \
+	TWICE_FOR(name1, name2, def_lifter, __VA_ARGS__) \
+	def_lifter(name3, __VA_ARGS__)
+
+#define FOR_4(name1, name2, name3, name4, def_lifter, ...) \
+	TWICE_FOR(name1, name2, def_lifter, __VA_ARGS__) \
+	TWICE_FOR(name3, name4, def_lifter, __VA_ARGS__)
```
Rot127:
```suggestion
```

Please simply repeat.

test/db/asm/riscv_32
Rot127, line 1: Please also add tests for the 64bit versions.

librz/arch/isa/riscv/riscv_il_compressed.h
Rot127, line 1: Also here, no function definition in the header.

librz/arch/isa/riscv/riscv_il_base.h
```diff
@@ -0,0 +1,226 @@
...
+	REQUIRE_3OPS(RISCV_OP_REG, RISCV_OP_REG, RISCV_OP_IMM); \
+	uint32_t rd = insn->detail->riscv.operands[0].reg; \
+	RzILOpBitVector *rs = CAST(32, IL_FALSE, riscv_il_get_reg(analysis->bits, insn->detail->riscv.operands[1].reg)); \
+	RzILOpBitVector *imm = SN(32, insn->detail->riscv.operands[2].imm);
+
+#define DECODE_RD_IMM_TRUNCATE32(analysis, insn) \
+	REQUIRE_64_BIT(analysis); \
+	REQUIRE_2OPS(RISCV_OP_REG, RISCV_OP_IMM); \
+	uint32_t rd = insn->detail->riscv.operands[0].reg; \
+	RzILOpBitVector *imm = SN(32, insn->detail->riscv.operands[1].imm);
```
Rot127, line 209: All the immediates are defined here as signed. But the ISA uses often enough unsigned (e.g. for AND, OR etc).
That can be a source of subtle bugs.

librz/arch/isa/riscv/riscv_il_integer_reg_names.h
```diff
@@ -0,0 +1,53 @@
...
+	/* x27 */ "s11",
+	/* x28 */ "t3",
+	/* x29 */ "t4",
+	/* x30 */ "t5",
+	/* x31 */ "t6",
+};
+
+static inline const char *riscv_integer_reg_name(uint32_t reg) {
+	int idx = (int)reg - RISCV_REG_X0;
+	if (idx < 0 || idx >= 32) {
```
Rot127: RZ_ARRAY_SIZE

Rot127: Also, that should warn here.

Rot127 (changes requested):
Please finish those ones before continuing to work on the others.

There are still a few failing and some leaks.

librz/arch/isa/riscv/riscv_il_base.h
```diff
@@ -0,0 +1,226 @@
...
+		if (insn->detail->riscv.operands[i].type == RISCV_OP_REG) {
+			RZ_LOG_ERROR("  REG = %d\n", insn->detail->riscv.operands[i].reg);
+		} else if (insn->detail->riscv.operands[i].type == RISCV_OP_IMM) {
+			RZ_LOG_ERROR("  IMM = 0x%" PFMT64x "\n", (ut64)insn->detail->riscv.operands[i].imm);
+		} else if (insn->detail->riscv.operands[i].type == RISCV_OP_MEM) {
+			RZ_LOG_ERROR("  MEM base = %d, disp = 0x%" PFMT64x "\n", insn->detail->riscv.operands[i].mem.base, (ut64)insn->detail->riscv.operands[i].mem.disp);
+		}
+	}
+}
+
```
Rot127, line 83: Also please remove this logging function before merging.

Rot127:
Ouh, and please add a test which prints instructions with `aoi` and check no errors are thrown.
For 32 and 64 bit binaries.

Because there are already a few:

```
ERROR: op_str: s6, 0x10
ERROR: need_effective_addr: 0
ERROR: op_count: 2
ERROR: operands[0].type: 1
ERROR:   REG = 64
ERROR: operands[1].type: 2
ERROR:   IMM = 0x10
Invalid instruction of lifting not implemented.
```

librz/arch/p/analysis/analysis_riscv_cs.c
Rot127: You should be able to handle those mismatches in rz-tracetest.
There are functions for these kind of cases IIRC.

librz/arch/isa/riscv/riscv_il_base.h
moste00: Hmmm, I don't think `ANDI` and `ORI` are unsigned, actually, almost all immediate-using instruction use sign-extension, not unsigned extension.

The only exceptions are:

1- fences, implemented as `NOP`s here

2- Shift instructions, but their 5-bit immediates are tiny and sign-extending them is always correct

3- CSR instructions, same situation as shift instructions

4- LUI and AUIPC, their immediates are 12-bit and 20-bit respectively, and sign-extending them also won't cause any problems, since capstone presents them in their unshifted form.

Other than that, I remember everything else was signed.

moste00: I will go over every one again and review the spec but I think it's fine.

Rot127: Sorry, my bad. I missed the note in the ISA about it. All good.

Rot127 (changes requested):
The branch instructions leak:
```
==34077==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 40 byte(s) in 1 object(s) allocated from:
    #0 0x75d3c0ae68a3 in calloc (/lib64/libasan.so.8+0xe68a3) (BuildId: 25975f766867e9e604dc5a71a8befeaed3301942)
    #1 0x75d3bb244749 in rz_il_op_new_bitv_from_ut64 ../librz/il/il_opcodes.c:180
    #2 0x75d3b39571d3 in riscv_il_get_reg ../librz/arch/isa/riscv/riscv_il_base.h:17
    #3 0x75d3b3959a58 in rz_riscv_lift_c_beqz ../librz/arch/isa/riscv/riscv_il_compressed.c:69
    #4 0x75d3b393d8ad in rz_riscv_lift_instr ../librz/arch/isa/riscv/riscv_il.c:152
    #5 0x75d3b3f3d9c9 in analyze_op ../librz/arch/p/analysis/analysis_riscv_cs.c:1929
    #6 0x75d3b408cad1 in rz_analysis_op ../librz/arch/op.c:126
    #7 0x75d3c067d960 in print_disassembly_output ../librz/main/rz-asm.c:331
    #8 0x75d3c068393e in rz_main_rz_asm ../librz/main/rz-asm.c:896
    #9 0x0000004007ad in main ../binrz/rz-asm/rz-asm.c:8
    #10 0x75d3bcc105b4 in __libc_start_call_main (/lib64/libc.so.6+0x35b4) (BuildId: 88c03969b4319fc0140734fe863825ec4a1cdb0d)
    #11 0x75d3bcc10667 in __libc_start_main@@GLIBC_2.34 (/lib64/libc.so.6+0x3667) (BuildId: 88c03969b4319fc0140734fe863825ec4a1cdb0d)
    #12 0x0000004006c4 in _start (/home/user/.local/bin/rz-asm+0x4006c4) (BuildId: 7e9c8cb46628eb0719490f7b5b48ea6d6bb3e8b9)

Indirect leak of 32 byte(s) in 1 object(s) allocated from:
    #0 0x75d3c0ae68a3 in calloc (/lib64/libasan.so.8+0xe68a3) (BuildId: 25975f766867e9e604dc5a71a8befeaed3301942)
    #1 0x75d3bd61a391 in rz_bv_new ../librz/util/bitvector.c:101
    #2 0x75d3bd627c56 in rz_bv_new_from_ut64 ../librz/util/bitvector.c:1747
    #3 0x75d3bb244725 in rz_il_op_new_bitv_from_ut64 ../librz/il/il_opcodes.c:176
    #4 0x75d3b39571d3 in riscv_il_get_reg ../librz/arch/isa/riscv/riscv_il_base.h:17
    #5 0x75d3b3959a58 in rz_riscv_lift_c_beqz ../librz/arch/isa/riscv/riscv_il_compressed.c:69
    #6 0x75d3b393d8ad in rz_riscv_lift_instr ../librz/arch/isa/riscv/riscv_il.c:152
    #7 0x75d3b3f3d9c9 in analyze_op ../librz/arch/p/analysis/analysis_riscv_cs.c:1929
    #8 0x75d3b408cad1 in rz_analysis_op ../librz/arch/op.c:126
    #9 0x75d3c067d960 in print_disassembly_output ../librz/main/rz-asm.c:331
    #10 0x75d3c068393e in rz_main_rz_asm ../librz/main/rz-asm.c:896
    #11 0x0000004007ad in main ../binrz/rz-asm/rz-asm.c:8
    #12 0x75d3bcc105b4 in __libc_start_call_main (/lib64/libc.so.6+0x35b4) (BuildId: 88c03969b4319fc0140734fe863825ec4a1cdb0d)
    #13 0x75d3bcc10667 in __libc_start_main@@GLIBC_2.34 (/lib64/libc.so.6+0x3667) (BuildId: 88c03969b4319fc0140734fe863825ec4a1cdb0d)
    #14 0x0000004006c4 in _start (/home/user/.local/bin/rz-asm+0x4006c4) (BuildId: 7e9c8cb46628eb0719490f7b5b48ea6d6bb3e8b9)

SUMMARY: AddressSanitizer: 72 byte(s) leaked in 2 allocation(s).
```

Also many others, please run the asm tests with LSAN

Rot127:
Also:

```
rz-test test/db/asm/riscv_64
Running rz-test on linux-x64
Running from /home/user/repos/rizin_wt_I/test
Loaded 308 tests.
[XX] 00h00m03s817ms /home/user/repos/rizin_wt_I/test/db/asm/riscv_64 <asm> jal 2
-- <asm> jal 2 <--- 0920 ---> <IL>
-- disassembly
--- expected
+++ actual
@@ -1,1 +1,1 @@
-jal 2
+invalid

-- IL
--- expected
+++ actual
@@ -1,1 +1,0 @@
-(seq (set ra (bv 32 0x2)) (jmp (bv 32 0x2)))


```

Rot127:
@moste00 I converted it to a draft for now. Please mark it as ready for review when it is green.

moste00:
@Rot127 can you take one more look ? 

~By the way, all CI failures are due to M68K failures from updating the capstone version, I tried both the tip of capstone and the commit before it, both produce failures in the CI here.~

moste00:
@Rot127 ping

moste00:
@Rot127 I think it's ready.

librz/arch/isa/riscv/riscv_il.c
```diff
@@ -0,0 +1,168 @@
...
+#include "riscv_il_m.h"
+#include "riscv_il_compressed.h"
+
+static void label_ecall(RzILVM *vm, RzILOpEffect *op) {
+	// stub: ecall is handled at the analysis layer
+}
+
+static void label_ebreak(RzILVM *vm, RzILOpEffect *op) {
+	// stub: ebreak is handled at the analysis layer
+}
```
Rot127, line 16: What means analysis layer here?

It doesn't need to be handled, but please add some logs.

librz/arch/isa/riscv/riscv_il.c
```diff
@@ -0,0 +1,168 @@
...
+	if (insn->id == RISCV_INS_INVALID) {
+		return NULL;
+	}
+	if (insn->id >= RISCV_INS_ENDING) {
+		RZ_LOG_ERROR("Invalid RISC-V instruction id %u (0x%08x)\n", insn->id, rz_read_le32(insn->bytes));
+		return NULL;
+	}
+	if (insn->id >= RZ_ARRAY_SIZE(riscv_lifters)) {
+		return NULL;
+	}
```
Rot127, line 146: Merge this a single rz_return_val_if_fail please.

librz/arch/isa/riscv/riscv_il_base.h
```diff
@@ -0,0 +1,195 @@
+// SPDX-FileCopyrightText: 2026 Mostafa Mahmoud <ubermenchun@gmail.com>
+// SPDX-License-Identifier: BSD-3-Clause
+
+#ifndef RISCV_IL_BASE_H
+#define RISCV_IL_BASE_H
+
+#include "riscv.h"
+#include "rz_types.h"
```
Rot127, line 8:
```suggestion
#include <rz_types.h>
```

librz/arch/isa/riscv/riscv_il_base.h
```diff
@@ -0,0 +1,195 @@
...
+	return reg != RISCV_REG_X0 ? VARG(riscv_integer_reg_name(reg)) : UN(bits, 0);
+}
+
+static inline RzILOpEffect *riscv_il_set_reg(uint32_t reg, RZ_OWN RZ_NONNULL RzILOpBitVector *value) {
+	if (reg != RISCV_REG_X0) {
+		return SETG(riscv_integer_reg_name(reg), value);
+	}
+	// assigning to x0 is a no-op, so the value will never be referenced in the IL tree
+	rz_il_op_pure_free(value);
+	return NOP();
```
Rot127, line 25:
```suggestion
	return EMPTY();
```

NOP increments the PC

librz/arch/isa/riscv/riscv_il_base.h
```diff
@@ -0,0 +1,195 @@
...
+
+#define DEFINE_LIFTER_WITH_EFFECT(name, decoder, effect) \
+	RzILOpEffect *rz_riscv_lift_##name(RZ_BORROW RZ_NONNULL RzAnalysis *analysis, \
+		RZ_NONNULL RzAnalysisOp *op, RZ_NONNULL cs_insn *insn, ut64 current_addr, size_t size) { \
+		decoder(analysis, insn); \
+		return effect; \
+	}
+
+// oneway jumps are those that don't have a destination register
+#define DEFINE_LIFTER_FOR_ONEWAY_JUMP DEFINE_LIFTER_WITH_EFFECT
```
Rot127, line 53: Those are commonly called "direct jumps".
Indirect ones are the jumps which get there target address from a register.

librz/arch/isa/riscv/riscv_il_base.h
```diff
@@ -0,0 +1,195 @@
...
+
+// A more intuitive definition would be:
+//   static const RiscvInstructionLifter rz_riscv_lift_##alias = ...;
+// but C (unlike C++) does not treat const-qualified variables as constant expressions
+// (C11 §6.6), so they cannot appear in a static array initializer. Most compilers
+// accept it tolerantly, but TCC enforces the standard strictly and rejects it.
+// A forced-inline wrapper is an alternative for 0-indirection aliasing.
```
Rot127, line 60:
```suggestion
```

It's ok. Makes sense as it is.

librz/arch/isa/riscv/riscv_il_base.h
```diff
@@ -0,0 +1,195 @@
...
+		return true;
+	}
+	return false;
+}
+
+static inline bool riscv_il_require_64_bit(RZ_NONNULL RzAnalysis *analysis, RZ_NONNULL cs_insn *insn) {
+	if (rz_analysis_get_bits(analysis) == 64) {
+		return true;
+	}
+	// RZ_LOG_ERROR("[%s (%d)] Expected 64-bit analysis, found %d bits\n", insn->mnemonic, insn->id, rz_analysis_get_bits(analysis));
```
Rot127, line 78:
```suggestion

```

librz/arch/isa/riscv/riscv_il_base.h
```diff
@@ -0,0 +1,195 @@
...
+// accept it tolerantly, but TCC enforces the standard strictly and rejects it.
+// A forced-inline wrapper is an alternative for 0-indirection aliasing.
+#define DEFINE_ALIAS_LIFTER(alias, name) \
+	RzILOpEffect *rz_riscv_lift_##alias(RZ_BORROW RZ_NONNULL RzAnalysis *analysis, RZ_NONNULL RzAnalysisOp *op, RZ_NONNULL cs_insn *insn, ut64 current_addr, size_t size) { \
+		return rz_riscv_lift_##name(analysis, op, insn, current_addr, size); \
+	}
+
+#if RZ_CHECKS_LEVEL > 0
+static inline bool riscv_il_require_op(RZ_NONNULL cs_insn *insn, ut64 current_addr, int idx, int type, const char *type_name) {
+	if (insn->detail->riscv.operands[idx].type == type && insn->detail->riscv.operands[idx].type != RISCV_OP_INVALID) {
```
Rot127, line 68: Please add a macro to access the operands.
Something like `INSN_OP(idx) (insn->detail->riscv.operands[idx])`

Then replace the pattern everywhere. It makes the code way easier to read.

librz/arch/isa/riscv/riscv_il_integer_reg_names.h
```diff
@@ -0,0 +1,54 @@
...
+	/* x25 */ "s9",
+	/* x26 */ "s10",
+	/* x27 */ "s11",
+	/* x28 */ "t3",
+	/* x29 */ "t4",
+	/* x30 */ "t5",
+	/* x31 */ "t6",
+};
+
+static inline const char *riscv_integer_reg_name(uint32_t reg) {
```
Rot127, line 45: Is there an issue with the capstone register name getter?

librz/arch/isa/riscv/riscv_il_m.c
```diff
@@ -0,0 +1,41 @@
...
+	return UN(bits, UT64_MAX);
+}
+
+static inline RzILOpBitVector *riscv_il_sdiv(ut32 bits, RZ_OWN RzILOpBitVector *rs1, RZ_OWN RzILOpBitVector *rs2) {
+	return ITE(IS_ZERO(DUP(rs2)), riscv_il_div_by_zero_result(bits), SDIV(rs1, rs2));
+}
+
+static inline RzILOpBitVector *riscv_il_divu(ut32 bits, RZ_OWN RzILOpBitVector *rs1, RZ_OWN RzILOpBitVector *rs2) {
+	return ITE(IS_ZERO(DUP(rs2)), riscv_il_div_by_zero_result(bits), DIV(rs1, rs2));
+}
```
Rot127, line 22: Did we discuss this somewhere?
Is this actually defined in the RISC-V ISA?

librz/arch/isa/riscv/riscv_il_base.h
Rot127: Ah, I've seen you added riscv_operand().
Please use this one then.

librz/arch/p/analysis/analysis_riscv_cs.c
```diff
@@ -69,6 +28,134 @@ static RzRegItem *riscv_reg_get(const RzReg *reg, const char *name, int type) {
 	return rz_reg_get(reg, name, type);
 }
 
+static cs_riscv_op *riscv_operand(cs_insn *insn, ut8 n) {
```
Rot127, line 31:
```suggestion
static const cs_riscv_op *riscv_operand(cs_insn *insn, ut8 n) {
```

librz/arch/p/analysis/analysis_riscv_cs.c
```diff
@@ -69,6 +28,134 @@ static RzRegItem *riscv_reg_get(const RzReg *reg, const char *name, int type) {
...
+static cs_riscv_op *riscv_memory_operand(cs_insn *insn, ut8 n) {
+	rz_return_val_if_fail(riscv_operand_is(insn, n, RISCV_OP_MEM), NULL);
+	return riscv_operand(insn, n);
+}
+
+static bool riscv_memory_operand_is_based_on(cs_insn *insn, ut8 n, ut32 reg) {
+	return riscv_operand_is(insn, n, RISCV_OP_MEM) && riscv_memory_operand(insn, n)->mem.base == reg;
+}
+
+static cs_riscv_op *find_memory_reference(cs_insn *insn) {
```
Rot127, line 64:
```suggestion
static const cs_riscv_op *find_memory_reference(cs_insn *insn) {
```

librz/arch/p/analysis/analysis_riscv_cs.c
```diff
@@ -69,6 +28,134 @@ static RzRegItem *riscv_reg_get(const RzReg *reg, const char *name, int type) {
...
+	rz_return_val_if_fail(riscv_operand_is(insn, n, RISCV_OP_REG), RISCV_REG_INVALID);
+	return riscv_operand(insn, n)->reg;
+}
+
+static st64 riscv_imm(cs_insn *insn, ut8 n) {
+	rz_return_val_if_fail(riscv_operand_is(insn, n, RISCV_OP_IMM), INT64_MAX);
+	return riscv_operand(insn, n)->imm;
+}
+
+static cs_riscv_op *riscv_memory_operand(cs_insn *insn, ut8 n) {
```
Rot127, line 55:
```suggestion
static const cs_riscv_op *riscv_memory_operand(cs_insn *insn, ut8 n) {
```

librz/arch/p/analysis/analysis_riscv_cs.c
```diff
@@ -119,13 +206,14 @@ static int parse_reg_name(RzRegItem *reg, csh handle, cs_insn *insn, int reg_num
 	if (!reg) {
 		return -1;
 	}
-	switch (OPERAND(reg_num).type) {
+	cs_riscv_op *op = riscv_operand(insn, reg_num);
+	switch (op->type) {
 	case RISCV_OP_REG:
-		reg->name = (char *)cs_reg_name(handle, OPERAND(reg_num).reg);
+		reg->name = (char *)cs_reg_name(handle, op->reg);
```
Rot127, line 212:
```suggestion
		reg->name = cs_reg_name(handle, op->reg);
```

Not your change, but I am surprised it doesn't crash.
Probably some atrocity again of mixing static with heap pointers.
Please copy the string. If it shows a leak, be so kind and fix it.

librz/arch/p/analysis/analysis_riscv_cs.c
```diff
@@ -119,13 +206,14 @@ static int parse_reg_name(RzRegItem *reg, csh handle, cs_insn *insn, int reg_num
...
 		break;
 	case RISCV_OP_MEM:
-		if (OPERAND(reg_num).mem.base != RISCV_REG_INVALID) {
-			reg->name = (char *)cs_reg_name(handle, OPERAND(reg_num).mem.base);
+		if (op->mem.base != RISCV_REG_INVALID) {
+			reg->name = (char *)cs_reg_name(handle, op->mem.base);
```
Rot127, line 216: Here as well.

librz/arch/p/analysis/analysis_riscv_cs.c
Rot127: Ping on this one.

test/db/analysis/riscv
```diff
@@ -835,19 +835,19 @@ nth      paddr      vaddr bind   type   size lib name
...
+/ sym.test_reductions(int64_t arg1, int64_t arg2, int64_t arg3, int64_t arg4, int64_t arg5, int64_t arg8);
 |       :   ; arg int64_t arg1 @ a0
 |       :   ; arg int64_t arg2 @ a1
 |       :   ; arg int64_t arg3 @ a2
 |       :   ; arg int64_t arg4 @ a3
 |       :   ; arg int64_t arg5 @ a4
-|       :   ; arg int64_t arg6 @ a5
+|       :   ; arg int64_t arg8 @ a7
 |       :   ; var int64_t var_c0h @ stack - 0xc0
-|       :   ; arg int64_t arg_7fffffffffffffcfh @ stack + 0x7fffffffffffffcf
```
Rot127, line 846: lol

Rot127 (changes requested):

librz/arch/isa/riscv/riscv_il_base.h
moste00: Oh I didn't mean the source of the jump here, I meant the written destination register. A "one-way" jump is that which doesn't record its current location (the "link" address, as it's called by the RISC-V spec, which implement calls), so it doesn't write to a destination, so it can be implemented as a single direct effect.

The ones that do write their current location to their destination register are linking jumps and they must write first to `rd` so they have their own macros that hardcode the shape of the "write then jump" logic. 

(Technically speaking, `DEFINE_LIFTER_WITH_EFFECT` is extremly general and can arbitarily do anything using `SEQ`, including writing to any register, but the convention I made here is to only use it for "simple" one-off effects that specifically don't involve writing to registers, `SEQ` effects are only used inside their own lifter definitions not macros, they are mostly needed in the atomic and the floating point PRs rather than here.)

librz/arch/isa/riscv/riscv_il_m.c
moste00: Yup, if you remember I told you about it once in the mattermost chat.

> defined in the RISC-V ISA?

Indeed: 
image

along with the rationale:

image

moste00:
>> image

@Rot127 
> Ping on this one.

I don't know, I'm very strongly in favor of doing whatever QEMU does. QEMU leaves it at 64 bits regardless of the ISA bitness, so I'm leaning towards doing this as well.

If adherence to the spec is strictly necessary, I can convert to 32 bit and do an adaptation trick in the comparison in `rz-tracetest`, but I don't see why would it be useful if something as big as QEMU deviates from the spec.

(If you're worried that our branch of QEMU is too old I can check on the current upstream to see if they fixed it to 32 bits or not)

librz/arch/isa/riscv/riscv_il.c
moste00: Understood, ignore the comment, it's gesturing at some outside-of-IL mechanism to handle syscalls, does this not exist at all in Rizin ? so users know syscalls are complete no-ops ?

librz/arch/isa/riscv/riscv_il.c
moste00: Ok, I made the `INS_INVALID` and the `RZ_ARRAY_SIZE` into `rz_return_if_fail` calls but I have to keep `RISCV_INS_ENDING` check with its logging, and it can't be converted into an `&&` expression because `rz_log` returns void.

librz/arch/isa/riscv/riscv_il_integer_reg_names.h
moste00: Ehh, it would require passing a capstone handle to all the lifters, would depend on capstone syntax options, and it would be hard to edit there if  the register profile drifted. (Capstone API is very conservative, it's a hassle to change a name there if we ever wanted to change the names in the register profile without breaking capstone's consumers.)

I would say we keep it, a little duplication buying us decoupling from Capstone and being explicit about which register is accessed is worth it IMO. But you can override me on this if  you really hate this duplication.

PR #6291, opened 2026-04-27
RzReg: Associate roles with RzRegItem instead of name strings

thestr4ng3r:

**Detailed description**

Preparation for #6178

Register roles (RzRegisterId) are now associated directly with an RzRegItem in RzReg rather than mapping to name strings, which previously needed an additional hashtable lookup to get more information about the register. Conversely, if the name is needed from an RzRegItem, it is available directly as a member.

This is not a pure refactor as there were cases before where a register name was assigned to a role in the register profile, but no register actually existed under that name. Such cases will now cause a warning to be printed during profile load and the role association will be ignored. Changes in register profiles in this commit are for fixing such cases.

**Test plan**

Tests should pass with no relevant changes in behavior

### DIFF
--- a/librz/arch/fcn.c
+++ b/librz/arch/fcn.c
@@ -762,8 +762,8 @@ static RzAnalysisBBEndCause run_basic_block_analysis(RzAnalysisTaskItem *item, R
 			gotoBeach(RZ_ANALYSIS_RET_END);
 		}
 
-		const char *bp_reg = analysis->reg->name[RZ_REG_NAME_BP];
-		const char *sp_reg = analysis->reg->name[RZ_REG_NAME_SP];
+		const char *bp_reg = rz_reg_get_name(analysis->reg, RZ_REG_NAME_BP);
+		const char *sp_reg = rz_reg_get_name(analysis->reg, RZ_REG_NAME_SP);
 		bool has_stack_regs = bp_reg && sp_reg;
 
 		if (analysis->opt.nopskip && fcn->addr == at) {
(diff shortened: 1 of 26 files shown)


librz/debug/p/native/reg/darwin-arm64.h
```diff
@@ -11,16 +11,16 @@
 #endif
 return rz_str_dup(
 	"=PC	pc\n"
-	"=SP	sp\n" // XXX
-	"=BP	x30\n" // XXX
+	"=SP	sp\n"
+	"=BP	x29\n"
```
wargio, line 15: is this correct?

wargio (commented):
Nice! i would love to also maybe map the registers to bit vectors instead of using arenas.

wargio (approved):

wargio:
```
[XX] 00h00m00s103ms db/analysis/sysz sysz regs
```

librz/arch/fcn.c
```diff
@@ -2260,14 +2260,16 @@ static void __analysis_fcn_check_bp_use(RzAnalysis *analysis, RzAnalysisFunction
 			}
 			switch (op.type) {
 			case RZ_ANALYSIS_OP_TYPE_MOV:
-			case RZ_ANALYSIS_OP_TYPE_LEA:
-				if (can_affect_bp(analysis, &op) && op.src[0] && op.src[0]->reg && op.src[0]->reg->name && strcmp(op.src[0]->reg->name, analysis->reg->name[RZ_REG_NAME_SP])) {
+			case RZ_ANALYSIS_OP_TYPE_LEA: {
+				const char *sp = rz_reg_get_name(analysis->reg, RZ_REG_NAME_SP);
+				if (can_affect_bp(analysis, &op) && op.src[0] && op.src[0]->reg && op.src[0]->reg->name && sp && strcmp(op.src[0]->reg->name, sp)) {
```
notxvilka: Probably can be split into two lines for readability

notxvilka (approved):
Good simplification!

librz/debug/p/native/reg/darwin-arm64.h
thestr4ng3r: Yes, x30 is lr and x29 is fp. I double-checked in an actual debug session.

thestr4ng3r:
> i would love to also maybe map the registers to bit vectors instead of using arenas.

This might work well since bitvectors already have the apis for reading and writing at arbitrary bit offsets.

wargio:
we can work on this later on

thestr4ng3r:
Yes. I would also like to decouple register value storage from register definition.

PR #6325, opened 2026-05-08
asm.arch: Default to 64 bits

kazarmy:

**Detailed description**

Based on https://lwn.net/Articles/1035727/ I think it's safe nowadays for `asm.arch` to default to 64 bits if given a choice, and this pr does so. This should also fix the s390x build.

**Test plan**

**Closing issues**

...


kazarmy:
Hmm probably not a good idea

librz/arch/asm.c
```diff
@@ -445,10 +445,10 @@ static ut32 asm_get_first_default_bits(RzAsmPlugin *h) {
 		return RZ_SYS_BITS << 3;
 	}
 
-	if (h->bits & 32) {
-		return 32;
-	} else if (h->bits & 64) {
+	if (h->bits & 64) {
 		return 64;
+	} else if (h->bits & 32) {
+		return 32;
```
wargio, line 451: i had a similar issue with this code path. is a mess.

PR #6328, opened 2026-05-08
Feature: Implement gadget cache 

### LINKED ISSUE #5544: Implement `rop.cache` to cache ROP gadgets
### Work environment

| OS/arch/bits (mandatory)                             | darwin arm 64
| File format of the file you reverse (mandatory)      | raw binary
| Architecture/bits of the file (mandatory)            |  risc-v 64
| `rizin -v` full output, **not truncated** (mandatory)  | rizin 0.9.0 @ darwin-arm-64 commit: 856c28f1e8c7e2d3f6b9789d4d0f4aa70c6d242f

### Expected behavior
when `rop.cache` enabled, the time for searching a rop gadgets should be significantly reduced

### Actual behavior
almost no difference

I am sry I can't share the origin binary, but I test it in a just random data, same issue.
I use following cmd to parse and search
```
rizin -a riscv b 64 -B 0x10000000 -nn test
```
```
/R/ jalr.a.
``` 

Image

MrQuantum1915:

**Detailed description**

ok so this required more work then i thought.
- [x] is there a leak of global_var_tree - NO
- [x]  try to solve 26  mem leaks (https://github.com/rizinorg/rizin/pull/6354)?

**a small flowchart :)**
screenshot_20260515_025706

**Commit-1:** Refactor - seperate search from filtering
   - seperated filter from building gadgets (first buidl then filter)
   - this needed to be done to allow caching, as we need to cache all gadgets without filtering

---

**Commit-2:** Fix custom search , constrain the interval instead of raw search
   - this seemed to be a mistake from OG dev, see:
   1. the first thing is that this path only hits when `context->from/to` is set using the RZ_API, normal flow when using `/R` command will not hit this ever, because from and to are both initialised to 0 when the new ` RzGadgetSearchContext` object is made.
   2. and when hit, it directly called the function on raw range instead of passing it to boundary fetching logic. 
    
   - So i fixed that and instead constraint the search interval and keep normal flow...

---

**Commit-3:**  Implemented cache hit + miss
   - Used RBTree and not Hashtable because we need to ITERATE when printing, hash table is not good here.
   - the Cache metadata lives in `RzGadgetCache` struct under `RzAnalysis` , which decides validity of cache
   - each node (representing a gadget) in tree contains:
       - gadget start address (key used to order tree)
       - the hitlist - which is `RzPVector` having pointers to `RzCoreAsmHit` objects which contains address of the instruction and length. **NOTE:** No code string is set during build either. So in short we still need to do disassembly on printing from cache which is apparently fine.
       - delay size and intrusive RB Node for left and right child pointers.

   - u can see flowchart above
    
   - Another problem was `filter_gadget()` function was passed a `io` buffer originally, which was determined while fetching boundaries for building gadgets. But since for printing from cache, we just have full range (`search.from/to`) which is by default 0 to `UINT64_MAX` we cant allocate that much buffer :0
   - so what i did is that if buffer is not provided to `filter_gadget()` function, i allocate a sufficient 32 bytes buffer in the function itself and read the instruction there.

---

**Commit-4:** Still the semantic filtering and detail filtering was broke for gadget cache path
   - Seperated semantic and detail filtering from building gadgets, I extracted them to a common function so that when i am printing gadgets from cache I can  directly call this filtering functions instead of `process_disassemly`.
   - call this common filter function in `print_gadget_from_cache()`
   - I also fixed `search.maxhits` (max_count) which stopped stoped search when limit reached. I kinda used a trick so that when cache is enabled, and maxhit reaches while iterating, Dont print gadget but continue to search and insert gadgets into tree. This allowed to make full tree instead of partial one. 
    
---
**Performance metric**: of single run though, not average :)

- Binary : `test/bins/elf/libstdc++.so.6` (14MB)
- command = `/R`

|  | cold start (initial search + cache build) | warm start (cached search) | del |
| :--- | :--- | :--- | :--- |
| **exe time** | 10.342648 s | 0.130789 s | **~98.7% reduction  (~79x speedup)** |
| **mem (RAM usage)** | 387 MB (before cache building) | 413 MB  (after cache building)| **+26 MB (~6.72% increase)** |
	
- time was measured by `time /R` command
- RAM usage of rizin session using `BTOP++` TUI

I did some maths for the space usage , as we have total 15788 ROP gadgets, the total comes to 4.58MB
the 26MB spike is may be heap fragmentation or somthing...

I think this much space trade off is tolerable.
 
---
**QUESTIONS**: @wargio @notxvilka @Rot127  
- ~should we enable cache by default, currently the default is false in config.~
- ~i added logs just for debugging purpose, i can remove them if u want...~
- ~Should we have an option for user to clear cache when he needs??~
- ~How am I supposed to write tests for this?~

**Test plan**

...

**Closing issues**

closes #5544

### DIFF
--- a/librz/arch/analysis.c
+++ b/librz/arch/analysis.c
@@ -532,7 +532,7 @@ RZ_API void rz_analysis_set_xrefs_to(RZ_NONNULL RzAnalysis *analysis, HtUP *xref
  * \param type The RzGadgetType of gadget cache to retrieve.
  * \return A pointer to the requested RzGadgetCache, or NULL if the type is invalid.
  */
-RZ_API RZ_BORROW RzGadgetCache *rz_analysis_get_gadget_cache(RZ_NONNULL RzAnalysis *analysis, ut8 type) {
+RZ_API RZ_BORROW RzGadgetCache *rz_analysis_get_gadget_cache(RZ_NONNULL RzAnalysis *analysis, RzGadgetType type) {
 	rz_return_val_if_fail(analysis, NULL);
 	if (type >= 3) {
 		return NULL;
@@ -551,7 +551,7 @@ RZ_API RZ_BORROW RzGadgetCache *rz_analysis_get_gadget_cache(RZ_NONNULL RzAnalys
  * \param gadget_cache Pointer to the RzGadgetCache object to set.
  * \param type The RzGadgetType of gadget cache to set.
  */
-RZ_API void rz_analysis_set_gadget_cache(RZ_NONNULL RzAnalysis *analysis, RZ_NULLABLE RzGadgetCache *gadget_cache, ut8 type) {
+RZ_API void rz_analysis_set_gadget_cache(RZ_NONNULL RzAnalysis *analysis, RZ_NULLABLE RzGadgetCache *gadget_cache, RzGadgetType type) {
 	rz_return_if_fail(analysis);
 	if (type < 3) {
 		if (analysis->gadget_cache[type] == gadget_cache) {
--- a/librz/arch/analysis.c
+++ b/librz/arch/analysis.c
@@ -165,7 +165,9 @@ RZ_API RzAnalysis *rz_analysis_new(RZ_NULLABLE const char *sdb_types_path) {
 	}
 	analysis->ht_global_var = ht_sp_new(HT_STR_DUP, NULL, (HtSPFreeValue)rz_analysis_var_global_free);
 	analysis->ht_gadget_semantics = NULL;
-	analysis->ht_gadget = NULL;
+	analysis->gadget_cache[0] = (RzGadgetCache){ 0 };
+	analysis->gadget_cache[1] = (RzGadgetCache){ 0 };
+	analysis->gadget_cache[2] = (RzGadgetCache){ 0 };
 	analysis->global_var_tree = NULL;
 	analysis->il_vm = NULL;
 	analysis->hash = rz_hash_new();
@@ -221,6 +223,9 @@ RZ_API void rz_analysis_free(RZ_NULLABLE RzAnalysis *a) {
 	rz_str_constpool_fini(&a->constpool);
 	ht_sp_free(a->ht_global_var);
 	ht_up_free(a->ht_gadget_semantics);
+	rz_rbtree_free(a->gadget_cache[0].tree, a->gadget_cache[0].free, NULL);
+	rz_rbtree_free(a->gadget_cache[1].tree, a->gadget_cache[1].free, NULL);
+	rz_rbtree_free(a->gadget_cache[2].tree, a->gadget_cache[2].free, NULL);
 	ht_sp_free(a->plugins);
 	rz_analysis_debug_info_free(a->debug_info);
 	ht_sp_free(a->ht_virtual_xrefs);
@@ -511,6 +516,33 @@ RZ_API RZ_BORROW HtUP *rz_analysis_get_xrefs_to(RZ_NONNULL RzAnalysis *analysis)
 RZ_API void rz_analysis_set_xrefs_to(RZ_NONNULL RzAnalysis *analysis, HtUP *xrefs_to) {
 	rz_return_if_fail(analysis);
 	analysis->ht_xrefs_to = xrefs_to;
+}
+
+/**
+ * \brief Get the gadget cache for a specific gadget type.
+ * \param analysis Pointer to the RzAnalysis object.
+ * \param type The RzGadgetType of gadget cache to retrieve.
+ * \return A pointer to the requested RzGadgetCache, or NULL if the type is invalid.
+ */
+RZ_API RZ_BORROW RzGadgetCache *rz_analysis_get_gadget_cache(RZ_NONNULL RzAnalysis *analysis, ut8 type) {
+	rz_return_val_if_fail(analysis, NULL);
+	if (type >= 3) {
+		return NULL;
+	}
+	return &analysis->gadget_cache[type];
+}
+
+/**
+ * \brief Set the gadget cache in the analysis object for a specific gadget type.
+ * \param analysis Pointer to the RzAnalysis object.
+ * \param gadget_cache Pointer to the RzGadgetCache object to set.
+ * \param type The RzGadgetType of gadget cache to set.
+ */
+RZ_API void rz_analysis_set_gadget_cache(RZ_NONNULL RzAnalysis *analysis, RzGadgetCache *gadget_cache, ut8 type) {
+	rz_return_if_fail(analysis);
+	if (type < 3) {
+		analysis->gadget_cache[type] = *gadget_cache;
+	}
 }
 
 RZ_API RZ_BORROW HtUP *rz_analysis_get_gadget_semantics(RZ_NONNULL RzAnalysis *analysis) {
--- a/librz/arch/analysis.c
+++ b/librz/arch/analysis.c
@@ -165,9 +165,9 @@ RZ_API RzAnalysis *rz_analysis_new(RZ_NULLABLE const char *sdb_types_path) {
 	}
 	analysis->ht_global_var = ht_sp_new(HT_STR_DUP, NULL, (HtSPFreeValue)rz_analysis_var_global_free);
 	analysis->ht_gadget_semantics = NULL;
-	analysis->gadget_cache[0] = (RzGadgetCache){ 0 };
-	analysis->gadget_cache[1] = (RzGadgetCache){ 0 };
-	analysis->gadget_cache[2] = (RzGadgetCache){ 0 };
+	analysis->gadget_cache[0] = NULL;
+	analysis->gadget_cache[1] = NULL;
+	analysis->gadget_cache[2] = NULL;
 	analysis->global_var_tree = NULL;
 	analysis->il_vm = NULL;
 	analysis->hash = rz_hash_new();
@@ -185,6 +185,14 @@ RZ_API void plugin_fini(RzAnalysis *analysis) {
 		RZ_LOG_ERROR("analysis plugin '%s' failed to terminate.\n", p->name);
 	}
 	analysis->plugin_data = NULL;
+}
+
+static void gadget_cache_free(RzGadgetCache *gadget_cache) {
+	if (!gadget_cache) {
+		return;
+	}
+	rz_rbtree_free(gadget_cache->tree, gadget_cache->free, NULL);
+	free(gadget_cache);
 }
 
 void __block_free_rb(RBNode *node, void *user);
@@ -223,9 +231,9 @@ RZ_API void rz_analysis_free(RZ_NULLABLE RzAnalysis *a) {
 	rz_str_constpool_fini(&a->constpool);
 	ht_sp_free(a->ht_global_var);
 	ht_up_free(a->ht_gadget_semantics);
-	rz_rbtree_free(a->gadget_cache[0].tree, a->gadget_cache[0].free, NULL);
-	rz_rbtree_free(a->gadget_cache[1].tree, a->gadget_cache[1].free, NULL);
-	rz_rbtree_free(a->gadget_cache[2].tree, a->gadget_cache[2].free, NULL);
+	for (int i = 0; i < 3; i++) {
+		gadget_cache_free(a->gadget_cache[i]);
+	}
 	ht_sp_free(a->plugins);
 	rz_analysis_debug_info_free(a->debug_info);
 	ht_sp_free(a->ht_virtual_xrefs);
@@ -529,19 +537,33 @@ RZ_API RZ_BORROW RzGadgetCache *rz_analysis_get_gadget_cache(RZ_NONNULL RzAnalys
 	if (type >= 3) {
 		return NULL;
 	}
-	return &analysis->gadget_cache[type];
+	return analysis->gadget_cache[type];
 }
 
 /**
  * \brief Set the gadget cache in the analysis object for a specific gadget type.
+ *
+ * Takes ownership of `gadget_cache` pointer.
+ * The caller must not free it after passing it to this function.
+ * If a cache already exists for the given type, it will be freed before setting the new one.
+ *
  * \param analysis Pointer to the RzAnalysis object.
  * \param gadget_cache Pointer to the RzGadgetCache object to set.
  * \param type The RzGadgetType of gadget cache to set.
  */
-RZ_API void rz_analysis_set_gadget_cache(RZ_NONNULL RzAnalysis *analysis, RzGadgetCache *gadget_cache, ut8 type) {
+RZ_API void rz_analysis_set_gadget_cache(RZ_NONNULL RzAnalysis *analysis, RZ_NULLABLE RzGadgetCache *gadget_cache, ut8 type) {
 	rz_return_if_fail(analysis);
 	if (type < 3) {
-		analysis->gadget_cache[type] = *gadget_cache;
+		if (analysis->gadget_cache[type] == gadget_cache) {
+			return;
+		}
+
+		// delete old cache
+		if (analysis->gadget_cache[type]) {
+			gadget_cache_free(analysis->gadget_cache[type]);
+		}
+
+		analysis->gadget_cache[type] = gadget_cache;
 	}
 }
 
--- a/librz/arch/analysis_private.h
+++ b/librz/arch/analysis_private.h
@@ -88,7 +88,7 @@ struct rz_analysis_t {
 	RzPlatformTargetIndex *platform_target;
 	HtSP *ht_global_var; // global variables
 	HtUP *ht_gadget_semantics; ///< cache gadget semantic information
-	HtUP *ht_gadget; ///< cache gadget address list
+	RzGadgetCache gadget_cache[3]; ///< 3 separate caches 0=ROP, 1=JOP and 2=COP gadgets
 	RBTree global_var_tree; // global variables by address. must not overlap
 	RzHash *hash;
 	RzAnalysisDebugInfo *debug_info; ///< store all debug info parsed from DWARF, etc..
--- a/librz/arch/analysis_private.h
+++ b/librz/arch/analysis_private.h
@@ -88,7 +88,7 @@ struct rz_analysis_t {
 	RzPlatformTargetIndex *platform_target;
 	HtSP *ht_global_var; // global variables
 	HtUP *ht_gadget_semantics; ///< cache gadget semantic information
-	RzGadgetCache gadget_cache[3]; ///< 3 separate caches 0=ROP, 1=JOP and 2=COP gadgets
+	RzGadgetCache *gadget_cache[3]; ///< 3 separate caches 0=ROP, 1=JOP and 2=COP gadgets
 	RBTree global_var_tree; // global variables by address. must not overlap
 	RzHash *hash;
 	RzAnalysisDebugInfo *debug_info; ///< store all debug info parsed from DWARF, etc..
--- a/librz/core/gadget.c
+++ b/librz/core/gadget.c
@@ -1404,7 +1404,7 @@ static void update_search_context(const RzGadgetSearchContext *context, const ch
 	}
 }
 
-static bool filter_gadget(RzCore *core, const ut8 *buf, int start_idx, RzGadgetSearchContext *context,
+static bool filter_gadget(RzCore *core, const ut8 *buf, RzGadgetSearchContext *context,
 	RzList /*<char *>*/ *rx_list, RzPVector /*<RzCoreAsmHit *>*/ *hitlist) {
 
 	bool is_greparg = !(context->mask & (RZ_GADGET_PRINT_DETAIL | RZ_GADGET_ANALYZE)) && !(context->detail_mask | RZ_GADGET_DETAIL_SEARCH_NON) && context->greparg;
@@ -1476,7 +1476,7 @@ static bool filter_gadget(RzCore *core, const ut8 *buf, int start_idx, RzGadgetS
 	return pass;
 }
 
-static RzPVector /*<RzCoreAsmHit *>*/ *build_gadget_hitlist_raw(RzCore *core, ut8 *buf, int idx,
+static RzPVector /*<RzCoreAsmHit *>*/ *build_gadget_hitlist_raw(RzCore *core, ut8 *buf, ssize_t idx,
 	RzGadgetSearchContext *context, RzGadgetEndListPair *end_gadget, RZ_OUT RzStrBuf **sb_out) {
 
 	RzPVector *hitlist = rz_pvector_new((RzPVectorFree)rz_core_asm_hit_free);
@@ -1595,7 +1595,7 @@ static RzPVector /*<RzCoreAsmHit *>*/ *deep_copy_hitlist(const RzPVector /*<RzCo
 	return new_hitlist;
 }
 
-static RzPVector /*<RzCoreAsmHit *>*/ *construct_gadget(RzCore *core, ut8 *buf, int idx, RzGadgetSearchContext *context,
+static RzPVector /*<RzCoreAsmHit *>*/ *construct_gadget(RzCore *core, ut8 *buf, ssize_t idx, RzGadgetSearchContext *context,
 	RzList /*<char *>*/ *rx_list, RzGadgetEndListPair *end_gadget) {
 
 	RzStrBuf *sb = NULL;
@@ -1626,7 +1626,7 @@ static RzPVector /*<RzCoreAsmHit *>*/ *construct_gadget(RzCore *core, ut8 *buf, 
 		}
 	}
 
-	if (!filter_gadget(core, buf, idx, context, rx_list, hitlist)) {
+	if (!filter_gadget(core, buf, context, rx_list, hitlist)) {
 		rz_pvector_free(hitlist);
 		return NULL;
 	}
@@ -2013,7 +2013,7 @@ static bool apply_post_build_filters(RzCore *core, RzGadgetSearchContext *contex
 	return true;
 }
 
-static bool process_disassembly(RzCore *core, ut8 *buf, const int idx, RzGadgetSearchContext *context,
+static bool process_disassembly(RzCore *core, ut8 *buf, const ssize_t idx, RzGadgetSearchContext *context,
 	RzList /*<char *>*/ *rx_list, RzGadgetEndListPair *end_gadget) {
 	RzAsmOp *asmop = rz_asm_op_new();
 	bool status = false;
@@ -2057,7 +2057,7 @@ fini:
 	return status;
 }
 
-static bool update_end_gadget(int *i, const int gadget_depth, RzGadgetEndListPair **end_gadget, const RzGadgetSearchContext *context) {
+static bool update_end_gadget(ssize_t *i, const int gadget_depth, RzGadgetEndListPair **end_gadget, const RzGadgetSearchContext *context) {
 	rz_return_val_if_fail(end_gadget && context, false);
 	if (*i > (*end_gadget)->instr_offset) {
 		// We've exhausted the first end-gadget section,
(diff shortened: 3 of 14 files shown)


librz/core/gadget.c
```diff
@@ -1419,39 +1482,107 @@ static RzPVector /*<RzCoreAsmHit *>*/ *construct_gadget(RzCore *core, ut8 *buf,
...
+	if (context->cache) {
+		ut64 gadget_addr = context->from + idx;
+		RzGadgetCache *gadget_cache = rz_analysis_get_gadget_cache(core->analysis, context->type);
+		RzGadgetCacheNode *node = RZ_NEW(RzGadgetCacheNode);
+		node->addr = gadget_addr;
+		node->hitlist = deep_copy_hitlist(hitlist);
+		node->delay_size = end_gadget->delay_size;
+		void *data = &node->addr;
+		if (!rz_rbtree_insert(&gadget_cache->tree, data, &node->rb, gadget_cache_node_cmp, NULL)) {
+			// TODO: should invalidate cache? idts as this path will rarely hit, dont deserve writing extra code?
```
MrQuantum1915: @Rot127 @notxvilka @wargio 
wdyt about this comment??

MrQuantum1915:
sorry for a delay , i was having fever for couple days and then was getting used to apis of RBTree and other stuffs.

MrQuantum1915:
I suspect those 26 leaks is in `armil_32.c` , opened a draft PR (https://github.com/rizinorg/rizin/pull/6354)
lets see...

librz/core/gadget.c
```diff
@@ -1419,39 +1482,107 @@ static RzPVector /*<RzCoreAsmHit *>*/ *construct_gadget(RzCore *core, ut8 *buf,
...
+			RZ_LOG_ERROR("Failed to cache gadget for address 0x%" PFMT64x "\n", gadget_addr);
+			free_gadget_cache_node(&node->rb, NULL);
+		}
```
wargio: move this into a function that returns boolean.

librz/core/gadget.c
```diff
@@ -1419,39 +1482,107 @@ static RzPVector /*<RzCoreAsmHit *>*/ *construct_gadget(RzCore *core, ut8 *buf,
...
 		rz_pvector_free(hitlist);
 		rz_strbuf_free(sb);
 		return NULL;
 	}
 	rz_strbuf_free(sb);
+
+	if (context->cache) {
+		ut64 gadget_addr = context->from + idx;
+		RzGadgetCache *gadget_cache = rz_analysis_get_gadget_cache(core->analysis, context->type);
+		RzGadgetCacheNode *node = RZ_NEW(RzGadgetCacheNode);
```
wargio: check for null

librz/core/gadget.c
```diff
@@ -1876,6 +2022,62 @@ static bool update_end_gadget(int *i, const int gadget_depth, RzGadgetEndListPai
 	return true;
 }
 
+static bool print_gadgets_from_cache(RzCore *core, RzGadgetCache *gadget_cache, RzGadgetSearchContext *context, RzList /*<char *>*/ *rx_list) {
+	void *addr = &context->from;
+	RBIter iter = rz_rbtree_lower_bound_forward(gadget_cache->tree, addr, gadget_cache_node_cmp, NULL);
```
wargio, line 2111: @notxvilka this should be returning an RzIterator

@MrQuantum1915 ignore this comment.

librz/core/gadget.c
```diff
@@ -1906,7 +2108,7 @@ static int handle_gadget_search_address(RzCore *core, RzGadgetSearchContext *con
 	RzGadgetEndListPair *end_gadget = rz_list_pop(context->end_list);
 	// Start at just before the first end gadget.
 	const int next = end_gadget->instr_offset;
-	for (int i = 0; i < delta && context->max_count; i += context->increment) {
+	for (int i = 0; i < delta && (context->cache || context->max_count); i += context->increment) {
```
wargio:
```suggestion
	for (size_t i = 0; i < delta && (context->cache || context->max_count); i += context->increment) {
```

librz/core/gadget.c
```diff
@@ -1971,10 +2173,48 @@ RZ_API RzCmdStatus rz_core_gadget_search(RZ_NONNULL RzCore *core, RZ_NONNULL RzG
...
+				RZ_LOG_INFO("core: Parameters change detected, Rebuilding cache\n");
+			}
+		}
+
+		// cache miss
+		// delete old cache (if exist)
+		if (gadget_cache) {
+			rz_rbtree_free(gadget_cache->tree, free_gadget_cache_node, NULL);
+		}
+		RZ_LOG_INFO("core: Building gadget cache for address range 0x%" PFMT64x " - 0x%" PFMT64x "\n", context->from, context->to);
```
wargio: RZ_LOG_DEBUG

librz/core/gadget.c
MrQuantum1915: @wargio

then when it returns false, should i [ free the hitlist + return null ] and as a result dont print that gadget??

or do nothing and continue building and printing?

librz/core/gadget.c
MrQuantum1915: @wargio 

this is the same issue we discussed here : https://github.com/rizinorg/rizin/pull/6108#discussion_r3124399788

this same `i` is used in many function ahead, and at one place this is used with `RzAnalysisOp aop.size` for subtraction  which is `int`

i suggest solving this in another PR

librz/core/gadget.c
wargio: if returns false means the allocation failed and should just gracefully terminate and print the error.

librz/core/gadget.c
wargio: use `ssize_t` then

MrQuantum1915:
Fixed all reviews @wargio 

>- should we enable cache by default, currently the default is false in config.
>- ~i added logs just for debugging purpose, i can remove them if u want...~
>- Should we have an option for user to clear cache when he needs??
>- How am I supposed to write tests for this?

librz/arch/analysis_private.h
```diff
@@ -88,7 +88,7 @@ struct rz_analysis_t {
 	RzPlatformTargetIndex *platform_target;
 	HtSP *ht_global_var; // global variables
 	HtUP *ht_gadget_semantics; ///< cache gadget semantic information
-	HtUP *ht_gadget; ///< cache gadget address list
+	RzGadgetCache gadget_cache[3]; ///< 3 seperate caches 0=ROP, 1=JOP and 2=COP gadgets
```
notxvilka: "separate"

librz/arch/analysis.c
```diff
@@ -513,6 +518,21 @@ RZ_API void rz_analysis_set_xrefs_to(RZ_NONNULL RzAnalysis *analysis, HtUP *xref
 	analysis->ht_xrefs_to = xrefs_to;
 }
 
+RZ_API RZ_BORROW RzGadgetCache *rz_analysis_get_gadget_cache(RZ_NONNULL RzAnalysis *analysis, ut8 type) {
```
notxvilka: Add Doxygen to these please

librz/core/gadget.c
notxvilka: Good point. Maybe we should create an issue to track all iterator-like functions to use RzIterator only

librz/core/gadget.c
```diff
@@ -1906,7 +2129,7 @@ static int handle_gadget_search_address(RzCore *core, RzGadgetSearchContext *con
 	RzGadgetEndListPair *end_gadget = rz_list_pop(context->end_list);
 	// Start at just before the first end gadget.
 	const int next = end_gadget->instr_offset;
-	for (int i = 0; i < delta && context->max_count; i += context->increment) {
+	for (ssize_t i = 0; i < delta && (context->cache || context->max_count); i += context->increment) {
```
notxvilka, line 2194: Why signed?

notxvilka (approved):
Good overall but couple things need to be addressed first.

notxvilka:
> > * How am I supposed to write tests for this?

This could be done with the unit test. Do the search, then check the contents of the cache. See, for example, `test/unit/test_io_cache.c` which tests the RzIO cache.

librz/core/gadget.c
MrQuantum1915: @notxvilka 
context: https://github.com/rizinorg/rizin/pull/6328#discussion_r3245451100

notxvilka (approved):

MrQuantum1915:
@notxvilka 

I added the unit test, (its my first time writing any unit test...)
Is it fine or i need to add more checks, like for each node too??

I didnt use the previous test function because it called functions that were inside test file...
i needed to call `rz_core_gadget_search()` and not local `setup_hitlist()` function 

I didnt find a way to use them to test cache, as I wanted to call the API function and test it as a black box

MrQuantum1915:
dont know why codecov test fails for that unit test, it passes locally though...
will see later today

test/unit/test_rop.c
```diff
@@ -143,8 +143,50 @@ bool test_rz_direct_solver() {
 	mu_end;
 }
 
+bool test_rz_gadget_cache() {
+	RzCore *core = setup_rz_core("x86", 64);
```
Rot127: Add a test for Sparc to test delay slots.

test/unit/test_rop.c
```diff
@@ -143,8 +143,50 @@ bool test_rz_direct_solver() {
...
+	mu_assert_notnull(core, "setup_rz_core failed");
+	int size = sizeof(x86_64_buf_str) / sizeof(x86_64_buf_str[0]);
```
Rot127: `RZ_ARRAY_SIZE`

test/unit/test_rop.c
```diff
@@ -143,8 +143,50 @@ bool test_rz_direct_solver() {
...
+	mu_assert_null(gadget_cache, "Requesting cache for invalid gadget type should return NULL");
+	gadget_cache = rz_analysis_get_gadget_cache(core->analysis, RZ_GADGET_TYPE_ROP);
+	mu_assert_notnull(gadget_cache, "ROP gadget cache is NULL");
+	RzCmdStatus status = rz_core_gadget_search(core, context);
+	mu_assert_eq(status, RZ_CMD_STATUS_OK, "ROP gadget search failed");
+	RBIter iter;
+	int cnt = 0;
+	RzGadgetCacheNode *node;
+	rz_rbtree_foreach (gadget_cache->tree, iter, node, RzGadgetCacheNode, rb) {
+		cnt++;
```
Rot127, line 192: Please also test the gadget content. So we know that the cache doesn't manipulate them.

test/unit/test_rop.c
```diff
@@ -143,8 +143,50 @@ bool test_rz_direct_solver() {
...
+	}
+	mu_assert_eq(cnt, 9, "ROP gadget cache size mismatch");
+
+	status = rz_core_gadget_search(core, context);
+	mu_assert_eq(status, RZ_CMD_STATUS_OK, "ROP gadget search failed on using cache");
+	mu_assert_eq(cnt, 9, "ROP gadget cache size should not change on serving from cache");
```
Rot127: Count was not updated. This check always passes.

librz/include/rz_analysis.h
```diff
@@ -1370,6 +1382,8 @@ RZ_API RZ_BORROW HtUP *rz_analysis_get_xrefs_from(RZ_NONNULL RzAnalysis *analysi
 RZ_API void rz_analysis_set_xrefs_from(RZ_NONNULL RzAnalysis *analysis, HtUP *xrefs_from);
 RZ_API RZ_BORROW HtUP *rz_analysis_get_xrefs_to(RZ_NONNULL RzAnalysis *analysis);
 RZ_API void rz_analysis_set_xrefs_to(RZ_NONNULL RzAnalysis *analysis, HtUP *xrefs_to);
+RZ_API RZ_BORROW RzGadgetCache *rz_analysis_get_gadget_cache(RZ_NONNULL RzAnalysis *analysis, ut8 type);
+RZ_API void rz_analysis_set_gadget_cache(RZ_NONNULL RzAnalysis *analysis, RzGadgetCache *gadget_cache, ut8 type);
```
Rot127: Missing type annotations

librz/include/rz_analysis.h
```diff
@@ -43,6 +43,18 @@ typedef struct {
 	PJ *pj;
 } RzAnalysisMetaUserItem;
 
+/**
+ * \brief Gadget cache
+ */
+typedef struct rz_gadget_cache_t {
+	RBTree tree; ///< root node
+	ut64 from; ///< cached range start
+	ut64 to; ///< cached range end
```
Rot127:
```suggestion
	ut64 from; ///< cached address range start
	ut64 to; ///< cached address range end
```

librz/include/rz_analysis.h
```diff
@@ -43,6 +43,18 @@ typedef struct {
 	PJ *pj;
 } RzAnalysisMetaUserItem;
 
+/**
+ * \brief Gadget cache
```
Rot127, line 48: Add more detail how it caches.

librz/include/rz_analysis.h
```diff
@@ -43,6 +43,18 @@ typedef struct {
...
+ */
+typedef struct rz_gadget_cache_t {
+	RBTree tree; ///< root node
+	ut64 from; ///< cached range start
+	ut64 to; ///< cached range end
+	ut8 max_instr; ///< gadget.len used during cache build
+	bool allow_conditional; ///< gadget.conditional used during cache build
+	RBNodeFree free; ///< free function
```
Rot127: free function for what?

librz/core/gadget.c
```diff
@@ -2027,10 +2250,48 @@ RZ_API RzCmdStatus rz_core_gadget_search(RZ_NONNULL RzCore *core, RZ_NONNULL RzG
...
+			rz_rbtree_free(gadget_cache->tree, free_gadget_cache_node, NULL);
+		}
+		RZ_LOG_DEBUG("core: Building gadget cache for address range 0x%" PFMT64x " - 0x%" PFMT64x "\n", context->from, context->to);
+		gadget_cache->tree = NULL;
+		gadget_cache->free = free_gadget_cache_node;
+		gadget_cache->from = rz_config_get_integer(core->config, "search.from");
+		gadget_cache->to = rz_config_get_integer(core->config, "search.to");
+		gadget_cache->max_instr = context->max_instr;
+		gadget_cache->allow_conditional = context->allow_conditional;
+		rz_analysis_set_gadget_cache(core->analysis, gadget_cache, context->type);
```
Rot127: Please move all this into a static function for readability.

librz/core/gadget.c
```diff
@@ -2027,10 +2250,48 @@ RZ_API RzCmdStatus rz_core_gadget_search(RZ_NONNULL RzCore *core, RZ_NONNULL RzG
...
+					goto cleanup;
+				}
+			} else {
+				RZ_LOG_INFO("core: Parameters change detected, Rebuilding cache\n");
+			}
+		}
+
+		// cache miss
+		// delete old cache (if exist)
+		if (gadget_cache) {
```
Rot127: Move the tree free into the previous if case.

librz/core/gadget.c
```diff
@@ -2027,10 +2250,48 @@ RZ_API RzCmdStatus rz_core_gadget_search(RZ_NONNULL RzCore *core, RZ_NONNULL RzG
...
+			rz_rbtree_free(gadget_cache->tree, free_gadget_cache_node, NULL);
+		}
+		RZ_LOG_DEBUG("core: Building gadget cache for address range 0x%" PFMT64x " - 0x%" PFMT64x "\n", context->from, context->to);
```
Rot127: It says here that it builds the cache from `cintext->from/to` but below it uses the config values `search.from/to`.
Please use only one source of truth.

Rot127 (changes requested):
> should we enable cache by default, currently the default is false in config.

Yes, please.

> i added logs just for debugging purpose, i can remove them if u want...

All logs I have seen are reasonable.

> Should we have an option for user to clear cache when he needs??

Yes, please add a sub command.

> How am I supposed to write tests for this?

See suggestions below.

TODO:

- Please open an issue for Hexagon support.
- Add a test which runs a gadget search twice with commands. Store the output each time in a file. Then diff the files.

librz/include/rz_analysis.h
MrQuantum1915: the function that frees the `RzGadgetCacheNode` :  `free_gadget_cache_node()`
its there so that that function is available to `RzAnalysis` 

the free function is defined in `gadget.c` and `static`
i used same approach as 
- `__block_free_rb` defined in `block.c`
- `free_gadget_cache_node()` defined in `gadget.c`

MrQuantum1915: I will update the doxygen to be more descriptive...

MrQuantum1915:
@notxvilka @Rot127 @wargio 

As per rot127 reviews I made some more changes in last 5 commits: please review them

**1. Cache gadgets by default**: You can see i updated the test, FYI: Its just the ordering.
I checked why the ordering when serverd from cache and fresh search are different. The cache is keyed by start address, but during fresh search the gadgets are dicovered grouped by end-gadget (ret position), scanning backwards from each. 

I thought of it and There is no simple way to make them both consistent. 

This is the reason you will see in Unit test that I compare the outputs after SORTING (see below).

So it remains that order of output will be different when serving from cache and fresh search.
And i think its fine... because actually ordering by start address of gadget is better...
And i think order does not affect us as such...

...

**2. Fix critical error of violating privatisation of RzAnalysis :** 
- Noticed this due to rot127 feedback ([comment1](https://github.com/rizinorg/rizin/pull/6328#discussion_r3257457412) , [comment2](https://github.com/rizinorg/rizin/pull/6328#discussion_r3257433600 ))
- I made a critical mistake in handling of gadget cache struct. Before i violated the new privatisation of `RzAnalysis` and mistakenly declared struct array in RzAnalysis itself (now i delclare a pointer array instead of data array). I also  passed the address of the struct itself in getter function which gave direct access to the `RzGadgetCache` instead. Also i was deleting old cache from `gadget.c`. Now I moved that to `analysis.c` setter function.

...

**3. Subcommand to clear cache:** `/R-`, `/C-`,`/J-`
**4. Extract cache handling to static function: as per review**

...

**5. Unit test for gadget cache:**
- First search: Build cache,  store the node count and compare on second search
- as per review, I added testing for cache content too. 
- store the result in **SORTED** List for both fresh search and second search
- then iterate the list for comparing the strings
- added test for both x86 and mips for delay slot as per review

librz/core/gadget.c
MrQuantum1915: shifted deletion to `analysis.c`  setter function
see https://github.com/rizinorg/rizin/pull/6328#issuecomment-4492985946

test/unit/test_rop.c
MrQuantum1915: I added for MIPS, is it ok?

MrQuantum1915:
see this ss for example: the difference in order
screenshot_20260520_034407

MrQuantum1915:
>Please open an issue for Hexagon support.

@Rot127 I didnt get what you mean here?? There are already open issue for Hexagon support ig??

notxvilka (approved):

test/unit/test_rop.c
Rot127: Yes, as long as it tests for the delayed slot.

Rot127:
> I didnt get what you mean here?? There are already open issue for Hexagon support ig??

Hexagon is a Very Long Instruction Word (VLIW) architecture. Which executes 1-4 instructions in parallel. Those 1-4 instructions are called a packet.

So a ROP/COP etc gadget search would need to look at these packets, instead of single instructions.

> see this ss for example: the difference in order

Will review later. ~~But can you sort the gadgets before printing by address?~~ Sorry, missed the message above

MrQuantum1915:
This mem leaks are somewhere in upper level of config management 
https://github.com/rizinorg/rizin/actions/runs/26131592492/job/76857736924?pr=6328
and there are actually more than just 4 leaks all indirect

this same mem leaks appeared in previous PR - around 35 leaks : https://github.com/rizinorg/rizin/actions/runs/24354188848/job/71116621362?pr=6130

so this leaks are associated with many configs, which means the issue is somewhere on upper level...

i am trying to fix this...

Rot127:
> i am trying to fix this...

Ignore them. They happen in some test for every new config

notxvilka:
> Hexagon is a Very Long Instruction Word (VLIW) architecture. Which executes 1-4 instructions in parallel. Those 1-4 instructions are called a packet.
> 
> So a ROP/COP etc gadget search would need to look at these packets, instead of single instructions.

I suggest doing that in a separate PR after opening an issue to track it. I think the current scope of this PR is good as is. @Rot127 @wargio WDYT?

MrQuantum1915:
> I suggest doing that in a separate PR after opening an issue to track it. I think the current scope of this PR is good as is. @Rot127 @wargio WDYT?

I opened the issue : https://github.com/rizinorg/rizin/issues/6392

librz/include/rz_analysis.h
```diff
@@ -1370,6 +1389,8 @@ RZ_API RZ_BORROW HtUP *rz_analysis_get_xrefs_from(RZ_NONNULL RzAnalysis *analysi
 RZ_API void rz_analysis_set_xrefs_from(RZ_NONNULL RzAnalysis *analysis, HtUP *xrefs_from);
 RZ_API RZ_BORROW HtUP *rz_analysis_get_xrefs_to(RZ_NONNULL RzAnalysis *analysis);
 RZ_API void rz_analysis_set_xrefs_to(RZ_NONNULL RzAnalysis *analysis, HtUP *xrefs_to);
+RZ_API RZ_BORROW RzGadgetCache *rz_analysis_get_gadget_cache(RZ_NONNULL RzAnalysis *analysis, ut8 type);
+RZ_API void rz_analysis_set_gadget_cache(RZ_NONNULL RzAnalysis *analysis, RZ_NULLABLE RzGadgetCache *gadget_cache, ut8 type);
```
Rot127: Please pass the enum instead, not ut8

librz/include/rz_analysis.h
```diff
@@ -43,6 +43,25 @@ typedef struct {
...
+ * The cache is built for the given parameters (`from`, `to`, `max_instr`, `allow_conditional`)
+ * and is only valid for those parameters.
+ * When these parameters change, the cache must be rebuilt.
+ * The `free` callback is used to release per-node resources when the cache is cleared.
+ */
+typedef struct rz_gadget_cache_t {
+	RBTree tree; ///< root node
+	ut64 from; ///< cached address range start
+	ut64 to; ///< cached address range end
+	ut8 max_instr; ///< gadget.len used during cache build
```
Rot127:
```suggestion
	size_t max_instr; ///< gadget.len used during cache build
```
The compiler would most like pad it to size_t anyways.

librz/core/gadget.c
```diff
@@ -1404,20 +1404,83 @@ static void update_search_context(const RzGadgetSearchContext *context, const ch
...
+	ut64 delta = context->to - context->from;
+	void **it;
+	rz_pvector_foreach (hitlist, it) {
+		RzCoreAsmHit *hit = *it;
+		const ut8 *op_buf;
+		ut64 op_size;
+		ut8 local_buf[32];
+
+		if (buf) {
+			int buf_offset = (int)(hit->addr - context->from);
```
Rot127:
```suggestion
			st64 buf_offset = (st64)(hit->addr - context->from);
```

librz/core/gadget.c
```diff
@@ -1467,39 +1530,129 @@ static RzPVector /*<RzCoreAsmHit *>*/ *construct_gadget(RzCore *core, ut8 *buf,
...
+	if (sb_out) {
+		*sb_out = sb;
+	} else {
+		rz_strbuf_free(sb);
+	}
+	return hitlist;
+}
+
+static void free_gadget_cache_node(RBNode *node, void *user) {
+	RzGadgetCacheNode *n = container_of(node, RzGadgetCacheNode, rb);
```
Rot127, line 1563: Please check for NULL and return early if the case.
Maybe it is not possible in the current code, but free functions always should accept NULL (by general convention).

librz/core/gadget.c
```diff
@@ -1843,52 +1996,67 @@ static bool match_constraints(const RzGadgetInfo *gadget_info, const RzPVector /
...
+static bool process_disassembly(RzCore *core, ut8 *buf, const ssize_t idx, RzGadgetSearchContext *context,
+	RzList /*<char *>*/ *rx_list, RzGadgetEndListPair *end_gadget) {
+	RzAsmOp *asmop = rz_asm_op_new();
+	bool status = false;
+	const int ret = rz_asm_disassemble(core->rasm, asmop, buf + idx, context->to - context->from - idx);
+	if (!ret) {
+		goto fini;
+	}
+
+	rz_asm_set_pc(core->rasm, context->from + idx);
```
Rot127, line 2052: Reset the pc before the function returns?

librz/core/gadget.c
Rot127: +1 for issue + new PR.
Also can be combined by making the RzIterator passed via the stack not as heap allocated.

Rot127 (approved):
Please address those nitpicks. Otherwise lgtm

librz/include/rz_analysis.h
MrQuantum1915: `RzGadgetType` is defined in `rz_gadget.h` so its not available in `rz_analysis.h`
so i had to use `ut8` instead

librz/include/rz_analysis.h
MrQuantum1915: its defined as `ut8` in the main context struct , so to keep it consistent i defined it as ut8 here too

```c
typedef struct rz_gadget_search_context_t {
	RzGadgetType type; ///< Type of gadget to search for.
	ut8 max_instr; ///< Gadget search max length.
	bool subchains; ///< Display every length gadget from gadget.len=X to 2.
	bool allow_conditional; ///< Include conditional jump, calls and returns in gadget search.
	bool comments; ///< Display comments in gadget search output.
	char *greparg; ///< Grep argument string.
	const char *arch; ///< Architecture of the binary.
	bool regexp; ///< Regular expression argument flag.
	bool cache; ///< Cache the search results.
	RzGadgetRequestMask mask; ///< Mask for kind of gadget request operation.
	RzGadgetDetailSearchMask detail_mask; ///< Mask for searching gadgets given details.
	RzCmdStateOutput *state; ///< Command state output.
	int increment; ///< Gadget search increment value.
	ut64 max_count; ///< Maximum number of hits (0: no limit).
	ut64 from; ///< Start address to start gadget search.
	ut64 to; ///< End address to stop gadget search.
	RzList /*<RzGadgetEndListPair *>*/ *end_list; ///< List of end gadgets.
	HtSU *unique_hitlists; ///< Cache unique gadget hitlists.
	bool ret_val; ///< Flag to indicate return the search results.
	RzStrBuf *buf; ///< String buffer for storing search results.
	RzPVector /*<RzGadgetConstraint *>*/ *constraints; ///< User constraints for filtering.
} RzGadgetSearchContext;
```

librz/include/rz_analysis.h
Rot127: Why not include `rz_gadget.h`?

Rot127: Because of

```
../librz/egg/egg_lang.c:69:9: error: redeclaration of enumerator ‘SYSCALL’
   69 |         SYSCALL,
      |         ^~~~~~~
In file included from ../librz/include/rz_analysis.h:25,
                 from ../librz/include/rz_parse.h:10,
                 from ../librz/include/rz_asm.h:13,
                 from ../librz/include/rz_egg.h:4,
                 from ../librz/egg/egg_lang.c:4:
../librz/include/rz_gadget.h:70:9: note: previous definition of ‘SYSCALL’ with type ‘enum rz_gadget_il_instr_type’
   70 |         SYSCALL, ///< syscall
      |         ^~~~~~~
[322/709] Compiling C object librz/arch/librz_arch.so.0.9.0.p/var.c.o

```

Please rename the enum members then in the gadget header.

librz/include/rz_analysis.h
Rot127: Please change both then to `size_t`

MrQuantum1915:
@Rot127 all reviews addressed

Rot127 (approved):

librz/core/gadget.c
MrQuantum1915: are you expecting me to open the issue or you guys will?

PR #6371, opened 2026-05-17
Prevent RzBuffer's `Oxff_priv` from overriding `io.0xff`

kazarmy:

**Detailed description**

This pr prevents RzBuffer's `Oxff_priv` from overriding the `io.0xff` config var. ~~Not sure whether it'll seriously break stuff.~~ It probably won't break stuff.

**Test plan**

The changes make sense. All builds are green.

**Closing issues**

...


wargio (approved):

librz/bin/format/pe/pe_section.c
```diff
@@ -177,7 +177,9 @@ RzList /*<char *>*/ *PE_(section_flag_to_rzlist)(ut64 flag) {
 
 bool PE_(read_image_section_header)(RzBuffer *b, ut64 addr, PE_(image_section_header) * section_header) {
 	ut8 buf[sizeof(PE_(image_section_header))];
+	b->Oxff_priv_override = true;
```
wargio: this is a big hack

kazarmy: ok i think 3295e11bb549a98200b5943c83b3866407ecb64a is better

kazarmy: actually i think incomplete section headers should be removed since the corresponding sections are partially / fully hallucinations (unless Windows actually has defined behavior in this situation) but currently it appears the section headers are not stored in a data structure that allows for easy removal of individual section headers

kazarmy: removal of individual section headers is not impossible ofc even with the current data structure used but time is limited

kazarmy: `readpe` and the `iH` command actually says that the number of sections is 82 (!) so 3295e11bb549a98200b5943c83b3866407ecb64a is basically only about making sure the tests pass

librz/bin/format/pe/pe_section.c
```diff
@@ -177,9 +177,11 @@ RzList /*<char *>*/ *PE_(section_flag_to_rzlist)(ut64 flag) {
...
 bool PE_(read_image_section_header)(RzBuffer *b, ut64 addr, PE_(image_section_header) * section_header) {
 	ut8 buf[sizeof(PE_(image_section_header))];
-	b->Oxff_priv_override = true;
-	rz_buf_read_at(b, addr, buf, sizeof(buf));
-	b->Oxff_priv_override = false;
+	st64 bytes_read = rz_buf_read_at(b, addr, buf, sizeof(buf));
+	if (bytes_read < sizeof(buf)) {
+		// For incomplete section header, fill the rest of the section header with 0
+		memset(buf + bytes_read, 0, sizeof(buf) - bytes_read);
+	}
```
wargio: how about `ut8 buf[sizeof(PE_(image_section_header))] = { 0 };`

kazarmy: that works only if `rz_buf_read_at()` doesn't fill in the read shortfall itself, but currently it always does and i think it should do so because buffers that are partially uninitialized -- maybe the caller forgot to fully initialize the buffer --can cause intermittent test errors that are hard to debug, since the contents of the uninitialized portion of the buffer can be different across builds

wargio: i have many questions, one of which is, can a pe-section being less than the expected size? because this to me seems a good opportunity to fail reading.

kazarmy: i don't think it can be less than the expected size, but i do think it's an issue for a different pr

wargio: then this should be addressed before this PR can be merged, otherewise we will forget about it

wargio: One proper way to parse this structure is to use the RzBuffer offset api which will guarantee the successful read (or failure)

kazarmy: without documentation/evidence to the contrary, i'm going to assume that incomplete section header even by one byte --> section does not exist

librz/include/rz_util/rz_buf.h
```diff
@@ -59,6 +59,7 @@ struct rz_buf_t {
 	ut8 *whole_buf;
 	bool readonly;
 	ut8 Oxff_priv;
+	bool Oxff_priv_override;
```
Rot127: Document please

test/db/formats/pe/foldedhdr
```diff
@@ -1,5 +1,7 @@
 NAME=PE: corkami foldedhdr.exe - open
 FILE=bins/pe/foldedhdr.exe
+# HACK
```
Rot127: Document what exactly the hack is here.

librz/util/buf.c
```diff
@@ -16,6 +16,27 @@
 
 #define GET_STRING_BUFFER_SIZE 32
 
+static ut8 Oxff_global = 0xff;
+static bool Oxff_global_enabled = true;
```
Rot127: Do we really need to add global variables?
Is there no other way around it?

Also the setters for them are no where used?

Rot127 (changes requested):

test/db/formats/pe/foldedhdr
kazarmy: It's a hack because the cause of the timeout is not fixed, and one should not have to use the (possibly obscure) hack.

librz/util/buf.c
kazarmy: > Do we really need to add global variables?
> Is there no other way around it?

I'm open to suggestions.

> Also the setters for them are no where used?

The setters are in cconfig.c.

librz/include/rz_util/rz_buf.h
kazarmy: Ok 736baeace0f759590cafbb3e676028ddb8d0fff0

librz/util/buf.c
kazarmy: > Do we really need to add global variables?

1. You could pass the `Oxff` byte to be used, as a parameter, say:

```c
RZ_API st64 rz_buf_read(RZ_NONNULL RzBuffer *b, RZ_NONNULL ut8 RZ_OUT *buf, ut64 len, ut8 Oxff)
```

the problem here is that now it's the caller that decides what `Oxff` byte should be used, and that decision logic has to be repeated every time there is a call to `rz_buf_read()`. This is suboptimal for the following reasons:

1. Code has to be duplicated.
2. The caller can mess things up.

With (static) global variables, the `Oxff` logic is centralized in one place.

----

2. Why not depend on `core->config` then in RzBuffer code, since that's where the `io.0xff` and `io.ff` config vars live? Well, RzBuffer is a utility in `librz/util`, and utility code should be usable even without the Rizin core.

----

> Is there no other way around it?

The alternatives are worse in my opinion.

librz/bin/format/pe/pe_section.c
wargio: indeed. does windows run these bins? if yes, then maybe it is valid.

kazarmy: even if windows ran those bins, it doesn't mean the sections are valid and should be on the `iS` list ... windows could just be ignoring invalid sections ... btw, pe-bear 0.7.1 doesn't recognize the sections in `bins/pe/virtsectblXP.exe`, so I suppose rizin should ignore those invalid sections as well

kazarmy: i do really think the question of what is an invalid pe section and what is valid is beyond the scope of this pr

librz/util/buf.c
Rot127: The `io.0xff` config var is a property of RzIO.
So it also just affects buffers which are a wrapper around RzIO. Is this correct?

Maybe we can give the `RzBuffer` object the pointer to the underlying RzIO config value?

Sure, if the RzIO object is freed before the buffer this is undefined behavior (if the pointer is not NULLed). But this should never be the case. Because the `LIFETIME(RzIO) > LIFETIME(RzBuffer(RzIO))`, otherwise the buffer cannot read at all.

librz/bin/format/pe/pe_section.c
kazarmy: resolved in #6397

librz/bin/format/pe/pe_section.c
kazarmy: resolved in #6397

librz/util/buf.c
kazarmy: > So it also just affects buffers which are a wrapper around RzIO. Is this correct?

Sounds good to me.

> Maybe we can give the `RzBuffer` object the pointer to the underlying RzIO config value?

Ok f0394dc3012de3b313d4b592a55d77e11a180545 c05f3f1d2e2806712780ca4bb039652d7ceea095 32d5175173c413928d69d5c2e57737d3b9a9d505 c366caba7258d8c257263c73124a75fef3f4ed82 105a056b09e1fbe75e27683730dc3e6357fb5e6f

test/db/formats/pe/foldedhdr
kazarmy: This issue is no longer relevant due to 363d8735b63d4d255d17f97c73e85558499af43b.

wargio (approved):

Rot127 (approved):

notxvilka (approved):

PR #5627, opened 2025-12-16
Optimize queue to act also as proper channel

wargio:
**Your checklist for this pull request**
- [ ] I've read the [guidelines for contributing](https://github.com/rizinorg/rizin/blob/dev/CONTRIBUTING.md) to this repository.
- [ ] I made sure to follow the project's [coding style](https://github.com/rizinorg/rizin/blob/dev/DEVELOPERS.md#code-style).
- [ ] I've documented every `RZ_API` function and struct this PR changes.
- [ ] I've added tests that prove my changes are effective (required for changes to `RZ_API`).
- [ ] I've updated the [Rizin book](https://github.com/rizinorg/book) with the relevant information (if needed).
- [ ] I've used AI tools to generate fully or partially these code changes and I'm sure the changes are not copyrighted by somebody else.

**Detailed description**

Implements properly, sync mechanisms that are part of the queue feature in rizin.

PR #6383, opened 2026-05-18
Fix deadlock in rz_th_queue_close_when_empty()

thestr4ng3r:

**Detailed description**

empty_cond was never signalled and the termination depended only on the timeout in rz_th_queue_close_when_empty() causing a re-check of emptiness.
However, the rz_th_cond_timed_wait() implementation, which was used there, was flawed because it expected a relative timeout but passed that directly to pthread_cond_timedwait() which expected an absolute time value, practically causing it to time out immediately. Depending on the pthread_cond implementation, this possibly created a situation where the mutex could never be acquired by another thread, effectively causing a deadlock. This behavior was observed on Mac OS X 10.5 (ppc) when running the test_core_bin test.
We solve this by not using a timeout at all and signalling the condition variable at the appropriate time.

Since there are no other users of `rz_th_cond_timed_wait()` and making it work reliably requires a bit more work (see e.g. https://git.sr.ht/~thestr4ng3r/chiaki/tree/ab55cf4c95f7723faae08a546fbe14e0f6bddf45/item/lib/src/thread.c#L228-283 for reference) I decided it's probably best to remove it until there is a real need for it.

**Test plan**

All tests should still pass and test_core_bin should reliably run (not hang) on Mac OS X 10.5 (ppc).

### DIFF
--- a/librz/util/thread_queue.c
+++ b/librz/util/thread_queue.c
@@ -19,7 +19,7 @@ struct rz_th_queue_t {
 	RzThreadCond *reader_cond; ///< Cond for readers
 	size_t reader_awaiting; ///< Number of readers awaiting to read
 
-	RzThreadCond *empty_cond; ///< Cond any thread waiting for an empty queue
+	RzThreadCond *empty_cond; ///< Cond for predicate `closed || rz_list_empty(list)`
 
 	RzThreadLock *data_lock; ///< Lock used to modify the RzThreadQueue data
 	RzThreadQueueSize max_size; ///< Max queue size or unlimited
@@ -123,27 +123,23 @@ RZ_API RZ_OWN RzThreadQueue *rz_th_queue_from_pvector(RZ_NONNULL RZ_BORROW RzPVe
 /**
  * \brief  Closes a RzThreadQueue but only when empty (once closed you cannot read/write data).
  *
+ * This function will block until the queue is empty and then close it.
+ * If the queue is closed from somewhere else, it returns early, even if the queue is not empty.
+ *
  * \param  queue The RzThreadQueue to close
  */
 RZ_API void rz_th_queue_close_when_empty(RZ_NONNULL RzThreadQueue *queue) {
 	rz_return_if_fail(queue);
-
 	rz_th_lock_enter(queue->data_lock);
-	if (queue->closed) {
-		// already closed.
-		rz_th_lock_leave(queue->data_lock);
-		return;
-	}
-
-	while (!rz_list_empty(queue->list)) {
+	while (!queue->closed && !rz_list_empty(queue->list)) {
 		// the list is not empty, so we wait for it to be empty.
-		rz_th_cond_timed_wait(queue->empty_cond, queue->data_lock, 100);
+		rz_th_cond_wait(queue->empty_cond, queue->data_lock);
 	}
-
-	// we finally close the queue & notify all awating readers
-	queue->closed = true;
-	rz_th_cond_signal_all(queue->reader_cond);
-
+	if (!queue->closed) {
+		// we finally close the queue & notify all awating readers
+		queue->closed = true;
+		rz_th_cond_signal_all(queue->reader_cond);
+	}
 	rz_th_lock_leave(queue->data_lock);
 }
 
@@ -159,6 +155,7 @@ RZ_API void rz_th_queue_close(RZ_NONNULL RzThreadQueue *queue) {
 	if (!queue->closed) {
 		queue->closed = true;
 		rz_th_cond_signal_all(queue->reader_cond);
+		rz_th_cond_signal_all(queue->empty_cond);
 	}
 	rz_th_lock_leave(queue->data_lock);
 }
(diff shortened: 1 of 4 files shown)


librz/util/thread_queue.c
```diff
@@ -274,6 +274,9 @@ RZ_API bool rz_th_queue_pop(RZ_NONNULL RzThreadQueue *queue, bool tail, RZ_NONNU
 	} else {
 		*data = rz_list_pop_head(queue->list);
 	}
+	if (rz_list_empty(queue->list)) {
+		rz_th_cond_signal(queue->empty_cond);
```
wargio: this is not guaranteed that all the threads will receive it.

wargio:
with this code i have the opposite issue, where some threads were too slow to get the signal. keep the timeout signal one.

thestr4ng3r:
> with this code i have the opposite issue, where some threads were too slow to get the signal. keep the timeout signal one.

@wargio Do you have a concrete way to reproduce this? I think it should be solvable.

wargio:
The issue is quite simple, if you want all the thread to receive its cond notification then they need to be in the wait state, but if they are not in such state, they will not get the message and wait forever. The [new implementation of the queue](https://github.com/rizinorg/rizin/pull/5627) behaves as a channel and we cannot guarantee that the consumer is in such state.

wargio:
If as you mention, the waiting time on ppc macos is absolute, so we need to fix the implementation of the timed wait

thestr4ng3r:
> The issue is quite simple, if you want all the thread to receive its cond notification then they need to be in the wait state, but if they are not in such state, they will not get the message and wait forever. The [new implementation of the queue](https://github.com/rizinorg/rizin/pull/5627) behaves as a channel and we cannot guarantee that the consumer is in such state.

@wargio This is actually not the whole story. When you use a condition variable, you almost always associate it with some predicate that it is signalled for. In this case the predicate is `rz_list_empty(queue->list)` and it is guarded by the `queue->data_lock` mutex.
Threads will only enter the waiting state if the predicate is still false, and it is guaranteed by the mutex that there is no race between some other thread emptying the queue and entering the waiting state.

So in the event that the queue gets empty and a single other thread has started to call `rz_th_queue_close_when_empty()` at some point, there are two cases:
* The queue is already empty at the time rz_th_queue_close_when_empty() reaches the waiting code, in which case it will detect that the predicate `rz_list_empty(queue->list)` is already true and never try to wait on the cond
* The thread waiting for the queue to be empty is currently waiting on the condition variable. It is guaranteed to receive the event in that case because of how condition variables are specified to behave.

Now there is another aspect, which is the case of multiple threads calling `rz_th_queue_close_when_empty()` on the same queue. We could either just specify that `rz_th_queue_close_when_empty()` must be called only by one thread, or we can solve it using either `pthread_cond_broadcast()` or chaining signalling the condition variable once more after waking up.

Does that solve your issues?

wargio:
as i said before the issue is the same, as we would need to add a new lock just to make the wait unique. the idea was that multiple threads can wait at the same time via mutex+cond

We could resolve with the additional lock as the other threads would be waiting for the one who has it to release it.

thestr4ng3r:
@wargio I am not sure I get the full context. Are your concerns only about the case of more than one thread waiting for the emptiness?

wargio:
yes.

thestr4ng3r:
Ok, that makes it clear. I will try to find a good solution for it. I am pretty sure we can find one without an extra mutex, either using broadcast or chained signalling.

wargio:
i'm ok if we allow only one waiting thread by adding an additional lock before we call the cond wait.

This lock needs to be acquired before checking for the list length.

thestr4ng3r:
@wargio I was able to solve it by just using `rz_th_cond_signal_all()`. The added test quite reliably deadlocks on my local machine with `rz_th_cond_signal()` but works fine with `rz_th_cond_signal_all()`.

librz/util/thread_queue.c
```diff
@@ -274,6 +274,9 @@ RZ_API bool rz_th_queue_pop(RZ_NONNULL RzThreadQueue *queue, bool tail, RZ_NONNU
 	} else {
 		*data = rz_list_pop_head(queue->list);
 	}
+	if (rz_list_empty(queue->list)) {
+		rz_th_cond_signal_all(queue->empty_cond);
```
Rot127, line 275: The signalling also has to happen when the queue is closed via `rz_th_queue_close_when_empty` or `rz_th_queue_close`

thestr4ng3r: @Rot127 Can you elaborate more on why you think this is the case?
It should only be necessary to signal the condition variable when the state of its predicate `rz_list_empty(queue->list)` transitions from false to true, and that is only be the case in `rz_th_queue_pop()`.

wargio: because normally you would need to signal the thread when you close the queue.

Rot127: @thestr4ng3r Because two threads could theoretically try to close the queue.
If one calls `rz_th_queue_close_when_empty` and waits on the `empty_cond` (releasing the `data_lock`), the second thread could close the queue with `rz_th_queue_close`.
If it doesn't signal the the `empty_cond`, the thread calling `rz_th_queue_close` would wait forever for the other thread to join (which still waits on the `empty_cond`).

Or did I miss something?

thestr4ng3r: @Rot127 Indeed, if we allow calling `rz_th_queue_close()` on the same queue that `rz_th_queue_close_when_empty()` is called, then the only sensible outcome is for the `rz_th_queue_close_when_empty()` to return early even though the queue is not yet empty. I have modified the code so the predicate for `empty_cond` is now `closed || rz_list_empty(list)` and added a test for this case.

Rot127 (approved):

PR #6395, opened 2026-05-23
[stable] backport: Fix double free / reject invalid search.in values (#6327)

dkgkdfg65:
backport of `045fff363b` (PR #6327) to the stable branch.

the search command's `bytes_search_collection` is freed by the search-context cleanup and then re-freed in `byte_pattern_search` when a custom `search.in` value points at something the collection no longer owns. on dev this is a real heap-use-after-free under ASan; the fix moves `rz_core_get_boundaries` into the set-callback so invalid modes are rejected up front (`Invalid mode 'section' for [0x0, 0xffffffffffffffff]`) instead of triggering UAF at search time.

verified with rizin built at parent (`f1251293a4`) vs fix (`045fff363b`) with `-Db_sanitize=address` in ubuntu:22.04:
  pre-fix: `rizin -qc 'e search.in=section; /x deadbeef' malloc://1024` -> `AddressSanitizer: heap-use-after-free READ at bytes_search.c:51`, exit 99.
  post-fix: rejected at set-time, exit 0.

clean cherry-pick except for one additive conflict in `test/db/cmd/cmd_search_x` (stable didn't have the surrounding invalid-utf8 test; resolved by accepting the two added test cases verbatim, since they were the new tests for this fix).

CVE-2026-45324 (already public, fix on dev). per SECURITY.md this isn't classed as a security issue, but figured it was worth backporting since stable users hit it with crafted `search.in` values.

### DIFF
--- a/librz/core/cbounds.c
+++ b/librz/core/cbounds.c
@@ -704,6 +704,75 @@ RZ_API RZ_OWN RzList /*<RzIOMap *>*/ *rz_core_get_boundaries_debug_program(RZ_NO
  * \brief      Returns a list of boundaries (as RzIOMap), based on the selected mode; see [search/analysis/zoom/[in/from/to]] for available modes.
  *
  * \param      core      The RzCore to use
+ * \param      interval  The address range of the boundaries.
+ * \param      region    The region of the boundaries. Check valid ones with `e search.in=?`.
+ *
+ * \return     On success a valid pointer (can be an empty list), otherwise NULL
+ */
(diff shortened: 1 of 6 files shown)


Rot127 (approved):

Rot127:
I am not sure if we will merge this though, until we finalize the release.

Rot127:
Never mind. You seem to be a bot.
image

wargio:
these prs are useless as we will do this upon release

Issue #5092, opened 2025-04-20
`Float16` and `Float128` support


notxvilka:
- `fp16` is commonly used in AI tasks
- `fp128` is often used in scientific computations

Support both in:
- [ ] `t` (types) commands
- [ ] `pf` (print format) commands
- [ ] analysis steps

PR #6410, opened 2026-05-26
Rewrite the `pf` parser and rework the grammar

### LINKED ISSUE #782: Colorize the `pf` output (especially in `rizin -nn binary` case)
Currently rizin prints both `pf` and built-in `pf` output in `rizin -nn` mode without any colors, even if they are enabled. It would be amazing to improve the readability to colorize them according to the current color options and colorscheme.

image

See the `pf` implementation in `librz/core/cmd_print.c` and `librz/type/format.c`

I think it's better to do only **AFTER** `librz/core/cmd_print.c` is completely ported to the newshell.

### LINKED ISSUE #783: `pf` enhancements
- [ ] Support `pf.` nested struct/union name autocompletion (currently works only for the 1st level)
 - [ ] Support bitfields (they are supported in `t` command parsing but not in the `pf` syntax, thus printing it from `tp` command is impossible since it uses `pf` commands for that.
 - [ ] Add custom terminator support (for strings, etc)
 - [ ] Improve (and sort, may be group) help for the command
 - [ ] Better API to load/store `pf.` formats
 - [ ] Add two-byte  and four-byte octal value format specifiers.
 - [ ] Rethink on the `pf` syntax to be more consistent and harmonious

See the `pf` implementation in `librz/core/cmd/cmd_print.c` and `librz/type/format.c`

### LINKED ISSUE #905: Rewrite `pf` formats parser
Current parser located in `librz/type/format.c` is a total mess and hard to understand/maintain. It would be beneficial to port this grammar to the tree-sitter too. It will allow to change/extend the syntax easier too.

Probably makes sense to do together with https://github.com/rizinorg/rizin/issues/783

notxvilka:
# SQUASH ME

(I tried to separate commits initially but it is hard to keep them working individually given how pervasive `pf` formats in the Rizin)

# Detailed description

This series replaces `librz/type/format.c` (the legacy single-character
`pf` engine) with a layered parser / reader / renderer pipeline under
`librz/type/pf_parser*.c`, exposed through a new public header
`<rz_pf.h>`. The user-visible command surface (`pf`, `pfj`, `pfo`,
`pfd`, `pfc`, `pfn`, `pfs`, `pfv`, `pfl`, `pf.`, `pf*`, `tp`, `tpx`,
`avgp`, `tu`, `tuc`) stays identical; what changes is the DSL accepted
to the right of `pf` and the quality of the output and diagnostics.

### Cheatsheet

image

[PDF version](https://github.com/user-attachments/files/28337256/pf-cheatsheet.pdf)

### Sized integers with explicit endianness

The old `xdoub` codes had implicit, context-dependent width and
endianness. The new spec is explicit: a base letter (`x` `d` `u` `o`
`b`) plus a size digit (`1`/`2`/`4`/`8`), with case selecting endianness
(lowercase = LE, uppercase = BE). Plus a new `n{N}` code for fields
whose endianness genuinely follows the surrounding context (e.g. ELF
header fields driven by the file's data byte).

```text
pf "x4 magic"       # 32-bit hex, little-endian
pf "X4 magic"       # 32-bit hex, big-endian
pf "u8 size"        # 64-bit unsigned LE
pf "D2 count"       # 16-bit signed BE
pf "n4 entry"       # 32-bit hex, context endianness
```

### Per-base rendering

Same bytes, different display: `b` (binary), `d` (decimal), `o` (octal),
`x` (hex). Both LE and BE for each.

```text
pf "x4d4o4b4 hex dec oct bin"   # one word, four interpretations
```

### Strings with explicit encoding and optional length prefix

```text
pf "z greeting"             # NUL-terminated, UTF-8 default
pf "z(8bit) greeting"       # single-byte / Latin-1 passthrough
pf "z(utf16le) text"        # UTF-16 little-endian
pf "z(ebcdicus) host"       # EBCDIC US (mainframe)
pf "z[1] greeting"          # 1-byte LE length prefix, then body
pf "z[2] greeting"          # 2-byte LE length prefix
pf "s pointer"              # pointer to NUL-terminated string
```

Supported encoding names (parsed by `rz_str_enc_string_as_type`):
`guess`, `ascii`, `8bit`, `utf8`, `mutf8`, `utf16le`, `utf16be`,
`utf32le`, `utf32be`, `ibm037`, `ibm290`, `ebcdices`, `ebcdicuk`,
`ebcdicus`, `settings`.

#### Length-prefixed string example (Pascal-style)

Input bytes `05 48 65 6c 6c 6f`:

```text
$ rizin -qc 'wx 0548656c6c6f; pf "z[1] greeting"' =
0x00000000 : greeting = "Hello"
```

The `[1]` declares a 1-byte LE length prefix. `z[2]` would consume a
2-byte LE length and then the body.

#### EBCDIC US example (mainframe-style)

Input bytes `c8 85 93 93 96 81 40 40 40 40` (EBCDIC for `"Helloa    "`,
where `40` is the EBCDIC space):

```text
$ rizin -qc 'wx c8859393968140404040; pf "z(ebcdicus) hello"' =
0x00000000 : hello = "Helloa    "
```

### Parameterised timestamps

The old engine had `t` (any kind of seconds-since-epoch) and `T`
(deprecated). The new engine takes a `t(format)` parameter:

| Format        | Width | Description                                 |
|---------------|-------|---------------------------------------------|
| `unix32`      | 4 B   | Unix seconds since 1970 (32-bit)            |
| `unix64`      | 8 B   | Unix seconds since 1970 (64-bit)            |
| `unixms`      | 8 B   | Unix milliseconds                           |
| `unixus`      | 8 B   | Unix microseconds                           |
| `unixns`      | 8 B   | Unix nanoseconds                            |
| `filetime`    | 8 B   | Windows FILETIME (100-ns ticks since 1601)  |
| `dos`         | 4 B   | DOS date/time (16+16 bits)                  |
| `hfs`         | 4 B   | HFS+ seconds since 1904-01-01               |
| `oletime`     | 8 B   | OLE automation date (IEEE-754 double, days) |
| `webkit`      | 8 B   | WebKit microseconds since 1601              |
| `cocoa`       | 8 B   | Cocoa NSDate seconds since 2001             |

### GUIDs in three layouts

```text
pf "G uuid"          # 16-byte GUID, Microsoft mixed-endian (default)
pf "G(be) uuid"      # RFC 4122 big-endian network order
pf "G(le) uuid"      # all little-endian
```

### Inline bitfields

No typedb registration required for one-off bitfields:

```text
pf "B4(R=1,W=2,X=4) perm"            # 4-byte field, three flags
pf "B1(LSB=1,MID=2,MSB=4) byte"      # works on any size
```

### Bit fields and alignment

`:N` reads N bits from a packed byte. Default ordering is MSB-first;
append `<` for LSB-first. Adjacent `:N` fields share a bit cursor
within the current byte; when a non-bit field follows, the cursor
snaps to the next byte boundary.

```text
$ rizin -qc 'wx ab; pf ":3:5 high low"' =
0x00000000 : high = 5 (3-bit)
0x00000000 :  low = 11 (5-bit)
```

Byte `0xab` = `0b10101011`. MSB-first: `:3` = `0b101` = 5, `:5` =
`0b01011` = 11.

LSB-first reverses the slicing order within the byte:

```text
$ rizin -qc 'wx ab; pf ":3<:5< high low"' =
0x00000000 : high = 3 (3-bit)
0x00000000 :  low = 21 (5-bit)
```

`:3<` = `0b011` = 3, `:5<` = `0b10101` = 21.

`@N` advances to the next N-byte-aligned offset, consuming no name:

```text
pf "@4 x4 aligned"   # advance to next 4-byte boundary, then read
```

### TLV records

Tag-Length-Value records with five configuration keys (`t`/`l`/`e`/
`h`/`d`):

```text
pf "V record"                    # defaults: t=u1, l=u2 LE, value=raw bytes
pf "V(t=u2,l=u2,e=be) hdr"       # u2 BE tag and length
pf "V(t=u1,l=u4) chunk"          # 1-byte tag, 4-byte LE length
pf "V(t=u1,l=u2,d=mytab)"        # dispatch via tlv.mytab.<hex-tag>
```

#### Concatenated TLVs example

Three concatenated TLV records, each with a 1-byte tag and a 2-byte
big-endian length:

```text
$ rizin -qc 'wx 01000652697a696e21 020004010203 04050008deadbeefcafebabe; \
             pfn tlvs "V(t=u1,l=u2,e=be)V(t=u1,l=u2,e=be)V(t=u1,l=u2,e=be) \
                       hello version checksum"; pf tlvs' =
0x00000000 :    hello = tag=0x1 len=6
  52 69 7a 69 6e 21
0x00000009 :  version = tag=0x2 len=4
  01 02 03 04
0x00000010 : checksum = tag=0x5 len=8
  de ad be ef ca fe ba be
```

### Arrays

```text
pf "[8]x1 bytes"          # fixed-count array of 1-byte hex
pf "[5]d4 numbers"        # fixed-count array of 32-bit signed LE
pf "[4]U2 values"         # fixed-count array of 16-bit unsigned BE
pf "u4[@count]u1 count payload"    # length-by-reference: u1 N-tuple
```

#### Length-by-reference example

`[@count]u1` reads N `u1`s where N comes from the earlier `count` field:

```text
$ rizin -qc 'wx 04000000aabbccdd; pf "u4[@count]u1 count payload"' =
0x00000000 :   count = 4 [LE]
0x00000004 : payload = [ 170, 187, 204, 221 ] [LE]
```

#### Multi-base array example

Same eight bytes shown as a fixed-count hex array:

```text
$ rizin -qc 'wx deadbeefcafef00d; pf "[8]x1 bytes"' =
0x00000000 : bytes = [ 0xde, 0xad, 0xbe, 0xef, 0xca, 0xfe, 0xf0, 0x0d ] [LE]
```

### Short-offset mode (`scr.pf.short`)

A new `scr.pf.short` config option (default `false`) re-renders
MODE_TEXT offsets as deltas from the format base instead of absolute
addresses. Useful for self-contained struct dumps where the load
address is just noise:

```text
$ rizin -e scr.pf.short=true -qc 'wx 00007a452a4b9a02; pf fcb1d4 a b c d' =
   0 : a = 4000 [LE]
  +4 : b = '*'
  +5 : c = 0b0100_1011 [LE]
  +6 : d = 666 [LE]
```

Nested struct children reset to `0` inside their own block; trailing
fields after the block continue from the outer base. The
`RzPfRenderOpts::base_offset` API field lets programmatic callers pin
the base to any value they want.

### `pfw` write mode

The legacy parser implemented `pfw` by *emitting rizin commands*
(`wv8 42 @ 0x...\n`) into its output, which a `.` prefix would then
execute. The rewritten parser ports `pfw` to a direct I/O write
path: parse the format, locate the target field by name, encode the
value (honouring the field's endianness and string encoding), call
`rz_io_write_at`. After the write, the slot is re-read and a
right-aligned confirmation line is emitted:

```text
[0x00000000]> pfw troll.hu 42
   hu : 0x00000000 = 42
[0x00000000]> pfw troll.swagg "B*A AA"
swagg : 0x00000004 = "B*A AABLABLA"
[0x00000000]> pfw troll.plop 4000
 plop : 0x00000011 = 4000
```

The string write supports both UTF-8 (raw byte write) and the
wide encodings (`Z` / `z(utf16le)` / `z(utf16be)`) via
`rz_str_utf8_to_utf16`. Scalars use `rz_pf_field_size` to pick the
correct byte width and the field's recorded LE/BE/CTX flag for the
byte order. Unknown field names and missing values produce single
actionable error messages rather than silent fall-throughs.

### Enum symbol resolution

`pf E (name)field` looks up the enum-member name via the typedb at
render time and appends it as a `; SYMBOL` suffix. This matches the
legacy `pf` output that downstream users (and the type-inspection
workflow `pd $r @ elf_header`) depend on:

```text
[0x00000000]> pf. elf_header
0x00000000 :     ident = struct<elf_ident> {
   0x00000000 :   magic = "\x7fELF"
   0x00000004 :   class = 0x00000002 ; ELFCLASS64
   0x00000005 :    data = 0x00000001 ; ELFDATA2LSB
   0x00000006 : version = 0x00000001 ; EV_CURRENT
}
0x00000010 :      type = 0x00000003 ; ET_DYN
0x00000012 :   machine = 0x0000003e ; EM_X86_64
...
```

When no typedb is supplied at render time (e.g. a unit test calling
`rz_pf_format` with a NULL ctx), the symbolic suffix is omitted and
the numeric value stands alone -- the parser never crashes for lack
of a typedb. The typedb is passed via the new `RzPfRenderOpts.typedb`
field; the `pf` core integration populates it from the analysis
context, so users see the suffix without touching their config.

### Better diagnostics for `pf.` and `pfo`

Two existing UX papercuts addressed:

- **`pf.` with an unknown name** used to fall through to the raw
  parser, emitting one "unknown specifier" warning per letter. It
  now resolves the name (and any `name.field` prefix) against the
  typedb up front and emits a single actionable error if the name
  isn't registered:

  ```text
  $ rizin -nn -qc 'pf. elf32' = 
  ERROR: pf.: no registered format named 'elf32'. Run `pfn` to list
  known formats, or `pfo` to load a format definition file (e.g.
  `pfo elf64` then `pf. elf_header`).
  ```

- **`pfo` (no args)** used to list each file twice when the same FDF
  appeared in both the home and the system format directories. The
  listing is now de-duplicated.

### Positioned diagnostics

Errors now have a category (`SYNTAX` / `SEMANTIC` / `RANGE` /
`DEPRECATED` / `DATA` / `DEPTH`), a column position, and a severity.
Examples:

```text
$ rizin -qc 'pf "Q4 foo"' =
WARNING: pf: unknown specifier 'Q' at position 0, skipping
WARNING: pf: unknown specifier '4' at position 1, skipping

$ rizin -qc 'pf "f foo"' =
WARNING: pf: bare 'f' deprecated, use 'f4'
0x00000000 : foo = 0 [LE]

$ rizin -qc 'pf "V record"' =
WARNING: pf: V: bare 'V' uses defaults t=u1,l=u2 (no dispatch); add
  '(t=...,l=...,d=table)' to configure

$ rizin -qc 'pf "V(zoo=u2)"' =
WARNING: pf: V: unknown key 'zoo' (expected t, l, e, h, or d)

$ rizin -qc 'pfn rec "?(rec) inner"; pf rec' =
WARNING: pf: struct-size recursion limit at 'rec'
WARNING: pf: max depth for 'rec'
```

The recursion bound (default 32) prevents the legacy engine's stack
overflows on self-referential structs.

### Inline colorisation

When `scr.color` is enabled, text-mode output is colourised using the
active console palette (`pal.offset`, `pal.fname`, `pal.meta`,
`pal.num`, `pal.flag`). The theme is the one selected via `eco`, so
users get consistent colours across the whole UI.

### Bitvectors (`v(N)`) for forensics-friendly bitmap display

A new specifier `v(N)` reads N individual bits (1..4096) from
`ceil(N/8)` bytes of input and exposes them as N separate 0/1 scalars.
The text renderer prints them grouped by 8 with a `|` separator:

```
> wx abcd
> pf v(12) bits
0x00000000 : bits = [ 1 0 1 0 1 0 1 1 | 1 1 0 0 ] (12-bit)
```

JSON renders the vector as a compact string of `'0'`/`'1'` characters
plus a `bit_width` sibling; quiet mode emits the bits without
decoration. Bit order within each byte defaults to MSB-first and can
be flipped with `,lsb`/`,msb`. Width is clamped to `[1, 4096]` with a
diagnostic.

Use cases this addresses cleanly with one specifier instead of a
hand-rolled loop: page-frame allocation maps, NTFS `$Bitmap` clusters,
ext4 block/inode bitmaps, ELF `DT_FLAGS_1`, PE characteristics, ACL
bitmasks. Unlike `:N`, `v(N)` does not interact with the packed-bit
cursor -- each field stands alone and consumes whole bytes.

## What features/behavior was removed from DSL grammar

This is the deliberately-broken list. Each item has a documented
rationale or replacement.

- **Legacy anonymous nested struct `?(format inner_names)`** is no
  longer supported. The new engine implements only the named form
  `? (typename) name`. Reasoning: the inline form silently re-parses
  its argument as a sub-format and binds positional names from the
  parent's name list, which is ambiguous in the presence of arrays
  and TLV. The named form is unambiguous and lets typedb-stored
  formats be reused. The one test (`cmd_pf2 anon-nested-struct`) is
  marked `BROKEN=1` with a `# moved to ?(name) form` comment.
- **Bare single-letter scalars** `b d i o q u w x X D O f F Z` still
  parse, but the parser emits a one-line `WARNING: pf: bare 'X'
  deprecated, use 'X4'` diagnostic suggesting the sized form. The
  underlying width is `4` (or `8` for `q/Q`), preserving the legacy
  semantics for existing Cf metadata and FDF files. Authors of new
  formats should use `x4`/`u8`/`d4`/etc.
- **Implicit big-endian-by-spec-char `N`** (legacy `N{N}` for context-
  endian) is replaced by `n{1,2,4,8}` (lowercase) for context-endian
  scalars and the explicit `X1/X2/X4/X8` (uppercase) for forced
  big-endian. Old `N` parses with a deprecation warning.
- **The legacy DSL conversion shim** (`format.c`'s pre-parser that
  rewrote some specifiers before handing them to the legacy reader)
  is removed in commit 6. The new parser is the only path; no
  fallback.
- **Single-pass-only constructs** that the legacy reader implemented
  with side effects across the format pointer (the global `nargs`
  counter, the in-place format rewrite for `b` -> `x4`, the
  per-letter `is_struct`/`is_pointer` flags carried on the format
  pointer) are gone. The new engine has explicit AST nodes for
  arrays, pointers, and struct dereference; nothing is decided by
  scanning forward in the source string.

## What commands were removed or changed

No public command was removed. Behaviour changes:

- **`pfw`** -- previously printed the bytes it *would* write and
  asked the user to commit them manually. Now actually writes and
  reads back the slot to confirm, decoding the readback in the
  field's own type (so `pfw "z(utf16le)" foo` writes UTF-16 and
  reads it back as UTF-16, not as raw bytes). UTF-16 decode-readback
  is delegated to the same encoding path the reader uses.
- **`pf.<name>`** -- when `<name>` is not a registered format,
  previously printed nothing and returned silently. Now emits
  `ERROR: pf.: no registered format named 'X'. Run \`pfn\` to list
  known formats, or \`pfo\` to load a format definition file (e.g.
  \`pfo elf64\` then \`pf. elf_header\`).` -- with an actionable hint.
- **`pfo`** -- previously listed every Format Definition File it
  found, including duplicates across `$HOME/.config/rizin/format`
  and `$RZ_PREFIX/share/rizin/format`. Now deduplicates by basename;
  the home directory wins.
- **`pf *z`** -- previously read the pointed-to string from offset 0
  of the current segment, ignoring the seek address. Now reads from
  the current seek, which is what callers actually want when
  inspecting a struct field.
- **`pf. name[N]`, `pf. name[N].sub`, `pf. outer.inner`** -- new
  path-navigation grammar in the dot form, implemented by a new
  `pf_path_navigate` helper. The legacy parser handled only the
  flat case.
- **`scr.pf.short`** -- new config option (default off). When on,
  `pf` text mode prints `+<delta>` deltas from the seek address
  instead of full 16-hex-digit offsets. The full-offset form is
  preserved by default so existing test snapshots and scripts that
  scrape `pf` output keep working.

The new bare-letter deprecation warnings are emitted at `WARN`
severity, which is filtered out of normal CLI output (only `ERROR`
goes to stderr by default). They appear when `RZ_LOG_LEVEL` is set
to `info` or below, or when the test harness's `EXPECT_ERR` block
explicitly looks for them. So existing scripts that pipe `pf` output
to `grep` are not affected.

## What stays the same

- Public command surface: `pf`, `pfj`, `pfs`, `pfd`, `pfq`, `pf?`,
  `pfo`, `pfv`, `pfl`, `pf.`, `pf*`, `tp`, `tpx`, `avgp`, `tu`, `tuc`.
- Named-format storage in the typedb (`pf.<name>=...`) is unchanged
  on disk; existing typedb dumps continue to work.
- The deprecated single-character codes (`b`, `c`, `d`, `f`, `o`,
  `q`, `s`, `t`, `u`, `w`, `x`, `z`, `i`, `i:N`) still parse with a
  deprecation warning; the legacy DSL conversion shim is removed in
  the final commit so the new parser is the only path.
- Integration tests `cmd_pf{,2,_elf,_write,_dsl}`, `cmd_pfd`,
  `types_format`, `cmd_pFb`, `cmd_avg`, `types` -- preserved or
  improved (see numbers below).

## Architecture

The new engine is a three-stage pipeline. Each stage is its own public
API entry point:

```text
   parse                 read                       render
  +--------+         +------------------+         +---------------------+
  | source |  -->    | format -> values |  -->    | values -> string    |
  | string |         | (RzPfFormat,     |         | (text / json /      |
  |        |         |  RzPfValue[])    |         |  cstruct / quiet /  |
  +--------+         +------------------+         |  dot / write)       |
                                                  +---------------------+
```

| Stage      | Entry point     | Inputs                                    | Output         |
|------------|-----------------|-------------------------------------------|----------------|
| parse      | `rz_pf_parse`   | format string                             | `RzPfFormat *` |
| read       | `rz_pf_read`    | format, buffer, base addr, ctx            | `RzPfValue *`  |
| render     | `rz_pf_render`  | values, count, mode, opts                 | `char *`       |
| all-in-one | `rz_pf_format`  | source + buffer + ctx + mode + opts       | `char *`       |

See `librz/type/README.md` and `doc/pf.md` for the full reference.

Screenshot 2026-05-27 at 4 56 59 PM

Screenshot 2026-05-27 at 4 59 09 PM

Screenshot 2026-05-27 at 5 00 08 PM

Current `pf?` help:
```
[0x100000960]> pf?
Usage: pf[?]   # Print formatted data
╭ pf[jq] <format>          # Show data using given format string
│ pf- <formatname>         # Remove named format
│ pf-*                     # Remove all named formats
│ pfa <format>             # Apply format string at given address and define flags for each field
│ pfc <format>             # Show data using given format string with C syntax
│ pfd <format>             # Show data using given format string as DOT
│ pf.[jq] [<formatname>]   # Show data using given named format
│ pfn [<formatname> [<formatstring>]] # List named formats/Print named format string/Define a new named format
│ pfo [<file>]             # Load a Format Definition File (FDF)
│ pfs <format>             # Print the size of format in bytes
│ pfv <format>             # Print the value for named format
╰ pfw <format> [<value>]   # Write data using given format string

Sized integers (lowercase=LE UPPER=BE):
| x1 / x2 / x4 / x8 # hex unsigned, N bytes
| d1 / d2 / d4 / d8 # decimal signed, N bytes
| u1 / u2 / u4 / u8 # decimal unsigned, N bytes
| o1 / o2 / o4 / o8 # octal, N bytes
| b1 / b2 / b4 / b8 # binary, N bytes
| n1 / n2 / n4 / n8 # hex unsigned, N bytes, context-endian (follows ctx.big_endian; use for file headers whose endianness depends on the file's data byte, e.g. ELF)
| f2 / f4 / f8      # IEEE 754 float, 2/4/8 bytes (half/single/double)

Special scalars:
| c     # single byte rendered as character
| p     # pointer (size from ctx.bits: 2 / 4 / 8 bytes)
| Q     # uint128_t (16 bytes, byte-sequential)
| r     # raw hex byte dump (count via [N])
| U / L # ULEB128 / SLEB128 (variable length)

Strings (encoding-aware):
| z              # inline NUL-terminated string
| z(utf16le)     # encoding override (utf8, utf16le, utf16be, utf32le, utf32be, ibm037, ebcdic_us, ...)
| z[N]           # length-prefixed string, N-byte prefix (1/2/4/8), length is in characters
| z(utf16le)[2b] # length prefix in BYTES (MS BSTR style); suffix 'b' switches the unit
| s              # pointer to NUL-terminated string (dereferences via read_at)

Timestamps (parameterised):
| t(unix32) / T(unix32) # wire format inside the parens; case selects LE/BE
| supported formats     # unix32, unix64, unixms, unixus, unixns, filetime (alias ntfs), dos, hfs, oletime, webkit, cocoa

Typed composites:
| E (enum_type)     # 4-byte enum, name resolved via typedb
| B (bitfield_type) # 4-byte bitfield, name resolved via typedb
| B4(R=1,W=2,X=4)   # inline bitfield with named flags (size = 1/2/4/8 byte BN)
| ? (struct_type)   # nested struct, name resolved via typedb
| 0...              # leading '0' marks the format as a union (all fields share offset 0)

DSL extensions:
| @N                   # align cursor up to next N-byte boundary (no value, no name)
| :N                   # read N bits (1..64) from packed bitstream; MSB-first by default
| :N< / :N>            # explicit bit order: < = LSB-first, > = MSB-first
| G                    # 16-byte GUID/UUID, mixed-endian (MS) layout by default
| G(le) / G(be)        # GUID layout: all-LE or RFC 4122 BE
| V(t=u1,l=u2,d=table) # TLV record; t=tag, l=length, e=le/be, h=v/a (header inclusion), d=dispatch table
| [@field_name]T       # array whose length comes from an earlier scalar field

Skip / repeat / pointers:
| .      # skip 1 byte; [N]. skips N bytes
| 3...   # leading integer repeats the whole format N times
| {N}... # alternate repeat syntax
| *T     # field is a pointer to T (dereferenced via read_at)
| [N]T   # fixed-size array of N elements of T

Examples:
| pf 'x4 d2 u8 z magic ver size name' # header with hex u32, signed s16, unsigned u64, NUL-terminated string
| pf 'B4(R=1,W=2,X=4,KERN=0x100) perms' # inline bitfield with named flags
| pf 'u1 [@count] x4 count items'     # u8 count, then that many u32 hex values
| pf 'z(utf16le)[2b] bstr'            # BSTR-style: 2-byte length prefix counting bytes, UTF-16 LE body
| pf ':1 :7 x4 flag rest tail'        # bit-level: 1 bit then 7 bits, then 4 bytes
| pf 'G(be) uuid'                     # RFC 4122 big-endian UUID
| pf 'V(t=u1,l=u2,d=usb_desc) record' # TLV dispatched via the 'usb_desc' tag-to-format table
| pf 't(filetime) created'            # 8-byte Windows FILETIME timestamp
| pf '3 0x4d4 a b'                    # repeat 3 times, union with fields x4 and d4
| pf '? (mytype) field'               # nested struct of typedb format 'mytype'

Notes:
| bit order                     # :N defaults to MSB-first (matches DWARF and most network protocols); use <N for LSB-first
| endian via case               # lowercase specifier = little-endian, UPPERCASE = big-endian
| context endian                # n1/n2/n4/n8 follow the surrounding RzPfCtx.big_endian flag instead of being pinned by case
| skip vs alignment             # '.' / '[N].' skip an exact number of bytes; '@N' pads to the next N-byte boundary
| deprecation                   # bare-letter codes from the legacy parser (b, c, d, f, o, q, s, t, u, w, x, z) still parse with a deprecation warning; new code should use the sized forms
| pf 'X2D4u8 bigWord beef qword' # BE u16, BE s32, LE u64 -- one of every endianness/sign combination
| pfn foo 'rr (eax)reg1 (eip)reg2' # Create object foo referencing two registers
| pf 't(unix32)t(unix32) troll plop' # Print two unix-epoch (32-bit) timestamps with labels 'troll' and 'plop'
```

**Test plan**

- CI is green
- Play manually with these commands:
         - `pf`
         -  `tp`
         -  `avgp`

I used the custom (LLM-generated) harness for the property-based testing with the help of [Theft library](https://github.com/silentbicycle/theft), just like in https://github.com/rizinorg/rizin/pull/4996

Here it is: 

[pf-property-harness.zip](https://github.com/user-attachments/files/28339903/pf-property-harness.zip)

**Closing issues**

Closes https://github.com/rizinorg/rizin/issues/905
Closes https://github.com/rizinorg/rizin/issues/783
Closes https://github.com/rizinorg/rizin/issues/782
Partially addresses https://github.com/rizinorg/rizin/issues/5092


wargio:
wpuld be nice to have as output RzStructuredData

notxvilka:
> wpuld be nice to have as output RzStructuredData

Your wish is granted. I also split `pf_parser.c` and `pf_renderer.c` into smaller chunks, per used "feature" and output mode.

wargio (approved):
I think with this we could implement native visualization in Cutter for any `RzStructuredData` output :)

PR #6422, opened 2026-05-29
type: add a pf-format to C-declaration converter

### LINKED ISSUE #2268: Support defining `t` types from `pf` format strings, including from existing saved ones
There are currently two type format supported by rizin. The `pf` format strings, and the C11 syntax supported by `t`. While you can get a `pf` format string for a `t` type and even use `tp` to instantly use it to print (You will have to edit the string if you want it saved as a `pf.something`). There doesn't seem to be a way to quickly create a `t` type from a `pf` string. This matters because only types defined with `tl` do type analysis, defining types using `Cf` doesn't. (P.S. Is there any interaction between `tl` and `Cf`.

Support a command to quickly define a `t` type using a `pf` format string, using appropriate standard fixed sized types.

Do it manually. How fun.

Alternatively, there could be an even tighter integration between the two features.

notxvilka:

**Detailed description**

format.c can already turn an RzType into a pf format string. Add the inverse in the same file, next to its counterparts: `rz_type_format_to_c_declaration()` parses a pf format string with `rz_pf_parse()` and emits an equivalent C struct/union declaration built from the standard fixed-width types (`uint8_t`, `int32_t`, `float`, ...).

The conversion is structural: it consumes only the parsed `RzPfFormat` (field kinds, widths, array counts, pointer flags), never a byte buffer, so it runs without any target data. Every field kind the engine produces is mapped; a handful of specifiers with no exact static C form are mapped best-effort (documented inline): `@N` alignment is dropped, an unknown-length inline `z` string becomes `char *`, `LEB128` widens to its largest decoded integer, and `?(Name)` / `E(Name)` emit `struct`/`enum` references that must themselves be defined to parse.

Added new `tdf` command that exposes that conversion to the user

**Test plan**

- CI is green
- Try new `tdf` command

**Closing issues**

Closes https://github.com/rizinorg/rizin/issues/2268

### DIFF
--- a/librz/core/cmd_descs/cmd_type.yaml
+++ b/librz/core/cmd_descs/cmd_type.yaml
@@ -70,11 +70,31 @@ commands:
             summary: Remove all calling conventions
             args: []
   - name: td
-    cname: type_define
-    summary: Define type from C definition
-    args:
-      - name: type
-        type: RZ_CMD_ARG_TYPE_STRING
+    summary: Define types from a C definition or a pf format string
+    subcommands:
+      - name: td
+        cname: type_define
+        summary: Define type from C definition
+        args:
+          - name: type
+            type: RZ_CMD_ARG_TYPE_STRING
+      - name: tdf
+        cname: type_define_from_format
+        summary: Define a type from a pf format string or a saved pf.<name>
+        details:
+          - name: "Examples"
+            entries:
+              - text: "tdf"
+                arg_str: ' rgba "x1x1x1x1 r g b a"'
+                comment: "define struct rgba from an inline pf format"
+              - text: "tdf"
+                arg_str: " elf_header elf_header"
+                comment: "define a type from the saved pf.elf_header format"
+        args:
+          - name: name
+            type: RZ_CMD_ARG_TYPE_STRING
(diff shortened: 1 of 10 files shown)


wargio (approved):

librz/core/cmd_descs/cmd_type.yaml
```diff
@@ -70,11 +70,31 @@ commands:
...
+          - name: type
+            type: RZ_CMD_ARG_TYPE_STRING
+      - name: tdf
+        cname: type_define_from_format
+        summary: Define a type from a pf format string or a saved pf.<name>
+        details:
+          - name: "Examples"
+            entries:
+              - text: "tdf"
+                arg_str: ' rgba "bbbb r g b a"'
```
Rot127:
```suggestion
                arg_str: ' rgba "x1x1x1x1 r g b a"'
```

Rot127 (changes requested):

librz/core/cmd_descs/cmd_type.yaml
notxvilka: Good catch!

PR #6433, opened 2026-05-29
librz/type: parse, render and size C23 enum underlying types (#3498)

### LINKED ISSUE #3498: Support enhanced enumerations in C parser (input and output)
C recently introduced a way to specify the underlying type for enum:
```c
enum F : int { A = 0x8000 } f;
enum E : long long { R, G, B } e;
```
Depends on https://github.com/tree-sitter/tree-sitter-c/issues/132

notxvilka:

**Detailed description**
C23 lets an enum fix its underlying type, e.g. "enum E : long long { ... }". The C grammar already exposes it as the "underlying_type" field of an enum_specifier, and RzBaseType already has a "type" slot documented as used by enums, but the parser ignored the field and always left it NULL.

```c
enum F : int { A = 0x8000 } f;
enum E : long long { R, G, B } e;
```

- parse_enum_node() now reads the "underlying_type" field and stores the parsed type on RzBaseType::type (reusing parse_type_node_single(), so primitive, sized and typedef'd integer types are all handled). Classic enums keep a NULL underlying type.
- The pretty printer emits " : <type>" between the enum name and its body when an underlying type is present, so "tc"/"tcd"/"tec" round-trip it.
- enum_bitsize() now derives the width from the underlying type instead of the hardcoded 32-bit default (resolving the long-standing FIXME); it still falls back to 32 for a classic enum.

Single-token underlying types (int, uint64_t, char, ...) work end-to-end with the bundled grammar revision: the existing "enhanced enum" db test is updated to round-trip "enum v : int" and a unit test covers "enum EU : uint64_t". Multi-word underlying types (long long, unsigned int) are added as BROKEN db tests; the bundled grammar parses them to an ERROR node and drops the underlying type, so these tests fail for now and will pass once rizin-grammar-c accepts sized type specifiers as the enum underlying type. No further rizin change is needed for that step.

**Test plan**

CI is green

Closes #3498


librz/type/serialize_types.c
```diff
@@ -76,6 +76,20 @@ static TypeFormatPair *get_enum_type(Sdb *sdb, const char *sname) {
 	}
 	free(members);
 
+	// C23 fixed underlying type, stored by save_enum() under "enum.<name>.@type"
+	RzStrBuf utkey;
+	char *underlying = sdb_get(sdb, rz_strbuf_initf(&utkey, "enum.%s.@type", sname));
+	rz_strbuf_fini(&utkey);
+	if (underlying) {
+		char *error_msg = NULL;
+		RzType *ut = rz_type_parse_string_single(typedb->parser, underlying, &error_msg);
```
thestr4ng3r, line 90: Could there be a case where the types referred to by `underlying` have not yet been loaded because of the order they are read from the sdb?

thestr4ng3r (changes requested):

librz/type/serialize_types.c
notxvilka: Yeah, it's absolutely possible. Let me think about it, how it's best to handle such situations. It might work out of the box actually but I need to test it first (and add a test to avoid regressions in the future).

notxvilka: Yes, it was handled already, just added a test to make sure it works and doesn't break in the future.

PR #6434, opened 2026-05-30
librz/arch/tms320: improve the C55x+ analysis, add basic RzIL

notxvilka:
# DO NOT SQUASH!

# Overview

- First commit is the small fix to allow open TI COFF files properly
- Second commit implements RzIL for both C55x and C55x+ variants
- Third commit rewrites the old string-based disassembly/analysis engine (two separate ones - one for C55x and another for C55x+) and implements a new, byte-level one, with a shared infrastructure between both variants, allowing reusing analysis and RzIL uplifting code whenever possible.
- Fourth commit fixes remaining gaps - missing RzIL uplifting for remaining instructions to get complete coverage (except FIR, MAC, and similar instructions)

# Detailed description

All changes did after trying Rizin on the real TMS320C55x+ firmware, which is why these bugs were found in the first place.

**IMPORTANT**: There are 4 commits, one is the COFF changes; second one is the extending and fixing *existing* code, implementing RzIL and covering everything with tests; third one is the full refactoring to share the common code between C55x and C55x+, and, eventually - C54x; forth one is to cover gaps found during the testing the refactored engine.

### Disassembler correctness

Fixes to C55x+ operand decoding and instruction tables. On the 5 MB Wrigley3G
image ~99.7% of instructions decode; the residual ~0.3% is a small unhandled
opcode tail (notably the `0x24` prefix), scattered through code rather than data.

### Analysis plugin
- Register profile: `ac0-3` (40-bit), `ar0-7`/`xar0-7`, `t0-3`, `sp`, `pc`,
  and `st0_55` with the documented bit layout (TC1 = bit 13, TC2 = bit 12,
  CARRY = bit 11).
- Function prologue preludes for `aap` (the two-word `0e ?? 0e ??` and the
  three-push entry pattern), which recovers ~6970 functions on the image.
- Branch/call typing: `bcc`/`bccu` -> CJMP with jump/fail, `call` -> CALL,
  `b` -> JMP.

### RzIL lifter (`c55plus_il.c`) — *correct-or-NULL*

Every form below is lifted only when its semantics are exactly known; anything
else returns NULL (left unlifted) rather than being approximated. Over a 20k
instruction window, IL coverage is ~62-66%.

- **mov / copy**: immediate load; register copy (equal-width, low-bit narrowing
  e.g. `mov ac0, xar0`, and sign-extending widen from a 16-bit data register);
  accumulator half writes (`mov #k, ac0.l/.h`).
- **Data-memory addressing** (the bulk of the image — `mov`/`copy` are ~45% of
  all instructions): indirect `*arN`, post-inc/dec `*arN+`/`*arN-`,
  register-indexed `*arN(short(#K))`, sp-relative `*sp(#K)`, and k24 absolute
  `*(#K)`; with `byte()` (8-bit), default word (16-bit) and `dbl()` (32-bit)
  access sizes, and `uns()` zero-extension vs the default sign-extension.
- **Bitwise** `and`/`or`/`xor`/`not` (incl. the 3-operand same-register form).
- **Arithmetic** `add`/`sub` (16-bit and accumulator).
- **Shifts** `sfts` (arithmetic) / `sftl` (logical), 40-bit accumulator with a
  6-bit signed count.
- **Control flow**: unconditional `b` (JMP); conditional `bcc`/`bccu` lifted to
  `BRANCH(cond, JMP(target), NOP())`, where the condition is either a status
  flag (`tc1`/`tc2`/`carry` and their `!`-negations, read from `st0_55`) or a
  register/accumulator-half compared against an immediate (signed for `bcc`,
  unsigned for `bccu`); `call` lifted to a JMP to the target (the return-address
  save is intentionally not modelled — rizin's call/return analysis tracks
  linkage separately).

### Calling convention
`librz/arch/types/cc-tms320-32.sdb.txt` defines a `c55x` convention (the default
for `arch=tms320, bits=32`): argument registers AR0-AR3 then T0/T1, overflow on
the stack, return in T0. This is a positional approximation of the
type-segregated TI C55x ABI; AR5 is callee-saved (consistent with the firmware's
`psh dbl(xar5)` prologues), though callee-saved registers are not expressible in
the simple cc format.

## Verification highlights (RzIL VM execution)

- `mov *ar5(short(#0x8)), ac0.l` with `xar5=0x10` and word `0xabcd` at byte
  `0x30` -> `ac0 = 0x000000abcd` (added as an `rzil/tms320` regression test).
- `bccu #t, ac1.l < #0x3`: branch taken (VM PC -> jump) when `ac1.l=1`, falls
  through when `ac1.l=5`.
- `bcc #t, !tc1`: falls through when `st0_55` bit 13 is set, branches when clear.
- The pre-existing word load/store + post-increment + accumulator execution
  tests are unchanged and still pass.

## Known semantic caveat — pointer mode

The lifter models data addressing in **word-pointer mode**: a register base
(XARn/SP) is a 23-bit word address (byte address = base << 1), the K16
displacement is added in word units (`(base + K) << 1`, keeping word/dbl
accesses aligned), and the k24 absolute constant is a direct 24-bit byte
address. The C55x+ also offers a **byte-pointer mode** (the manual's 'Ryujin'
byte addressing for C `char` data); the prevalence of `byte()` register-relative
accesses and odd k24 byte addresses in this firmware is suggestive of byte-pointer
mode at runtime, in which case register-relative effective addresses are scaled
by an extra factor of 2. This is the single unverified addressing assumption and
is localised to one helper (`c55_mem_byte_addr`), so it can be flipped in one
place if the firmware's runtime ST configuration is pinned.

# Test plan

- CI is green

It works on the real baseband firmware:

Screenshot 2026-06-07 at 1 21 46 AM
Screenshot 2026-06-07 at 1 23 56 AM
Screenshot 2026-06-07 at 1 24 38 AM
Screenshot 2026-06-07 at 1 21 46 AM
Screenshot 2026-06-07 at 3 02 20 AM
Screenshot 2026-06-15 at 4 29 26 PM


wargio (approved):

notxvilka:
It's incomplete yet. As I am in the middle of the RE session using Rizin, I will improve it as I go. Will mark it ready after all necessary fixes pushed.

PR #6439, opened 2026-05-30
librz/type: support C bitfield members in structs and unions

### LINKED ISSUE #1240: Add proper bitfields support in RzType

notxvilka:

**Detailed description**

The C grammar already exposes a struct/union member's bitfield width as a "bitfield_clause" node and the parser already recognized "int a : 4;", but it threw the width away (the member was stored with size 0 and a FIXME). As a result "tc" rendered the member as a plain "int a;", "ts" produced a full-int "pf" format ("pf d4d4d4 a b c") and "tp" mis-read every field as a whole integer.

RzTypeStructMember and RzTypeUnionMember already carry a "size" field documented as "in bits"; it is now used as the bitfield width, where 0 means the member is not a bitfield:

- The struct and union member parsers store the parsed bit width on the member instead of discarding it.
- The pretty printer emits " : <width>" before the member's trailing ';', so "tc"/"tcd" round-trip "struct qwe { int a : 4; int b : 16; int c : 3; }".
- rz_base_type_as_format() (used by "ts"/"tp") now emits the pf packed-bits spec ":N" for a bitfield member, with the bit order matching the target endianness ("<" little-endian / ">" big-endian), instead of the member type's full format. "ts qwe" becomes 'pf ":4<:16<:3< a b c"' and "tp" unpacks the fields correctly (a=5, b=0x1234, c=0 for 0x00012345 on x86).

Serialization keeps the width in the 3rd comma-field of the existing "struct.<name>.<member>" / "union.<name>.<member>" value, i.e. "type,offset,bitsize". That field was already written (always as 0) and ignored on load, so the on-disk layout is unchanged and only its meaning is now honored. The project version is bumped to 24 with an additive no-op migration: an absent or 0 bitsize deserializes as a regular, non-bitfield member.

Tests: a unit test parses a bitfield struct and union and checks the member widths, the rendered string and the "pf" format; a db/cmd/types test covers the full "td"/"tcd"/"ts"/"tp" flow; a project-migration test with a new v23-bitfield.rzdb fixture covers v23->v24, and the "migrate info" db test is updated for the new version.

The PDB type parser (librz/arch/pdb_process.c) still drops bitfield members because their LF_BITFIELD member type resolves to NULL; wiring LF_BITFIELD up to the new member "size" field, with tests in test/unit/test_pdb.c, is left as a follow-up.

**Test plan**

CI is green

**Closing issues**

Closes https://github.com/rizinorg/rizin/issues/1240


librz/type/parser/types_parser.c
```diff
@@ -556,7 +556,7 @@ int parse_struct_node(CParserState *state, TSNode node, const char *text, Parser
 					.name = membname,
 					.type = rz_type_clone(membtpair->type),
 					.offset = 0, // FIXME
-					.size = 0, // FIXME
+					.size = bits, // bitfield width in bits, 0 if not a bitfield
```
thestr4ng3r: Update the docs of the field in `librz/include/rz_type.h` instead of a comment here.

librz/type/format.c
```diff
@@ -140,6 +140,16 @@ static void base_type_to_format_unfold(const RzTypeDB *typedb, RZ_NONNULL RzBase
 	case RZ_BASE_TYPE_KIND_STRUCT: {
 		RzTypeStructMember *memb;
 		rz_vector_foreach (&type->struct_data.members, memb) {
+			if (memb->size && memb->size < rz_type_db_get_bitsize(typedb, memb->type)) {
+				// C bitfield member (size set and narrower than the member's type; a width
+				// equal to the type size, as stored by loaders like DWARF for plain
+				// members, is not a bitfield): emit the pf ":N" packed-bits spec with the
+				// bit order matching the target endianness ('<' little, '>' big).
```
thestr4ng3r: Extract this condition to an api-accessible function on the `RzTypeStructMember`.

Also instead of checking `memb->size < rz_type_db_get_bitsize(typedb, memb->type)`, should we perhaps make all loaders like dwarf set it to 0 for non-bitfield members?
Since serialization previously only wrote 0, we should not have any issues making this change, should we?

librz/type/serialize_types.c
```diff
@@ -217,17 +222,22 @@ static TypeFormatPair *get_union_type(RzTypeDB *typedb, Sdb *sdb, const char *sn
...
 			char *error_msg = NULL;
 			RzType *ttype = rz_type_parse_string_single(typedb->parser, value, &error_msg);
 			if (!ttype || error_msg) {
 				free(values);
 				goto error;
 			}
 
+			// the 3rd comma-field (after the union member offset) is the
+			// bitfield width in bits, 0 if the member is not a bitfield.
+			const char *member_bitsize = member_rest ? strchr(member_rest, ',') : NULL;
```
thestr4ng3r: Why not use `sdb_anext()` here too?

thestr4ng3r (changes requested):

librz/type/format.c
notxvilka: Yes, that would work.

librz/type/base.c
```diff
@@ -326,6 +326,32 @@ RZ_API RZ_OWN char *rz_type_db_base_type_as_pretty_string(RZ_NONNULL const RzTyp
...
 
+/**
+ * \brief Returns true if the struct member is a C bitfield
+ *
+ * A member is a bitfield exactly when a non-zero width was recorded in its
+ * \p size field (in bits). Producers that do not deal in bitfields (e.g. the
+ * DWARF loader for a plain member) leave \p size at 0.
+ *
+ * \param member struct member to inspect
+ */
```
thestr4ng3r: This documentation is too verbose imo and unnecessarily exposes implementation details. For me it would be fine to have no docs comment at all for this since the name `rz_type_struct_member_is_bitfield` already says everything.

librz/include/rz_type.h
```diff
@@ -93,14 +93,14 @@ typedef struct rz_type_struct_member_t {
 	char *name;
 	RzType *type;
 	size_t offset; // in bytes
-	size_t size; // in bits?
+	size_t size; // bitfield width in bits, or 0 if the member is not a bitfield
```
thestr4ng3r:
```suggestion
	size_t size; ///< bitfield width in bits, or 0 if the member is not a bitfield
```

test/unit/test_serialize_types.c
```diff
@@ -67,17 +67,19 @@ bool test_types_save() {
 	RzType *mtype = rz_type_parse_string_single(typedb->parser, "char *", &error_msg);
 	mu_assert_notnull(mtype, "member type parsing");
 	member.type = mtype;
-	member.size = rz_type_db_get_bitsize(typedb, mtype);
-	mu_assert_eq(member.size, 64, "member type size");
+	ut64 gillian_bitsize = rz_type_db_get_bitsize(typedb, mtype);
+	mu_assert_eq(gillian_bitsize, 64, "member type size");
+	member.size = 0; // member.size is the bitfield width; 0 means not a bitfield
```
thestr4ng3r: The comment here is redundant, same for the ones below.

thestr4ng3r (changes requested):

librz/type/format.c
```diff
@@ -170,6 +178,14 @@ static void base_type_to_format_unfold(const RzTypeDB *typedb, RZ_NONNULL RzBase
 		rz_strbuf_append(format, "0");
 		RzTypeUnionMember *memb;
 		rz_vector_foreach (&type->union_data.members, memb) {
+			if (rz_type_union_member_is_bitfield(memb)) {
+				// C bitfield member: emit the pf ":N" packed-bits spec with the
+				// bit order matching the target endianness ('<' little, '>' big).
+				rz_strbuf_appendf(format, ":%u%s", (unsigned)memb->size,
+					(typedb->target && typedb->target->big_endian) ? ">" : "<");
```
Rot127, line 187: Warn if target is NULL.

librz/type/format.c
```diff
@@ -140,6 +140,14 @@ static void base_type_to_format_unfold(const RzTypeDB *typedb, RZ_NONNULL RzBase
 	case RZ_BASE_TYPE_KIND_STRUCT: {
 		RzTypeStructMember *memb;
 		rz_vector_foreach (&type->struct_data.members, memb) {
+			if (rz_type_struct_member_is_bitfield(memb)) {
+				// C bitfield member: emit the pf ":N" packed-bits spec with the
+				// bit order matching the target endianness ('<' little, '>' big).
+				rz_strbuf_appendf(format, ":%u%s", (unsigned)memb->size,
+					(typedb->target && typedb->target->big_endian) ? ">" : "<");
```
Rot127, line 148: Warn if target is NULL.

Rot127 (changes requested):
Two nitpicks.

PR #6440, opened 2026-05-30
librz/cons/histogram: revamp visual horizontal histogram (#6365, #4431)

### LINKED ISSUE #4431: rizin 0.8.0 Segfault p==v
### Work environment

| OS/arch/bits (mandatory)                             | Linux Mint 21.1 x64
| `rizin -v` full output, **not truncated** (mandatory)  | rizin 0.8.0 @ linux-x86-64 commit: 73d85d270a81c8c42c1b8e601d167ce2ad93157f

### Expected behavior
Working command

### Actual behavior
Segfault

```
truncate -s 8000M zeros; rizin -c 'p==v; q!!' zeros
```

### Additional Info
The segfault appears to happen **reproducible** at a given boundary of file size.
The file size appears to be **dependent however on the environment** that rizin is run in.
I searched for the file size using weighted binary search and found the following sizes to be the smallest file size that still results in a segfault:

|Byte|MiB|KiB|when running in
|-|-|-|-
|849314816 | ~ 809.9...  |  829409 | terminal
|846644224 | ~ 807.4...  | 826801 | ide
|847307776 | ~ 808.0...  | 827449 | ide terminal

### LINKED ISSUE #6365: Visual horizontal histogram mode improvements
Lets take `rizin-testbins/elf/busybox-powerpc`.
If we check the entropy histogram (`p==e`) it shows this:

Image

But if we run the visual mode (`p==ev`) that allows scrolling, vital information is missing:
- [ ] No vertical ruler indicating values (see also https://github.com/rizinorg/rizin/issues/5290)
- [ ] No horizontal ruler indicating where we are (apart from the red line - cursor)
- [ ] It would be nice to show the percentage in the right bottom corner as well
- [ ] Somehow it misses half of the diagram
- [ ] I think it would make sense to show also the minimap at the top.

Image

notxvilka:

**Detailed description**

This rewrites the look of the interactive horizontal histogram in Rizin - the output of `p==v`, `p==ev`, `p==mv`, `p==0v`, `p==fv`, `p==pv`, `p==zv`, `p==Sv` and the analysis-based interactive variants. It is built on top of the static histogram so it reuses the RzHistogramOptions struct and its helpers.

Five sub-tasks from #6365:
  - Context-aware vertical ruler with up to 5 anchor labels (top, bottom, quarter, half, three-quarter) honouring opts->value_min..value_max, value_precision, value_scale and value_unit. Reuses the static helpers compute_ruler_gutter, render_ruler_gutter, label_value_at and select_ruler_label_rows so visual and static stay consistent.
  - X-axis byte-offset ruler at the bottom: `^` ticks every (addr_w + 3) cols followed by absolute byte offsets computed from opts->offpos + realj * blocksize.
  - Cursor offset + percentage shown right-aligned on the status line, with "Index N data V" on the left.
  - Top minimap: Unicode block characters (▁▂▃▄▅▆▇█) for the density of each slice of the whole data, plus a ┏━━━┓ window indicator showing the visible slice. Always rendered when the new `scr.hist.minimap` config is true. When the terminal is wide (hist->w > 200) the minimap shrinks by 43 columns to make room for a two-line `px 0x20`-style hex preview panel on the right showing the 32 bytes at the cursor's file offset. Each byte is wrapped in its px-style colour code (b0x00 / b0x7f / b0xff / btext / other) when opts->color is set, matching rz_print's rz_print_byte_color exactly. A trailing safety pad keeps the canvas from clipping the last byte of the first hex row.
  - Missing-half bug on `p==v` when the cursor is at offset 0 fixed by clamping `adder` to [0, histogramwidth - span]. The old expression `barnumber + 1 - histogramwidth/(zoom*2)` was always negative for the default barnumber=0, causing the rendering loop to read `data[-N]` (segfault on large files, missing left half on small ones).

Cursor visibility:
  - The cursor column is drawn as a CONTINUOUS vertical line connecting two plain markers at the top and bottom, ALWAYS exactly one character wide. The line itself uses the dedicated `wordhl` palette colour (default red background, configurable via `ec wordhl ...`), drawn on every chart row so the cursor is always a full-height vertical strip. The markers (`▼` at the top and `▲` at the bottom, or ASCII `v` / `^` when scr.utf8=false) are intentionally left un-highlighted so they read as a clean pair of arrows pointing at the cursor column.
  - The cursor screen column is computed up-front (j_cursor) by inverting the data-to-column map (rel * zoom * width / histogramwidth). Two distinct widening bugs are avoided this way:
      1. sizeofonebar > 1 (high zoom) - each data index spans several screen columns; only j == j_cursor && kbar == 0 renders as the cursor, the remaining kbar columns fall through to the gradient.
      2. histogramwidth < width (chart much wider than data) - several adjacent screen columns map to the same data index via integer truncation; only the j_cursor column may render as the cursor.

Interactive keybindings & live config:
  - The `:` hotkey drops into rz_core_visual_prompt_input, matching the rest of Rizin's visual modes. Lets the user run arbitrary rizin commands without leaving the histogram.
  - The `?` help text now uses the same colour-coded format as the visual / visual-bit-editor modes (rz_core_visual_append_help with pal.args for keys and pal.help for descriptions), shown via rz_cons_less_str.
  - The config-driven opts (scr.hist.minimap, scr.hist.block, scr.utf8, scr.color, hex.offset) are re-read on every redraw via refresh_visual_opts_from_config, so `:` `e scr.hist.minimap=true` <Enter> takes effect immediately without having to quit and re-enter. The canvas's `color` field is refreshed alongside so `scr.color` changes take effect on the same redraw.

Hex preview panel:
  - When the terminal is wide (hist->w > 200) and the minimap is enabled, the visual mode shows a two-line hex preview on the right of the minimap rows: 32 bytes at the cursor's file offset, formatted as 8 pairs of 2 bytes separated by spaces (`abcd ef00 1234 5678 ...`), matching `px 0x20` minus the header / offsets / ASCII column. The bytes are fetched live via rz_io_read_at_mapped each redraw, so moving the cursor (`h` / `l`) updates the preview.
  - Each byte gets its px-style colour code: green for 0x00, red for 0xff, yellow for 0x7f, btext (white) for printable ASCII, "other" (magenta) for non-printable. Mirrors rz_print_byte_color so the histogram preview reads consistently with `px`.
  - A trailing safety pad keeps the canvas from clipping the last hex byte of the first row (a side effect of UTF-8 minimap glyphs interacting with the canvas's width tracking when the row fills the canvas exactly).
  - Implemented via two new fields on RzHistogramInteractive (`cursor_bytes`, `cursor_bytes_len`) that the caller fills in just before the render call and clears right after. The minimap helper grows two extra parameters that the visual function passes through; when the panel is disabled (narrow terminal, no cursor_bytes, or shrinking the minimap would leave it < 40 cols) the helper falls back to the previous full-width minimap.

Closes #4431 in full:
  - The negative-offset crash above is the immediate segfault from the bug report.
  - `print_histogram_bytes` now samples one byte per block instead of reading nblocks contiguous bytes from core->offset. For an 8 GB file shown across 80 bars the original code rendered the first 80 bytes of the file; the new code samples at offsets brange->from + i * blocksize so the chart represents the full span.
  - The inner `int i` in the column-aggregation loop is renamed to `k` to drop the shadow over the outer `size_t i`.

Refactor on the cmd_print.c side:
  - New default_visual_opts(core, offset) returns an RZ_OWN RzHistogramOptions* pre-populated for the visual commands (ruler=true, minimap from scr.hist.minimap, offpos from caller, palette and screen-mode toggles from config via refresh_visual_opts_from_config). The nine print_visual_bytes call sites now build opts via this helper, then pass it to print_visual_bytes which takes ownership. Entropy sets value_max=8 / value_precision=1 / data_f=fdata so the visual histogram shows the Shannon range matching the static side.
  - `print_visual_bytes(core, opts, data, brange)` now propagates opts cleanup along every error path; rz_histogram_interactive_new no longer leaves a heap-allocated opts pointer dangling. The redraw loop fetches 32 cursor bytes via rz_io_read_at_mapped, points hist->cursor_bytes at a stack buffer for the call, then NULLs it back so the next iteration's fetch is independent.
  - RzHistogramInteractive gains `blocksize`, `cursor_bytes` and `cursor_bytes_len` fields.

New config option:
  - `scr.hist.minimap` (bool, default true) controls whether the top minimap is shown for p==v / p==ev. Surfaces as `opts->minimap` and is honoured by `rz_histogram_interactive_horizontal`. When true, the minimap is ALWAYS rendered (provided there's room) - even when the chart already shows the full data, in which case the window indicator spans the whole map. Changes via `:` `e scr.hist.minimap=...` <Enter> are picked up on the very next redraw.

Tests (33 total, 10 new for the visual side):
  - test_histogram_interactive_horizontal_basic - smoke test with barnumber=0 (pins the #4431 crash regression).
  - test_histogram_interactive_horizontal_ruler_percent - fractional labels with value_max=100 / value_scale=0.01 / unit="%".
  - test_histogram_interactive_horizontal_ruler_default - the legacy 0..255 byte ruler.
  - test_histogram_interactive_horizontal_no_negative_adder - covers `p==v` at offset 0 on a small data set.
  - test_histogram_interactive_horizontal_percent - status-line percent indicator present.
  - test_histogram_interactive_horizontal_cursor_markers - the ▼/▲ cursor markers (and ASCII v/^ fallback) appear on the cursor column, left un-highlighted.
  - test_histogram_interactive_horizontal_cursor_full_line - the cursor bar is rendered on every chart row between the markers regardless of the data threshold (continuous vertical line).
  - test_histogram_interactive_horizontal_cursor_width - pins single- char width across BOTH cursor-widening bugs: high zoom (sizeofonebar
    > 1) AND chart wider than data (histogramwidth < width).
  - test_histogram_interactive_horizontal_minimap_toggle - pins the scr.hist.minimap gating across {zoomed, not zoomed} when opts->minimap=true / =false.
  - test_histogram_interactive_horizontal_hex_preview - 4 cases: wide terminal + cursor_bytes shows the hex panel; narrow terminal suppresses it; missing cursor_bytes suppresses it; colour mode emits ANSI escape sequences for the bytes. Both p== integration tests in test/db/cmd/cmd_print pass with their regenerated EXPECT blocks (the per-block sampling change moves the visible bars for small buffers).

Screenshot 2026-05-31 at 1 24 49 AM

Screenshot 2026-05-31 at 1 25 23 AM

Screenshot 2026-05-31 at 1 25 51 AM

And this one with the zoomed in display:

Screenshot 2026-05-31 at 1 31 55 AM

**Test plan**

- CI is green
- `p==` subcommands
- `p==ev` command and other `p==` subcommands with the `v` suffix

**Closing issues**

Closes #6365.
Closes #4431.


librz/cons/histogram.c
```diff
@@ -500,30 +500,325 @@ RZ_API void rz_histogram_interactive_zoom_out(RzHistogramInteractive *hist) {
...
+	const unsigned char *data, int histogramwidth, unsigned int map_w,
+	int adder, int visible_span, int gutter_w,
+	bool hex_panel, const ut8 *hex_bytes, int hex_bytes_len) {
+	const char *levels_utf8[9] = { " ", "\xe2\x96\x81", "\xe2\x96\x82", "\xe2\x96\x83",
+		"\xe2\x96\x84", "\xe2\x96\x85", "\xe2\x96\x86", "\xe2\x96\x87", "\xe2\x96\x88" };
+	const char *levels_ascii[9] = { " ", ".", ":", "-", "=", "+", "*", "#", "@" };
+	const char **levels = opts->unicode ? levels_utf8 : levels_ascii;
+	const char *win_left = opts->unicode ? "\xe2\x94\x8f" /* ┏ */ : "[";
+	const char *win_right = opts->unicode ? "\xe2\x94\x93" /* ┓ */ : "]";
+	const char *win_mid = opts->unicode ? "\xe2\x94\x81" /* ━ */ : "-";
```
Rot127: Define the characters in macros. Also not in bytes please. Or is there a reason for that?

librz/cons/histogram.c
```diff
@@ -500,30 +500,325 @@ RZ_API void rz_histogram_interactive_zoom_out(RzHistogramInteractive *hist) {
...
+
+	for (int g = 0; g < gutter_w; g++) {
+		rz_strbuf_append(buf, " ");
```
Rot127: Could be done with `"%.*s", 5, ...`

librz/cons/histogram.c
```diff
@@ -500,30 +500,325 @@ RZ_API void rz_histogram_interactive_zoom_out(RzHistogramInteractive *hist) {
...
+
+// Render the top minimap (density gradient + window indicator) for the visual
+// histogram. Called only when the data is wider than the visible window so
+// the user can see at a glance where the visible slice sits in the whole data.
+// When `hex_panel` is true, a two-line `px 0x20`-style hex preview is rendered
+// on the right of each minimap row (16 bytes per row), drawn from
+// hist->cursor_bytes. The minimap takes (map_w) columns and the hex panel
+// takes (HEX_PANEL_W) columns separated by a leading space.
+static void render_visual_minimap(RzStrBuf *buf, const RzHistogramOptions *opts,
+	const unsigned char *data, int histogramwidth, unsigned int map_w,
```
Rot127: In the whole function it uses unsigned int etc. Use our internal types please

Rot127: Also, most of those parameters don't need to be signed if I see this correctly.

librz/cons/histogram.c
```diff
@@ -500,30 +500,325 @@ RZ_API void rz_histogram_interactive_zoom_out(RzHistogramInteractive *hist) {
...
+	const char *win_mid = opts->unicode ? "\xe2\x94\x81" /* ━ */ : "-";
+
+	for (int g = 0; g < gutter_w; g++) {
+		rz_strbuf_append(buf, " ");
+	}
+	// Density row: each screen column averages a slice of the full data.
+	for (unsigned int c = 0; c < map_w; c++) {
+		int start_idx = (int)((ut64)c * histogramwidth / map_w);
+		int end_idx = (int)((ut64)(c + 1) * histogramwidth / map_w);
+		if (end_idx <= start_idx) {
```
Rot127, line 611: Is this the overflow case? Then it is better to simply check `map_w` before. Instead of doing all that checking and casting inside here.

librz/cons/histogram.c
```diff
@@ -500,30 +500,325 @@ RZ_API void rz_histogram_interactive_zoom_out(RzHistogramInteractive *hist) {
...
+		rz_strbuf_append(buf, levels[lvl]);
+	}
+	if (hex_panel) {
+		rz_strbuf_append(buf, "   "); // separator margin
+		append_hex_preview_row(buf, opts, hex_bytes, hex_bytes_len, 0);
+	}
+	rz_strbuf_append(buf, "\n");
+	// Window indicator: '[====]' over the visible slice.
+	for (int g = 0; g < gutter_w; g++) {
+		rz_strbuf_append(buf, " ");
```
Rot127: As above

librz/cons/histogram.c
```diff
@@ -500,30 +500,325 @@ RZ_API void rz_histogram_interactive_zoom_out(RzHistogramInteractive *hist) {
...
+// Render the bottom status line for the visual histogram: cursor index + raw
+// data value on the left, file offset + percent through the chart on the
+// right (right-aligned within the screen width).
+static void render_visual_status(RzStrBuf *buf, const RzHistogramInteractive *hist,
+	const unsigned char *data) {
+	const RzHistogramOptions *opts = hist->opts;
+	ut64 cursor_off = opts->offpos + (ut64)hist->barnumber * hist->blocksize;
+	int percent = 0;
+	if (hist->size > 1) {
+		percent = (int)((ut64)hist->barnumber * 100 / (ut64)(hist->size - 1));
```
Rot127, line 646: Why is percent int?

librz/cons/histogram.c
```diff
@@ -500,30 +500,325 @@ RZ_API void rz_histogram_interactive_zoom_out(RzHistogramInteractive *hist) {
...
+	int left_len = snprintf(left, sizeof(left), "Index %d data %d",
+		hist->barnumber, data[hist->barnumber]);
+	int right_len = snprintf(right, sizeof(right), "0x%" PFMT64x "  %3d%%",
+		cursor_off, percent);
+	if (left_len < 0) {
+		left_len = 0;
+	}
+	if (right_len < 0) {
+		right_len = 0;
+	}
```
Rot127: Please don't YOLO over the error cases.

librz/cons/histogram.c
```diff
@@ -500,30 +500,325 @@ RZ_API void rz_histogram_interactive_zoom_out(RzHistogramInteractive *hist) {
...
+	unsigned int total_w = hist->w > 0 ? (unsigned int)hist->w : 80;
+	rz_strbuf_append(buf, left);
+	if ((unsigned int)(left_len + right_len) + 1 <= total_w) {
+		int pad = (int)total_w - left_len - right_len;
```
Rot127: Also here, a lot of casting probably can be prevented by do bounds check at the beginning. Also `int` for an always positive value?

librz/cons/histogram.c
```diff
@@ -500,30 +500,325 @@ RZ_API void rz_histogram_interactive_zoom_out(RzHistogramInteractive *hist) {
...
+	}
+	rz_strbuf_append(buf, "\n");
+	for (int g = 0; g < gutter_w; g++) {
+		rz_strbuf_append(buf, " ");
+	}
+	unsigned int c = 0;
+	for (int t = 0; t < ticks; t++) {
+		int start_col = tick_cols[t];
+		while ((int)c < start_col) {
+			rz_strbuf_append(buf, " ");
```
Rot127: `%.s`

librz/cons/histogram.c
```diff
@@ -500,30 +500,325 @@ RZ_API void rz_histogram_interactive_zoom_out(RzHistogramInteractive *hist) {
...
+		if (next_tick < ticks && (int)c == tick_cols[next_tick]) {
+			rz_strbuf_append(buf, "^");
+			next_tick++;
+		} else {
+			rz_strbuf_append(buf, " ");
+		}
+	}
+	rz_strbuf_append(buf, "\n");
+	for (int g = 0; g < gutter_w; g++) {
+		rz_strbuf_append(buf, " ");
```
Rot127: `%.s`

librz/cons/histogram.c
```diff
@@ -566,55 +884,192 @@ RZ_API RZ_OWN RzStrBuf *rz_histogram_interactive_horizontal(RZ_NONNULL RzHistogr
...
+		//     the minimap itself (otherwise the preview would crowd it).
+		const int HEX_PANEL_W = 43;
+		bool show_hex = (hist->cursor_bytes && hist->cursor_bytes_len > 0 &&
+			hist->w > 200 && (int)map_w > HEX_PANEL_W + 40);
+		if (show_hex) {
+			map_w -= HEX_PANEL_W;
+		}
+		render_visual_minimap(buf, opts, data, histogramwidth, map_w,
+			adder, visible_span, gutter_w,
+			show_hex, hist->cursor_bytes, hist->cursor_bytes_len);
```
Rot127, line 1024: Could the code before `render_visual_minimap` be part of `render_visual_minimap`?

librz/cons/histogram.c
```diff
@@ -500,30 +500,325 @@ RZ_API void rz_histogram_interactive_zoom_out(RzHistogramInteractive *hist) {
...
+		addr_w = 1;
+	}
+	int ticks = (int)data_cols / (addr_w + label_pad);
+	if (ticks < 2) {
+		ticks = 2;
+	}
+	if ((unsigned int)ticks > data_cols) {
+		ticks = (int)data_cols;
+	}
+	int *tick_cols = calloc(ticks, sizeof(int));
```
Rot127: RZ_NEW

librz/cons/histogram.c
```diff
@@ -566,55 +884,192 @@ RZ_API RZ_OWN RzStrBuf *rz_histogram_interactive_horizontal(RZ_NONNULL RzHistogr
...
 			}
-			for (k = 0; k < sizeofonebar; k++) {
-				if (realj < hist->size && realj >= 0 && (255 - curdata < threshold || (i + 1 == rows))) {
-					if (realj == hist->barnumber) {
-						if (opts->thinline) {
-							rz_strbuf_appendf(buf, "%s%s%s", colofcurbar, vline, Color_RESET);
-						} else {
-							rz_strbuf_appendf(buf, "%s%s%s", colofcurbar, block, Color_RESET);
-						}
+			for (kbar = 0; kbar < sizeofonebar; kbar++) {
```
Rot127: Please move all those loops into static functions to make everything more readable.

librz/core/cmd/cmd_print.c
```diff
@@ -5158,15 +5158,55 @@ static void showcursor(RzCore *core, int x) {
...
+	opts->color = rz_config_get_i(core->config, "scr.color");
+}
+
+// Build a heap-allocated RzHistogramOptions for the visual histogram. Unlike
+// default_histogram_opts (used by the static commands and returning a value),
+// rz_histogram_interactive_new owns its opts pointer so this helper returns
+// RZ_OWN. The caller fills value_min / value_max / value_precision / unit /
+// data_f if they apply, then passes the result to print_visual_bytes which
+// takes ownership.
+static RZ_OWN RzHistogramOptions *default_visual_opts(RzCore *core, ut64 offset) {
```
Rot127, line 5175: Comments like these are a little too verbose for a simple function like this.
- The comparison is not really needed.
- Also duplicates stuff like "pointer is owned", which is obvious due to `RZ_OWN` in the declaration.
- If the user is expected to set value min/max, then add parameters to the function please.
  - Also, this is always ignored below.
- "then passes the result to print_visual_bytes" these are semantic instructions which don't really make sense here.

librz/core/cmd/cmd_print.c
```diff
@@ -5235,14 +5280,26 @@ static RzCmdStatus print_visual_bytes(RzCore *core, RZ_NONNULL const unsigned ch
...
-		case '?':
+		case '?': {
+			static const char *help_msg_visual_hist[] = {
+				"?", "", "show this help",
+				"hl", "", "move cursor left / right one bar",
+				"+/-", "", "zoom in / out",
+				":cmd", "", "run a rizin command",
+				"q", "", "back to Visual mode (or Q / Space)",
+				NULL
+			};
```
Rot127: Move outside of function.

librz/core/cmd/cmd_print.c
```diff
@@ -5226,7 +5257,21 @@ static RzCmdStatus print_visual_bytes(RzCore *core, RZ_NONNULL const unsigned ch
 		}
 		hist->w = w;
 		hist->h = h;
+		// Pre-fetch up to 32 bytes at the cursor's file offset for the hex
+		// preview panel that appears on the right of the minimap when the
+		// terminal is wide (hist->w > 200). The buffer lives on the stack
+		// of this loop iteration; hist->cursor_bytes points at it for the
+		// duration of one render call.
+		ut8 cursor_bytes_buf[32];
```
Rot127: init with 0, program passively etc.

librz/core/cmd/cmd_print.c
```diff
@@ -5226,7 +5257,21 @@ static RzCmdStatus print_visual_bytes(RzCore *core, RZ_NONNULL const unsigned ch
...
+		ut64 cursor_off = brange->from + (ut64)hist->barnumber * brange->blocksize;
+		int cb_len = (int)sizeof(cursor_bytes_buf);
+		memset(cursor_bytes_buf, 0, sizeof(cursor_bytes_buf));
```
Rot127: Ah, lol

librz/core/cmd/cmd_print.c
```diff
@@ -5226,7 +5257,21 @@ static RzCmdStatus print_visual_bytes(RzCore *core, RZ_NONNULL const unsigned ch
...
+		rz_io_read_at_mapped(core->io, cursor_off, cursor_bytes_buf, sizeof(cursor_bytes_buf));
+		hist->cursor_bytes = cursor_bytes_buf;
+		hist->cursor_bytes_len = cb_len;
```
Rot127: `sizeof(cursor_bytes_buf);`

Rot127 (changes requested):
Aren't those `┏━━━┓ ` the wrong way around? They belong to the minimap, right? So either they need to be on top of the minimap or the other way around. In the screenshots they are at the bottom of the minimap.

I didn't check the tests yet. But it would be nice to add one interactive test in `cmd_interactive_modes`. Just to increase code coverage more and prevent segfault changes in the future.

I give the AI that all our TUI code looks like that, and it is indeed pretty hard to make it different.
TUI has no architecture design or anything (or I can't see it) and hence no API.

Due to that, I think this is mergable once the worst things are fixed (like silently ignoring snprintf errors, lol).
The feature is pretty nice in the end.

Even though it definitely is not up to our code quality standards, I think it is pretty hard to do it right for our current TUI environment.
Even if manually implemented.

notxvilka:
> Aren't those ┏━━━┓ the wrong way around? They belong to the minimap, right? So either they need to be on top of the minimap or the other way around. In the screenshots they are at the bottom of the minimap.

No, it represents the "zoom" showing how minimap piece zoomed into the full diagram. Moved it to the top though.

wargio:
i like it. is more readable than before.

wargio (approved):

PR #6445, opened 2026-05-31
Add `pf` commands autocomplete

notxvilka:

# Detailed description

- Implemented autocomplete for the `pf` commands
- Updated `doc/pf.md` and `pf?` help to reflect better the current state and removed some inconsistencies

Three `RzCmdArgType` values were added, six commands gain completion, eleven tab
points light up:

| Arg type | Used by | What you get |
|---|---|---|
| `RZ_CMD_ARG_TYPE_PF_FORMAT_NAME` | `pf-`, `pfa`, `pfc`, `pfd`, `pfn`, `pfs`, `pfv` | All registered named formats, prefix-filtered. |
| `RZ_CMD_ARG_TYPE_PF_FORMAT_PATH` | `pf.`, `pfw` | `name[.field[N]?]*` walked to arbitrary depth via STRUCT-typed children. |
| `RZ_CMD_ARG_TYPE_PF_FDF_FILE` | `pfo` | FDF basenames from the user's home + system formats dirs (deduped). |

Examples:
```
[0x00000000]> pfn elf_<TAB>
elf_header  elf_phdr  elf_shdr ...

[0x00000000]> pf. elf_header.<TAB>
e_ident  e_type  e_machine  e_version  e_entry ...

[0x00000000]> pfn inner "x4d4 first second"
[0x00000000]> pfn outer "?(inner) body"
[0x00000000]> pf. outer.body.<TAB>
first  second

[0x00000000]> pfn troll "[3]?(plop) str"
[0x00000000]> pfn plop "d4x2x8x2 one two three four"
[0x00000000]> pf. troll.str[1].<TAB>
one  two  three  four

[0x00000000]> pfo elf<TAB>
elf32  elf64  elf_dynamic
```

The path completer does three things that matter for UX:

1. **Suppresses the trailing space** after the format name. If the user
   tabs `pf. ut_header<TAB>` and there's a single match, the cursor stays
   right after `ut_header` (no inserted space) so the next keystroke can
   be `.` to descend into fields.
2. **Rewrites the completion `start`** past the last `.` once the user
   has typed it, so the field name replaces only the field portion of
   the buffer and leaves the format name (and dot, and any intermediate
   segments) alone.
3. **Refuses to complete inside `[...]` or right after `]`** — the user
   is mid-index or mid-descent, where identifier completion would
   produce a syntax error if accepted.

## Syntax note

The supported field-path syntax is what `pf_path_navigate` accepts:
`name[idx]?(.name[idx]?)*`. The `[N]` is a **suffix on a name segment**,
not a segment of its own — so `troll.str[1].two` works, but
`troll.str.[1].two` (with a dot before the `[`) doesn't (the engine
itself rejects it). The completer matches the engine.

## What's deliberately out of scope

- **`pf <raw_format>`** — the bare `pf` command takes an arbitrary
  raw format string. The pf DSL has dozens of single-char codes and
  hundreds of combinations; useful completion would require something
  much smarter than enumeration. Skipped.
- **Numeric `[N]` completion** — when the cursor sits inside `[...]`,
  the completer is silent. Completing numeric indices isn't useful
  without knowing the array's element count, which would require
  parsing the parent format and walking to the field — overkill for
  the marginal benefit.

# Test plan

- CI is green
- Manual testing of `pf` commands


doc/pf.md
```diff
@@ -117,8 +117,11 @@ canonical form is `t(format)`.
     of `=` inside the parens is the discriminator.
 - `G[(layout)]` -- 16-byte GUID. Layouts:
   - `ms` (default) -- Microsoft mixed-endian (D1 LE u32, D2/D3 LE u16, D4 raw)
-  - `be` -- RFC 4122 big-endian network order
-  - `le` -- all little-endian
+  - `be` -- RFC 4122 big-endian network order (D1/D2/D3 BE, D4 raw)
+  - `le` -- D1/D2/D3 little-endian; D4 stays in buffer order, so the
```
Rot127, line 125: Also RFC 4122? Then add it here please

doc/pf.md
```diff
@@ -134,10 +137,14 @@ canonical form is `t(format)`.
   prefix it consumes one byte.
 - `U` / `L` -- ULEB128 / SLEB128 (variable-length unsigned / signed). Decoded
   greedily up to 10 bytes (enough for a full 64-bit value).
-- `n` / `N` -- context-sized integer:
-  - `n1`/`n2`/`n4`/`n8` -- forced byte width (1/2/4/8) in LE.
-  - `N1`/`N2`/`N4`/`N8` -- same, BE.
-  - Bare `n`/`N` defaults to `ctx.bits/8` bytes.
+- `n` -- context-endian sized integer (unsigned hex):
+  - `n1`/`n2`/`n4`/`n8` -- 1/2/4/8 byte width. Endianness follows
+    `RzPfCtx::big_endian` instead of being pinned by the spec's case,
```
Rot127: `RzPfCtx` is a random struct in the code the user doesn't know about.
Please add some other descirption from where it determines the endianess,

notxvilka:
Ignore the leaks CI job - it's a leak already existed on `dev`, fixed in https://github.com/rizinorg/rizin/pull/6448

doc/pf.md
```diff
@@ -134,10 +137,14 @@ canonical form is `t(format)`.
...
+    which is what file-format readers want when a header byte determines
+    the rest of the file's endianness (ELF `EI_DATA`, MachO magic, etc.).
```
Rot127: The example is so specific, it seems like `n` is only there for file headers.
Maybe rephrase it together with the explanation where the endianess decision comes from in  https://github.com/rizinorg/rizin/pull/6445/changes#r3330533309

librz/core/cmd_descs/cmd_print.yaml
```diff
@@ -625,9 +625,19 @@ commands:
...
-            comment: "TLV record; t=tag, l=length, e=le/be, h=v/a (header inclusion), d=dispatch table"
+            comment: >
+              TLV record; t=tag, l=length, e=le/be, h=v/l/a (len
+              covers value / len+value / tag+len+value), d=dispatch
+              table
+          - text: "v(N) / v(N,lsb) / v(N,msb)"
+            comment: >
+              bitvector: N individual bits (1..4096) exposed as
+              separate 0/1 scalars; consumes ceil(N/8) bytes; default
+              bit order is MSB-first
```
Rot127: "MSB-first" means MSB is on the left or right hand side (lower or higher address)?

librz/core/cautocmpl.c
```diff
@@ -435,6 +439,270 @@ static void autocmplt_cmd_arg_any_type(RzCore *core, RzLineNSCompletionResult *r
...
+ * if present). On success, *out_name_len is the length of the name
+ * portion and *out_has_idx is true iff a complete `[N]` was present.
+ *
+ * Returns 0 if the segment is malformed in a way that should stop the
+ * walk (e.g. an empty name, an unterminated `[`, or a non-numeric
+ * index). The cursor is treated as a hard stop: parsing never reads
+ * past \p end. */
+static size_t pf_path_seg_consume(const char *p, const char *end,
+	size_t *out_name_len, bool *out_has_idx) {
+	const char *q = p;
```
Rot127: Why no regex?

Rot127: If not possible for any reason, please rename `q` and `p` to something descriptive.

librz/core/cautocmpl.c
```diff
@@ -435,6 +439,270 @@ static void autocmplt_cmd_arg_any_type(RzCore *core, RzLineNSCompletionResult *r
...
+	}
+	rz_list_free(list);
+}
+
+/* Helper: parse a single path segment starting at \p p.
+ *
+ * A segment is `name`, optionally followed by `[N]`. Returns the byte
+ * count consumed (including the optional `[N]` and the trailing `.`
+ * if present). On success, *out_name_len is the length of the name
+ * portion and *out_has_idx is true iff a complete `[N]` was present.
```
Rot127: So it looks something like this: `<name>[<N>]`?
And returns `strlen(name)` and `out_has_idx == true` if there is an `[<n>]` in the string?

librz/core/cautocmpl.c
```diff
@@ -435,6 +439,270 @@ static void autocmplt_cmd_arg_any_type(RzCore *core, RzLineNSCompletionResult *r
...
+ *
+ * \p committed is the prefix of the user's input up to and including
+ * the last `.` -- i.e. every fully-typed segment. The walk consumes
+ * one segment per loop iteration, looks the segment's name up in the
+ * current format, and re-parses the named struct's body when a
+ * `type_name` is present.
+ *
+ * The returned RzPfFormat is owned by the caller and must be freed
+ * with rz_pf_format_free(). */
+static RzPfFormat *pf_resolve_path_format(RzTypeDB *typedb,
```
Rot127:
```suggestion
 **/
static RZ_OWN RzPfFormat *pf_resolve_path_format(RzTypeDB *typedb,
```

Rot127: Also for the function: wouldn't a regex work here as well?

librz/core/cautocmpl.c
```diff
@@ -435,6 +439,270 @@ static void autocmplt_cmd_arg_any_type(RzCore *core, RzLineNSCompletionResult *r
...
+ *     suppress the trailing space so `.` can be typed next without
+ *     space insertion.
+ *   - `.` is the segment separator. The cursor sits in the final
+ *     segment; everything before the last `.` is the committed path.
+ *   - Inside `[...]` or immediately after `]` (no trailing `.`), the
+ *     user is mid-index, not mid-identifier: nothing useful to offer.
+ *   - The committed path is walked through STRUCT-typed children
+ *     (whose `type_name` resolves to another registered format).
+ *     Walking past a scalar or an inline struct returns nothing. */
+static void autocmplt_cmd_arg_pf_format_path(RzCore *core,
```
Rot127, line 624: Also, regex question.

Rot127 (changes requested):

notxvilka:
Regarding the regex @Rot127 it's possible but slower to use the regex.

Rot127:
@notxvilka Slower, yes. But this code is not a hot path at all, so I think maintainability and ease of understanding > performance in that case.

But your call.

notxvilka:
> @notxvilka Slower, yes. But this code is not a hot path at all, so I think maintainability and ease of understanding > performance in that case.
> 
> But your call.

I would merge as is, then later this week I can check again and substitute. Is this ok?

wargio:
LGTM

PR #6450, opened 2026-06-01
analysis: render flag-enum operands as an OR of members (#2344)

### LINKED ISSUE #2344: Command `aht` doesn't work for custom enum type
### Work environment

| OS/arch/bits (mandatory)                             | Parrot 5.0 64 bit
| Architecture/bits of the file (mandatory)            | x86/64
| `rizin -v` full output, **not truncated** (mandatory)  | rizin 0.4.0-git @ linux-x86-64
commit: 12a935dddd0d9d04cce1c0f19115ecb04eca93b3, build: 2022-02-21__11:40:56

### Expected behavior
Command `aht` should replace `int value` in disasm code to enum alias

### Actual behavior
Command `aht` did nothing.

- I'm having a test file code
```
#include <unistd.h>
#include <stdio.h>


int main() {
  printf("%d\n", access("/tmp/run", 6));
  return 0;
}
```
Compile `gcc test.c -o run`

I added custom enum type here https://github.com/dmknght/rizin/blob/dev/librz/analysis/d/types-linux.sdb.txt#L73 and added type to function `access` here https://github.com/dmknght/rizin/blob/dev/librz/analysis/d/functions-linux.sdb.txt#L99

- Compile and install new rizin
- Start analysis with `rizin -A run`
- Command `te` showed new enum type 
image
At first, nothing changed from disasm code except the comment 
image
- As rizin devs recommended on telegram channel, i tried `aht access_def` but nothing happened
image

notxvilka:

**Detailed description**

An enum type hint set on an operand (`ahie`) only replaced the immediate when its value matched an enum member exactly, via `rz_type_db_enum_member_by_val()`. A value that is the bitwise OR of several flag members -- e.g. the access(2) mode R_OK|W_OK == 6 from the issue -- matched no single member and was left as a raw number.

`replace_enum_hint()` in `rz_parse` now falls back to` rz_type_db_enum_get_bitfield()` when there is no exact member, so the value is rendered as the OR of the matching members, e.g. "access_def.W_OK | access_def.R_OK".

`rz_type_db_enum_get_bitfield()` is reworked along the way: it was unused and buggy (capped at 32 bits, reused a stale match for bits without a member, and emitted a "0x.. : " debug prefix). It now walks all 64 bits, returns the matching members qualified with the enum name and joined by " | ", and returns NULL when the value is 0, the type is not an enum, or any set bit has no member (so a value that is not cleanly a combination of flags stays a number).

**Test plan**

CI is green

**Closing issues**

Closes #2344

### DIFF
--- a/librz/type/type.c
+++ b/librz/type/type.c
@@ -566,47 +566,57 @@ RZ_API RZ_OWN RzList /*<char *>*/ *rz_type_db_find_enums_by_val(const RzTypeDB *
 }
 
 /**
- * \brief Returns all matching bitfields as an OR mask given the resulting value
+ * \brief Renders a value as the OR of the flag-enum members that compose it
+ *
+ * Each set bit of \p val is looked up among the members of the bitfield (flag)
+ * enum \p name; the matching member names, each qualified with the enum name,
+ * are joined by " | " into a newly-allocated string, for example
+ * "access_def.W_OK | access_def.R_OK". \p val is decomposed bit by bit, so a
+ * member whose value is not a single bit is never matched against the whole
+ * value -- use rz_type_db_enum_member_by_val() for that exact-match case (the
+ * disassembly filter tries it first).
  *
  * \param typedb Types Database instance
- * \param name The name of the bitfield enum
- * \param val The value to search for
+ * \param name The name of the bitfield (flag) enum
+ * \param val The value to decompose
+ * \return The qualified OR mask, or NULL if \p name is not an enum, \p val is 0,
+ *         or any set bit of \p val has no matching member (so \p val is not
+ *         cleanly a combination of the enum's flag members).
  */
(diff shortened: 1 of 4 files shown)


librz/type/type.c
```diff
@@ -566,47 +566,56 @@ RZ_API RZ_OWN RzList /*<char *>*/ *rz_type_db_find_enums_by_val(const RzTypeDB *
...
 				break;
 			}
 		}
-		if (isFirst) {
-			isFirst = false;
-		} else {
-			ret = rz_str_append(ret, " | ");
+		if (!member) {
+			// a set bit with no matching member: not a clean flag combination
+			free(ret);
```
Rot127, line 613: Warn here? Or is that a legal state?

Rot127 (approved):

librz/type/type.c
notxvilka: No, it's illegal. Let me add a warning.

PR #6453, opened 2026-06-01
Changed mmio_address in RzAnalysisOp to mmios[4]

SSharshunov:
Changed mmio_address in RzAnalysisOp to mmios[4] to fully replace register names from the SDBC database if multiple registers are used simultaneously.
For example: 
`add DPP3:0x025c, DPP0`
`add DPP3:0x025c, SORIC`

**Detailed description**

...

**Test plan**

...

**Closing issues**

...


librz/arch/p/analysis/analysis_c166.c
```diff
@@ -91,7 +91,6 @@ bool set_reg_val(RzAnalysis *analysis, const char *name, const ut64 value) {
 
 static void c166_set_mimo_addr_from_reg(RzAnalysisOp *op, ut8 reg) {
 	if (reg < 0xF0) {
-		op->mmio_address = BASE_SFR_ADDR + (2 * reg);
```
Rot127, line 94: Would you mind removing the lonely `if` statement here doing nothin?

Rot127: Also the one below please.

librz/arch/op.c
```diff
@@ -30,7 +30,9 @@ RZ_API void rz_analysis_op_init(RzAnalysisOp *op) {
 	op->refptr = 0;
 	op->val = UT64_MAX;
 	op->disp = UT64_MAX;
-	op->mmio_address = UT64_MAX;
+	for (int i = 0; i < 4; i++) {
```
Rot127: `RZ_ARRAY_SIZE`

librz/core/disasm.c
```diff
@@ -1127,6 +1116,18 @@ static void ds_build_op_str(RzDisasmState *ds, bool print_color) {
 	} else {
 		ds_opstr_try_colorize(ds, print_color);
 	}
+	for (int i = 0; i < 4; i++) {
```
Rot127: RZ_ARRAY_SIZE

librz/include/rz_analysis.h
```diff
@@ -874,7 +874,8 @@ typedef struct rz_analysis_op_t {
 	int ptrsize; /* f.ex: zero extends for 8, 16 or 32 bits only */
 	st64 stackptr; /* stack pointer */
 	int refptr; /* if (0) ptr = "reference" else ptr = "load memory of refptr bytes" */
-	ut64 mmio_address; // mmio address
+	ut8 mmios_count; // used mmio count
+	ut64 mmios[4]; // mmio address
```
Rot127:
```suggestion
	ut8 mmios_count; ///< used mmio count
	ut64 mmios[4]; ///< mmio address
```

Rot127 (changes requested):
Have you used AI by any chance for that one?
If yes, please check more carefully the changes before marking the PR as ready for review.

librz/arch/p/analysis/analysis_c166.c
SSharshunov: I have another PR in preparation, so I didn’t delete much of what I was going to change in it, so as not to create confusion.

SSharshunov:
> Have you used AI by any chance for that one? If yes, please check more carefully the changes before marking the PR as ready for review.

I use Gemini only to find information about obscure architectural details. Sometimes I use to understand the Rizin internals, but I don't use copy-paste.

SSharshunov:
In general, situations that seem to suggest AI use occur when I leave myself marks for future development, which reviewers then ask me to remove. As a result, some changes are lost in the future, resulting in illogical code.

Rot127:
sure, no problem.
In general you can also use AI, even to write (some) code.
Just check it very carefully.

librz/arch/p/analysis/analysis_c166.c
Rot127: Please remove them though. So we have a clean state in between PRs.
It's not a big deal if you add them later again.

librz/arch/p/analysis/analysis_c166.c
```diff
@@ -791,7 +790,6 @@ static void c166_op_mov_reg_data(RzAnalysis *analysis, RzAnalysisOp *op, const u
 	op->dst = c166_new_reg_value(analysis, reg, false);
 	op->src[0] = c166_new_imm_value(data, true);
 	if (op->dst != NULL) {
-		op->mmio_address = op->dst->base;
 	}
```
wargio: what?

wargio:
i do not understand the issue. how can there be multiple MMIO at the same address?

SSharshunov:
> i do not understand the issue. how can there be multiple MMIO at the same address?

I have mnemonics that represent several register names. While reverse engineering the RIZIN code, I noticed that values ​​like 0xfe06 are replaced with DPP3 if a value is set in mmio_address. A code search revealed that this is the only place where this field is used. So, I simply adapted the code for this architecture. There's also a place in AVR that uses this field.

SSharshunov:
image
as example in Keil for c166 (address 0x0c)

wargio:
for me that means that the map loaded for that rom image is wrong, and should load the correct one. i do not think you can just swap MMIO devices easily in AVR.

Is not an arch like the gameboy one where you have the concept of ram slots thus you have at the same address different data depending how a register is configured

SSharshunov:
> Is not an arch like the gameboy one where you have the concept of ram slots thus you have at the same address different data depending how a register is configured
> 
Then I'm asking for help in explaining how memory mapping should work correctly so that the analysis can see a 24-bit address. 
image

wargio:
oh this kinda works like a page table. let me double check the changes

wargio:
ok, so i think you should check if it matches with a DPPx register, then extract the upper 2 bits and print as `DPP<upper 2 bits>:<remaining bits as hex>`
you can keep the old mmio value and add an additional one for DPPx. maybe set it always to an invalid value, but when valid you print DPPn:offset

this is because you dont know what are the contents of DPP0-3 when disassembling

wargio:
you can also do this directly when resolving the mmio. just map the address as `DPP` and when you resolve the address, check if `DPP` and extract the upper 2 bits and print the DPPn register before the offset

SSharshunov:
How will this solve the substitution of address names from the sdb txt? Or are you suggesting that DPP3 be issued to the disassembler?

wargio:
yes to the second one. the address range is defined by sdb but we know that when resolved it could be a DPPn and the disassembler replaces this with the correct one

notxvilka (changes requested):
This definitely requires a test

wargio (approved):

Rot127:
Please rebase.

Issue #6095, opened 2026-03-25
Optimize the performance of the RzList implementation and algorhtms


notxvilka:
# Optimization Suggestions for Rizin's `RzList`

After analyzing the header and implementation, here are concrete optimizations ranked roughly by expected impact.

---

## 1. Node Pool / Slab Allocator (Highest Impact)

Every `rz_list_append`/`rz_list_prepend` calls `malloc` for a single 24-byte `RzListIter`. This is the **dominant cost** — per-node heap allocation causes massive overhead and poor cache locality [^1].

**Proposed:** Allocate nodes in bulk slabs and hand them out from a freelist:

```c
#define RZ_LIST_SLAB_SIZE 256

typedef struct rz_list_slab_t {
    RzListIter nodes[RZ_LIST_SLAB_SIZE];
    struct rz_list_slab_t *next_slab;
} RzListSlab;

typedef struct rz_list_pool_t {
    RzListSlab *slabs;
    RzListIter *freelist;
} RzListPool;

static inline RzListIter *pool_alloc(RzListPool *pool) {
    if (RZ_UNLIKELY(!pool->freelist)) {
        RzListSlab *slab = calloc(1, sizeof(RzListSlab));
        if (!slab) return NULL;
        slab->next_slab = pool->slabs;
        pool->slabs = slab;
        // chain all nodes onto freelist
        for (int i = 0; i < RZ_LIST_SLAB_SIZE - 1; i++) {
            slab->nodes[i].next = &slab->nodes[i + 1];
        }
        slab->nodes[RZ_LIST_SLAB_SIZE - 1].next = NULL;
        pool->freelist = &slab->nodes[0];
    }
    RzListIter *n = pool->freelist;
    pool->freelist = n->next;
    n->next = n->prev = NULL;
    n->val = NULL;
    return n;
}

static inline void pool_free(RzListPool *pool, RzListIter *node) {
    node->next = pool->freelist;
    pool->freelist = node;
}
```

This eliminates per-node `malloc`/`free`, keeps nodes contiguous for better prefetcher behaviour, and makes operations like `rz_list_purge` dramatically cheaper. An optional `RzListPool *pool` field on `RzList` makes this backward-compatible (fall back to `malloc` when `pool == NULL`).

---

## 2. Fix `rz_list_purge` — Avoid Per-Node Pointer Patching

[...]

PR #6465, opened 2026-06-06
librz/util/list: minor RzList performance optimizations

notxvilka:
# SQUASH ME

Only self-contained and easy to review changes with measurable benefits:

## Before / after (per operation)

Numbers below are median of 42 interleaved trials, and are stable across repeated runs. Absolute
nanoseconds depend on the host and on list size; the **ratio** is the portable
result.

| operation | pattern         | before        | after        | speedup            |
|-----------|-----------------|---------------|--------------|--------------------|
| prepend   | build N         | ~9.5 ns/elem  | ~8.5 ns/elem | **~1.10x**         |
| get_n     | uniform-random  | ~69 µs/call   | ~42 µs/call  | **~1.6–2.0x**      |
| get_n     | tail-region     | ~145 µs/call  | ~10 µs/call  | **~14–15x**        |
| set_n     | tail-region     | ~155 µs/call  | ~9 µs/call   | **~15–20x**        |
| del_n     | interior        | ~2.5 µs/op    | ~2.5 µs/op   | **~1.0x (neutral)**|
| append    | build N         | unchanged     | unchanged    | not modified       |
| clone     | copy N          | unchanged     | unchanged    | not modified       |
| merge sort| random N        | unchanged     | unchanged    | not modified       |

Notes per change:

* **get_n / set_n** — walk from whichever end is closer, i.e. `O(min(n, len-n))` instead of always from the head. Uniform-random access halves the average walk (len/4 vs len/2) → roughly 2x; access near the tail becomes a short backward walk → an order of magnitude. `set_n` shares the same traversal as `get_n`.

* **del_n** — **performance-neutral**. This change removes a block of manual neighbour/head/tail pointer patching that `rz_list_delete` already performs, and fixes the NULL-value traversal bug. The unlink is O(1) and is dominated by the O(n) walk to the n-th node, so the speed is unchanged; the value of the change is correctness and removing duplicated logic.

* **it->val guard removal** — primarily a correctness fix. It also removes one load + branch per loop iteration in `insert`, `reverse` and `insertion_sort`, but those are not hot paths and the effect is within noise.

**Test plan**

- CI is green
- Bench is faster

**Closing issues**

Partially addresses https://github.com/rizinorg/rizin/issues/6095

### DIFF
--- a/librz/util/list.c
+++ b/librz/util/list.c
@@ -440,20 +440,8 @@ RZ_API ut32 rz_list_del_n(RZ_NONNULL RzList *list, ut32 n) {
 
 	rz_return_val_if_fail(list, false);
 
-	for (it = list->head, i = 0; it && it->val; it = it->next, i++) {
+	for (it = list->head, i = 0; it; it = it->next, i++) {
 		if (i == n) {
-			if (!it->prev && !it->next) {
-				list->head = list->tail = NULL;
-			} else if (!it->prev) {
-				it->next->prev = NULL;
-				list->head = it->next;
-			} else if (!it->next) {
-				it->prev->next = NULL;
-				list->tail = it->prev;
-			} else {
-				it->prev->next = it->next;
-				it->next->prev = it->prev;
-			}
 			rz_list_delete(list, it);
 			return true;
 		}
--- a/librz/util/list.c
+++ b/librz/util/list.c
@@ -328,7 +328,7 @@ RZ_API RZ_BORROW RzListIter *rz_list_append(RZ_NONNULL RzList *list, RZ_NONNULL 
 RZ_API RZ_BORROW RzListIter *rz_list_prepend(RZ_NONNULL RzList *list, RZ_NONNULL void *data) {
 	rz_return_val_if_fail(list, NULL);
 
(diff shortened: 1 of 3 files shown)


librz/util/list.c
```diff
@@ -513,7 +501,7 @@ RZ_API RZ_BORROW RzListIter *rz_list_add_sorted(RZ_NONNULL RzList *list, RZ_NONN
 	for (it = list->head; it && it->val && cmp(data, it->val, user) > 0; it = it->next) {
 	}
 	if (it) {
-		item = RZ_NEW0(RzListIter);
+		item = RZ_NEW(RzListIter);
```
wargio, line 504: i can understand the usage of malloc instead of calloc to optimize performance (as it wont have to call memset) but i strongly discourage its usage as we can have non-clean pointers when calling `rz_list_free`.

Rot127: Second this.

Rot127 (changes requested):
I guess the `set_n` changes make sense. But the others don't even seem to lie outside of one std dev?
Can you add the full table of the benchmark please?

PR #6466, opened 2026-06-06
Add χ² , index of coincidence, min-entropy, serial correlation statistical indicators

notxvilka:

# Detailed description

After some experience with [scrutari](https://codeberg.org/xvilka/scrutari) and dissecting various firmwares, binary blobs, compressed and encrypted data, I decided that some of these statistical indicators would be useful also in Rizin. I chose the most high-impact ones but that are relatively small and simple enough to include in Rizin out of the box.

Served best with https://github.com/rizinorg/rizin/pull/6440

## Summary

`rz-hash` already ships three statistical plugins — `entropy`, `entropy_fract` and
`temperature`. All three are functions of the **byte histogram** and collapse a block to a
single "how random does this look" number. That number has well-known blind spots when
triaging firmware images and binary blobs. This PR adds four metrics that measure what
entropy cannot, each as an independent plugin with a public `rz_hash_*()` API, a unit test
and `test/db` tests:

1. **`chisquare`** — goodness-of-fit to a uniform distribution. Tells **encrypted apart
   from merely compressed** when both already report entropy ≈ 8.0.
2. **`ioc`** — index of coincidence (collision probability). The classic **single-byte /
   repeating-key XOR** detector — and it recovers the **key period**.
3. **`minentropy`** — worst-case entropy `-log2(max pᵢ)`. The **conservative floor** that
   cryptographers actually use (NIST SP 800-90B); exposes a dominant byte the *average*
   hides.
4. **`serialcorr`** — lag-1 autocorrelation. The only **order-aware** metric here:
   separates **code / structured data from random / compressed**, which no histogram
   statistic can.

The first three give a richer view of **distribution shape** than the single entropy
number; the fourth adds an orthogonal axis entropy ignores entirely — **byte order**.

Each metric is exposed **two ways**: as an `rz-hash` plugin (with a public `rz_hash_*()`
API) and as a per-block mode in the `p=` / `p==` block-visualization commands, alongside
the existing `p=e` entropy histogram — under keys `C` (chi-square), `I` (IoC),
`M` (min-entropy) and `S` (serial correlation).

### 1. Chi-square vs uniform (`chisquare`)

**Problem it solves.** Entropy near 8.0 means "looks random," but it cannot tell *truly
uniform* data (a stream cipher / CSPRNG output) from *high-entropy-but-structured* data (a
zlib/LZMA stream, a packed section, a table of pointers). Both can read entropy ≈ 7.99.

**What it measures.** Pearson's χ² of the 256-bin byte histogram against the uniform
expectation. A genuinely uniform source lands near **255** (the degrees of freedom);
residual structure — compression headers, alignment padding, Huffman tables, repeated
records — drives it sharply higher.

**Example.** You carve a region out of a firmware image; entropy says 7.99, so it "looks
encrypted." χ² ≈ 255 ⇒ consistent with encryption/CSPRNG. χ² in the thousands ⇒ it is not
encrypted, it is compressed or packed and still carries structure. This is a routine,
high-value question when deciding whether a blob is worth attacking as ciphertext.

*Computed anchors:* a constant block `"aaaaaaaa"` → **2040.0** (maximally non-uniform);
`"admin"` → **251.0**.

### 2. Index of coincidence (`ioc`)

**Problem it solves.** Single-byte and repeating-key XOR are everywhere in malware configs
and firmware, and entropy barely moves under them.

**What it measures.** The probability that two bytes drawn from the block are equal:
≈ **1/256 (0.0039)** for uniform data, ≈ **0.065** for English text, up to **1.0** for a
single repeated byte. IoC depends only on the *shape* of the histogram, so it is preserved
by single-byte XOR (XOR with a constant just permutes which value each count attaches to) —
which is exactly why a text- or code-like IoC **survives** the XOR and fingerprints the
underlying plaintext.

**Example (the killer use): recover a repeating-XOR key length.** For a key of period *P*,
slice the data into *P* columns (every *P*-th byte). Each column is single-byte-XOR'd
plaintext, so each column's IoC stays elevated (≈ the plaintext's), while for the *wrong*
period the columns look random (IoC ≈ 1/256). Sweep *P* and watch for the spike — that is
the Friedman/Kasiski test, and it is IoC's signature application. Entropy cannot do this.

*Computed anchors:* eight distinct bytes `"abcdefgh"` → **0.0** (no collisions);
`"aaaaaaaa"` → **1.0**; `"password"` (two `s`) → **0.0357**.

### 3. Min-entropy (`minentropy`)

**Problem it solves.** Shannon entropy is an *average*. It can stay high while one symbol
is very predictable — and "the most predictable symbol" is precisely what matters when you
ask "how many bits here can I actually trust?"

**What it measures.** H∞ = `-log2(max_i pᵢ)`, governed by the single most likely byte. It
is the worst-case bound used by NIST SP 800-90B for validating entropy sources, and it is a
conservative floor: **H∞ ≤ Shannon ≤ 8**. A large gap between the two is a tell that the
distribution is skewed by a dominant value.

**Example.** A buffer in which half the bytes are `0x00` and the rest are uniform has
Shannon entropy ≈ 5.0 bits, but **min-entropy = 1.0** (because `max p = 0.5`,
`-log2 0.5 = 1`). If you are judging whether a region is a real RNG output or usable as a
key/nonce, min-entropy is the honest number; Shannon would over-state it. It also flags
padding-heavy or sparse regions that average entropy smooths over.

*Computed anchors:* a uniform block → **8.0**; a constant block → **0.0**; `"abcdefgh"`
(8 equiprobable symbols) → **3.0** = log2 8.

### 4. Serial correlation (`serialcorr`)

**Problem it solves.** Metrics 1–3 are all **histogram** statistics — shuffle the bytes and
none of them change. But *order* is exactly what distinguishes machine code, text, counters
and sampled signals from random or compressed data.

**What it measures.** The lag-1 serial correlation coefficient (with wrap-around, as in the
classic `ent` tool): how strongly each byte predicts the next. ≈ **0** for
compressed/encrypted data (adjacent bytes are linearly unrelated); clearly **non-zero** for
executable code, ASCII text, incrementing counters, gradients and PCM audio.

**Example.** A table of incrementing 32-bit counters or a buffer of PCM samples shows strong
serial correlation even at moderate-to-high entropy, whereas compressed data of the *same*
entropy sits at ≈ 0. So serial correlation is the axis that says "this high-entropy blob is
actually structured/code," not "this is random." It is the natural partner to the histogram
metrics (see the table below).

*Computed anchors:* `"abcdefgh"` (a monotone ramp, damped by the wrap term) → **+0.333**;
a constant block or fewer than two bytes (undefined) → **0.0**.

---

## Why all four together

Each metric isolates a different property, and read jointly they discriminate cases that any
one number leaves ambiguous:

| entropy | chi-square | serial corr | most likely content |
|--------:|-----------:|------------:|---------------------|
| high (~8) | ~255 | ~0 | encrypted / CSPRNG |
| high (~8) | large | ~0 | compressed / packed |
| high (~8) | large | ≠ 0 | executable code |
| low–mid | large | varies | text / tables / structured records |

A high-IoC spike at stride *P* on top of any of these additionally flags repeating-key XOR
with period *P*.

Screenshot 2026-06-07 at 1 03 20 AM
Screenshot 2026-06-07 at 1 03 49 AM
Screenshot 2026-06-07 at 1 04 22 AM
Screenshot 2026-06-07 at 1 04 51 AM
Screenshot 2026-06-07 at 1 05 19 AM

# Test plan

- CI is green
- Play with new commands: `p==C`, `p==I`, `p==M`, `p==S`


librz/hash/algorithms/ioc/ioc.c
```diff
@@ -0,0 +1,40 @@
...
+	if (ctx->size > 1) {
+		double numerator = 0.0;
+		for (size_t i = 0; i < 256; i++) {
+			double c = (double)ctx->count[i];
+			numerator += c * (c - 1.0);
+		}
+		ioc = numerator / ((double)ctx->size * ((double)ctx->size - 1.0));
+	}
+	rz_write_be_double(digest, ioc);
+	return true;
```
wargio, line 39: early return

librz/hash/algorithms/minentropy/minentropy.c
```diff
@@ -0,0 +1,45 @@
...
+}
+
+bool rz_minentropy_final(ut8 *digest, RzMinEntropy *ctx) {
+	rz_return_val_if_fail(ctx && digest, false);
+	// Min-entropy H_inf = -log2(max_i p_i): the conservative (worst-case)
+	// entropy used by NIST SP 800-90B. Equals 8.0 for a uniform block and
+	// drops as soon as one byte value dominates, exposing non-uniformity
+	// that Shannon entropy can hide.
+	double min_entropy = 0.0;
+	if (ctx->size) {
```
wargio, line 30: early return

wargio (approved):

notxvilka:
@wargio these early returns would require either `goto` or code duplication, since there is still write after both condition outcomes. I think it should be good enough as is.

PR #20, opened 2020-10-07
Introduce FAKE command descryptor type ##newshell

ret2libc:
**Your checklist for this pull request**
- [ ] I've read the [guidelines for contributing](https://github.com/rizinorg/rizin/blob/master/DEVELOPERS.md) to this repository
- [ ] I made sure to follow the project's [coding style](https://github.com/rizinorg/rizin/blob/master/DEVELOPERS.md#code-style)
- [ ] I've added tests that prove my fix is effective or that my feature works (if possible)
- [ ] I've updated the documentation and the [rizin book](https://github.com/rizinorg/rizinbook) with the relevant information (if needed)

**Detailed description**

I have introduced a new `RzCmdDescType` which is FAKE, used to just show help entries in the help structure. They don't have an handler or anything, it's just help.

**Test plan**

Try `?`, you should see `|?`, `@?` and `>?` (for now) in the help. `@@?`, `~?` coming. Also, `@?` and the others should show various uses of those things.

PR #6514, opened 2026-06-15
GSoC'26: prelude detection plugin

MrQuantum1915:

# Issues to open

- [ ] using the byte search (used by command: `/x`) for `rz_core_search_prelude`/`rz_core_search_preludes`.
- [ ] Consider a suitable "reduction" algorithm for the resulting preludes.

---

**Detailed description**

Full blog : https://mrquantum1915.github.io/blogs/rizin-function-prologues-detection/ 

---
---

Prelude = Prologue

The function prologues generator plugin is a core plugin. 

**Full design**

TLDR : Extract prologues from binary having symbol info loading them in a prefix tree. Use that prefix tree to generalize trie bitmask generation for prologoues. After that generate a list of prologues (kw + mask).  Use this list of prologues to analyse target binary (like a striped bin) by searching for hit for the prologue signatures and create functions at appropriate places.

- **Extract**: Take binaries that have symbol tables. Read the first L bytes from each function start. Insert those byte sequences into the prefix tree, track how many times each byte appears at each position.
- **Generalize:** Walk the trie and produce generalised prologue signatures : (kw + mask)
- **Search:** Scan the target stripped binary for matches against those patterns, create function on hit using `aapf` and `aapk` commands.

The pipeline in rizin commands is fully automated end to end for simplicity of tool as discussed earlier. Foor user needing fine grained control should first look into using the `rz-prologues` binrz utility tool and then look into using APIs. We can add other APIs as per usage requirements in future.

The design is fully ephemeral (no persistency like earlier design). Each command is one shot command. Nothing is stored in rizin session. 

---

 **CMDs**
```bash
Usage: pg[jad]   # Core plugin to generate prologues from given set of binaries having symbol info.
| pg[j]        # Generate prologues only from current open file
| pga[j]       # Generate prologues for all open files (arch will be adopted from first open binary)
| pgd[j] <dir> # Generate prologues for all files in a directory (arch will be adopted from first valid binary)
```
**CMD semantics**:
| cmd | reads from |  writes to | |
|---|---|---|---|
|pg | current binary's symbol table| stdout | |
|pga | all open binaries	| stdout	| skips mismatched arch, already duplicates |
|pgd |	directory of binaries |stdout | opens each file temporarily, closes after processing |

---

**Utility**
```bash
Usage: rz-prologues (-f <file> | -d <dir>) [options]
 -f file      Input binary file to process
 -d dir       Input directory containing binaries to process (non-recursive)
 -o file      Write output to file instead of stdout
 -j           Output in JSON format (default is YAML)
 -l len       Prologue length in bytes (default: 16)
 -e threshold Entropy threshold for wildcard generalization [0.0-1.0] (default: 0.8)
 -r           Output raw prologues without entropy-based generalization
 -t           Output prefix tree (trie) structure instead of prologue list
 -a arch      Filter / target architecture (e.g. x86, arm)
 -b bits      Filter / target bitness of arch (e.g. 16, 32, 64)
 -B           Target big-endian architecture (default: little-endian)
 -q           Quiet mode (suppress log messages)
 -v           Show version information
 -h           Show this help message

Notes:
  When processing a directory (-d), if architecture (-a) or bits (-b) are not
  specified, the tool adopts the architecture of the first valid binary to
  prevent mixing cross-architecture prologues.

  Generated JSON prologues (-j) can be loaded into Rizin for function analysis
  using the 'aapf <file>' command.

Examples:
  # Extract generalized JSON prologues from a binary
  rz-prologues -f input.bin -j -o output.json

  # Extract 64-bit x86 prologues from a directory in quiet mode
  rz-prologues -d /path/to/bins -a x86 -b 64 -jq -o x86_64.json

  # Extract raw prologues without entropy-based generalization
  rz-prologues -f input.bin -rj -o raw_prologues.json

  # Dump the prefix tree (trie) structure without generalization to YAML
  rz-prologues -f input.bin -tr -o trie.yaml
```
---
**Data Poisoning Prevention:**
The trie accumulates prologues samples and they need to be clean samples. Inserting garbage corrupts the generalisation. Safeguards:

- Symbol filtering: Only FUNC and METH type symbols with valid physical addresses are used.
- Duplicate address filtering: Symbols pointing to the same address (ICF, weak symbols) are deduplicated using an Set.
- File: APIs gives feature to pass a RzSet of files to prevent same file being processed again
- Architecture gating: You cannot mix x86 and ARM prologues in the same trie. Mismatched files rejected based on target arch specified.
- Only use symbols from symbol table and dose not use functions created by any analysis cmds (aa/aaa/aaaa)

---

**Output formats**
The prologues json format (used by pgL for loading, pgpj for export):
```json
{
  "arch": "x86",
  "bits": 64,
  "endian": "little",
  "prologue_length": 16,
  "count": 3,
  "prologues": [
    {
      "bytes": "554889e5...",
      "mask": "ffffffffffff..."
    },
    ...
  ]
}
```
- `prologue_length`: all prologues in the file have this many bytes
- `bytes`: hex string of  prologue
- `mask`: hex string of  bitmask (same length as bytes)
- `count`:  number of prologuges

The trie json (used by pgtpj) includes trie specific metdata:
```json
{
  "arch": "x86",
  "bits": 64,
  "endian": "little",
  "prologue_length": 16,
  "total_prologues_analyzed": 1234,
  "files":[
      "netbsd-calculate-x64",
      "glibc-heap-2.27",
      ...
   ]
  "prefix_tree": [
    {"node_id": 1, "parent_id": 0, "bit_val":1, "hit_cnt": 1200},
    ...
  ]
}
```
---

**Prefix Tree (Trie) `RzUtil` Lib:** https://github.com/rizinorg/rizin/pull/6638

---

I build this web visualizer tool to inspect the generated trees, will help in deciding further algorithm and direction by experimenting on real binaries.
- upload exported trie json file to [Visualizer website](https://rz-prelude-visualize.vercel.app/) to inspect the trie.
- Would advice to reduce the tree depth on website to reduce clutter and also **speed up** rendering (dagre math) . Other details for usage are mentioned on site and github readme of visualizer.

---

**Questions:**
- [x] Is use of `rz_core_search_prelude` API acceptable for our purpose or does not seem right?  
- [x] Should we allow user to set session metadata (arch) manually or current one is fine (first operation wins).
- [x] Should we exclude "Imported dynamic symbols" from prologue generation?

---
**Links:**
- First Draft - [Detailed Google Doc](https://docs.google.com/document/d/1SgOFjlYZMUbGxa72dkApYdcgv_I8Q7ec_teZEBuE7VE/edit?usp=sharing)
- [Augmented Draft](https://hackmd.io/@CHOc-XAxRZmze0ILiNEXeg/SJwTHmk-Gx)
- [Prologues Prefix Tree Visualiser website](https://rz-prelude-visualize.vercel.app/) | [Github](https://github.com/MrQuantum1915/rz-prelude-visualize)

---

TODOs:
- [x] Generic RzTrie Lib. Almost 100% coverage by unit test :_)
- [x] `pga` for generating trie for multiple binfiles at once. with safeguards as much as possible for data poisoning.
- [x]  add files used for tree in output.
- [x] noticed some binaries have same address tagged as different symbols. So need to deduplicate symbols by address before processing OR do on the go. Currently every symbol is processed = poisioned data by repeated insertion. 
- [x] Add `pgd` command to analyse those 100s of bins in directory, because `zipall` is broken and also probably bad idea to load that much bins at one time. Instead iterate over files and close file after use.
- [x] Update the cmd_eval hexagon test after #6567  gets merged. 
- [x]  use Bitvector for bit trie or current one is fine?
- [x] add architecture data in trie, also show it in print (and hence export)
- [x] binrz wrapper for standalone implementation outside rizin session.
- [x] better remove the state persistence completely create unnecessary confusion  (hence remove arch metadata in ctx)
- [ ] Figure out testing and benchmarking flow! 
- [x] Generalization algorithm.
- [x] there's another hardship - little endian! For such bins, the nodes in trie diverge very early and meet at bottom. yet we cannot just insert reversed (8) bytes windows (ok for like arm), because of non-fixed length arch like x86. 
- [ ] should we CAP the max prologue length? as prelude API uses int for length of signature, it cant exceed int , currently its ut64 because config require ut64.
- [x]  config names too long? but have to use plugin.<plugin_name> as this is how plugin configs are deleted and eval.
- [x] use string for entorpy threshold or scaled integer (0,100) - percentage
- [ ] Choose optimal value of `DEFAULT_PROLOGUE_LEN` config after experimenting.
- [x] improve cmd descriptions help to avoid confusion in commands.
- [ ] do we need to add this plugin in auto analysis callback (plugin analysis that runs at end of auto analysis cmd)?
- [x] Tests
- [ ] finish all TODOs in code
- [ ] check the file name and warn the user if it has a .a suffix. Or you ask them once, when the command is run, if it should try to unpack them with rz-ar.
- [ ] finish all Doxygen (and rizin book?)
- [ ] Epilogue?

---

**Side Note**: just fixed doxygen in 2 files, caught it when visiting the file and my IDE didn't colorize the doxygen properly :]
 
`librz/bin/bobj.c`
`librz/core/canalysis.c`

---

**Test plan**

...
**Closing issues**

...


librz/core/p/prologues_generator.c
```diff
@@ -0,0 +1,274 @@
...
+
+	Trie *trie = trie_new();
+	if (!trie) {
+		RZ_FREE(ctx);
+		return false;
+	}
+	ctx->pg_trie = trie;
+
+	RzCmd *rcmd = core->rcmd;
+	RzCmdDesc *root_cd = rz_cmd_get_root(core->rcmd);
```
Rot127, line 930: NULL checks please

librz/core/p/prologues_generator/prefix_tree.h
Rot127, line 1: The Trie implementations should be in RzUtil and be independently tested as well.

librz/core/p/prologues_generator/prologues_gen.c
```diff
@@ -0,0 +1,274 @@
...
+ *
+ * TODO: short desc of algo, flow etc.
+ */
+
+#include <rz_core.h>
+#include <rz_cons.h>
+#include "prefix_tree.h"
+
+// may be config?
+#define PROLOGUE_LEN 16
```
Rot127: Yes, should be configurable.

librz/core/p/prologues_generator.c
```diff
@@ -0,0 +1,274 @@
...
+
+static const RzCmdDescHelp cmd_prefix_tree_clear_help = {
+	.summary = "Clear the prefix tree generated so far",
+	.args = cmd_prefix_tree_clear_args,
+};
+
+static const RzCmdDescHelp cmd_prefix_tree_print_help = {
+	.summary = "Print the prefix tree generated so far",
+	.args = cmd_prefix_tree_print_args,
+};
```
Rot127: Definitions like these can happen in an extra file.
E.g. .inc and include them then here.

librz/core/p/prologues_generator/prefix_tree.h
```diff
@@ -0,0 +1,41 @@
...
+ * For generic we need to add is_end flag in node, so that shorter sequence are preserved.
+ */
+
+#include <rz_types.h>
+#include <rz_vector.h>
+
+typedef struct trie_node_t {
+	ut64 hit_cnt;
+	RzPVector /*<TrieNode *>*/ children;
+	ut8 byte_val;
```
Rot127: Please don't implement a Trie which hard codes the assumption that a node must hold a byte.
Prefixes can be of any form: bit strings, instructions, substrings, you name it.

Rot127 (changes requested):
> select the other file and again run pg, as trie is persistent it will just update the trie. 

That won't work if I want to sample from a thousand binaries.
So that step should be automated.

And although that case should be handled by a stand alone tool, please implement this right from the beginning with many binaries in mind.

The "sample prologues from a single binary" is the special case.
"Sample from many binaries" is the default case.

librz/include/rz_util/rz_trie.h
```diff
@@ -0,0 +1,59 @@
...
+ * initialize it manually (or reinitialize after trie clear as
+ * `rz_trie_clear` frees `root->data` using callback).
+ *
+ * TODO: would radix tree be better or even usefull?
+ *
+ * Used vector rather than fixed size children array (as in https://en.wikipedia.org/wiki/Trie),
+ * because for current use, for sparse data its wasteful. In future we can
+ * give user an option to select fixed array or vector for children, based
+ * on their use case, likely using an enum for RzTrieChildrenMode. Implementation
+ * for API would branch based on mode. Adaptive Radix Tree is an option but its a complex DS.
```
Rot127: Please open an issue about these things and remove the comment.

librz/include/rz_util/rz_trie.h
```diff
@@ -0,0 +1,59 @@
...
+ * \file rz_trie.h
+ * \brief Trie (Prefix Tree).
+ * TODO: description
+ *
+ * Trie's root node is meant to be empty(dummy). If caller
+ * manipulates `root->data`, then its caller's responsibility to
+ * initialize it manually (or reinitialize after trie clear as
+ * `rz_trie_clear` frees `root->data` using callback).
+ *
+ * TODO: would radix tree be better or even usefull?
```
Rot127: Good idea

librz/util/trie.c
Rot127, line 1: Please add test cases right from the beginning.
If something goes wrong in such a fundamental data structure, you can easily waste a lot of time in debugging.

librz/include/rz_util/rz_trie.h
```diff
@@ -0,0 +1,59 @@
...
+typedef void (*RzTriePreVisitCb)(RzTrieNode *n, void *user);
+typedef void (*RzTriePostVisitCb)(RzTrieNode *n, void *user);
+typedef void (*RzTrieNodeOnHitCb)(RzTrieNode *n);
+
+// TODO: Add apis for find, delete, prefix search etc..
+RZ_API RZ_OWN RzTrie *rz_trie_new(RzTrieNodeMatchCb match, RzTrieNodeInitCb init, RzTrieNodeFreeCb free);
+RZ_API bool rz_trie_insert(RzTrie *t, const void *key, size_t len, RZ_NULLABLE RzTrieNodeOnHitCb on_hit);
+RZ_API void rz_trie_dfs(RzTrieNode *n, RzTriePreVisitCb cb_pre, RzTriePostVisitCb cb_post, void *user);
+RZ_API void rz_trie_clear(RzTrie *t);
+RZ_API void rz_trie_free(RzTrie *t);
```
Rot127: Add type annotations, const, RZ_BORROW etc.

librz/include/rz_util/rz_trie.h
```diff
@@ -0,0 +1,59 @@
...
+ * TODO: would radix tree be better or even usefull?
+ *
+ * Used vector rather than fixed size children array (as in https://en.wikipedia.org/wiki/Trie),
+ * because for current use, for sparse data its wasteful. In future we can
+ * give user an option to select fixed array or vector for children, based
+ * on their use case, likely using an enum for RzTrieChildrenMode. Implementation
+ * for API would branch based on mode. Adaptive Radix Tree is an option but its a complex DS.
+ *
+ * \note Not yet fully generic. Also it assumes a constant length sequence.
+ * For generic we need to add is_end flag in node, so that shorter sequence are preserved.
```
Rot127: Better implement it right from the beginning.

MrQuantum1915: ok then will fully complete the TRIE first and everything after that :)

test/unit/test_trie.c
Rot127, line 1: Looks good. Please also add a test with `RzBitVector` for example.

wargio:
@Rot127 wouldn't it make more sense to be an utility rather than commands in rizin?

MrQuantum1915:
> @Rot127 wouldn't it make more sense to be an utility rather than commands in rizin?

do you mean standalone binrz utility like `rz-prelude` (which is yet to be implemented as wrapper)?
if so i was just concerned about how users would use the generated prologues while analyzing a binary inside rizin session...

if its necessary then we can do like this : keep generation as utility and add loading/analysis cmds in rizin...

Rot127:
Just read the PR description and still need to review. But as early comments:

> wouldn't it make more sense to be an utility rather than commands in rizin?

Yes, I think we agreed on this? That is what I meant with "let's make it a stand alone tool like rz-ax" etc.
In any case, the API for the different extraction steps can stay the same I guess. But let me do the code review first.

Edit: I loosely remember that I was ok have the commands for testing. Because the command functions can be easily converted API functions. But that the final result was supposed to be a utility IIRC.

@MrQuantum1915 Please consider the following use cases, what are the steps for a user to solve it?

`I have a directory with 100 ELF binaries for Sparc with debug info. How do I extract all the preludes from them?`

`The plugin found a set of N prologues, how do share them with my colleague?`

`I have a stripped binary and the prologues my colleague gave me. How do I load them? How are the prologues used from there on?`

> and add loading/analysis cmds in rizin...

Please check how the current prologues are used. To keep things simple, just append the new ones to them one way or the other for now.

librz/util/trie.c
Rot127, line 1: Please extract the trie to a separate PR.

librz/core/p/prologues_generator.c
```diff
@@ -0,0 +1,1019 @@
...
+	ut64 trie_prologue_len; // cfg: prologue length used for generation, may differ from the length of prologues in the store
+
+	// prologues store (reset by pgp-)
+	RzVector /*<PGPrologue>*/ *prologues; // has independent lifecycle then pg_trie
+	ut64 store_prologue_len; // length of prologues in store
+
+	// session metadata (set by first pg[a|d]/pgL and reset only by pg-)
+	char *arch;
+	int bits;
+	bool big_endian;
```
Rot127, line 48: Document them with `///<`

librz/core/p/prologues_generator.c
```diff
@@ -0,0 +1,1019 @@
...
+	RzStructuredData *arr;
+	HtPU *ids;
+	ut64 curr_id;
+} TrieDFSContext;
+
+typedef struct {
+	RzVector /*<PGPrologue>*/ *prologues;
+	ut8 *buf;
+	size_t depth;
+} ProloguesDFSContext;
```
Rot127, line 66: Doxygen please.

librz/core/p/prologues_generator.c
```diff
@@ -0,0 +1,1019 @@
...
+	if (!binfiles) {
+		return RZ_CMD_STATUS_ERROR;
+	}
+
+	RzListIter *it;
+	RzBinFile *curr_file;
+	bool res = true;
+	size_t fcnt = 0;
+	rz_list_foreach (binfiles, it, curr_file) {
+		if (!curr_file || !curr_file->o || !curr_file->o->info) {
```
Rot127, line 298: There are getters for `->o` and the others IIRC. Please use those.

librz/core/p/prologues_generator.c
```diff
@@ -0,0 +1,1019 @@
...
+			continue;
+		}
+
+		if (rz_set_s_contains(ctx->fpaths, curr_file->file)) {
+			RZ_LOG_WARN("Skipping file '%s', already processed.\n", curr_file->file);
+			continue;
+		}
+
+		const RzBinInfo *info = curr_file->o->info;
+		if (!pg_check_file_arch(ctx, info, rz_config_get(core->config, "asm.arch"))) {
```
Rot127, line 310: If you change the API it would be better to pass the arch requirements as a single struct to it.
This way you don't have to fiddle with `core->config` here, one less dependency.

librz/core/p/prologues_generator.c
```diff
@@ -0,0 +1,1019 @@
...
+	}
+	RzList *files = rz_sys_dir(dir_path);
+	RzListIter *it;
+	char *file;
+	size_t fcnt = 0;
+	bool res = true;
+	rz_list_foreach (files, it, file) {
+		if (!strcmp(file, ".") || !strcmp(file, "..")) {
+			continue;
+		}
```
Rot127: For such kind of checks, please use only stuff from `rz_file.h`.

librz/core/p/prologues_generator.c
```diff
@@ -0,0 +1,1019 @@
...
+	const char *file_arch = arch_js->str_value;
+	int file_bits = (int)bits_js->num.u_value;
+	bool file_big_endian = !strcmp(endian_js->str_value, "big");
+
+	if (!ctx->arch) {
+		// first operation in session, adopt arch from file
+		ctx->arch = rz_str_dup(file_arch);
+		ctx->bits = file_bits;
+		ctx->big_endian = file_big_endian;
+	} else if (strcmp(ctx->arch, file_arch) != 0 ||
```
Rot127:
```suggestion
	} else if (RZ_STR_EQ(ctx->arch, file_arch) ||
```

librz/core/p/prologues_generator.c
```diff
@@ -0,0 +1,1019 @@
...
+	}
+
+	const RzJson *prologue_len_js = rz_json_get(json, "prologue_length");
+	if (!prologue_len_js || prologue_len_js->type != RZ_JSON_INTEGER) {
+		RZ_LOG_ERROR("Invalid JSON format: 'prologue_length' key is not an integer in file: %s\n", file_path);
+		goto cleanup_json;
+	}
+	size_t expected_len = prologue_len_js->num.u_value;
+
+	rz_vector_clear(ctx->prologues); // replace existing prologues
```
Rot127, line 480: Shouldn't it warn instead? If prologues are in the process of being overwritten?

librz/core/p/prologues_generator.c
```diff
@@ -0,0 +1,1019 @@
...
+ *
+ * TODO: short desc of algo, flow etc.
+ */
+
+#include <rz_core.h>
+#include <rz_cons.h>
+#include <rz_util.h>
+#include "prologues_generator.inc"
+
+#define DEFAULT_PROLOGUE_LEN 16
```
Rot127, line 16: Please add to the todo list to experiment with this value.

librz/core/p/prologues_generator.c
```diff
@@ -0,0 +1,1019 @@
...
+	}
+	ctx->trie_prologue_len = val;
+	return true;
+}
+
+static bool isValidSymbol(RzBinSymbol *symbol) {
+	if (symbol && symbol->type) {
+		const char *type = symbol->type;
+		return (symbol->paddr != UT64_MAX) &&
+			(!strcmp(type, RZ_BIN_TYPE_FUNC_STR) || !strcmp(type, RZ_BIN_TYPE_METH_STR));
```
Rot127: RZ_STR_EQ

librz/core/p/prologues_generator.c
```diff
@@ -0,0 +1,1019 @@
...
+
+	// first run
+	if (!ctx->arch) {
+		ctx->arch = rz_str_dup(file_arch);
+		ctx->bits = info->bits;
+		ctx->big_endian = info->big_endian;
+		return true;
+	}
+
+	return strcmp(ctx->arch, file_arch) == 0 &&
```
Rot127: RZ_STR_EQ

MrQuantum1915:
> Yes, I think we agreed on this? That is what I meant with "let's make it a stand alone tool like rz-ax" etc

Yeah, and I am yet to code that standalone wrapper, cause it just needs to call this commands we are having for core plugin. We will have both interactive prelude tool for user to do the same thing in rizin session as the standalone utility tool...

For your questions I think I should have added some example usage in description... sorry...will do that too

> I have a directory with 100 ELF binaries for Sparc with debug info. How do I extract all the preludes from them?

use `pgd` command, give the directory path to the cmd and it will generate prologues from all those files with single cmd. It iterates on each file in the directory , opens the file and runs the extraction and builds the trie. Then user needs to run `pgg` cmd to generalize prologues and store them in store. then he can export the prologues using `pgpj > sparc_preludes.json`. 

> The plugin found a set of N prologues, how do share them with my colleague?

As said above , use `pgpj` to export prologues in a standard json format. Share the json file with colleague.

> I have a stripped binary and the prologues my colleague gave me. How do I load them? How are the prologues used from there on?

Start a rizin session and load the stripped binary.
 Then u can load the shared json file using `pgL sparc_preludes.json`. The tool loads the prologues from json file and store them as kw + mask pairs in vector. Then when u run `pgs` command , it uses that stored prologues for analysis and searches the stripped binary and creates functions on hit of each of the prologues signatures.

librz/core/p/prologues_generator.c
```diff
@@ -0,0 +1,1019 @@
...
+
+RzCorePlugin rz_core_plugin_prologues_generator = {
+	.name = "prologues_generator",
+	.desc = "TODO",
+	.license = "LGPL-3.0-only",
+	.author = "MrQuantum1915",
+	.version = "1.0",
+	.init = rz_cmd_prologues_gen_init,
+	.fini = rz_cmd_prologues_gen_fini,
+	.analysis = NULL, // TODO
```
Rot127, line 1007: This is where you could implement the "scan for all loaded prologues in the current binary" I think.

wargio:
> i was just concerned about how users would use the generated prologues

probably we would need to extract them to some format and compile them (or allow external resources to be loaded)

Rot127 (changes requested):
Please see my questions about "how would the user do it?" above.

But from my POV I would expect a workflow like that:

- Me being a researcher, I have a stripped binary of some funny arch with bad function detection.
- I go and compile a bunch of code with debug info for that arch. All of them in `my_dir_with_debug_bins`. 
- _Either_: Run `rz-prologue-finder -o found_prologues.json ./my_dir_with_debug_bins/` (user provides arch config via command line options _or_ the first loaded binary decides it).
  - Load stripped binary I am actually interested in and run `aaa`: `rizin -c "pgL found_prologues.json ; aaa" stripped_binary.bin`
- _Or_: Run `rizin -c "<pg_cmd_gen_all_from> found_prologues.json ./my_dir_with_debug_bins/ ; pgL found_prologues.json ; aaa" stripped_binary.bin` (inherit arch config from `stripped_binary.bin`)

(If it is a common arch, Rizin should ship with a set of `ARCH_prologues.json`. See below)

In either case I think the number of commands should be reduced to two: `generate_prologues_from_sample_binaries`, `load_prologues_from_file (current pgL)`
Otherwise it just gets complicated and hard to use IMHO.

If they need fine grained control, they can write their own tool linking against the API of the plugin. (@wargio Or is this something we don't do for core plugins?)

That means each of those current command handlers could be replaced with an `RZ_API` equivalent, taking the arguments they need, instead of `char **argv, int argc`.

>  Is use of rz_core_search_prelude API acceptable for our purpose or does not seem right?

Frankly, I would refactor `rz_core_search_preludes()`:

- Remove the `analysis.prelude` option. Instead `aap` can take the byte sequence as argument and do what it did before.
- But if `aap` is run without argument, it search for all prologues detected by your plugin (using `rz_cmd_prologues_search_handler` for example). Of course, Rizin should ship with a set `ARCH_prologues.json`. `aap` can just pick and load the appropriate one and warn if there is non. Asking the user to generate such a file with the `pg` command or the standalone tool.

---

Then there is `rz_analysis_preludes`. Adding a callback to your plugin there seems the simplest way to me. THe callback can check if there is an appropriate `ARCH_prologues.json` to load. If not: warn.
If yes, return the prologues.

> Should we allow user to set session metadata (arch) manually or current one is fine (first operation wins).

See above

> Should we exclude "Imported dynamic symbols" from prologue generation?

For now, yes. It sounds like an edge case to me, wouldn't you say? It is more important to get this one to a round package. Then we can improve it.

MrQuantum1915:
>Please see my questions about "how would the user do it?" above.

@Rot127 , yeah i explained current workflow in above comment : https://github.com/rizinorg/rizin/pull/6514#issuecomment-5078549297

Rot127:
Yes, saw the comment after adding the review.
Would you disagree with the "reduce to two commands" point?

librz/util/trie.c
MrQuantum1915: hmm... the commits are all mingled here.
I will do this : pick the trie related files in a new branch and open the PR. And for this PR, i would squash all 33 commits so far and remove trie realted files and then rebase this one on that new branch. I think this will avoid those ugly merge conflicts

Rot127: Yes, that seems to be the easiest way.

MrQuantum1915:
@Rot127 

> Would you disagree with the "reduce to two commands" point?

Nah, i agree.  But i think we can have a win win solution here (simplicity + user has fine grained control too if he wants without writing their own tool with API)

I dont think we need to remove those individual commands + only give two commands. Those individual cmds  are also useful for inspecting internal state in between like of trie or similar, and also helpful in dev for now.

What we can do is: Keep those exisiting individual command, and add one master command which just concatenates those existing cmds pipeline = automated. Something like `pgG` command which takes `<dir>` and prints prologues.

This way you have to use only two commands and no need to touch other ones if you dont want.

We already have that loading prologues command `pgL`. We just need another command that "simply accepts bins and outputs prologues". Every process in between is automated...

WDYT?

---

- Would you disagree with having both : Utility + Cmds in Rizin? Utility can be just a wrapper around the cmds. We have `rz_core_cmd` APIs for calling cmds in rizin.

- Do we still want plugins functions to be APIs given the above point?  And arent plugin meant to be isolated and self contained? Wouldnt replacing with APIs would make this a part of core code??

---
>Frankly, I would refactor rz_core_search_preludes():
> Remove the analysis.prelude option. Instead aap can take the byte sequence as argument and do what it did before.
But if aap is run without argument, it search for all prologues detected by your plugin (using rz_cmd_prologues_search_handler for example). Of course, Rizin should ship with a set ARCH_prologues.json. aap can just pick and load the appropriate one and warn if there is non. Asking the user to generate such a file with the pg command or the standalone tool.

So what i think is, you are proposing a new structure for the existing prelude analysis as whole , right?
Like what your vision seems :
- Currently the particular arch analysis plugins define their callbacks for asking for arch prelude signatures during `aap`. You want to remove that and centralize all arch preludes into arch specific json files (database).
- `aap` loads this arch specific file during analysis if exist. If not ask user to give one. While also giving an option to manually give custom preludes even though there exist a file for that arch in shipped rizin.

That seems to be some larger refactoring of core code of `aap` and all archs analysis files.

Also after removing `analysis.prelude` ,  how do you pass single prologue to `aap` . Like we need to pass both bytes + mask, `aap <bytes> <mask>`? 

And this also means remove pgL command. `aap <file>`  loads custom file generator by plugin/utility and uses prologues from them for analysis. Also remove `pgs` command (no analysis code or callback in plugin) ?

Running `aap` alone will check for existing arch progloues file and error out if none found (asking user for manually giving it).

Then there's a conflict.  Like for single prologue : `aap <bytes> <mask>` or but we also want `aap <file>`? Both cant have same cmd.

This makes the pipeline look like this : 
- Plugin = Generator
- `aap` = Consumer



Is this what you mean? Or i got it wrong?

---

>For now, yes. It sounds like an edge case to me, wouldn't you say? It is more important to get this one to a round package. Then we can improve it.

hmm... like `RzBinSymbol` has `is_imported` field , we can ignore symbols that have this boolean true

Rot127:
> > Would you disagree with the "reduce to two commands" point? (https://github.com/rizinorg/rizin/pull/6514#pullrequestreview-4779209427)
>  But i think we can have a win win solution here (simplicity + user has fine grained control too if he wants without writing their own tool with API)

Let's also ask the others.
@wargio @notxvilka What is your opinion on it? Reduce number of cmds to two, or keep current set + one more to do all of the steps?

> Utility + Cmds in Rizin? Utility can be just a wrapper around the cmds. We have rz_core_cmd APIs for calling cmds in rizin.
> Do we still want plugins functions to be APIs given the above point?

Yes, it can be just a wrapper. But don't use `rz_core_cmd`. Always use proper API functions.
We spent the last years removing these `rz_core_cmd` calls from the code base, please don't add more to it.
This means either API functions or an equivalent struct to `RzCorePlugin`. But API functions sound simpler to me.

> Currently the particular arch analysis plugins define their callbacks for asking for arch prelude signatures during aap. You want to remove that and centralize all arch preludes into arch specific json files 

No, keep the arch specific ones. Just add the ones from the json files.

> Also after removing analysis.prelude , how do you pass single prologue to aap . Like we need to pass both bytes + mask, aap <bytes> <mask>?

For example. Even better is if you use `rz_search_parse_byte_pattern()`. It takes a pattern string and returns an object with the mask and all. You maybe have to implement some getters to get the `mask` and `bytes` from `RzSearchBytesPattern`.

>  Like for single prologue : aap <bytes> <mask> or but we also want aap <file>? Both cant have same cmd.
> This makes the pipeline look like this : Plugin = Generator ; aap = Consumer

I didn't think about that. But `aap <file>` makes sense. Not sure if we can do something like `aap -d <dir/file>`/`aap -b <prologue_pattern>`.



Generally, the thing I am concerned about is this (we don't want to be the last picture):

image

Rot127:
> hmm... like RzBinSymbol has is_imported field , we can ignore symbols that have this boolean true

Skip imported symbols for now. It is too annoying to check if they are actually loaded and you have to deal with redirections and such.

notxvilka:
> In either case I think the number of commands should be reduced to two: `generate_prologues_from_sample_binaries`, `load_prologues_from_file` (current pgL)

Agree with that. As you already mentioned, they can use the C API for a finer grained control. Or, if we will get actual request from users.

wargio:
i think we can reduce the commands, as we should stick to the KISS (keep it simple stupid) strategy

librz/core/cmd/cmd_analysis.c
```diff
@@ -6069,8 +6069,207 @@ RZ_IPI RzCmdStatus rz_autoname_all_functions_noreturn_handler(RzCore *core, int
...
+RZ_IPI RzCmdStatus rz_load_and_analyze_all_preludes_handler(RzCore *core, int argc, const char **argv) {
+
+	const char *file_path = argv[1];
+	if (!rz_file_exists(file_path)) {
+		RZ_LOG_ERROR("File does not exist: %s\n", file_path);
+		return RZ_CMD_STATUS_ERROR;
+	}
+
+	RzCmdStatus status = RZ_CMD_STATUS_ERROR;
+	RzList /* <RzSearchKeyword *> */ *prologues = rz_list_newf((RzListFree)rz_search_keyword_free);
```
wargio, line 6092: check for null.

librz/core/cmd/cmd_analysis.c
```diff
@@ -6069,8 +6069,207 @@ RZ_IPI RzCmdStatus rz_autoname_all_functions_noreturn_handler(RzCore *core, int
...
+
+	// validate arch metadata from file against curr open binary
+	const RzJson *arch_js = rz_json_get(json, "arch");
+	const RzJson *bits_js = rz_json_get(json, "bits");
+	const RzJson *endian_js = rz_json_get(json, "endian");
+	if (!arch_js || arch_js->type != RZ_JSON_STRING ||
+		!bits_js || bits_js->type != RZ_JSON_INTEGER ||
+		!endian_js || endian_js->type != RZ_JSON_STRING) {
+		RZ_LOG_ERROR("Invalid or missing arch metadata (arch, bits, endian) in file: %s\n", file_path);
+		goto cleanup;
```
wargio, line 6119: Add a null check here as `rz_json_get` does return NULL

librz/core/cmd/cmd_analysis.c
```diff
@@ -6069,8 +6069,207 @@ RZ_IPI RzCmdStatus rz_autoname_all_functions_noreturn_handler(RzCore *core, int
...
+		res = rz_hex_str2bin(mask_js->str_value, mask_buf);
+		if (res <= 0 || (size_t)RZ_ABS(res) != expected_len) {
+			RZ_LOG_ERROR("Failed to convert hex string to binary for prologue mask: %s\n", mask_js->str_value);
+			RZ_FREE(bytes_buf);
+			RZ_FREE(mask_buf);
+			goto cleanup;
+		}
+		RzSearchKeyword *kw = rz_search_keyword_new(bytes_buf, expected_len, mask_buf, expected_len, NULL);
+		RZ_FREE(bytes_buf);
+		RZ_FREE(mask_buf);
```
wargio, line 6235: maybe move this into a `static` function, so is cleaner.
Just take as input the `bytes_js` and `mask_js` and return `RzSearchKeyword *`

librz/core/cmd/cmd_search.c
```diff
@@ -302,70 +319,77 @@ RZ_API int rz_core_search_prelude(RzCore *core, ut64 from, ut64 to, const ut8 *b
...
-	const char *prelude = rz_config_get(core->config, "analysis.prelude");
-	ut64 limit = rz_config_get_i(core->config, "analysis.prelude.limit");
+/**
+ * \brief Search for function prologues and analyze them as functions.
+ * \param [in] core 	 RzCore instance to perform the search on.
+ * \param [in] prologues List of RzSearchKeywords to search for. If NULL, the default prologues from the
+ * 						 static database of the current architecture will be used if it exists.
+ * \returns The number of prologues hit if successful, -1 otherwise.
+ */
+RZ_API int rz_core_search_preludes(RzCore *core, RZ_NULLABLE RzList /*<RzSearchKeyword *>*/ *prologues) {
```
wargio: maybe return `ssize_t` or `st64`

librz/core/p/prologues_generator.c
```diff
@@ -0,0 +1,1146 @@
+// SPDX-FileCopyrightText: 2026 MrQuantum1915 <darshanpatelgdh@gmail.com>
+// SPDX-License-Identifier: LGPL-3.0-only
+
+/**
+ * \file prologues_generator.c
+ * \brief Dynamic prologues generator plugin.
+ *
+ * TODO: short desc of algo, flow etc.
+ */
```
wargio, line 9: doxygen here and for each public API (i.e. RZ_API). 
Also check if the public structures in the headers have the doxygen comments

librz/core/p/prologues_generator.c
```diff
@@ -0,0 +1,1146 @@
...
+		if (!buf) {
+			RZ_LOG_WARN("Failed to open buffer for file: %s\n", file_path);
+			RZ_FREE(file_path);
+			continue;
+		}
+
+		RzBinOptions opt;
+		rz_bin_options_init(&opt, -1, 0, 0, false);
+		opt.filename = file_path;
+		RzBinFile *bf = rz_bin_open_buf(bin, buf, &opt);
```
wargio, line 581: check for null.

librz/core/p/prologues_generator.c
wargio: Yes, if you implement this function, it will be called when `aaa` is executed.

librz/core/p/prologues_generator.c
```diff
@@ -0,0 +1,1146 @@
...
+	if (!rz_core_plugin_cmd_desc_remove(core, ctx->cmd_desc)) {
+		return false;
+	}
+	RZ_FREE(ctx);
+	return true;
+}
+
+RzCorePlugin rz_core_plugin_prologues_generator = {
+	.name = "prologues_generator",
+	.desc = "TODO",
```
wargio: missing

librz/core/cconfig.c
```diff
@@ -3026,7 +3026,6 @@ RZ_API int rz_core_config_init(RzCore *core) {
 	SETDESC(n, "Select the architecture to use");
 	update_analysis_arch_options(core, n);
 	SETCB("analysis.cpu", RZ_SYS_ARCH, &cb_asm_cpu_set, "Specify the analysis.cpu to use");
-	SETPREF("analysis.prelude", "", "Specify an hexpair to find preludes in code");
```
wargio, line 3030: you need to bump the projects version as you are removing this variable.

librz/search/bytes_search.c
```diff
@@ -74,6 +74,28 @@ RZ_API size_t rz_search_bytes_pattern_len(RZ_NONNULL const RzSearchBytesPattern
...
 
+/**
+ * \brief Get the read-only pointer to bytes buffer of the pattern.
+ *
+ * \return The bytes of the pattern or NULL in case of failure.
+ */
+RZ_API const ut8 *rz_search_bytes_pattern_get_bytes(RZ_NONNULL const RzSearchBytesPattern *bp) {
+	rz_return_val_if_fail(bp, NULL);
+	const ut8 *bytes = bp->bytes; // TODO: direct return
+	return bytes;
```
wargio:
```suggestion
	return bp->bytes;
```

librz/search/bytes_search.c
```diff
@@ -74,6 +74,28 @@ RZ_API size_t rz_search_bytes_pattern_len(RZ_NONNULL const RzSearchBytesPattern
...
+
+/**
+ * \brief Get the read-only pointer to the mask buffer of the pattern.
+ *
+ * \return The mask of the pattern or NULL in case of failure.
+ */
+RZ_API const ut8 *rz_search_bytes_pattern_get_mask(RZ_NONNULL const RzSearchBytesPattern *bp) {
+	rz_return_val_if_fail(bp, NULL);
+	const ut8 *mask = bp->mask; // TODO: direct return
+	return mask;
```
wargio:
```suggestion
	return bp->mask;
```

wargio: Is there a usecase where you only need the bytes or the mask? if not, why not merging this api and return the bytes & mask from the arguments (when not-null)?

wargio (changes requested):
Missing doxygen and please implement some unit & regression tests to verify the correctness of the C apis & commands.

test/db/cmd/cmd_list
```diff
@@ -69,13 +69,16 @@ EOF
...
-dex  LGPL-3.0-only deroad 1.0     Suite of dex commands, type `dex` for more info
-java LGPL-3.0-only deroad 1.0     Suite of java commands, type `java` for more info
+prologues_generator: TODO (Made by MrQuantum1915, v1.0, LGPL-3.0-only)
+[{"name":"dex","description":"Suite of dex commands, type `dex` for more info","author":"deroad","version":"1.0","license":"LGPL-3.0-only"},{"name":"java","description":"Suite of java commands, type `java` for more info","author":"deroad","version":"1.0","license":"LGPL-3.0-only"},{"name":"prologues_generator","description":"TODO","author":"MrQuantum1915","version":"1.0","license":"LGPL-3.0-only"}]
+name                license       author        version description                                       
+----------------------------------------------------------------------------------------------------------
+dex                 LGPL-3.0-only deroad        1.0     Suite of dex commands, type `dex` for more info
+java                LGPL-3.0-only deroad        1.0     Suite of java commands, type `java` for more info
+prologues_generator LGPL-3.0-only MrQuantum1915 1.0     TODO
```
wargio: still a TODO in the description

Rot127 (changes requested):
It crashes right on start of Rizin:

```
> rizin
 
==11965==ERROR: AddressSanitizer: heap-use-after-free on address 0x773cd69f0d68 at pc 0x789cd80b7d22 bp 0x7ffdb46b9610 sp 0x7ffdb46b9608
READ of size 8 at 0x773cd69f0d68 thread T0
    #0 0x789cd80b7d21 in cb_cfgdebug ../librz/core/cconfig.c:804
    #1 0x789ce956723a in rz_config_set_cb ../librz/config/old_config.c:159
    #2 0x789cd80e9e4c in rz_core_config_init ../librz/core/cconfig.c:3330
    #3 0x789cd81d9c5e in rz_core_init ../librz/core/core.c:1809
    #4 0x789cd81c59f5 in rz_core_new ../librz/core/core.c:941
    #5 0x789ceb01b0c0 in rz_main_rizin ../librz/main/rizin.c:496
    #6 0x000000401508 in main ../binrz/rizin/rizin.c:57
    #7 0x789cef908680 in __libc_start_call_main (/lib64/libc.so.6+0x3680) (BuildId: 5bd941be836f538fe5e10eff508f7f5dd94905a6)
    #8 0x789cef908797 in __libc_start_main@GLIBC_2.2.5 (/lib64/libc.so.6+0x3797) (BuildId: 5bd941be836f538fe5e10eff508f7f5dd94905a6)
    #9 0x0000004008b4 in _start (/home/user/.local/bin/rizin+0x4008b4) (BuildId: 71ac89e5d798e83128ad34b1a88489a965ebce0b)

0x773cd69f0d68 is located 19304 bytes inside of 23848-byte region [0x773cd69ec200,0x773cd69f1f28)
freed by thread T0 here:
    #0 0x789cef2ee71f in realloc.part.0 (/lib64/libasan.so.8+0xee71f) (BuildId: 61e8165ca90f7a4f06b73f064a938c76deaae079)
    #1 0x789ceb61165c in rz_vector_insert ../librz/util/vector.c:317
    #2 0x789ceb61330d in rz_vector_insert_sorted ../librz/util/vector.c:422
    #3 0x789ce95647d9 in config_add_entry ../librz/config/config_internal.h:54
    #4 0x789ce956a805 in rz_config_set_b ../librz/config/old_config.c:344
    #5 0x789cd80b7bb8 in cb_cfgdebug ../librz/core/cconfig.c:802
    #6 0x789ce956723a in rz_config_set_cb ../librz/config/old_config.c:159
    #7 0x789cd80e9e4c in rz_core_config_init ../librz/core/cconfig.c:3330
    #8 0x789cd81d9c5e in rz_core_init ../librz/core/core.c:1809
    #9 0x789cd81c59f5 in rz_core_new ../librz/core/core.c:941
    #10 0x789ceb01b0c0 in rz_main_rizin ../librz/main/rizin.c:496
    #11 0x000000401508 in main ../binrz/rizin/rizin.c:57
    #12 0x789cef908680 in __libc_start_call_main (/lib64/libc.so.6+0x3680) (BuildId: 5bd941be836f538fe5e10eff508f7f5dd94905a6)
    #13 0x789cef908797 in __libc_start_main@GLIBC_2.2.5 (/lib64/libc.so.6+0x3797) (BuildId: 5bd941be836f538fe5e10eff508f7f5dd94905a6)
    #14 0x0000004008b4 in _start (/home/user/.local/bin/rizin+0x4008b4) (BuildId: 71ac89e5d798e83128ad34b1a88489a965ebce0b)

previously allocated by thread T0 here:
    #0 0x789cef2ee71f in realloc.part.0 (/lib64/libasan.so.8+0xee71f) (BuildId: 61e8165ca90f7a4f06b73f064a938c76deaae079)
    #1 0x789ceb61165c in rz_vector_insert ../librz/util/vector.c:317
    #2 0x789ceb61330d in rz_vector_insert_sorted ../librz/util/vector.c:422
    #3 0x789ce95647d9 in config_add_entry ../librz/config/config_internal.h:54
    #4 0x789ce956b035 in rz_config_set ../librz/config/old_config.c:379
    #5 0x789ce9566fed in rz_config_set_cb ../librz/config/old_config.c:157
    #6 0x789cd80e78d4 in rz_core_config_init ../librz/core/cconfig.c:3200
    #7 0x789cd81d9c5e in rz_core_init ../librz/core/core.c:1809
    #8 0x789cd81c59f5 in rz_core_new ../librz/core/core.c:941
    #9 0x789ceb01b0c0 in rz_main_rizin ../librz/main/rizin.c:496
    #10 0x000000401508 in main ../binrz/rizin/rizin.c:57
    #11 0x789cef908680 in __libc_start_call_main (/lib64/libc.so.6+0x3680) (BuildId: 5bd941be836f538fe5e10eff508f7f5dd94905a6)
    #12 0x789cef908797 in __libc_start_main@GLIBC_2.2.5 (/lib64/libc.so.6+0x3797) (BuildId: 5bd941be836f538fe5e10eff508f7f5dd94905a6)
    #13 0x0000004008b4 in _start (/home/user/.local/bin/rizin+0x4008b4) (BuildId: 71ac89e5d798e83128ad34b1a88489a965ebce0b)

SUMMARY: AddressSanitizer: heap-use-after-free ../librz/core/cconfig.c:804 in cb_cfgdebug
Shadow bytes around the buggy address:
  0x773cd69f0a80: fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd
  0x773cd69f0b00: fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd
  0x773cd69f0b80: fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd
  0x773cd69f0c00: fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd
  0x773cd69f0c80: fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd
=>0x773cd69f0d00: fd fd fd fd fd fd fd fd fd fd fd fd fd[fd]fd fd
  0x773cd69f0d80: fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd
  0x773cd69f0e00: fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd
  0x773cd69f0e80: fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd
  0x773cd69f0f00: fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd
  0x773cd69f0f80: fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd
Shadow byte legend (one shadow byte represents 8 application bytes):
  Addressable:           00
  Partially addressable: 01 02 03 04 05 06 07 
  Heap left redzone:       fa
  Freed heap region:       fd
  Stack left redzone:      f1
  Stack mid redzone:       f2
  Stack right redzone:     f3
  Stack after return:      f5
  Stack use after scope:   f8
  Global redzone:          f9
  Global init order:       f6
  Poisoned by user:        f7
  Container overflow:      fc
  Array cookie:            ac
  Intra object redzone:    bb
  ASan internal:           fe
  Left alloca redzone:     ca
  Right alloca redzone:    cb
==11965==ABORTING
Aborted                    rizin

> git log -n 1
commit 1f0eada486fd9711714025248f9c43f1d9fde14b (HEAD, MrQuantum1915/gsoc26/prelude-detection-plugin)
Author: Mr.Quantum_1915 <darshanpatelgdh@gmail.com>
Date:   Mon Aug 24 11:48:35 2026 +0530

    Replace Persistence state design with Ephermeral design

```

librz/core/cconfig.c
Rot127: @MrQuantum1915 This removal also triggers the use after free. Probably because you did not remove the option below it.

Rot127 (changes requested):
- [ ] The entropy threshold leaks:
	```
	[0x00000000]> e plugins.prologues_generator.
	plugins.prologues_generator.entropy_threshold=0.8
	plugins.prologues_generator.prologue_len=16
	==21789==ERROR: LeakSanitizer: detected memory leaks

	Direct leak of 4 byte(s) in 1 object(s) allocated from:
	    #0 0x7f2ff52ef24f in calloc (/lib64/libasan.so.8+0xef24f) (BuildId: 61e8165ca90f7a4f06b73f064a938c76deaae079)
	    #1 0x7f2ff1d0b439 in rz_str_newf ../librz/util/str.c:834
	    #2 0x7f2fe1a95d4d in config_entropy_threshold_getter ../librz/core/p/prologues_generator.c:125
	    #3 0x7f2ff1ac1f53 in config_var_bind_get_value ../librz/config/config.c:604
	    #4 0x7f2ff1ac2b3d in rz_config_var_get_string ../librz/config/config.c:668
	    #5 0x7f2ff1acaac6 in rz_config_entry_get_string ../librz/config/config.c:1705
	    #6 0x7f2fe180d895 in core_config_print_var_as_string ../librz/core/cconfig.c:1696
	    #7 0x7f2fe180e69e in core_config_print_iterator ../librz/core/cconfig.c:1761
	    #8 0x7f2ff1abe12c in rz_config_iterate_over ../librz/config/config.c:183
	    #9 0x7f2fe180f1ec in rz_core_config_print_all ../librz/core/cconfig.c:1826
	    #10 0x7f2fe1c36e6e in rz_eval_getset_handler ../librz/core/cmd/cmd_eval.c:527
	    #11 0x7f2fe1c0a104 in argv_call_cb ../librz/core/cmd/cmd_api.c:743
	    #12 0x7f2fe1c0af7d in call_cd ../librz/core/cmd/cmd_api.c:799
	    #13 0x7f2fe1c0af64 in call_cd ../librz/core/cmd/cmd_api.c:795
	    #14 0x7f2fe1c0b019 in rz_cmd_call_parsed_args ../librz/core/cmd/cmd_api.c:812
	    #15 0x7f2fe1bd8016 in handle_ts_arged_stmt_internal ../librz/core/cmd/cmd.c:1169
	    #16 0x7f2fe1bd6dfd in handle_ts_arged_stmt ../librz/core/cmd/cmd.c:1117
	    #17 0x7f2fe1bf88fa in handle_ts_stmt ../librz/core/cmd/cmd.c:2773
	    #18 0x7f2fe1bfa16b in handle_ts_statements_internal ../librz/core/cmd/cmd.c:2830
	    #19 0x7f2fe1bf9220 in handle_ts_statements ../librz/core/cmd/cmd.c:2795
	    #20 0x7f2fe1bfb71d in core_cmd_tsrzcmd ../librz/core/cmd/cmd.c:2945
	    #21 0x7f2fe1bfb8b4 in rz_core_cmd ../librz/core/cmd/cmd.c:2965
	    #22 0x7f2fe18d6216 in rz_core_prompt_exec ../librz/core/core.c:2064
	    #23 0x7f2fe18d4432 in rz_core_prompt_loop ../librz/core/core.c:1939
	    #24 0x7f2ff18215b7 in rz_main_rizin ../librz/main/rizin.c:1527
	    #25 0x000000400fbf in main ../binrz/rizin/rizin.c:57
	    #26 0x7f2ff0e0a680 in __libc_start_call_main (/lib64/libc.so.6+0x3680) (BuildId: 5bd941be836f538fe5e10eff508f7f5dd94905a6)
	    #27 0x7f2ff0e0a797 in __libc_start_main@GLIBC_2.2.5 (/lib64/libc.so.6+0x3797) (BuildId: 5bd941be836f538fe5e10eff508f7f5dd94905a6)
	    #28 0x000000400734 in _start (/home/user/.local/bin/rizin+0x400734) (BuildId: 1192fcf155b4a16c0a68fc18fa229403426574aa)

	SUMMARY: AddressSanitizer: 4 byte(s) leaked in 1 allocation(s).
	Aborted                    rizin
	```
- [ ] Opening an library archive and running `pg` doesn't give a clear error message, but fails:
	```
	rizin libc++.a 
	 -- Use scr.accel to browse the file faster!
	[0x00000000]> pg
	ERROR: Prologues trie is empty. Build it first with pg, pga, or pgd.
	ERROR: Failed to generalize prologues from trie
	[0x00000000]> pga
	ERROR: Prologues trie is empty. Build it first with pg, pga, or pgd.
	ERROR: Failed to generalize prologues from trie
	```

librz/core/p/prologues_generator.c
```diff
@@ -0,0 +1,1066 @@
...
+	}
+
+	// check split entropy and merge subtrees based on it (skip if not generalizing)
+	if (pgctx->generalize && rz_pvector_len(&n->children) == 2) {
+		double entropy = shanon_entropy_of_split(n);
+		if (entropy > pgctx->entropy_threshold) {
+			// set next depth bit to 0 in mask
+			size_t child_byte_idx = pgctx->depth / 8;
+			size_t child_bit_pos = 7 - (pgctx->depth % 8);
+			pgctx->mask_buf[child_byte_idx] &= ~(1u << child_bit_pos);
```
Rot127, line 733: Add an assert here to check if `child_byte_idx` is oob.
Because `pgctx->depth` is incremented "somewhere", it can be in theory arbitrarily large.
So here must be a bounds check.

librz/core/p/prologues_generator.c
```diff
@@ -0,0 +1,1066 @@
...
+	pgctx->depth++;
+}
+
+static void post_visit_prologues(RzTrieNode *n, void *user) {
+	rz_return_if_fail(n && n->data && user);
+	ProloguesDFSContext *pgctx = user;
+	// reset child's mask bit to 1
+	size_t child_byte_idx = (pgctx->depth - 1) / 8;
+	size_t child_bit_pos = 7 - ((pgctx->depth - 1) % 8);
+	pgctx->mask_buf[child_byte_idx] |= (1u << child_bit_pos);
```
Rot127: Same as https://github.com/rizinorg/rizin/pull/6514/changes#r3842701078

librz/core/p/prologues_generator.c
```diff
@@ -0,0 +1,1066 @@
...
+typedef struct {
+	RzStructuredData *arr;
+	HtPU *ids;
+	ut64 curr_id;
+} TrieDFSContext;
+
+typedef struct {
+	RzVector /*<RzPrologue>*/ *prologues;
+	ut8 *byte_buf;
+	ut8 *mask_buf;
```
Rot127, line 49: I think it is better to make those RzBuffer.
That ensures we know at all times the length of the buffers and can prevent oob reads/writes.
The dependency on `depth` is not good because depth is dynamic, but the buffer size is not.

librz/core/p/prologues_generator.inc
```diff
@@ -0,0 +1,39 @@
...
+
+static const RzCmdDescHelp cmd_raw_prologues_gen_all_help = {
+	.summary = "Generate raw prologues prefix tree for all open files, without any generalisation",
+	.args = cmd_raw_prologues_gen_all_args,
+};
+
+static const RzCmdDescHelp cmd_raw_prologues_gen_dir_help = {
+	.summary = "Generate raw prologues prefix tree for all files in a directory, without any generalisation",
+	.args = cmd_raw_prologues_gen_dir_args,
+};
```
Rot127: What does "without any generalisation" mean?

Either remove it or make it more clear what it means please.

Rot127 (changes requested):
Some more things I found while using it.

- [ ] `pg` doesn't show up in the help of `p?~pg`.

librz/core/p/prologues_generator.inc
MrQuantum1915: Sorry i forgot to update  the  cmd help after the recent changes.

librz/core/cconfig.c
MrQuantum1915: Actually the use after free is due to a hidden bug in callback function of "cfg.debug" config.
It just hit us here by luck...

opened a PR to fix that , this fixes the crash on rizin startup: https://github.com/rizinorg/rizin/pull/6685

librz/main/rz-prologues.c
```diff
@@ -0,0 +1,333 @@
...
+	char *output = NULL;
+	if (json_mode) {
+		output = rz_structured_data_to_json(sd);
+	} else {
+		output = rz_structured_data_to_yaml(sd);
+	}
+	rz_structured_data_free(sd);
+	if (!output) {
+		RZ_LOG_ERROR("Failed to convert structured data to %s\n", json_mode ? "JSON" : "YAML");
+	}
```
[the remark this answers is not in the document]
wargio: do not commit this inline if

librz/core/p/prologues_generator.c
MrQuantum1915: yeah the issue was of oob access on the leaf nodes in the post_visit_cb
so there it needs check
but here the `len(vec)==2` check ensures that this doesn't oob

fixed that it will run now, i will look into using RzBuffer now

wargio:
```
In file included from ../librz/core/cmd/cmd.c:32:
../librz/core/cmd/cmd_analysis.c:6249:82: error: format specifies type 'int' but the argument has type 'int64_t' (aka 'long') [-Werror,-Wformat]
 6249 |         RZ_LOG_INFO("aapf: Found %" PFMT32d " prologue hits and analyzed functions.\n", hits);
      |         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~
../librz/include/rz_util/rz_log.h:50:44: note: expanded from macro 'RZ_LOG_INFO'
   50 |         __LINE__, RZ_LOGLVL_INFO, NULL, fmtstr, ##__VA_ARGS__);
      |                                         ~~~~~~    ^~~~~~~~~~~
```

librz/core/cconfig.c
```diff
@@ -3027,7 +3027,6 @@ RZ_API int rz_core_config_init(RzCore *core) {
...
```
Rot127, line 3030: This removal still requires to increment the project version.
This PR shows you how to do it: https://github.com/rizinorg/rizin/pull/6461/changes#top

librz/core/cmd_descs/cmd_analysis.yaml
```diff
@@ -122,18 +122,38 @@ commands:
...
-          - name: Search a custom prelude
+          - name: Examples
             entries:
-              - text: "e analysis.prelude='90AEF630'"
-                arg_str: ""
-                comment: Set new prelude
               - text: "aap"
                 arg_str: ""
-                comment: Search for 90AEF630 and create a new function
+                comment: Search for static arch preludes and create new functions on hit
```
Rot127, line 150:
```suggestion
                comment: Search for function prologues and create new functions on hit.
```

librz/core/cmd_descs/cmd_analysis.yaml
```diff
@@ -122,18 +122,38 @@ commands:
             cname: autoname_all_functions_noreturn
             args: []
       - name: aap
-        summary: Analyze all preludes
-        cname: analyze_all_preludes
-        args: []
+        summary: Analyze preludes
+        subcommands:
+          - name: aap
+            summary: Analyze all static arch preludes
```
Rot127, line 128:
```suggestion
        summary: Analyze prologues
        subcommands:
          - name: aap
            summary: Analyze all static arch prologues
```

I get confused as well all the time. But lets try to switch to the way better "prologue" naming.

librz/core/cmd_descs/cmd_analysis.yaml
```diff
@@ -122,18 +122,38 @@ commands:
...
+              - name: file
+                type: RZ_CMD_ARG_TYPE_FILE
+                optional: false
+          - name: aapk
+            summary: Search for a prologue keyword and create new functions on hit
+            cname: analyze_all_preludes_keyword
+            args:
+              - name: keyword
+                type: RZ_CMD_ARG_TYPE_STRING
+                optional: false
```
Rot127, line 144:
```suggestion
            cname: load_and_analyze_all_prologues
            args:
              - name: file
                type: RZ_CMD_ARG_TYPE_FILE
                optional: false
          - name: aapk
            summary: Search for a prologue keyword and create new functions on hit
            cname: analyze_all_prologues_keyword
            args:
              - name: keyword
                type: RZ_CMD_ARG_TYPE_STRING
                optional: false
```

librz/include/rz_prologues.h
```diff
@@ -0,0 +1,44 @@
...
+RZ_API st64 rz_prologues_trie_feed_directory(RZ_NONNULL RzTrie *pg_trie, RZ_NONNULL RzBin *bin, RZ_NONNULL const char *dir_path,
+	ut64 prologue_len, RZ_NULLABLE RzProloguesArchInfo *arch_info, RZ_NULLABLE RzSetS *processed_files);
+
+RZ_API RZ_OWN RzVector /*<RzPrologue>*/ *rz_prologues_generalize_and_extract(RzTrie *pg_trie, ut64 prologue_len, double entropy_threshold);
+RZ_API RZ_OWN RzVector /*<RzPrologue>*/ *rz_prologues_extract_raw_from_trie(RzTrie *pg_trie, ut64 prologue_len);
+
+RZ_API RZ_OWN RzStructuredData *rz_prologues_trie_to_structured_data(RZ_NONNULL const RzTrie *pg_trie,
+	ut64 prologue_len, RZ_NULLABLE const RzProloguesArchInfo *arch_info, RZ_NULLABLE const RzSetS *files);
+RZ_API RZ_OWN RzStructuredData *rz_prologues_to_structured_data(RZ_NONNULL const RzVector /*<RzPrologue>*/ *prologues,
+	ut64 prologue_len, RZ_NULLABLE const RzProloguesArchInfo *arch_info);
```
Rot127, line 44: ownerships for pointers are missing.

librz/main/rz-prologues.c
```diff
@@ -0,0 +1,334 @@
...
+	rz_print_colored_help(options, RZ_ARRAY_SIZE(options), false);
+	printf("\n" Color_CYAN "Notes:" Color_RESET "\n"
+	       "  When processing a directory (-d), if architecture (-a) or bits (-b) are not\n"
+	       "  specified, the tool adopts the architecture of the first valid binary to\n"
+	       "  prevent mixing cross-architecture prologues.\n\n"
+	       "  Generated JSON prologues (-j) can be loaded into Rizin for function analysis\n"
+	       "  using the 'aapf <file>' command.\n");
+
+	printf("\n%s%s%s\n", Color_CYAN, "Examples:", Color_RESET);
+	printf("\n" Color_CYAN "Examples:" Color_RESET "\n"
```
Rot127, line 41:
```suggestion
	printf(
```

Duplicate "Examples"

librz/main/rz-prologues.c
```diff
@@ -0,0 +1,334 @@
...
+	} else {
+		output = rz_structured_data_to_yaml(sd);
+	}
+	rz_structured_data_free(sd);
+	if (!output) {
+		RZ_LOG_ERROR("Failed to convert structured data to %s\n", json_mode ? "JSON" : "YAML");
+		return false;
+	}
+
+	if (output_file) {
```
Rot127: Invert condition and early return after the printf.
Makes it easier to read.

Rot127 (changes requested):
Looks really good! My testing so far was also giving some pretty decent results.

I also edited the PR description to add a list of future issues we need to open. Feel free to extend it.


Lastly,
- update the PR description with the new command set of `pg,pga,pgd`
- focus on tests. Don't forget to add also tests which pass invalid arguments to the command and `rz-prologue` parameters.

librz/core/cconfig.c
MrQuantum1915: Thanks

MrQuantum1915:
added the test and addressed many reviews, will address remaining reviews later

added 3 tests:
- regression tests for `rz-prologues` tool
- regresssion tests for `cmd_pg` in rizin session
- added unit tests (integration tests) for `rz_prologues.h` APIs

TODO:
- [ ] Add test for `aapf` and `aapk`  and new byte search api

...

the unit test fails on ASAN (passes without asan):
idk if its related to my plugin but will look into it later

```
../librz/bin/format/elf/elf_relocs_addend.c:7:27: runtime error: shift exponent 32 is too large for 32-bit type 'unsigned int'
    #0 0x7f6f45d29162 in arm_rotr32 ../librz/bin/format/elf/elf_relocs_addend.c:7
    #1 0x7f6f45d29162 in get_reloc_addend_arm ../librz/bin/format/elf/elf_relocs_addend.c:37
    #2 0x7f6f45d29162 in Elf32_rz_bin_elf_reloc_get_addend ../librz/bin/format/elf/elf_relocs_addend.c:258
    #3 0x7f6f45fc20d1 in patch_relocs_obj ../librz/bin/p/bin_elf.inc:708
    #4 0x7f6f45fc2e3a in relocs_obj ../librz/bin/p/bin_elf.inc:1052
    #5 0x7f6f45d81852 in rz_bin_set_and_process_relocs ../librz/bin/bobj_process_reloc.c:36
    #6 0x7f6f45d77242 in rz_bin_object_process_plugin_data ../librz/bin/bobj_process.c:298
    #7 0x7f6f45d72f8a in rz_bin_object_new ../librz/bin/bobj.c:534
    #8 0x7f6f45d42e58 in rz_bin_file_new_from_buffer ../librz/bin/bfile.c:180
    #9 0x7f6f45d52d28 in rz_bin_open_buf ../librz/bin/bin.c:295
    #10 0x7f6f441fb725 in rz_prologues_trie_feed_directory ../librz/core/p/prologues_generator.c:581
    #11 0x55f4d11b04b3 in test_prologues_generate ../test/integration/test_prologues.c:167
    #12 0x55f4d11b1f90 in all_tests ../test/integration/test_prologues.c:233
    #13 0x55f4d11ac44a in main ../test/integration/test_prologues.c:237
    #14 0x7f6f42827780  (/usr/lib/libc.so.6+0x27780) (BuildId: 503200d7fda94a5dc6058d7e0694e5d1dcb2e372)
    #15 0x7f6f428278b8 in __libc_start_main (/usr/lib/libc.so.6+0x278b8) (BuildId: 503200d7fda94a5dc6058d7e0694e5d1dcb2e372)
    #16 0x55f4d11ac494 in _start (/mnt/windows/1_Darshan/CS/Programing/Projects/Rizin/rizin/build-asan/test/integration/test_prologues+0x5494) (BuildId: f8cb113309ce91e1d179c8d85904201c3e130a05)

SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior ../librz/bin/format/elf/elf_relocs_addend.c:7:27 
```

Rot127:
@MrQuantum1915 That is an UBSAN (undefined behavior sanitizer), not an ASAN (address sanitizer) error. It probably triggers because your plugin disassembles all these instructions in the binary, and triggers the UB in there. Be so kind and fix it. It shouldn't be more than a one line change.

MrQuantum1915:
> @MrQuantum1915 That is an UBSAN (undefined behavior sanitizer), not an ASAN (address sanitizer) error. It probably triggers because your plugin disassembles all these instructions in the binary, and triggers the UB in there. Be so kind and fix it. It shouldn't be more than a one line change.

there are actually several UB being hit for a particular elf binary when our tool is run...
they are in elf relocations and there are also mem leaks in elf_symbols being triggered (have yet to look if its my plugin's bug or `elf_symbol.c`)

opened PR #6727 
might have to look for other UBs in in elf processing

PR #6530, opened 2026-06-18
RzConfig:  API to remove specific key and load plugin AFTER RzConfig is initialised

MrQuantum1915:

**Detailed description**

TODO:
- [x]  Add unit test

1. Load static plugins after config is initialised, so that core plugins can add their own configs when loaded (that plugin's `init` is called)

2. Added new api to remove specific key.

...

**Test plan**

see unit test

...

**Closing issues**

...

### DIFF
--- a/librz/config/config.c
+++ b/librz/config/config.c
@@ -502,6 +502,30 @@ RZ_API bool rz_config_add_bind(RZ_NONNULL RzConfig *cfg, RZ_NONNULL const char *
 }
 
 /**
+ * \brief Removes a config key and frees associated memory
+ * \param cfg The RzConfig which holds the key
+ * \param name Name of the variable to remove
+ *
+ * \return On success returns true, otherwise false
+ */
+RZ_API bool rz_config_remove(RZ_NONNULL RzConfig *cfg, RZ_NONNULL const char *name) {
+	rz_return_val_if_fail(cfg && RZ_STR_ISNOTEMPTY(name), false);
+	size_t length = rz_vector_len(&cfg->sorted_vars);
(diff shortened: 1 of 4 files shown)


librz/config/config.c
```diff
@@ -501,6 +501,30 @@ RZ_API bool rz_config_add_bind(RZ_NONNULL RzConfig *cfg, RZ_NONNULL const char *
...
+RZ_API bool rz_config_remove(RZ_NONNULL RzConfig *cfg, RZ_NONNULL const char *name) {
+	rz_return_val_if_fail(cfg && RZ_STR_ISNOTEMPTY(name), false);
+	size_t length = rz_vector_len(&cfg->sorted_vars);
+	size_t index = rz_vector_find_sorted(&cfg->sorted_vars, (void *)name, (RzVectorComparator)find_variable, NULL);
+	if (index >= length) {
+		RZ_LOG_ERROR("config: variable '%s' does not exists.\n", name);
+		return false;
+	}
+
+	RzConfigEntry *entry = rz_vector_index_ptr(&cfg->sorted_vars, index);
```
MrQuantum1915, line 520: here i know that `config_find_entry` should be used, but then we dont get the index (which is needed to remove the var from vector), so I duplicated that code directly so that i have index to remove it from vector too.

Rot127 (changes requested):
As mentioned in mattermost. It would be nice to have similar APIs around the `core->plugin_configs`. But I wouldn't allow a plugin to manipulate the core config (which is even owned by RzCore, not by the plugin). It just adds unnecessary entanglement.

wargio:
Closed as per discussion on Mattermost

PR #6533, opened 2026-06-18
librz/util/pj: Fix JSON depth limit handling

MrQuantum1915:

**Detailed description**

It was failing silently when hardcoded JSON depth limit was exceeded.

now if depth limit is exceeded, it logs error and returns NULL instead of invalid  json
also once this error happens , any subsequent addition will be rejected (to preserve error till end)  and finally return NULL

...

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/librz/include/rz_util/rz_pj.h
+++ b/librz/include/rz_util/rz_pj.h
@@ -14,7 +14,7 @@ typedef struct pj_t {
 	bool is_first;
 	bool is_key;
 	char braces[RZ_PRINT_JSON_DEPTH_LIMIT];
-	int level;
+	int level; ///< Current nesting depth. 0 = top-level, >0 = inside braces, <0 = error (depth limit exceeded).
 } PJ;
 
 /* lifecycle */
--- a/librz/include/rz_util/rz_pj.h
+++ b/librz/include/rz_util/rz_pj.h
@@ -14,7 +14,7 @@ typedef struct pj_t {
 	bool is_first;
 	bool is_key;
 	char braces[RZ_PRINT_JSON_DEPTH_LIMIT];
-	int level; ///< Current nesting depth. 0 = top-level, >0 = inside braces, <0 = error (depth limit exceeded).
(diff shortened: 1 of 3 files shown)


librz/util/pj.c
```diff
@@ -48,19 +54,27 @@ RZ_API void pj_reset(PJ *j) {
 }
 
 RZ_API char *pj_drain(PJ *pj) {
-	rz_return_val_if_fail(pj && pj->level == 0, NULL);
+	if (!pj) {
+		return NULL;
+	}
+	if (pj->level != 0) {
+		pj_free(pj);
```
Rot127, line 62: When does this happen?
I just find it suspicious that here you use pj_free but below it is normal free?

Please document the `level` member while you are at it.

Rot127 (approved):
Just this nitpick.

librz/util/pj.c
MrQuantum1915: >When does this happen?

on error path, json is partial when this may be called. So level will be non zero

>I just find it suspicious that here you use pj_free but below it is normal free?

in error path, we want to return NULL and not the invalid json
 `pj->sb` still owns its internal buffer.
while below its calling `rz_strbuf_drain_nofree` which extracts the buffer from `sb` (transferring ownership to `res`), so  after that plain free used

>Please document the level member while you are at it.

done.

librz/include/rz_util/rz_pj.h
```diff
@@ -14,7 +14,7 @@ typedef struct pj_t {
 	bool is_first;
 	bool is_key;
 	char braces[RZ_PRINT_JSON_DEPTH_LIMIT];
-	int level;
+	int level; ///< Current nesting depth. 0 = top-level, >0 = inside braces, <0 = error (depth limit exceeded).
```
wargio: use st64

Rot127 (approved):

wargio (approved):

MrQuantum1915:
@notxvilka

notxvilka (approved):

PR #6546, opened 2026-06-21
Update cconfig.c to make =? behave like =??

### LINKED ISSUE #301: Remove `e search.in=??` syntax and just keep `e search.in=?`
### Expected behavior
`e search.in=??` -> error
`e search.in=?` -> current output of `e search.in=??`.

### Actual behavior
Some eval accept only a set of options, like `search.in`. To get the list of accepted options of a eval you can now do either `e search.in=?` or `e search.in=??`. The former just prints the possible values, the latter also adds a description. Considering getting the list option should be quite easy with autocompletion, it's probably better to just list the options with descriptions.

This issue is to study and, if regarded as appropriate, remove `=??` and make `=?` behaves as `=??`.

Suggested by @Maijin .

Farhan-25:

**Detailed description**
I updated all commands to use =?.
It now displays the verbose output (including descriptions) instead of only listing the available nodes/options.
The usage of =?? has been removed throughout the codebase.
I also added strcmp() checks to ensure that only a single ? is accepted. Previously, commands such as: `e asm.features=?` `e asm.features=??` `e asm.features=???`all produced the same output. 
With the new check, only =? is valid, while any input containing more than one ? results in an error.

Its a good first issue **BUT AI WAS USED** just to find the `cconfig.c` file which had the code related for this. All changes are manually applied.
...

**Test plan**
NA
...

**Closing issues**

closes #301  

...


librz/core/cconfig.c
```diff
@@ -222,9 +222,11 @@ static bool cb_diff_sort(void *_core, void *_node) {
 static bool cb_asm_features_set(void *user, void *data) {
 	RzCore *core = (RzCore *)user;
 	RzConfigNode *node = (RzConfigNode *)data;
-	if (*node->value == '?') {
-		update_asmfeatures_options(core, node);
-		print_node_options(node);
+	if (node->value[0] == '?') {
+		if (!strcmp(node->value, "?")) {
```
Rot127: Why not simply do a `RZ_STR_EQ(node->value, "?")`?

Rot127: Or even simple, print the help whenever the first char is a ?

Rot127 (changes requested):

librz/core/cconfig.c
Farhan-25: I simply looked how its done in `cb_analysiscc `function and just did that.

Rot127: Both is fine for me. But if you want it like that, please make the condition an AND and use RZ_STR_EQ

Rot127 (approved):

wargio:
I'm starting to think this should be part of rz_config as "long description" and not as it is now. this patch is more a hack than an actual solution for the future.

(@Farhan-25 this comment is not intended for you, as your patch is correct)

wargio (approved):

PR #6547, opened 2026-06-21
M68k rzil

### LINKED ISSUE #5847: Motorola M68k RzIL uplifting
See files:
- `librz/arch/isa/m68k`
- `librz/arch/asm/p/asm_m68k_cs.c`
- `librz/arch/analysis/p/analysis_m68k_cs.c`

b1llow:

**Detailed description**

This PR implements a complete RzIL lifter for the Motorola 68000 (m68k) architecture, closing one of the remaining architectures without RzIL support (#5847). This enables `aez*` emulation, IL-based analysis and future esil retirement for m68k.

- Integer arithmetic/logic, shifts/rotates, bitfield instructions (`bf*`), `div`/`mul` variants, BCD (`abcd`/`sbcd`/`nbcd`), `move`/`movem`/`movea`/`moveq`/`exg`/`swap`, `lea`/`pea`, all branches/`bcc`/`scc`/`dbcc`, `jmp`/`jsr`/`rts`/`rte`/`rtd`, `link`/`unlk`, `chk`, `trap`/`trapcc`/`trapv`/`bkpt`, etc.
- FPU: full `f*` instruction set with 80-bit extended precision (`fp0`–`fp7`), condition codes in `fpsr`, exact remainder quotient bits, and dynamic rounding-mode handling via `fpcr`.
- ColdFire MAC/EMAC instructions (`mac`, `msac`, `movacc`, ...) with `acc0`–`acc3`, `accext01`/`23`, `macsr`, `mask`.
- Supervisor/MMU instructions (`movec`, `moves`, `ptest`, `pflush`, `plpa`, ...), cache ops (`cpushl`, `cinv*`, ...), debug module ops, coprocessor interface, `callm`/`rtm`, and system ops (`reset`, `stop`, `lpstop`, `halt`, `pulse`).
- Precise CCR flag modeling (X/N/Z/V/C) per the M68000 PRM, including subtle cases like `movem` predecrement CPU-specific base values (68000 vs 68020+).
- Behavior that cannot be modeled in pure RzIL (traps, privilege violations, MMU/cache/debug/coprocessor side effects, unimplemented insns) is lifted into dedicated `EFFECT_LABEL_SYSCALL` labels (`m68k_trap`, `m68k_illegal`, `m68k_privilege`, `m68k_chk`, `m68k_fpu`, `m68k_mmu`, `m68k_cache`, `m68k_debug`, `m68k_coprocessor`, `m68k_callm`, `m68k_rtm`, `m68k_system`, `m68k_unimplemented`) so it remains visible to the VM instead of silently dropping semantics.

**Test plan**

trace test summary (https://github.com/BinaryAnalysisPlatform/qemu/pull/48 https://github.com/rizinorg/rz-tracetest/pull/34):

| CPU | IDs | Clean-pass IDs | Untestable IDs | Failed IDs | Trace tests |
|---|---:|---:|---:|---:|---:|
| `68000` | 117 | 101 (86.3%) | 6 (5.1%) | 10 (8.5%) | 6,965 |
| `68010` | 121 | 102 (84.3%) | 7 (5.8%) | 12 (9.9%) | 7,117 |
| `68020` | 325 | 162 (49.8%) | 109 (33.5%) | 54 (16.6%) | 13,800 |
| `68030` | 331 | 162 (48.9%) | 107 (32.3%) | 62 (18.7%) | 14,181 |
| `68040` | 352 | 70 (19.9%) | 105 (29.8%) | 177 (50.3%) | 13,913 |
| `68060` | 355 | 161 (45.4%) | 107 (30.1%) | 87 (24.5%) | 13,919 |
| `cpu32` | 319 | 168 (52.7%) | 115 (36.1%) | 36 (11.3%) | 13,036 |
| `coldfire` | 337 | 130 (38.6%) | 157 (46.6%) | 50 (14.8%) | 10,897 |
| `cfv1` | 143 | 85 (59.4%) | 40 (28.0%) | 18 (12.6%) | 7,426 |
| `cfv2` | 136 | 77 (56.6%) | 40 (29.4%) | 19 (14.0%) | 7,440 |
| `cfv3` | 136 | 78 (57.4%) | 40 (29.4%) | 18 (13.2%) | 7,440 |
| `cfv4` | 145 | 85 (58.6%) | 40 (27.6%) | 20 (13.8%) | 7,596 |
| `cfv4e` | 333 | 58 (17.4%) | 89 (26.7%) | 186 (55.9%) | 10,893 |
| `cfv5` | 152 | 85 (55.9%) | 44 (28.9%) | 23 (15.1%) | 7,756 |

- `rz-test test/db/rzil/m68k` — new RzIL suite, including an end-to-end `aez` emulation test that decrypts the flag in `bins/elf/m68k/emulateme-m68060` (68060 FPU code path) and a regression test for CPU-specific `movem` predecrement semantics.
- `rz-test test/db/analysis/m68k` — rewritten analysis expectations (now exercising IL paths).
- `rz-test test/db/asm/m68k_*` — greatly expanded per-CPU disassembly/lifting coverage: 68000, 68010, 68020, 68030, 68040, 68060, cpu32, cfv1–cfv5. Instruction tests were added following the coverage report shared in review to cover `m68k_il.c` except unreachable error handling.
- `rz-test test/db/cmd/cmd_aL test/db/cmd/cmd_list test/db/formats/elf/core` — updated for the new IL support and core-file register sync; obsolete `test/db/cmd/m68k` removed.

AI disclosure: code in this PR was partially generated with AI assistance (Codex), then reviewed, corrected and tested by hand.

**Closing issues**

closes #5847


thestr4ng3r (changes requested):
I did a coverage run on `rz-test test/db/asm/m68k_*` and while coverage is quite good already, there are still some missing parts of `librz/arch/isa/m68k/m68k_il.c`.
Please add instruction tests to `test/db/asm/m68k_*` to cover everything except the error handling that should not be reachable normally.
See this for reference: http://cloud.metallic.software/priv/iwohp8Ie/coverage-html/librz/arch/isa/m68k/m68k_il.c.gcov.html

librz/arch/p/analysis/analysis_alpha_cs.c
```diff
@@ -197,198 +197,198 @@ static void alpha_op_set_type(RzAsmAlphaContext *ctx, RzAnalysisOp *op) {
 		op->type = RZ_ANALYSIS_OP_TYPE_UNK;
 		break;
 	}
-	case Alpha_INS_BEQ:
```
notxvilka: Could you please extract that change into a separate PR?

notxvilka (changes requested):

Rot127:
@b1llow Just created a new release with some more Alpha changes: https://github.com/capstone-engine/capstone/releases/tag/6.0.0-Alpha10

Rot127:
Merged the PR with the Capstone update.
You can rebase.

librz/arch/isa/m68k/m68k_cs.c
```diff
@@ -0,0 +1,520 @@
...
+RZ_IPI bool rz_m68k_reg_name_is_mmu_root_pointer(const char *name) {
+	return name && (!strcmp(name, "srp") || !strcmp(name, "crp"));
+}
+
+RZ_IPI bool rz_m68k_reg_is_fpu(m68k_reg reg) {
+	return reg >= M68K_REG_FP0 && reg <= M68K_REG_FP7;
+}
+
+RZ_IPI bool rz_m68k_op_is_fpu_reg(const cs_m68k_op *op) {
+	if (!op || (op->type != M68K_OP_REG && op->type != M68K_OP_INVALID)) {
```
Rot127: Why doesn't it check for the FPU op types?

librz/arch/isa/m68k/m68k_cs.c
```diff
@@ -0,0 +1,520 @@
...
+	case M68K_REG_URP:
+	case M68K_REG_SRP:
+		return true;
+	default:
+		return false;
+	}
+}
+
+RZ_IPI bool rz_m68k_op_is_control_reg(const cs_m68k_op *op) {
+	if (!op || op->type != M68K_OP_REG) {
```
Rot127: Shouldn't all the `!op` case be an assert? Or is this legal here?

librz/arch/isa/m68k/m68k_cs.c
```diff
@@ -0,0 +1,520 @@
...
+
+RZ_IPI bool rz_m68k_op_is_fpu_control_reg(const cs_m68k_op *op) {
+	if (!op || op->type != M68K_OP_REG) {
+		return false;
+	}
+	return rz_m68k_reg_is_fpu_control(op->reg);
+}
+
+RZ_IPI bool rz_m68k_insn_uses_fpu_operand(const cs_m68k *m68k) {
+	if (!m68k) {
```
Rot127: Assert?

librz/arch/isa/m68k/m68k_cs.c
```diff
@@ -0,0 +1,520 @@
...
+	return fallback_bits;
+}
+
+RZ_IPI ut32 rz_m68k_bits_access_bytes(ut32 bits) {
+	return RZ_MAX(1, bits / 8);
+}
+
+RZ_IPI ut32 rz_m68k_reg_stack_access_bytes(m68k_reg reg, ut32 bits) {
+	ut32 bytes = rz_m68k_bits_access_bytes(bits);
+	return reg == M68K_REG_A7 && bytes == 1 ? 2 : bytes;
```
Rot127, line 201: Is the A7 register not more than >16 bits?
Or is this intended. If it is a bug in Capstone we would need to fix it there.

librz/arch/isa/m68k/m68k_cs.c
Rot127: Also, sometimes it checks for !op sometimes it doesn't.

librz/arch/isa/m68k/m68k_cs.c
```diff
@@ -0,0 +1,520 @@
...
+	return insn == M68K_INS_TBLS || insn == M68K_INS_TBLSN;
+}
+
+RZ_IPI bool rz_m68k_tbl_insn_is_unrounded(m68k_insn insn) {
+	return insn == M68K_INS_TBLSN || insn == M68K_INS_TBLUN;
+}
+#endif
+
+RZ_IPI bool rz_m68k_fpu_insn_extension_word(const cs_insn *insn, ut16 *extension) {
+	if (!insn || insn->size < 4 || !extension) {
```
Rot127: Are the NULL cases asserts?

librz/arch/isa/m68k/m68k_cs.c
```diff
@@ -0,0 +1,520 @@
...
+		return false;
+	}
+	memset(op, 0, sizeof(*op));
+	op->type = M68K_OP_REG;
+	op->reg = reg;
+	return true;
+}
+
+RZ_IPI const cs_m68k_op *rz_m68k_fpu_insn_second_op_or_hidden_dst(const cs_insn *insn, cs_m68k_op *hidden_dst) {
+	if (!insn || !insn->detail) {
```
Rot127: Assert?

librz/arch/isa/m68k/m68k_cs.c
```diff
@@ -0,0 +1,520 @@
...
+	if (!op || !rz_m68k_reg_is_fpu(reg)) {
+		return false;
+	}
+	memset(op, 0, sizeof(*op));
+	op->type = M68K_OP_REG;
+	op->reg = reg;
+	return true;
+}
+
+RZ_IPI const cs_m68k_op *rz_m68k_fpu_insn_second_op_or_hidden_dst(const cs_insn *insn, cs_m68k_op *hidden_dst) {
```
Rot127: What is the hidden destination all about?
Is that a bug?

librz/arch/isa/m68k/m68k_cs.c
```diff
@@ -0,0 +1,520 @@
...
+}
+
+RZ_IPI bool rz_m68k_fpu_op_detail_is_invalid(const cs_m68k_op *op) {
+	return !op ||
+		op->type == M68K_OP_INVALID ||
+		(op->type == M68K_OP_REG && op->reg == M68K_REG_INVALID) ||
+		(op->type == M68K_OP_MEM && op->address_mode == M68K_AM_NONE);
+}
+
+RZ_IPI bool rz_m68k_fpu_op_is_illegal_read(const cs_m68k *m68k, const cs_m68k_op *op) {
```
Rot127: Please add a short comment why each case is an illegal read.

librz/arch/isa/m68k/m68k_cs.c
```diff
@@ -0,0 +1,520 @@
...
+		return true;
+	}
+	if (m68k->op_size.type == M68K_SIZE_TYPE_FPU) {
+		return !rz_m68k_op_is_mem_addr(op);
+	}
+	return op->type != M68K_OP_REG && !rz_m68k_op_is_mem_addr(op);
+}
+
+RZ_IPI bool rz_m68k_fpu_insn_data_alias_has_dst(const cs_insn *insn) {
+	if (!insn || !insn->detail || insn->detail->m68k.op_count < 1) {
```
Rot127: Assert?

librz/arch/isa/m68k/m68k_cs.h
```diff
@@ -4,10 +4,18 @@
 #ifndef RZ_M68K_CS_H
 #define RZ_M68K_CS_H
 
+#include <string.h>
+
```
Rot127:
```suggestion
```

librz/arch/isa/m68k/m68k_cs.h
```diff
@@ -4,10 +4,18 @@
...
+#include <rz_types.h>
 #include <rz_util/rz_str.h>
 #include <capstone/capstone.h>
 
-#define M68K_LONGEST_INSTRUCTION 22
+#ifdef CAPSTONE_M68K_H
+#include <capstone/m68k.h>
+#endif
+
```
Rot127:
```suggestion
```

librz/arch/isa/m68k/m68k_cs.h
```diff
@@ -62,4 +70,58 @@ static inline cs_mode rz_m68k_cs_mode(const char *cpu) {
 	return CS_MODE_M68K_040;
 }
 
+#ifdef CAPSTONE_M68K_H
```
Rot127: Is there any case when this is not defined?

librz/arch/isa/m68k/m68k_il.h
```diff
@@ -0,0 +1,19 @@
...
+// SPDX-License-Identifier: LGPL-3.0-only
+
+#ifndef RZ_M68K_IL_H
+#define RZ_M68K_IL_H
+
+#include <rz_analysis.h>
+#include <rz_il.h>
+#include <capstone/capstone.h>
+
+#ifdef CAPSTONE_M68K_H
```
Rot127: When is `CAPSTONE_M68K_H` not defined?

librz/arch/p/asm/asm_m68k_cs.c
```diff
@@ -11,28 +11,36 @@
...
+
 static int m68k_disassemble(const RzAsm *a, RzAsmOp *op, const ut8 *buf, int len) {
 	if (!buf) {
 		return -1;
 	}
 	CapstoneContext *ctx = (CapstoneContext *)a->plugin_data;
 	char *buf_asm = NULL;
 	cs_insn *insn = NULL;
-	int ret = 0, n = 0;
+	int op_size = -1, cs_ret = CS_ERR_OK;
```
Rot127: Don't mix types here.
cs_ret is an cs_err enum.

librz/arch/p/asm/asm_m68k_cs.c
```diff
@@ -11,28 +11,36 @@
...
+	if (len < M68K_MIN_OP_SIZE) {
+		goto beach;
+	}
+
 	if (!ctx->handle) {
-		ret = cs_open(CS_ARCH_M68K, mode, &ctx->handle);
-		if (ret) {
-			ret = -1;
+		cs_ret = cs_open(CS_ARCH_M68K, mode, &ctx->handle);
+		if (cs_ret != CS_ERR_OK) {
```
Rot127:
```suggestion
		if (cs_open(CS_ARCH_M68K, mode, &ctx->handle) != CS_ERR_OK) {
```

librz/arch/p/asm/asm_m68k_cs.c
```diff
@@ -43,44 +51,35 @@ static int m68k_disassemble(const RzAsm *a, RzAsmOp *op, const ut8 *buf, int len
...
 	}
 
-	if (len > M68K_LONGEST_INSTRUCTION) {
-		len = M68K_LONGEST_INSTRUCTION;
-	}
-
-	n = cs_disasm(ctx->handle, buf, len, a->pc, 1, &insn);
-	if (n < 1) {
-		ret = -1;
+	cs_ret = cs_disasm(ctx->handle, buf, len, a->pc, 1, &insn);
```
Rot127: cs_disasm returns size_t not cs_err

librz/arch/p/asm/asm_m68k_cs.c
```diff
@@ -43,44 +51,35 @@ static int m68k_disassemble(const RzAsm *a, RzAsmOp *op, const ut8 *buf, int len
...
-	}
-	if (insn->size < 1) {
-		ret = -1;
+	if (insn->id == M68K_INS_INVALID || insn->size < 1) {
+		op_size = m68k_set_invalid(op, insn->size > 0 ? insn->size : M68K_MIN_OP_SIZE);
 		goto beach;
 	}
-	if (op && !op->size) {
+	op_size = insn->size;
+	if (op) {
```
Rot127: Do the `!op` check right at the beginning of the function please.

librz/arch/p/asm/asm_m68k_cs.c
```diff
@@ -43,44 +51,35 @@ static int m68k_disassemble(const RzAsm *a, RzAsmOp *op, const ut8 *buf, int len
...
-	if (op && buf_asm) {
-		buf_asm = rz_str_replace(buf_asm, "$", "0x", true);
-		if (buf_asm) {
-			rz_str_replace_char(buf_asm, '#', 0);
-			rz_asm_op_set_asm(op, buf_asm);
+		if (!buf_asm) {
+			goto beach;
 		}
+		buf_asm = rz_str_replace(buf_asm, "$", "0x", true);
+		rz_str_replace_char(buf_asm, '#', 0);
```
Rot127:
```suggestion
		rz_str_replace_char(buf_asm, '#', '');
```

librz/arch/p/analysis/analysis_m68k_cs.c
```diff
@@ -1391,21 +1628,13 @@ static char *m68k_get_reg_profile(RzAnalysis *analysis) {
...
-		"gpr	fp6 	.32	104	0\n" // FPU register 6, 96bits to write and read max
-		"gpr	fp7 	.32	108	0\n" // FPU register 7, 96bits to write and read max
 		"gpr	fpcr	.32	112	0\n"
 		"gpr	fpsr	.32	116	0\n"
 		"gpr	fpiar	.32	120	0\n"
 		"gpr	sr 		.16	70	0\n" // only available for read and write access during supervisor mode
 		"gpr	sfc 	.32	124	0\n" // source function code register
 		"gpr	dfc		.32	128	0\n" // destination function code register
-		"gpr	usp		.32	60	0\n" // user stack point this is an shadow register of A7 user mode, SR bit 0xD is 0
+		"gpr	usp		.32	240	0\n" // user stack point this is an shadow register of A7 user mode, SR bit 0xD is 0
```
Rot127: Please move it down to the gap in the register list.

Rot127 (changes requested):
Still have to look at `m68k_il.c` and `analysis_m68k_cs.c`

librz/arch/isa/m68k/m68k_cs.c
b1llow: FP register operands use M68K_OP_REG; M68K_OP_FP_SINGLE and M68K_OP_FP_DOUBLE represent floating-point immediates.

librz/arch/isa/m68k/m68k_cs.c
b1llow: A7 is 32 bits. This function returns the byte delta for predecrement/postincrement, not the register width. A byte access through A7 changes the 32-bit stack pointer by 2 bytes to preserve stack word alignment.

librz/arch/isa/m68k/m68k_cs.c
b1llow: It is not a decoder bug. Capstone deliberately collapses some FPU read-modify-write encodings to one operand when source and destination are the same—for example, `f2000025` is printed as frem fp0. The extension word still carries the destination, so the synthetic operand expands that shorthand into a uniform two-operand form for analysis and IL.

librz/arch/isa/m68k/m68k_cs.c
Rot127: And the M68K_OP_INVALID? Is there some edge case missing in Capstone?

librz/arch/p/analysis/analysis_m68k_cs.c
```diff
@@ -257,6 +243,9 @@ static RzAnalysisValue *m68k_value_from_operand(RzAnalysis *a, csh handle, const
 			return NULL;
 		}
 		break;
+	case M68K_OP_INVALID:
+		rz_analysis_value_free(value);
```
Rot127, line 247: rz_warn_if_reached()

librz/arch/p/analysis/analysis_m68k_cs.c
```diff
@@ -301,6 +290,21 @@ static RzAnalysisValue *m68k_value_from_operand(RzAnalysis *a, csh handle, const
...
+	RzAnalysisValue *value = rz_analysis_value_new();
+	if (!value) {
+		return NULL;
+	}
+	value->access = access;
+	value->type = RZ_ANALYSIS_VAL_REG;
+	value->reg = m68k_reg_get(a, handle, reg);
+	if (!value->reg) {
+		rz_analysis_value_free(value);
+		return NULL;
```
Rot127, line 303: rz_warn_if_reached

librz/arch/p/analysis/analysis_m68k_cs.c
```diff
@@ -769,10 +963,17 @@ static void op_fillval(RzAnalysis *a, RzAnalysisOp *op, csh handle, cs_insn *ins
 	}
 }
 
+static inline void m68k_invalid_il_nop(RzAnalysisOp *op, RzAnalysisOpMask mask) {
+	if (mask & RZ_ANALYSIS_OP_MASK_IL) {
+		op->il_op = rz_il_op_new_nop();
```
Rot127: IIRC we shouldn't NOP invalid ops.
Because otherwise the VM can't distinguish between nop and invalid.
And invalid != nop.

The interpreter can replace it if necessary.

librz/arch/p/analysis/analysis_m68k_cs.c
```diff
@@ -813,9 +1009,16 @@ static int m68k_analyze_op(RzAnalysis *a, RzAnalysisOp *op, ut64 addr, const ut8
...
 	}
+	if (mask & RZ_ANALYSIS_OP_MASK_IL) {
+		op->il_op = rz_m68k_cs_get_il_op(ctx->handle, mode, insn, addr);
+	}
 	switch (insn->id) {
 	case M68K_INS_INVALID:
 		op->type = RZ_ANALYSIS_OP_TYPE_ILL;
+		if (mask & RZ_ANALYSIS_OP_MASK_IL) {
+			rz_il_op_effect_free(op->il_op);
+			op->il_op = rz_il_op_new_nop();
```
Rot127: Also here, no NOP.

librz/arch/isa/m68k/m68k_il.c
```diff
@@ -0,0 +1,5881 @@
+// SPDX-FileCopyrightText: 2026 RizinOrg <info@rizin.re>
+// SPDX-License-Identifier: LGPL-3.0-only
+
+#include "m68k/m68k_il.h"
+#include "m68k/m68k_cs.h"
+
+#include <string.h>
+
```
Rot127, line 7:
```suggestion
```

librz/arch/isa/m68k/m68k_il.c
```diff
@@ -0,0 +1,5881 @@
...
+#define M68K_FPSR_CC_I               25
+#define M68K_FPSR_CC_NAN             24
+#define M68K_FPSR_CC_MASK            ((1u << M68K_FPSR_CC_N) | (1u << M68K_FPSR_CC_Z) | (1u << M68K_FPSR_CC_I) | (1u << M68K_FPSR_CC_NAN))
+#define M68K_FPSR_QUOTIENT_SHIFT     16
+#define M68K_FPSR_QUOTIENT_MASK      (0xffu << M68K_FPSR_QUOTIENT_SHIFT)
+#define M68K_FMOVEM_FP_REG_BITS_MASK 0x00ff0000u
+#define M68K_FMOVEM_EXTENDED_BITS    96
+#define M68K_FMOVEM_EXTENDED_BYTES   12
+
+#ifndef M68K_BF_REG_FLAG
```
Rot127: Does this get defined somewhere?

librz/arch/isa/m68k/m68k_il.c
```diff
@@ -0,0 +1,5881 @@
...
+	const char *addr_local;
+	RzILOpEffect *post;
+} M68KRWOperand;
+
+typedef struct {
+	const cs_m68k_op *op;
+	bool memory;
+	const char *addr_local;
+	RzILOpEffect *post;
+} M68KBitfieldTarget;
```
Rot127: Please document them.

librz/arch/isa/m68k/m68k_il.c
```diff
@@ -0,0 +1,5881 @@
...
+static RzILOpPure *fpu_ext96_to_fp80(RzILOpPure *ext96);
+static void m68k_ea_fini(M68KEA *ea);
+
+static RzILOpPure *u32(ut64 v) {
+	return U32((ut32)v);
+}
+
+static RzILOpPure *s32(st64 v) {
+	return S32((st32)v);
+}
```
Rot127: Why not use `S32`/`U32` directly?

librz/arch/isa/m68k/m68k_il.c
```diff
@@ -0,0 +1,5881 @@
...
+	}
+	if (reg == M68K_REG_CCR) {
+		return bits == 8 ? cast_unsigned(8, VARG("sr")) : cast_unsigned(bits, VARG("sr"));
+	}
+	if (reg == M68K_REG_SR) {
+		return bits == 16 ? VARG(name) : cast_unsigned(bits, VARG(name));
+	}
+	if (rz_m68k_reg_is_fpu(reg)) {
+		return bits == 80 ? VARG(name) : cast_unsigned(bits, VARG(name));
+	}
```
Rot127: Please convert them to a switch.
Not sure if it is optimized by the compile with two function invocations in it.

librz/arch/isa/m68k/m68k_il.c
```diff
@@ -0,0 +1,5881 @@
...
+	return seq;
+}
+
+static RzILOpEffect *set_ccr_from_value(RzILOpPure *new_ccr) {
+	RzILOpPure *ccr8 = cast_unsigned(8, new_ccr);
+	return SETG("sr", LOGOR(LOGAND(VARG("sr"), U16(0xff00)), cast_unsigned(16, ccr8)));
+}
+
+static RzILOpPure *ccr_bit(ut32 bit) {
```
Rot127:
```suggestion
static RzILOpBool *ccr_bit(ut32 bit) {
```

Rot127: Consider adding an assert about the bit not being < 8

librz/arch/isa/m68k/m68k_il.c
```diff
@@ -0,0 +1,5881 @@
...
+	return NON_ZERO(LOGAND(cast_unsigned(8, VARG("sr")), U8(1u << bit)));
+}
+
+static RzILOpBool *fpsr_bit(ut32 bit) {
+	return NON_ZERO(LOGAND(cast_unsigned(32, VARG("fpsr")), U32(1u << bit)));
```
Rot127: Consider adding an assert about the bit not being < 32

librz/arch/isa/m68k/m68k_il.c
```diff
@@ -0,0 +1,5881 @@
...
+	}
+	if (reg == M68K_REG_PC) {
+		return JMP(cast_unsigned(32, value));
+	}
+	if (reg == M68K_REG_CCR) {
+		return set_ccr_from_value(value);
+	}
+	if (reg == M68K_REG_SR) {
+		bool has_master_stack = ctx->mode == CS_MODE_M68K_020 || ctx->mode == CS_MODE_M68K_030 ||
+			ctx->mode == CS_MODE_M68K_040 || ctx->mode == CS_MODE_M68K_060;
```
Rot127: Move into static inline bool function

librz/arch/isa/m68k/m68k_il.c
```diff
@@ -0,0 +1,5881 @@
...
+			OR(EQ(VARL("format"), U16(0)), EQ(VARL("format"), U16(1))),
+			OR(EQ(VARL("format"), U16(2)), EQ(VARL("format"), U16(12))));
+	}
+#endif
+	return OR(
+		OR(OR(EQ(VARL("format"), U16(0)), EQ(VARL("format"), U16(1))),
+			OR(EQ(VARL("format"), U16(2)), EQ(VARL("format"), U16(3)))),
+		OR(OR(EQ(VARL("format"), U16(4)), EQ(VARL("format"), U16(7))),
+			OR(OR(EQ(VARL("format"), U16(8)), EQ(VARL("format"), U16(9))),
+				OR(EQ(VARL("format"), U16(10)), EQ(VARL("format"), U16(11))))));
```
Rot127: Couldn't it check against not being 5,6 and >11?

Rot127: Same for the things above. It, they probably can be shortened quite a bit.

librz/arch/isa/m68k/m68k_il.c
```diff
@@ -0,0 +1,5881 @@
...
+static RzILOpPure *rte_frame_size(cs_mode mode) {
+	RzILOpPure *size = ITE(EQ(VARL("format"), U16(0)), U32(8),
+		ITE(EQ(VARL("format"), U16(1)), U32(8),
+			ITE(EQ(VARL("format"), U16(2)), U32(12),
+				ITE(EQ(VARL("format"), U16(3)), U32(12),
+					ITE(EQ(VARL("format"), U16(4)), U32(16),
+						ITE(EQ(VARL("format"), U16(7)), U32(60),
+							ITE(EQ(VARL("format"), U16(8)), U32(58),
+								ITE(EQ(VARL("format"), U16(9)), U32(20),
+									ITE(EQ(VARL("format"), U16(10)), U32(32), U32(92))))))))));
```
Rot127: I am not sure if possible currently. But defining an extra memory space, which only has a lookup table for these values would be way more efficient.
Can you check out if it is possible? If not, I guess it is fine.

librz/arch/isa/m68k/m68k_il.c
```diff
@@ -0,0 +1,5881 @@
...
+#else
+	(void)mode;
+#endif
+	return size;
+}
+
+static RzILOpPure *read_mem_sized(ut32 bits, RzILOpPure *addr) {
+	if (bits == 8) {
+		return LOAD(addr);
+	}
```
Rot127:
```suggestion
```

Or just use LOADW anywhere?

librz/arch/isa/m68k/m68k_il.c
```diff
@@ -0,0 +1,5881 @@
...
+	return LOADW(bits, addr);
+}
+
+static RzILOpEffect *write_mem_sized(ut32 bits, RzILOpPure *addr, RzILOpPure *value) {
+	if (bits == 8) {
+		return STORE(addr, cast_unsigned(8, value));
+	}
+	return STOREW(addr, cast_unsigned(bits, value));
```
Rot127: Same as above

librz/arch/isa/m68k/m68k_il.c
```diff
@@ -0,0 +1,5881 @@
...
+}
+
+static RzILOpPure *index_value(M68KILCtx *ctx, const cs_m68k_op *op) {
+	if (op->mem.index_reg == M68K_REG_INVALID) {
+		return U32(0);
+	}
+	RzILOpPure *idx = read_reg_sized(ctx, op->mem.index_reg, op->mem.index_size ? 32 : 16);
+	if (!idx) {
+		return NULL;
```
Rot127: rz_warn_if_reached

librz/arch/isa/m68k/m68k_il.c
```diff
@@ -0,0 +1,5881 @@
...
+			return 4;
+		}
+		if (op->br_disp.disp_size == M68K_OP_BR_DISP_SIZE_WORD) {
+			return 2;
+		}
+		return 0;
+	case M68K_OP_MEM:
+		return mem_extension_bytes(op);
+	default:
+		return 0;
```
Rot127: rz_warn_if_reached?

librz/arch/isa/m68k/m68k_il.c
```diff
@@ -0,0 +1,5881 @@
...
+	case M68K_OP_FP_DOUBLE:
+		return 8;
+	case M68K_OP_BR_DISP:
+		if (op->br_disp.disp_size == M68K_OP_BR_DISP_SIZE_LONG) {
+			return 4;
+		}
+		if (op->br_disp.disp_size == M68K_OP_BR_DISP_SIZE_WORD) {
+			return 2;
+		}
+		return 0;
```
Rot127: rz_warn_if_reached?

librz/arch/isa/m68k/m68k_il.c
```diff
@@ -0,0 +1,5881 @@
...
+			return offset;
+		}
+		if (!found) {
+			fallback = offset;
+			found = true;
+		}
+	}
+	return fallback;
+}
+#endif
```
Rot127: Can you please move these special functions handling the old Capstone output inot another file please? m68k is already extremely big.

librz/arch/isa/m68k/m68k_il.c
```diff
@@ -0,0 +1,5881 @@
...
+	return branch_disp_base(ctx);
+#else
+	return ctx->addr + pc_relative_extension_offset(ctx, op);
+#endif
+}
+
+static RzILOpPure *branch_disp_target(const M68KILCtx *ctx, const cs_m68k_op *op) {
+	rz_return_val_if_fail(ctx && op, NULL);
+	if (op->type != M68K_OP_BR_DISP) {
+		return NULL;
```
Rot127: rz_warn_if_reached?

librz/arch/isa/m68k/m68k_il.c
```diff
@@ -0,0 +1,5881 @@
...
+	}
+	return u32(branch_disp_base(ctx) + op->br_disp.disp);
+}
+
+static RzILOpEffect *set_addr_reg_delta(M68KILCtx *ctx, m68k_reg reg, st32 delta) {
+	if (!rz_m68k_reg_is_areg(reg)) {
+		return NULL;
```
Rot127: rz_warn_if_reached?

librz/arch/isa/m68k/m68k_il.c
```diff
@@ -0,0 +1,5881 @@
...
+
+static RzILOpPure *cast_signed(ut32 bits, RzILOpPure *v) {
+	return CAST(bits, MSB(DUP(v)), v);
+}
+
+static RzILOpPure *extend_to_32(RzILOpPure *v, bool sign_extend) {
+	return sign_extend ? cast_signed(32, v) : cast_unsigned(32, v);
+}
+
+static RzILOpPure *append_addr_add(RzILOpPure *base, RzILOpPure *delta) {
```
Rot127:
```suggestion
static RzILOpPure *addr_add(RzILOpPure *base, RzILOpPure *delta) {
```

append suggests it actually uses `APPEND` somewhere.

librz/arch/isa/m68k/m68k_il.c
```diff
@@ -0,0 +1,5881 @@
...
+		ea.addr = ADD(ADD(read_mem_sized(32, ADD(base, s32(op->mem.in_disp))), idx), s32(op->mem.out_disp));
+		break;
+	case M68K_AM_PCI_DISP:
+		ea.addr = u32((st64)(ut32)pc_relative_base(ctx, op) + op->mem.disp);
+		break;
+	case M68K_AM_PCI_INDEX_8_BIT_DISP:
+	case M68K_AM_PCI_INDEX_BASE_DISP: {
+		st64 disp = op->address_mode == M68K_AM_PCI_INDEX_BASE_DISP ? op->mem.in_disp : op->mem.disp;
+		idx = index_value(ctx, op);
+		ea.addr = ADD(u32((st64)(ut32)pc_relative_base(ctx, op) + disp), idx);
```
Rot127: Those castings seem unnecessary

librz/arch/isa/m68k/m68k_il.c
```diff
@@ -0,0 +1,5881 @@
...
+		break;
+	default:
+		if (rz_m68k_op_is_mem_addr(op) && base_reg != M68K_REG_INVALID) {
+			ea.addr = ADD(reg_value(ctx, base_reg), s32(op->mem.disp));
+		} else if (rz_m68k_op_is_mem_addr(op)) {
+			ea.addr = u32(rz_m68k_op_absolute_address(op, false));
+		}
+		break;
+	}
+	(void)bytes;
```
Rot127:
```suggestion
```

librz/arch/isa/m68k/m68k_il.c
```diff
@@ -0,0 +1,5881 @@
...
+	RzILOpPure *res = delta >= 0 ? ADD(base, U32((ut32)delta)) : SUB(base, U32((ut32)-delta));
+	return write_reg_sized(ctx, reg, 32, res);
+}
+
+static M68KEA effective_addr(M68KILCtx *ctx, const cs_m68k_op *op, ut32 bits) {
+	M68KEA ea = { 0 };
+	RzILOpPure *base = NULL;
+	RzILOpPure *idx = NULL;
+	m68k_reg base_reg = rz_m68k_op_base_reg(op);
+	ut32 bytes = rz_m68k_bits_access_bytes(bits);
```
Rot127:
```suggestion
```

librz/arch/isa/m68k/m68k_il.c
```diff
@@ -0,0 +1,5881 @@
...
+		*pre = NULL;
+	}
+	if (post) {
+		*post = NULL;
+	}
+	if (rz_m68k_op_is_mem_addr(op)) {
+		M68KEA ea = effective_addr(ctx, op, bits);
+		if (!ea.addr) {
+			m68k_ea_fini(&ea);
+			return NULL;
```
Rot127: rz_warn_if_reached?

librz/arch/isa/m68k/m68k_il.c
```diff
@@ -0,0 +1,5881 @@
...
+		}
+		if (pre) {
+			*pre = ea.pre;
+		}
+		if (post) {
+			*post = ea.post;
+		}
+		return read_mem_sized(bits, ea.addr);
```
Rot127: The ea pre/post will leak here if pre/post are NULL

librz/arch/isa/m68k/m68k_il.c
```diff
@@ -0,0 +1,5881 @@
...
+	case M68K_OP_IMM:
+		return UN(bits, op->imm);
+	case M68K_OP_BR_DISP:
+		return branch_disp_target(ctx, op);
+	case M68K_OP_FP_SINGLE:
+		return F2BV(F32(op->simm));
+	case M68K_OP_FP_DOUBLE:
+		return F2BV(F64(op->dimm));
+	default:
+		return NULL;
```
Rot127: rz_warn-if_reached?

librz/arch/isa/m68k/m68k_il.c
```diff
@@ -0,0 +1,5881 @@
...
+			m68k_ea_fini(&ea);
+			return NULL;
+		}
+		if (pre) {
+			*pre = ea.pre;
+		}
+		if (post) {
+			*post = ea.post;
+		}
+		return write_mem_sized(bits, ea.addr, value);
```
Rot127: leak of ea.pre/post

librz/arch/isa/m68k/m68k_il.c
```diff
@@ -0,0 +1,5881 @@
...
+	}
+	if (post) {
+		*post = NULL;
+	}
+	if (rz_m68k_op_is_mem_addr(op)) {
+		M68KEA ea = effective_addr(ctx, op, bits);
+		if (!ea.addr) {
+			rz_il_op_pure_free(value);
+			m68k_ea_fini(&ea);
+			return NULL;
```
Rot127: rz_warn_if_reached

librz/arch/isa/m68k/m68k_il.c
```diff
@@ -0,0 +1,5881 @@
...
+	}
+}
+
+static RzILOpEffect *m68k_label(const char *label) {
+	if (!strcmp(label, "m68k_illegal")) {
+		RzILOpEffect *seq = SETL("trap_op", U32((ut32)M68K_TRAP_OP_ILLEGAL));
+		seq = seq_append(seq, SETL("trap_vector", U32((ut32)M68K_VECTOR_ILLEGAL)));
+		return seq_append(seq, GOTO(label));
+	}
+	if (!strcmp(label, "m68k_privilege")) {
```
Rot127:
```suggestion
	if (RZ_STR_EQ(label, "m68k_illegal")) {
		RzILOpEffect *seq = SETL("trap_op", U32((ut32)M68K_TRAP_OP_ILLEGAL));
		seq = seq_append(seq, SETL("trap_vector", U32((ut32)M68K_VECTOR_ILLEGAL)));
		return seq_append(seq, GOTO(label));
	}
	if (RZ_STR_EQ(label, "m68k_privilege")) {
```

librz/arch/isa/m68k/m68k_il.c
```diff
@@ -0,0 +1,5881 @@
...
+	}
+	return GOTO(label);
+}
+
+static RzILOpEffect *guard_supervisor(RzILOpEffect *effect) {
+	return effect ? BRANCH(supervisor_mode(), effect, m68k_label("m68k_privilege")) : NULL;
+}
+
+static RzILOpEffect *m68k_invalid(void) {
+	return NOP();
```
Rot127: invalid != NOP

Rot127 (changes requested):
Came until m68k_il.c:919. Rest is still missing.

- [ ] Please move all these small helper and auxiliary functions `fini`, `is_x_reg` etc into a separate files.

librz/arch/p/analysis/analysis_m68k_cs.c
b1llow: `M68K_OP_INVALID` comes from Capstone’s decoded data, so it represents a possible decoder/input failure rather than an unreachable internal state.

The callers already handle this failure normally:

- `m68k_value_from_operand()` frees `value` and returns `NULL`.
- `m68k_add_src_value()` and `m68k_set_dst_value()` safely ignore `NULL`.

Using `rz_warn_if_reached()` here would emit misleading programmer-error warnings when analyzing malformed or unsupported machine code.

Therefore, the ordinary `return NULL` is intentional. The `rz_warn_if_reached()` is not appropriate for decoder-controlled input.

I assume you suggested using `rz_warn_if_reached` just to make debugging easier?

librz/arch/isa/m68k/m68k_cs.c
b1llow: M68K_OP_INVALID is not an FPU operand type. Capstone can leave it on malformed or unsupported encodings; we skip it rather than treating it as FPU. Not a Capstone bug to fix here.

librz/arch/isa/m68k/m68k_il.c
b1llow: Yes — it is a local fallback for Capstone 4/5. Newer Capstone already defines these helpers, so the `#ifndef` keeps the encoding local to the lifter.

librz/arch/isa/m68k/m68k_il.c
b1llow: Not currently possible: analysis IL memory is IO-backed only (`RzAnalysisILMemType` has no ROM-table type, and `RzAnalysisILConfig` has no hook to preload a lookup mem). Nested ITE stays.

librz/arch/p/analysis/analysis_m68k_cs.c
b1llow: Done — invalid ops now leave `il_op` as NULL instead of NOP.

librz/arch/p/analysis/analysis_m68k_cs.c
b1llow: Done — `M68K_INS_INVALID` now sets `il_op` to NULL.

librz/arch/isa/m68k/m68k_il.c
b1llow: Done — `m68k_invalid()` now returns NULL, not NOP().

librz/arch/isa/m68k/m68k_il.c
b1llow: Fixed — ownership of `pre`/`post` is transferred to the caller, then `m68k_ea_fini()` releases anything left.

librz/arch/isa/m68k/m68k_il.c
b1llow: Fixed — same as the read path: transfer `pre`/`post`, then `m68k_ea_fini()`.

librz/arch/isa/m68k/m68k_il.c
b1llow: Moved the old-Capstone PC-relative helpers (and the other small EA/flag/reg auxiliaries) into `m68k_il_helpers.c`.

librz/arch/p/analysis/analysis_m68k_cs.c
b1llow: Same as the `M68K_OP_INVALID` case: this is a decoder/input failure, not an unreachable internal state. Callers already handle NULL.

librz/arch/isa/m68k/m68k_il.c
b1llow: Same as the `M68K_OP_INVALID` case: this is a decoder/input failure, not an unreachable internal state. Callers already handle NULL.

librz/arch/isa/m68k/m68k_il.c
b1llow: Same as the `M68K_OP_INVALID` case: unknown operand types are decoder/input failures, not an unreachable internal state.

librz/arch/isa/m68k/m68k_il.c
b1llow: Same as the `M68K_OP_INVALID` case: a non-branch operand here is decoder/input failure, not an unreachable internal state. Callers already handle NULL.

librz/arch/isa/m68k/m68k_il.c
b1llow: Same as the `M68K_OP_INVALID` case: a non-address register here is decoder/input failure, not an unreachable internal state. Callers already handle NULL.

librz/arch/isa/m68k/m68k_il.c
b1llow: Same as the `M68K_OP_INVALID` case: a missing EA is a decoder/input failure, not an unreachable internal state. Callers already handle NULL.

librz/arch/isa/m68k/m68k_il.c
b1llow: Same as the `M68K_OP_INVALID` case: unknown operand types are decoder/input failures, not an unreachable internal state. Callers already handle NULL.

librz/arch/isa/m68k/m68k_il.c
b1llow: Same as the `M68K_OP_INVALID` case: a missing EA is a decoder/input failure, not an unreachable internal state. Callers already handle NULL.

librz/arch/isa/m68k/m68k_il.c
b1llow: Byte displacements have no extension word, so returning 0 is the valid case, not unreachable.

b1llow:
Helpers (`fini`, EA, flags, old-Capstone PC-relative, `is_*_reg`, etc.) from the first part of `m68k_il.c` are now in `librz/arch/isa/m68k/m68k_il_helpers.c` / `m68k_il_helpers.h`.

librz/arch/isa/m68k/m68k_il_helpers.c
```diff
@@ -0,0 +1,826 @@
...
+
+#include "m68k/m68k_il_helpers.h"
+
+#include <rz_il/rz_il_opbuilder_begin.h>
+#include <rz_util/rz_assert.h>
+
+RZ_IPI RzILOpPure *m68k_reg_value(M68KILCtx *ctx, m68k_reg reg) {
+	const char *name = cs_reg_name(ctx->handle, reg);
+	if (!name) {
+		return NULL;
```
Rot127, line 13: rz_warn_if_reached.

librz/arch/isa/m68k/m68k_il_helpers.c
```diff
@@ -0,0 +1,826 @@
...
+	}
+	return VARG(name);
+}
+
+RZ_IPI RzILOpPure *m68k_read_reg_sized(M68KILCtx *ctx, m68k_reg reg, ut32 bits) {
+	const char *name = cs_reg_name(ctx->handle, reg);
+	if (!name) {
+		return NULL;
```
Rot127, line 22: rz_warn_if_reached.

librz/arch/isa/m68k/m68k_il_helpers.c
```diff
@@ -0,0 +1,826 @@
...
+	m68k_reg index_reg = ((word & 0x8000) ? M68K_REG_A0 : M68K_REG_D0) +
+		((word >> 12) & 7);
+	if (op->mem.index_reg != index_reg) {
+		return false;
+	}
+	if (op->mem.index_size != ((word & 0x0800) ? 1 : 0)) {
+		return false;
+	}
+	ut8 scale = 1 << ((word >> 9) & 3);
+	return (op->mem.scale ? op->mem.scale : 1) == scale;
```
Rot127: Please make the magic values used in the `word & ...` checks enums with descriptive names.

Also, `word` is a word of the instruction bytes? If so, please rename it to something like `insn_op_enc`, `insn_word`, `insn_ops_word`. Anything which shows where the word is coming from. You probably can come up with a better name.

librz/arch/isa/m68k/m68k_il_helpers.h
```diff
@@ -0,0 +1,199 @@
...
+/**
+ * Read/write operand state. op and addr_local are borrowed; post is owned and
+ * must be transferred or released with m68k_rw_operand_fini().
+ */
+typedef struct {
+	const cs_m68k_op *op;
+	const char *addr_local;
+	RzILOpEffect *post;
+} M68KRWOperand;
+
```
Rot127, line 63: Please also add doxygen to the members.
Also remove the "borrowing" and "owning" stuff. It is clear from the context, because they are const and not const.

Rot127: Same for M68KBitfieldTarget

librz/arch/isa/m68k/m68k_il_helpers.c
```diff
@@ -0,0 +1,826 @@
...
+	default:
+		break;
+	}
+	if (rz_m68k_reg_name_is_mmu_root_pointer(name)) {
+		return bits == 64 ? VARG(name) : UNSIGNED(bits, VARG(name));
+	}
+	return bits == 32 ? VARG(name) : UNSIGNED(bits, VARG(name));
+}
+
+RZ_IPI RzILOpEffect *m68k_seq_append(RzILOpEffect *a, RzILOpEffect *b) {
```
Rot127: Please remove this and use simply `SEQ2` everywhere.

The reason is, that this fails silently. And I seems to be barely any case where NULL was given as valid `a` or `b`.

librz/arch/isa/m68k/m68k_il.c
```diff
@@ -0,0 +1,5000 @@
...
+	seq = append_fpu_state_metadata(ctx, seq);
+	return m68k_guard_supervisor(m68k_seq_append(seq, m68k_label("m68k_fpu")));
+}
+
+static RzILOpPure *callm_frame_addr(ut32 offset) {
+	return offset ? ADD(VARL("callm_frame"), U32(offset)) : VARL("callm_frame");
+}
+
+static m68k_reg callm_reg_from_code(ut32 code) {
+	static const m68k_reg regs[] = {
```
Rot127: Move outside of function.

librz/arch/isa/m68k/m68k_il.c
```diff
@@ -0,0 +1,5000 @@
...
+	return effect;
+}
+
+static RzILOpEffect *callm_load_selected_reg(M68KILCtx *ctx) {
+	RzILOpEffect *effect = EMPTY();
+	for (st32 code = 14; code >= 0; code--) {
+		m68k_reg reg = callm_reg_from_code((ut32)code);
+		RzILOpEffect *write = m68k_write_reg_sized(ctx, reg, 32, VARL("callm_data"));
+		if (!write) {
+			return NULL;
```
Rot127, line 561: It seems like all these cases of NULLs being returned for RzILOpEffects, should warn if reached.
Because it is not expected to fail, right?

librz/arch/isa/m68k/m68k_il.c
```diff
@@ -0,0 +1,5000 @@
...
+		return M68K_MMU_OP_MOVE;
+	case M68K_INS_PLPAR:
+	case M68K_INS_PLPAW:
+		return M68K_MMU_OP_LOAD_PHYSICAL;
+	default:
+		return M68K_MMU_OP_MOVE;
+	}
+}
+
+static int mmu_read_write_access(ut32 insn_id) {
```
Rot127, line 709: Document meaning of return values.

librz/arch/isa/m68k/m68k_il.c
```diff
@@ -0,0 +1,5000 @@
...
+				return NULL;
+			}
+			seq = m68k_seq_append(seq, SETL("mmu_addr", UNSIGNED(M68K_ADDR_BITS, addr)));
+		} else if (op->type == M68K_OP_IMM) {
+			seq = m68k_seq_append(seq, SETL("mmu_addr", U32((ut32)op->imm)));
+		} else if (!rz_m68k_op_is_mem_addr(op)) {
+			if (op->type != M68K_OP_REG || append_unmodeled_mmu) {
+				return m68k_label("m68k_mmu");
+			}
+			return NOP();
```
Rot127: Is this really a NOP in this case?

librz/arch/isa/m68k/m68k_il.c
```diff
@@ -0,0 +1,5000 @@
...
+			seq = m68k_seq_append(seq, ea.pre);
+			seq = m68k_seq_append(seq, SETL("mmu_addr", UNSIGNED(M68K_ADDR_BITS, ea.addr)));
+			seq = m68k_seq_append(seq, ea.post);
+		}
+	}
+	if (append_unmodeled_mmu) {
+		seq = append_mmu_metadata(ctx, seq);
+		return m68k_seq_append(seq, m68k_label("m68k_mmu"));
+	}
+	return seq ? seq : NOP();
```
Rot127: Also here, is this a NOP? Seems more like an error.

librz/arch/isa/m68k/m68k_il.c
```diff
@@ -0,0 +1,5000 @@
...
+	RzILOpEffect *write_acc = m68k_write_reg_sized(ctx, acc_op->reg, 32, VARL("res"));
+	if (!write_acc) {
+		rz_il_op_effect_free(seq);
+		return NULL;
+	}
+	return m68k_seq_append(seq, write_acc);
+}
+
+static RzILOpEffect *lift_mac_dual_acc(M68KILCtx *ctx, bool subtract_first, bool subtract_second) {
+	if (ctx->m68k->op_count < 4) {
```
Rot127: Please make all the checks op_counts asserts.

librz/arch/isa/m68k/m68k_il.c
```diff
@@ -0,0 +1,5000 @@
...
+	seq = m68k_seq_append(seq, SETL("dst", APPEND(m68k_read_reg_sized(ctx, high_reg, 32), m68k_read_reg_sized(ctx, low_reg, 32))));
+	RzILOpEffect *normal = SEQ5(
+		SETL("quot", div_quotient_expr(sign, "dst", "src")),
+		SETL("rem", div_remainder_expr(sign, "dst", "src")),
+		write_low,
+		write_high,
+		m68k_set_flags_div(VARL("quot"), 32, IL_FALSE));
+	RzILOpBool *overflow = div_quotient_overflows(sign, 32, div_quotient_expr(sign, "dst", "src"));
+	RzILOpEffect *do_divide = BRANCH(overflow, m68k_set_flags_div(U32(0), 32, IL_TRUE), normal);
+	return m68k_seq_append(seq, BRANCH(IS_ZERO(VARL("src")), m68k_exception(M68K_TRAP_OP_DIV_ZERO, M68K_VECTOR_ZERO_DIVIDE, "m68k_trap"), do_divide));
```
Rot127: Division by zero is handled by our VM.
Why is the check done here?

Rot127: Also for the other cases.

librz/arch/isa/m68k/m68k_il.c
```diff
@@ -0,0 +1,5000 @@
...
+}
+
+static RzILOpEffect *lift_div(M68KILCtx *ctx, bool sign) {
+	if (ctx->m68k->op_count < 2) {
+		return NULL;
+	}
+	const cs_m68k_op *dst = &ctx->m68k->operands[1];
+	ut32 bits = rz_m68k_detail_op_bits(ctx->m68k, 16);
+	if (bits != 16 && bits != 32) {
```
Rot127: These bit checks seem also like asserts to me, are they?

librz/arch/isa/m68k/m68k_il.c
```diff
@@ -0,0 +1,5000 @@
...
+	} else {
+		seq = m68k_seq_append(seq, SETL("src", m68k_read_reg_sized(ctx, src->reg, bits)));
+		seq = m68k_seq_append(seq, bits == 16 ? SEQ2(movep_store_byte(0, 8), movep_store_byte(2, 0)) : SEQ4(movep_store_byte(0, 24), movep_store_byte(2, 16), movep_store_byte(4, 8), movep_store_byte(6, 0)));
+	}
+	seq = m68k_seq_append(seq, ea.post);
+	return seq;
+}
+
+static m68k_reg movem_reg_for_bit(ut32 bit) {
+	static const m68k_reg regs[] = {
```
Rot127: I think this array existed above already.

librz/arch/isa/m68k/m68k_il.c
```diff
@@ -0,0 +1,5000 @@
...
+		M68K_REG_D0, M68K_REG_D1, M68K_REG_D2, M68K_REG_D3,
+		M68K_REG_D4, M68K_REG_D5, M68K_REG_D6, M68K_REG_D7,
+		M68K_REG_A0, M68K_REG_A1, M68K_REG_A2, M68K_REG_A3,
+		M68K_REG_A4, M68K_REG_A5, M68K_REG_A6, M68K_REG_A7
+	};
+	return bit < RZ_ARRAY_SIZE(regs) ? regs[bit] : M68K_REG_INVALID;
+}
+
+static ut32 movem_reg_count(ut32 reg_bits) {
+	ut32 count = 0;
```
Rot127: There should be a bit counting function in rz_bits.h.
If not, please add it there.

librz/arch/isa/m68k/m68k_il_helpers.h
```diff
@@ -0,0 +1,199 @@
...
+	RzILOpEffect *post;
+} M68KBitfieldTarget;
+
+typedef enum {
+	M68K_BIN_ADD,
+	M68K_BIN_SUB,
+	M68K_BIN_AND,
+	M68K_BIN_OR,
+	M68K_BIN_XOR
+} M68KBinOp;
```
Rot127: These ones seem to be a used as places where a `RzILOpPureCode` would be sufficient.
I would vote for removing them and use the `RzILOpPureCode` values instead. To reduce code duplication. But your call.

Rot127 (changes requested):
Got until `m68k_il.c::1742 - lift_addx_subx`.

Since it is an enormous amount of code (as any lifting), we definitely need to trace test IMHO.
QEMU seems tp support it: https://wiki.qemu.org/Documentation/Platforms/m68k

librz/arch/isa/m68k/m68k_il_helpers.c
b1llow: Added `rz_warn_if_reached()` when `cs_reg_name()` returns NULL. A register we chose to look up is expected to have a Capstone name.

librz/arch/isa/m68k/m68k_il_helpers.c
b1llow: Same here, and on the matching write path in `m68k_write_reg_sized()`.

librz/arch/isa/m68k/m68k_il_helpers.c
b1llow: Named the extension-word bits (`M68K_EXT_WORD_FULL_FORMAT`, `M68K_EXT_WORD_INDEX_SUPPRESS`, `M68K_EXT_WORD_INDEX_ADDR_REG`, `M68K_EXT_WORD_INDEX_LONG`) and renamed the parameter to `ext_word`.

librz/arch/isa/m68k/m68k_il_helpers.h
b1llow: Added member docs on `M68KILCtx`, `M68KEA`, `M68KRWOperand`, and `M68KBitfieldTarget`, and dropped the borrowed/owned wording. Const vs non-const is enough.

librz/arch/isa/m68k/m68k_il_helpers.c
b1llow: Removed `m68k_seq_append`. Sequences now use `SEQ2` (or `SEQ3` where the ops are known non-NULL). Optional EA `pre`/`post` are still skipped when they are NULL, which is a valid empty side-effect, not a failed lift.

librz/arch/isa/m68k/m68k_il.c
b1llow: Moved the D0–A7 table to file scope as `m68k_data_addr_regs`.

librz/arch/isa/m68k/m68k_il.c
b1llow: Agreed for these paths. `callm_store_selected_reg` / `callm_load_selected_reg` now `rz_warn_if_reached()` before returning NULL, since a selected D/A register should always lift.

librz/arch/isa/m68k/m68k_il.c
b1llow: Documented: `0` read, `1` write, `-1` when the instruction has no transfer direction.

librz/arch/isa/m68k/m68k_il.c
b1llow: Reworked during the SEQ2 refactor: the NOP() path is gone. Unmodeled MMU operand kinds now fall through to the `m68k_mmu` label instead of quietly lifting to a no-op.

librz/arch/isa/m68k/m68k_il.c
b1llow: Same refactor removed this trailing NOP() as well: `lift_mmu_address_effects()` now always appends the MMU metadata and ends with the `m68k_mmu` label, and lift failures return NULL.

librz/arch/isa/m68k/m68k_il.c
b1llow: Done. The operand-count checks guaranteed by the dispatcher are now `rz_return_val_if_fail()` asserts. Plain checks remain only where Capstone genuinely produces a variable operand count (coprocessor cpGEN forms and the PMOVE fallbacks), where fewer operands select the metadata-only lift rather than indicating a broken decode.

librz/arch/isa/m68k/m68k_il.c
b1llow: The RzIL VM defines a result for division by zero, but it does not raise the architectural M68K Zero Divide exception. The branch models the trap to vector 5 (M68000PRM, DIVU/DIVS: a zero divisor takes the Zero Divide exception) so consumers see the `m68k_trap` label and vector where hardware traps, instead of the VM's default division result. So the check is about exception semantics, not about protecting the VM; I kept it (same for the other DIV cases).

librz/arch/isa/m68k/m68k_il.c
b1llow: Yes, they are now `rz_return_val_if_fail(bits == 16 || bits == 32, NULL)` asserts.

librz/arch/isa/m68k/m68k_il.c
b1llow: Right. After moving the D0-A7 table to file scope, `movem_reg_for_bit()` now indexes the shared `m68k_data_addr_regs` instead of keeping its own copy.

librz/arch/isa/m68k/m68k_il.c
b1llow: `rz_bits.h` already provides it; `movem_reg_count()` now uses `rz_bits_count_ones_ut32()`.

librz/arch/isa/m68k/m68k_il_helpers.h
b1llow: Good idea, done: removed `M68KBinOp`; `lift_binop()`/`lift_status_binop()` now take an `RzILOpPureCode` (`RZ_IL_OP_ADD`/`RZ_IL_OP_SUB`/`RZ_IL_OP_LOGAND`/`RZ_IL_OP_LOGOR`/`RZ_IL_OP_LOGXOR`) directly.

Rot127:
When you have some time, please extract the Capstone update into a separate PR.

librz/arch/isa/m68k/m68k_il.c
```diff
@@ -0,0 +1,7699 @@
...
+static bool fpu_operand_to_float_local(M68KILCtx *ctx, const char *name, const char *raw80_name, const cs_m68k_op *op, ut32 bits, RzILOpEffect **seq) {
+	RzILOpFloat *value = NULL;
+	RzILOpEffect *pre = NULL;
+	RzILOpEffect *post = NULL;
+	bool raw80_set = false;
+
+	if (op->type == M68K_OP_MEM && op->address_mode == M68K_AM_NONE) {
+		return false;
+	}
+	if (rz_m68k_op_is_fpu_reg(op)) {
```
Rot127, line 5010: Please make these if conditions a switch (Github is broken, so can't check if `rz_m68k_op_is_fpu_reg` prevents that).

Rot127 (approved):
Guess it looks good. Hard to say with this size.
Some of the generated IL ops are massive though.
Optimizing them would be a nice to have

PR #6576, opened 2026-06-28
Implement file download from the remote machine in GDB protocol

### LINKED ISSUE #3975: Download file and symbols via GDB remote protocol
GDB Remote protocol allows the client to download the binary, if possible.

See:
- https://sourceware.org/gdb/download/onlinedocs/gdb/File-Transfer.html
- https://www.sourceware.org/gdb/onlinedocs/gdb/GDB_002fMI-File-Transfer-Commands.html

moste00:

**Detailed description**

There is an issue to download binary files from a remote machine during a GDB remote session https://github.com/rizinorg/rizin/issues/3975, this file is a simple implementation of that addresses the simplest case.

Implementation  as follows:

(1) An existing `gdbr_read_file` that implemented a bounded read of a remote file into a caller-provided buffer was refactored, it now takes an additional argument `start_offset`, and starts reading from the remote file at this offset. The more general logic is named `gdbr_pread_file`, and the behaviour of `gdbr_read_file`  is recovered by calling it with a `start_offset` of 0.

(2) Using `gdbr_pread_file` as a generic primitive, file downloading is then implemented as an outer loop repeatedly calling it with increasing offsets until a zero-length or shorter-than-expected read (indicating the remote file is complete).

(3) A sub-command `download_file` was added where the user controls both the remote source and the local destination

(4) A convenience was added to the main function so that if the user runs `rizin -D gdb gdb://...`, it offers to download the remote executable being deubugged if any of the following is true (a) the remote host is NOT localhost (b)  the file doesn't exist by its full path locally, even if the remote host is localhost. 

Note: the localhost check is intentionally conservative, it checks for a literal "gdb://localhost:" in the prefix of the URI, so (e.g.) it rejects `127.0.0.1` and considers it "not local". This is because parsing IP addresses doesn't have any well-established utilities in Rizin AFAICT, and I don't want to implement a half-baked buggy and slow implementation of half of an IP parser in the main function. Note that `127.0.0.1` is not unique as a loopback address anyway, and sometimes it can be behind container/namespace routing that make its filesystem not identical to the local filesystem Rizin is running on. 

So for now, `localhost` is treated as a magic escape when the user knows the file doesn't need to be downloaded, and any other remote address is treated as remote.

**Test plan**

There are no automated tests, sorry, it would be awkward to test network code like this.

Here's a concrete UX difference:

1- First, run a gdb server with a binary that is inaccessible to Rizin, this can be done with VMs or SSH, but a much simpler way to do it with bwrap is:

`sudo bwrap --ro-bind / / --dev-bind /dev /dev --proc /proc --tmpfs /tmp --bind "$tmp/remote-sleep" /tmp/remote-sleep gdbserver :2345 /tmp/remote-sleep 1000`

(bwrap is only a convience wrapper over namespaces here, so this is doable on any modern linux without bwrap as well.)

2- Run Rizin without this PR (or with this PR but select no when the prompt for downloading the remote file appears)

3- run `is`, the reasonable expectation is: symbols should appear. But the error `ERROR: No binary object currently selected.` appears instead. 

4- Even after running a full `aaaa` analysis pass, a simple debug command like `dcu main` doesn't actually stop at main. The error `WARNING: Breakpoint won't be placed since it's not in a valid map.
You can bypass this check by setting dbg.bpinmaps to false.` appears instead, and the sleep command appears to run to completion before control is handed back to Rizin.

With this PR, choosing `y` at the prompt will make (3) and (4) work as expected. 

**Closing issues**

Closes https://github.com/rizinorg/rizin/issues/3975 but might need extra features.

**Future/RFC**

The following concerns are reasonably too big/unknown to handle in this PR, but arguably should:

1- If the binary being debugged in the remote machine loads `.so` objects, does the current prompt in main naturally extend to download them as well ? should it ? what if loading is lazy and happens mid-execution? Do we need a new prompt for every one ?

2- If the file is not downloaded by the prompt or manually by the user, should other triggers (e.g. the `is` command) re-prompt the user or should we treat user refusal as final, given that the manual command is always available in case user changes their mind ?

3- Right now, for simplicity, the convenience prompt generates a new temp file on every download and deletes it on exit, should we make any attempt to build a hash-keyed persistent local database ?

4- Instead of downloading to the filesystem for simplicity, maybe it's better to download to memory instead.

Almost all of those problems/nitpicks are around the triggers and UX of downloading, downloading itself once the decision is made is straightforward enough and uses existing code for the most part.

### DIFF
--- a/librz/io/p/io_gdb.c
+++ b/librz/io/p/io_gdb.c
@@ -15,6 +15,9 @@
 
 #define RZ_GDB_MAGIC rz_str_djb2_hash("gdb")
 
+#define MIN_DOWNLOAD_CHUNK 64
+#define MAX_DOWNLOAD_CHUNK 65536
+
 static int __close(RzIODesc *fd);
 
 static bool __plugin_open(RzIO *io, const char *file, bool many) {
@@ -209,19 +212,20 @@ extern int read_packet(libgdbr_t *instance, bool vcont);
 
 static bool gdbr_download_file(libgdbr_t *g, const char *remote, const char *local) {
 	rz_return_val_if_fail(g && remote && local, false);
-
+	// empty the destination file by writing nothing in non-append mode
 	if (!rz_file_dump(local, NULL, 0, false)) {
 		return false;
 	}
-
+	// open the remote file on the stub side
 	if (gdbr_open_file(g, remote, O_RDONLY, 0) < 0) {
 		return false;
 	}
 
 	uint32_t off = 0;
 	bool ok = true;
-	ut32 chunk = RZ_MAX(64, g->stub_features.pkt_sz / 2);
-	chunk = RZ_MIN(chunk, 65536);
+	// clamp the download size between a reasonable MIN floor and MAX ceiling
+	ut32 chunk = RZ_MAX(MIN_DOWNLOAD_CHUNK, g->stub_features.pkt_sz / 2);
+	chunk = RZ_MIN(chunk, MAX_DOWNLOAD_CHUNK);
 	ut8 *buf = malloc(chunk);
 	if (!buf) {
 		gdbr_close_file(g);
@@ -249,9 +253,8 @@ static bool gdbr_download_file(libgdbr_t *g, const char *remote, const char *loc
 	return ok;
 }
 
-RZ_API bool rz_io_gdb_download_file(RzIODesc *fd, const char *remote, const char *local) {
-	rz_return_val_if_fail(fd && remote && local, false);
-	if (!fd->plugin || strcmp(fd->plugin->name, "gdb")) {
+static bool io_gdb_download_file(RzIO *io, RzIODesc *fd, const char *remote, const char *local) {
+	if (!fd || !fd->data) {
 		return false;
 	}
 	return gdbr_download_file(fd->data, remote, local);
@@ -492,6 +495,7 @@ RzIOPlugin rz_io_plugin_gdb = {
 	.system = __system,
 	.getpid = __getpid,
 	.gettid = __gettid,
+	.download_file = io_gdb_download_file,
 	.isdbg = true
 };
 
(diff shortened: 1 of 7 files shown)


librz/io/p/io_gdb.c
```diff
@@ -207,6 +207,56 @@ static int __gettid(RzIODesc *fd) {
...
+	}
+
+	if (gdbr_open_file(g, remote, O_RDONLY, 0) < 0) {
+		return false;
+	}
+
+	uint32_t off = 0;
+	bool ok = true;
+	ut32 chunk = RZ_MAX(64, g->stub_features.pkt_sz / 2);
+	chunk = RZ_MIN(chunk, 65536);
```
wargio: `65536` should go into a `#define`

notxvilka (changes requested):
Ideally, would be great to test it with our mock gdbserver Python script like we do already for some other stuff.

moste00:
@notxvilka Hey, added a test and did some neat-ing up of the code.

One slighlty hack-ish thing I had to do is adding a `download_file` callback to an IO plugin, only GDB implements it for now. This is so `librz/main/rizin.c` can download the file after user prompt without resorting to abusing the `system` callback.

If it's a deal breaker then if we remove it, we can't have the prompt-user-to-download feature, I personally like this feature though, it saves an extra `R! download_file` command (which a new user might not even know exists).

moste00:
I know I need to document this in the rizin book and I already located the section where I need to edit, I'm just waiting for this PR to be approved so that the feature behaviour is fixed and then I can document it.

moste00:
@wargio @notxvilka ping.

wargio (approved):

notxvilka (approved):
Sorry, missed the notification. `librz/main/rizin.c` should be refactored but that's definitely out of the scope of this PR.

Issue #6543, opened 2026-06-20
Rizin for AI/LLM use (by the models themselves)


notxvilka:
### A1 — Structured JSON error envelope + status propagation to scripts/pipe  **(P0, M)**
**Problem.** There is an `RzCmdStatus` enum (`OK/ERROR/INVALID/WRONG/NONEXISTINGCMD/EXIT`) but **no
JSON error output** anywhere in the command layer, and a failing command is largely invisible to a
pipe consumer (empty stdout, maybe stderr noise). An agent driving Rizin over `rzpipe` or `rizin -c`
cannot reliably tell success from failure.
**Action.** In JSON output mode, emit a consistent envelope on non-OK status, e.g.
`{"error":{"code":"WRONG_ARGS","message":"...","command":"afl"}}`. Expose the last command's
`RzCmdStatus` to `rzpipe` (a way to read status alongside output), and make `rizin -c`/batch return a
non-zero process exit when a command errors.
**Where.** `librz/core/cmd_api.c` (RzCmd dispatch), `librz/cons` + `librz/socket/rzpipe.c`, batch
entry in `binrz/rizin`.
**Done when.** A failing command in `-j`/pipe context yields a parseable error object; rzpipe exposes
the status; CI test asserts non-zero exit on a known-bad command.

---

### B1 — Audit and close JSON-mode gaps on data-producing commands  **(P0, M)**
**Problem.** Only **315 / 1451 leaf commands (~21%)** declare `RZ_OUTPUT_MODE_JSON`. Many commands
legitimately produce no listable output (seek, config, write, navigation), but a substantial set of
*data* commands still have no `j` mode, forcing agents to scrape human text.
**Action.** Audit the data-producing subset (listings, `i*` info, analysis dumps, disassembly,
xrefs, types, flags, search) and add `RZ_OUTPUT_MODE_JSON` where missing. Add a CI lint that fails
when a *new* data command is added without a JSON mode.
**Where.** `librz/core/cmd_descs/*.yaml` + the corresponding handlers.
**Done when.** Every command that emits a list/record/table offers JSON; the lint is in CI.

### B2 — Publish versioned JSON Schemas for the core commands  **(P0, M)**
**Problem.** JSON shapes are undocumented and unversioned; agents (and downstream tools) can't rely
on field names/types, and a refactor can break consumers silently.
**Action.** Author/maintain JSON Schema for the top ~30–50 high-value commands (functions list,
[...]

PR #6597, opened 2026-07-04
Implemented a command catalog exporting as a json

moste00:

**Detailed description**

This implements suggestion C1 of https://github.com/rizinorg/rizin/issues/6543, namely, it exports the full-command tree as a json tree. The export can be triggered by both a command line option `--cmd-catalog json`, and via an internal help statement like `?*`, named `?+j`.

**Test plan**

For now, none. I plan to add it soon. The following will be implemented:

(1) A test that asserts the top-level number of children is the same as the number of commands outputed by `?`

(2) Manual comparison of the output against the yaml description of the commands, by both humans and several LLMs

**Closing issues**

None, small part of addressing https://github.com/rizinorg/rizin/issues/6543.


moste00:
@notxvilka please review ?

moste00:
@notxvilka ping

librz/core/cmd/cmd_api.c
```diff
@@ -1519,9 +1515,15 @@ RZ_API bool rz_cmd_get_help_json(RzCmd *cmd, const RzCmdDesc *cd, PJ *j) {
 		CASE_CDTYPE(RZ_CMD_DESC_TYPE_ARGV_STATE, "argv_state");
 #undef CASE_CDTYPE
 	default:
-		break;
+		return "unknown";
 	}
-	pj_ks(j, "type", type);
+}
+
+RZ_API bool rz_cmd_get_help_json(RzCmd *cmd, const RzCmdDesc *cd, PJ *j) {
```
notxvilka: Missing Doxygen, also add `RZ_NONNULL` where needed

librz/core/cmd/cmd_api.c
```diff
@@ -1539,6 +1541,127 @@ RZ_API bool rz_cmd_get_help_json(RzCmd *cmd, const RzCmdDesc *cd, PJ *j) {
...
+	}
+}
+
+/**
+ * \brief Visit every real command descriptor in a subtree.
+ *
+ * The pre callback is called before visiting children. The post callback is
+ * called after all children have been visited. Traversal starts at the root descriptor.
+ */
+RZ_API void rz_cmd_desc_foreach_tree(RzCmd *cmd, RzCmdDescVisitCb pre, RzCmdDescVisitCb post, void *user) {
```
notxvilka: NONNULL where appropriate

librz/core/cmd/cmd_api.c
```diff
@@ -1539,6 +1541,127 @@ RZ_API bool rz_cmd_get_help_json(RzCmd *cmd, const RzCmdDesc *cd, PJ *j) {
...
+	cmd_desc_foreach_tree(cmd, rz_cmd_get_root(cmd), pre, post, user);
+}
+
+/**
+ * \brief Visit every real command descriptor in a subtree.
+ *
+ * The pre callback is called before visiting children. The post callback is
+ * called after all children have been visited. Traversal starts at \p begin.
+ */
+RZ_API void rz_cmd_desc_foreach_tree_from(RzCmd *cmd, RzCmdDesc *begin, RzCmdDescVisitCb pre, RzCmdDescVisitCb post, void *user) {
```
notxvilka: And here

librz/core/cmd/cmd_api.c
```diff
@@ -1539,6 +1541,127 @@ RZ_API bool rz_cmd_get_help_json(RzCmd *cmd, const RzCmdDesc *cd, PJ *j) {
...
+	CmdTreeJson *ctx = user;
+	PJ *j = ctx->j;
+	pj_end(j);
+	pj_end(j);
+}
+
+/**
+ * \brief Generates a recursive JSON representation of the real command descriptor tree.
+ */
+RZ_API bool rz_cmd_get_tree_json(RzCmd *cmd, PJ *j) {
```
notxvilka: NONNULL

librz/core/cmd/cmd_api.c
```diff
@@ -1539,6 +1541,127 @@ RZ_API bool rz_cmd_get_help_json(RzCmd *cmd, const RzCmdDesc *cd, PJ *j) {
...
+	rz_return_val_if_fail(cmd && j, false);
+	RzStrBuf args;
+	rz_strbuf_init(&args);
+	CmdTreeJson ctx = {
+		.j = j,
+		.args = &args,
+	};
+	pj_o(j);
+	pj_ki(j, "version", 1);
```
notxvilka: Maybe make a define?

librz/main/rizin.c
```diff
@@ -513,6 +515,38 @@ RZ_API int rz_main_rizin(int argc, const char **argv) {
 	char *debugbackend = rz_str_dup("native");
 
 	RzGetopt opt;
+	for (int i = 1; i < argc; i++) {
```
notxvilka: This probably can be extracted into a small static function

librz/main/rizin.c
```diff
@@ -513,6 +515,38 @@ RZ_API int rz_main_rizin(int argc, const char **argv) {
...
+					RZ_LOG_ERROR("Unsupported command catalog format '%s'\n", argv[i + 1]);
+					ret = 1;
+					goto beach;
+				}
+				remove_next_arg = true;
+			}
+		} else if (!strcmp(argv[i], "--cmd-catalog=json")) {
+			do_print_cmd_tree_json = true;
+			remove_arg = true;
+		} else if (rz_str_startswith(argv[i], "--cmd-catalog=")) {
```
notxvilka: This solution looks hackish a bit, tbh, we have custom getopt, we should use it somehow

notxvilka (changes requested):

moste00:
@notxvilka I did most of the suggestions, for the objection about parsing of the `--cmd-catalog` flag, I discovered that Rizin has no support for long options to be begin with. Is that true or did I miss something ?

So I added this long form flag support for now, I made sure it doesn't affect any existing logic for option parsing. It's bit more complex than short options, for one we don't use a string to describe it, we use a structured array of structured data.

If it's too irrleveant for this PR, our options are:

1- Define it in a seperate PR and make this PR based on it.

2- Remove the long form flag entirely and rely on Rizin's command execution functionality to execute `?+j` from the command line directly.

Other than that, all other comments are addressed.

notxvilka:
> @notxvilka I did most of the suggestions, for the objection about parsing of the `--cmd-catalog` flag, I discovered that Rizin has no support for long options to be begin with. Is that true or did I miss something ?

Yes, it doesn't. But maybe it's time to add it properly. @wargio @Rot127 @thestr4ng3r WDYT?

librz/include/rz_getopt.h
```diff
@@ -7,6 +7,15 @@
 extern "C" {
 #endif
 
+#define RZ_GETOPT_LONG_BASE 0x100
+
+typedef struct rz_getopt_long_t {
+	const char *name;
+	int val;
+	const char **values;
+	const char *default_value;
```
Rot127: Add doxygen please.

librz/include/rz_getopt.h
```diff
@@ -17,9 +26,11 @@ typedef struct rz_getopt_t {
 	int argc;
 	const char **argv;
 	const char *ostr;
+	const RzGetoptLong *longopts;
```
Rot127: Doxygen please.

librz/core/cmd/cmd.c
```diff
@@ -1447,7 +1447,10 @@ DEFINE_HANDLE_TS_FCN_AND_SYMBOL(redirect_stmt) {
 
 DEFINE_HANDLE_TS_FCN_AND_SYMBOL(help_stmt) {
 	size_t node_str_len = strlen(node_string);
-	if (node_str_len >= 2 && !strcmp(node_string + node_str_len - 2, "?*")) {
+	if (RZ_STR_EQ(node_string, "?+j")) {
+		const char *argv[1] = { node_string };
+		return rz_cmd_catalog_json_handler(state->core, 1, argv);
+	} else if (node_str_len >= 2 && !strcmp(node_string + node_str_len - 2, "?*")) {
```
Rot127: RZ_STR_EQ

librz/core/cmd/cmd_api.c
```diff
@@ -1,13 +1,16 @@
 // SPDX-FileCopyrightText: 2009-2020 pancake <pancake@nopcode.org>
 // SPDX-License-Identifier: LGPL-3.0-only
 
+#include "rz_types.h"
```
Rot127: <>

librz/core/cmd_descs/cmd_descs.yaml
```diff
@@ -160,6 +160,10 @@ commands:
     summary: Search help commands
     type: RZ_CMD_DESC_TYPE_GROUP
     subcommands: cmd_help
+  - name: "?+j"
```
Rot127, line 163: Please use the output modes here. Hardcoding it for json is going against the CMD design.

librz/include/rz_getopt.h
```diff
@@ -7,6 +7,15 @@
 extern "C" {
 #endif
 
+#define RZ_GETOPT_LONG_BASE 0x100
+
+typedef struct rz_getopt_long_t {
+	const char *name;
+	int val;
+	const char **values;
```
Rot127: Same here, make it a vector please.

librz/include/rz_getopt.h
Rot127: Also, is this an array of longopts? Why not a vector?

librz/util/getopt.c
```diff
@@ -14,7 +14,72 @@
 #define BADARG (int)':'
 #define EMSG   ""
 
-RZ_API void rz_getopt_init(RzGetopt *opt, int argc, const char **argv, const char *ostr) {
+static bool longopt_has_values(const RzGetoptLong *desc) {
+	return desc->values && desc->values[0];
+}
+
+static bool longopt_value_allowed(const RzGetoptLong *desc, const char *value) {
+	if (!value || !*value) {
```
Rot127, line 23: Assert? Or is this valid input?

librz/util/getopt.c
```diff
@@ -14,7 +14,72 @@
...
+		return false;
+	}
+	for (size_t i = 0; desc->values[i]; i++) {
+		if (!strcmp(desc->values[i], value)) {
```
Rot127: RZ_STR_EQ

librz/util/getopt.c
```diff
@@ -14,7 +14,72 @@
...
+	}
+	return false;
+}
+
+static int rz_getopt_long_next(RzGetopt *opt, const char *arg) {
+	const char *name = arg + 2;
+	const RzGetoptLong *desc = NULL;
+
+	for (const RzGetoptLong *it = opt->longopts; it && it->name; it++) {
+		if (!strcmp(name, it->name)) {
```
Rot127: RZ_STR_EQ

librz/util/getopt.c
```diff
@@ -14,7 +14,72 @@
...
+			desc = it;
+			break;
+		}
+	}
+	if (!desc) {
+		opt->opt = '-';
+		opt->ind++;
+		if (opt->err && *opt->ostr != ':') {
+			eprintf("%s: illegal option -- %s\n", opt->argv[0], name);
```
Rot127: RZ_LOG_ERROR

librz/util/getopt.c
```diff
@@ -14,7 +14,72 @@
...
+		consume_value = true;
+	} else {
+		value = desc->default_value;
+	}
+
+	if (!longopt_value_allowed(desc, value)) {
+		opt->opt = '-';
+		opt->ind += consume_value ? 2 : 1;
+		if (opt->err && *opt->ostr != ':') {
+			eprintf("%s: invalid argument '%s' for option -- %s\n", opt->argv[0], value ? value : "", desc->name);
```
Rot127: RZ_LOG_ERROR?

librz/util/getopt.c
```diff
@@ -36,11 +106,19 @@ RZ_API int rz_getopt_next(RzGetopt *opt) {
 			place = EMSG;
 			return -1;
 		}
-		if (place[1] && *++place == '-') { // found "--"
+		if (place[1] && place[1] == '-') {
```
Rot127: rz_str_starts_wth

Rot127 (changes requested):
The help doesn't work for me:

> ?+?
ERROR: core: Error while executing command: ?+?

Rot127:
> Yes, it doesn't. But maybe it's time to add it properly

Adding it would be nice. But it is a bigger task for sure.
Because the parsing is way way more complex.

librz/core/cmd_descs/cmd_descs.yaml
moste00: Is there a reasonable non-json encoding though ? What would the tree look like in plain text (indentation?) ?

moste00:
> The help doesn't work for me:
> 
> > ?+?
> > ERROR: core: Error while executing command: ?+?

I think you accidentally invoked your suggested renaming of the command, it's `?+j` not `?+` (for now).

image

moste00:
> > Yes, it doesn't. But maybe it's time to add it properly
> 
> Adding it would be nice. But it is a bigger task for sure. Because the parsing is way way more complex.

I'm open to refactoring the parsing of long options into another PR after implementing your comments, but for now the logic is pretty simple and runs on  200-250 lines, it goes something like this:

1- Every long options is described by a descriptor struct, the options initialization function is extended to accept a descriptor array for describing long options (the old semantics of no long options still exists as a thin wrapper)

2- Every long option either has a valid list of values or is just a flag (no values, its presence or absence is itself a boolean value), those that do have a list of values can have also a default value

3- When a long option is given as `--option value`, both strings are consumed, if it's given as `--option -o` or `--option --another-option` (followed by another short or long option immediately), then only `--option` is consumed and it's given a default value if it has a default value or becomes an error if it doesn't.

4- When a long option is given an invalid value `--option value-that-doesnt-appear-on-the-valid-list`, both strings are consumed and it's considered a parsing error (even if the option has a default value, the provided value is still consumed and considered as its invalid value)

5- The function that consumes short options is the same that consumes long options, since it returns an `int`, it can be (ab)used to return a big (> 256) enum value for long options, making sure it can always unamibigously indicate a long option, you can then add `case:`s for those enum values in any switch statement you have around it.

That's it. I think it's pretty simple as far as designs go, any comment or objects on it ? is it too much to implement for a single option for now and should be postponed ?

librz/include/rz_getopt.h
moste00: The rest of the parser uses old school C style of arrays, like `argv`, so to maintain consistency I used the same style, but I changed it to vectors now.

librz/util/getopt.c
moste00: Can't be an assert, this would make rizin crash when given no option value for a required option, which is worse UX than just detecting and logging it with a readable diagnosis.

librz/util/getopt.c
moste00: This will make the code inconsistent IMHO, whole function uses pointer gymnastics extensively and heavily relies on that `place` cursor, rewriting it seems dangerous, and introducing `rz_str_starts_with` for just this one small part is annoyingly inconsistent.

You still want to proceed with this ?

librz/core/cmd_descs/cmd_descs.yaml
Rot127: It doesn't need to be plain text. Maybe we add some XML output later, or csv or whatever. Then this command has to be refactored pretty heavily.
The command line design was intentional made extensible. If we start to add exceptions it makes it unmaintainable.

librz/util/getopt.c
```diff
@@ -32,15 +105,31 @@ RZ_API int rz_getopt_next(RzGetopt *opt) {
...
+					return ret;
+				}
+				// otherwise an error
+				opt->ind++;
+				place = EMSG;
+				return -1;
+			}
+			place++;
+		}
 	}
```
Rot127, line 133: Also here, please use the rz_str functions

librz/util/getopt.c
Rot127: The point is, that all the rz_str functions have additional checks. NULL checks, early return if pointer matches, and properly tested, and have descriptive names etc.

Using them here means following better software design and defensive programming:
1. You reuse the code which was already written and better tested.
2. Use the functions which include asserts.
3. Using these functions makes the code more readable by being more descriptive, all with less LOC.

That increases the code quality on average.

The rest of `getopt.c` is super old and doesn't do any of these things.
We shouldn't orient ourselves at it. If consistency means being consistent with lower standards, we don't want to be consistent with it.

librz/core/cmd_descs/cmd_descs.yaml
Rot127: It also breaks the user assumption how commands work. Which is `<cmd><output mode>`.

Rot127 (changes requested):

moste00:
@Rot127 I implemented your suggestions, also refactored `place` to be non-static by lifting to the `RzGetOpt` struct (it already contains a mutable cursor-like variable, the `ind` field).

I think this is about as far as we can get without a wholesale refactor of argument parsing, which I can also do in a seperate PR if you think it's worth it.

Also refactored to output modes instead of hardcoded `j`.

librz/core/cmd/cmd_api.c
```diff
@@ -1519,9 +1518,22 @@ RZ_API bool rz_cmd_get_help_json(RzCmd *cmd, const RzCmdDesc *cd, PJ *j) {
 		CASE_CDTYPE(RZ_CMD_DESC_TYPE_ARGV_STATE, "argv_state");
 #undef CASE_CDTYPE
 	default:
-		break;
+		return "unknown";
 	}
-	pj_ks(j, "type", type);
+}
+/**
```
Rot127, line 1526: Missing newline

librz/core/cmd/cmd_api.c
```diff
@@ -1519,9 +1518,22 @@ RZ_API bool rz_cmd_get_help_json(RzCmd *cmd, const RzCmdDesc *cd, PJ *j) {
...
+* \brief Generates a JSON tree describing the command tree
+* \param cmd Reference to RzCmd
+* \param cd Reference to RzCmdDesc
+* \param j Reference to the PJ builder helper object
+* 
+* \return returns false if any argument is null, otherwise it builds in PJ and returns true.
+*/
+RZ_API bool rz_cmd_get_help_json(RZ_NONNULL RzCmd *cmd, RZ_NONNULL const RzCmdDesc *cd, RZ_NONNULL PJ *j) {
```
Rot127:
```suggestion
RZ_API bool rz_cmd_get_help_json(RZ_NONNULL RzCmd *cmd, RZ_NONNULL const RzCmdDesc *cd, RZ_NONNULL RZ_OUT PJ *j) {
```

librz/core/cmd/cmd_api.c
```diff
@@ -1519,9 +1518,22 @@ RZ_API bool rz_cmd_get_help_json(RzCmd *cmd, const RzCmdDesc *cd, PJ *j) {
...
-		break;
+		return "unknown";
 	}
-	pj_ks(j, "type", type);
+}
+/**
+* \brief Generates a JSON tree describing the command tree
+* \param cmd Reference to RzCmd
+* \param cd Reference to RzCmdDesc
+* \param j Reference to the PJ builder helper object
```
Rot127:
```suggestion
* \param j PJ builder to write the JSON into.
```

librz/core/cmd/cmd_api.c
```diff
@@ -1519,9 +1518,22 @@ RZ_API bool rz_cmd_get_help_json(RzCmd *cmd, const RzCmdDesc *cd, PJ *j) {
...
 	default:
-		break;
+		return "unknown";
 	}
-	pj_ks(j, "type", type);
+}
+/**
+* \brief Generates a JSON tree describing the command tree
+* \param cmd Reference to RzCmd
+* \param cd Reference to RzCmdDesc
```
Rot127:
```suggestion
* \param cmd The RzCmd instance to get the comments from.
* \param cd ???
```

That they are references is clear from the header.

librz/core/cmd/cmd_api.c
```diff
@@ -1539,6 +1551,127 @@ RZ_API bool rz_cmd_get_help_json(RzCmd *cmd, const RzCmdDesc *cd, PJ *j) {
...
+		cmd_desc_foreach_tree(cmd, child, pre, post, user);
+	}
+
+	if (post) {
+		post(cmd, cd, user);
+	}
+}
+
+/**
+ * \brief Visit every real command descriptor in a subtree.
```
Rot127, line 1585: What is an unreal descriptor?

librz/core/cmd/cmd_api.c
```diff
@@ -1539,6 +1551,127 @@ RZ_API bool rz_cmd_get_help_json(RzCmd *cmd, const RzCmdDesc *cd, PJ *j) {
...
+ *
+ * The pre callback is called before visiting children. The post callback is
+ * called after all children have been visited. Traversal starts at the root descriptor.
+ */
+RZ_API void rz_cmd_desc_foreach_tree(RZ_NONNULL RzCmd *cmd, RzCmdDescVisitCb pre, RzCmdDescVisitCb post, void *user) {
```
Rot127: Missing doxygen for parameters.

librz/core/cmd/cmd_api.c
```diff
@@ -1539,6 +1551,127 @@ RZ_API bool rz_cmd_get_help_json(RzCmd *cmd, const RzCmdDesc *cd, PJ *j) {
...
+	cmd_desc_foreach_tree(cmd, rz_cmd_get_root(cmd), pre, post, user);
+}
+
+/**
+ * \brief Visit every real command descriptor in a subtree.
+ *
+ * The pre callback is called before visiting children. The post callback is
+ * called after all children have been visited. Traversal starts at \p begin.
+ */
+RZ_API void rz_cmd_desc_foreach_tree_from(RZ_NONNULL RzCmd *cmd, RZ_NONNULL RzCmdDesc *begin, RzCmdDescVisitCb pre, RzCmdDescVisitCb post, void *user) {
```
Rot127: Missing doxygen for parameters.

librz/core/cmd/cmd_api.c
```diff
@@ -1539,6 +1551,127 @@ RZ_API bool rz_cmd_get_help_json(RzCmd *cmd, const RzCmdDesc *cd, PJ *j) {
...
+	rz_return_if_fail(cmd && begin);
+	cmd_desc_foreach_tree(cmd, begin, pre, post, user);
+}
+
+static void cmd_tree_json_modes(PJ *j, const RzCmdDesc *cd) {
+	RzCmdDesc *exec_cd = rz_cmd_desc_get_exec((RzCmdDesc *)cd);
+	pj_ka(j, "modes");
+	if (exec_cd) {
```
Rot127: Invert and return early

librz/core/cmd/cmd_api.c
```diff
@@ -1539,6 +1551,127 @@ RZ_API bool rz_cmd_get_help_json(RzCmd *cmd, const RzCmdDesc *cd, PJ *j) {
...
+	rz_return_val_if_fail(cmd && j, false);
+	RzStrBuf args;
+	rz_strbuf_init(&args);
+	CmdTreeJson ctx = {
+		.j = j,
+		.args = &args,
+	};
+	pj_o(j);
+	pj_ki(j, "version", CMD_EXPORT_VERSION);
+	pj_ks(j, "generated_from", "runtime");
```
Rot127: What other source do you have in mind?

librz/include/rz_getopt.h
```diff
@@ -1,25 +1,39 @@
...
 #endif
 
+// 256, all long options values are >= RZ_GETOPT_LONG_BASE so that they can be distinguished from single-char short options
+#define RZ_GETOPT_LONG_BASE 0x100
+
+typedef struct rz_getopt_long_t {
+	const char *name; ///< Name of the long option. e.g. --name
+	int val; ///< enum value > RZ_GETOPT_LONG_BASE returned by rz_getopt_next when the long option is consumed
+	const RzPVector /*<const char *>*/ *values; ///< All possible valid cmdline values for the long option. If NULL, long option is a yes/no flag only.
+	const char *default_value; ///< Default of the long option when it's given no cmdline value, nullable
```
Rot127:
```suggestion
	RZ_NULLABLE const char *default_value; ///< Default of the long option when it's given no cmdline value.
```

librz/util/getopt.c
```diff
@@ -8,42 +8,132 @@
  * $Id: getopt.c,v 1.2 1998/01/21 22:27:05 billm Exp $ *
  */
 
+#include "rz_util/rz_log.h"
+#include "rz_util/rz_str.h"
```
Rot127:
```suggestion
#include <rz_util/rz_log.h>
#include <rz_util/rz_str.h>
```

librz/util/getopt.c
```diff
@@ -8,42 +8,132 @@
...
+		}
+		return BADCH;
+	}
+
+	opt->arg = value;
+	opt->ind += consume_value ? 2 : 1;
+	return desc->val;
+}
+
+RZ_API void rz_getopt_init_long(RzGetopt *opt, int argc, const char **argv, const char *ostr, const RzVector *longopts) {
```
Rot127: doxygen missing

librz/util/getopt.c
```diff
@@ -8,42 +8,132 @@
...
 	opt->reset = 0;
 	opt->arg = NULL;
+	opt->place = EMSG;
 	opt->argc = argc;
 	opt->argv = argv;
 	opt->ostr = ostr;
+	opt->longopts = longopts;
+}
+
+RZ_API void rz_getopt_init(RzGetopt *opt, int argc, const char **argv, const char *ostr) {
```
Rot127, line 115: Please add doxygen

Rot127 (changes requested):
A few more nitpicks.
But generally looks good

librz/core/cmd/cmd_api.c
moste00: I think that's just a slightly awkward phrasing of the intent of avoiding "virtual" nodes in the tree like help groupings.

image

Looking at the code, it does NOT actually do that, and it does print "fake" commands like the `>` redirection operator. I think that's more correct since it's more complete and a full reflection of the real command tree as it's in memory. 

So I will remove this phrasing, it's something either I intended to implement but never got around to implementing or implemented then deleted, I can't remember which.

librz/core/cmd/cmd_api.c
moste00: Sorry, that's a leftover from the first MVP by GPT.

The field value is alluding to the actual approach that actually got implemented, the other one I considered was ingesting the yaml description of the tree at build time by a python script and generating dumb, tree-strucutred C that is specialized to printing a single command tree shape. This approach was rejected because (1) generated code is frowned upon/a hassle in the project (2) The tree is dynamic via (at least) user-defined aliases (3) The dynamic approach adds a useful control structure `foreach` that can be used for other tree operations not related to printing.

I don't see any usefulness for it, I will delete it unless you think it's useful for something in the future.

librz/util/getopt.c
```diff
@@ -8,42 +8,147 @@
...
+			RZ_LOG_ERROR("%s: invalid argument '%s' for option -- %s\n", opt->argv[0], value ? value : "", desc->name);
+		}
+		return BADCH;
+	}
+
+	opt->arg = value;
+	opt->ind += consume_value ? 2 : 1;
+	return desc->val;
+}
+/**
```
Rot127, line 85:
```suggestion

/**
```

librz/util/getopt.c
```diff
@@ -8,42 +8,147 @@
...
 	opt->arg = NULL;
+	opt->place = EMSG;
 	opt->argc = argc;
 	opt->argv = argv;
 	opt->ostr = ostr;
+	opt->longopts = longopts;
+}
+
+/**
+ * A variant of rz_getopt_init_long that hardcodes long options to nothing
```
Rot127, line 108: What does this mean?

librz/util/getopt.c
```diff
@@ -8,42 +8,147 @@
...
+ * \param argc the cmdline count passed to the program
+ * \param argv the cmdline string passed to the program
+ * \param ostr the short option configuration string describing all valid short options
+ * \param longopts the long option configuration describing all valid long option
+ */
+RZ_API void rz_getopt_init(RzGetopt *opt, int argc, const char **argv, const char *ostr) {
+	rz_getopt_init_long(opt, argc, argv, ostr, NULL);
 }
 
 RZ_API int rz_getopt_next(RzGetopt *opt) {
```
Rot127, line 119: Please document this as well, while you are at it.

Rot127 (changes requested):
Last nitpicks. Otherwise looks good to me.

librz/util/getopt.c
moste00: The previous option-initialization function was called `rz_getopt_init` but it didn't take the long options description struct, so I had to the make `_long` variant that does take the long options description struct (or NULL if no long options exist), but I didn't want to change all calllers of it, so I kept the old one but made it a thin wrapper around the `_long` variant.

Rot127: I see. Please document the function independently though. Without reference to `rz_getopt_init_long`

Rot127 (approved):

PR #6307, opened 2026-05-03
Mark bytes as data and show as such in disassembly if referenced as DATA

bubblepipe:
**Your checklist for this pull request**
- [x] I've read the [guidelines for contributing](https://github.com/rizinorg/rizin/blob/dev/CONTRIBUTING.md) to this repository.
- [x] I made sure to follow the project's [coding style](https://github.com/rizinorg/rizin/blob/dev/DEVELOPERS.md#code-style).
- [x] I've documented every `RZ_API` function and struct this PR changes.
- [x] I've added tests that prove my changes are effective (required for changes to `RZ_API`).
- [x] I've updated the [Rizin book](https://github.com/rizinorg/book) with the relevant information (if needed).
- [x] I've used AI tools to generate fully or partially these code changes and I'm sure the changes are not copyrighted by somebody else.

**Detailed description**

I introcued an analysis pass to properly detect and mark `DATA`. 

**Test plan**

I have introduced tests for: 
- `bins/elf/libarm64.so`
- `bins/elf/analysis/guess-number-riscv64`
- `bins/elf/analysis/mipsbe-ip`

**Closing issues**

https://github.com/rizinorg/rizin/issues/3641

PR #6612, opened 2026-07-08
librz/core/canalysis.c: Fix data xref analysis

MrQuantum1915:

**Detailed description**

while processing for marking xrefs as data
1. classify target using `xref_ref_kind` for data section too, previously it was only classified if target was in exec segment. Which caused false positive when the target was in non-exec section. Happens when the immediate value is small and it points in data section.

3. restrict data block from bleeding into other sections. Currently it correctly caps data block  at next "detected" function (or next data) but when the function is not detected yet (like in stripped bins) and the area onwards from data ref is empty, the data block bleeds into other sections specifically executable section. This should never happen. 

while experimenting on my fix i also noticed that this fails when the target is in a section gap (for any reason :D) and it fallback to caping on map_end.
so we need to iterate over sections if any , to cap upper at next section.

fallback to map end (when target is in last section before unmapped region)

Relevant PRs: 
1st bug : https://github.com/rizinorg/rizin/pull/6506
2nd bug : https://github.com/rizinorg/rizin/pull/6307

...

**Test plan**

- **For 1st bug**
before:
```bash
❯ rizin test/bins/elf/hello-freebsd-x64
 -- Use the 'id' command to see the source line related to the current seek
[0x002015a0]> e io.cache=true
[0x002015a0]> s 0x2015ae
[0x002015ae]> wa "and dword [rbp-8], 0x0020001f"
[0x002015ae]> aaa
...
[0x002015ae]> Cdl~0x0020001f
0x0020001f data Cd 5505
[0x002015ae]> 
```

after:
```bash
❯ rizin test/bins/elf/hello-freebsd-x64
 -- Temporally drop the verbosity prefixing the commands with ':'
[0x002015a0]> e io.cache=true
[0x002015a0]> s 0x2015ae
[0x002015ae]> wa "and dword [rbp-8], 0x0020001f"
[0x002015ae]> aaa
...
[0x002015ae]> Cdl~0x0020001f
[0x002015ae]>
```

- **For 2nd bug i have already mentioned outputs on dev branch in test itself, checkout (those lines which says it should not be....). On dev that are the outputs and hence this test fails on dev**
...

**Closing issues**

...

### DIFF
--- a/librz/core/canalysis.c
+++ b/librz/core/canalysis.c
@@ -4004,11 +4004,6 @@ static bool addr_in_exec_section(RzBinObject *bo, ut64 addr) {
 	return sec && (sec->perm & RZ_PERM_X);
 }
 
-static bool addr_in_exec_segment(RzBinObject *bo, ut64 addr) {
-	RzBinSection *seg = rz_bin_get_segment_at(bo, addr, true);
-	return seg && (seg->perm & RZ_PERM_X);
-}
-
 /** \brief How a DATA xref's instruction relates to its target address. */
 typedef enum {
 	XREF_REF_OTHER = 0, ///< address computation, control flow, etc. — handled as before
@@ -4106,11 +4101,9 @@ static void analysis_mark_xrefs_as_data(RzCore *core) {
 		if (!bo) {
 			continue;
 		}
-		// Only executable-region targets need classifying: reject constants that merely
-		// equal a code address (e.g. `sub sp, sp, 0x810`), while skipping the decode for
-		// the common data-section case.
-		bool in_exec_segment = addr_in_exec_segment(bo, target);
-		XrefRefKind ref_kind = in_exec_segment ? xref_ref_kind(core, xref->from, target) : XREF_REF_OTHER;
+
+		// reject constants that merely equal a code address (e.g. `sub sp, sp, 0x810`)
+		XrefRefKind ref_kind = xref_ref_kind(core, xref->from, target);
 		if (ref_kind == XREF_REF_COINCIDENTAL_IMM) {
 			continue;
 		}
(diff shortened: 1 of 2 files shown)


librz/core/canalysis.c
```diff
@@ -4202,6 +4197,44 @@ static void analysis_mark_xrefs_as_data(RzCore *core) {
...
+			rz_pvector_foreach (bo->sections, iter) {
+				s = *iter;
+				if (s->is_segment) {
+					continue;
+				}
+				ut64 vaddr = rz_bin_object_addr_with_base(bo, s->vaddr);
+				if (vaddr > target && vaddr < next_sec_start) {
+					next_sec_start = vaddr;
+				}
+			}
```
MrQuantum1915, line 4222: wondering if we should create api fxn for `rz_bin_get_next_section` or not

Rot127:
Can you give some context why this is needed please?

MrQuantum1915:
> Can you give some context why this is needed please?

i discussed this in mattermost rizin dev channel
its a bug that prevented `aap` (preludes analysis) from creating a function at proper place

Rot127:
Sorry, just saw the message.

Rot127 (approved):

Rot127:
Please rebase

Rot127:
@notxvilka Please take a look

Rot127:
Please rebase

Rot127 (approved):
Redness is unrelated

librz/core/canalysis.c
notxvilka: Try to search for similar patterns in the code - if there is more than 3 - likely.

notxvilka (approved):

librz/core/canalysis.c
MrQuantum1915: i checked; it seems to be used only here, so yeah adding that API is not worth it maybe

Issue #1461, opened 2021-08-16
Add Trans theory and rounding modes into RZIL


XVilka:
Currently only supported theories are:
- Bool
- Bitvector
- Mem (Array)
- Effect

The biggest missing chunk is the Float theory

- [x] Implement IEE754 types and helper APIs in RzUtil
- [x] Implement Fbasic operations https://github.com/rizinorg/rizin/pull/3184
- [x] Cover these by unit tests
- [ ] Add rounding modes (Rmode) operations
- [ ] Add Trans (transcendental) operations

See:
- http://binaryanalysisplatform.github.io/bap/api/odoc/bap-core-theory/Bap_core_theory/Theory/Float/index.html
- http://binaryanalysisplatform.github.io/bap/api/odoc/bap-core-theory/Bap_core_theory/Theory/Rmode/index.html
- http://binaryanalysisplatform.github.io/bap/api/odoc/bap-core-theory/Bap_core_theory/Theory/module-type-Fbasic/index.html
- http://binaryanalysisplatform.github.io/bap/api/odoc/bap-core-theory/Bap_core_theory/Theory/module-type-Float/index.html
- http://binaryanalysisplatform.github.io/bap/api/odoc/bap-core-theory/Bap_core_theory/Theory/module-type-Trans/index.html
- http://binaryanalysisplatform.github.io/bap/api/odoc/bap-core-theory/Bap_core_theory/Theory/IEEE754/index.html

Implementation is at https://github.com/BinaryAnalysisPlatform/bap/blob/master/lib/bap_core_theory/bap_core_theory_definition.ml

Note, I think that **Trans** theory should be separate from the **Float** in the code.

http://binaryanalysisplatform.github.io/bap/api/odoc/bap-core-theory/Bap_core_theory/Theory/module-type-Core/index.html

XVilka:
Better description of the rounding modes is here: http://binaryanalysisplatform.github.io/bap/api/master/bap-core-theory/Bap_core_theory/Theory/module-type-Float/index.html#rounding-modes

@brightprogrammer @Heersin @imbillow

Rot127:
The current float operations don't support dynamic rounding mode selection. They must be static. This needs refactoring, because most architectures select the rounding mode via some flags in a float status register.
(I assume this is what you meant with "Add rounding modes (Rmode) operations" as well?)

notxvilka:
>  (I assume this is what you meant with "Add rounding modes (Rmode) operations" as well?)

[...]

PR #6626, opened 2026-07-18
Improve RzIL floating-point support

b1llow:

- [x] I've read the guidelines for contributing (https://github.com/rizinorg/rizin/blob/dev/CONTRIBUTING.md) to this repository.
- [x] I made sure to follow the project's coding style (https://github.com/rizinorg/rizin/blob/dev/DEVELOPERS.md#code-style).
- [x] I've documented every RZ_API function and struct this PR changes.
- [x] I've added tests that prove my changes are effective (required for changes to RZ_API).
- [x] I've updated the Rizin book (https://github.com/rizinorg/book) with the relevant information (if needed).

**Detailed description**

This PR improves IEEE-754 edge-case handling in `RzFloat` and extends `RzIL` floating-point support.

The `RzFloat` changes fix or improve:

- binary80 explicit-integer-bit, pseudo-value, infinity, and NaN handling;
- binary16 conversions;
- gradual underflow and directed rounding;
- overflow, underflow, invalid-operation, and inexact exception reporting;
- exception propagation through nested conversions and arithmetic operations;
- binary80 fused multiply-add rounding, including reduced-precision and double-rounding edge cases;
- thread-local SoftFloat state, preventing rounding state from leaking between threads.

The `RzIL` changes add scoped binary80 precision support through `RzFloatRPrecision` and `FWITH_RPREC`. The supported precisions are 32, 64, and 80. Precision scopes restore the
previous thread-local SoftFloat state after successful evaluation and evaluation failures.

Runtime rounding modes are represented explicitly by dedicated pure opcodes:

- `FCONVERT_WITH_RMODE`
- `FROUND_WITH_RMODE`
- `FSQRT_WITH_RMODE`
- `FADD_WITH_RMODE`
- `FSUB_WITH_RMODE`
- `FMUL_WITH_RMODE`
- `FDIV_WITH_RMODE`
- `FMOD_WITH_RMODE`

Their rounding-mode operand is a 32-bit IL bitvector whose values correspond to `RzFloatRMode`: RNE, RNA, RTP, RTN, and RTZ. Invalid operand widths are rejected by validation,
while invalid runtime values cause evaluation to fail with an error.

Dedicated opcodes keep runtime-controlled floating-point expressions compact. This is useful for architectures whose rounding mode is selected from register state and avoids the
expression duplication caused by expanding every operation into nested `ITE` branches.

The new operations are supported by:

- construction, duplication, and destruction;
- type and operand validation;
- VM evaluation;
- plain, Unicode, and JSON exporters;
- graph output and opcode stringification.

`FEXCEPT` now emits a VM event only when the queried exception is present, while preserving exceptions raised by nested conversions and arithmetic operations.

**Test plan**

...

**Closing issues**

Partially addresses https://github.com/rizinorg/rizin/issues/1461


librz/util/float/float.c
```diff
@@ -1660,38 +1812,146 @@ RZ_API RZ_OWN RzFloat *rz_float_convert(RZ_NONNULL RzFloat *f, RzFloatFormat for
 	rz_return_val_if_fail(f, NULL);
 
 	if (rz_float_is_nan(f)) {
```
b1llow, line 1954: **[correctness] Converting a signaling NaN quiets it without raising `RZ_FLOAT_E_INVALID_OP`.**

`rz_float_is_nan()` matches both SNAN and QNAN, so an sNaN input takes this early return and produces a fresh quiet NaN with only the sign and accumulated `f->exception` preserved. IEEE 754 requires the invalid-operation exception when an sNaN is converted, and SoftFloat's conversion routines (used for every finite path added below) do raise it. Concretely: an IL `fconvert` of an sNaN f32→f64 now yields a qNaN with no invalid-op flag, diverging from the SoftFloat semantics the rest of this function relies on. Routing NaN inputs through the same SoftFloat conversion calls (which quiet sNaNs and set `softfloat_flag_invalid`) would make this special case unnecessary.

meson.build
```diff
@@ -760,8 +760,35 @@ if r.returncode() == 1 and get_option('subprojects_check')
   error(subproject_clean_error_msg)
 endif
 
-softfloat_dep = dependency('softfloat', required: get_option('use_sys_softfloat'), static: is_static_build)
-if not softfloat_dep.found()
+use_sys_softfloat = get_option('use_sys_softfloat')
+softfloat_dep = dependency('softfloat', required: use_sys_softfloat, static: is_static_build)
+use_bundled_softfloat = not softfloat_dep.found()
+if softfloat_dep.found()
+  softfloat_tls_probe = '''
```
b1llow, line 760: **[build] The probe validates hand-written `extern __thread` declarations, not the `<softfloat.h>` declarations rizin actually compiles against.**

The probe links its own ad-hoc declarations, so it proves the *library's symbols* are TLS — but rizin's `float.c` sees the state through `softfloat.h`, whose `#ifndef THREAD_LOCAL → empty` fallback yields plain (non-TLS) externs unless the installed headers also carry the platform patch. A distro that builds SoftFloat with TLS state but ships unpatched headers passes this probe, and then librz_util fails at ninja time with `TLS definition ... mismatches non-TLS reference` (a hard link error with GNU ld/lld), even with `use_sys_softfloat=auto` — the configure-time guard never fires. Having the probe `#include <softfloat.h>` through `softfloat_dep` and reference the state via the real declarations checks exactly the contract rizin depends on.

librz/util/float/float.c
```diff
@@ -1374,8 +1475,58 @@ RZ_API RZ_OWN RzFloat *rz_float_fma_ieee_bin(RZ_NONNULL RzFloat *a, RZ_NONNULL R
 		float128_t b_resized = extF80_to_f128(to_float80(b));
 		float128_t c_resized = extF80_to_f128(to_float80(c));
 
+		if (extF80_roundingPrecision != RZ_FLOAT_RPREC_32 && extF80_roundingPrecision != RZ_FLOAT_RPREC_64) {
+			float128_t fma_resized = f128_mulAdd(a_resized, b_resized, c_resized);
```
b1llow: **[correctness] The default precision-80 binary80 FMA still double-rounds through f128.**

The new staging below fixes double-rounding only for the reduced-precision (32/64) cases, but the default path keeps `f128_mulAdd` → `f128_to_extF80`: an extF80 product of two 64-bit significands needs up to 128 significant bits while f128 carries 113, so the intermediate rounding is lossy and the common (precision-80) case can be off by 1 ulp — the same defect class this block exists to eliminate. The deeper fix is to add upstream's `extF80_mulAdd.c` to the rizinorg/softfloat fork's meson source list: it rounds once, honors `extF80_roundingPrecision` (32/64/80), and computes exact-cancellation zero signs natively, which would also let most of this staging block be deleted.

librz/il/theory_fbasic.c
```diff
@@ -452,6 +475,144 @@ void *rz_il_handler_fmod(RzILVM *vm, RzILOpPure *op, RzILTypePure *type) {
 	return ret;
 }
 
+static RzFloatRMode runtime_rmode_from_value(const RzBitVector *value) {
```
b1llow: **[correctness/design] Invalid runtime rounding-mode values are silently coerced to RNE.**

Any 32-bit rmode bitvector whose value is outside RNE..RTZ (e.g. 5, or a garbage register from a buggy lifter or foreign core) evaluates with round-to-nearest — a rounding mode the IL never asked for — with no validation failure, no IL exception event, and no logged warning. That is silently-wrong floating-point output, and it is inconsistent with the strictness applied everywhere else in this series (the validator rejects wrong rmode *widths*, and static rmode enums are checked). Consider failing evaluation (NULL) or raising an IL exception on unrecognized values. (As a side note, since the enum is contiguous 0..4 with `_UNK` as terminator, `v < RZ_FLOAT_RMODE_UNK ? (RzFloatRMode)v : <fallback>` expresses the same check in two lines.)

librz/il/il_validate.c
```diff
@@ -652,15 +666,124 @@ VALIDATOR_PURE(float_binop_with_round) {
 	VALIDATOR_ASSERT(sy.type == RZ_IL_TYPE_PURE_FLOAT, "Right operand of %s op is not a float.\n", rz_il_op_pure_code_stringify(op->code));
 
 	// flatten validator assert
-	VALIDATOR_ASSERT(sx.props.f.format == sy.props.f.format, "Op %s formats of left operand (%s) and right operand (%s) do not agree.\n",
-		rz_il_op_pure_code_stringify(op->code),
-		rz_il_sort_pure_stringify(sx),
-		rz_il_sort_pure_stringify(sy));
+	if (sx.props.f.format != sy.props.f.format) {
+		rz_warn_if_reached();
+		char *sx_s = rz_il_sort_pure_stringify(sx);
```
b1llow, line 723: **[consistency] The stringify leak fixed here survives in the sibling float validators.**

This hunk frees the `rz_il_sort_pure_stringify()` results on the failure path of `float_binop_with_round`, but the identical pattern still leaks at `VALIDATOR_PURE(forder)` (il_validate.c:611-612) and `float_terop_with_round` (il_validate.c:803-804): both pass `rz_il_sort_pure_stringify(sx/sy)` straight into `VALIDATOR_ASSERT`, which never frees them, so every failed forder/fmad validation leaks two strings — the same defect commit "Fix IL float binop validation leaks" removed here. Applying the same sx_s/sy_s treatment to those two validators finishes the job in one place.

test/db/archos/linux-x64/dbg_cont_back
```diff
@@ -35,6 +35,7 @@ EOF
 RUN
 
 NAME=dbg.manycontback
+TIMEOUT=1200
```
b1llow, line 38: **[hygiene] A 20x timeout raise (to 1200s) lands in an unrelated RzIL float branch with no root cause.**

If the reverse-debug continue-back test recently got slower, this masks a real regression in CI permanently; if it's long-standing flakiness, it belongs in its own change with a linked issue so the underlying slowness is tracked rather than papered over. Either way it would be better split out of this branch.

librz/util/float/float.c
```diff
@@ -983,6 +1045,25 @@ RZ_API bool rz_float_is_zero(RZ_NONNULL RzFloat *f) {
 	return type == RZ_FLOAT_SPEC_ZERO;
 }
 
+static inline bool is_f80_infinity(RzFloat *f) {
```
b1llow, line 1081: **[altitude/efficiency] Binary80 pseudo-infinity canonicalization is now open-coded at five consumer sites.**

`detect_spec` (integer bit cleared before the zero-mantissa check), this `is_f80_infinity` bit-loop, `rz_float_is_equal`'s forced-true ternaries, `normalize_f80_infinity` in `rz_float_cmp`, and the succ/pred boundary guards each spell the same rule ("exponent all-ones + zero fraction = infinity regardless of the explicit integer bit") differently. The next bit-level consumer will compare raw bitvectors and reintroduce the bug this series fixed piecemeal — a single shared predicate (or normalizing the encoding in `of_float80`) would keep the rule in exactly one place. Incidentally, `is_f80_infinity` is equivalent to `f->r == RZ_FLOAT_IEEE754_BIN_80 && (spec == PINF || spec == NINF)` via the already-updated `rz_float_detect_spec`, and the unconditional ~79-iteration scan per `rz_float_is_equal` call (plus the two per-bit ternaries in the loop) can be gated on `x->r == RZ_FLOAT_IEEE754_BIN_80` so f32/f64 comparisons stay a straight word-wise compare.

librz/util/float/float.c
```diff
@@ -1660,38 +1812,146 @@ RZ_API RZ_OWN RzFloat *rz_float_convert(RZ_NONNULL RzFloat *f, RzFloatFormat for
...
-	// 1.MM..M * 2^exp_no_bias == 1MM..M * 2^(exp_no_bias - man_len)
-	// 0.MM..M * 2^exp_no_bias == 00..1X..X * 2^(exp_no_bias - man_len)
-	RzFloat *ret = round_float_bv_new(sign, exp, sig, old_format, format, mode);
-
-	rz_bv_free(sig);
+	set_float_rounding_mode(mode);
+	softfloat_exceptionFlags = 0;
+	RzFloat *ret = NULL;
+	switch (f->r) {
+	case RZ_FLOAT_IEEE754_BIN_16: {
```
b1llow: **[simplification] The 5x5 nested switch is twenty near-identical cases of copy-paste.**

Each source-format case repeats the same skeleton (`to_floatXX`, inner switch over the other four formats, `of_floatYY` wrap), so adding or fixing a conversion path means touching up to ten symmetric sites; the generic implementation this replaced handled every pair in ~20 lines. A five-element macro invocation per source format (generating the inner switch) or a [5][5] table of conversion thunks collapses the dispatch to one row per format while keeping the new SoftFloat routing and exception accumulation.

librz/il/theory_fbasic.c
```diff
@@ -452,6 +475,144 @@ void *rz_il_handler_fmod(RzILVM *vm, RzILOpPure *op, RzILTypePure *type) {
...
+	default:
+		return RZ_FLOAT_RMODE_RNE;
+	}
+}
+
+void *rz_il_handler_float_with_rmode(RzILVM *vm, RzILOpPure *op, RzILTypePure *type) {
+	rz_return_val_if_fail(vm && op && type, NULL);
+
+	RzILOpBitVector *rmode_op;
+	switch (op->code) {
```
b1llow: **[simplification/reuse] The opcode→rmode-operand mapping is written out three times across two files.**

This 8-case extraction switch is duplicated character-for-character in `VALIDATOR_PURE(float_with_runtime_rmode)` (il_validate.c:684), and each file then adds a second near-identical 8-case switch to build the delegated op — and the five binop cases only differ in the spelling of union members that all share one typedef (`RzILOpArgsFloatAlgBinopWithRmode`). Adding the next runtime-rmode op means editing four synchronized switches that will silently drift. A single shared accessor next to the `RzILOpArgs*WithRmode` structs in rz_il_opcodes.h, plus grouped case labels for the binops, removes ~70 lines and one whole class of divergence.

subprojects/packagefiles/softfloat/include/platform.h
```diff
@@ -0,0 +1,76 @@
...
+/* SOFTFLOAT_FAST_INT64: (as sourced from http://www.jhauser.us/arithmetic/SoftFloat-3/doc/SoftFloat-source.html)
+ * Can be defined to indicate that the build target's implementation of 64-bit
+ * arithmetic is efficient. For newer 64-bit processors, this macro should
+ * usually be defined. For very small microprocessors whose buses and registers
+ * are 8-bit or 16-bit in size, this macro should usually not be defined. Whether
+ * this macro should be defined for a 32-bit processor may depend on the target
+ * machine and the applications that will use SoftFloat. */
+#define SOFTFLOAT_FAST_INT64 1
+/* For Rizin, we assume that the only architectures we'll be running on will have
+ * a 64-bit integer. The fast part is not too relevant either, since we don't
```
Rot127, line 50: > For Rizin, we assume that the only architectures we'll be running on will have a 64-bit integer.

Do we?
@notxvilka @wargio

librz/il/theory_fbasic.c
```diff
@@ -452,6 +499,108 @@ void *rz_il_handler_fmod(RzILVM *vm, RzILOpPure *op, RzILTypePure *type) {
 	return ret;
 }
 
+static bool runtime_rmode_from_value(const RzBitVector *value, RzFloatRMode *out) {
+	if (rz_bv_len(value) != 32) {
+		return false;
```
Rot127: rz_warn_if_reached?

librz/il/theory_fbasic.c
```diff
@@ -452,6 +499,108 @@ void *rz_il_handler_fmod(RzILVM *vm, RzILOpPure *op, RzILTypePure *type) {
...
+	}
+	ut64 v = rz_bv_to_ut64(value);
+	if (v >= RZ_FLOAT_RMODE_UNK) {
+		return false;
```
Rot127: rz_warn_if_reached?

test/unit/test_float.c
```diff
@@ -1757,9 +2133,13 @@ bool f80_ieee_cast_test(void) {
 }
 
 bool all_tests() {
+#if !defined(__TINYC__)
```
Rot127, line 2140: Document why it is excluded please.

Rot127 (changes requested):
Those nitpicks. Otherwise lgtm.

subprojects/packagefiles/softfloat/include/platform.h
notxvilka: Yes, even if it is 32bit arch it should support int64 in C

librz/il/il_validate.c
```diff
@@ -652,15 +674,69 @@ VALIDATOR_PURE(float_binop_with_round) {
 	VALIDATOR_ASSERT(sy.type == RZ_IL_TYPE_PURE_FLOAT, "Right operand of %s op is not a float.\n", rz_il_op_pure_code_stringify(op->code));
 
 	// flatten validator assert
-	VALIDATOR_ASSERT(sx.props.f.format == sy.props.f.format, "Op %s formats of left operand (%s) and right operand (%s) do not agree.\n",
-		rz_il_op_pure_code_stringify(op->code),
-		rz_il_sort_pure_stringify(sx),
-		rz_il_sort_pure_stringify(sy));
+	if (sx.props.f.format != sy.props.f.format) {
+		rz_warn_if_reached();
```
thestr4ng3r:
```suggestion
```
This check is not an assertion

librz/il/il_validate.c
```diff
@@ -674,12 +750,18 @@ VALIDATOR_PURE(float_terop_with_round) {
...
 	VALIDATOR_ASSERT(sz.type == RZ_IL_TYPE_PURE_FLOAT, "3rd operand of %s op is not a float.\n", rz_il_op_pure_code_stringify(op->code));
 
-	VALIDATOR_ASSERT((sx.props.f.format == sy.props.f.format) && (sx.props.f.format == sz.props.f.format),
-		"types of operand in op %s do not agree: operand1 (%s) operand2 (%s) operand3 (%s)",
-		rz_il_op_pure_code_stringify(op->code),
-		rz_il_sort_pure_stringify(sx),
-		rz_il_sort_pure_stringify(sy),
-		rz_il_sort_pure_stringify(sz));
+	if ((sx.props.f.format != sy.props.f.format) || (sx.props.f.format != sz.props.f.format)) {
+		rz_warn_if_reached();
```
thestr4ng3r:
```suggestion
```

librz/include/rz_il/rz_il_opcodes.h
```diff
@@ -410,6 +410,27 @@ typedef struct rz_il_op_args_float_fconvert_t {
...
+/**
+ * \brief Evaluate a binary80 expression using the requested significand precision.
+ *
+ * [FWITH_RPREC] temporarily sets SoftFloat binary80 arithmetic precision for its
+ * child expression. The child must have binary80 format.
+ */
+typedef struct rz_il_op_args_float_with_rprec_t {
+	RzFloatRPrecision precision;
+	RzILOpFloat *f;
+} RzILOpArgsFwithRprec;
```
thestr4ng3r: This op breaks the premise that a pure op is equivalent to a pure function as the temporary setting is essentially a side-effect.
If the RPrecision is a relevant parameter to downstream computations, the only ways to express this are either to attach it to float values or as an explicit argument to every computation op that respects it, depending on which fits better.

librz/include/rz_il/rz_il_opcodes.h
```diff
@@ -410,6 +410,27 @@ typedef struct rz_il_op_args_float_fconvert_t {
...
 
+/**
+ * \brief convert between different float formats using a runtime rounding mode
+ * [FCONVERT_WITH_RMODE] evaluates rmode as a 32-bit bitvector before converting f.
+ */
+typedef struct rz_il_op_args_float_fconvert_with_rmode_t {
+	RzFloatFormat format;
+	RzILOpBitVector *rmode;
+	RzILOpFloat *f;
+} RzILOpArgsFconvertWithRmode;
```
thestr4ng3r: Does this not contradict the pr description here?
```
These helpers expand to ordinary LET/ITE expressions and do not introduce a dynamic rounding-mode opcode.
```

I see there are actual new opcodes like `RZ_IL_OP_FCONVERT_WITH_RMODE` and no helpers.

thestr4ng3r (changes requested):
Maybe I am missing something, but what is the initial goal of introducing runtime rounding modes? Where is this going to be used?

Rot127:
@thestr4ng3r 
> Maybe I am missing something, but what is the initial goal of introducing runtime rounding modes? Where is this going to be used?

Many archs can change the rounding modes in runtime. It is a common issue sadly.

librz/il/il_validate.c
Rot127: It is good for debugging though. The time I spend figuring out why something returned NULL, just to end up at such an `if` case which silently returned, is uncountable.

thestr4ng3r:
> @thestr4ng3r
> 
> > Maybe I am missing something, but what is the initial goal of introducing runtime rounding modes? Where is this going to be used?
> 
> Many archs can change the rounding modes in runtime. It is a common issue sadly.

Do you have an example? Especially between how many different rounding modes can be switched dynamically for a single operation. My question aims specifically at the idea from the description here, which the implementation in this pr from what I can tell at the moment does *not* match:

> These helpers expand to ordinary LET/ITE expressions and do not introduce a dynamic rounding-mode opcode.

Whether it would not make more sense to just emulate the switch with ITE rather than introducing additional complexity to the language.

librz/il/il_validate.c
thestr4ng3r: @Rot127 That is right for assertions, but this case is an expected code block, not something that should not be reached like a NULL assertion. On top of that, there is already the report `Op %s formats of left operand (%s) and right operand (%s) do not agree.` below, which should give you enough info to start debugging.

Rot127: True, just checked again and I wasn't sure in my comments as well.

Rot127:
> Do you have an example? 

Sparc for example: https://github.com/rizinorg/rizin/blob/dd618bb30a3ca7dea2f41d0c3e31d448d3070a1d/librz/arch/isa/sparc/sparc_il.h#L86-L94

I must admit though, I checked this PR only briefly. And didn't think about the implementation.
For the rounding mode it is probably better to make the `rmode` argument in all the float IL ops a `RzILOpPure`?

b1llow:
> Maybe I am missing something, but what is the initial goal of introducing runtime rounding modes? Where is this going to be used?

Because it is very common to select the round mode at runtime, we used to handle it with multiple nested ITEs. However, expressing the round mode with ITEs also has its problems. For example, in some cases, the recursion depth is very deep, which can cause an IL OP to be printed as a string that may reach tens of MiB.

librz/include/rz_il/rz_il_opcodes.h
b1llow: This was the initial implementation, which later shifted to the current direction rather than the original ITE implementation. I will update the description.

librz/include/rz_il/rz_il_opcodes.h
b1llow: Our current SoftFloat implementation internally uses global variables to store states such as the round mode. This OP ensures that a fixed round mode is used during execution, rather than relying on global state.

If we don't adopt a similar approach, the scope of changes might be very large in order to maintain the purity of float operations? Because a large number of float operations involve rounding.

notxvilka:
The original BAP Core theory has Rmode support, so this change will just close that gap of RzIL compared to the original: 
- https://binaryanalysisplatform.github.io/bap/api/master/bap-core-theory/Bap_core_theory/Theory/Rmode/index.html
- https://github.com/BinaryAnalysisPlatform/bap/blob/034bbd09646a589d5eb1f10b740177d628da80fd/lib/bap_core_theory/bap_core_theory_value.ml#L274

librz/include/rz_il/rz_il_opcodes.h
thestr4ng3r: Ok, I see. What if we hardcode setting it to `RZ_FLOAT_RPREC_80` before all operations? Or are there any cases where a precision lower than this is necessary for results accurate to the emulated architecture? I am wondering also because BAP does not seem to have a notion of this precision from what I can tell.

thestr4ng3r:
Thanks all for the comments, I get the context much better now. With all that in mind, the dynamic round mode operand instead of `ITE` makes sense. What I do not like yet though is the fact that with this, all the ops are doubled for static and dynamic rmode. This would become a real pain for users of the IL like the abstract interpreter.

What about simply making all regular operations always take a dynamic rmode? If I read it right, this is also what @Rot127 suggested here:

>  For the rounding mode it is probably better to make the rmode argument in all the float IL ops a RzILOpPure?

Alternatively, in case doing so would bloat the resulting lifted ILs too much for cases where dynamic is not needed, what about something like this?

```c
typedef struct rz_il_op_arg_float_rmode_t {
	bool static;
	union {
		RzFloatRMode static;
		RzILOpBitVector *dynamic;
	} val;
} RzILOpArgFloatRMode;

typedef struct rz_il_op_args_float_alg_binop_with_rmode_t {
	RzILOpArgFloatRMode rmode;
	RzILOpFloat *x;
	RzILOpFloat *y;
} RzILOpArgsFloatAlgBinopWithRMode;
```

Finally, what I am not fond of is that the bitvector type is abused to express the rmode. The clean way would be to introduce a new `RzILTypePure` type specifically for the rmode, but I am not sure if that is really worth it for us so I am fine with keeping it as a bitvector for now.

librz/il/theory_fbasic.c
```diff
@@ -452,6 +499,110 @@ void *rz_il_handler_fmod(RzILVM *vm, RzILOpPure *op, RzILTypePure *type) {
...
 
+static bool runtime_rmode_from_value(const RzBitVector *value, RzFloatRMode *out) {
+	if (rz_bv_len(value) != 32) {
+		rz_warn_if_reached();
+		return false;
+	}
+	ut64 v = rz_bv_to_ut64(value);
+	if (v >= RZ_FLOAT_RMODE_UNK) {
+		rz_warn_if_reached();
+		return false;
```
thestr4ng3r: > while invalid runtime values cause evaluation to fail with an error

I think it would be better to fall back to some default value instead of the error. The idea is that type-checked code should be guaranteed to never fail and since the type checker can not ensure the dynamic rmode will always be a valid enum value, a fallback is the best we can do.

thestr4ng3r (changes requested):
Please rename occurrences of `Rmode` to `RMode` to be consistent with `RzFloatRMode` for example.

thestr4ng3r (approved):
Thank you!

librz/include/rz_il/rz_il_opcodes.h
```diff
@@ -368,14 +368,37 @@ typedef RzILOpArgsUnopFloat RzILOpArgsFabs;
 typedef RzILOpArgsUnopFloat RzILOpArgsFsucc;
 typedef RzILOpArgsUnopFloat RzILOpArgsFpred;
 
+/**
+ * \brief Selects how a floating-point rounding-mode argument is represented.
+ */
+typedef enum rz_il_op_arg_float_rmode_kind_t {
+	RZ_IL_OP_ARG_FLOAT_RMODE_STATIC,
+	RZ_IL_OP_ARG_FLOAT_RMODE_DYNAMIC,
```
Rot127: Please document these. What the difference between static and dynamic is most importantly.

librz/include/rz_il/rz_il_opcodes.h
```diff
@@ -368,14 +368,37 @@ typedef RzILOpArgsUnopFloat RzILOpArgsFabs;
...
+} RzILOpArgFloatRModeKind;
+
+/**
+ * \brief A floating-point rounding-mode argument.
+ *
+ * Static rounding modes are stored directly. Dynamic rounding modes are
+ * represented by a 32-bit bitvector IL expression whose value is a
+ * RzFloatRMode.
```
Rot127: Static missing in docs

Rot127 (changes requested):
Sorry if I missed it, what was the rational again behind having static and dynamic? Instead of only dynamic?
I feel like having both is pretty weird to handle if one has to implement a VM.
Lifting seems easier, sure. But writing the lifting code is only done once, implementing a VM consuming RzIL is done not just by us, but also by Rizin users. So the rounding mode should be more comfortable for that case, or is there an advantage I don't see?

b1llow:
> Sorry if I missed it, what was the rational again behind having static and dynamic? Instead of only dynamic? I feel like having both is pretty weird to handle if one has to implement a VM. Lifting seems easier, sure. But writing the lifting code is only done once, implementing a VM consuming RzIL is done not just by us, but also by Rizin users. So the rounding mode should be more comfortable for that case, or is there an advantage I don't see?



Prior to this PR, floating-point operations only accepted a static `RzFloatRMode`. Lifters that needed a runtime-determined rounding mode already worked around this with nested ITEs, where each branch performed the same operation under a different static mode. This pattern appears in many existing call sites.

Moving to a dynamic-only representation would require rewriting all of these static call sites and ITE patterns, along with their construction and serialized representation — significantly expanding an already large PR.

Supporting both forms instead keeps existing lifting code valid while letting new lifters use a more compact dynamic expression. For the VM, the distinction can be localized to a single helper that either returns the stored static enum or evaluates and decodes the dynamic 32-bit bitvector; the floating-point operation itself only ever sees a resolved `RzFloatRMode`. Static modes also avoid the cost of allocating and evaluating a constant bitvector node in the common fixed-mode case.

In short, this is primarily a backward-compatibility and incremental-migration decision, not a case where both representations are semantically necessary. If a dynamic-only design is preferable long-term, I'd treat that as a separate, deliberate migration.

Rot127 (approved):
Thanks, makes sense.

b1llow:
> Thanks, makes sense.

Can this PR be merged?

Rot127:
Yes, could you please open an issue so we can track which archs have to be converted to dynamic rounding mode.

Sparc, Mips, x86, hexagon are the ones which require or would profit from a refactor.

PR #6664, opened 2026-08-09
Fix aar ignoring vmap memory.

Rot127:

**Detailed description**

That fixes a really annoying bug of `aar` ignoring any `vmap`. If `core_search_for_xrefs_in_boundaries()` gets an interval which lies within a `vmap`, `rz_io_size()` returns the size equivalent `mmap` (not `vmap`).
Which can be smaller and make the if case fail.

This seems to be a bug for all virtual files mapped in memory. If a virtual file of size `0x100` is mapped to a memory map of size `0x180`, it creates a `vmap` of `0x100` for the file content and a `0x80 mmap` for the empty bytes.
These maps are names `vmap.filename`, `mmap.filename`.

`rz_io_size()` now seems to only ever return the size of the `mmap`. I don't know why it does that.
And because it is one of the stone age functions I am not going down there to look.

This bug breaks `aar` for Qualcomm split firmware images (`.mdt`). Removing this check should not have an impact, because I added a check for the memory read.

**Test plan**

Added

I checked the failing test in `db/cmd/cmd_coverage`. The "missing xrefs" were false positives.

**Closing issues**

...

### DIFF
--- a/librz/core/canalysis.c
+++ b/librz/core/canalysis.c
@@ -2115,6 +2115,7 @@ RZ_API int rz_core_analysis_search_xrefs(RZ_NONNULL RzCore *core, ut64 from, ut6
 		}
 		int bytes_read = rz_io_nread_at(core->io, at, buf, bsz);
 		if (bytes_read <= 0) {
+			rz_cons_break_pop();
 			free(buf);
 			free(block);
 			return -1;
--- a/librz/core/canalysis.c
+++ b/librz/core/canalysis.c
@@ -2115,6 +2115,8 @@ RZ_API int rz_core_analysis_search_xrefs(RZ_NONNULL RzCore *core, ut64 from, ut6
 		}
 		int bytes_read = rz_io_nread_at(core->io, at, buf, bsz);
 		if (bytes_read <= 0) {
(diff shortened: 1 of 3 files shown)


notxvilka (approved):

librz/core/canalysis.c
```diff
@@ -2110,11 +2109,14 @@ RZ_API int rz_core_analysis_search_xrefs(RZ_NONNULL RzCore *core, ut64 from, ut6
...
 	while (at < to && !rz_cons_is_breaked()) {
-		int i = 0, ret = bsz;
+		int ret = bsz;
 		if (!rz_io_is_valid_offset(core->io, at, RZ_PERM_X)) {
 			break;
 		}
-		(void)rz_io_read_at_mapped(core->io, at, buf, bsz);
+		int bytes_read = rz_io_nread_at(core->io, at, buf, bsz);
+		if (bytes_read <= 0) {
+			return -1;
```
wargio, line 2118: aren't we leaking `buf`?

wargio: also `block` probably

PR #6668, opened 2026-08-16
Reload malloc:// bin info from the live buffer with obR

### LINKED ISSUE #6607: [rz-test] Enable format autodetection for virtual URIs (e.g., malloc://)
**Description**

Currently, rz-test supports automatic format detection when loading standard binary files. However, this behavior does not extend to virtual URIs or memory-allocated buffers specified via environment variables like FILE=malloc://32.

When testing shellcodes, raw architecture payloads, or specific fuzzing edge cases using virtual or blank spaces, rz-test fails to detect or falls back unpredictably because it expects a physical file for format probing, or doesn't pass the proper flags/bins to the core analysis engine during initialization.

**Expected Behavior**

rz-test should attempt format autodetection regardless of whether the target FILE is a physical file on disk or a virtual URI scheme supported by Rizin io plugins (like malloc://, mmap://, etc.), provided the buffer contains enough bytes for a signature match or the test configuration implies probing.

**Actual Behavior**
Autodetection is either skipped or fails when FILE does not point to a valid file path, forcing developers to explicitly define the format/architecture even for dynamic or temporary memory allocations where Rizin could otherwise guess the format from subsequent writes or pre-filled magic bytes.

atirna:

**Detailed description**

empty `malloc://` lands on the dummy `any` plugin at open. writing elf/mach0 magic into it used to leave format stuck on `any` unless you reopened the file, which throws away the buffer.

`obR` already reloads bin info from the current IO buffer (the help text even mentions malloc://). this PR uses that instead of probing on every `wx`:

- `wx` does not change format. scripts that set `asm.arch` then write code are not surprised by a reload.
- `obR` re-runs plugin autodetection on the live buffer and does not discard it.
- `hex://` with magic already in the uri still detects at open (`rz_core_bin_load`).
- `rizin -` loads bin info after stdin is written into `malloc://`.
- `om` / `rz_core_file_malloc_copy_chunk` writes the copied bytes before bin load, so the probe sees the data.
- `core_bin_reload` keeps the replacement binfile on the core file after the old one is deleted.

reverted an unrelated `avr` `check_buffer` tweak that was in an earlier revision of this PR.

AI: Cursor Agent / cursor-grok-4.6-high for this rework. earlier revisions also used Codex / gpt-5.6-terra and Hermes Agent / glm-5.3. i reviewed the diff and checks.

**Test plan**

- `./build/test/unit/test_core_bin` 13/13 (write stays `any`; `obR` reloads elf and keeps the IO bytes; mach0 via `obR`; hex:// elf at open)
- `rz-test db/cmd/file_malloc` 6/6
- manual: `rizin -q malloc://128` then `wx 7f45...`; `i~format` stays `any`; `obR`; `i~format` is `elf`; `e io.va=0; p8 4 @ 0` still `7f454c46`

**Closing issues**

closes #6607


librz/core/cfile.c
```diff
@@ -851,7 +851,7 @@ RZ_IPI void rz_core_bin_autodetect_virtual_io(RzCore *core) {
 	}
 	RzBinPlugin *plugin = rz_bin_get_binplugin_by_buffer(core->bin, buf);
 	rz_buf_free(buf);
-	if (!plugin) {
+	if (!plugin || strcmp(plugin->name, "elf")) {
```
wargio: i do not think this is correct as you want to detect any valid file format.

atirna:
done. the elf special case is gone: after a write the re-probe now accepts any plugin that claims the buffer, so mach0/elf64/pe/... all reload bin info, not just elf32.

that broad gate was the reason the elf check existed though, some check_buffer implementations are loose enough to claim shellcode. so instead of narrowing the gate i tightened the claimers that actually misfire:

- avr no longer claims from the reset vector alone (arm `ldr` bytes decode as avr rjmp too). it now requires a full interrupt vector table, smallest known board has 19 vectors, so foreign shellcode stays on `any`
- when a plugin claims by magic but fails to load, `rz_bin_reload` now keeps the current bin file instead of leaving the bin empty

the implicit re-probe also stopped clobbering `asm.arch`/`asm.cpu` and seeking to the entrypoint mid-script, only the detected plugin becomes current.

checked: `wx cefaedfe070000010300000002000000; i~format` gives `mach0` now, elf64 works the same way, arm shellcode stays `any`, and `db/cmd/file_malloc` + `db/formats/avr` pass 17/17 with the stricter avr check.

librz/bin/p/bin_avr.c
```diff
@@ -665,17 +665,44 @@ static void avr_parse_interrupt_vectors(RzBuffer *buf, RzVector /*<ut64>*/ *inte
...
 		return false;
 	}
 
-	size_t n_bytes = 2;
-	if (is_jmp(buf, 0) || is_rjmp32(buf, 0)) {
-		n_bytes = 4;
-	} else if (!is_rjmp(buf, 0)) {
+	size_t n_bytes = avr_interrupt_vector_size(buf);
+	if (!n_bytes) {
 		return false;
```
wargio, line 679: this is not the same output? also, why you are touching this when your PR is unrelated to this?

librz/bin/bin.c
```diff
@@ -252,6 +252,13 @@ RZ_API RzBinFile *rz_bin_reload(RzBin *bin, RzBinFile *bf, ut64 baseaddr) {
 	opt.filename = bf->file;
 	rz_buf_seek(bf->buf, 0, RZ_BUF_SET);
 	RzBinFile *nbf = rz_bin_open_buf(bin, bf->buf, &opt);
+	if (!nbf) {
+		// The new buffer was rejected (e.g. a plugin claimed the data by
+		// magic but failed to load it). Keep the current file selected
+		// instead of leaving the bin without one.
```
wargio: these comments are useless

librz/core/cfile.c
```diff
@@ -812,6 +812,73 @@ static bool core_file_do_load_for_io_plugin(RzCore *r, ut64 baseaddr, ut64 loada
...
+	// changes how the written code disassembles. So only make the detected
+	// plugin current (so "i" reflects it) and leave everything else alone.
+	ut64 cur_offset = core->offset;
+	if (bf) {
+		RzBinFile *nbf = rz_bin_reload(core->bin, bf, UT64_MAX);
+		if (!nbf) {
+			// The claimed plugin could not load the data (weak magic
+			// match); keep the current bin state untouched.
+			return;
+		}
```
wargio: these comments are not needed.

wargio (changes requested):
I feel like this should be done via a command to reload the bin & autodetect without discarding the current buffer as autodetecting this may do weird stuff.

wargio:
@notxvilka @thestr4ng3r @Rot127  please have a look

librz/bin/p/bin_avr.c
atirna: yes avr hunk is unrelated, i'll revert it

librz/bin/bin.c
atirna: will drop

librz/core/cfile.c
atirna: will drop

atirna:
dropping probing on wx as its not needed.

obR already reloads bin info from the current buffer without discarding it, so that can be used. wx leaves format as any, obR picks up elf/mach0 and the malloc bytes can stay. hex:// still detects at open because the buffer is already filled.

Issue #6663, opened 2026-08-07
Add a simple regex asm text search


Rot127:
The assembly search is not really in a good shape currently. But it is blocked on the RzArch refactor (see https://github.com/rizinorg/rizin/issues/5352).

Nonetheless, it would be nice to have a very simple "disassemble all instructions and match them against this regex". There is `/ao` already and it should be refactored. In its current state it is not useful anymore. 

That would be helpful and super straight forward to implement.
It can copy one of the modern searches like the string search and do it.

The command should be of the form
```
/ao <extended-regex>
```

It should definitely make use of the multi threading. And it should abortable.

I marked this as "good-first-issue" but it is only one for people who are familiar with C. It is too hard for programming beginners.

ksd-2005:
Hi,
I am new to open source contributions and have experience in C/C++ . 
I'd like to work on this issue.

My Plan:
rewrite `/ao` to disassemble over the search range and match each instruction's text against an extended regex, modelled on the string search rather than the current implementation. Use the thread pool for parallel scanning, and make the search abortable so long scans can be interrupted.

Happy to adjust the approach if you'd rather want it a different way.

ksd-2005:
Hi could you please share if this is assigned to me, I am interested in contributing to this.

Rot127:
You can just work on it and open a PR.

ksd-2005:
Following up on my Aug 11 comment - I'm implementing this on top of the RzSearch collection framework so it's actually multithreaded and honors search.in boundaries, replacing the legacy /ao handler rather than adding a parallel path. 
One design question I have is:
 since rz_arch already depends on rz_search, an asm collection can't live in librz/search. Is promoting rz_search_collection_new_bytes_space() + RzSearchFindBytesCallback to RZ_API (so core registers its own find callback and drives it via rz_search_on_io) the direction you'd want? Happy to coordinate with #6673 .

tahmeedahmed06-pixel:
Hi! I'd like to work on this issue. Is it still available, and would it be okay if I take it up?

Rot127:
See the discussion here: https://github.com/rizinorg/rizin/pull/6673#discussion_r3824911192

The RzArch modules first have to be made thread safe.
Feel free to coordinate with @Yashwin-i about it.

Issue #4055, opened 2023-12-30
Remove all globals & static variables where possible


wargio:
Currently rizin contains a lot of global variables, ideally we want to remove most of them.

Many fields shall not be touched like the `float.c` ones, etc..

Can be found via `$ find librz/ -type f -name "*.c" -o -name "*.h" -o -name "*.inc"  | xargs -I % ctags  -x  --c-kinds=v % | grep 'static ' | grep -v ' static const \|{\|\[\]' | grep -v ' const static \|{\|\[\]'`

```
is_interrupted_cb variable     35 librz/util/print.c static RzPrintIsInterruptedCallback is_interrupted_cb = NULL;
env              variable     39 librz/util/sys.c static char **env = NULL;
fd2close         variable   1511 librz/util/sys.c static HtUU *fd2close;
is_child         variable   1438 librz/util/sys.c static bool is_child = false;
is_child         variable   1515 librz/util/sys.c static bool is_child = false;
sys_pipe_mutex   variable   1437 librz/util/sys.c static RzThreadLock *sys_pipe_mutex;
sys_pipe_mutex   variable   1514 librz/util/sys.c static RzThreadLock *sys_pipe_mutex;
has_procthreadattr variable     53 librz/util/subprocess.c static bool has_procthreadattr = false;
lpDeleteProcThreadAttributeList variable     35 librz/util/subprocess.c static DeleteProcThreadAttributeList_t lpDeleteProcThreadAttributeList = NULL;
lpInitializeProcThreadAttributeList variable     33 librz/util/subprocess.c static InitializeProcThreadAttributeList_t lpInitializeProcThreadAttributeList = NULL;
lpUpdateProcThreadAttribute variable     34 librz/util/subprocess.c static UpdateProcThreadAttribute_t lpUpdateProcThreadAttribute = NULL;
mode_stderr      variable     57 librz/util/subprocess.c static DWORD mode_stderr;
mode_stdin       variable     55 librz/util/subprocess.c static DWORD mode_stdin;
mode_stdout      variable     56 librz/util/subprocess.c static DWORD mode_stdout;
pipe_id          variable     54 librz/util/subprocess.c static volatile long pipe_id = 0;
refcount         variable     52 librz/util/subprocess.c static long refcount = 0;
sigchld_pipe     variable    761 librz/util/subprocess.c static int sigchld_pipe[2];
sigchld_thread   variable    762 librz/util/subprocess.c static RzThread *sigchld_thread;
[...]

PR #6673, opened 2026-08-19
WIP: Refactor /ao for regex assembly search (#6663)

Yashwin-i:
[x] I've read the guidelines for contributing to this repository.

[x] I made sure to follow the project's coding style.

[ ] I've documented every RZ_API function and struct this PR changes. (N/A - no RZ_API changes)

[ ] I've added tests that prove my changes are effective (required for changes to RZ_API). (WIP)

[ ] I've updated the Rizin book with the relevant information (if needed).

[x] I've used AI tools to generate fully or partially these code changes and I'm sure the changes are not copyrighted by somebody else.

### Detailed description

This PR implements the core functionality for issue #6663 (Refactor /ao for regex assembly search). I have built a functional /ao <extended-regex> command that correctly reads memory, disassembles instructions, and matches them against the user's regex.

(Note: The original issue requested multithreading, but per @wargio's feedback, RzArch is not yet thread-safe. Therefore, this has been refactored into a single-threaded static function inside cmd_search.c to respect the current architecture).

What is currently working:
-> Added a new static helper do_asm_regex_search directly in cmd_search.c.
-> Disassembles current memory blocks using rz_asm_disassemble.
-> Compiles and executes extended regex queries against op.buf_asm.
-> The command is abortable using rz_cons_is_breaked().
-> Seamlessly integrates into the existing /ao command switch using public Rizin APIs.

(Note: I used Google's Gemini AI assistant to help format my code, and I have reviewed the code to ensure architectural soundness).

**Test plan**

Build the project using Meson/Ninja and enter the developer environment.

Open an empty memory block: rizin -

Verify default bytes with pd 3 (e.g., 0x00000000      jnbe  0x63).

Run the new regex search to match instructions:

> [0x00000000]> /ao jnbe
> 0x00000000   jnbe 0x63

You can also write custom assembly as a single string and search for it:

> [0x00000000]> "wa mov eax, 1; mov ebx, 2; add eax, ebx"
> [0x00000000]> /ao mov

Closing issues

Addresses #6663

### DIFF
--- a/librz/core/cmd/cmd_search.c
+++ b/librz/core/cmd/cmd_search.c
@@ -45,6 +45,60 @@ struct search_parameters {
 	bool regex_search;
 };
 
+static void do_asm_regex_search(RzCore *core, const char *regex_pattern) {
+    if (!regex_pattern || !*regex_pattern) {
+        RZ_LOG_ERROR("Usage: /ao <extended-regex>\n");
+        return;
+    }
+
+    RzRegexMulti *re = rz_regex_new_multi(regex_pattern, RZ_REGEX_EXTENDED, RZ_REGEX_DEFAULT, RZ_REGEX_UTF8);
+    if (!re) {
+        RZ_LOG_ERROR("Failed to compile regex: %s\n", regex_pattern);
+        return;
+    }
+
+    ut64 from = core->offset;
+    ut64 to = from + core->blocksize;
+    ut64 buf_size = to - from;
+
+    if (buf_size == 0) {
+        rz_regex_free_multi(re);
+        return;
+    }
+
+    ut64 current_offset = 0;
+    while (current_offset < buf_size) {
+        if (rz_cons_is_breaked()) {
+            break;
+        }
+
+        RzAsmOp op;
+        rz_asm_op_init(&op);
+        rz_asm_set_pc(core->rasm, from + current_offset);
+
+        int instr_len = rz_asm_disassemble(core->rasm, &op, core->block + current_offset, (int)(buf_size - current_offset));
+
+        if (instr_len < 1) {
+            rz_asm_op_fini(&op);
+            current_offset += 1;
+            continue;
+        }
+
+        RzPVector *matches = rz_regex_match_all(re->re8, rz_strbuf_get(&op.buf_asm), RZ_REGEX_ZERO_TERMINATED, 0, RZ_REGEX_DEFAULT);
+        
+        if (matches && !rz_pvector_empty(matches)) {
+            rz_cons_printf("0x%08" PFMT64x "   %s\n", from + current_offset, rz_strbuf_get(&op.buf_asm));
+        }
+        
+        rz_pvector_free(matches);
+        rz_asm_op_fini(&op);
+        
+        current_offset += instr_len;
+    }
+
+    rz_regex_free_multi(re);
+}
+
 static RzCmdStatus gadget_info(RzCore *core, int argc, const char **argv, RzCmdStateOutput *state, RzGadgetType gadget_type) {
 	const char *input = argc > 1 ? argv[1] : "";
 	if (!input) {
@@ -1498,7 +1552,7 @@ reread:
 			do_asm_search(core, &param, input + 2, 'c', search_itv);
 		} else if (input[1] == 'o') { // "/ao"
 			dosearch = 0;
-			do_asm_search(core, &param, input + 2, 'o', search_itv);
+            do_asm_regex_search(core, input + 2);
 		} else if (input[1] == 'a') { // "/aa"
 			dosearch = 0;
 			do_asm_search(core, &param, input + 2, 'a', search_itv);
--- a/librz/core/cmd/cmd_search.c
+++ b/librz/core/cmd/cmd_search.c
@@ -46,57 +46,57 @@ struct search_parameters {
 };
 
 static void do_asm_regex_search(RzCore *core, const char *regex_pattern) {
-    if (!regex_pattern || !*regex_pattern) {
-        RZ_LOG_ERROR("Usage: /ao <extended-regex>\n");
-        return;
-    }
+	if (!regex_pattern || !*regex_pattern) {
+		RZ_LOG_ERROR("Usage: /ao <extended-regex>\n");
+		return;
+	}
 
-    RzRegexMulti *re = rz_regex_new_multi(regex_pattern, RZ_REGEX_EXTENDED, RZ_REGEX_DEFAULT, RZ_REGEX_UTF8);
-    if (!re) {
-        RZ_LOG_ERROR("Failed to compile regex: %s\n", regex_pattern);
-        return;
-    }
+	RzRegexMulti *re = rz_regex_new_multi(regex_pattern, RZ_REGEX_EXTENDED, RZ_REGEX_DEFAULT, RZ_REGEX_UTF8);
+	if (!re) {
+		RZ_LOG_ERROR("Failed to compile regex: %s\n", regex_pattern);
+		return;
+	}
 
-    ut64 from = core->offset;
-    ut64 to = from + core->blocksize;
-    ut64 buf_size = to - from;
+	ut64 from = core->offset;
+	ut64 to = from + core->blocksize;
+	ut64 buf_size = to - from;
 
-    if (buf_size == 0) {
-        rz_regex_free_multi(re);
-        return;
-    }
+	if (buf_size == 0) {
+		rz_regex_free_multi(re);
+		return;
+	}
 
-    ut64 current_offset = 0;
-    while (current_offset < buf_size) {
-        if (rz_cons_is_breaked()) {
-            break;
-        }
+	ut64 current_offset = 0;
+	while (current_offset < buf_size) {
+		if (rz_cons_is_breaked()) {
+			break;
+		}
 
-        RzAsmOp op;
-        rz_asm_op_init(&op);
-        rz_asm_set_pc(core->rasm, from + current_offset);
+		RzAsmOp op;
+		rz_asm_op_init(&op);
+		rz_asm_set_pc(core->rasm, from + current_offset);
 
-        int instr_len = rz_asm_disassemble(core->rasm, &op, core->block + current_offset, (int)(buf_size - current_offset));
+		int instr_len = rz_asm_disassemble(core->rasm, &op, core->block + current_offset, (int)(buf_size - current_offset));
 
-        if (instr_len < 1) {
-            rz_asm_op_fini(&op);
-            current_offset += 1;
-            continue;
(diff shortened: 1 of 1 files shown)


ksd-2005:
Ran into the same question when looking at this issue, so what I found  is : 
in the build graph 

`librz/arch/meson.build` already lists `rz_search_dep` among `rz_arch`'s dependencies, and `librz/search/meson.build` only depends on `rz_util`, `rz_io`, `rz_magic`, `rz_crypto`, `rz_hash`. So an asm search collection can't live in `librz/search/` next to `string_search.c` — pulling `RzAsm` in there would make the dependency cycle `rz_search -> rz_arch -> rz_search`.

That leaves your first option (expose the collection API to core) as the workable one.
`RzSearchHit`, `RzSearchOpt`, `RzSearchFindOpt` and `rz_search_on_io()` are already `RZ_API` in `rz_search.h`. The only pieces behind the module boundary are `rz_search_collection_new_bytes_space()` (`RZ_IPI`, `librz/search/collection.c:51`) and the `RzSearchFindBytesCallback` typedef in `search_internal.h`.

On thread safety: `core->rasm` can't be shared across find threads - `rz_asm_set_pc()` mutates it, The collection's private user data would need one `RzAsm` per worker, configured from core's arch/bits/cpu/syntax/endianness up front.

For transparency: I picked this issue up in the issue thread on Aug 11 and I'm working on an implementation along these lines. Happy to combine efforts rather than duplicate - let me know what you'd prefer.

Yashwin-i:
Thanks for the detailed architectural analysis! I already have the core disassembly and regex engine prototyped and working here in this draft. Regarding the module boundary, I want to hold off on promoting `rz_search_collection_new_bytes_space` to the public RZ_API until the core maintainers confirm that is their preferred approach, as it affects the public ABI. Let's see what they advise before we rewire the headers!

ksd-2005:
On ABI point: 

Promoting an `RZ_IPI` symbol to `RZ_API` is additive to the export table rather than a break. 
The build uses `-fvisibility=hidden` (once check the `meson.build` file), so those symbols are just hidden from `librz_search.so`, marking one `RZ_API` exports it without touching any existing signature or struct layout. 
So it's a question of whether maintainers want that symbol in the public surface , so i think it doesn't lead to ABI break.

Since it's small change , so I will keep building on my branch against the promoted symbol so there's something concrete to look at when they  check in -  easy to change later if needed.

On coordinating: 
are you planning to carry on with the threaded implementation yourself? I am interested in working on the threaded version, want to make sure we're both not writing the same code.

Yashwin-i:
Thanks for the heads-up! Since altering the headers impacts the public API surface, I'm going to hold off on doing that on my branch and wait for the maintainers to weigh in on their preferred architectural direction first. Let's see what they advise before we coordinate any further code changes.

Yashwin-i:
Hey @ksd-2005, I’ve actually already coded the multithreaded `RzSearchCollection` approach locally to test it out. I also sorted out the cascading header dependencies (moving `RzSearchIsEmptyCallback` and `RzSearchFreeCallback` over to `rz_search.h` to keep the compiler happy).

That said, since exposing these internal collections alters the public ABI, I am holding off on pushing the changes to this PR until one of the core maintainers gives the green light. I’ve got the fully working code ready to push, so no need for you to duplicate the effort!

@Rot127 (or other maintainers) - let me know if you are comfortable with this API promotion, or if there is a different internal bridge you'd prefer I use!

wargio:
We cannot multithread asm search as RzArch is not thread safe.(yet)

librz/core/cmd/cmd_search_asm.c
```diff
@@ -0,0 +1,64 @@
+// SPDX-FileCopyrightText: 2024-2025 RizinOrg
+// SPDX-License-Identifier: LGPL-3.0-only
+
+#include <rz_core.h>
+#include <rz_search.h>
+#include <rz_util/rz_regex.h>
+
+void do_asm_regex_search(RzCore *core, const char *regex_pattern) {
```
wargio: this should be in `cmd_search.c` as it makes no sense to have it in a new file. also this should be a `static` function.

Yashwin-i: Hey @wargio, thanks for the review and the clarification on `RzArch` thread-safety!

Since the original issue description explicitly mentioned multithreading, I initially went down that path by creating a separate file and trying to hook into the search collection APIs. However, your feedback makes perfect sense. I've completely scrapped the multithreaded approach for now.

I did a clean reset of my branch, moved the single-threaded implementation directly into `cmd_search.c` as a static function, and force-pushed the clean commit. Let me know if this implementation looks good to go or if you'd like any other tweaks!

Rot127: Sorry, I thought RzArch is already thread safe. Will review this next week. Please ping me if I forget it.

wargio: RzArch is not implemented yet, and the current RzAnalysis/RzAsm code is mostly thread safe, but not all of it. also there are still some archs that are not thread safe. (see https://github.com/rizinorg/rizin/issues/4055)

Rot127: Seems like ARC, ARCcompat, Lanai, and TMS320 are the only ones left (the ARM assembler is irrelevant for this one). TMS320 is done by @notxvilka .
Let me try to fix the ARC ones then later this week.

Yashwin-i: No worries at all, @Rot127! That makes total sense given the ongoing architectural updates. Sounds great—take your time with the review, and I'll gladly ping you late next week if it slips off the radar. Thanks!

Yashwin-i: Hey @Rot127, just following up as promised ,it's been about a week, wanted to check if you've had a chance to look this over, or if it's still in the queue. No rush, just didn't want it to slip through the cracks!

Rot127: I actually started with the ARC PR. But somehow lost in what branch I did it :roll_eyes: 

I see two ways:
- you implement the search multi-threaded any ways. It can be tested on binaries which already have thread safe plugins. We can't merge it before the last archs are updated though. Sadly. But I think most of them are relatively easy to fix. Maybe even with AI.
- You could fix the RzArch plugins and open a PR. Then we can do that one again. This would help even more tbh.

Yashwin-i: Hey @Rot127, I'm happy to take **Option 2** and help fix the `RzArch` plugins first to make them thread-safe. That seems like the healthiest path forward for the codebase.

I'm actually aiming to stick around and propose a project for **GSoC '27** next year, so tackling core architectural debt like this is exactly the kind of deep dive I'm looking for. Let me know which architectures are the biggest priority, or if there's a specific pattern you want me to follow, and I can start opening PRs for those.

Separately, while waiting, I was digging into the CI failures on this branch. I already pushed the `clang-format` fix, but for `linux-gcc-tests-lsan`, I ran an ASAN/LSAN build locally and tracked down the memory leak.

My `do_asm_regex_search` pairs `rz_regex_new_multi` / `rz_regex_free_multi` correctly on every path. The leak is actually upstream in `librz/util/regex.c:291-294`, inside `rz_regex_new_multi`.

When the pattern itself fails to compile (e.g. `/ao [`), `re->re8` stays `NULL` and the function does a bare:

```c
free(re);
```

However, `re->ccontext` (created at line 258) is never freed, which results in an 88-byte leak.

Interestingly, the JIT-stack failure case a few lines down at `310-314` already handles this correctly by calling:

```c
rz_regex_free_multi(re);
```

So the fix should essentially be:

```diff
- free(re);
+ rz_regex_free_multi(re);
```

Since this shared utility leak is what's tripping the CI, should I include this one-line fix in the current PR so we have a clean slate, or would you prefer that I open a quick separate PR for it while we pivot to the `RzArch` thread-safety work?

Yashwin-i: Hey @Rot127 , just checking in whenever you get a chance , happy to start on whichever RzArch plugin you think is highest priority, or on the regex.c leak fix, whichever's more useful. No rush, just want to make sure this didn't slip off your radar.

Rot127: > However, re->ccontext (created at line 258) is never freed, which results in an 88-byte leak.

Good catch. Would you mind making a separate PR fixing it pelase?

> I'm happy to take Option 2 and help fix the RzArch plugins first to make them thread-safe.

Thanks! This would be of great help!

You can check this issue: https://github.com/rizinorg/rizin/issues/4055
Every file in `librz/arch/` belongs to an RzArch plugin. I would suggest to start with a plugin which has very few global or static variables left.

You can fix them by passing the variables as part of an object to the function.
So: move the global/static to an object, then every function using this variable has to get a pointer to that object, the object itself is initialized in the RzAsmPlugin->init() functions.

About wording: RzArch is the term we use for a plugin which does analysis and disassembly for a specific architecture. But internally they are split between disassembly and analysis. RzAsmPlugin does disassembly. And RzAnalysisPlugin is  analysis stuff. We will merge them in the near future to one. But for now they are split.

Yashwin-i: Hey @Rot127, opened the separate PR for the memory leak fix as requested: https://github.com/rizinorg/rizin/pull/6721

I'll come up with a concrete plan for next steps on the work before moving ahead , will follow up here or anywhere else if you say so, once I've got that together.

Yashwin-i: Hey @Rot127, picking Lanai since you're on ARC.

Only real static var: `opcodes_sorted` in `lanai-dis.c`, guarding a lazy `qsort()` of `lanai_opcodes[]` (declared non-static in `lanai-opc.c` - genuinely shared table, not per-instance). Also a `static char *op_names[]` in the same file, but that's read-only, just missing `const`.

This one doesn't fit "move to object + init()" cleanly since the array itself has to stay shared. Plan is `RzThreadLock` around the check-and-sort instead. Rizin already has `rz_th_lock_new`/`enter`/`leave` in `rz_th.h`.

One thing I haven't sorted out: where does the lock itself get created without the same race one level up? Lazy-creating it inside `print_insn_lanai` just moves the problem. Is there a standard place plugins do one-time setup before any thread touches them, or is a module-scope lock just initialized as a static and that's considered fine in practice here?

Rot127: Don't introduce global locks please. We have to end up without a single global variable.
Just move all of the globals into a struct like `ArcCtx`, including the pointer to the table, initialize it in `init()`, and pass it to all functions which require it.
Other functions can still update the table pointer.
Then you can add the locks (if required) in the struct.

Please open a draft PR for it. Then we can discuss it better.

PR #6705, opened 2026-09-05
Add RzBin plugin for `ar` archives

### LINKED ISSUE #6680: Add RzBin plugin for `ar` archives

It would be useful to have an bin plugin to open `ar` archives with static libraries inside.
Useful for commands which only need the bytes to do its thing, like the new prologue detection (https://github.com/rizinorg/rizin/pull/6514).

Simply setting the correct architecture and mapping the objects linear in memory would already be helpful.

If there are different architecture binaries in there, it should be reported or similar handled like fat binaries for Mac.

If the `ar` archive doesn't contain any object file, it can still unpack and open it. But notify the user that it couldn't detect object files.

Unpacking them manually. But that is a little uncomfortable. 

**Additional context**

https://en.wikipedia.org/wiki/Ar_(Unix)

kx7m2qd:
Adds a new RzBin plugin (`ar`) that recognizes and loads Unix `ar` archives (static libraries, `.a` files) by parsing the common-format member headers and mapping each member raw bytes linearly into memory, 16-byte aligned.

* Supports GNU short names, GNU long names via the `//` extended filename table, and BSD `#1/N` extended names
* Best-effort detects the architecture of each member by peeking at ELF/Mach-O magic, similar in spirit to how fat Mach-O binaries are handled; disagreeing members are flagged via a warning rather than failing to load
* Archives with no recognizable object members still open, with a warning instead of a hard failure

Closes #6680

- [x] I followed the project's coding style.
- [x] This PR does not change any documented `RZ_API` functions or structs.
- [x] No tests for `RZ_API` changes are required.
- [x] The Rizin book does not require an update for this internal bin-plugin addition.
- [x] I used AI tools to generate part of these changes and verified that the resulting code is original.

**Detailed description**

This pull request adds a new `RzBin` plugin for Unix `ar` archives, commonly used for static libraries such as `.a` files.

The plugin:

- Recognizes archives using the standard `!<arch>\n` magic.
- Parses archive member headers and maps each member's raw data into virtual memory.
- Supports GNU short names and long names through the `//` extended filename table.
- Supports BSD `#1/N` extended member names.
- Ignores symbol-table and filename-table members when exposing archive contents.
- Performs best-effort architecture detection for ELF and Mach-O members.
- Reports mixed architectures without rejecting the archive.
- Exposes archive members through bin maps, sections, metadata, and structured bin information.
- Allows archives with unrecognized member formats to open with a warning.

**Test plan**

Validation performed:

```sh
git diff --check

### DIFF
--- a/librz/bin/meson.build
+++ b/librz/bin/meson.build
@@ -1,5 +1,6 @@
 bin_plugins_list = [
   'any',
+  'ar',
   'art',
   'avr',
   'bf',
@@ -107,6 +108,7 @@ rz_bin_sources = [
   'relocs_patch.c',
   'stabs.c',
   'p/bin_any.c',
+  'p/bin_ar.c',
   'p/bin_art.c',
   'p/bin_avr.c',
   'p/bin_bf.c',
--- a/librz/bin/p/bin_ar.c
+++ b/librz/bin/p/bin_ar.c
@@ -1,0 +1,491 @@
+// SPDX-FileCopyrightText: 2026 Karna
+// SPDX-License-Identifier: LGPL-3.0-only
+
+// Unix `ar` archive RzBin plugin.
+//
+// Parses the common-format `ar` archive (as produced by GNU binutils `ar`
+// and used for static libraries, `.a` files) well enough to map every
+// member's raw bytes linearly into memory. This is useful for commands
+// that only need the raw bytes of the contained object files (e.g.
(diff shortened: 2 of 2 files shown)


Rot127:
See the discussion from https://github.com/rizinorg/rizin/issues/6680#issuecomment-5560168049 onward

PR #6722, opened 2026-09-09
core: read past the block boundary for null/unprintable-delimited ps

### LINKED ISSUE #4857: `ps` doesn't print long strings by default, because it depends on `core->block`.

kisnner26:
### Update: the first version of this PR broke CI, here's what changed and why

The first push failed Windows CI (both `RUN_TESTS=true` legs). Six tests failed, and none of them were flaky — every one pointed at a real gap in the original patch. Leaving this here since it explains design decisions a reviewer would otherwise have to ask about.

**1. `test/db/cmd/cmd_psj` "psj utf8 (#9064)"** uses `psj @!13 ...` — `@!13` temporarily narrows the block to 13 bytes for that one command. The original patch grew straight past that, ignoring the deliberate narrowing.

**2 through 4. `test/db/rzil/tricore`, `test/db/rzil/ppc64`, and (by the same mechanism, confirmed by inspection though I couldn't run them locally without the missing `rizin-testbins` submodule) `test/db/cmd/write` "wx pcache" and `test/db/cmd/cmd_r` "extend file"** all use `ps @! N` (N = 2, 3, 5) for the same reason: get an exact, short byte-for-byte view. Same underlying problem as #1 — those are all real, load-bearing uses of `@!N` as "exactly N bytes, no more," which the original patch didn't distinguish from an untouched default block.

**Fix for 1–4:** only grow when `core->blocksize == RZ_CORE_BLOCKSIZE` (the untouched default). There's no flag anywhere in core for "was this narrowed deliberately vs. is it just the default" — `core->tmpseek` covers seek modifiers but is not set for the `@!N` blocksize-override handler specifically (checked directly: it's `false` during `@!13`'s execution). The default-value check is the only signal available without deeper plumbing, and it's consistent with every one of these tests: all of them narrow well below the 256-byte default, so none of them are affected once growth is gated on being at the untouched default.

**5. `test/db/rzil/ppc64` "emulateme-big-endian"** (and `test/db/cmd/cmd_foreach` "iter string", same mechanism, not run locally for the same submodule reason) uncovered a second, independent bug: `ps` on a short, cleanly-terminated string (`"Hello from RzIL!"`) started printing CJK-looking garbage. Root cause: `str.encoding=guess` was being resolved *after* growth, so the heuristic was scanning the grown ~2048-byte window instead of the original block. Real binaries have plenty of zero-padding and byte patterns past any given string, and `rz_str_guess_encoding_from_buffer()` doesn't stop at a delimiter — it scans the entire buffer it's given. Over a big enough window it started misdetecting UTF-16.

**Fix:** resolve encoding from the original in-block buffer, before the growth decision. Verified directly — same input bytes, same code otherwise, produces `"type":"utf16be"` with encoding-after-growth and `"type":"ascii"` with encoding-before-growth.

Both fixes are additive to the original design (still an `RZ_CORE_BLOCKSIZE` check and a reordered two-line move), not a rewrite. Added two more `test/db/cmd/cmd_ps` cases covering exactly these two failure modes (`@!N` no longer grows; growth no longer corrupts an encoding guess), on top of the two from the original description. All four pass locally, and I re-verified the original issue's own repro still grows correctly, `delimiter=block` is still untouched, and the four previously-failing scenarios now reproduce their pre-patch output exactly — not just "no crash," the actual byte-for-byte expected values.

---

Fixes #4857.

## The bug

`core_print_string_in_block()` (the shared implementation behind `ps`, `psu`, `psw`, `psW`, `psm`, `psM`, `psn`, `ps+`) always capped the buffer it hands to `rz_str_stringify_raw_buffer()` at `core->blocksize - offset`:

```c
const ut8 *buffer = core->block + offset;
const ut32 length = core->blocksize - offset;
```

`delimiter=null`/`delimiter=unprintable` are documented to print until a real stop condition is found in the data — but they were never given more than the current block (256 bytes by default) to search in, so a string longer than that got cut off mid-character with no indication it was truncated:

```
$ rizin -qc "psW" test/bins/cmd/search/string_encodings/Georgian-Lipsum.utf32le
ლორემ იფსუმ დოლორ სით ამეთ, ილლუმ ელაბორარეთ ად ჰას, ილლუმ იისყუ
```

`delimiter=block` already documents this limit and tells users to raise it manually with `b <size>` — that mode is working as intended and is untouched by this PR.

## The fix

When `stop_at_nil` or `stop_at_unprintable` is set **and the block is at its untouched default size**, read further directly from IO via `rz_io_nread_at()`, capped at the existing `search.str.max_length` config (`RZ_BIN_STRING_SEARCH_BUFFER_SIZE`, 2048 bytes by default) — the same value the string-search subsystem already uses as its own ceiling, rather than a new magic number. `core->block` itself is left untouched; only the local `buffer`/`length` used for this one print call point at the wider IO-backed copy when it actually grew.

Encoding (including `str.encoding=guess`) is resolved from the original, un-grown buffer before that decision is made — see the update above for why.

## Scope, and why

The issue asks for a deeper refactor: routing this whole code path through `RzBuffer`/`RzIO` (`rz_buf_get_whole_hot_paths()`, per @Rot127's comment) instead of `core->block`. I read through the discussion (`well-mannered-goat`'s attempt, the back-and-forth with `#4762`) and that refactor has stalled multiple times over close to a year across several contributors, apparently because it touches `RzStrStringifyOpt`, `core_print_string_in_block()` and `core_print_raw_buffer()` together and keeps needing to be coordinated with unrelated work.

This PR deliberately does not attempt that. It fixes the literal, reported, user-visible bug — default `ps`/`psW` truncating a string that has a real terminator further in the file — with a change scoped to the one function that has the bug, without touching `core->block` semantics or any other caller. If the broader `RzBuffer` refactor is still wanted, this doesn't preclude it and isn't meant to close that door — it's a smaller, independently reviewable step.

## Testing

- Added four `test/db/cmd/cmd_ps` cases:
  - a `null`-delimited UTF-32LE string spanning multiple blocks now prints in full instead of stopping at the block boundary
  - the same buffer with `delimiter=block` prints identically before and after this patch, confirming that mode's documented behavior is untouched
  - `ps @! 2` on a buffer with more real content past byte 2 still prints exactly the 2 requested bytes — growth does not override an explicit `@!N`
  - a short, cleanly NUL-terminated ASCII string followed by binary padding past the default block no longer misdetects as UTF-16 under `str.encoding=guess`
- Manually verified against the `Georgian-Lipsum.utf32le` repro from the issue (I don't have the `rizin-testbins` submodule checked out locally, so I couldn't wire that specific binary into the `db` test, but the behavior is what the added tests exercise directly)
- `ps` on a short string, and `ps` at a non-zero offset (`s <off>; ps`), both unaffected
- `psj`/`psWj` (JSON output) also read the wider buffer correctly
- Confirmed the added read is negligible in practice: it's one command invocation per call site (`ps` and friends are single-shot, not looped over many candidate strings — that's a separate code path in the string-search subsystem), so the extra `malloc`+`rz_io_nread_at` only happens once per user command, not in any hot loop
- `clang-format -style=file` reports no changes to the touched file

## Environment

- macOS 27.0 (arm64), rizin built from `dev` at `5082af3230d`
- Confirmed the exact same fix and test pass under `meson --buildtype=debug`


librz/core/cmd/cmd_print.c
```diff
@@ -1745,14 +1745,52 @@ static void core_print_raw_buffer(RzStrStringifyOpt *opt) {
...
+				int n = rz_io_nread_at(core->io, core->offset + offset, grown, (size_t)max_length);
+				if (n > (int)length) {
+					buffer = grown;
+					length = (ut32)n;
+				} else {
+					RZ_FREE(grown);
+				}
+			}
+		}
+	}
```
wargio, line 1790: this should go into a function.

librz/core/cmd/cmd_print.c
```diff
@@ -1745,14 +1745,52 @@ static void core_print_raw_buffer(RzStrStringifyOpt *opt) {
 
 static RzCmdStatus core_print_string_in_block(RzCore *core, bool stop_at_nil, bool stop_at_unprintable, ut32 offset, RzOutputMode mode, RzStrEnc str_encoding) {
 	const ut8 *buffer = core->block + offset;
-	const ut32 length = core->blocksize - offset;
+	ut32 length = core->blocksize - offset;
```
wargio, line 1748: also this might not be a good idea. use `size_t`

librz/core/cmd/cmd_print.c
```diff
@@ -1745,14 +1745,52 @@ static void core_print_raw_buffer(RzStrStringifyOpt *opt) {
...
+	// this many bytes" is respected rather than overridden. There is no
+	// flag distinguishing "ambient default" from "explicit override" at
+	// this point (core->tmpseek covers seek but not blocksize overrides),
+	// so the default-value check is the only available signal.
+	if ((stop_at_nil || stop_at_unprintable) && core->blocksize == RZ_CORE_BLOCKSIZE) {
+		ut64 max_length = core->bin->str_search_cfg.max_length;
+		if (max_length > length && max_length <= UT32_MAX) {
+			grown = malloc((size_t)max_length);
+			if (grown) {
+				int n = rz_io_nread_at(core->io, core->offset + offset, grown, (size_t)max_length);
```
wargio, line 1781: if you use it as size_t everywhere, just declare it as such

wargio (changes requested):
remove the comments.

wargio:
you cannot remove the template; please read the guidelines on AI/LLM usage

wargio:
Also closed because it is forbidden to use AI for `good first issue` tagged issues

PR #6723, opened 2026-09-09
core: add gadget.bytes.length to limit table-mode gadget bytes column

### LINKED ISSUE #6359: Gadget Search: The printed bytes length should be limited by an evar
In Table Mode of Gadget search, the bytes length should be limited by an evar.

Current Output for `/Rt`

```
❯ rizin malloc://32
[0x00000000]> e asm.arch=arm
[0x00000000]> e asm.bits=32
[0x00000000]> e cfg.bigendian=false
[0x00000000]> wx 04e02de50000a0e10110a0e10080bd081eff2fe10220a0e100f020e3
[0x00000000]> /Rt
      addr                                    bytes disasm                                                                      
――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――
0x00000000 04e02de50000a0e10110a0e10080bd081eff2fe1 str lr, [sp, -4]!; ; mov r0, r0; ; mov r1, r1; ; ldmeq sp!, {pc}; ; bx lr; 
0x00000004         0000a0e10110a0e10080bd081eff2fe1 mov r0, r0; ; mov r1, r1; ; ldmeq sp!, {pc}; ; bx lr; 
0x00000008                 0110a0e10080bd081eff2fe1 mov r1, r1; ; ldmeq sp!, {pc}; ; bx lr; 
0x0000000c                         0080bd081eff2fe1 ldmeq sp!, {pc}; ; bx lr; 
0x00000010                                 1eff2fe1 bx lr; 
```

Instead use a config variable for length of bytes shown (with a suitable default)
Truncate the bytes when the length is greater than that variable 
And show ellipsis (...) for truncated output.

kisnner26:
Fixes #6359.

## The problem

`/Rt` (Table Mode gadget search) prints the full accumulated hex bytes of the whole gadget chain in the "bytes" column, with no way to cap it. Long chains produce a very wide, hard-to-read column:

```
❯ rizin malloc://32
[0x00000000]> e asm.arch=arm; e asm.bits=32; e cfg.bigendian=false
[0x00000000]> wx 04e02de50000a0e10110a0e10080bd081eff2fe10220a0e100f020e3
[0x00000000]> /Rt
      addr                                    bytes disasm
――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――
0x00000000 04e02de50000a0e10110a0e10080bd081eff2fe1 str lr, [sp, -4]!; ; mov r0, r0; ; mov r1, r1; ; ldmeq sp!, {pc}; ; bx lr;
0x00000004         0000a0e10110a0e10080bd081eff2fe1 mov r0, r0; ; mov r1, r1; ; ldmeq sp!, {pc}; ; bx lr;
...
```

## The fix

Adds a `gadget.bytes.length` config var, `0` by default (no limit, matching the existing convention used by `search.maxhits` and `asm.flags.maxname`). When set to a positive value, the "bytes" column of table-mode output is truncated to that many bytes with a trailing ellipsis, via the existing `rz_str_trunc_ellipsis()` helper — no new truncation logic, reusing what `agraph.c` already uses for the same kind of thing.

```
[0x00000000]> e gadget.bytes.length=4
[0x00000000]> /Rt
      addr    bytes disasm
――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――
0x00000000  04e0... str lr, [sp, -4]!; ; mov r0, r0; ; mov r1, r1; ; ldmeq sp!, {pc}; ; bx lr;
0x00000004  0000... mov r0, r0; ; mov r1, r1; ; ldmeq sp!, {pc}; ; bx lr;
0x00000008  0110... mov r1, r1; ; ldmeq sp!, {pc}; ; bx lr;
0x0000000c  0080... ldmeq sp!, {pc}; ; bx lr;
0x00000010 1eff2fe1 bx lr;
```

(A row whose bytes already fit within the limit, like the last one, is left untouched rather than padded — nothing to truncate there.)

## Scope

- Only `RZ_OUTPUT_MODE_TABLE` is touched, at the single `rz_table_add_rowf()` call site in `gadget.c` that builds the "bytes" column. JSON mode (`/Rj`) doesn't have a bytes field for hits and is unaffected; quiet mode (`/R`) prints each instruction's own hex separately (not the accumulated chain) via a different code path and is also unaffected — verified both manually.
- `gadget.bytes.length=0` (the default) reproduces the original output byte-for-byte — checked by diffing against pre-patch output.
- The context field is populated once in `rz_core_gadget_search_context_new()` alongside the other `gadget.*` config reads, following the existing pattern rather than querying config from inside `gadget.c`.

## Testing

- Reproduced the issue's own example exactly (same bytes, same `/Rt` command) to confirm the bug and then the fix
- Added two `test/db/cmd/cmd_rop` cases: one with `gadget.bytes.length=4` showing the truncated table, one with `=0` showing the full table is unchanged
- Both pass locally (`rz-test`)
- `clang-format -style=file` reports no changes needed on the touched files
- Full `ninja` build, no new warnings


Rot127:

PR #6730, opened 2026-09-12
Fix tms320 tests on s390x

kazarmy:

**Detailed description**

This pr fixes the tms320 tests on s390x.

**Test plan**

The Travis S390X build is green.

**Closing issues**

...


kazarmy:
btw, i think that `asm.arch` and `analysis.arch` should automatically set `cfg.bigendian`

wargio (approved):

wargio:
they will in RzArch, and technically they do when you switch arch, but only if current endianness cannot be set with the new arch.

notxvilka (approved):
Some architectures support both BE and LE, that's why in some cases, like this one, it's important to specify the endianness in tests.

kazarmy:
There should be one default endianness for each arch like `tms320`, rather than the endianness being dependent on the host machine's endianness, because the latter is somewhat unexpected.

wargio:
there is no concept of "default endianness" unless it is the host one.

Also according to the plugin, the arch support both https://github.com/rizinorg/rizin/blob/dev/librz/arch/p/asm/asm_tms320.c#L161

PR #6739, opened 2026-09-13
arch/lanai: remove global opcode table sort, make it per-instance

Yashwin-i:
- [x] I made sure to follow the project's coding style.
- [ ] I've documented every `RZ_API` function and struct this PR changes. (N/A - `lanai_dis_context_init`/`fini` are internal, marked `RZ_IPI`, not part of the public API)
- [ ] I've added tests that prove my changes are effective (required for changes to `RZ_API`). (N/A - no RZ_API changes; verified manually, see Test plan)
- [ ] I've updated the Rizin book with the relevant information (if needed). (N/A - internal refactor, no user-facing behavior change)
- [ ] I've used AI tools to generate fully or partially these code changes and I'm sure the changes are not copyrighted by somebody else. (Used Claude (Anthropic) to help design and write the struct/context refactor after diagnosing the thread-safety issue; reviewed, tested, and iterated on every change myself, including addressing automated review feedback.)

**Detailed description**

Fixes the Lanai portion of #4055, following up on the thread-safety discussion in #6673.

The only non-const static in `lanai-dis.c` was `opcodes_sorted`, guarding a one-time `qsort()` of `lanai_opcodes[]`. That array is declared non-static in `lanai-opc.c`, so it's shared across every plugin instance - sorting it in place on first use is a data race if two threads hit it concurrently.

Each `LanaiContext` now allocates and sorts its own private copy of the opcode table once, in `init()`. The original `lanai_opcodes[]` global (and `bfd_lanai_num_opcodes`) are now `const`, per review - no lock needed, since nothing shared is ever mutated after program start.

A new header, `lanai_ctx.h`, shares the `LanaiContext` struct between `asm_lanai_gnu.c` and `lanai-dis.c` (it forward-declares `struct lanai_opcode` rather than including the full header a second time, since the vendored `common_gnu/opcode/lanai.h` doesn't have an include guard and double-including it causes redeclaration errors).

`is_delayed_branch()` now takes the opcode table as a parameter instead of reaching for the global directly.

`lanai_dis_context_init()` now returns `bool` and propagates allocation failure up through `lanai_gnu_init()`, which frees the context and fails plugin init cleanly instead of leaving a NULL table to be dereferenced later. Also replaced the `qsort` comparator's function-pointer cast (technically UB - casting a `char *`-taking function to a `const void *`-taking one) with a small, correctly-typed wrapper.

**Test plan**

Built normally and confirmed the plugin still disassembles correctly:

    meson setup build && ninja -C build
    ./build/binrz/rizin/rizin -a lanai -
    [0x00000000]> pd 4
                0x00000000      sub.f pc , 0xffff , r30
                0x00000004      punt
                0x00000008      punt
                0x0000000c      punt

Rebuilt with AddressSanitizer and repeated the same test, no errors reported:

    meson setup build-asan -Db_sanitize=address
    ninja -C build-asan
    ./build-asan/binrz/rizin/rizin -a lanai -
    [0x00000000]> pd 4
    [0x00000000]> exit

Re-verified after the OOM-safety changes - output unchanged.

**Closing issues**

Related to #4055 (Lanai only - the broader tracking issue covers many other files and isn't closed by this PR). Follows from discussion in #6673.


Yashwin-i:
@Rot127 what say?

librz/arch/isa_gnu/lanai/lanai_ctx.h
```diff
@@ -0,0 +1,20 @@
...
+
+typedef struct {
+	unsigned long Offset;
+	RzStrBuf *buf_global;
+	unsigned char bytes[4];
+	struct lanai_opcode *opcodes;
+} LanaiContext;
+
+void lanai_dis_context_init(LanaiContext *ctx);
+void lanai_dis_context_fini(LanaiContext *ctx);
```
Rot127: Please add type annotations.
Also those functions should be RZ_IPI.

Rot127 (changes requested):
Yes, looks good.

Please make `lanai_opcodes` `const`.
And `bfd_lanai_num_opcodes` directly a macro.

Yashwin-i:
@Rot127 the macro doesn't work here - `NUMOPCODES` is used in `lanai-dis.c`, which only sees the `extern` declaration, not the full array definition. Since `lanai_opcodes[]` is an incomplete type there, `sizeof` can't be used on it.

It only works inside `lanai-opc.c`, where the array is actually defined.

The only way I see to make `bfd_lanai_num_opcodes` a macro and have it usable from `lanai-dis.c` as well would be to give the array a fixed size in the header (`lanai_opcodes[N]` instead of `lanai_opcodes[]`), so the size is known at every declaration.

Is that what you meant, or is there another approach you'd prefer for this one?

Rot127:
> Is that what you meant, or is there another approach you'd prefer for this one?

Sorry, what I actually was thinking (but didn't spell it out) that the table should be moved to a header file.
But given that the code here is from the 90s it doesn't matter. Just make the table and the count const and done.

Yashwin-i:
@Rot127 done -  `lanai_opcodes` and `bfd_lanai_num_opcodes` are now `const`, and `lanai_dis_context_init` / `fini` have `RZ_IPI` + `RZ_NONNULL` annotations. Pushed the changes.

Good to take this out of draft, or is there anything else you'd like me to change first?

Rot127 (approved):

Yashwin-i:
Addressed all three of Copilot's findings - I had overlooked the OOM path on `malloc` failure.

`lanai_dis_context_init` now returns `bool`, `lanai_gnu_init` bails out and frees `ctx` if initialization fails, and I replaced the `qsort` comparator cast with a properly typed wrapper function.

Pushed the changes.

Yashwin-i:
@Rot127 dug into the logs a bit more - both failures are in `unicode` / `unicode/utf16` / `unicode/utf8` and a few unrelated `db/` corpus tests, all with exit status `-1` (killed, rather than a mismatch).

The Meson unit tests (`137/137`) and integration tests (`20/20`) both pass cleanly; it's specifically the `rz-test` step. Nothing in the failing groups touches Lanai or anything changed by this PR.

This looks more like memory pressure on the ASAN job while running the full corpus in parallel than a regression from my changes, but I can't rerun the job myself to confirm whether it's transient. Let me know if this is a known/expected issue with this suite.

Separately, a heads-up: I'll be pretty limited on availability for the next week, so I might be a bit slow to respond if anything else comes up here. Not gone, just slower for a bit.

Rot127:
The red tests are unrelated

Rot127:
@Yashwin-i Please fix the AI comments btw. At least the ones which are correct.
Then resolve the comments.

Issue #6751, opened 2026-09-16
tms320c6x: compact (16-bit) instructions mis-decoded as 32-bit / `.fphead` when the caller's buffer doesn't reach the fetch-packet header


notxvilka:
## Summary

On the TMS320C6x decoder (`tms320` arch, `c674x`/`c64x`/`c66x` CPUs), a 16-bit **compact** instruction is only decoded as compact when the 32-byte fetch packet's **header word** (packet offset `0x1c`) happens to lie within the byte buffer the caller passed to `c6x_decode()`. When it doesn't — single-instruction disassembly (`rz-asm -d`), or `pd`/`rz_asm_disassemble` at the tail of a mapped region — the function silently falls through to the 32-bit path and:

* returns the **wrong instruction**, and
* returns **size 4 instead of 2**, which desyncs everything after it, and
* for any word whose top nibble is `0xe` it emits a spurious **`.fphead`** (a fetch-packet *header*, i.e. metadata, in the middle of real code).

The RzAsm plugin calls `c6x_decode()` directly with whatever length it is handed, so it hits this. The RzAnalysis plugin does **not** — it reconstructs the whole fetch packet first — so `pd` (asm) and `ao`/analysis can disagree on the same address, and the asm output is buffer-length-dependent.

Found while loading a real TMS320C674x firmware image and cross-checking the disassembly. The decoder is otherwise solid: it passes all 211 vectors in `test/db/asm/tms320_c674x_32`. This is specifically about the length/context handling on the RzAsm entry path.

## Version

* rizin `dev` @ `6dd01b84f472734e8efa193bb64d0818d8257126`
* arch `tms320`, `asm.cpu=c674x`, `asm.bits=32`

## Reproducer (`wx`)

The 32 bytes below are one real C674x compact fetch packet. Its header word is `0xe0208000` (top nibble `0xe`); the layout field (bits 27:21) is `0x01`, meaning **only word 0 holds two 16-bit compact instructions** (`0x0246` at `+0`, `0xeaba` at `+2`); words 1–6 are ordinary 32-bit instructions; word 7 is the header.

### Correct — whole fetch packet is present (header reachable)

```
$ rizin -a tms320 -c c674x -b 32 malloc://0x20
[0x00000000]> wx 4602baea2874ca01e805e001740280012ae05002ea05600276220002008020e0
[0x00000000]> pd 4
0x00000000      0246  mv .l1 a4,a0
[...]

PR #6755, opened 2026-09-16
Fixes for the TI TMS320C6x compact mode

### LINKED ISSUE #6752: tms320c6x: branch targets are twice the correct distance inside a compact fetch packet
A `B`/`BNOP` sitting in a compact (16-bit) fetch packet resolves to double its
real displacement, on both the RzAsm and the RzAnalysis path. In an ordinary
32-bit packet the same instructions are correct, so this is specific to compact
packets — which is most of the code in a real C674x image.

The 32 bytes below are one fetch packet emitted by `cl6x` 7.4.4, the official TI
compiler, from `int sum(int *p, int n) { int i, s = 0; for (i = 0; i < n; i++) s += p[i]; return s; }`
built with `cl6x -mv6740 --abi=eabi -O2 -k -c`. It is a software-pipelined loop;
ten of its twelve instructions are compact and word 7 is a real fetch-packet
header. `dis6x` from the same toolchain resolves the branch at `0x04` to
`$C$L4`, the loop exit at `0x18`, twenty-four bytes along:

```
00000004   300ca121    [!B0]  BNOP.S1       $C$L4 (PC+24 = 0x00000018),5
00000018       81ef           BNOP.S2       B3,4
0000001c   ef200401           .fphead       n, l, W, BU, nobr, nosat, 1111001b
```

Rizin decodes every slot and every size in that packet identically to `dis6x` —
the compact pairs, the parallel bits and the header all agree — but puts the
branch target at `0x30` instead of `0x18`:

```
rizin -N -qc 'e asm.arch=tms320; e asm.cpu=c674x; e cfg.bigendian=false
wx 2712a60621a10c30a2039026660c3c0c6e4c6e0c675cd062ef81c682010420ef
pd 5; s 4; ao~^jump:' malloc://64

            0x00000000      cmplt .l2 0x0,b4,b0
            0x00000002      ||    mvk .l1 0x0,a5
        ,=< 0x00000004      [!b0] bnop .s1 0x30,0x5     <-- dis6x says 0x18
        |   0x00000008      ||    [b0] mvc .s2 b4,ilc
        |   0x0000000c      sploop 1
jump: 0x00000030                                        <-- analysis agrees with itself, not dis6x
```

An ordinary packet is fine, which is what scopes this. From `void g(int x) { if (x) f(); }`
at `-O0`, where nothing is compact, rizin matches `dis6x` exactly — `PC+20 = 0x14`
and `PC+0 = 0x0`:

```
rizin -N -qc 'e asm.arch=tms320; e asm.cpu=c674x; e cfg.bigendian=false
wx d80f1000900200d0100000c062638cd000000000
pd 4' malloc://64

            0x00000000      mv    .l1 a4,a0
        ,=< 0x00000004      [!a0] b .s1 0x14            <-- correct
        |   0x00000008      [a0]  b .s1 0x0             <-- correct
        |   0x0000000c      [!a0] bnop .s2 b3,0x3
```

The factor is exactly two in the case above, which points at the base the
displacement is added to rather than the displacement itself: a C6000 branch is
relative to the fetch packet address (PCE1), and in a compact packet the slots
are two bytes apart rather than four, so a base computed from a slot index times
four lands twice as far along. I have not confirmed that is the mechanism — it is
where I would look first. The affected code is the `PCE1`/displacement handling
around `librz/arch/isa/tms320/c6x/c6x_decode.c:1185` (`sext(BITS(w, 27, 7), 21) * 4`)
and whatever supplies the packet base to it.

notxvilka:
# DO NOT SQUASH

**Detailed description**

- Fix branch decoding in some cases inside the compact-encoded instruction packet
- Partially fix the the misdecoded compact (16-bit) instructions as full (32-bit) when in the middle of the packet but the decoding starts not from the packet start.

**Test plan**

CI is green

**Closing issues**

- Closes https://github.com/rizinorg/rizin/issues/6752
- Partially addresses https://github.com/rizinorg/rizin/issues/6751 (full fix requires a lookahead support for the RzArch/RzAsm for packet-aware decoding, which requires RzAsm API changes).


test/db/analysis/tms320.c6x_32
```diff
@@ -534,3 +534,54 @@ EXPECT=<<EOF
...
+NAME=c6x compact fetch packet from cl6x 7.4.4 (-mv6740 -O2)
+FILE=malloc://64
+CMDS=<<EOF
+e asm.arch=tms320
+e cfg.bigendian=false
+e asm.cpu=c674x
+e scr.color=0
+e asm.lines=false
+e asm.bytes=false
+wx 2712a60621a10c30a2039026660c3c0c6e4c6e0c675cd062ef81c682010420ef
```
wargio: these two tests can probably be merged.

Rot127 (approved):

PR #6756, opened 2026-09-16
Preserve empty Base32 encoding without leaking memory

Shubham-Padkonde:

**Detailed description**

Encoding empty input with `rz_base32_encode_dyn` returns `NULL` and leaks the allocated one-byte buffer. The wrapper mistakes the encoder's valid zero-length result for an error.

Keep the allocated, NUL-terminated empty result, as for other valid inputs. Document this behavior and add a regression test that frees the result.

AI disclosure: Codex prepared the code, regression test, and description. The exact model identifier is not exposed. Contributor checklist attestations remain for the submitting human to verify.

**Test plan**

```sh
meson setup build -Denable_tests=true
meson test -C build base32 --print-errorlogs
```

The new empty-input test fails on the original implementation and passes with the fix; all seven Base32 unit tests pass. Changed C lines pass clang-format 20 formatting.

This should be submitted as a draft until upstream CI is green, as required for new contributors. Local tests do not replace that requirement.

**Closing issues**

None. Searches for Base32 empty-input issues and PRs found no existing report.


librz/util/base32.c
```diff
@@ -210,9 +211,7 @@ RZ_API RZ_OWN char *rz_base32_encode_dyn(RZ_NONNULL const ut8 *src, size_t n) {
 	if (!out) {
 		return NULL;
 	}
-	if (rz_base32_encode(out, src, n) == 0) {
-		return NULL;
-	}
+	rz_base32_encode(out, src, n);
```
wargio: instead of this we should check for an empty string and produce an empty string.

Shubham-Padkonde: Changed this to an explicit 
 == 0` early return of an allocated empty string. The input is binary, so the check uses its length rather than `src[0]`: one NUL byte must still encode as `AA======`.

The encoder error check is restored for nonempty input, with `free(out)` before returning `NULL`. The regression also covers a nonempty buffer passed with length zero and a one-byte binary zero.

Implemented in a0cb255. All seven Base32 unit tests pass; changed C lines pass clang-format 20. The local Meson build used the pinned dependency revisions, with subproject checking disabled because its checker compares the libmspack v1.11 tag text to the detached commit hash. The full Rizin test suite was not run.

librz/util/base32.c
```diff
@@ -205,12 +206,16 @@ RZ_API size_t rz_base32_encode(RZ_OUT RZ_NONNULL char *dest, RZ_NONNULL const ut
  */
 RZ_API RZ_OWN char *rz_base32_encode_dyn(RZ_NONNULL const ut8 *src, size_t n) {
 	rz_return_val_if_fail(src, NULL);
+	if (!n) {
+		return rz_str_dup("");
+	}
```
wargio:
```suggestion
	if (n < 1) {
		return rz_str_dup("");
	}
```

Shubham-Padkonde: Applied the suggested `n < 1` guard in 4f63124. The Base32 unit target passes, including the empty-input and binary-zero regressions.

wargio:
is this PR linked to an issue?

Shubham-Padkonde:
No, there is no separate issue linked to this PR. The original issue search did not find an existing report, and the description lists “Closing issues: None.” The problem was found while reviewing `rz_base32_encode_dyn`: its empty-input path returned `NULL` and leaked the allocation. The PR includes the reproducing regression and the explicit empty-input handling adjusted in response to your reviews. If you prefer a separate tracking issue, I can add one with that reproduction.

This reply was prepared with Codex/GPT-6 assistance.

Issue #5728, opened 2026-01-04
Add missing header (`iH`) information to some bin formats


notxvilka:
Currently, not all formats create header information in the `RzStructuredData` format that can be printed with the `iH` command. 

> [!IMPORTANT]
> Each bin requires its own PR, **DO NOT ATTEMPT TO OPEN A PR FIXING MULTIPLE PLUGINS**

The missing are:

- [ ] ART
- [x] AVR (maybe just some entry point, cpu and other useful info?)
- [ ] BF (maybe some stats & file size ?)
- [x] bFLT
- [ ] BIOS (like AVR?)
- [ ] CGC (reuse elf)
- [x] DEX
- [ ] JAVA (see java core plugin)
- [x] LUAC (https://github.com/rizinorg/rizin/pull/5555)
- [ ] MDMP
- [x] MENUET
- [ ] NES
- [ ] NIN3DS
- [ ] NINDS
- [x] NINGB (https://github.com/rizinorg/rizin/pull/5926)
- [ ] NINGBA
- [ ] NRO
- [ ] NSO
- [x] OMF
- [ ] P9
- [ ] PEBBLE
- [ ] PRC
- [ ] PSXEXE
- [x] PYC
- [x] SMD (https://github.com/rizinorg/rizin/pull/5929)
- [ ] SPC700
- [ ] TE
- [ ] VSF
- [x] WASM
- [x] XBE
- [x] ZIMG

aviralgarg05:
I would like to solve the issue, please assign me

Rot127:
You can just work on them. No need for assignment.

Ayushd785:
I am now working on the PYC format.

Ayushd785:
Working on AVR

Ayushd785:
working on BF

PR #6764, opened 2026-09-18
bin/psxexe: expose header info via iH (bin_structure)

kisnner26:
part of #5728 (psxexe)

per the issue's rule, this PR touches only the `psxexe` plugin.

`psxexe` didn't implement `bin_structure`, so `iH` printed nothing for PS-X EXE binaries even though the plugin already parses the full header for `sections`/`entries` (`t_addr`/`t_size`, `pc0`, etc.).

Adds `bin_structure()` following the same pattern as the completed SMD/NES-style plugins: reuses the existing endian-safe `rz_buf_fread_at(bf->buf, 0, (ut8 *)&psxheader, "8c17i", 1)` read already used in `sections()`/`entries()`, and exposes every field of `psxexe_header` (entry point, gp/text/data/bss/stack addresses and sizes, saved exception-handler registers) as a `psxexe` map via `RzStructuredData`.

Added a `test/db/formats/psx` case for `iH` against the existing real test binary (`bins/psx/HITPSX.EXE`).

Verified locally: built with meson/ninja, ran `iH` against `bins/psx/HITPSX.EXE` — values match what `ia`/`iS` already report for the same file (`t_addr`/`t_size`, entry point). Ran `rz-test` on `test/db/formats/psx`: 2/2 pass.


Rot127:

PR #6765, opened 2026-09-18
bin/p9: expose header info via iH (bin_structure)

kisnner26:
part of #5728 (p9)

per the issue's rule, this PR touches only the `p9` plugin.

`p9` didn't implement `bin_structure`, so `iH` printed nothing for Plan9 executables even though the plugin already parses the full header for `sections()`/`size()` (text/data/syms/spsz sizes at fixed offsets).

Adds `bin_structure()`, reading the header with the same endianness convention already used elsewhere in this exact file: `magic` big-endian (matching `rz_bin_p9_get_arch`), the rest of the header (`text`/`data`/`bss`/`syms`/`entry`/`spsz`/`pcsz`) little-endian (matching `sections()`/`size()`). Exposes all fields as a `p9` map via `RzStructuredData`.

Added a `test/db/formats/plan9` case for `iH`.

Verified locally: built with meson/ninja, ran `iH` against `test/bins/plan9/bin` — `text`/`data` values match what `iS` already reports for the same file, and `magic` (0x1eb) matches `I_MAGIC`, consistent with `iI` reporting x86. Ran `rz-test` on `test/db/formats/plan9`: 2/2 pass.


Rot127:

PR #6766, opened 2026-09-18
bin/pebble: expose header info via iH (bin_structure)

kisnner26:
part of #5728 (pebble)

per the issue's rule, this PR touches only the `pebble` plugin.

`pebble` didn't implement `bin_structure`, so `iH` printed nothing for Pebble watch apps even though the plugin already parses and keeps the full `PebbleAppInfo` header in `bf->o->bin_obj` for `info()`/`sections()`/`entries()` (name, company, versions, reloc table, entry point).

Adds `bin_structure()`, exposing every `PebbleAppInfo` field (header sentinel, struct/sdk/app versions, size, entry offset, CRC, name, company, icon resource id, symbol table address, flags, reloc table info, uuid) as a `pebble` map via `RzStructuredData`. No new parsing — it just reads the already-persisted struct.

Added a `test/db/formats/peeble` case for `iH`.

Verified locally: built with meson/ninja, ran `iH` against `test/bins/pebble/pebble-app.bin` — prints the real app metadata (name "Pedometer", company "Jathusan Thiruchelvanathan"), consistent with what `iI` already reports for the same file. Ran `rz-test` on `test/db/formats/peeble`: 2/2 pass.


Rot127:

